PR #224, opened 2019-01-30
Implement settable timeouts in Functions.

bklimt:
1. Update okhttp to a newer version.
2. Change the default timeout for Functions to 60 seconds, to match the backend.
3. Add methods to change the timeout for a Function.
4. Ensure all exceptions coming from Functions are wrapped in FirebaseFunctionsException.


bklimt:
These oss-bot failures don't look like they have anything to do with my changes.

bklimt:
This was implemented with okhttp 3, because the newer version: 1) lets us specify a single timeout for the whole call, instead of separate timeouts for connect, read, and write, and 2) lets us easily set timeouts for particular calls, rather than having to create a new client for each callable.

https://github.com/square/okhttp/wiki/Recipes#timeouts

The downside is this may increase library size for developers using both Firestore and Functions, since they will no longer share the same okhttp version.

CodingDoug:
Is 60s the best number on the client?  If that's measured by the moment the client attempts to connect, to the moment it receives the response, the function on the server may not actually have its own full 60s to yield a response, since the client could also be waiting for network latency and such.

bklimt:
60s is what had been agreed upon for the API, and it's a big improvement over the previous, unspecified default. But I agree it would be reasonable to have a longer default to account for connection time. We should make the default the same across all client SDKs. What length default timeout would you propose?

CodingDoug:
I'd guess that a buffer of 5 extra seconds is enough to cover 98% of the connections out there.

bklimt:
After some discussion offline, I've upped the timeout to 70s. Thanks for the feedback @CodingDoug.

firebase-functions/firebase-functions.gradle
```diff
@@ -55,7 +55,7 @@ dependencies {
     }
     implementation 'com.google.firebase:firebase-iid-interop:16.0.1'
 
-    implementation 'com.squareup.okhttp:okhttp:2.7.5'
+    implementation 'com.squareup.okhttp3:okhttp:3.12.1'
```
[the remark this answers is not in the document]
ashwinraghav: I might try to revive some of those bugs on grpc

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallOptions.java
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.functions;
+
+import java.util.concurrent.TimeUnit;
+import okhttp3.OkHttpClient;
+
+/** An internal class for keeping track of options applied to an HttpsCallableReference. */
+class HttpsCallOptions {
```
ashwinraghav: Question: Are there some functions that could take more time?
Should this be configurable/part of public API?

ashwinraghav: NVM. I see it down there :)

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.java
```diff
@@ -89,6 +94,28 @@
...
+    return functionsClient.call(name, null, options);
+  }
+
+  /**
+   * Changes the timeout for calls from this instance of Functions. The default is 60 seconds.
+   *
+   * @param timeout The length of the timeout, in the given units.
+   * @param units The units for the specified timeout.
+   */
+  public void setTimeout(long timeout, TimeUnit units) {
```
ashwinraghav: Does this require an API review?

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -243,17 +244,28 @@ public void useFunctionsEmulator(String origin) {
       request = request.header("Firebase-Instance-ID-Token", context.getInstanceIdToken());
     }
 
+    OkHttpClient callClient = options.apply(client);
```
ashwinraghav: Question: Do we create a client and apply these options to create a new Client for some reason?
I am wondering whether we can reduce state by just saying options.apply(new OKHttpClient()) or possibly options.createClient()

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.java
bklimt: Yes, the API has been approved. go/callable-timeouts

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
bklimt: The way `options.apply()` is implemented is the recommended way to do it. To change the timeout, you have to create a new client from the existing client, but they share most of their state, so it's not as expensive as it looks. The okhttp API is admittedly not obvious.

https://github.com/square/okhttp/wiki/Recipes#per-call-configuration

bklimt: As to why we don't just create a new one every time: doing it this way lets us share one connection pool and dispatcher across all of the function references for an app.

firebase-functions/firebase-functions.gradle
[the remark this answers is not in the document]
bklimt: > would you be open to picking that back up?

I'm open to it, but we'd have to discuss its priority relative to the other things I'm working on.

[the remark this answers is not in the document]
bklimt: From the comments on those grpc issues, it sounds like maybe they haven't considered the bloat from developers also using okhttp3 as a reason for them to update their dependency. Perhaps we could follow up and suggest that as a reason to reopen the issue.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
ashwinraghav: Thanks for clarifying

PR #231, opened 2019-02-04
Adjust sqlite size calculation to account for WAL

rsgowman:
SQLite supports two methods of commiting; a rollback journal and a
write-ahead-log (WAL). By default, a rollback journal is used. However,
on Android 9 (P), the default changed slightly to a WAL-compatible mode:
https://source.android.com/devices/tech/perf/compatibility-wal. This
mode causes a WAL to be written to disk.

Our prior method of calculating the database size did not take the WAL
into account. In practice, this isn't too awful, as the contents of the
WAL would eventually be written back into the main database file, but in
the meantime, our calculation could be off by upwards of a few MB. This
commit changes the size calulation to ask sqlite rather than looking at
the size of the file on disk.

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Unreleased
+- [fixed] Fixed calculation of SQLite database size on Android 9 (P) devices.
+  Previous method could be off by a few MBs on these devices, potentially
+  delaying garbage collection.
 
 # 18.0.1
 - [fixed] Fixed an issue where Firestore would crash if handling write batches
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-- [fixed] Fixed calculation of SQLite database size on Android 9 (P) devices.
+- [fixed] Fixed calculation of SQLite database size on Android 9 Pie devices.
   Previous method could be off by a few MBs on these devices, potentially
   delaying garbage collection.
 
--- a/firebase-firestore/firebase-firestore.gradle
+++ b/firebase-firestore/firebase-firestore.gradle
@@ -83,6 +83,7 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    testOptions.unitTests.includeAndroidResources = true
 }
 
 dependencies {
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
@@ -34,7 +34,6 @@ import com.google.firebase.firestore.model.DatabaseId;
 import com.google.firebase.firestore.util.Consumer;
 import com.google.firebase.firestore.util.Logger;
 import com.google.firebase.firestore.util.Supplier;
-import java.io.File;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.ArrayList;
@@ -77,7 +76,6 @@ public final class SQLitePersistence extends Persistence {
   private final OpenHelper opener;
   private final LocalSerializer serializer;
   private SQLiteDatabase db;
-  private File databasePath;
   private boolean started;
   private final SQLiteQueryCache queryCache;
   private final SQLiteRemoteDocumentCache remoteDocumentCache;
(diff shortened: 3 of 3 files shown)


rsgowman:
cc @ashwinraghav

firebase-firestore/firebase-firestore.gradle
```diff
@@ -83,6 +83,7 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    testOptions.unitTests.includeAndroidResources = true
```
mikelehen: This is technically tangential to your PR, right?  Please confirm you're including it intentionally.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Unreleased
+- [fixed] Fixed calculation of SQLite database size on Android 9 (P) devices.
```
wilhuff: "Android 9 (P)" is a nonstandard way of referring to this. Android documentation calls this "Android 9 Pie". We should follow since the changelog gets rolled up into our public release notes.

firebase-firestore/firebase-firestore.gradle
wilhuff: Also, please confirm that including this flag now does not generate a warning.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -199,7 +196,40 @@ void runTransaction(String action, Runnable operation) {
...
+  private long getPageCount() {
+    return pragma("page_count");
+  }
+
+  /**
+   * @param pragmaName Must be a valid 'pragma-name' and must refer to a long value.
+   * @see https://www.sqlite.org/pragma.html
+   */
+  private long pragma(String pragmaName) {
+    Cursor cursor = db.rawQuery("PRAGMA " + pragmaName, null);
```
wilhuff: As a matter of principle we should avoid composing SQL through string concatenation. This construct is subject to SQL injection attacks if we're not careful. Better to just have the caller pass the whole query string.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -199,7 +196,40 @@ void runTransaction(String action, Runnable operation) {
...
+   */
+  private long getPageCount() {
+    return pragma("page_count");
+  }
+
+  /**
+   * @param pragmaName Must be a valid 'pragma-name' and must refer to a long value.
+   * @see https://www.sqlite.org/pragma.html
+   */
+  private long pragma(String pragmaName) {
```
wilhuff: Not all pragmas resolve to a single long value. Maybe name this simpleQueryForLong?

I suggest that name because for queries that resolve to a single 1x1 result, SQLite supports running them with `SQLiteStatement.simpleQueryForLong` in a way that avoids allocating a cursor.

firebase-firestore/CHANGELOG.md
rsgowman: Done.

firebase-firestore/firebase-firestore.gradle
rsgowman: > This is technically tangential to your PR, right? Please confirm you're including it intentionally.

Mostly tangential, though it causes the issue to manifest itself. (Also required by Ashwin's changes.)

> Also, please confirm that including this flag now does not generate a warning.

This does not seem to introduce any new warnings.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
rsgowman: No longer relevant.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
rsgowman: As discussed, it's not possible to use a prepared statement here. We could instead allow the caller to pass the *entire* sql statement, but that's probably worse. Instead, we decided to un-refactor this. (Done).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -199,7 +196,38 @@ void runTransaction(String action, Runnable operation) {
...
+    return getPageCount() * getPageSize();
+  }
+
+  /**
+   * Gets the page size of the database. Typically 4096.
+   *
+   * @see https://www.sqlite.org/pragma.html#pragma_page_size
+   */
+  private long getPageSize() {
+    Cursor cursor = db.rawQuery("PRAGMA page_size", null);
```
wilhuff: You could cut down on the verbosity here by writing this as:

```java
return query("PRAGMA page_size").firstValue(row -> row.getLong(0));
```

PR #235, opened 2019-02-05
Fix version in x-goog-api-client-value and logcat

rsgowman:
Was previously hardcoded. Now picked up from the gradle build.


rsgowman:
@vkryachko Note that the version doesn't seem to actually pickup the '-SNAPSHOT' portion; so even when I publish to a local maven repo, the version is still presented as 18.0.1 (rather than 18.0.1-SNAPSHOT).

vkryachko:
@rsgowman You're right, now looking at the code, the snapshot suffix is only added during maven artifact creation. Does this behavior work for you?

rsgowman:
@vkryachko Jonny mentioned that, while not necessarily ideal, it was fine. Being able to separate out dev instances is a secondary concern. Ideally, it would be nice to move toward that ideal, but I don't think we need to do that *now*.

That said, if there's something relatively easy on your end that you can do to make the -SNAPSHOT flow through in instances like this, that would be even better. (But I'd like to proceed with what we've got now regardless, as it's a clear improvement. And coincidentally easy, thanks to the work you did.)

vkryachko:
@rsgowman No easy solution comes to mind off the top of my head, I'll file a bug to track this

Issue #239, opened 2019-02-12
Using PropertyName value as enum value


wikes82:
## What feature would you like to see?
saving PropertyName annotation value for enum, instead of enum name

## How would you use it?

```
public enum MyEnum {
  @PropertyName("one")
   ONE("one"),
 @PropertyName("secondOne")
   SECOND_ONE("secondOne")
}
```
so, instead saving it into Firestore as ONE or SECOND_ONE,  we use PropertyName value of one or secondOne.

We generate our POJO using jsonschema2pojo, and it capitalized all enum names, see : https://github.com/joelittlejohn/jsonschema2pojo/wiki/Reference#enum

We use custom annotator to add PropertyName annotation into our POJO but can't change the way jsonschema2pojo behave on naming the enum name without modifying their source codes.

wilhuff:
This seems like a pretty good idea!

Unfortunately, we're completely booked on other work right now so if you're motivated to make this happen your best path forward would be to create a pull request that implements it.

The code that does the POJO mapping is in [CustomClassMapper](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java). In particular you'll see that when mapping:

  * from an object we're unconditionally just taking the [name of the enumeration member](https://github.com/firebase/firebase-android-sdk/blob/26e848e07812e86b83773c7376b1787023ab41e4/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java#L158)
  * to an object we're unconditionally [calling Enum.valueOf with the name](https://github.com/firebase/firebase-android-sdk/blob/26e848e07812e86b83773c7376b1787023ab41e4/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java#L322).

In the forward direction it should be possible to either use [annotatedName](https://github.com/firebase/firebase-android-sdk/blob/26e848e07812e86b83773c7376b1787023ab41e4/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java#L930) or do something similar.

The reverse direction is trickier but the straightforward thing that just iterates over the enumeration members to find a match is probably OK as a fallback from `Enum.valueOf`.

I'd start by forking the repo, and then following along with our [getting started instructions](https://github.com/firebase/firebase-android-sdk/#getting-started). Make sure `./gradlew :firebase-firestore:check` passes before making changes.

[...]

PR #240, opened 2019-02-13
Use PropertyName value for enum serialization/deserialization

wikes82:
Pull request for this issue : https://github.com/firebase/firebase-android-sdk/issues/239

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
@@ -156,17 +156,13 @@ public class CustomClassMapper {
     } else if (o.getClass().isArray()) {
       throw serializeError(path, "Serializing Arrays is not supported, please use Lists instead");
     } else if (o instanceof Enum) {
-      String enumField = ((Enum<?>) o).name();
+      String enumName = ((Enum<?>) o).name();
       try {
-        if (o.getClass().getField(enumField).isAnnotationPresent(PropertyName.class)) {
-          PropertyName annotation =
-              o.getClass().getField(enumField).getAnnotation(PropertyName.class);
-          return annotation.value();
-        }
-      } catch (Exception ex) {
-        return enumField;
+        Field enumField = o.getClass().getField(enumName);
+        return BeanMapper.propertyName(enumField);
+      } catch (NoSuchFieldException ex) {
+        return enumName;
       }
-      return enumField;
     } else if (o instanceof Date
         || o instanceof Timestamp
         || o instanceof GeoPoint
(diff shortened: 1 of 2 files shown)


wikes82:
I signed CLA

wikes82:
CLA signed

wikes82:
please check CLA again

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -324,6 +334,19 @@ public static Object convertToPlainJavaTypes(Object object) {
       String value = (String) object;
       // We cast to Class without generics here since we can't prove the bound
       // T extends Enum<T> statically
+
+      // try to use PropertyName if exist
+      Field[] enumFields = clazz.getFields();
+      for (Field field : enumFields) {
+        if (field.isAnnotationPresent(PropertyName.class)) {
```
wilhuff: This should reuse `propertyName` (or at least `annotationName`).

Also, not all fields in an enum class are necessarily enum constants, so I think you need to check `Field.isEnumConstant`. That would make the inner part of this loop look like this:

```java
if (field.isEnumConstant()) {
  String propertyName = propertyName(field);
  if (value.equals(propertyName)) {
    value = field.getName();
    break;
  }
}
```

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -709,7 +709,10 @@ public static void setValue1(String value1) {
 
   private enum ComplexEnum {
     One("one"),
-    Two("two");
+    Two("two"),
+
+    @PropertyName("three")
+    THREE("three");
```
wilhuff: Please make either the PropertyName or the value different here so there's no chance the test is accidentally succeeding.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -1974,18 +1979,22 @@ public void enumsAreSerialized() {
...
   public void enumsAreParsed() {
-    String json = "{'enumField': 'Bar', 'enumValue': 'Foo', 'complexEnum': 'One'}";
+    String json =
+        "{'enumField': 'Bar', 'enumValue': 'Foo', 'complexEnum': 'One', 'enumUsingPropertyName': 'three'}";
     EnumBean bean = deserialize(json, EnumBean.class);
     assertEquals(bean.enumField, SimpleEnum.Bar);
     assertEquals(bean.enumValue, SimpleEnum.Foo);
     assertEquals(bean.complexEnum, ComplexEnum.One);
+    assertEquals(bean.enumUsingPropertyName, ComplexEnum.THREE);
   }
```
wilhuff: Please add a test for a pathological case like this enum:

```java
enum PathologicalEnum {
  @PropertyName("Two")
  One,

  @PropertyName("One")
  Two,
}
```

wikes82:
not sure why those 2 tests failed

wilhuff:
Those aren't your fault. We've recently been having trouble with some of the tasks in those tests which has made them flaky. This submission wouldn't depend on getting those resolved.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
wikes82: I made this change.. PropertyName is Three and value is three

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
wikes82: added this test

PR #243, opened 2019-02-14
Add gradle task for generating JaCoCo coverage report for unit tests.

yifanyang:
This commit will add a gradle task to run the instrumented unit test and generate a jacoco coverage report.

The command for the gradle task is `./gradlew :<firebase-project>:coverage` for specific sdk or `./gradlew coverage` for all sdks.

Generated html report and xml report reside in `build/reports/jacoco/coverage` under each sdk directory.

### DIFF
--- a/root-project.gradle
+++ b/root-project.gradle
@@ -97,10 +97,10 @@ configure(subprojects) {
     }
     apply plugin: 'jacoco'
     jacoco {
-        toolVersion = '0.8.2'
+        toolVersion = '0.8.3'
     }
     task(['type': JacocoReport, 'dependsOn': 'check', 'group': 'Coverage',
-          'description': 'Generates JaCoCo unit test coverage reports.'], 'coverage') {
+          'description': 'Generates JaCoCo unit test coverage reports.'], 'checkCoverage') {
         def excludes = ['**/R.class', '**/R$*.class', '**/BuildConfig.*', '**Manifest*.*']
         classDirectories = files([
             fileTree(dir: "$buildDir/intermediates/javac/debug", excludes: excludes)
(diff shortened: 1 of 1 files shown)


yifanyang:
A sample report generated with the gradle task in this pr can be downloaded from [here](https://drive.google.com/file/d/14dWaY9da34JqkQHH82jt8bZ9M7w-QLZH/view?usp=sharing).

root-project.gradle
```diff
@@ -95,6 +95,27 @@ configure(subprojects) {
         exclude 'src/testUtil/java/com/google/firebase/firestore/testutil/ThrowingRunnable.java'
         exclude '**/package-info.java'
     }
+    apply plugin: 'jacoco'
+    jacoco {
+        toolVersion = '0.8.2'
```
allisonbm92: The latest version is 0.8.3. Can you see if this works with the latest?

root-project.gradle
```diff
@@ -95,6 +95,27 @@ configure(subprojects) {
...
+    }
+    task(['type': JacocoReport, 'dependsOn': 'check', 'group': 'Coverage',
+          'description': 'Generates JaCoCo unit test coverage reports.'], 'coverage') {
```
allisonbm92: It might be better to name this checkCoverage, as this is the coverage of the tests run by the check task. What do you think?

root-project.gradle
yifanyang: Done. Verified locally that 0.8.3 works.

allisonbm92:
FYI, the smoke tests are flaky. Looking at the failures, you can probably merge this even if they don't pass.

PR #259, opened 2019-03-06
Add okhttp3 settings to Functions proguard.

bklimt:
I'm not sure if this is the best way to implement this. But this was needed to make the warnings go away for the test app so that it would pass smoke tests.


bklimt:
@davidmotson IIUC, if we want to fix this in `firebase-functions/proguard.txt`, it'll have to be cherrypicked into the current release, correct?

bklimt:
Moved it.

PR #261, opened 2019-03-08
 Implement firebase-firestore-ktx Kotlin library.

vkryachko:


firebase-firestore-ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -0,0 +1,107 @@
...
+inline fun <reified T> DocumentSnapshot.toObject(
+    serverTimestampBehavior: DocumentSnapshot.ServerTimestampBehavior
+): T? =
+        toObject(T::class.java, serverTimestampBehavior)
+
+/**
+ * Returns the value at the field, converted to a POJO, or null if the field or document doesn't
+ * exist.
+ */
+inline fun <reified T : Any> DocumentSnapshot.getField(field: String): T? = get(field, T::class.java)
```
wilhuff: `reified T : Any` as a type parameter here seems wrong. If the function can return `null` (because the return type is `T?`) shouldn't this be just `reified T`?

firebase-firestore-ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreTests.kt
```diff
@@ -0,0 +1,163 @@
...
+
+const val APP_ID = "APP_ID"
+const val API_KEY = "API_KEY"
+
+const val EXISTING_APP = "existing"
+
+abstract class BaseTestCase {
+    @Before
+    fun setUp() {
+        FirebaseApp.initializeApp(
```
wilhuff: Shouldn't this be `Firebase.app`?

firebase-firestore-ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreTests.kt
```diff
@@ -0,0 +1,163 @@
...
+        val app = Firebase.app(EXISTING_APP)
+        assertThat(Firebase.firestore(app)).isSameAs(FirebaseFirestore.getInstance(app))
+    }
+}
+
+@RunWith(RobolectricTestRunner::class)
+class LibraryVersionTest : BaseTestCase() {
+    @Test
+    fun `library version should be registered with runtime`() {
+        val uaPublisher = Firebase.app.get(UserAgentPublisher::class.java)
```
wilhuff: "ua" isn't very clear. Maybe just "publisher"?

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -293,6 +293,21 @@ public WriteBatch batch() {
     return new WriteBatch(this);
   }
 
+  /**
+   * Executes a batchFunction on a newly created {@link WriteBatch} and then commits all of the
+   * writes made by the batchFunction as a single atomic unit.
+   *
+   * @param batchFunction The function to execute within the batch context.
+   * @return A Task that will be resolved when the write finishes.
```
wilhuff: This comment inconsistently refers to the batch. Above it's "batch context" but here it's "write finishes". 

How about: `@return A Task that will be resolved when the batch has been committed.`?

firebase-firestore-ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
vkryachko: Good catch, `Any` was not intentional, just an oversight, removed here and below

firebase-firestore-ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreTests.kt
vkryachko: Since this is a robolectric test, the default app is not automatically initialized, hence I am initializing it explicitly. Changed to kotlin equivalent though: `Firebase.initialize()`

firebase-firestore-ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreTests.kt
vkryachko: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
vkryachko: Changed

firebase-firestore-ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -0,0 +1,107 @@
...
+
+/** Returns the contents of the document converted to a POJO or null if the document doesn't exist. */
+inline fun <reified T> DocumentSnapshot.toObject(
+    serverTimestampBehavior: DocumentSnapshot.ServerTimestampBehavior
+): T? =
+        toObject(T::class.java, serverTimestampBehavior)
+
+/**
+ * Returns the value at the field, converted to a POJO, or null if the field or document doesn't
+ * exist.
```
wilhuff: Are these public javadoc-equivalents? If so shouldn’t these have `@param` and otherwise match the java documentation?

We can do a clean-up pass in a separate PR if it helps.

vkryachko: A followup PR sounds good, will create one shortly

Issue #264, opened 2019-03-08
Use RestrictTo annotations and then add a linter for all public methods


samtstern:
## Part 1 - Use RestrictTo annotations

If we have `public` methods that are only meant to be called by other Firebase libraries (like things in `common` or `core`) we should annotate them with `RestrictTo(Scope.LIBRARY_GROUP)`.

https://developer.android.com/reference/android/support/annotation/RestrictTo.Scope

## Part 2 - Add a lint check

Any `public` method or class in our SDKs should be annotated with one of these three annotations:

  * `PublicApi` - actually meant to be used by 3p developers
  * `RestricTo(Scope....)` - only meant to be used by us
  * `VisibleForTesting` - not a great idea, but sometimes you need this one

## Part 3 - Abort on error

We should clean up all existing violations in the codebase and then add `lintOptions { abortOnError true }` so that we cannot footgun ourselves anymore.

samtstern:
cc @mikelehen

samtstern:
@google-oss-bot back off or I will pull your plug.

schmidt-sebastian:
I open #266 to add these annotations to the Storage API. To do this, I created a script that can be used to get the ball rolling (the result likely needs to be modified a bit before it can be submitted). The script is here: https://gist.github.com/schmidt-sebastian/c467724506c8c3b9937118f4c6823a18

It should be called with one argument - the main folder of the source tree to modify. That way, it doesn't modify any of our test files. Since the script does not format the source afterwards, the Gradle source formatter should be run afterwards.

For now, I tried to restrict the number of annotations that I am adding. If a class isn't public, but it's methods are, I opted not to add the annotations. While this may make linting harder, it reduces the number of changes significantly. Furthermore, I am also excluding methods marked as `@Override`.

PR #266, opened 2019-03-11
Adding RestrictTo annotations to Firebase Storage

schmidt-sebastian:
This adds `@RestrictTo` annotations to FirebaseStorage using the script I linked to in https://github.com/firebase/firebase-android-sdk/issues/264. 

The only manual modification was to remove superfluous annotations from StorageTaskManager. StorageTaskManager is a package-private class but the "AI" in the script treats it as a public class.

I verified that Android Studio 3.1.4 does not let me use the classes and their methods anymore.

### DIFF
--- a/firebase-storage/src/main/java/com/google/firebase/storage/internal/ActivityLifecycleListener.java
+++ b/firebase-storage/src/main/java/com/google/firebase/storage/internal/ActivityLifecycleListener.java
@@ -17,6 +17,7 @@ package com.google.firebase.storage.internal;
 import android.app.Activity;
 import android.support.annotation.MainThread;
 import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
 import android.util.Log;
 import com.google.android.gms.common.api.internal.LifecycleActivity;
 import com.google.android.gms.common.api.internal.LifecycleCallback;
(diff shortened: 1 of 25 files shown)


firebase-storage/src/main/java/com/google/firebase/storage/internal/ActivityLifecycleListener.java
```diff
@@ -33,6 +34,7 @@
  * @hide
  */
 @SuppressWarnings("JavaDoc")
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
mikelehen: Rather than adding this to every class, can we put it on the package?  e.g. add a package-info.java in .../internal/ with:

```
@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
package com.google.firebase.storage.internal;
```

I don't want to have to remember to put this on every class if possible.

schmidt-sebastian: Yes, that works as well. Updated the PR.

firebase-storage/src/main/java/com/google/firebase/storage/network/connection/package-info.java
```diff
@@ -13,4 +13,7 @@
 // limitations under the License.
 
 /** @hide */
-package com.google.firebase.storage.network.connection;
\ No newline at end of file
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
mikelehen: I take it the one on network/ doesn't cascade to network/connection?

schmidt-sebastian: No, unfortunately it doesn't.

PR #267, opened 2019-03-11
firebase-firestore-ktx KDoc improvements.

vkryachko:


firebase-firestore-ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -34,10 +34,24 @@ val Firebase.firestore
 /** Returns the [FirebaseFirestore] instance of a given [FirebaseApp]. */
 fun Firebase.firestore(app: FirebaseApp): FirebaseFirestore = FirebaseFirestore.getInstance(app)
 
-/** Returns the contents of the document converted to a POJO or null if the document doesn't exist. */
+/**
+ * Returns the contents of the document converted to a POJO or null if the document doesn't exist.
+ *
+ * @param T The Java class to create.
```
wilhuff: Though this text exactly matches the corresponding Javadoc on `DocumentSnapshot` this doesn't seem right.

In the Javadoc, this is `@param valueType`, where it's describing that you actually have to pass the instance of `java.lang.Class` to make this work. For example, to get a value of type `Model`, you need to pass `Model.class`. The Javadoc is documenting that instance you pass, not the type parameter.

In the KDoc, the class instance is hidden behind a reified generic, and what you're documenting here is the type. Secondarily, we're going out of our way to hide the "java class" that you get via `T::class.java`.

A proposed alternative would be:

```
@param T The type of the object to create.
```

Here and throughout.

wilhuff: Quote about `T`: currently `T` is using the default bound of `Any?`. This would allow you to request a value of type `Int`, despite the fact that this can't actually work--the parameter must be a structured type, containing fields.

Java's generic syntax is sufficiently constrained that we can't express this constraint. Is there a way to do so in Kotlin? If not, no worries, but if possible without too much verbosity this would be worthwhile.

vkryachko: Done

vkryachko: Afaik it is not possible. There is nothing special about `Int`, `Float` at the type level, they are just subtypes of `Any` just like any other user-defined type.

wilhuff:
I thought the `T: Any` in `inline fun <reified T: Any> QueryDocumentSnapshot.toObject(): T = toObject(T::class.java)` was intended.

Unlike `DocumentSnapshot.toObject`, `QueryDocumentSnapshot.toObject` isn't nullable.

PR #270, opened 2019-03-12
Add RestrictTo annotations to Firebase Database

schmidt-sebastian:
Same as https://github.com/firebase/firebase-android-sdk/pull/266 but for the RTDB and firebase-database-collection.

Addresses #264

### DIFF
--- a/firebase-database-collection/firebase-database-collection.gradle
+++ b/firebase-database-collection/firebase-database-collection.gradle
@@ -22,8 +22,11 @@ android {
 }
 
 dependencies {
+    implementation "com.google.android.gms:play-services-base:$playServicesVersion"
+
     testImplementation 'junit:junit:4.12'
     testImplementation 'net.java:quickcheck:0.6'
+
     testAnnotationProcessor 'net.java:quickcheck-src-generator:0.6'
     testAnnotationProcessor 'net.java.quickcheck:quickcheck-src-generator:0.6'
 }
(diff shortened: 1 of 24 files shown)


firebase-database/src/main/java/com/google/firebase/database/InternalHelpers.java
```diff
@@ -27,6 +28,7 @@
  *
  * @hide
  */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
mikelehen: FWIW- I think this whole wrapper class could theoretically be removed in favor of adding RestrictTo to the underlying methods being called.  But I also think it's not worth doing.

firebase-database-collection/firebase-database-collection.gradle
```diff
@@ -22,8 +22,11 @@ android {
 }
 
 dependencies {
+    implementation "com.google.android.gms:play-services-base:$playServicesVersion"
```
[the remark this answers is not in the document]
schmidt-sebastian: This adds an indirect dependency on the Android Support Annotations package that includes `@RestrictTo`. Rather than including it directly (and keep updating the version as Play Services updates its dependencies), I pulled in the play-services-base package which is already used by both Firestore and RTDB.

firebase-database/src/main/java/com/google/firebase/database/InternalHelpers.java
schmidt-sebastian: My helper script is not capable of doing this yet, so I am going to punt on this :)

PR #271, opened 2019-03-12
Add RestrictTo annotations to Firestore

schmidt-sebastian:
Same as #266 and #270 but for Firestore.

I fixed a couple of places where we were missing the `@PublicApi` annotation. See cr/151329304 (internal) for reference.

Addresses #264

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
@@ -21,6 +21,7 @@ import static java.util.Collections.singletonList;
 
 import android.app.Activity;
 import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.android.gms.tasks.Tasks;
@@ -71,6 +72,7 @@ public class DocumentReference {
   }
 
   /** @hide */
+  @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
   public static DocumentReference forPath(ResourcePath path, FirebaseFirestore firestore) {
     if (path.length() % 2 != 0) {
       throw new IllegalArgumentException(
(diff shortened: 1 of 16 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
```diff
@@ -71,6 +72,7 @@
   }
 
   /** @hide */
+  @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
rsgowman: We recently dropped the 'public' from this, so we don't need to explicitly "RestrictTo" it. i.e. if you merge master, you can revert your changes to this file.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
```diff
@@ -313,6 +314,7 @@ public Object get(
    * @return The value at the given field or null.
    */
   @Nullable
+  @PublicApi
```
rsgowman: Note that despite lacking this annotation, these methods were still showing up in our public api.

firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
```diff
@@ -46,12 +47,14 @@ private SetOptions(boolean merge, @Nullable FieldMask fieldMask) {
   }
 
   /** @hide */
+  @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
rsgowman: Similar to above, this method is no longer public (but note that getFIeldMask() still remains as a public method.)

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
schmidt-sebastian: Merged and removed.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
schmidt-sebastian: `@PublicApi` was used when we obfuscated our codebase. I am wondering if Proguard does not obfuscate methods when overloads are marked `@PublicApi`. But we hopefully never have to find an answer to that question :)

firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
schmidt-sebastian: Done

PR #272, opened 2019-03-12
Reject non-standard Number classes during POJO serialization

schmidt-sebastian:
This change adds custom error messages for BigDecimal, BigInteger, AtomicInteger, etc during serialization and deserialization. 

Note that we currently do throw errors as well.
- Deserialization:
Old:  "Unknown primitive type: class java.math.BigInteger"
New: "Could not deserialize object. Deserializing to BigInteger is not supported"

- Serialization: 
Old: "Invalid data. Unsupported type: java.math.BigInteger (found in field bigInteger)"
New: "Could not serialize object. BigInteger is not supported, please use int, long, float or double"

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
@@ -113,26 +113,21 @@ public class CustomClassMapper {
     if (o == null) {
       return null;
     } else if (o instanceof Number) {
-      if (o instanceof Float) {
-        return ((Float) o).doubleValue();
-      } else if (o instanceof Short) {
-        throw serializeError(path, "Shorts are not supported, please use int or long");
-      } else if (o instanceof Byte) {
-        throw serializeError(path, "Bytes are not supported, please use int or long");
-      } else if (o instanceof Long || o instanceof Integer || o instanceof Double) {
+      if (o instanceof Long || o instanceof Integer || o instanceof Double || o instanceof Float) {
         return o;
       } else {
         throw serializeError(
             path,
-            o.getClass().getSimpleName()
-                + " is not supported, please use int, long, float or double");
+            String.format(
+                "%s is not supported, please use an int, long, float or double",
+                o.getClass().getSimpleName()));
       }
     } else if (o instanceof String) {
       return o;
     } else if (o instanceof Boolean) {
       return o;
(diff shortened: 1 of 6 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -119,9 +119,13 @@ public static Object convertToPlainJavaTypes(Object object) {
         throw serializeError(path, "Shorts are not supported, please use int or long");
       } else if (o instanceof Byte) {
         throw serializeError(path, "Bytes are not supported, please use int or long");
-      } else {
-        // Long, Integer, Double
+      } else if (o instanceof Long || o instanceof Integer || o instanceof Double) {
```
wilhuff: The underlying UserDataConverter handles `Float` types so you can combine the cases that return a value. With this general error message it no longer seems worthwhile to handle `Short` and `Byte` separately.

Something like this:

```java
if (o instanceof Long || o instanceof Integer || o instanceof Double || o instanceof Float) {
  return o;
} else {
  throw serializeError(...);  // your new message
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -320,7 +324,8 @@ public static Object convertToPlainJavaTypes(Object object) {
     } else if (Character.class.isAssignableFrom(clazz) || char.class.isAssignableFrom(clazz)) {
       throw deserializeError(path, "Deserializing to chars is not supported");
     } else {
-      throw new IllegalArgumentException("Unknown primitive type: " + clazz);
+      throw deserializeError(
```
wilhuff: Similarly, do we need to special case `short`, `byte`, and `char` if we give a nearly equivalent message here?

schmidt-sebastian: We can probably reduce our code duplication here as well. I removed the extra checks.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
schmidt-sebastian: Yeah, that is a good suggestion. I moved some of the code duplication to the test cases instead...

PR #277, opened 2019-03-13
Firebase Storage Component Registration

schmidt-sebastian:
This definitely falls under the umbrella "I don't know what I am doing". But here we go...

This follows the pattern in Firestore to register the component for Firebase Storage. It changes FirebaseStorage to only ever work with a single app, but keeps the option to get different instances for different storage buckets.

I also added `getInstanceImpl` as a valid name for the FirebaseApp.get() call, as I didn't want to add yet another `getInstance()` overload (which would result in an ambiguous null cast).

So what does everyone think? :)

### DIFF
--- a/firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
+++ b/firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
@@ -14,8 +14,6 @@
 
 package com.google.firebase.storage;
 
-import static com.google.android.gms.common.internal.Preconditions.checkNotNull;
-
 import android.net.Uri;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
@@ -27,6 +25,7 @@ import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
 import com.google.firebase.annotations.PublicApi;
 import com.google.firebase.auth.internal.InternalAuthProvider;
+import com.google.firebase.inject.Provider;
 import com.google.firebase.storage.internal.Util;
 import java.io.UnsupportedEncodingException;
 
@@ -48,7 +47,7 @@ public class FirebaseStorage {
   private static final String STORAGE_BUCKET_WITH_PATH_EXCEPTION =
       "The storage Uri cannot contain a path element.";
   @NonNull private final FirebaseApp mApp;
-  @Nullable private final InternalAuthProvider mAuthProvider;
+  @Nullable private final Provider<InternalAuthProvider> mAuthProvider;
   @Nullable private final String mBucketName;
   private long sMaxUploadRetry = 10 * DateUtils.MINUTE_IN_MILLIS; //  10 * 60 * 1000
   private long sMaxDownloadRetry = 10 * DateUtils.MINUTE_IN_MILLIS; //  10 * 60 * 1000
@@ -57,7 +56,7 @@ public class FirebaseStorage {
   FirebaseStorage(
       @Nullable String bucketName,
       @NonNull FirebaseApp app,
-      @Nullable InternalAuthProvider authProvider) {
+      @Nullable Provider<InternalAuthProvider> authProvider) {
     mBucketName = bucketName;
     mApp = app;
     mAuthProvider = authProvider;
@@ -70,9 +69,9 @@ public class FirebaseStorage {
       throw new IllegalArgumentException(STORAGE_BUCKET_WITH_PATH_EXCEPTION);
     }
 
-    checkNotNull(app, "Provided FirebaseApp must not be null.");
+    Preconditions.checkNotNull(app, "Provided FirebaseApp must not be null.");
     FirebaseStorageComponent component = app.get(FirebaseStorageComponent.class);
-    checkNotNull(component, "Firebase Storage component is not present.");
+    Preconditions.checkNotNull(component, "Firebase Storage component is not present.");
     return component.get(bucketName);
   }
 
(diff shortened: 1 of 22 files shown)


firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorageComponent.java
```diff
@@ -0,0 +1,50 @@
...
+  synchronized FirebaseStorage get(@Nullable String bucketName) {
+    FirebaseStorage storage = instances.get(bucketName);
+    if (storage == null) {
+      storage = FirebaseStorage.newInstance(bucketName, app);
+      instances.put(bucketName, storage);
+    }
+    return storage;
+  }
+
+  /** @hide */
```
vkryachko: I think `@hide` is not needed for non-public classes/symbols

firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorageComponent.java
```diff
@@ -0,0 +1,50 @@
...
+  /** A static map from storage buckets to FirebaseStorage instances. */
+  private static final Map<String, FirebaseStorage> instances = new HashMap<>();
+
+  private final FirebaseApp app;
+
+  FirebaseStorageComponent(@NonNull FirebaseApp app) {
+    this.app = app;
+  }
+
+  /** Provides instances of Firestore for given database names. */
```
vkryachko: `s/Firestore/Storage/`

firebase-storage/src/main/java/com/google/firebase/storage/StorageRegistrar.java
```diff
@@ -16,19 +16,31 @@
...
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class StorageRegistrar implements ComponentRegistrar {
   @Override
+  @Keep
   public List<Component<?>> getComponents() {
-    return Collections.singletonList(
+    return Arrays.asList(
+        Component.builder(FirebaseStorageComponent.class)
+            .add(Dependency.required(FirebaseApp.class))
+            .add(Dependency.optional(InternalAuthProvider.class))
```
vkryachko: With my limited understanding of storage it looks like `InternalAuthProvider` is not needed right away(when storage initializes) unlike in `Firestore`, but only used in response to user calls to downloads and what not. If that is the case consider switching to `Dependency.optionalProvider(InternalAuthProvider.class)`, what it will do is defer initialization of `InternalAuthProvider` until it is needed as well as prevent potential dependency cycles in the future.

That said, currently the only implementor of `InternalAuthProvider` is `FirebaseAuth` which is eager, so deferred initialization will not apply, but it seems a relatively easy change to do(especially if the `Provider<T>` is hidden behind storage.getAuthProvider()) to bulletproof storage for the future. wdyt?

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
```diff
@@ -61,6 +62,7 @@
   private final AdaptiveStreamBuffer mStreamBuffer;
   // Active, current mutable state.
   private final AtomicLong mBytesUploaded = new AtomicLong(0);
+  @Nullable final InternalAuthProvider mAuthProvider;
```
vkryachko: could it be private?

firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
```diff
@@ -151,6 +145,8 @@ public static FirebaseStorage getInstance(@NonNull FirebaseApp app) {
   public static FirebaseStorage getInstance(@NonNull FirebaseApp app, @NonNull String url) {
     // noinspection ConstantConditions
     Preconditions.checkArgument(app != null, "Null is not a valid value for the FirebaseApp.");
+    Preconditions.checkArgument(
```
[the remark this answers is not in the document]
schmidt-sebastian: The rest of the SDK doesn't use static imports, so I changed this file to use the qualified import checkNotNull.

firebase-storage/src/main/java/com/google/firebase/storage/StorageRegistrar.java
schmidt-sebastian: Your understanding is correct. We only fetch a token once a user requests a file upload or download, so we don't need the AuthProvider until later. The change for this was relatively minor, so I updated this PR.

BTW, if wasn't immediately obvious to me whether optionalProvider  makes Provider nullable or whether it makes Provider.get() nullable. I think it's the former (from looking through the code and playing with it). Maybe we can document this?

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
schmidt-sebastian: Yes. FIxed

firebase-storage/src/main/java/com/google/firebase/storage/StorageRegistrar.java
```diff
@@ -16,19 +16,31 @@
...
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
 
 /** @hide */
 @Keep
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class StorageRegistrar implements ComponentRegistrar {
   @Override
+  @Keep
```
vkryachko: nit: no need to `@Keep` this method, just the class is enough

schmidt-sebastian: Removed

PR #280, opened 2019-03-14
Database: Rename test helpers

schmidt-sebastian:
While working on the component registration for the RTDB, I got super confused because we have two TestHelper files in the same package. One is used for Unit Tests and one for Integration Tests. They used to be part of the same package, but I split them up when we moved to GitHub.

This PR:
- Renames them to IntegrationTestHelpers and UnitTestHelpers respectivley
- Moves ReadFuture and WriteFuture from testUtil to androidTest, since they require the IntegrationTestHelpers
- Removes some unused methods from the test helpers


schmidt-sebastian:
Padded!

PR #281, opened 2019-03-14
Fix Firestore POJO tests

schmidt-sebastian:
Looks like #272 broke the Firestore Integration Tests, which I noticed in the failures for #277.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/POJOTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/POJOTest.java
@@ -301,19 +301,19 @@ public class POJOTest {
     pojo.bigIntegerValue = new BigInteger("0");
     expectedErrorMessages.put(
         pojo,
-        "Could not serialize object. BigInteger is not supported, please use an int, long, float or double (found in field 'bigIntegerValue')");
+        "Could not serialize object. Numbers of type BigInteger are not supported, please use an int, long, float or double (found in field 'bigIntegerValue')");
 
     pojo = new InvalidPOJO();
     pojo.byteValue = 0;
     expectedErrorMessages.put(
         pojo,
-        "Could not serialize object. Byte is not supported, please use an int, long, float or double (found in field 'byteValue')");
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/POJOTest.java
```diff
@@ -301,19 +301,19 @@ public void testCantWriteNonStandardNumberTypes() {
     pojo.bigIntegerValue = new BigInteger("0");
     expectedErrorMessages.put(
         pojo,
-        "Could not serialize object. BigInteger is not supported, please use an int, long, float or double (found in field 'bigIntegerValue')");
+        "Could not serialize object. Numbers of type BigInteger are not supported, please use an int, long, float or double (found in field 'bigIntegerValue')");
```
rsgowman: So remember in that other PR? Where I was talking about over-specified error message checking in tests? :)

(This is still fine; no action required.)

mikelehen: FWIW- As a reviewer, I actually like that I can just look at the tests to see the exact error message that comes out without worrying that we're missing a space somewhere or that some generated part of the message is coming out goofy.

Also, as a reviewer, I should actually check if the tests pass before merging PRs.
box_of_shame

(in any case, there's definitely a balance to be struck with these sorts of tests... if we feel they're too fragile, I don't mind revisiting)

mikelehen:
Since Sebastian is in a different timezone and tests are broken without this, I'm merging this on his behalf again (I haven't learned my lesson apparently).  somke-tests are failing but I'm confident it's unrelated to this change.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/POJOTest.java
schmidt-sebastian: I actually am a big fan of having at least one place that checks the full error message. A lot of our error messages are composed of different parts, and it is very easy to omit words/spaces or put extra spaces when we do this.

PR #284, opened 2019-03-14
[datatransport] Implement EventStore backed by SQLite.

vkryachko:
The database contains 3 tables:

* `events`: contains individual events
* `event_metadata`: contains many-to-one metadata key-value pairs
  associated with events
* `backends`: contains metadata about known/registered backends


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  private static Priority toPriority(int value) {
+    if (value < 0 || value >= ALL_PRIORITIES.length) {
+      return Priority.DEFAULT;
+    }
+    return ALL_PRIORITIES[value];
+  }
+
+  private static class OpenHelper extends SQLiteOpenHelper {
```
wilhuff: Your OpenHelper could be better set up to handle schema changes.

Consider instead structuring this as a series of migrations, and then run the migrations required to get from one version to the next. See [Firestore's implementation](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java#L75) for ideas.

This also makes it easy to test that the migrations are happening correctly in downgrade/upgrade scenarios. See [SQLiteSchemaTest](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteSchemaTest.java)

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,355 @@
...
+          if (rowsUpdated < 1) {
+            values.put("name", backendName);
+            db.insert("backends", null, values);
+          }
+          return null;
+        });
+  }
+
+  @Override
+  public Iterable<PersistedEvent> loadAll(String backendName) {
```
wilhuff: This function is huge. Consider breaking it up into methods whose names document what you're doing.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,355 @@
...
+      }
+    }
+    query.append(')');
+    getDb().compileStatement(query.toString()).execute();
+  }
+
+  @Override
+  @Nullable
+  public Long getNextCallTime(String backendName) {
+    try (Cursor cursor =
```
wilhuff: Consider using `SQLiteStatement.simpleQueryForString` for cases like this. It avoids constructing a Cursor and makes the call much less expensive.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,355 @@
+// Copyright 2018 Google LLC
```
wilhuff: 2019

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,355 @@
...
+  private SQLiteDatabase db;
+
+  @Inject
+  SQLiteEventStore(Context applicationContext) {
+    this.openHelper = new OpenHelper(applicationContext);
+  }
+
+  private SQLiteDatabase getDb() {
+    if (db == null) {
+      db = openHelper.getWritableDatabase();
```
wilhuff: Your in-memory EventStore synchronizes many methods leading me to believe you're defending against multiple threads storing data this way.

This assignment is unsynchronized, which seems like you could end up with multiple opens happening concurrently, which possibly could result in multiple OpenHelpers attempting to create the schema concurrently.

I don't think making `getDb()` synchronized would be enough, because your implementation of atomically essentially assumes that `db` has a current transaction in force. Multiple concurrent callers would attempt to use that same transaction from multiple threads.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,355 @@
...
+    values.put("backend_id", backendName);
+    values.put("transport_name", event.getTransportName());
+    values.put("priority", event.getPriority().ordinal());
+    values.put("timestamp_ms", event.getEventMillis());
+    values.put("uptime_ms", event.getUptimeMillis());
+    values.put("payload", event.getPayload());
+    values.put("num_attempts", 0);
+
+    long newRowId =
+        inTransaction(
```
wilhuff: I thought the idea was that callers would call something like

```
store.atomically({
  store.persist(foo, bar);
})
```

Doesn't that mean you're creating a nested transaction by default in this usage?

If atomically is required, persist and friends should assert that they're called in the block rather than opening a nested transaction.

A type safe way to make this work is to have persist and similar writing methods on a nested type like EventStore.Writer and have atomically pass that in.

Then you get

```
store.atomically(writer -> {
  writer.persist(foo, bar);
})
```

This would also make it easy to have a thread-local writer that would allow concurrent usage.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: Good catch, but it is not the case. We are explicitly designing the library to be single-threaded as it greatly simplifies the design. Essentially all db access will happen on a single thread sequentially.

What's not entirely clear to me is how my setup will behave in a multiprocess setup which we have to plan for being a low-level library that's going to be used by other SDKs. Specifically, below I issue `PRAGMA busy_timeout = 5000` in the hope that concurrent access to the db file won't immediately result in `sqlite busy` but would instead wait for the transaction lock to get released and then proceed normally. Does this approach seem reasonable to you?

wilhuff: That's not clear to me.

Within Firestore we tie which SQLite DB we open to the current FirebaseApp and then open then exclusively. We disallow multiple processes to access the same DB concurrently.

The mechanism you're using isn't one I'm familiar with.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
wilhuff: db.compileStatement does allow binding. See bindFoo() on SQLiteProgram

It's something like this:
```java
SQLiteStatement query = db.compileStatement("SELECT next_request_ms FROM backends WHERE name = ?");
query.bindString(1, backendName);
nextRequestMs  = query.simpleQueryForLong();
```

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
wilhuff: You can refactor when you need it for sure.

The problematic case is when you do ship schema v2, you need to ensure that your migrations are idempotent after a user downgrades to 1 and then re-upgrades to 2. Firestore didn't do this initially and so we ended up failing trying to create tables that already existed.

If you're not going to use migrations now, make sure to leave a monster TODO indicating SCHEMA_VERSION cannot change until you can handle re-upgrades after downgrades.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
wilhuff: Fair enough.

Reading a bit more about this, it seems like there's [no real penalty to using nested transactions this way](https://stackoverflow.com/questions/26106237/sqlitedatabase-nested-transaction-and-workaround) (and no benefit either) so it's not harmful.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: OK, so just to document my findings.

`PRAGMA busy_timeout` is only supported in sqlite `3.7.15+`, which is on present only api21+ devices so we can't use that, instead I will implement blocking/waiting manually.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: cool, that's a good one to know about. I will keep my implementation in this case since I am not guaranteed that the database contains any matching rows, and this will throw an exception and I'll need to have try-catch everywhere.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: Fair enough. Will document in properly

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+import java.util.ListIterator;
+import java.util.Map;
+import java.util.Set;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/** {@link EventStore} implementation backed by a SQLite database. */
+@Singleton
+@WorkerThread
+public class SQLiteEventStore implements EventStore, SynchronizationGuard {
```
ashwinraghav: Chatted with @vkryachko offline to determine if Room is a good fit for us

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/MultiProcessSynchronizationGuardTest.java
```diff
@@ -0,0 +1,160 @@
...
+    assertThat(rpc.tryAcquireLock(0)).isTrue();
+    try {
+      runnable.run();
+    } finally {
+      rpc.releaseLock();
+    }
+  }
+
+  private static void assertCanLock(IRemoteLockRpc rpc, boolean can) throws RemoteException {
+    boolean locked = rpc.tryAcquireLock(0);
```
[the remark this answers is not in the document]
vkryachko: it can't be, since `locked` will not be in scope in the finally block

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+    values.put("backend_id", backendName);
+    values.put("transport_name", event.getTransportName());
+    values.put("priority", event.getPriority().ordinal());
+    values.put("timestamp_ms", event.getEventMillis());
+    values.put("uptime_ms", event.getUptimeMillis());
+    values.put("payload", event.getPayload());
+    values.put("num_attempts", 0);
+
+    long newRowId =
+        inTransaction(
```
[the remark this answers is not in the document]
vkryachko: Not really, we need the transaction to make sure all inserts happen atomically: `event+metadata`. And then we just return the id of the event, if it were a single insert, we would not have needed a transaction.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+              return newEventId;
+            });
+
+    return PersistedEvent.create(newRowId, backendName, event);
+  }
+
+  @Override
+  public void recordFailure(Iterable<PersistedEvent> events) {
+    StringBuilder query =
+        new StringBuilder("UPDATE events SET num_attempts = num_attempts + 1 WHERE _id in (");
```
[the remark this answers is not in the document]
vkryachko: yes, that is correct and intended

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+      query.append(iterator.next().getId());
+      if (iterator.hasNext()) {
+        query.append(',');
+      }
+    }
+    query.append(')');
+    inTransaction(
+        db -> {
+          db.compileStatement(query.toString()).execute();
+          db.compileStatement("DELETE FROM events WHERE num_attempts >= " + MAX_RETRIES).execute();
```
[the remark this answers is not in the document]
vkryachko: We anticipate 2 failure modes:
* transient error: backend down, connection problem
* non transient: bad data in one of the events, e.g. we receive 400

We only use num_attempts for transient errors, for non transient we just delete the events with the assumption that they are invalid.

> expected to fail shortly after

we will have exponential back-off in an effort to increase the probability that the backend is up by the time we run out of attempts.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+    }
+  }
+
+  @Override
+  public void recordNextCallTime(String backendName, long timestampMs) {
+    inTransaction(
+        db -> {
+          ContentValues values = new ContentValues();
+          values.put("next_request_ms", timestampMs);
+          int rowsUpdated = db.update("backends", values, "name = ?", new String[] {backendName});
```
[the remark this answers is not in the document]
vkryachko: nope, our supported version range for sqlite3 is 3.7.4(ICS) through 3.22.0(Pie), upsert is new in version 3.24

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  /** Loads all events for a backend. */
+  private List<PersistedEvent> loadEvents(SQLiteDatabase db, String backendName) {
+    List<PersistedEvent> events = new ArrayList<>();
+    try (Cursor cursor =
+        db.query(
+            "events",
+            new String[] {
+              "_id", "transport_name", "priority", "timestamp_ms", "uptime_ms", "payload"
```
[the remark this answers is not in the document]
vkryachko: I don't think there is, if we do that it would make sense to also rewrite queries to use these constants, and that would effectively make them unreadable, while not using them in queries will defeat the purpose of having constants in the first place. wdyt?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+  @Override
+  public void recordSuccess(Iterable<PersistedEvent> events) {
+    StringBuilder query = new StringBuilder("DELETE FROM events WHERE _id in (");
+    Iterator<PersistedEvent> iterator = events.iterator();
+    while (iterator.hasNext()) {
+      query.append(iterator.next().getId());
+      if (iterator.hasNext()) {
+        query.append(',');
+      }
+    }
```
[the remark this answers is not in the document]
vkryachko: Done

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,443 @@
...
+              }
+              return newEventId;
+            });
+
+    return PersistedEvent.create(newRowId, backendName, event);
+  }
+
+  @Override
+  public void recordFailure(Iterable<PersistedEvent> events) {
+    StringBuilder query =
```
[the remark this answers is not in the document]
vkryachko: Good catch, it is safe but why make the call if we know it will have no effect. Added a check.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
ashwinraghav: **Ashwin**: Did you consider using Room ?

**Vlad**: yep, it transitively pulls hundreds of KB of dependencies
I essentially decided to punt this decision until/if we have a sofisticated enough setup(now it is just 3 simple tables)

**Ashwin**: Just tried it on my app. I do see it adding 100K without proguard. The interesting thing is that clearcut adds more than this (for many products, but not for some others since there are common transitive deps). So the overall could be a net gain

**Vlad**: well, we are a bit over 200KB ourselves already and we are not feature complete yet, so I would not add another 100 lightly

**Vlad**: (that size includes our transitive deps of course)

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/RemoteLockRpc.java
```diff
@@ -0,0 +1,71 @@
...
+class RemoteLockRpc extends IRemoteLockRpc.Stub {
+  private final Executor executor;
+  private final Locker<Boolean> acquireReleaseLocker = new Locker<>();
+
+  RemoteLockRpc(Executor executor) {
+    this.executor = executor;
+  }
+
+  @Override
+  public boolean tryAcquireLock(long lockTimeoutMs) {
```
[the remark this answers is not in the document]
vkryachko: 1. Not sure I fully understand your question, what we guarantee is that `Boolean result = sectionEnteredLocker.await();` block until critical section is entered and calls `setResult(true)`.

2. There is no such guarantee as I mentioned in 1. It does not matter what happens first `setResult()` or `await()` they will synchronize on each other regardless and unblock each other.

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/LocalBinderService.java
```diff
@@ -0,0 +1,32 @@
...
+import android.content.Intent;
+import android.os.IBinder;
+import android.util.Log;
+import com.google.android.datatransport.runtime.TransportRuntime;
+import java.util.concurrent.Executors;
+
+/** Service instance that runs in the main process of the Android application. */
+public class LocalBinderService extends Service {
+  @Override
+  public IBinder onBind(Intent intent) {
```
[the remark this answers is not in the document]
vkryachko: Done

transport/transport-runtime/src/androidTest/aidl/com/google/android/datatransport/runtime/IRemoteLockRpc.aidl
```diff
@@ -0,0 +1,14 @@
+package com.google.android.datatransport.runtime;
+
+/** RPC for manipulating SynchronizationGuard locking across process boundary. */
+interface IRemoteLockRpc {
```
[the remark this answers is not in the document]
vkryachko: done

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,444 @@
...
+  @Override
+  public <T> T runCriticalSection(long lockTimeoutMs, CriticalSection<T> criticalSection) {
+    SQLiteDatabase db = getDb();
+    ensureBeginTransaction(db, lockTimeoutMs);
+    try {
+      T result = criticalSection.execute();
+      db.setTransactionSuccessful();
+      return result;
+    } finally {
+      db.endTransaction();
```
ashwinraghav: Does this throw any errors that may result in the transaction not ending and causing others to wait?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,444 @@
...
+@WorkerThread
+public class SQLiteEventStore implements EventStore, SynchronizationGuard {
+
+  static final int MAX_RETRIES = 10;
+
+  private static final Priority[] ALL_PRIORITIES = Priority.values();
+
+  @VisibleForTesting
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  public static final int LOCK_RETRY_BACK_OFF = 50;
```
ashwinraghav: Rename to LOCK_RETRY_BACK_OFF_MILLIS

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,444 @@
...
+      long lockTimeoutMs, Producer<T> retriable, Function<Throwable, T> failureHandler) {
+    long startTime = monotonicClock.getTime();
+    do {
+      try {
+        return retriable.produce();
+      } catch (SQLiteDatabaseLockedException ex) {
+        if (monotonicClock.getTime() >= startTime + lockTimeoutMs) {
+          return failureHandler.apply(ex);
+        }
+        SystemClock.sleep(LOCK_RETRY_BACK_OFF);
```
ashwinraghav: In practice, will there be different threads competing for this?
Does it have to be a read-write lock or do transactional writes make concurrent reads thread safe?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -0,0 +1,444 @@
...
+      }
+    } while (true);
+  }
+
+  interface Producer<T> {
+    T produce();
+  }
+
+  interface Function<T, U> {
```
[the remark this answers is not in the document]
vkryachko: No

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: The only documented failure is if someone is holding a shared lock on the db. It is not possible in our case as we use `exclusive` transactions

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: There won't be as all db access will happen on one dedicated thread. We are also intentionally not using WAL so concurrent writes are not allowed

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: Done

PR #286, opened 2019-03-15
Firebase Database Component Registration

schmidt-sebastian:
This is based on what I learned in #277 and adds Component Registration to the RTDB SDK while also removing the usages of FirebaseApp.getToken().

I kept the AndroidAuthTokenProvider, which (when available) wraps the InternalAuthProvider. With this PR, I no longer have access to the RTDB's internal Executor when I initialize the AndroidAuthTokenProvider. To make sure that the callbacks still run on the same executor, I changed the only callsite of `getToken()` to forward all calls to the existing executor.

The integration tests also no longer initialize a default app, since that is done by our dependencies.


schmidt-sebastian:
Re: Optional provider.

I debated doing so, but decided against it. It would change `AndroidAuthTokenProvider` quite a bit (unless I hide it internally, which doesn't have as much benefit). On top of that, since we follow the same network usage patterns as Firestore, I don't think there is much gain. I suspect that almost all users obtain a database reference and almost immediately connect to the network. I am willing to tackle this in a follow up if we feel that there is opportunity for performance improvements.

PR #294, opened 2019-03-19
Firebase Test Lab plugin

vkryachko:
Registers firebase test lab as device provider making `deviceCheck` task
to run tests on it.

Note: requires gcloud configured with correct IAM permissions.


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
```diff
@@ -0,0 +1,58 @@
...
+    String getName() {
+        return "firebase-test-lab"
+    }
+
+    @Override
+    void uploadApks(String variantName, File testApk, File testedApk) {
+        // test lab requires an "app" apk, so we give an empty apk to it.
+        String testedApkPath = testedApk ?: "$project.rootDir/buildSrc/resources/dummy.apk"
+
+        project.logger.warn("Uploading for $variantName: testApk=$testApk, testedApk=$testedApkPath")
```
ashwinraghav: Ensure `$testApk` prints well.

transport/transport-runtime/transport-runtime.gradle
```diff
@@ -34,6 +35,17 @@ android {
     }
 }
 
+firebaseTestLab {
```
ashwinraghav: Will it be sufficient to run these on testLab when on CI, but locally otherwise?
Otherwise, we should document this as setup instructions for the project.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    String getName() {
+        return "firebase-test-lab"
+    }
+
+    @Override
+    void uploadApks(String variantName, File testApk, File testedApk) {
+        // test lab requires an "app" apk, so we give an empty apk to it.
+        String testedApkPath = testedApk ?: "$project.rootDir/buildSrc/resources/dummy.apk"
```
ashwinraghav: Consider generating dynamically inside `$project.buildDir/outputs/apk/androidTest/dummy.apk`

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
vkryachko: It does

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
vkryachko: There is no easy way to generate a `valid` empty APK, hence using one I generated upfront

transport/transport-runtime/transport-runtime.gradle
ashwinraghav: Chatted with Vlad.
This will change only the deviceCheck tasks and not the connectedCheck tasks as I initially assumed.
It is indeed the case that connectedCheck can be run locally without Test Lab

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
ashwinraghav: Did not know it had to be valid. #FeatureNotABug

PR #324, opened 2019-03-28
Implement backend discovery.

vkryachko:
The mechanism is the following:

Each android libary that provides a backend needs to implement a
`BackendFactory` and register it in its manifest. Runtime will
discover it via manifest and create the backend via the factory,
providing `CreationContext` to it.

Design details: http://doc/17ln50RetnwUiwsVrBxQg6Ocqu8-yppaRHXJhKo3DrVo#heading=h.ny13u8okebw3


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendFactory.java
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.android.datatransport.runtime.backends;
+
+/** Used by backend discovery to create {@link TransportBackend}s. */
```
VinayGuthal: can't we  use TransportBackend directly?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+      return backends.get(name);
+    }
+    String backendProviderName = getBackendProviders().get(name);
+    if (backendProviderName == null) {
+      return null;
+    }
+
+    try {
+      BackendFactory provider =
+          (BackendFactory)
```
VinayGuthal: And do TransportBackend provider = (TransportBackend)  Class.forName(backendProviderName).getInstance() ?

transport/transport-backend-cct/src/main/AndroidManifest.xml
```diff
@@ -14,8 +14,18 @@
...
 
-  <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.INTERNET" />
+
+    <application>
+        <service
+            android:name="com.google.android.datatransport.runtime.backends.TransportBackendDiscovery"
+            android:exported="false">
+            <meta-data
+                android:name="backend:com.google.android.datatransport.cct.CctBackendFactory"
```
VinayGuthal: using CcTransportBackend here?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctBackendFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+import android.support.annotation.Keep;
+import android.support.annotation.VisibleForTesting;
+import com.google.android.datatransport.runtime.backends.BackendFactory;
+import com.google.android.datatransport.runtime.backends.CreationContext;
+import com.google.android.datatransport.runtime.backends.TransportBackend;
+
+@Keep
+public class CctBackendFactory implements BackendFactory {
+  private static final String URL = mergeStrings("hts/pa.ogepscmlgbth", "tp:/lygolai.o/o/ac");
```
VinayGuthal: is this for obfuscation?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendFactory.java
vkryachko: Doing so will impose a requirement on the backend class to either:
1. have a certain static method e.g. like you suggested below `getInstance()`
2. have a constructor that takes certain arguments

which may not sound like a big deal, but it  violates the `single responsibility principle`, the TransportBackend is now responsible for communicating with the backend and for integrating with the discovery mechanism. Whereas having separate classes for this separates these concerns.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendRegistry.java
vkryachko: see above

transport/transport-backend-cct/src/main/AndroidManifest.xml
vkryachko: see above

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctBackendFactory.java
vkryachko: yep

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/CreationContextFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.google.android.datatransport.runtime.time.Clock;
+import com.google.android.datatransport.runtime.time.Monotonic;
+import com.google.android.datatransport.runtime.time.WallTime;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/** Factory that produces instances of {@link CreationContext}. */
+@Singleton
+final class CreationContextFactory {
```
ashwinraghav: I feel like this does not warrant a factory. Consider injecting the clocks directly into the Registry

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/TransportBackendDiscovery.java
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.android.datatransport.runtime.backends;
+
+import android.app.Service;
+import android.content.Intent;
+import android.os.IBinder;
+
+/** Dummy services needed to be able to declare transport backends in AndroidManifest.xml files. */
```
ashwinraghav: Could you explain this a bit more?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctBackendFactory.java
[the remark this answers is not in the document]
vkryachko: b/112610020

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendFactory.java
```diff
@@ -0,0 +1,20 @@
...
+public interface BackendFactory {
```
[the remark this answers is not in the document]
vkryachko: We don't have components here, this library does not depend on firebase in any way by design.

Not sure what you mean by interop mechanism, what hard dependency are you trying to avoid?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+    if (backends.containsKey(name)) {
+      return backends.get(name);
+    }
+    String backendProviderName = getBackendProviders().get(name);
+    if (backendProviderName == null) {
+      return null;
+    }
+
+    try {
+      BackendFactory provider =
```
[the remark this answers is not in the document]
vkryachko: There isn't any reason right now, but I don't think the api should encode this in any way. The only contract between the runtime and the backends is: runtime will call you when an instance of your backend is needed, the rest is unspecified to enable future changes

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/BackendRegistry.java
```diff
@@ -0,0 +1,133 @@
...
+          (BackendFactory)
+              Class.forName(backendProviderName).getDeclaredConstructor().newInstance();
```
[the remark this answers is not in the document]
vkryachko: Good point, will document it in the interface definition

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/CreationContext.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Values that are provided to {@link BackendFactory}s when a backend is instantiated.
+ *
+ * <p>Provides access to useful services that are provided by the runtime.
+ */
+@AutoValue
+public abstract class CreationContext {
+
+  /** Returns the backend name that runtime is requesting. */
+  public abstract String getBackendName();
```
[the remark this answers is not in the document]
vkryachko: good catch, tbh this is a left-over from an idea I had, I agree that it does not add value at this point, and we can always add it later if we need it. removed

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/CreationContextFactory.java
vkryachko: agreed, removed

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/TransportBackendDiscovery.java
vkryachko: done

PR #327, opened 2019-03-29
add uploader

VinayGuthal:
Add uploader which the scheduler will use to upload to our sqlite database.


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -14,13 +14,98 @@
...
+          } finally {
+            callback.run();
+          }
+        });
+  }
+
+  void logAndUpdateState(String backendName, int attemptNumber) {
+    TransportBackend backend = backendRegistry.get(backendName);
+    List<EventInternal> eventInternals = new ArrayList<>();
+    Iterable<PersistedEvent> persistedEvents =
```
ashwinraghav: If I understand correctly, the explicit synchronization is only required because we want to support multi process thread safety ?

Questions
1. The API for the datastore seems to require that call sites responsibly call eventStore methods in critical sections. I see that  this cannot be done within the implementations of the eventStore since we invoke others apis from within critical sections, that are outside of the store. Would a single operation on datastore have to be run in a critical section too?
2. If the critical sections are scattered across the project, I would vote to bring them all into a single file and encapsulate them as higher level functions

vkryachko: > require that call sites responsibly call eventStore methods in critical sections

It's not the case, you can call them by themselves they will just be part of their own isolated transaction, the reason you'd want to call them within a critical section is to guarantee atomicity.

2. iiu your comment correctly `runCriticalSection` is exactly the higher-level function you're describing, so I am not sure what the suggestion is here, can you be more specific, e.g. provide an example?

ashwinraghav: Ok. So it seems like each operation on datastore itself need not be in a critical section. THis is requires only when multiple operations are to be atomic.

2. What I mean is -> All invocations of `runCriticalSection` could be centralized. We will have a single place to reason about all higher level operations that we need to be atomic. Is this file that place or are there other files that invoke `runCtiticalSection` ?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -14,13 +14,98 @@
...
+    this.context = context;
+    this.backendRegistry = backendRegistry;
+    this.eventStore = eventStore;
+    this.workScheduler = workScheduler;
+    this.executor = executor;
+    this.guard = guard;
+  }
+
+  boolean isNetworkAvailable() {
+    ConnectivityManager connectivityManager =
```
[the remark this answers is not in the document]
VinayGuthal: I don't think so. I have made it double safe however and returned false in the scenario i am even unable to get the connectivity manager.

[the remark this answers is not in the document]
vkryachko: @VinayGuthal I don't think you should have done it, it's guaranteed to be not null, why have a null check then?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
vkryachko: @VinayGuthal this is what I was referring to in 1. Namely, this is what's called as a result of `transport.send(event)`.

ashwinraghav: Ok. Defer to your judgement

transport/transport-runtime/src/main/AndroidManifest.xml
```diff
@@ -14,6 +14,7 @@
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.android.datatransport.runtime">
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
vkryachko: You are missing a service declaration in the manifest, it will cause an exception at runtime.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/TransportRuntimeModule.java
```diff
@@ -67,6 +68,17 @@ static WorkScheduler workScheduler(
     }
   }
 
+  @Provides
+  static Uploader uploader(
```
vkryachko: You can remove this as the uploader is annotated with `@Inject`

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -14,13 +14,107 @@
...
+    for (PersistedEvent persistedEvent : persistedEvents) {
+      eventInternals.add(persistedEvent.getEvent());
+    }
+    BackendResponse response = backend.send(BackendRequest.create(eventInternals));
+    guard.runCriticalSection(
+        LOCK_TIMEOUT,
+        () -> {
+          if (response.getStatus() == BackendResponse.Status.OK) {
+            eventStore.recordSuccess(persistedEvents);
+            eventStore.recordNextCallTime(transportContext, response.getNextRequestWaitMillis());
```
vkryachko: this looks incorrect, instead of persisting the offset e.g. `2 minutes` you need to persist `clock.now() + 2 minutes`

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/JobInfoScheduler.java
```diff
@@ -81,7 +81,12 @@ public void schedule(TransportContext transportContext, int attemptNumber) {
...
-    long timeDiff = eventStore.getNextCallTime(transportContext) - clock.getTime();
+    Long backendTime = eventStore.getNextCallTime(transportContext);
+
+    long timeDiff = 0;
+    if (backendTime != null) {
+      timeDiff = backendTime - clock.getTime();
+    }
     // Schedule the build.
     PersistableBundle bundle = new PersistableBundle();
     bundle.putInt(SchedulerUtil.ATTEMPT_NUMBER, attemptNumber);
```
vkryachko: `setMinimumLatency()` below is wrong, you should not add offset to `clock.now()`.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/TransportRuntimeModule.java
vkryachko: Also please mark the executor above as `@Singleton`

transport/transport-runtime/src/main/AndroidManifest.xml
```diff
@@ -14,10 +14,16 @@
...
     <!--Although the *SdkVersion is captured in gradle build files, this is required for non gradle builds-->
     <!--<uses-sdk android:minSdkVersion="14"/>-->
 
     <application>
+        <service
+            android:name=".scheduling.jobscheduling.JobInfoSchedulerService"
+            android:label="JobScheduler Service"
+            android:permission="android.permission.BIND_JOB_SERVICE" >
```
vkryachko: please add `android:exported="false"`, also it seems we don't need the label(none of our services/content-providers are labeled)

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/TransportRuntimeModule.java
vkryachko: > You can remove this as the uploader is annotated with `@Inject`
> Also please mark the executor above as `@Singleton`

These 2 are still not done, please change

PR #336, opened 2019-04-04
Add a limit on the maximum amount of storage the `events` database uses.

vkryachko:
New events are dropped until old ones are discarded.


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -411,6 +426,40 @@ private static Priority toPriority(int value) {
...
+   * Gets the number of pages in the database file. Multiplying this with the page size yields the
+   * approximate size of the database on disk (including the WAL, if relevant).
+   */
+  private long getPageCount() {
+    return getDb().compileStatement("PRAGMA page_count").simpleQueryForLong();
+  }
+
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  @VisibleForTesting
+  void setMaxDbSizeOnDisk(long byteSize) {
```
ashwinraghav: Can this be injected instead of a setter?

PR #341, opened 2019-04-08
Synchronize access to AndroidConnectivityMonitor.callbacks

rsgowman:


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
```diff
@@ -99,15 +101,19 @@ public void run() {
...
+          callback.accept(NetworkStatus.REACHABLE);
+        }
       }
     }
 
     @Override
     public void onLost(Network network) {
-      for (Consumer<NetworkStatus> callback : callbacks) {
-        callback.accept(NetworkStatus.UNREACHABLE);
+      synchronized (callbacks) {
```
wilhuff: I can't comment there so consider this a comment on line 133.

Don't we also need to synchronize in `NetworkReceiver` for pre-N devices?

rsgowman: Yes, absolutely. Fixed/thanks.

PR #345, opened 2019-04-11
Re-implement 3p license inference implementation.

vkryachko:
The implementation relied on an oss gradle plugin. The plugin had
several issues:

* Did not respect project level dependencies
* Did not respect some gradle configurations that propagate to the pom,
  like runtimeOnly
* Had resolution issues where the order of task execution would affect
  the outcome of the license report
* Failed the build for projects that have debug buildType disabled.

### DIFF
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/GenerateLicensesTask.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/GenerateLicensesTask.groovy
@@ -15,20 +15,25 @@
 package com.google.firebase.gradle.plugins.license
 
 import groovy.json.JsonBuilder
-import groovy.json.JsonSlurper
+import groovy.transform.CompileStatic
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
+import org.gradle.api.artifacts.Configuration
 import org.gradle.api.tasks.*
 import org.gradle.api.tasks.incremental.IncrementalTaskInputs
 
+import javax.inject.Inject
+
+@CompileStatic
 class GenerateLicensesTask extends DefaultTask {
-    def NEW_LINE_LENGTH = "\n".getBytes().length;
+    private static final int NEW_LINE_LENGTH = "\n".getBytes().length
+
+    private final LicenseResolver licenseResolver
+    private final Configuration inputConfiguration
+
 
     @Input
     List<ProjectLicense> additionalLicenses
-
-    @InputFile
-    File licenseReportFile
 
     @InputDirectory
     File licenseDownloadDir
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/GenerateLicensesTask.groovy
```diff
@@ -15,37 +15,54 @@
...
 
+    @Inject
+    GenerateLicensesTask(Configuration configuration) {
+        this(configuration, new LicenseResolver())
+    }
+
+    GenerateLicensesTask(Configuration configuration, LicenseResolver licenseResolver) {
+        this.inputConfiguration = configuration
+        this.licenseResolver = licenseResolver
+        outputs.upToDateWhen { false }
```
ashwinraghav: We always run the task? Could you add a comment why?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/LicenseResolver.groovy
```diff
@@ -0,0 +1,91 @@
...
+            Project depProject = ((ProjectDependency) it).dependencyProject
+            Set<ProjectLicense> transitive = determineProjectLicenses(depProject.configurations.getByName(configuration.name))
+            return [determineLicense(depProject)] + transitive
+        }.findAll { it != null } as Set
+    }
+
+    private static Set<ProjectLicense> determineExternalLicenses(Project project, Configuration configuration) {
+        def conf = project.configurations.create('internalAllExternalArtifacts')
+
+        configuration.allDependencies.each {
```
ashwinraghav: Four our future selves, can you leave a comment on why these tactics are necessary?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/ProjectLicense.groovy
```diff
@@ -12,10 +12,69 @@
...
+            def name = project.name
+            def licenses = project.licenses.license.collect { URI.create(it.url as String) }
+            return new ProjectLicense(
+                    name: name,
+                    explicitLicenseUris: licenses,
+                    parent: determineParent(pomFile.parentFile.parentFile.parentFile.parentFile.parentFile, project.parent))
+        }
+    }
+
+    private static ProjectLicense determineParent(File baseDir, GPathResult parentNode) {
```
ashwinraghav: Worth leaving a comment on why we need to look up licenses from the parent for some libs

buildSrc/src/test/groovy/com/google/firebase/gradle/plugins/license/LicenseResolverPluginSpec.groovy
```diff
@@ -80,7 +80,7 @@ class LicenseResolverPluginSpec extends Specification {
```
ashwinraghav: Wow! So the test actually served a purpose and passess?

vkryachko: yep, with one minor capitalization change

PR #349, opened 2019-04-11
use contexts package itself

VinayGuthal:


transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerSchedulerTest.java
```diff
@@ -55,8 +54,10 @@ private Intent getIntent() {
     Uri.Builder intentDataBuilder = new Uri.Builder();
     intentDataBuilder.appendQueryParameter(
         SchedulerUtil.BACKEND_NAME, TRANSPORT_CONTEXT.getBackendName());
+    assertThat(context.getPackageName()).isEqualTo("org.robolectric.default");
```
vkryachko: we should not test robolectric's behavior

PR #350, opened 2019-04-11
Version bumps for next release

ashwinraghav:


firebase-functions/gradle.properties
```diff
@@ -1,3 +1,3 @@
-version=16.2.0
-latestReleasedVersion=16.1.3
+version=16.3.1
+latestReleasedVersion=16.3.0
```
vkryachko: this seems odd, can you verify it is intended?

ashwinraghav: confirmed

PR #352, opened 2019-04-11
Metrics Collection Refactoring (1)

yifanyang:
Move apksize and coverage gradle tasks defination and plugins to different packages.


buildSrc/src/test/groovy/com/google/firebase/gradle/plugins/measurement/apksize/ApkSizeTestProject.groovy
```diff
@@ -154,10 +156,7 @@ class ApkSizeTestProject extends ExternalResource {
         def dest = buildSrc.resolve(src)
 
         Files.createDirectories(dest)
-        Files.list(src)
-            .map { path -> path.getFileName().toString() }
-            .filter { name -> name.endsWith(".groovy") }
-            .forEach { name -> Files.copy(src.resolve(name), dest.resolve(name)) }
+        FileUtils.copyDirectory(src.toFile(), dest.toFile())
```
allisonbm92: I'm surprised this works, since it comes from Android. I'd prefer to keep the Java interface. 

Is there something you're trying to achieve with this change?

yifanyang: Changed to use `org.apache.commons.io.FileUtils` now.

PR #353, opened 2019-04-12
Remove dependency on functions emulator.

vkryachko:
This makes it possible to execute the tests in FTL.


firebase-functions/src/androidTest/functionsForTests/.firebaserc
```diff
@@ -0,0 +1 @@
+{}
```
bklimt: Every other platform calls this directory `backend`, rather than `functionsForTests`. Could we change it, for consistency?

firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java
```diff
@@ -97,16 +85,15 @@ public void testToken() throws InterruptedException, ExecutionException {
...
-                return Tasks.forResult(context);
-              }
+            () -> {
+              HttpsCallableContext context = new HttpsCallableContext("token", null);
+              return Tasks.forResult(context);
             });
-    useTestURL(functions);
+    // 'Authorization' is a special header that does not reach the function if it does not contain a
+    // valid token.
+    functions.replaceAuthorizationHeaderForTest("Test-Authorization");
```
bklimt: Are you sure? The test functions aren't implemented as callables, so I don't think the token gets checked. We haven't had to do this on iOS or JavaScript, so I'd really like to not have to add it here.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -191,6 +195,18 @@ public void useFunctionsEmulator(String origin) {
     urlFormat = origin + "/%2$s/%1$s/%3$s";
   }
 
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  @VisibleForTesting
+  void replaceAuthorizationHeaderForTest(String headerName) {
```
bklimt: As mentioned above, I doubt this is really necessary.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -191,6 +195,18 @@ public void useFunctionsEmulator(String origin) {
...
+    authorizationHeader = checkNotNull(headerName, "Null header name");
+  }
+
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  @VisibleForTesting
+  void replaceIidHeaderForTest(String headerName) {
```
bklimt: Surely this isn't necessary? The IID token isn't used for anything automatically.

firebase-functions/src/androidTest/functionsForTests/.firebaserc
vkryachko: done

firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java
vkryachko: You're right, I don't know why it was failing for me on Friday, now it is passing. Removed.

PR #354, opened 2019-04-15
Defer SharedPrefs read until it is needed.

vkryachko:
The motivation behind the change is to:
* Fix crash when application is started in direct-boot mode(b/116149732).
* Avoid a disk read at application startup, which causes strict mode
  violation(b/128433971). Note that this does not completely solve the
  issue as eager SDKs can access this value which will cause a disk
  read.


firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -551,33 +533,13 @@ protected FirebaseApp(Context applicationContext, String name, FirebaseOptions o
...
-            && applicationInfo.metaData.containsKey(DATA_COLLECTION_DEFAULT_ENABLED)) {
-          return applicationInfo.metaData.getBoolean(DATA_COLLECTION_DEFAULT_ENABLED);
-        }
-      }
-    } catch (PackageManager.NameNotFoundException e) {
-      // This shouldn't happen since it's this app's package, but fall through to default if so.
-    }
-    return true;
+    dataCollectionConfigStorage =
+        new Lazy<>(
```
davidmotson: I think I'm confused about the use of lazy in this context over just provider. 

Whats gained? Is it that lazy is memoized and we expect this provider to be called multiple times?

vkryachko: Yeah, the key is that we want to memoize the value once it is created, but only want to create it when it is needed and not earlier

PR #356, opened 2019-04-15
Metrics Collection Refactoring (2)

yifanyang:
- Rename coverage gradle tasks.
- Assign apk size gradle tasks into a new category "ApkSize"
  to make them discoverable via `./gradlew tasks`.
- Misc clean up.

### DIFF
--- a/tools/measurement/apksize/apksize.gradle
+++ b/tools/measurement/apksize/apksize.gradle
@@ -33,6 +33,9 @@ apply from: "src/functions/functions.gradle"
  * @param -Ppull_request the pull request number to be included in the report
  */
 task generateMeasurements(type: GenerateMeasurementsTask) {
+    description 'Builds all supported variants and calculates APK sizes.'
+    group 'ApkSize'
+
     sdkMapFile = file("sdks.csv")
     reportFile = file("$buildDir/size-report.json")
 }
@@ -44,6 +47,9 @@ task generateMeasurements(type: GenerateMeasurementsTask) {
  * @param -Ppull_request the pull request number to be included in the report
  */
 task uploadMeasurements(type: UploadMeasurementsTask) {
+    description 'Uploads APK size measurements to a database via the uploader tool.'
+    group 'ApkSize'
+
     dependsOn generateMeasurements
 
     reportFile = file("$buildDir/size-report.json")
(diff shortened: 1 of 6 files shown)


tools/measurement/apksize/apksize.gradle
```diff
@@ -33,6 +33,9 @@ apply from: "src/functions/functions.gradle"
  * @param -Ppull_request the pull request number to be included in the report
  */
 task generateMeasurements(type: GenerateMeasurementsTask) {
+    description 'Builds all supported variants and calculates APK sizes.'
+    group 'ApkSize'
```
allisonbm92: I think we should use a single group for all metrics. Either Metrics or Measurements sounds good to me.

tools/measurement/coverage/coverage.gradle
```diff
@@ -13,20 +13,22 @@
...
 
-import com.google.firebase.gradle.plugins.measurement.coverage.GenerateCoveragePercentsTask
+import com.google.firebase.gradle.plugins.measurement.coverage.GenerateMeasurementsTask
 import com.google.firebase.gradle.plugins.measurement.UploadMeasurementsTask
 
 
-task generateMeasurements(type: GenerateCoveragePercentsTask) {
+task generateMeasurements(type: GenerateMeasurementsTask) {
     description 'Runs checkCoverage task in all projects and calculates coverage percents.'
     group 'Coverage'
```
allisonbm92: Same here.

tools/measurement/apksize/apksize.gradle
```diff
@@ -44,6 +47,9 @@ task generateMeasurements(type: GenerateMeasurementsTask) {
  * @param -Ppull_request the pull request number to be included in the report
  */
 task uploadMeasurements(type: UploadMeasurementsTask) {
```
allisonbm92: FYI: We should probably have only one uploadMeasurements task in tools/measurement. We can do this in a future pull request. Then, Prow will only need to run the one task and all of our other metrics will be gathered.

yifanyang: Agreed.

Also, they need to be separate at the moment, if we want to use different database config for apk size and coverage.

tools/measurement/apksize/apksize.gradle
yifanyang: Done. They are all within group "Measurements" now.

Note: I have to rename the tasks as `{generate, upload}{ApkSize, Coverage}Measurements` since tasks with duplicate names under same category do not show properly with `./gradlew tasks`. This also means we need to update the prow config accordingly.

PR #358, opened 2019-04-16
Add `android:exported` manifest lint check.

vkryachko:
The intent is to make sure teams don't forget to explicitly set the
attribute on Android components.

### DIFF
--- a/tools/lint/src/main/kotlin/LintChecks.kt
+++ b/tools/lint/src/main/kotlin/LintChecks.kt
@@ -1,0 +1,65 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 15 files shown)


tools/lint/src/main/kotlin/LintChecks.kt
```diff
@@ -0,0 +1,65 @@
...
+        "Leaving this attribute out may unintentionally lead to an exported component, please " +
+                "specify the value explicitly.",
+        Category.SECURITY,
+        1,
+        Severity.ERROR,
+        Implementation(ManifestElementHasNoExportedAttributeDetector::class.java, Scope.MANIFEST_SCOPE))
+
+class ManifestElementHasNoExportedAttributeDetector : Detector(), XmlScanner {
+
+    override fun getApplicableElements(): Collection<String>? = listOf("provider", "receiver", "service")
```
allisonbm92: Should we add Activity?

vkryachko: Done

PR #371, opened 2019-04-23
Add integration tests

VinayGuthal:


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/TransportRuntime.java
```diff
@@ -83,6 +84,12 @@ public static TransportRuntime getInstance() {
     return localRef.getTransportRuntime();
   }
 
+  @VisibleForTesting
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  static void setInstance(TransportRuntimeComponent component) {
```
vkryachko: I'd suggest you replace this with a higher-order function like:

```java
static void withInstance(TransportRuntimeComponent component, Runnable runnable) {
  TransportRuntimeComponent original;
  synchronized (TransportRuntime.class) {
      original = INSTANCE;
      INSTANCE = component;
  }
  try {
    runnable.run();
  } finally {
    synchronized (TransportRuntime.class) {
      INSTANCE = original;
    }
  }

}
```

This will make sure that, no matter what failure happens, the runtime is restored to its original state. Otherwise there's big potential for cascading test failures that run after such a failed test.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/SchedulingConfigModule.java
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package com.google.android.datatransport.runtime.scheduling;
+
+import com.google.android.datatransport.runtime.scheduling.jobscheduling.SchedulerConfig;
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+public abstract class SchedulingConfigModule {
```
vkryachko: Do you need this module? it looks like config is injected via the builder

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/SchedulerConfig.java
```diff
@@ -0,0 +1,42 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.android.datatransport.runtime.scheduling.jobscheduling;
+
+import javax.inject.Inject;
+
+public class SchedulerConfig {
```
vkryachko: Please make this an autovalue, and since we don't want to expose `delta`, you can make it package-private.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/SchedulerConfig.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private long delta;
+  private long maxAllowedTime;
+  private long maximumDelay;
+
+  @Inject
+  public SchedulerConfig(long delta, long maxAllowedTime, long maximumDelay) {
```
vkryachko: An `@Autovalue.Builder` would be nice as well, because `new SchedulerConfig(1,2,3)` is not very readable at call sites.

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+@RunWith(AndroidJUnit4.class)
+public class SchedulerIntegrationTest {
+  private static final String TEST_KEY = "test";
+  private static final String TEST_VALUE = "test-value";
+  private static final String testTransport = "testTransport";
+  private TransportInternal transportInternalMock = mock(TransportInternal.class);
+  private TransportBackend mockBackend = mock(TransportBackend.class);
+  private TransportBackend mockBackend2 = mock(TransportBackend.class);
+  private BackendRegistry mockRegistry = mock(BackendRegistry.class);
```
vkryachko: Can you make all of the above final as well?

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,185 @@
...
+                invocation ->
+                    invocation
+                        .<EventInternal>getArgument(0)
+                        .toBuilder()
+                        .addMetadata(TEST_KEY, TEST_VALUE)
+                        .build());
+    runtime = TransportRuntime.getInstance();
+  }
+
+  String getBackendName() {
```
vkryachko: a better name for this method would be `generateBackendName()`, wdyt?

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,185 @@
...
+    byte[] array = new byte[8]; // length is bounded by 8
+    new Random().nextBytes(array);
+    return new String(array, Charset.forName("UTF-8"));
```
vkryachko: This will produce unreadable strings, can you `base64` encode them? or maybe just use uuid4, which will be longer but readable out of the box

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,185 @@
...
+        EventInternal.builder()
+            .setEventMillis(3)
+            .setUptimeMillis(1)
+            .setTransportName(testTransport)
+            .setPayload("TelemetryData".getBytes())
+            .build();
+    transport.send(stringEvent);
+    verify(mockBackend, times(1)).decorate(eq(expectedEvent));
+    try {
+      TimeUnit.SECONDS.sleep(5);
```
vkryachko: Try using `SystemClock.sleep()` instead, that way you won't have to deal with interrupts.

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  String getBackendName() {
+    byte[] array = new byte[8]; // length is bounded by 8
+    new Random().nextBytes(array);
+    return new String(array, Charset.forName("UTF-8"));
+  }
+
+  @Test
+  public void firstAttemptSchedule() {
```
vkryachko: The test naming pattern elsewhere in the codebase is `methodOrSystemUnderTest_whenConditions_shouldDoStuff`. Could you rename them to the same pattern pls? e.g. `scheduler_whenEventScheduledForFirstTime_shouldUploadEventWithinExpectedTime` or something

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/SchedulingConfigModule.java
VinayGuthal: we need this for the normal working?

vkryachko: my bad :) thought it was for tests, but now I see it is in the main sourceSet

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,180 @@
...
+            .build();
+    transport.send(stringEvent);
+    TransportFactory factory2 = runtime.newFactory(secondBackendName);
+    Transport<String> transport2 =
+        factory2.getTransport(testTransport, String.class, String::getBytes);
+    transport2.send(stringEvent);
+    verify(mockBackend, times(1)).decorate(eq(expectedEvent));
+    verify(mockBackend2, times(1)).decorate(eq(expectedEvent));
+    SystemClock.sleep(5000);
+    verify(mockUploader, times(2)).upload(any(), eq(1), any());
```
vkryachko: let's verify backend names for each call instead of having `times(2)` with `any()`

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/TransportRuntimeRule.java
```diff
@@ -0,0 +1,45 @@
...
+  public Statement apply(final Statement base, final Description description) {
+    return new Statement() {
+      @Override
+      public void evaluate() throws Throwable {
+        TransportRuntime.withInstance(
+            component,
+            () -> {
+              try {
+                base.evaluate();
+              } catch (Throwable throwable) {
```
vkryachko: I think this catch defeats the purpose of the tests as all tests will pass no matter what. I think the easiest way to fix would be to switch to a `Callable` in `withInstance()` and change its signature to something like:
```java
static <T> T withInstance(Component cmp, Callable<T> callable) throws Throwable
```

transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerSchedulerTest.java
```diff
@@ -47,8 +47,9 @@
   private final EventStore store = new InMemoryEventStore();
   private final AlarmManager alarmManager =
       spy((AlarmManager) context.getSystemService(Context.ALARM_SERVICE));
+  private final SchedulerConfig config = new SchedulerConfig(30000, 100000000, -1);
```
vkryachko: This won't compile, will it?

transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/JobInfoSchedulerTest.java
```diff
@@ -40,7 +40,9 @@
   private final EventStore store = new InMemoryEventStore();
   private final JobScheduler jobScheduler =
       (JobScheduler) context.getSystemService(Context.JOB_SCHEDULER_SERVICE);
-  private final JobInfoScheduler scheduler = new JobInfoScheduler(context, store, () -> 1);
+
+  private final SchedulerConfig config = new SchedulerConfig(30000, 100000000, -1);
```
vkryachko: same here

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/TransportRuntime.java
```diff
@@ -83,6 +85,24 @@ public static TransportRuntime getInstance() {
     return localRef.getTransportRuntime();
   }
 
+  @VisibleForTesting
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  static void withInstance(TransportRuntimeComponent component, Callable callable)
```
vkryachko: You should not use raw generic types, please change to `Callable<Void>`

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/SchedulingConfigModule.java
```diff
@@ -0,0 +1,32 @@
...
+
+  @Provides
+  static SchedulerConfig config() {
+    return SchedulerConfig.builder()
+        .setDelta(30000)
+        .setMaxAllowedTime(100000000)
```
vkryachko: can you extract delta and maxAllowed to constants that have meaningful names? e.g `THIRTY_SECONDS` and whatever that 1 hundred million is(or maybe change it to `TWENTY_FOUR_HOURS`

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/SchedulerIntegrationTest.java
```diff
@@ -0,0 +1,194 @@
...
+                    invocation
+                        .<EventInternal>getArgument(0)
+                        .toBuilder()
+                        .addMetadata(TEST_KEY, TEST_VALUE)
+                        .build());
+    doAnswer(
+            (Answer<Void>)
+                i -> {
+                  locker.setResult(true);
+                  i.<Runnable>getArgument(2).run();
```
vkryachko: Please document the above two lines:
* what the locker is needed for and how it synchronizes the tests with the service/broadcast receiver
* why we need to the the passed-in runnable.

Issue #370, opened 2019-04-22
Bug-Upgrading Cloud Firestore SDK causes crash loop for some users


highthunder:
### [READ] Step 1: Are you in the right place?

Yes, this is an issue with Cloud Firestore's internal on-device sqlite database and this repository appears to contain all of the code required to manage that database.

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.4
  * Firebase Component: Cloud Firestore
  * Component version: 17.0.2 upgrading to 18.2.0
  * Gradle version: 4.10.1
  * com.android.tools.build:gradle:3.3.2
  * Compile/Target SDK version: 28

### [REQUIRED] Step 3: Describe the problem

Immediately after releasing an app update which included the library upgrade we started to see users crash looping, seemingly starting when the app first restarted after updating.

Crashlytics Stack Trace
```
# Crashlytics - plaintext stacktrace downloaded by Developer at Mon, 22 Apr 2019 19:19:01 GMT
# URL: https://fabric.io/maven-machines-inc/android/apps/com.mavenmachines/issues/5c6af9dbf8b88c29634c7cec/sessions/5CBE11CA008800014C5BB70A3A4E0045_DNE_0_v2?
# Organization: Maven Machines, Inc.
# Platform: android
# Application: Maven Machines
# Version: 3.28.0-7 (177)
# Bundle Identifier: com.mavenmachines
# Issue ID: 5c6af9dbf8b88c29634c7cec
# Session ID: 5CBE11CA008800014C5BB70A3A4E0045_DNE_0_v2
# Date: 2019-04-22T19:11:00Z
# OS Version: 7.1.1
# Device: Galaxy Tab E 8.0
# RAM Free: 41.8%
# Disk Free: 81.2%

#0. Crashed: main
       at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$5(AsyncQueue.java:379)
       at com.google.firebase.firestore.util.AsyncQueue$$Lambda$5.run(AsyncQueue.java)
       at android.os.Handler.handleCallback(Handler.java:751)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6823)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1557)
[...]

PR #374, opened 2019-04-24
Firestore: Split schema migrations into batches

schmidt-sebastian:
Fixes: https://github.com/firebase/firebase-android-sdk/issues/370

When we migrate schemas with large datasets, we need to split our operations into multiple batches. Otherwise, we can exceed SQLite's CursorWindow.

This PR splits the three operations that potentially iterate through large datasets into batches. I have added tests that verify that the batching works, but I was not able to reproduce the original issue using the unit test. 

With this PR, I have been able to successfully migrate a schema in a test app that stored 10,000 documents. Without it, this migration fails with:

```
   Caused by: java.lang.RuntimeException: android.database.sqlite.SQLiteBlobTooBigException: Row too big to fit into CursorWindow requiredPos=15935, totalRows=4065
        at com.google.firebase.firestore.util.AsyncQueue.lambda$enqueue$3(com.google.firebase:firebase-firestore@@18.2.0:290)
        at com.google.firebase.firestore.util.AsyncQueue$$Lambda$3.run(Unknown Source:4)
...
```

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Unreleased
+- [fixed] Fixed an issue that prevented schema migrations for clients with
+  large offline datasets (#370, #734).
+
+# To be released - 19.0.0
 - [feature] You can now query across all collections in your database with a
   given collection ID using the `FirebaseFirestore.collectionGroup()` method.
 - [changed] The garbage collection process for on-disk persistence that
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 - [fixed] Fixed an issue that prevented schema migrations for clients with
-  large offline datasets (#370, #734).
+  large offline datasets (#370).
 
 # To be released - 19.0.0
 - [feature] You can now query across all collections in your database with a
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
@@ -51,13 +51,13 @@ class SQLiteSchema {
   static final int INDEXING_SUPPORT_VERSION = VERSION + 1;
 
   /**
-   * The batch size for operations that potentially operate on a large result set.
+   * The batch size for the sequence number migration in `ensureSequenceNumbers()`.
    *
    * <p>This addresses https://github.com/firebase/firebase-android-sdk/issues/370, where a customer
    * reported that schema migrations failed for clients with thousands of documents. The number has
    * been chosen arbitrarily.
    */
-  static final int REMOTE_DOCUMENTS_BATCH_SIZE = 1000;
+  private static final int SEQUENCE_NUMBER_BATCH_SIZE = 1000;
 
   private final SQLiteDatabase db;
 
@@ -376,7 +376,7 @@ class SQLiteSchema {
                     + "SELECT TD.path FROM target_documents AS TD "
                     + "WHERE RD.path = TD.path AND TD.target_id = 0"
                     + ") LIMIT ?")
-            .binding(REMOTE_DOCUMENTS_BATCH_SIZE);
+            .binding(SEQUENCE_NUMBER_BATCH_SIZE);
 
     boolean[] resultsRemaining = new boolean[1];
 
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
@@ -54,9 +54,10 @@ class SQLiteSchema {
    * The batch size for operations that potentially operate on a large result set.
    *
    * <p>This addresses https://github.com/firebase/firebase-android-sdk/issues/370, where a customer
-   * reported that schema migrations failed for clients with thousands of documents.
+   * reported that schema migrations failed for clients with thousands of documents. The number has
+   * been chosen arbitrarily.
    */
-  static final int REMOTE_DOCUMENTS_BATCH_SIZE = 100;
+  static final int REMOTE_DOCUMENTS_BATCH_SIZE = 1000;
 
   private final SQLiteDatabase db;
 
@@ -218,10 +219,6 @@ class SQLiteSchema {
           String uid = mutationQueueEntry.getString(0);
           long lastAcknowledgedBatchId = mutationQueueEntry.getLong(1);
 
-          // Unlike other migrations, removeAcknowledgedMutations() does not use batching. Since
-          // this query only returns acknowledged mutations that have not yet been committed to
-          // the `remote_documents` table, the total number of entries should fit within a single
-          // operation.
           SQLitePersistence.Query mutationsQuery =
               new SQLitePersistence.Query(
                       db, "SELECT batch_id FROM mutations WHERE uid = ? AND batch_id <= ?")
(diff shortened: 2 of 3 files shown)


firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Unreleased
+- [fixed] Fixed an issue that prevented schema migrations for clients with
+  large offline datasets (#370, #734).
```
mikelehen: What does 734 refer to?  https://github.com/firebase/firebase-android-sdk/issues/734 is a 404

Maybe you meant 374, which is this PR, but I'm not sure if that's particularly helpful?  Presumably this PR will be associated with issue 370 anyway, so folks can find their way to it easily enough if they really want to dig into it?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -50,6 +50,14 @@
   // Remove this constant and increment VERSION to enable indexing support
   static final int INDEXING_SUPPORT_VERSION = VERSION + 1;
 
+  /**
+   * The batch size for operations that potentially operate on a large result set.
+   *
+   * <p>This addresses https://github.com/firebase/firebase-android-sdk/issues/370, where a customer
+   * reported that schema migrations failed for clients with thousands of documents.
+   */
+  static final int REMOTE_DOCUMENTS_BATCH_SIZE = 100;
```
mikelehen: Is there some reason 100 is guaranteed to be safe?  From the code / comments I can't tell if it was chosen arbitrarily or if there's a specific rationale behind it.

mikelehen:
I guess to ask a pointed question that might help me understand what's going on, do you know what limit from https://www.sqlite.org/limits.html we're hitting?  Or is it a limit coming from somewhere else?

schmidt-sebastian:
I am also not able to find any documentation on limits that we are hitting, and the exception is thrown deep in the native SQLite code. In the meantime, I did come up with a repro though:

```
SQLiteOpenHelper opener =
        new SQLiteOpenHelper(this.getApplicationContext(), "foo", null, 1) {
            @Override
            public void onCreate(SQLiteDatabase db) {
            }

            @Override
            public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
            }
        };

db = opener.getWritableDatabase();

db.execSQL("DROP TABLE IF EXISTS a");
db.execSQL("DROP TABLE IF EXISTS b");

db.execSQL("CREATE TABLE a (path TEXT PRIMARY KEY)");
db.execSQL("CREATE TABLE b (path TEXT PRIMARY KEY)");

for (int i = 0; i < 100000; ++i) {
    db.execSQL("INSERT INTO a (path) VALUES ('foo" + i + "')");
}

Cursor cursor = db.rawQuery("SELECT path FROM a WHERE NOT EXISTS (SELECT 1 FROM b WHERE path = a.path)", new String[]{});
while (cursor.moveToNext()) {
    db.execSQL("INSERT INTO b (path) VALUES ('" + cursor.getString(0) + "')");
}
```

I am not able repro this without the `WHERE NOT EXISTS`. To me, this is starting to look like a SQLite bug (they should at least be throwing a different exception). I'd argue that we can probably drop the pagination from the v8 migration, but we might want to keep it for `ensureSequenceNumbers()`.

firebase-firestore/CHANGELOG.md
schmidt-sebastian: I removed the link to the PR. In general, I have been linking to both in changelogs, but the more common precedent in other SKDs seems to be to link to neither. One links seems to be a good comprise :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: This number is arbitrary. I picked a smaller number to make the unit tests run faster. I bumped it to 1000, which I believe to be safe as well since I have only ever hit this condition with 10k+ documents.

I will verify this change in another manual test run.

schmidt-sebastian: Test run succeeded with 1000.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -50,6 +50,15 @@
...
   static final int INDEXING_SUPPORT_VERSION = VERSION + 1;
 
+  /**
+   * The batch size for the sequence number migration in `ensureSequenceNumbers()`.
+   *
+   * <p>This addresses https://github.com/firebase/firebase-android-sdk/issues/370, where a customer
+   * reported that schema migrations failed for clients with thousands of documents. The number has
+   * been chosen arbitrarily.
+   */
+  private static final int SEQUENCE_NUMBER_BATCH_SIZE = 1000;
```
mikelehen: According to https://cloud.google.com/firestore/quotas, the maximum length of a path (document name) is 6 KiB.  Can you try your repro with 6K paths and the 1000 batch size and make sure it still doesn't hit the error?

Also, if we were to change this number in the future, how would one verify it is safe?  Our automated tests aren't enough, right?  So maybe throw your repro code in a gist and link it from here so somebody in the future could repro?  So then this comment would change to "This number was chosen somewhat arbitrarily but has been verified to avoid the reported issue, using the repro code at [your gist]." or whatever.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -359,17 +368,30 @@ private void ensureSequenceNumbers() {
...
+
+      untaggedDocumentsQuery.forEach(
+          row -> {
+            resultsRemaining[0] = true;
+            tagDocument.clearBindings();
+            tagDocument.bindString(1, row.getString(0));
+            tagDocument.bindLong(2, sequenceNumber);
+            hardAssert(tagDocument.executeInsert() != -1, "Failed to insert a sentinel row");
+          });
+    } while (resultsRemaining[0]);
```
mikelehen: This will always do at least one extra query.  You could avoid that by counting the rows and comparing to SEQUENCE_NUMBER_BATCH_SIZE.  But my guess is you know this and chose to keep it simpler, which is also fine.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteSchemaTest.java
```diff
@@ -307,6 +307,34 @@ public void addsSentinelRows() {
             });
   }
 
+  @Test
+  public void addsSequenceNumberInBatches() {
```
mikelehen: Technically this isn't verifying that it does it in batches...  and I think it's similar to the existing addsSentinelRows test?  So would it make sense to call it "addsSentinelRowsForManyDocuments" or something?

Also, probably worth a PORTING NOTE explaining why this test exists (presumably we won't port it to the other platforms).  I'd also call out that while this test exists to verify we can handle lots of documents, it *CANNOT* be used to reproduce the original issue (since it doesn't reproduce with robolectric or whatever).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: My verification steps use FriendlyEats and code that inserts documents in batches. There is no specific magic and most of the work revolves around restarting the client with different version of the SDK.  I don't feel it's worth linking to a code snippet. 
From my experience with SQLite, it also seems impossible to come up with a number that "just works" and increasing the document name to 6 KB ended up throwing a different error (" java.lang.IllegalStateException: Couldn't read row 114, col 0 from CursorWindow.  Make sure the Cursor is initialized correctly before accessing data from it.") with this batch size. Without further ado, I reduced the batch size back to 100, which works even when I use 6KB document names.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: I don't think it is worth the extra code to handle this case, since this will only ever run once.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteSchemaTest.java
schmidt-sebastian: Renamed the test to `addsSentinelRowsForLargeNumberOfDocuments` and added both a porting note and the additional comment that describes the limitations.

PR #375, opened 2019-04-24
Throw error in subsequent function calls after the client is shutdown

thebrianchen:
Context: b/38324305
JS SDK implementation: https://github.com/firebase/firebase-js-sdk/pull/1695

Please check that I used `verifyNotShutdown()` in the appropriate places. I wasn't sure since some of the public methods like `handleRemoteEvent()` are not present in the js api.


firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -255,33 +269,45 @@ private void initialize(Context context, User user, boolean usePersistence, long
...
 
+  private void verifyNotShutdown() {
+    if (this.isShutdown) {
+      throw new IllegalArgumentException("The client has already been shutdown");
+    }
+  }
+
   @Override
   public void handleRemoteEvent(RemoteEvent remoteEvent) {
+    this.verifyNotShutdown();
```
mikelehen: These methods are public in order to implement the `RemoteStoreCallback` interface and I don't think we should guard them with `verifyNotShutdown()`. In particular, if they get called after shutdown that is because the SDK is calling them, not because the user is calling them, and so throwing an exception is unhelpful and actually misleading (it's really a bug in the SDK).  If we *wanted*, we could `hardAssert(!this.isShutdown, ...);` to verify this never happens, but I'm inclined to keep it simple for now and just remove these.

PR #377, opened 2019-04-25
Update Readme to indicate google-services.plist is required

schmidt-sebastian:
We now require that the `google-services.plist` exists prior to importing the SDK from gradle. This PR updates the readme.


schmidt-sebastian:
This might not be needed after https://github.com/firebase/firebase-android-sdk/commit/036de23bae8d79b5c582c19f0d067550723e4c0f

Issue #382, opened 2019-04-27
firebase/firestore: getting internal error from AsyncQueue


mgenov:
### Version Information
  * Firebase Component: Firestore
  * Component version: 18.2.0
  * Affected Android Versions: 4.2.2, 5.1.1

### Description

It happens rarely and this is the stack trace from crashlytics:

```
Fatal Exception: java.lang.RuntimeException: Internal error in Firestore (18.2.0).
       at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$5(com.google.firebase:firebase-firestore@@18.2.0:379)
       at com.google.firebase.firestore.util.AsyncQueue$$Lambda$5.run(com.google.firebase:firebase-firestore@@18.2.0)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:135)
       at android.app.ActivityThread.main(ActivityThread.java:5254)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:902)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:697)
```

```
Caused by android.database.CursorWindowAllocationException: Cursor window allocation of 2048 kb failed. 
       at android.database.CursorWindow.<init>(CursorWindow.java:108)
       at android.database.AbstractWindowedCursor.clearOrCreateWindow(AbstractWindowedCursor.java:198)
       at android.database.sqlite.SQLiteCursor.fillWindow(SQLiteCursor.java:139)
       at android.database.sqlite.SQLiteCursor.getCount(SQLiteCursor.java:133)
       at android.database.AbstractCursor.moveToPosition(AbstractCursor.java:197)
       at android.database.AbstractCursor.moveToNext(AbstractCursor.java:245)
       at com.google.firebase.firestore.local.SQLitePersistence$Query.forEach(com.google.firebase:firebase-firestore@@18.2.0:430)
       at com.google.firebase.firestore.local.SQLiteRemoteDocumentCache.getAll(com.google.firebase:firebase-firestore@@18.2.0:98)
       at com.google.firebase.firestore.local.LocalStore.lambda$applyRemoteEvent$5(com.google.firebase:firebase-firestore@@18.2.0:385)
[...]

PR #386, opened 2019-04-29
Enable strict mode to detect leaks of SQLite objects

wilhuff:
In particular this will detect leaked Cursors in Firestore's integration tests.

This is in response to #382, which suggests leaked cursors as a possible culprit, but this shows that's not the case.


wilhuff:
> LGTM. I assume this is pretty cheap (won't significantly increase our test runtime)?

I couldn't perceive any difference in test execution time.

PR #387, opened 2019-04-29
Add clearPersistence()

thebrianchen:


firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -377,6 +383,22 @@ public static void setLoggingEnabled(boolean loggingEnabled) {
...
 
+  /**
+   * Clears the persistent storage.
+   *
+   * <p>Must be called while the client is not started (after the app is shutdown or when the app is
+   * first initialized). On startup, this method must be called before other methods (other than
+   * settings()). If the client is still running, an exception with a code of `failed-precondition`
+   * will be thrown.
+   */
+  @PublicApi
```
wilhuff: This would release the feature. For now, remove `@PublicApi` and `public` to make this package private.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
```diff
@@ -270,6 +260,7 @@ public static FirebaseFirestore testFirestore(
             new EmptyCredentialsProvider(),
             asyncQueue,
             /*firebaseApp=*/ null);
+    SQLitePersistence.clearPersistence(context, databaseId, persistenceKey);
```
wilhuff: Use the public API for this.

In the other comment I recommended making the public API package-private. To make it accessible, add a `clearPersistence` method like `shutdown` to `AccessHelper`.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -377,6 +383,22 @@ public static void setLoggingEnabled(boolean loggingEnabled) {
     }
   }
 
+  /**
+   * Clears the persistent storage.
+   *
+   * <p>Must be called while the client is not started (after the app is shutdown or when the app is
+   * first initialized). On startup, this method must be called before other methods (other than
+   * settings()). If the client is still running, an exception with a code of `failed-precondition`
```
wilhuff: settings is called `setFirestoreSettings` on Android. Errors have a `Code` that's an a proper enumeration here.

Make sure that any code literals you give are valid on the platform you're using.

Javadoc doesn't support backticks. Use `@code` annotations. (See elsewhere in our public API files for examples.)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -202,6 +203,15 @@ void runTransaction(String action, Runnable operation) {
     return value;
   }
 
+  public static void clearPersistence(
+      Context context, DatabaseId databaseId, String persistenceKey) {
+    String databaseName = SQLitePersistence.databaseName(persistenceKey, databaseId);
+    String sqlLitePath = context.getDatabasePath(databaseName).getPath();
+    String journalPath = sqlLitePath + "-journal";
+    new File(sqlLitePath).delete();
```
wilhuff: Previously these had warnings about ignoring a return value. Now that this is production code we probably shouldn't ignore anymore. Specifically if we run into permission problems or something where the file still exists after trying to delete it, this call should throw.

Use `java.nio.file.Path#delete()` instead and only fail here if the error is not file not found.

wilhuff: Actually, whoops! `java.nio.file.Path` Was added in API 26 and we need to support back to API 16. We still need to use the old way, but if there’s an error check to see if the file still exists. If still exists then throw

wilhuff: Actually, whoops! `java.nio.file.Path` Was added in API 26 and we need to support back to API 16. We still need to use the old way, but if there’s an error check to see if the file still exists. If still exists then throw

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
```diff
@@ -270,6 +259,11 @@ public static FirebaseFirestore testFirestore(
             new EmptyCredentialsProvider(),
             asyncQueue,
             /*firebaseApp=*/ null);
+    try {
+      AccessHelper.clearPersistence(firestore);
+    } catch (Exception e) {
```
wilhuff: This catch clause is over broad and should be avoided unless it's required. See http://go/java-practices/exceptions#catch_exception.

Meanwhile note that any stack trace that comes from clearPersistence failing is going to include it so there isn't much value in catching this exception only to (indirectly) rethrow. It's less code to just `waitFor(clearPersistence(firestore));` and it will be just as obvious what's failing if it doesn't work.

I realize you're following the pattern of the AsyncQueue constructor above, but it's unwarranted there as well. Feel free to remove that try/catch block also.

thebrianchen:
Thanks for the detailed review Gil! Your explanations and examples about SecurityException and general Exceptions were very helpful. I did have a face palm moment when I realized what you were getting at for the true purpose of FileUtil.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/FileUtil.java
```diff
@@ -0,0 +1,55 @@
...
+    } else {
+      LegacyFileDeleter.delete(file);
+    }
+  }
+
+  /** Deletes a file if it exists. Only used on API levels >= 26. */
+  @TargetApi(Build.VERSION_CODES.O)
+  private static class DefaultFileDeleter {
+    public static void delete(File file) throws IOException {
+      try {
```
rsgowman: I'm not sure why you're catching the exception, only to immediately rethrow it? Could the body of this method be simplified to:
```
public static void delete(File file) throws IOException {
  Files.deleteIfExists(file.toPath());
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -202,14 +207,31 @@ void runTransaction(String action, Runnable operation) {
     return value;
   }
 
+  public static void clearPersistence(Context context, DatabaseId databaseId, String persistenceKey)
+      throws FirebaseFirestoreException {
+    String databaseName = SQLitePersistence.databaseName(persistenceKey, databaseId);
+    String sqLitePath = context.getDatabasePath(databaseName).getPath();
+    String journalPath = sqLitePath + "-journal";
```
rsgowman: I think you'll also need to delete the WAL. (sqlitePath + "-wal"). This only applies to android 9/pie and newer devices, but should be safe enough to do unconditionally.

https://source.android.com/devices/tech/perf/compatibility-wal
https://www.sqlite.org/tempfiles.html#write_ahead_log_wal_files

PR #397, opened 2019-05-01
Add gradle tasks for collecting prow jobs latency info.

yifanyang:


tools/measurement/latency/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+      });
+  });
+}
+
+function isAndroidSdkJob(prowJob: ProwJob): boolean {
+  return TEST_TARGET_ID_MAP.has(prowJob.spec.job);
+}
+
+function convertProwJobToReplaceMeasurement(job: ProwJob): ReplaceMeasurement {
+  // test_id
```
Corrob: I don't find these comments helpful since it's just repeating the variable name.

tools/measurement/latency/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  const testId = job.metadata.name;
+  // test_target_id
+  const testTargetId = TEST_TARGET_ID_MAP.get(job.spec.job);
+  // test_start_time, test_end_time, test_duration
+  const startTime = moment(job.status.startTime);
+  const endTime = moment(job.status.completionTime);
+  const format = 'YYYY-MM-DD HH:mm:ss';
```
Corrob: I think it would be good to make this a constant at the top of the file: DATE_FORMAT.

tools/measurement/latency/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+        reject(`Query k8s cluter failed with error "${error}".`);
+      });
+  });
+}
+
+function isAndroidSdkJob(prowJob: ProwJob): boolean {
+  return TEST_TARGET_ID_MAP.has(prowJob.spec.job);
+}
+
+function convertProwJobToReplaceMeasurement(job: ProwJob): ReplaceMeasurement {
```
Corrob: Maybe make ReplaceMeasurement a class with a method exporting to the proper format?

tools/measurement/latency/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  status: {
+    startTime: string;
+    completionTime: string;
+    state: string;
+  };
+}
+
+export interface ReplaceMeasurement extends Array<string | number> {}
+
+export interface Table {
```
Corrob: Similar to my comments above, I think we could put some of those functions above here to be more object oriented.

tools/measurement/latency/src/index.ts
yifanyang: Done.

tools/measurement/latency/src/index.ts
yifanyang: Done.

tools/measurement/latency/src/index.ts
yifanyang: Done.

tools/measurement/latency/src/types.ts
yifanyang: Done.

tools/measurement/latency/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+function isAndroidSdkJob(prowJob: ProwJob): boolean {
+  return TEST_TARGET_ID_MAP.has(prowJob.spec.job);
+}
+
+function constructReport(prowJobs: ProwJob[]): Report {
+  /* Temporarily allow snake case naming for report object assembly. */
+  /* tslint:disable:variable-name */
+  const table_name = 'TestResults';
+  const column_names = [
```
allisonbm92: What schema is this? This is different from the schema that I made.

tools/measurement/latency/src/test_target_id_map.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Map } from 'immutable';
+
+export const TEST_TARGET_ID_MAP: Map<string, number> = Map({
```
allisonbm92: Ideally, we should make this a file that's not embedded in the source code.

tools/measurement/latency/src/test_target_id_map.ts
```diff
@@ -0,0 +1,32 @@
...
+  'postsubmit-gob-sync': 1,
```
allisonbm92: Some of these run on post-submit, and others, are non-blocking pre-submit. It may be useful to gather these, but we need to exclude them from total pre-submit time.

tools/measurement/latency/src/types.ts
```diff
@@ -0,0 +1,100 @@
...
+    type: string;
+  };
+  status: {
+    startTime: string;
+    completionTime: string;
+    state: string;
+  };
+}
+
+export class ReplaceMeasurement extends Array<string | number> {
```
allisonbm92: I'm not sure this should extend Array. It's probably best as a proper class (with fields) or as a function that simply returns the array. Can you add a method to ProwJob that converts it into the measurement array? That's another option.

tools/measurement/latency/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+    'test_id',
+    'test_target_id',
+    'test_start_time',
+    'test_end_time',
+    'test_duration',
```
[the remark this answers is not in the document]
yifanyang: Done.

tools/measurement/latency/src/index.ts
yifanyang: Right. I intend to make some change to the schema. The new column `test_id` (an UUID to specify a test run) might be most confusing. The reason is that we are querying prow regularly, and might get info for a prow job that we already got and uploaded from previous query. This column helps prevent us from uploading duplicate records.

What do you think? Cory and I discussed about it yesterday. You were not here so didn't get a chance to check with you.

tools/measurement/latency/src/test_target_id_map.ts
yifanyang: Then will it also be checked into the repo?

Another reason to have it here is that prow jobs for many repos are running on the same cluster, and we need to have a way to tell whether a job is for Android SDK or other project.

tools/measurement/latency/src/test_target_id_map.ts
yifanyang: I think eventually we will need to collect them, e.g. for greenness calculation. Do you know how I can tell whether a presubmit job is a blocking one or a non-blocking one?

tools/measurement/latency/src/types.ts
yifanyang: I'm not sure I agree. Isn't `ReplaceMeasurement` essentially an array?

JavaScript object and the its JSON serialization are almost the same syntactically. For example, `{'a': 123, 'b': 'hello world!'}` is both a valid JSON object and a valid JavaScript object literal. I think it's more burden to process mentally if we break this association, and replace it with our customized way of JSON serialization for an object.

yifanyang: After all "JSON" is short for "JavaScript Object Notation".

yifanyang:
> 1. This code fetches prow jobs from k8s api server, so why is this in this repo in the first place? Can't this be in a dedicated docker container configured to run on a periodic basis(to host such code we have a `test-infra` repo).

I think I see your point. Unlike apk size or coverage, the concept and the implementation for latency data collection rely on prow being the underlying CI system, and some names of prow jobs used in the code are coming from the prow config actually. For that it makes sense to have it located in test-infra.

But with that, the metrics collection code would be living in two repos, and as you mentioned, we will need to duplicate the upload logic. I don't feel strongly about either option. What would you suggest?

> 2. Why is gradle involved? Seems like the typescript binary can be invoked directly and the upload task is simple enough to implement in typescript directly

Gradle is not necessarily required. It's just wrapped with gradle so that it is in accordance with other metrics collection tasks. With regard to upload task, if there's no special considerations, I think I would prefer to reuse the code. What do you think?

vkryachko:
@yifanyang The fact that some metric collection lives in this repo makes sense to me, as the measurements they collect are based on the state of the codebase they are part of.

In this case it's different, you're just measuring how long Prow jobs take and iiuc you don't even need to check out this repository as the prow job definition has all required metadata.

So it seems like the only argument to put this code here is to be able to reuse the upload task, but again it seems to be pretty simple: download a jar, execute it.

yifanyang:
@vkryachko I think I agree with your statement. Will close this pull request and open a new one in our `test-infra` repo.

PR #401, opened 2019-05-02
Update APK Size Tool to use release builds.

allisonbm92:
Gradle is now building both the release and aggressive variants, so we need to accommodate that.

### DIFF
--- a/tools/measurement/apksize/sdks.csv
+++ b/tools/measurement/apksize/sdks.csv
@@ -1,4 +1,8 @@
 database-aggressive:4
+database-release:8
 firestore-aggressive:1
+firestore-release:5
 functions-aggressive:2
 storage-aggressive:3
+functions-release:6
+storage-release:7


tools/measurement/apksize/sdks.csv
```diff
@@ -2,3 +2,7 @@ database-aggressive:4
 firestore-aggressive:1
 functions-aggressive:2
 storage-aggressive:3
+database-release:8
```
Corrob: Why not "4"? Or move to the end of the file to preserve order?

allisonbm92: "4" is also used, but it was cut-off by the preview. Originally, they were in alphabetical order, but I messed up when inserting them into the database. I added the four new ones in the same order as before (alphabetical) but after all of the release options. I can re-order them in any way. What do you think?

Corrob: Got it. That is funny how 4 got perfectly cut out. My preference is fully alphabetical, but I don't have a strong opinion. Like:

database-aggressive:4
database-release:8
firestore-aggressive:1
firestore-release:5
functions-aggressive:2	
functions-release:6
storage-aggressive:3	
storage-release:7

Also, should these be comma's instead of colons since it's a csv file?

allisonbm92: Okay, they're alphabetical now. It's csv as in colon separated values.

PR #406, opened 2019-05-03
Initial refactor of the plugin structure.

vkryachko:
The short-term/immediate goal is to introduce a place for projects to
indicate whether javadoc needs to be generated for them, it's done here
via `firebaseLibrary.publishJavadoc`.

The long-term vision is to move all build and release logic be
controlled via the 'firebase-library' plugin instead of it being
injected into individual projects from the root-project.


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -0,0 +1,54 @@
+package com.google.firebase.gradle.plugins;
+
+import com.android.build.gradle.LibraryExtension;
+import com.google.common.collect.ImmutableMap;
+import com.google.firebase.gradle.plugins.ci.device.FirebaseTestServer;
+// Copyright 2019 Google LLC
```
allisonbm92: Shouldn't this be on the first line?

vkryachko: Oops, fixed

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override
+  public void apply(Project project) {
+    project.apply(ImmutableMap.of("plugin", "com.android.library"));
+
+    FirebaseLibraryExtension firebaseLibrary =
+        project
+            .getExtensions()
+            .create("firebaseLibrary", FirebaseLibraryExtension.class, project.getObjects());
+
+    LibraryExtension android = project.getExtensions().getByType(LibraryExtension.class);
```
ashwinraghav: Do we know that this is going to be the case for all libs like well known types and collections?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestLabPlugin.java
```diff
@@ -0,0 +1,40 @@
...
+import org.gradle.api.Project;
+
+/**
+ * Firebase Test Lab Integration plugin.
+ *
+ * @deprecated For library projects use 'firebase-library' plugin instead. App projects should still
+ *     use this plugin.
+ */
+@Deprecated
+public class FirebaseTestLabPlugin implements Plugin<Project> {
```
ashwinraghav: Why should we deprecate this instead of just removing it?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
ashwinraghav: Seems like we do :)

vkryachko: Ack

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestLabPlugin.java
vkryachko: Because application projects still use it, e.g. fiamui-app. I did not want to spend too much more time making the new plugin support apps as well. I will leave it as future work

PR #409, opened 2019-05-07
Automate KTX release process and version tracking

vkryachko:


yifanyang:
Please ignore my activity of closing and reopening this pull request. Meant to close another one...

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryExtension.java
```diff
@@ -14,22 +14,67 @@
...
+  /** Indicates whether sources are published alongside the library. */
   public boolean publishSources;
+
+  /** Firebase Test Lab configuration/ */
   public final FirebaseTestLabExtension testLab;
 
   @Inject
-  public FirebaseLibraryExtension(ObjectFactory objectFactory) {
-    this.testLab = new FirebaseTestLabExtension(objectFactory);
+  public FirebaseLibraryExtension(Project project) {
```
ashwinraghav: I would go out of the way to document that this is not equivalent to -HEAD dependencies that folks might be used to. The key distinction why it may not be extended to fit that use case might be that the libraries being co-released are at the exact same version.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -102,6 +105,10 @@ class PublishingPlugin implements Plugin<Project> {
                     publications {
                         mavenAar(MavenPublication) {
                             from sub.components.android
+                            if (sub.name == 'ktx') {
```
ashwinraghav: Why does this need to be explicitly set ? Can we not establish this build hierarchy in the respective gradle files ?

buildSrc/src/test/groovy/com/google/firebase/gradle/plugins/publish/PublishingPluginSpec.groovy
```diff
@@ -95,36 +102,13 @@ class PublishingPluginSpec extends Specification {
...
-        }
-        android.compileSdkVersion = 26
-        version = '0.9'
-        
-        dependencies {
-            implementation project(':childProject1')
-        }
-        """
-
     def "Publish all dependent projects succeeds"() {
```
ashwinraghav: Rename to "Publishing dependent projects succeeds"

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
vkryachko: It has to be here for now, as we are configuring all subprojects from the root project, I want to move a lot of this logic into FirebaseLibraryPlugin, but until then I added a todo here

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryExtension.java
vkryachko: I updated the javadoc of the `releaseWith` method.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryExtension.java
```diff
@@ -14,22 +14,67 @@
...
   }
 
-  void testLab(Action<FirebaseTestLabExtension> action) {
+  /** Configure Firebase Test Lab. */
+  public void testLab(Action<FirebaseTestLabExtension> action) {
     action.execute(testLab);
   }
+
+  /** Register to be released alongside another Firebase Library project. */
+  public void releaseWith(Project releaseWithProject) {
```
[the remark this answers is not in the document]
vkryachko: done

vkryachko:
> I think it is worth capturing that in addition of adding co-releasing build tools, we are also making the ktx projects children of their java equivalents (Are we?).

It is captured in one of my commit messages

PR #411, opened 2019-05-07
Add combined flavor to smoke tests

allisonbm92:
This adds a new build flavor to the smoke tests. The combined flavor includes the other flavors enabling us to test with a single executable and verify they build together. The individual variants can still be used to test products individually.


smoke-tests/src/combined/java/com/google/firebase/testing/combined/AllTests.java
```diff
@@ -0,0 +1,28 @@
+// Copyright 2018 Google LLC
```
yifanyang: 2019?

allisonbm92: I copied the header over from other files with the older date. I can update all of them in another change,

PR #419, opened 2019-05-08
Artifact publishing capability improvements.

vkryachko:
* Adds sources.jar artifact for opted in libraries
* Allows libraries to customize their pom.xml contents

### DIFF
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
@@ -14,10 +14,12 @@
 
 package com.google.firebase.gradle.plugins.publish
 
+import com.google.firebase.gradle.plugins.FirebaseLibraryExtension
 import digital.wup.android_maven_publish.AndroidMavenPublishPlugin
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.api.tasks.bundling.Jar
 import org.gradle.api.tasks.bundling.Zip
 
 /**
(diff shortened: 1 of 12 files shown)


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -88,8 +90,9 @@ class PublishingPlugin implements Plugin<Project> {
                 if (!sub.plugins.hasPlugin('firebase-library')) {
                     return
                 }
+                FirebaseLibraryExtension firebaseLibrary = sub.extensions.getByType(FirebaseLibraryExtension)
                 if (projectsToPublish.contains(sub)) {
-                    projectsToPublish.addAll(sub.firebaseLibrary.projectsToRelease)
+                    projectsToPublish.addAll(firebaseLibrary.projectsToRelease)
```
ashwinraghav: I thought we already did this?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -111,18 +114,13 @@ class PublishingPlugin implements Plugin<Project> {
...
-                                licenses {
-                                    license {
-                                        name = 'The Apache Software License, Version 2.0'
-                                        url = 'http://www.apache.org/licenses/LICENSE-2.0.txt'
-                                    }
-                                }
-                                scm {
-                                    connection = 'scm:git:https://github.com/firebase/firebase-android-sdk.git'
-                                    url = 'https://github.com/firebase/firebase-android-sdk'
+                            if (firebaseLibrary.publishSources) {
```
ashwinraghav: Thinking out loud:
While the pom files we generate here will link to sources, we are still dependent on our internal release process to publish sources, which may not be supported yet.

vkryachko: correct, it still needs to get plumbed through explicitly

PR #423, opened 2019-05-09
Alter a few tests to fix them within google3

rsgowman:
Without this change, these tests cause a:
"junit.framework.AssertionFailedError: android.content.res.Resources$NotFoundException: Unable to find resource ID #0x7f030000 "
error.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
@@ -15,6 +15,7 @@
 package com.google.firebase.firestore;
 
 import static com.google.firebase.firestore.AccessHelper.getAsyncQueue;
+import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollection;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollectionWithDocs;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testDocument;
@@ -33,12 +34,9 @@ import static junit.framework.Assert.assertTrue;
 import static junit.framework.Assert.fail;
 import static org.junit.Assert.assertNotEquals;
 
-import android.content.Context;
-import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
-import com.google.firebase.FirebaseApp;
 import com.google.firebase.Timestamp;
 import com.google.firebase.firestore.FirebaseFirestoreException.Code;
 import com.google.firebase.firestore.Query.Direction;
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
```diff
@@ -969,11 +967,9 @@ public void testMaintainsPersistenceAfterRestarting() {
     DocumentReference docRef = testDocument();
     FirebaseFirestore firestore = docRef.getFirestore();
     waitFor(docRef.set(map("foo", "bar")));
-    Context context = InstrumentationRegistry.getContext();
 
     waitFor(AccessHelper.shutdown(firestore));
-    FirebaseApp app2 = FirebaseApp.initializeApp(context);
-    FirebaseFirestore firestore2 = FirebaseFirestore.getInstance(app2);
+    FirebaseFirestore firestore2 = testFirestore();
```
wilhuff: This change breaks the intent of the test.

`testFirestore()` creates a new Firestore instance associated with a new Firestore app. This guarantees that we'll use a different persistence key that the one we actually use. This test will pass even if clearPersistence is unimplemented.

The purpose of what we're doing here is attempting to create a new Firestore instance associated with the same app as the one before.

rsgowman: > This test will pass even if clearPersistence is unimplemented.

Hmm... you're right. However, the *original* tests will also pass with an effectively empty clearPersistence. i.e.:
```
$ git diff
diff --git a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
index e935416..5021e17 100644
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -404,17 +404,8 @@ public class FirebaseFirestore {
     if (this.clientRunning) {
       throw new IllegalStateException("Persistence cannot be cleared while the client is running.");
     }
-    final TaskCompletionSource<Void> source = new TaskCompletionSource<>();
-    asyncQueue.enqueueAndForget(
-        () -> {
-          try {
-            SQLitePersistence.clearPersistence(context, databaseId, persistenceKey);
-            source.setResult(null);
-          } catch (FirebaseFirestoreException e) {
-            source.setException(e);
-          }
-        });
-    return source.getTask();
+
+    return Tasks.forResult(null);
   }
 
   FirestoreClient getClient() {
$ ANDROID_HOME=~/Android/Sdk ./gradlew :firebase-firestore:connectedCheck
<successful output here>
```

I'll look into this a bit more tomorrow.

rsgowman: Ok; I've reworked the tests. They now:
a) Properly fail when I noop clearPersistence (i.e. when I apply the above patch)
b) Pass within g3.

rsgowman:
Where you referring to the error within g3? If so, unfortunately I have *no* idea. :)

Something in the depths of gmscore is requesting an android resource when initializing firebaseapp for the second time, and for reasons that are unclear to me, failing to find it. Digging into it further looked like a pretty large time sink, so I skipped it.

PR #438, opened 2019-05-14
Fix for missing kotlin project coverage reports.

yifanyang:


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/coverage/GenerateMeasurementsTask.groovy
```diff
@@ -54,22 +54,25 @@ public class GenerateMeasurementsTask extends DefaultTask {
...
         try {
             def report = parser.parse(path)
-            def name = report.@name
             def lineCoverage = report.counter.find { it.@type == 'LINE' }
             if (lineCoverage) {
                 def covered = Double.parseDouble(lineCoverage.@covered.text())
                 def missed = Double.parseDouble(lineCoverage.@missed.text())
                 def percent = covered / (covered + missed)
-                return new Tuple(name, percent)
+                return percent
```
Corrob: How about just

return covered / (covered + missed)

?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/coverage/GenerateMeasurementsTask.groovy
```diff
@@ -84,12 +87,23 @@ public class GenerateMeasurementsTask extends DefaultTask {
...
+
+                // 'firebase-common-ktx' and 'firebase-firestore-ktx' has been
+                // made a subproject of their corresponding java projects in PR
+                // #409 and #426. Old mappings are kept for backward
+                // compatibility reason.
+                'firebase-common:ktx': 10,
+                'firebase-firestore:ktx': 11,
+
+                // 'firebase-storage:test-app' was added in PR #303.
+                'firebase-storage:test-app': 12
```
allisonbm92: This is a test app that is only used for instrumentation tests. I don't think we need to measure coverage for it.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/coverage/GenerateMeasurementsTask.groovy
```diff
@@ -54,22 +54,25 @@ public class GenerateMeasurementsTask extends DefaultTask {
...
                 def percent = covered / (covered + missed)
-                return new Tuple(name, percent)
+                return percent
             } else {
                 throw new IllegalStateException("Cannot find line coverage section in the report: $path.")
             }
         } catch (FileNotFoundException e) {
             project.logger.warn("Cannot find coverage report for project: $_project.")
-            return new Tuple(_project.name, 0)
+            return 0
```
allisonbm92: Should we treat this as an error?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/coverage/GenerateMeasurementsTask.groovy
yifanyang: Done.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/coverage/GenerateMeasurementsTask.groovy
yifanyang: Done. Removed.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/coverage/GenerateMeasurementsTask.groovy
yifanyang: Maybe.

I guess:
* if it is a problem with the coverage code itself, then it makes sense to throw an exception here;
* if it is due to an incorrectly configured project, then the code should just omit it and run through.

I haven't met a case in real life where this code path got hit, so as of now I don't know.

allisonbm92: What happens if the tests fail? Is there a coverage report? Or is that skipped due to the tests terminating early?

yifanyang: The report generation task depends on the unit test task. Should the unit test fail, gradle will stop early and won't get to the coverage report task.

allisonbm92: Right, that makes sense.

PR #444, opened 2019-05-16
Add support for Bill of Materials to smoke tests.

allisonbm92:
This change adds support to use a Bill of Materials to run the smoke tests. The Bill of Materials contains the versions of the Firebase libraries that should be used.

### DIFF
--- a/smoke-tests/configure.gradle
+++ b/smoke-tests/configure.gradle
@@ -12,4 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-enableFeaturePreview("IMPROVED_POM_SUPPORT")
+
+def configurePlatform() {
+  def bom = "com.google.firebase:firebase-bom:18.1.0"
+  if (project.hasProperty("firebase-bom")) {
+    bom = project.getProperty("firebase-bom")
+  }
+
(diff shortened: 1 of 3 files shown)


smoke-tests/configure.gradle
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+def configurePlatform() {
+  def bom = "com.google.firebase:firebase-bom:18.1.0"
+  if (project.hasProperty("firebase-bom")) {
+    bom = project.getProperty("firebase-bom")
+  }
```
Corrob: Personally, I prefer a ternary here. I like it when my variables are constant.

smoke-tests/configure.gradle
```diff
@@ -0,0 +1,33 @@
+// Copyright 2018 Google LLC
```
yifanyang: In JS SDK, we have a script that can fix the license header (or add if missing). It's also hooked with the git, so that whenever someone executes a git push, the script will get triggered automatically, fix the header and generate a new commit. We may consider to have something similar in Android SDK.

Issue #429, opened 2019-05-11
Error on uploading file to firebase cloud storage: “Listing objects in a bucket is disallowed for rules_version = 1”


fredzqm:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: TODO(Samuel Waxman)
  * Firebase Component: Storage
  * Component version: TODO(Samuel Waxman)

### [REQUIRED] Step 3: Describe the problem

When trying to upload an object.
```
"Listing objects in a bucket is disallowed for rules_version = "1". Please update storage security rules to rules_verison = "2" to use list."
```

#### Steps to reproduce:

TODO(fredzqm): add more detailed steps.

Reference:

https://stackoverflow.com/questions/56072204/error-on-uploading-file-to-firebase-cloud-storage-listing-objects-in-a-bucket

#### Relevant Code:

For more information, my upload code (in Kotlin) is
```
 fun uploadImage(uri: Uri, path: String): Task<Uri> {
            val storageRef = FirebaseStorage.getInstance().reference
            val storagePath = storageRef.child(path)
            return storagePath.putFile(uri).continueWithTask{ storageRef.downloadUrl }
 }
```
I call it with
```
public override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
        if (requestCode in 0..2 && resultCode == Activity.RESULT_OK) {
            val cropImageUri = CropImage.getActivityResult(data).uri
            val systemTime = System.currentTimeMillis()
            val path = "$userId/$systemTime"
            //IMAGE UPLOAD HERE:
            FirebaseImageResource.uploadImage(cropImageUri, path)
                    .addOnCompleteListener {
                if (it.isSuccessful) {
                    GlideApp.with(this)
                            .load(cropImageUri)
                            .into(imageViewForPosition(requestCode)!!)
                    imageUris[requestCode] = it.result.toString()
                }
            }
        }
[...]

PR #445, opened 2019-05-16
DownloadUrl at the root of a bucket throws proper error message

fredzqm:
#429

StorageReference.getDownloadUrl will throw if called on a bucket root reference.


fredzqm:
@schmidt-sebastian

I got this hacked together without the dev setup.
Please take a look, it should be simple, but addresses many people's confusion over the error message.

See #429

firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
```diff
@@ -43,7 +43,9 @@
 
     this.storageRef = storageRef;
     this.pendingResult = pendingResult;
-
+    if (storageRef.getRoot().getName().equals(storageRef.getName())) {
+      throw new IllegalArgumentException("Cannot get download url at the root of a bucket.");
```
schmidt-sebastian: Can you change this to use the method name (getDownloadUrl())? Furthermore, does the same problem exists if a user calls getMetadata()?

schmidt-sebastian: Can you make the error message something like: "getDownloadUrl() is not supported at the root of the Storage bucket."

firebase-storage/src/main/java/com/google/firebase/storage/GetMetadataTask.java
```diff
@@ -41,6 +41,9 @@
 
     this.mStorageRef = storageRef;
     this.mPendingResult = pendingResult;
+    if (storageRef.getRoot().getName().equals(storageRef.getName())) {
+      throw new IllegalArgumentException("Cannot get object metedata at the root of a bucket.");
+    }
```
schmidt-sebastian: Can you make the error message something like: "getMetadata() is not supported at the root of the Storage bucket."

fredzqm: Done

firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
fredzqm: Done

firebase-storage/CHANGELOG.md
```diff
@@ -4,3 +4,6 @@
   and are not mentioned in our documentation.
 - [internal] Updated the SDK initialization process and removed usages of
   deprecated methods.
+- [changed] Added validation to `StorageReference.getDownloadUrl` and 
+  `StorageReference.getMetadata` to return error if the reference is the root
```
schmidt-sebastian: s/to return error/to return an error/ :)

PR #446, opened 2019-05-16
Clean up FieldValueTests

wilhuff:
Reorganize to make things clearer:
  * Split out the tests that are really exercising the `UserDataConverter` into a `UserDataConverterTest`.
  * Move the Timestamp tests that are unrelated to `FieldValue` into `TimestampTest`

Also fix nullability annotations on `Timestamp`.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/Timestamp.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/Timestamp.java
@@ -18,6 +18,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import android.os.Parcel;
 import android.os.Parcelable;
+import android.support.annotation.NonNull;
 import com.google.firebase.annotations.PublicApi;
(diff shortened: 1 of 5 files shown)


firebase-firestore/src/main/java/com/google/firebase/Timestamp.java
```diff
@@ -37,7 +38,7 @@
 @PublicApi
 public final class Timestamp implements Comparable<Timestamp>, Parcelable {
 
-  @PublicApi
+  @PublicApi @NonNull
```
mikelehen: Newline?

wilhuff: I had a newline here but `googleJavaFormat` eated it.

mikelehen: How rude!

PR #449, opened 2019-05-20
Add runtime measurement for Perf spec tests

schmidt-sebastian:


firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -1063,14 +1063,20 @@ public void testSpecTests() throws Exception {
         Set<String> tags = getTestTags(testJSON);
 
         boolean runTest = shouldRunTest(tags) && (!exclusiveMode || tags.contains(EXCLUSIVE_TAG));
+        boolean measureRuntime = tags.contains(EXCLUSIVE_TAG);
```
wilhuff: Is there something in the spec that indicates it's a performance spec test? It seems like it would be better to tag them that way so that we get this by default, and could even build something to track that performance over time. Conflating this with "exclusive" means we'll never be able to do that.

schmidt-sebastian: I did mean to use the "benchmark" tag. I updated this PR.

PR #452, opened 2019-05-21
Implement firebase-functions-ktx Kotlin library.

rlazo:


firebase-functions/ktx/ktx.gradle
```diff
@@ -0,0 +1,58 @@
...
+        multiDexEnabled true
+        targetSdkVersion project.targetSdkVersion
+        versionName version
+    }
+    sourceSets {
+        main.java.srcDirs += 'src/main/kotlin'
+        test.java {
+            srcDir 'src/test/kotlin'
+            srcDir '../src/testUtil/java'
+            srcDir '../src/roboUtil/java'
```
vkryachko: I don't think you need these 2 lines, right?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.java
```diff
@@ -14,15 +14,18 @@
...
 public class HttpsCallableResult {
   // The actual result data, as generic types decoded from JSON.
   private final Object data;
 
-  HttpsCallableResult(Object data) {
+  HttpsCallableResult(@Nullable Object data) {
     this.data = data;
   }
 
+  @Nullable
```
vkryachko: @bklimt Is this correct semantically? can the returned object be null?

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -187,7 +193,7 @@ URL getURL(String function) {
    *
    * @param origin The origin of the local emulator, such as "http://10.0.2.2:5005".
    */
-  public void useFunctionsEmulator(String origin) {
+  public void useFunctionsEmulator(@NonNull String origin) {
     urlFormat = origin + "/%2$s/%1$s/%3$s";
```
vkryachko: might make sense to add a precondition check on `origin`?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.java
bklimt: Yes, that's correct.

bklimt:
Awesome!

I noticed there aren't any tests that actually call a Function from Kotlin. It might be nice to have some just to verify that the API works as intended. They could call some of the same functions that the Java integration tests call, such as `dataTest`. See `firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java`

I'm also curious whether we could make it so that instead of having to do `f.call(data)`, we could make it so that `f(data)` works in Kotlin. I don't think that's necessary for this PR, but it would be nice to know if it's possible.

vkryachko:
@bklimt `f(data)` is a great idea imo, and is definitely possible via an extension method. @rlazo can you add it?

firebase-functions/ktx/ktx.gradle
rlazo: You're right, removed.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
rlazo: Done.

rlazo:
@bklimt @vkryachko I addressed the comments. I made HttpsCallableReference invokable with extensions. PTAL

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
```diff
@@ -0,0 +1,101 @@
...
+import android.support.test.InstrumentationRegistry
+import android.support.test.runner.AndroidJUnit4
+
+@RunWith(AndroidJUnit4::class)
+class CallTests {
+    companion object {
+        lateinit var app: FirebaseApp
+
+        @BeforeClass @JvmStatic fun setup() {
+            FirebaseApp.initializeApp(InstrumentationRegistry.getContext())
```
vkryachko: can be changed to `app = Firebase.initialize(InstrumentationRegistry.getContext())` ?

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
```diff
@@ -0,0 +1,101 @@
...
+            "bool" to true,
+            "int" to 2,
+            "long" to 3L,
+            "string" to "four",
+            "array" to listOf(5, 6),
+            "null" to null
+        )
+
+        var function = functions.getHttpsCallable("dataTest")
+        val result = function(input)
```
vkryachko: looks great!

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
```diff
@@ -0,0 +1,101 @@
...
+        val actual = Tasks.await(result).getData()
+
+        assertThat(actual is Map<*, *>).isTrue()
```
vkryachko: Can be simplified as `assertThat(actual).isInstanceOf(Map::class.java)` ?

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
vkryachko: Also I think it is a good idea to `app.delete()` in the `@AfterClass` method, otherwise adding more test classes will make tests fail

rlazo: Done. Good catch with the missing app.delete()

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
rlazo: You're right, much better. Thanks!

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
rlazo: Also, I needed to force it to fail when null (!!) because without it the compiler complains about getting a FirebaseApp? when expecting FirebaseApp.

firebase-functions/ktx/src/androidTest/kotlin/com/google/firebase/functions/ktx/CallTests.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.support.test.InstrumentationRegistry
+import android.support.test.runner.AndroidJUnit4
+
+@RunWith(AndroidJUnit4::class)
+class CallTests {
+    companion object {
+        lateinit var app: FirebaseApp
+
+        @BeforeClass @JvmStatic fun setup() {
+            app = FirebaseApp.initializeApp(InstrumentationRegistry.getContext())!!
```
vkryachko: nit: `s/FirebaseApp.initializeApp/Firebase.initialize/`

rlazo: Done.

vkryachko:
@bklimt are talking specifically about how gms.Task failure should be propagated? If so, for tasks developers are ancouraged to use task coroutine integration where you can just use task.await() from async kotlin methods. Otherwise error handling is identical to java. wdyt?

firebase-functions/ktx/ktx.gradle
```diff
@@ -0,0 +1,59 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+plugins {
+    id 'firebase-library'
+    id 'kotlin-android'
+}
+
+firebaseLibrary {
```
vkryachko: Oh, actually. Before you merge pls add `ftl.enabled = true`. to make sure instrumentation tests run in CI.

bklimt:
@vkryachko No, I meant is there anything we should do to the `FirebaseFunctionsException` class or `FirebaseFunctionsException.Code` enum to make them more Kotlin friendly? I doubt there is, but want to make sure we consider the question.

https://firebase.google.com/docs/reference/android/com/google/firebase/functions/FirebaseFunctionsException.html

vkryachko:
@bklimt Yeah, we don't plan to do anything with the exceptions, afaik there is nothing kotlin specific we can do, and also can be added later if it comes up.

firebase-functions/ktx/src/main/kotlin/com/google/firebase/functions/ktx/Functions.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/** Returns the [FirebaseFunctions] instance of a given [FirebaseApp]. */
+fun Firebase.functions(app: FirebaseApp): FirebaseFunctions = FirebaseFunctions.getInstance(app)
+
+/** Returns the [FirebaseFunctions] instance of a given [FirebaseApp] and [region]. */
+fun Firebase.functions(app: FirebaseApp, region: String): FirebaseFunctions =
+        FirebaseFunctions.getInstance(app, region)
+
+/** Runs the callable. */
+operator fun HttpsCallableReference.invoke() = this.call()
```
rlazo: Hi @the-dagger @rosariopfernandes @SUPERCILEX I would like to get your feedback on this use of operator overload. It enables this use case

```
var myFunction = functions.getHttpsCallable("myFunction")
myFunction(data)
```

instead of 

```
var myFunction = functions.getHttpsCallable("myFunction")
myFunction.call(data)
```

Does it feel kotlin-idiomatic, or it may produce unwanted confusion?

SUPERCILEX: Hmmm, I think that might be overdoing it. Usually, the invoke overload is used to eliminate a method call that takes a lambda:
```
// Without
gradleTask.configure { ... }

// Overload
gradleTask { ... }
```

Would be curious to hear what other people think though.

rosariopfernandes: > Does it feel kotlin-idiomatic?

I'm not sure, because most times I find it being used with lambda functions, as @SUPERCILEX mentioned. But I like the fact that this makes it more similar to the JavaScript SDK.

> may produce unwanted confusion?

I don't think so, since `call` is the only action you can take from an instance of `HttpsCallableReference`.

rlazo: Thanks for the feedback @rosariopfernandes, @SUPERCILEX. I'll remove the overload as I tend to agree with you about using invoke mostly for cases that include lambdas as argument.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -187,7 +193,10 @@ URL getURL(String function) {
    *
    * @param origin The origin of the local emulator, such as "http://10.0.2.2:5005".
    */
-  public void useFunctionsEmulator(String origin) {
+  public void useFunctionsEmulator(@NonNull String origin) {
+    if (origin == null) {
+      throw new IllegalArgumentException("origin cannot be null");
```
rosariopfernandes: nit: Maybe use `Preconditions.checkNotNull()` instead? Since that's the approach used in `getInstance()`

rlazo: Agreed. I'll fix this one, which I added, together with the in `call()` in a follow up PR. Thanks!

PR #453, opened 2019-05-21
Wuandy/readme.md

Hui-Wu:
Add more instructions specific to Firestore to firebase-firestore/README.md


firebase-firestore/README.md
```diff
@@ -10,12 +10,81 @@ responsive apps that work regardless of network latency or Internet
...
+
+Run:
+```bash
+./gradlew -PprojectsToPublish="firebase-firestore" firebasePublish
+```
+
+This will build a `m2repository.zip` under `build/`, which contains a locally
+built Firestore SDK.
+
+This way you can pick up updates/fixes before official releases.
```
rsgowman: It may be worth stating (or linking to) brief instructions that detail how to use the m2repository.zip file. This is something we occasionally recommend to end users, and having a default set of instructions for that in our readme would allow us to simply link here.

firebase-firestore/README.md
```diff
@@ -10,12 +10,81 @@ responsive apps that work regardless of network latency or Internet
...
+```
+
+See [here](../README.md#code-formatting) if you want to be able to format code
+from within Android Studio.
+
+## Build Local Jar of Firestore SDK
+
+Run:
+```bash
+./gradlew -PprojectsToPublish="firebase-firestore" firebasePublish
```
rsgowman: Double check that this publishes the dependencies to the m2repository.zip file too. In particular, I think we want to see firebase auth and core. If they're not present, then any app that uses the locally built firestore, will end up using the publicly released firebase-auth/core dependencies. 99% of the time, that'll work, but may cause badness if auth/core has introduced a breaking change that is present in our repo, but hasn't been published yet.

PR #456, opened 2019-05-22
Add DocumentId annotation, mark planned interface changes and call site updates.

Hui-Wu:
This PR add a DocumentId annotation, and mark places where class/method interfaces need to be changed in order to accommodate populating and validating the annotated fields user might send to firestore api.

It also makes CustomClassMapper a `normal` class instead of an utility class, such that we don't need to pass documentId through all of the recursion calls.

The actual implementation/tests PR will come later.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -125,7 +125,8 @@ public class CollectionReference extends Query {
   public Task<DocumentReference> add(@NonNull Object data) {
     checkNotNull(data, "Provided data must not be null.");
     final DocumentReference ref = document();
-    return ref.set(data)
+    // TODO: Assert fields annotated with DocumentId in `data` are set to null.
+    return ref.setForAddition(data)
         .continueWith(
             Executors.DIRECT_EXECUTOR,
             task -> {
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -125,8 +125,7 @@ public class CollectionReference extends Query {
   public Task<DocumentReference> add(@NonNull Object data) {
     checkNotNull(data, "Provided data must not be null.");
     final DocumentReference ref = document();
-    // TODO: Assert fields annotated with DocumentId in `data` are set to null.
-    return ref.setForAddition(data)
+    return ref.set(data)
         .continueWith(
             Executors.DIRECT_EXECUTOR,
             task -> {
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
@@ -1,0 +1,43 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore;
+
+import com.google.firebase.annotations.PublicApi;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to mark an object field to be the ID of a firestore document.
+ *
+ * <p>This annotation is recognized by {@link com.google.firebase.firestore.util.CustomClassMapper}.
+ *
+ * <p>During conversions from documents to java objects, fields with this annotation will be
+ * populated with the document ID being converted.
+ *
+ * <p>When objects with the annotation is used to create new documents, its field value must be null
+ * to guarantee uniqueness, otherwise a runtime exception will be thrown.
+ *
+ * <p>When objects with the annotation is used to update documents, its field value must match the
+ * target documents, otherwise a runtime exception will be thrown.
+ *
+ * <p>This annotation can only be applied to fields of String or {@link DocumentReference},
+ * otherwise a runtime exception will be thrown.
+ */
+@PublicApi
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface DocumentId {}
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
@@ -21,21 +21,14 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Annotation used to mark an object field to be the ID of a firestore document.
- *
- * <p>This annotation is recognized by {@link com.google.firebase.firestore.util.CustomClassMapper}.
- *
- * <p>During conversions from documents to java objects, fields with this annotation will be
- * populated with the document ID being converted.
- *
- * <p>When objects with the annotation is used to create new documents, its field value must be null
- * to guarantee uniqueness, otherwise a runtime exception will be thrown.
- *
- * <p>When objects with the annotation is used to update documents, its field value must match the
- * target documents, otherwise a runtime exception will be thrown.
+ * Annotation used to mark a POJO field to be automatically populated with the document's ID when
+ * the POJO is created from a Firestore document (e.g. via {@link DocumentSnapshot#toObject}).
  *
  * <p>This annotation can only be applied to fields of String or {@link DocumentReference},
  * otherwise a runtime exception will be thrown.
+ *
+ * <p>When writing a POJO to Firestore, the @DocumentId-annotated field must either be null or match
+ * the document ID of the document being written to, else a runtime exception will be thrown.
  */
 @PublicApi
 @Retention(RetentionPolicy.RUNTIME)
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
@@ -154,13 +154,6 @@ public class DocumentReference {
     return set(data, SetOptions.OVERWRITE);
   }
 
-  // Handles specific logic for document addition before entering comment `set` logic.
-  /* package */ Task<Void> setForAddition(@NonNull Object data) {
-    checkNotNull(data, "Provided data must not be null.");
-    // TODO: Assert fields annotated with DocumentId in `data` are set to null.
-    return setImpl(data, SetOptions.OVERWRITE);
-  }
-
   /**
    * Writes to the document referred to by this DocumentReference. If the document does not yet
    * exist, it will be created. If you pass {@link SetOptions}, the provided data can be merged into
@@ -176,16 +169,10 @@ public class DocumentReference {
   public Task<Void> set(@NonNull Object data, @NonNull SetOptions options) {
     checkNotNull(data, "Provided data must not be null.");
     checkNotNull(options, "Provided options must not be null.");
-    // TODO: Assert fields annotated with DocumentId in `data` match `documentRef`.
-    return setImpl(data, options);
-  }
-
-  @NonNull
-  private Task<Void> setImpl(@NonNull Object data, @NonNull SetOptions options) {
     ParsedSetData parsed =
         options.isMerge()
-            ? firestore.getDataConverter().parseMergeData(data, options.getFieldMask())
-            : firestore.getDataConverter().parseSetData(data);
+            ? firestore.getDataConverter().parseMergeData(data, options.getFieldMask(), getId())
+            : firestore.getDataConverter().parseSetData(data, getId());
     return firestore
         .getClient()
         .write(parsed.toMutationList(key, Precondition.NONE))
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
@@ -171,8 +171,8 @@ public class DocumentReference {
     checkNotNull(options, "Provided options must not be null.");
     ParsedSetData parsed =
         options.isMerge()
-            ? firestore.getDataConverter().parseMergeData(data, options.getFieldMask(), getId())
-            : firestore.getDataConverter().parseSetData(data, getId());
+            ? firestore.getDataConverter().parseMergeData(data, options.getFieldMask())
+            : firestore.getDataConverter().parseSetData(data);
     return firestore
         .getClient()
         .write(parsed.toMutationList(key, Precondition.NONE))
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
@@ -154,6 +154,13 @@ public class DocumentReference {
     return set(data, SetOptions.OVERWRITE);
   }
 
+  // Handles specific logic for document addition before entering comment `set` logic.
+  /* package */ Task<Void> setForAddition(@NonNull Object data) {
+    checkNotNull(data, "Provided data must not be null.");
+    // TODO: Assert fields annotated with DocumentId in `data` are set to null.
+    return setImpl(data, SetOptions.OVERWRITE);
+  }
+
   /**
    * Writes to the document referred to by this DocumentReference. If the document does not yet
    * exist, it will be created. If you pass {@link SetOptions}, the provided data can be merged into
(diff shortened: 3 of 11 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -97,6 +98,15 @@ public static Object convertToPlainJavaTypes(Object object) {
...
   }
 
+  /**
+   * If there are fields annotated with {@link com.google.firebase.firestore.DocumentId}, make sure
+   * they are set to expected value by throwing runtime exception when it fails to comply.
+   *
+   * @param object Java object that might have DocumentId annotated fields.
+   * @param expected Expected value of the DocumentId annotated fields.
+   */
+  public static void assertAnnotatedDocumentIdEqual(Object object, @Nullable Object expected) {}
```
mikelehen: As I mentioned elsewhere, we will always allow either null or the expected value, so `expected` won't need to be `@Nullable`, but the implementation should always accept null as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
```diff
@@ -0,0 +1,43 @@
...
+ * populated with the document ID being converted.
+ *
+ * <p>When objects with the annotation is used to create new documents, its field value must be null
+ * to guarantee uniqueness, otherwise a runtime exception will be thrown.
+ *
+ * <p>When objects with the annotation is used to update documents, its field value must match the
+ * target documents, otherwise a runtime exception will be thrown.
+ *
+ * <p>This annotation can only be applied to fields of String or {@link DocumentReference},
+ * otherwise a runtime exception will be thrown.
```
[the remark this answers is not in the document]
Hui-Wu: I'll just steal this.

firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
```diff
@@ -125,7 +125,8 @@ public DocumentReference document(@NonNull String documentPath) {
   public Task<DocumentReference> add(@NonNull Object data) {
     checkNotNull(data, "Provided data must not be null.");
     final DocumentReference ref = document();
-    return ref.set(data)
+    // TODO: Assert fields annotated with DocumentId in `data` are set to null.
+    return ref.setForAddition(data)
```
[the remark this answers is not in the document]
Hui-Wu: Good point. I was over-thinking.

firebase-firestore/src/main/java/com/google/firebase/firestore/Transaction.java
```diff
@@ -88,6 +88,7 @@ public Transaction set(
     firestore.validateReference(documentRef);
     checkNotNull(data, "Provided data must not be null.");
     checkNotNull(options, "Provided options must not be null.");
+    // TODO: Assert fields annotated with DocumentId in `data` match `documentRef`.
```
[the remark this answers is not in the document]
Hui-Wu: Following dicussion with mcg@, the verification will be done during serialization within CustomClassMapper.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: No longer needed.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
```diff
@@ -169,6 +176,12 @@ public CollectionReference collection(@NonNull String collectionPath) {
   public Task<Void> set(@NonNull Object data, @NonNull SetOptions options) {
     checkNotNull(data, "Provided data must not be null.");
     checkNotNull(options, "Provided options must not be null.");
+    // TODO: Assert fields annotated with DocumentId in `data` match `documentRef`.
```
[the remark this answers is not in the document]
Hui-Wu: Ack

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -753,7 +770,7 @@ private Type resolveType(Type type, Map<TypeVariable<Class<T>>, Type> types) {
       }
     }
 
-    Map<String, Object> serialize(T object, ErrorPath path) {
+    Map<String, Object> serialize(T object, ErrorPath path, CustomClassMapper classMapper) {
```
wilhuff: Given that the `BeanMapper` is shared between instances of `CustomClassMapper` and that we already need to pass a reference to `CustomClassMapper` around, would in make sense to just leave everything static and instead have a `SerializationContext` that included both the document-id and the `ErrorPath`?

It doesn't seem like that would blow the change up that much but would at least keep the way the document-id is accessed uniform.

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
```diff
@@ -207,7 +223,8 @@ private FieldValue convertAndParseFieldData(Object input, ParseContext context)
    * conform to document data (i.e. it must parse into an ObjectValue model type) and will throw an
    * appropriate error otherwise.
    */
-  private ObjectValue convertAndParseDocumentData(Object input, ParseContext context) {
+  private ObjectValue convertAndParseDocumentData(
+      Object input, ParseContext context, String expectedId) {
```
mikelehen: Per Gil's feedback elsewhere, this probably needs to be a DocumentReference but if not, I'd recommend renaming `expectedId` to `expectedDocumentId` or similar, since `id` is probably too ambiguous by itself.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -66,18 +66,35 @@ private static void hardAssert(boolean assertion, String message) {
     }
   }
 
+  /**
+   * Returns a mapper with a given document ID. The document ID will be used to populate the {@link
```
mikelehen: nit: "with" => "that's associated with"  (just seems clearer to me)

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
```diff
@@ -171,8 +171,8 @@ public CollectionReference collection(@NonNull String collectionPath) {
     checkNotNull(options, "Provided options must not be null.");
     ParsedSetData parsed =
         options.isMerge()
-            ? firestore.getDataConverter().parseMergeData(data, options.getFieldMask())
-            : firestore.getDataConverter().parseSetData(data);
+            ? firestore.getDataConverter().parseMergeData(data, options.getFieldMask(), getId())
```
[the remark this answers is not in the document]
Hui-Wu: No longer needed.

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
```diff
@@ -198,7 +214,7 @@ public FieldValue parseQueryValue(Object input) {
 
   /** Converts a POJO to native types and then parses it into model types. */
   private FieldValue convertAndParseFieldData(Object input, ParseContext context) {
-    Object converted = CustomClassMapper.convertToPlainJavaTypes(input);
+    Object converted = CustomClassMapper.withDocumentId(null).convertToPlainJavaTypes(input);
```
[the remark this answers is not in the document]
Hui-Wu: No longer needed, since we will not validate on write-path

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
Hui-Wu: No longer needed

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: No longer needed.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -66,18 +66,35 @@ private static void hardAssert(boolean assertion, String message) {
...
+   * com.google.firebase.firestore.DocumentId} annotated fields of the POJO of the custom class;
+   * when converting a POJO to plain Java type, the @DocumentId-annotated fields should be either
+   * null or match the given document ID.
+   */
+  // TODO: actually implement documentId population/validation logic during (de)serialization.
+  public static CustomClassMapper withDocumentId(String documentId) {
```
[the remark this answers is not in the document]
Hui-Wu: no longer needed.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Made it back to all static with DeserializeContext, since we don't need the doc id for Serialization any more.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * <p>When writing a POJO to Firestore, the @DocumentId-annotated field must either be null or match
+ * the document ID of the document being written to, else a runtime exception will be thrown.
+ */
+@PublicApi
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface DocumentId {}
```
wilhuff: We can't merge this PR as-is with this class being public here. We should make it package private, either here or in `util`, depending on how you want to test it.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -899,7 +899,7 @@ public void setValue(String value) {
 
   private static <T> T deserialize(String jsonString, Class<T> clazz) {
     Map<String, Object> json = fromSingleQuotedString(jsonString);
-    return CustomClassMapper.convertToCustomClass(json, clazz);
+    return CustomClassMapper.convertToCustomClass(json, clazz, null);
```
wilhuff: Most tests here aren't going to pass a non-null value here, so consider adding a local method that defaults the parameter to null to reduce the boilerplate of each invocation.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -91,9 +91,12 @@ public static Object convertToPlainJavaTypes(Object object) {
    *
    * @param object The representation of the JSON data
    * @param clazz The class of the object to convert to
+   * @param documentId The value to set to {@link com.google.firebase.firestore.DocumentId}
+   *     annotated fields in the custom class.
    * @return The POJO object.
    */
-  public static <T> T convertToCustomClass(Object object, Class<T> clazz) {
+  public static <T> T convertToCustomClass(Object object, Class<T> clazz, String documentId) {
```
wilhuff: This parameter should be a `DocumentReference` or at least a `DocumentKey`. Just the string allows for

```java
@DocumentId
public String justTheLastPathComponent;
```
But we also want to support
```java
@DocumentId
public DocumentReference fullPath;
```

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.java
Hui-Wu: Moved to util

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Ah, should have fixed it. Done now.

mikelehen: FYI- It doesn't look like you've pushed the fix yet, but you also haven't assigned it back to me / wilhuff, so perhaps you know that. :-)  Anyway, please assign it back when it's ready for another look!

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/DocumentId.java
```diff
@@ -0,0 +1,38 @@
...
+ * <p>This annotation can only be applied to fields of String or {@link DocumentReference},
+ * otherwise a runtime exception will be thrown.
+ *
+ * <p>When writing a POJO to Firestore, the @DocumentId-annotated field will be ignored, to allow
+ * writing to any documents that are not the origin of the POJO.
+ */
+@PublicApi
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface DocumentId {}
```
wilhuff: Remove `public`.

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
```diff
@@ -73,14 +73,23 @@ public UserDataConverter(DatabaseId databaseId) {
     this.databaseId = databaseId;
   }
 
-  /** Parse document data from a non-merge set() call. */
+  /**
+   * Parse document data from a non-merge set() call.
+   *
+   * @param input A POJO object representing document data.
```
mikelehen: nit: We usually use "POJO object" to refer to custom developer-written classes representing Firestore data, but "input" could also just be a normal Map<String, Object>... which I wouldn't consider a POJO.  So I'd change this to "A Map or POJO object representing document data."

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
```diff
@@ -73,14 +73,23 @@ public UserDataConverter(DatabaseId databaseId) {
...
     ParseAccumulator accumulator = new ParseAccumulator(UserData.Source.Set);
     ObjectValue updateData = convertAndParseDocumentData(input, accumulator.rootContext());
     return accumulator.toSetData(updateData);
   }
 
-  /** Parse document data from a set() call with SetOptions.merge() set. */
+  /**
+   * Parse document data from a set() call with SetOptions.merge() set.
+   *
+   * @param input A POJO object representing document data.
```
mikelehen: nit: "A map or POJO object representing document data"

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -754,6 +757,7 @@ private Type resolveType(Type type, Map<TypeVariable<Class<T>>, Type> types) {
     }
 
     Map<String, Object> serialize(T object, ErrorPath path) {
+      // TODO: Add logic to skip @DocumentId annotated fields in serialization.
```
mikelehen: Please don't check in naked "TODO" comments.  If this is intended to be checked in, please reference a bug ("TODO(b/xyz)") or your username for context.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -1011,4 +1015,18 @@ public String toString() {
       }
     }
   }
+
+  static class DeserializeContext {
+    final ErrorPath errorPath;
```
mikelehen: I have a guess what this is for, but I'm not 100% sure so it probably needs a comment.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -1011,4 +1015,18 @@ public String toString() {
       }
     }
   }
+
+  static class DeserializeContext {
```
mikelehen: It looks like this isn't used yet.  If you're planning to check this PR in in its current state, I think I'd suggest removing this for now and adding it in the PR that uses it?  That way:
1) we don't have dead code sitting around
2) you don't end up revamping this in the next PR once you actually use it (e.g. "documentId" should probably be a DocumentReference or DocumentKey now?)

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
mikelehen: Did you intentionally not use the new method here? 😄

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
Hui-Wu: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Deleted.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Deleted. Will add comments when it's back.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/DocumentId.java
Hui-Wu: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
Hui-Wu: Yes..because `serialize` below already refers to mapper..

PR #458, opened 2019-05-23
Add List() support

schmidt-sebastian:
This PR adds List support to the Android Storage SDK.

API: go/list-v2
Web PR: https://github.com/firebase/firebase-js-sdk/pull/1610


firebase-storage/src/main/java/com/google/firebase/storage/ListResult.java
```diff
@@ -0,0 +1,118 @@
...
+import com.google.firebase.annotations.PublicApi;
+import java.util.ArrayList;
+import java.util.List;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/** A single paged result of a {@link StorageReference#list} call. */
+@PublicApi
+public class ListResult {
```
fredzqm:
```suggestion
public final class ListResult {
```

firebase-storage/src/main/java/com/google/firebase/storage/ListResult.java
```diff
@@ -0,0 +1,118 @@
...
+  private static final String ITEMS_KEY = "items";
+  private static final String NAME_KEY = "name";
+  private static final String PAGE_TOKEN_KEY = "nextPageToken";
+  private static final String PREFIXES_KEY = "prefixes";
+
+  private List<StorageReference> prefixes;
+  private List<StorageReference> items;
+  @Nullable private String pageToken;
+
+  ListResult(
```
fredzqm:
```suggestion
  private ListResult(
```

firebase-storage/src/main/java/com/google/firebase/storage/ListResult.java
```diff
@@ -0,0 +1,118 @@
...
+  public String getPageToken() {
+    return pageToken;
+  }
+
+  static class Builder {
+    private final JSONObject resultBody;
+    private final StorageReference storageRef;
+    private final FirebaseStorage firebaseStorage;
+
+    public Builder(JSONObject resultBody, StorageReference storageRef) {
```
fredzqm: I would suggest pass-in `StorageReference` in the constructor, and have `appendResponse(JSONObject )` to parse and append results. This way, `listAll` can take advantage of the `Builder`.
The fact that `build()` might blow up is quite surprising.

nit: Google3 Java usually add a static `toBuilder(StorageReference storageRef)`. No strong opinion here.

firebase-storage/src/main/java/com/google/firebase/storage/network/ListNetworkRequest.java
```diff
@@ -0,0 +1,76 @@
...
+  protected String getQueryParameters() throws UnsupportedEncodingException {
+    List<String> keys = new ArrayList<>();
+    List<String> values = new ArrayList<>();
+
+    String prefix = getPathWithoutBucket();
+    if (!TextUtils.isEmpty(prefix)) {
+      keys.add("prefix");
+      values.add(prefix + "/");
+    }
+    keys.add("delimiter");
```
fredzqm: Maybe add a comment here as this will be obscure to someone unfamiliar with GCS's List API.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,109 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+    Executor executor = StorageTaskScheduler.getInstance().getCommandPoolExecutor();
+    Task<ListResult> list = list(ListTask.DEFAULT_PAGE_SIZE);
+
+    Continuation<ListResult, Task<Void>> continuation =
+        new Continuation<ListResult, Task<Void>>() {
+          @Override
+          public Task<Void> then(@NonNull Task<ListResult> currentPage) {
+            ListResult result = currentPage.getResult();
+            prefixes.addAll(result.getPrefixes());
+            items.addAll(result.getItems());
```
fredzqm: This is O(n2) for multiple pages.
`ListResult.Builder#appendListResult()` would be nice here.

schmidt-sebastian: If we want to make this O(n) then we have to initialize the resulting array with the correct capacity upfront. I updated the code to do this, but am a little worried about the readability.

firebase-storage/src/main/java/com/google/firebase/storage/network/ListNetworkRequest.java
schmidt-sebastian: Done

firebase-storage/src/test/java/com/google/firebase/storage/ListTest.java
```diff
@@ -0,0 +1,144 @@
...
+      fail();
+    } catch (IllegalArgumentException e) {
+
+    }
+
+    try {
+      reference.list(1001);
+      fail();
+    } catch (IllegalArgumentException e) {
+
```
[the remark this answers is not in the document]
schmidt-sebastian: Done. I hope google-java-format doesn't add them back :)

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,119 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+  @PublicApi
+  public Task<ListResult> listAll() {
+    TaskCompletionSource<ListResult> pendingResult = new TaskCompletionSource<>();
+
+    List<ListResult> results = new ArrayList<>();
+    int[] prefixesCount = new int[] {0};
+    int[] itemsCount = new int[] {0};
+
+    Executor executor = StorageTaskScheduler.getInstance().getCommandPoolExecutor();
+    Task<ListResult> list = list(ListTask.DEFAULT_PAGE_SIZE);
```
fredzqm: One caveat that I should explain here.
1. When you pass in `maxResults` to Firebase Storage Backend, we run a loop to make sure either max results are reached or there is no more results.
2. When you do not pass in `maxResults`, we always call GCS once and return the data.

Both of them would filter invalid objects. If there is one invalid object (very common case if you create a directory in console), 
1. In the first case, we call GCS twice to fill the spot for that invalid object.
2. In the second case, we return `999=1000-1` elements.

So in practice, not passing in a `maxResults` is a lot more effecient for listAll.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,119 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+   *     than 0 and at most 1000.
+   * @param pageToken A page token from a previous call to list.
+   * @return A a {@link Task} that returns the next items and prefixes under the current
+   *     StorageReference.
+   */
+  @NonNull
+  @PublicApi
+  public Task<ListResult> list(int maxResults, @NonNull String pageToken) {
+    Preconditions.checkArgument(maxResults > 0, "maxResults must be greater than zero");
+    Preconditions.checkArgument(maxResults <= 1000, "maxResults must be less than 1000");
```
fredzqm:
```suggestion
    Preconditions.checkArgument(maxResults <= 1000, "maxResults must be less than or equal to 1000");
```

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,119 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
     TaskCompletionSource<Void> pendingResult = new TaskCompletionSource<>();
     StorageTaskScheduler.getInstance().scheduleCommand(new DeleteStorageTask(this, pendingResult));
     return pendingResult.getTask();
   }
+
+  /**
+   * List up to `maxResults` items (files) and prefixes (folders) under this StorageReference.
+   *
+   * <p>"/" is treated as a path delimiter. Firebase Storage does not support invalid object paths
+   * that end with "/" or contain two consecutive "//". All invalid objects in GCS will be filtered.
```
fredzqm:
```suggestion
   * that end with "/" or contain two consecutive "/"s. All invalid objects in GCS will be filtered.
```
Maybe mention if there are too many `unsupported` objects in GCS, list may always fail.

Consider mirror the doc from JS SDK PR to stay consistent.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
schmidt-sebastian: Used `maxResults must be at most 1000` here and above.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
schmidt-sebastian: Thanks for the explanation. Guess it makes sense for the SDK and the backend team to collaborate :) I made maxResults optional.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
schmidt-sebastian: I do slightly prefer having smaller docstrings. Ideally, the backend would throw an error that is explanatory enough if this edge case occurs.

schmidt-sebastian:
Adding @rsgowman for readability.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
fredzqm: I don't think correct capacity initialization is needed for amortized O(n).
Two growing lists should serve the purpose.

schmidt-sebastian: There are two approaches to merging a lot of lists:

The original proposal with uses `addAll()` repeatedly. For ArrayList (LinkedList are better here, but I don't want to return a LinkedList-backed List to the consumer) the list [grows 2n every time](https://cs.corp.google.com/piper///depot/google3/third_party/java_src/jdk/openjdk8/u60/trunk/jdk/src/share/classes/java/util/ArrayList.java?rcl=249564389&g=0&l=257). If we are adding 9 pages of 1000 items:

- We start out with a list of 0 elements.
- Page 1: We increase the size to 1000 in one step. No existing elements are copied. 1000 new elements are copied.
- Page 2: We increase the size to 2000. The first 1000 elements are copied again.  1000 new elements are copied.
- Page 3: We increase the size to 4000. The first 2000 elements are copied again.  1000 new elements are copied.
- Page 4: 1000 new elements are copied.
- Page 5: We increase the size to 8000.  The first 4000 elements are copied again.  1000 new elements are copied.
- Page 6: 1000 new elements are copied.
- Page 7: 1000 new elements are copied.
- Page 8: 1000 new elements are copied.
- Page 9: We increase the size to 16000.  The first 8000 elements are copied again.  1000 new elements are copied.

Your suggestion of doing this inside the Builder seems equivalent to me (I might not be following correctly).

If we are setting the initial capacity, we reduce the number of copies to 9 * 1000 (down from 24 * 1000) and do not allocate space for 7000 unused elements. The amortized cost of both operations is O(n) (if I remember my interview training correctly), but O(n) might not be the best measurement for small datasets on Mobile devices that are much more resource constrained.

fredzqm: The trade-off is (N is # of total result)
* List of List and merge in the end.
 * N copies + N creations
 * Max array N memory
* Two accumulating list
 * 2N~3N copies + N creations.
 * Max array 2N~4N memory
 * Better Readability

There is a constant performance improvement, but I would not sacrifice readability for it.
We could let the third-pair of eyes decide here.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,122 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+            itemsCount[0] += result.getItems().size();
+            results.add(result);
+
+            if (result.getPageToken() != null) {
+              Task<ListResult> nextPage = listHelper(/* maxResults= */ null, result.getPageToken());
+              nextPage.continueWithTask(executor, this);
+            } else {
+              List<StorageReference> prefixes = new ArrayList<>(prefixesCount[0]);
+              List<StorageReference> items = new ArrayList<>(itemsCount[0]);
+              for (ListResult previousResult : results) {
```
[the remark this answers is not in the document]
rsgowman: Option 6: Eliminate listAll() entirely. i.e. force the user to think about this situation. For prototyping, the user can just fetch the first 1000 and ignore the rest.

firebase-storage/src/main/java/com/google/firebase/storage/ListResult.java
```diff
@@ -0,0 +1,106 @@
...
+
+/** Contains the prefixes and items returned by a {@link StorageReference#list} call. */
+@PublicApi
+public final class ListResult {
+  private static final String ITEMS_KEY = "items";
+  private static final String NAME_KEY = "name";
+  private static final String PAGE_TOKEN_KEY = "nextPageToken";
+  private static final String PREFIXES_KEY = "prefixes";
+
+  private List<StorageReference> prefixes;
```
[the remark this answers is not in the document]
fredzqm: Items are ordered and unique.

Order consistency is quite important for displaying them in a UI.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,122 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
   @PublicApi
   public Task<Void> delete() {
     TaskCompletionSource<Void> pendingResult = new TaskCompletionSource<>();
     StorageTaskScheduler.getInstance().scheduleCommand(new DeleteStorageTask(this, pendingResult));
     return pendingResult.getTask();
   }
+
+  // region List
+
+  /**
```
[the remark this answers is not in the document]
fredzqm: Good idea.
I should add this to the JS SDK as well.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,122 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+    Preconditions.checkArgument(maxResults <= 1000, "maxResults must be at most 1000");
+    return listHelper(maxResults, /* pageToken */ null);
+  }
+
+  /**
+   * Resumes a previous call to {@link #list(int)}, starting after a pagination token. Returns the
+   * next set of items (files) and prefixes (folders) under this StorageReference.
+   *
+   * <p>"/" is treated as a path delimiter. Firebase Storage does not support invalid object paths
+   * that end with "/" or contain two consecutive "//". All invalid objects in GCS will be filtered.
```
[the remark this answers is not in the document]
fredzqm: To minimize the confusion, we decided on
"a second query run transparently to get the additional missing entry"

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,109 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+   *
+   * @param maxResults The maximum number of results to return in a single page. Must be greater
+   *     than 0 and at most 1000.
```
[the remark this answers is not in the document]
fredzqm: This is part of GCS's API limit.
https://cloud.google.com/storage/docs/json_api/v1/objects/list

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,122 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+   * @param pageToken A page token from a previous call to list.
```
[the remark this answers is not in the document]
fredzqm: "The pageToken is an encoded field that marks the name and generation of the last object in the returned list"

See [GCS doc](https://cloud.google.com/storage/docs/json_api/v1/objects/list)

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,122 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
+    Preconditions.checkArgument(
+        pageToken != null, "pageToken must be non-null to resume a previous list() operation");
+    return listHelper(maxResults, pageToken);
+  }
+
+  /**
+   * List all items (files) and prefixes (folders) under this StorageReference.
+   *
+   * <p>This is a helper method for calling list() repeatedly until there are no more results. The
+   * default pagination size is 1000. Consistency of the result is not guaranteed if objects are
```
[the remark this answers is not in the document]
fredzqm: `1000` is from GCS's API limit.

In the case of `listAll`, larger page size is more efficient (less RTT).

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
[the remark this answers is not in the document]
fredzqm: Order is important for List API. Users often display all items in the UI, consistent ordering matters here.
In practice, random access may not be that useful, but it will be confusing if the developers run into this case. I would rather introduce less caveats.
If we really care about efficiency here, `ListAdaptor` would be the way to go. Not convinced if the complexity is worthwhile though. It also increases JAR size.

In the order of preference, 2 > 1 > 4.

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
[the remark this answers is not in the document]
fredzqm: The API will be inconsistent compared to GCS API, but will be self-consistent given no object insertion or deletion.

firebase-storage/src/main/java/com/google/firebase/storage/ListTask.java
```diff
@@ -0,0 +1,83 @@
...
+  @Nullable private Integer maxResults;
+  private ListResult listResult;
+
+  ListTask(
+      @NonNull StorageReference storageRef,
+      @Nullable Integer maxResults,
+      @Nullable String pageToken,
+      @NonNull TaskCompletionSource<ListResult> pendingResult) {
+    Preconditions.checkNotNull(storageRef);
+    Preconditions.checkNotNull(storageRef);
```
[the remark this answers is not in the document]
rsgowman: Ouch. :)

firebase-storage/src/main/java/com/google/firebase/storage/StorageReference.java
```diff
@@ -563,14 +566,124 @@ public StreamDownloadTask getStream(@NonNull StreamDownloadTask.StreamProcessor
...
     return pendingResult.getTask();
   }
+
+  // region List
+
+  /**
+   * List up to `maxResults` items (files) and prefixes (folders) under this StorageReference.
+   *
+   * <p>"/" is treated as a path delimiter. Firebase Storage does not support invalid object paths
+   * that end with "/" or contain two consecutive "/"s. All invalid objects in GCS will be filtered.
```
fredzqm: One more thing.
Can we refer to them as `unsupported objects` to stay consistent?

schmidt-sebastian: Done

PR #461, opened 2019-05-24
Properly handle pom dependencies.

vkryachko:
Namely:
* Exclude multidex from POMs
* Set dependency scope to compile
* Properly set dependency type(jar, aar) based on resolved artifact name

### DIFF
--- a/buildSrc/src/test/groovy/com/google/firebase/gradle/plugins/publish/PublishingPluginSpec.groovy
+++ b/buildSrc/src/test/groovy/com/google/firebase/gradle/plugins/publish/PublishingPluginSpec.groovy
@@ -44,6 +44,7 @@ class PublishingPluginSpec extends Specification {
 	    }
         dependencies {
             <%dependencies.each { println "implementation project(':$it.name')" } %>
+            <%customDependencies.each { println "implementation '$it'" } %>
         }
         '''
         String name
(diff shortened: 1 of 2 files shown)


buildSrc/src/test/groovy/com/google/firebase/gradle/plugins/publish/PublishingPluginSpec.groovy
```diff
@@ -44,13 +44,15 @@ class PublishingPluginSpec extends Specification {
...
             <%dependencies.each { println "implementation project(':$it.name')" } %>
+            <%customDependencies.each { println "implementation '$it'" } %>
         }
         '''
         String name
         String group = 'com.example'
         String version = 'undefined'
         String latestReleasedVersion = ''
         Set<Project> projectDependencies = []
+        Set<String> customDependencies = []
```
rlazo: I think "customDependencies" is a bit confusing. In my mind there are "implementation project(...)" dependencies, and "implementation (...)" dependencies, so maybe "nonProjectDependencies", "external dependencies"?

vkryachko: Good point, changed to `externalDependencies`

rlazo: Much better, thanks!

PR #462, opened 2019-05-24
Move SSL initialization to a separate thread

schmidt-sebastian:
This PR moves the SSL Provider installation (which attempts to load Play Services) to a separate thread. Operations that do not require network should now be faster when run immediately after client startup.

First query performance number for reading a single document before this change:
724ms 43 ms 604ms 684ms 308ms (avg 550 ms)

With this change:
495ms 637ms 297ms 433ms 628ms (avg 498 ms)

This is on a Nexus 5X.  I verified that the client is able to connect to the backend on API Level 18 with GMS Core Play Services installed.

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
-- [changed] SSL initialization now happens on a separate thread, which should
-  improve query performance during client startup.
+- [changed] SSL initialization now happens on a separate thread, which reduces
+  the time taken to produce the first query result.
 
 # Unreleased (19.0.1)
 - [fixed] Fixed an issue that prevented schema migrations for clients with
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Unreleased
+- [changed] SSL initialization now happens on a separate thread, which should
+  improve query performance during client startup.
+
+# Unreleased (19.0.1)
 - [fixed] Fixed an issue that prevented schema migrations for clients with
   large offline datasets (#370).
 
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
@@ -16,12 +16,8 @@ package com.google.firebase.firestore.remote;
 
 import static com.google.firebase.firestore.util.Assert.hardAssert;
 
-import android.content.Context;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
-import com.google.android.gms.common.GooglePlayServicesNotAvailableException;
-import com.google.android.gms.common.GooglePlayServicesRepairableException;
-import com.google.android.gms.security.ProviderInstaller;
 import com.google.firebase.database.collection.ImmutableSortedSet;
 import com.google.firebase.firestore.core.OnlineState;
 import com.google.firebase.firestore.core.Transaction;
@@ -129,7 +125,7 @@ public final class RemoteStore implements WatchChangeAggregator.TargetMetadataPr
   private final OnlineStateTracker onlineStateTracker;
 
   private boolean networkEnabled = false;
-  private volatile boolean sslReady = false;
+  private boolean sslReady = false;
   private final WatchStream watchStream;
   private final WriteStream writeStream;
   @Nullable private WatchChangeAggregator watchChangeAggregator;
@@ -152,10 +148,10 @@ public final class RemoteStore implements WatchChangeAggregator.TargetMetadataPr
 
   public RemoteStore(
       RemoteStoreCallback remoteStoreCallback,
-      Context context,
       LocalStore localStore,
       Datastore datastore,
       AsyncQueue workerQueue,
+      SslProvider sslProvider,
       ConnectivityMonitor connectivityMonitor) {
     this.remoteStoreCallback = remoteStoreCallback;
     this.localStore = localStore;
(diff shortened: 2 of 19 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -162,6 +169,23 @@ public RemoteStore(
     onlineStateTracker =
         new OnlineStateTracker(workerQueue, remoteStoreCallback::handleOnlineStateChange);
 
+    new Thread() {
+      public void run() {
+        try {
+          ProviderInstaller.installIfNeeded(context);
+        } catch (GooglePlayServicesNotAvailableException /* Thrown by ProviderInstaller */
+            | GooglePlayServicesRepairableException /* Thrown by ProviderInstaller */
+            | IllegalStateException /* Thrown by Robolectric */ e) {
```
schmidt-sebastian: Note: Apparently there is a way to avoid this IllegalStateException by setting "com.google.android.gms.version" in the AndroidManifest that Robolectric uses, but this seems to be fully automated in our repo (and should be as of Robolectric 4) and I can't figure out how to do this.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -406,8 +430,9 @@ private void startWatchStream() {
         shouldStartWatchStream(),
         "startWatchStream() called when shouldStartWatchStream() is false.");
     watchChangeAggregator = new WatchChangeAggregator(this);
-    watchStream.start();
 
+    initializeSslIfNeeded();
```
wilhuff: I'm not a fan of blocking here. Firestore is an event loop and we should reformulate this in terms of an event we need to handle. This avoids exposes threading constructs into code that currently is happily single threaded. Specifically, once SSL initialization is complete, we post that completion into the async queue.

The question then becomes, how to avoid touching the network before this is ready? I think is easy because the RemoteStore already understands how to run with the network is disabled. `canUseNetwork()` already gates every action that can touch the network. Our goal should be to piggyback off of that.

In particular, I think you should add a boolean to RemoteStore called `networkInitialized` or something like that. Default it to false, and the SslProvider completion should flip it to true in something like `handleNetworkInitialized()`.

Then factor out everything but `networkEnabled = true;` from `enableNetwork` into a helper (`maybeStartNetwork`?) and call it from `handleNetworkInitialized` as well.

Change `canUseNetwork` to check `networkInitialized && networkEnabled`.

Note that the simpler approach of just delaying `RemoteStore.start` until the network is initialized doesn't quite work, because the first action the user takes could be to disable the network. We can't have the completion of SSL initialization overriding that.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -613,9 +638,26 @@ private void startWriteStream() {
...
+
+    initializeSslIfNeeded();
     writeStream.start();
   }
 
+  /** Waits for the SSL Provider to be installed. Logs a warning on failure. */
+  private void initializeSslIfNeeded() {
+    try {
+      sslBarrier.await();
+    } catch (InterruptedException e) {
```
wilhuff: Who interrupts the thread to generate this exception?

Normally interrupts mean that the system is trying to kill a thread. We shouldn't swallow this interrupt as we're doing here.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Unreleased
+- [changed] SSL initialization now happens on a separate thread, which should
```
wilhuff: "should improve" makes it sound like we didn't test it. How about "which reduces the time taken to produce the first query results." ?

wilhuff:
One thing that's not present in this PR was something that checks whether or not a sane set of required ciphers was present. If there are no ciphers, shouldn't we log a wordy error message about being unable to connect on this device?

firebase-firestore/CHANGELOG.md
schmidt-sebastian: That was not totally unintended. There is some overhead that I assume is much lower than blocking on GMSCore, but I only tested this change on the one device I have. With that being, I don't mind being more assertive and updated the message.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -125,6 +130,7 @@
   private final OnlineStateTracker onlineStateTracker;
 
   private boolean networkEnabled = false;
+  private final CountDownLatch sslBarrier = new CountDownLatch(1);
```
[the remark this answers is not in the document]
schmidt-sebastian: I wanted to push back on this, but it turns out that the "lazy initialization" you suggested doesn't quite work in the Spec tests. For that, I would somehow have to block on it. By using a separate class, I am able to use a different SSL Provider for the tests.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
schmidt-sebastian: I tried to figure out how to prevent blocking the thread, but my solution would have required some large changes in other parts of the SDK. The idea was basically that the client would no longer rely on the RemoteStore's ability to always be in a state that it can start the network from. 
The missing piece was that this doesn't have to be the case, since the RemoteStore can already catch up to the state it is supposed to be in (via `fillWritePipeline()`). Long story short - what you are suggesting here is a huge improvement and I wish I had come up with that myself. Thanks :)

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -613,9 +638,26 @@ private void startWriteStream() {
     hardAssert(
         shouldStartWriteStream(),
         "startWriteStream() called when shouldStartWriteStream() is false.");
+
+    initializeSslIfNeeded();
     writeStream.start();
   }
 
+  /** Waits for the SSL Provider to be installed. Logs a warning on failure. */
+  private void initializeSslIfNeeded() {
```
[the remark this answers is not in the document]
schmidt-sebastian: I removed this method, but the saving grace that I did consider naming this method `awaitSslInitialization()` at some point :)

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
schmidt-sebastian: My inclination here is deal with the error handling as part of b/132978871. My goal is to make a best effort at providing an updated SSL Cipher, but if I can't then this is only an error condition if the SDK fails to connect. That error handling will come in a follow-up and will be part of the stream handling rather than Remote Store.

schmidt-sebastian:
> One thing that's not present in this PR was something that checks whether or not a sane set of required ciphers was present. If there are no ciphers, shouldn't we log a wordy error message about being unable to connect on this device?

That's tracked in a separate issue (b/132978871) and will be a follow-up PR.

schmidt-sebastian:
@wilhuff Looks like I broke this SQLiteSpecTests. Still investigating.

schmidt-sebastian:
The approach in this PR can potentially break Transactions, since they don't go through the RemoteStore.

PR #464, opened 2019-05-27
Implement @DocumentId annotation

Hui-Wu:
Implements @DocumentId support for deserialization of firestore document to custom POJO.


firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -612,47 +648,49 @@ private static RuntimeException deserializeError(ErrorPath path, String reason)
...
-                            + method.getName()
-                            + " (conflicts with "
-                            + existingSetter.getName()
-                            + " defined on "
-                            + existingSetter.getDeclaringClass().getName()
-                            + ")");
-                  }
-                }
-              }
+          if (!shouldIncludeSetter(method)) {
```
Hui-Wu: To reviewer: no behavior change here (supposedly), only refactor to reduce number of nested if/else.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -740,6 +798,47 @@ T deserialize(
           }
         }
       }
+
+      // Populate @DocumentId annotated fields. If there is a conflict (@DocumentId annotation is
+      // applied to a property that is already deserialized from the firestore document)
+      // a runtime exception will be thrown.
```
mikelehen: Good comment, thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -740,6 +798,47 @@ T deserialize(
...
+      for (String docIdPropertyName : documentIdPropertyNames) {
```
mikelehen: FWIW- This class has overly large methods, but typically we try to keep methods shorter.  Consider separating this out into a populateDocumentIdFields() method or something.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -809,6 +913,19 @@ private void applyFieldAnnotations(Field field) {
...
+
+      if (field.isAnnotationPresent(DocumentId.class)) {
+        Class<?> fieldType = field.getType();
+        if (fieldType != String.class && fieldType != DocumentReference.class) {
+          throw new IllegalArgumentException(
+              "Field "
+                  + field.getName()
+                  + " is annotated with @DocumentId but is "
+                  + fieldType
+                  + " instead of String or DocumentReference.");
```
mikelehen: Excellent error message. 😄

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -834,6 +965,19 @@ private void applySetterAnnotations(Method method) {
...
+      if (method.isAnnotationPresent(DocumentId.class)) {
+        Class<?> paramType = method.getParameterTypes()[0];
+        if (paramType != String.class && paramType != DocumentReference.class) {
+          throw new IllegalArgumentException(
+              "Method "
+                  + method.getName()
+                  + " is annotated with @DocumentId but sets "
+                  + paramType
+                  + " instead of String or DocumentReference.");
+        }
```
mikelehen: Consider adding a helper to avoid replicating these checks / error message.

```
  private void validateDocumentIdType(String fieldDescription, String operation, Type type) {
      if (type != String.class && type != DocumentReference.class) {
         throw new IllegalArgumentException(
            fieldDescription
                + " is annotated with @DocumentId but "
                + operation
                + type
                + " instead of String or DocumentReference.");
        }
  }
```

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -899,8 +907,12 @@ public void setValue(String value) {
   }
 
   private static <T> T deserialize(String jsonString, Class<T> clazz) {
+    return deserialize(jsonString, clazz, null);
```
mikelehen: optional nit: Whenever passing literals like null, true, false, etc. I like to comment them so you can tell what they represent without having to look up the method signature.  i.e.:

```
return deserialize(jsonString, clazz, /*docRef=*/null);
```

In this case, since the methods are next to each other, consider this very optional.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -1498,6 +1510,21 @@ public void getterOverridesField() {
...
+  }
+
+  @Test
+  public void deserializeGetterBeanWithNoBackingFieldThrows() {
+    assertExceptionContains(
+        "No setter/field",
+        () -> {
+          deserialize("{'value': 'foo'}", GetterBeanNoField.class);
+        });
+  }
```
mikelehen: Thanks for adding these tests!

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -2247,4 +2274,205 @@ public void deserializationFailureIncludesPath() {
...
+    assertExceptionContains(
+        "instead of String or DocumentReference",
+        () -> deserialize("{'intField': 1}", GetterWithDocumentIdOnWrongTypeBean.class));
+
+    assertExceptionContains(
+        "instead of String or DocumentReference",
+        () -> serialize(new PropertyWithDocumentIdOnWrongTypeBean()));
+    assertExceptionContains(
+        "instead of String or DocumentReference",
+        () -> deserialize("{'intField': 1}", PropertyWithDocumentIdOnWrongTypeBean.class));
```
mikelehen: nit: Test boilerplate (repetition) is one of my big pet peeves.  It makes it harder to change implementation details because there are lots of tests to update for a single implementation change.  We end up with lots of repetition anyway, but I think it's always worth asking if/how it can be avoided.  So consider:
1) Putting the expected error text into a variable that's re-used, or even less repetition:
2) Create a list of lambdas containing all of these operations and just loop through them, rather than call assertExceptionContains() a bunch of times.

Note that 2 could make the test a bit harder to debug if it fails, so it's a trade-off.  You could also create a helper method (or local lambda function) that wraps the `assertExceptionContains("instead of String or DocumentReference", ...)`  call and just call it repeatedly.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
```diff
@@ -2247,4 +2274,205 @@ public void deserializationFailureIncludesPath() {
...
+    assertEquals(
+        "doc123",
+        deserialize("{'nestedDocIdHolder': {}}", DocumentIdOnNestedObjects.class, ref)
+            .nestedDocIdHolder
+            .docId);
+  }
+
+  @Test
+  public void documentIdsRoundTrip() {
+    // Implicitly verifies @DocumentId is ignore during serialization.
```
mikelehen: nit: ignore => ignored

mikelehen:
Oh, also, I'm very sorry for the super slow review. With the offsite and everything, this took a lot longer than I would have liked. :)

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -542,21 +562,36 @@ private static RuntimeException deserializeError(ErrorPath path, String reason)
...
     private final Map<String, Method> setters;
     private final Map<String, Field> fields;
 
-    // A list of any properties that were annotated with @ServerTimestamp.
+    // A set of property names that were annotated with @ServerTimestamp.
     private final HashSet<String> serverTimestamps;
 
+    // A set of property names that were annotated with @DocumentId. These properties will be
+    // populated with
+    // document ID values during deserialization, and be skipped during serialization.
```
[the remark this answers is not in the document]
Hui-Wu: Fixed.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -612,47 +648,49 @@ private static RuntimeException deserializeError(ErrorPath path, String reason)
...
-                            + ")");
-                  }
-                }
-              }
+          if (!shouldIncludeSetter(method)) {
+            continue;
+          }
+          String propertyName = propertyName(method);
+          String existingPropertyName = properties.get(propertyName.toLowerCase(Locale.US));
+          if (existingPropertyName != null && !existingPropertyName.equals(propertyName)) {
```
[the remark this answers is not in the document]
Hui-Wu: This is my fault. Now i recall why I was messing around with this block and realized without a public getter it won't work, and went ahead and fix that and did the refactoring (even added test cases for that!). I did not realize the public doc said that.

I am reverting this and the added test cases out of this PR. Thanks for catching this!

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -676,6 +714,19 @@ private static RuntimeException deserializeError(ErrorPath path, String reason)
       if (properties.isEmpty()) {
         throw new RuntimeException("No properties to serialize found on class " + clazz.getName());
       }
+
+      // Make sure we can write to @DocumentId annotated properties before proceeding.
+      for (String docIdProperty : documentIdPropertyNames) {
+        if (setters.containsKey(docIdProperty) || fields.containsKey(docIdProperty)) {
+          continue;
```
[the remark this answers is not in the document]
Hui-Wu: Totally agree, fixed.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -676,6 +714,19 @@ private static RuntimeException deserializeError(ErrorPath path, String reason)
...
+        }
+
+        throw new RuntimeException(
+            "@DocumentId is annotated on property "
+                + docIdProperty
+                + " which provides no way to write to on class "
```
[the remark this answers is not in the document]
Hui-Wu: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
Hui-Wu: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
Hui-Wu: Added.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
Hui-Wu: No problem!

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
Hui-Wu: Done

firebase-firestore/src/test/java/com/google/firebase/firestore/util/MapperTest.java
Hui-Wu: Implemented #1.

I think `assertExceptionContains` here is pretty short and straightforward, adding an extra wrapping leads to more cognition load without a lot benefits. And it's harder to debug too, as you pointed out.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -740,9 +795,58 @@ T deserialize(
...
+        Map<TypeVariable<Class<T>>, Type> types,
+        DeserializeContext context,
+        T instance,
+        HashSet<String> deserialzedProperties) {
+      for (String docIdPropertyName : documentIdPropertyNames) {
+        if (deserialzedProperties.contains(docIdPropertyName)) {
+          String message =
+              "'"
+                  + docIdPropertyName
+                  + "' is found from document "
```
mikelehen: nit: I'd change to: `"' was found on document at "`

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -834,6 +956,24 @@ private void applySetterAnnotations(Method method) {
                 + " is annotated with @ServerTimestamp but should not be. @ServerTimestamp can"
                 + " only be applied to fields and getters, not setters.");
       }
+
+      if (method.isAnnotationPresent(DocumentId.class)) {
+        Class<?> paramType = method.getParameterTypes()[0];
+        ensureValidDocumentIdType("Method", "sets", paramType);
```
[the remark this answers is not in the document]
Hui-Wu: Done.

PR #465, opened 2019-05-27
Upgrade grpc from 1.16.1 to 1.21.0 for firestore

rsgowman:


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -72,6 +72,20 @@ interface RemoteLicenseFetcher extends Serializable {
     }
   }
 
+  static final class YetAnotherApache2LicenseFetcher implements RemoteLicenseFetcher {
```
rsgowman: @vkryachko : Rather than adding more and more `RemoteLicenseFetcher`s (there's now 3 for apache2, and at least 2 for mit) perhaps getServiceUri() and get() could return sets. This would allow multiple uri's for each license. (Or, you could take it a step further and eliminate all `RemoteLicenseFetcher`s and just have a single set of uris. You'd loose the uri -> license mapping, but I'm not sure if that's actually used anywhere or not.)

(Obviously not something I've attempted here; just an idea.)

vkryachko: I am not very familiar with this code (cc @ashwinraghav, who wrote it), but it looks like each resolver encapsulates parsing logic in addition to the url to fetch the license from, I can see your point though and given that the number of apache resolvers is increasing, it makes sense to factor this into the design.

Another approach comes to mind: favor composition over inheritance and instead of creating multiple implementations of the fetcher, there could be one fetcher that takes a `url` and a `String -> String` function that applies html -> text transform if required.

PR #468, opened 2019-05-28
Enable importing Bill of Materials from local Maven.

allisonbm92:
This change allows the smoke tests to pull Bill of Materials files from
the local Maven repository used by the main Firebase project.

### DIFF
--- a/smoke-tests/build.gradle
+++ b/smoke-tests/build.gradle
@@ -74,6 +74,12 @@ android {
 repositories {
   google()
   jcenter()
+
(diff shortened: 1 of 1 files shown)


smoke-tests/build.gradle
```diff
@@ -74,6 +74,9 @@ android {
 repositories {
   google()
   jcenter()
+  maven {
+    url "../build/m2repository/"
```
Corrob: Maybe add a comment describing what is in this repository and how it is created?

allisonbm92: Done.

PR #470, opened 2019-05-30
Moar Smoke Tests

allisonbm92:
Add more smoke tests.

### DIFF
--- a/smoke-tests/src/database/java/com/google/firebase/testing/database/DatabaseTest.java
+++ b/smoke-tests/src/database/java/com/google/firebase/testing/database/DatabaseTest.java
@@ -41,6 +41,29 @@ public final class DatabaseTest {
   @Rule public final ActivityTestRule<Activity> activity = new ActivityTestRule<>(Activity.class);
 
   @Test
+  public void setValueShouldFailWithPermissionDenied() throws Exception {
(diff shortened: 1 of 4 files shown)


smoke-tests/src/database/java/com/google/firebase/testing/database/DatabaseTest.java
```diff
@@ -40,6 +40,29 @@
 
   @Rule public final ActivityTestRule<Activity> activity = new ActivityTestRule<>(Activity.class);
 
+  @Test
+  public void setValueShouldFailWithPermissionDenied() throws Exception {
+    FirebaseAuth auth = FirebaseAuth.getInstance();
+    FirebaseDatabase database = FirebaseDatabase.getInstance();
+
+    auth.signOut();
+    Thread.sleep(1000); // TODO(allisonbm92): Introduce a better way to reduce flakiness.
```
yifanyang: This is waiting for sign out to complete?

allisonbm92: Yes. Unfortunately, signOut doesn't return a Task. If sleeping here proves to be a problem, I think I might be able to work around it, but so far, I haven't seen any flakiness with the sleep.

PR #475, opened 2019-05-31
Use `Preconditions.checkNotNull` in FirebaseFunctions.

rlazo:


rlazo:
cc @rosariopfernandes 

This is a follow up for #452 to use preconditions.check consistently across the code in `FirebaseFunctions.java`

PR #476, opened 2019-05-31
Refactor: Move network classes from util to remote

schmidt-sebastian:
This PR moves FirestoreChannel and IncomingStreamObserver from the firestore.util to the firestore.remote package.


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC
+// Copyright 2019 Google LLC
```
wilhuff: Since this is just moving the file around, shouldn't the copyright be the same as before? In other words, copyright applies to the file contents, not the filename.

schmidt-sebastian: This is auto-updated.

wilhuff: What's doing the auto-updating? If it's in a trigger can you disable the trigger?

schmidt-sebastian: It's applied by AndroidStudio when it creates a new file:

https://github.com/firebase/firebase-android-sdk/blob/master/.idea/copyright/Apache_2___Google.xml

I reverted the change manually.

PR #477, opened 2019-05-31
Moving SSL and GRPC Stub initialization to separate thread

schmidt-sebastian:
This is a follow up to https://github.com/firebase/firebase-android-sdk/pull/462 that doesn't break transactions.

As #462, this does not deal with the existing errors when there are no SSL ciphers available. This will be addressed in a different PR.

Performance numbers for a read from cache at client initialization:
Without this PR: 315ms, 318ms, 430ms, 323ms, 314ms
With PR: 279ms, 290ms, 291ms, 282ms, 319 ms

Note that `ProviderInstaller.installIfNeeded(context)` is documented to take up to 350ms.


firebase-firestore/src/main/java/com/google/firebase/firestore/util/GrpcCallProvider.java
```diff
@@ -0,0 +1,224 @@
...
+            TimeUnit.MILLISECONDS,
+            new LinkedBlockingQueue<>(),
+            threadFactory,
+            IGNORE_REJECTIONS_HANDLER);
+
+    // We execute network initialization on an internal queue to not block operations that
+    // depend on the AsyncQueue.
+    channelQueue.execute(
+        () -> {
+          FirestoreGrpc.FirestoreStub firestoreStub =
```
wilhuff: This seems like the reverse order to what came before (we did the provider installer stuff first in the initialize). Is this intended?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -0,0 +1,277 @@
...
+  private ManagedChannel initChannel(
+      AsyncQueue workerQueue, Context context, DatabaseInfo databaseInfo) {
+    try {
+      // We need to upgrade the Security Provider before any network channels are initialized.
+      // `OkHttp` maintains a list of supported providers that is initialized when the JVM first
+      // resolves the static dependencies of ManagedChannel.
+      ProviderInstaller.installIfNeeded(context);
+    } catch (GooglePlayServicesNotAvailableException /* Thrown by ProviderInstaller */
+        | GooglePlayServicesRepairableException /* Thrown by ProviderInstaller */
+        | IllegalStateException e /* Thrown by Robolectric */) {
```
schmidt-sebastian: Note that we can get rid of this if we change MockDatastore to not call the main Datastore constructor, which right now always initializes GrpcCallProvider.

schmidt-sebastian:
> While this is a neat approach, I do worry that it's a bit of a heavy-handed way to solve this problem. In particular, we're creating a dedicated thread that's not really necessary (after the initial SSL init) and we're creating a mostly-unnecessary layer of indirection and asynchrony over GrpcCall (grpc is already asynchronous).
> 
> How hard would it be to revive the original approach but change the transaction code so that it waits for SSL initialization before running the transaction or something?

There were three approaches. Note that all of them created a new thread:

- The original approach that blocked the AsyncQueue if the SSL initialization did not finish before the first request. That one is easy to port.

- The second approach that used `canUseNetwork()` inside of RemoteStore. That didn't work with transactions, we could have special-handled transactions by also blocking. The second approach also added race conditions to our Spec tests since it was never quite clear whether the streams were actually running.

- This approach.

An option that I considered was to drop the second thread once the initialization is done. This would mean that we only use the queue until the SSL channel is initialized, after which we would directly call into GRPC. Note that this adds some amount of complexity and would probably require dedicated testing (the existing code relies on the integration tests for coverage).

On a somewhat related note, I also hope that at some point we can move some of the Protobuf serialization and the WatchChangeAggregator from the main queue into a dedicated thread. With a lot of luck, we could use the executor in this PR for this.

Let me know what you think.

mikelehen:
Every active thread comes with a memory footprint of ~1MB I think (for the stack), so it's not great practice to have long-lived threads that aren't strictly necessary.  So if we can create the thread and then shut it down, that would be better.

AFAIK we don't want to block the AsyncQueue if we can avoid it...  So the second approach would be the one to revive.  If it makes things easier (for spec tests, etc.) I could maybe see the asynchronous wait for the SSL provider happening in AbstractStream.start() (as well as somewhere in the transaction call path) instead of in RemoteStore.  But I'll defer to you/Gil since you've spent more time puzzling over this.

I'm _partially_ won over by your hope to move protobuf serialization, etc. over to an alternate queue, but I think the code will end up looking substantially different at that point.  GrpcCall will be the wrong layer to enqueue to the AsyncQueue, etc...  So I don't think this change gets us materially closer to that.

So I would be interested in seeing the second approach revived and see how it compares to this one.  But I'll defer to you and Gil on whether this is warranted or not. If you're both happy with the current approach, I can acquiesce. :-)

schmidt-sebastian:
Updated the PR to with a lot of shiny new features:
- Instead of queuing everything on an Executor, we only create a short-lived thread for the ProviderInstaller and the Stub creation. 
FYI: This process kicks off the Auth-Token retrieval, which GRPC dispatches to the AsyncQueue when done (and any network call blocks on this inside of GRPC).

- We use the Task API to chain any operations that require the network on the completion of the above mentioned initialization.

- For the most part, Datastore hides the async behavior that this PR introduces. `runStreamingResponseRpc` and `runRpc` are already async, and `runBidiStreamingRpc` hides the async changes behind its slightly modified contract:
     - There is now an `onOpen()` callback (I could re-use the onReady() callback, but I am afraid that GRPC will eventually use this callback for something useful). The returned stream can only accept messages after the `onOpen` callback fires.
     - AbstractStream transitions streams from Initial to Started only when the `onOpen()` fires.

I still need to verify that Transactions work as expected and that the performance gain from this approach matches my previous measurements.

schmidt-sebastian:
> I still need to verify that Transactions work as expected and that the performance gain from this approach matches my previous measurements.

Verified: Transaction work on device, and the runtime of a read from cache is similar to the numbers reported in the PR description (282, 287, 269, 307, 289 and 304 ms).

schmidt-sebastian:
Note: All changes are in 176bb66.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -0,0 +1,198 @@
...
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.MethodDescriptor;
+import io.grpc.android.AndroidChannelBuilder;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Manages the gRPC channel and encapsulates all SSL and gRPC initialization.
+ *
+ * <p>All operations are dispatched to an internal worker queue and are not executed until the SSL
+ * and gRPC Stub initialization completes.
```
mikelehen: stale?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    ManagedChannelBuilder<?> channelBuilder;
+    if (overrideChannelBuilderSupplier != null) {
+      channelBuilder = overrideChannelBuilderSupplier.get();
+    } else {
+      channelBuilder = ManagedChannelBuilder.forTarget(databaseInfo.getHost());
+      if (!databaseInfo.isSslEnabled()) {
+        // Note that the boolean flag does *NOT* indicate whether or not plaintext should be
+        // used
```
mikelehen: Any chance you have any idea what this comment means?  It seems nonsensical.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/IncomingStreamObserver.java
```diff
@@ -25,7 +25,10 @@
   /** A message was received on the stream. */
   void onNext(RespT response);
 
-  /** The stream is "ready" (What the hell does that mean?!). */
+  /** The stream is open and able to accept messages. */
+  void onOpen();
+
+  /** The stream is "ready". */
```
mikelehen: You seem to have forgotten part of the comment. 😛 

Perhaps add a link to https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientCall.Listener.html#onReady-- and/or say "(additional messages can be sent to the stream without incurring excessive buffering)"

We could also just remove it since we don't use it for anything...  though I guess the log message might be meaningful.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -0,0 +1,198 @@
...
+      DatabaseInfo databaseInfo,
+      CallCredentials firestoreHeaders) {
+    this.asyncQueue = asyncQueue;
+
+    TaskCompletionSource<ManagedChannel> channelReady = new TaskCompletionSource<>();
+    this.channelTask = channelReady.getTask();
+
+    // We execute network initialization on a separate thread to not block operations that
+    // depend on the AsyncQueue.
+    new Thread("FirestoreGrpcInit") {
```
mikelehen: Rather than manually create our own thread, I wonder if it would be better to use `Tasks.call(AsyncTask.THREAD_POOL_EXECUTOR, ...)`

https://developers.google.com/android/reference/com/google/android/gms/tasks/Tasks.html#call(java.util.concurrent.Executor,%20java.util.concurrent.Callable%3CTResult%3E)
https://developer.android.com/reference/android/os/AsyncTask#THREAD_POOL_EXECUTOR

Unfortunately I don't see this pattern in existing Firebase code (or any usage of AsyncTask at all for that matter), so I'm a tiny bit scared there might be pitfalls.

Oh... Or even easier, can we just use https://developers.google.com/android/reference/com/google/android/gms/security/ProviderInstaller.html#installIfNeededAsync(android.content.Context,%20com.google.android.gms.security.ProviderInstaller.ProviderInstallListener) and not mess with threads ourselves?  That would call us back on the UI thread which isn't quite ideal but might be fine...

schmidt-sebastian: Re 2: That works, but it doesn't allow me to move the Stub initialization. I discovered the Async method on Friday when I was already fully committed to the idea that we should also move the initialization of the GRPC stub.

mikelehen: Heh.  Is the initialization of the gRPC stub measurably slow? :)

wilhuff: In my perf findings doc, of the 80ms we were spending on initialization, only half of it was the ProviderInstaller. Another ~30ms was gRPC stubs.

mikelehen: Cool, thanks.

schmidt-sebastian: Changed the code to run on Android's magical AsyncTask queue. I feel like a proper Android developer now.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
schmidt-sebastian: I am going to suspect this talks about Protobuf versus Plaintext, so uhm I updated it to the best of my knowledge.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/IncomingStreamObserver.java
schmidt-sebastian: I may have deliberately removed it since I found it kind of quirky :/ I just went one step further and removed the callback. If you think we need this logging (I am thinking we don't), we could also log in GrpcCallProvider.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
```diff
@@ -245,14 +247,6 @@ public void start() {
     // Note that Starting is only used as intermediate state until onOpen is called asynchronously,
     // since auth handled transparently by gRPC
```
mikelehen: I think this comment is kinda' stale and should probably be removed  (sorry I didn't notice before).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -75,23 +71,19 @@ public static void overrideChannelBuilder(
     this.asyncQueue = asyncQueue;
 
     TaskCompletionSource<ManagedChannel> channelReady = new TaskCompletionSource<>();
-    this.channelTask = channelReady.getTask();
 
-    // We execute network initialization on a separate thread to not block operations that
+    // We execute network initialization on a separate queue to not block operations that
```
mikelehen: Not certain why you changed thread => queue.  I think "thread" was and is more accurate.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -75,23 +71,19 @@ public static void overrideChannelBuilder(
     this.asyncQueue = asyncQueue;
 
     TaskCompletionSource<ManagedChannel> channelReady = new TaskCompletionSource<>();
```
mikelehen: unused now?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -116,8 +108,8 @@ private ManagedChannel initChannel(Context context, DatabaseInfo databaseInfo) {
     } else {
       channelBuilder = ManagedChannelBuilder.forTarget(databaseInfo.getHost());
       if (!databaseInfo.isSslEnabled()) {
-        // Note that the boolean flag does *NOT* indicate whether or not plaintext should be
-        // used
+        // Note that the boolean flag does *NOT* switch the wire format from Protobuf to Plaintext.
+        // It merely turns off SSL encryption.
```
mikelehen: Oohhh...  heh.  Good call.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
schmidt-sebastian: Removed

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
schmidt-sebastian: Undone. Not sure what the best term here is since the thread might already exist.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
schmidt-sebastian: Yup. Removed.

PR #479, opened 2019-06-02
Support only using ktx artifact

SUPERCILEX:
See #480


SUPERCILEX:
See https://github.com/firebase/firebase-android-sdk/issues/480#issuecomment-498807177

PR #484, opened 2019-06-03
Reduce flakiness of license fetcher.

allisonbm92:
This commit is a simple fix to reduce the flakiness of downloading
licenses. The fetcher will now try up to three times and will insert a
small delay between each attempt. This delay will only affect builds
that would otherwise fail.


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -14,155 +14,159 @@
...
-      TEXT_FORMATTER.getPlainText(doc.select('#body-content > div.jd-descr > div')[0])
-    }
+  /** Extracts the license text from the rest of the document. */
+  String processDocument(Document doc) {
+    return TEXT_FORMATTER.getPlainText(doc)
   }
 
-  static final class Apache2LicenseFetcher implements RemoteLicenseFetcher {
-    private URI APACHE_2_LICENSE_URI = URI.create("http://www.apache.org/licenses/LICENSE-2.0.txt")
+  static final class AndroidSdkTermsFetcher extends RemoteLicenseFetcher {
```
yifanyang: Is it proper to have these Fetchers live in their own files, instead of being a nested static class under `RemoteLicenseFetcher`?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -14,155 +14,159 @@
...
     }
-  }
-
-  static final class YetAnotherApache2LicenseFetcher implements RemoteLicenseFetcher {
-    private URI APACHE_2_LICENSE_URI = URI.create("http://www.apache.org/licenses/LICENSE-2.0")
 
     @Override
-    URI getServiceUri() {
-      APACHE_2_LICENSE_URI
+    String processDocument(Document doc) {
```
yifanyang: Looks like the difference among all `processDocument` lies essentially on the CSS selector part. Maybe just have the selector as a field so that we don't have to override `processDocument` for each fetcher?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
allisonbm92: I think the nested classes are fine here.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
allisonbm92: For the most part, that is true. However, some of the licenses are plain-text, so we skip the select method entirely. Having a method makes that a bit easier.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -14,155 +14,159 @@
...
+      super("http://www.gnu.org/software/classpath/license.html")
     }
 
     @Override
-    String get() {
-      def doc = Jsoup.connect(GNU_CLASSPATH_LICENSE_URI.toString()).get()
-
-      TEXT_FORMATTER.getPlainText(doc.select('body > table > tbody > tr:nth-child(2) > td:nth-child(2) > table > tbody > tr:nth-child(3) > td > en > blockquote'))
+    String processDocument(Document doc) {
+      return TEXT_FORMATTER.getPlainText(doc.select("body > table > tbody > tr:nth-child(2) > td:nth-child(2) > table > tbody > tr:nth-child(3) > td > en > blockquote").get(0))
```
Corrob: Keep line length <100 characters?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -14,155 +14,159 @@
...
   }
 
-  static final class GnuClasspathLicenseFetcher implements RemoteLicenseFetcher {
-    private URI GNU_CLASSPATH_LICENSE_URI = URI.create("http://www.gnu.org/software/classpath/license.html")
+  static final class GnuClasspathLicenseFetcher extends RemoteLicenseFetcher {
 
-    @Override
-    URI getServiceUri() {
-      GNU_CLASSPATH_LICENSE_URI
+    // TODO(allisonbm92, vkryachko): This does not fetch the actual license!
```
Corrob: Can you put the missing work in this TODO? Why doesn't it fetch the actual license?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
yifanyang: OK. I just feel the full name could be a little confusing, since it is now `<the-package-name>.RemoteLicenseFetcher.AndroidSdkTermsFetcher`. But it is totally up to you.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
vkryachko: +1 to @yifanyang 's comment, this is a good candidate for "prefer composition over inheritance", although to keep the code as expressive as it currently is, I'd suggest to have a member `Function<Document, String>` which could either be identity or "find by selector". so instead of this class hierarchy we end up with a list of values of the form:
```java
LicenseFetcher("http://example.com", d -> extractFrom(d));
```

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -14,155 +14,159 @@
...
-      APACHE_2_LICENSE_URI
+    AndroidSdkTermsFetcher() {
+      super("https://developer.android.com/studio/terms.html")
     }
 
     @Override
-    String get() {
-      APACHE_2_LICENSE_URI.toURL().getText()
+    String processDocument(Document doc) {
+      // TODO(vkryachko, allisonbm92): This fails silently.
```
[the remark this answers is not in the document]
vkryachko: yeah, why do you return an empty string?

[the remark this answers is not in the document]
allisonbm92: Yes, I've expanded details in the comments and removed the empty string.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
allisonbm92: Done.

PR #485, opened 2019-06-03
Storage: Port test changes from cr/250550507

schmidt-sebastian:
Manual port since we don't have Copybara set up yet.

https://critique.corp.google.com/#review/250550507

### DIFF
--- a/firebase-storage/src/test/java/com/google/firebase/storage/PathingTest.java
+++ b/firebase-storage/src/test/java/com/google/firebase/storage/PathingTest.java
@@ -36,7 +36,7 @@ import org.robolectric.annotation.Config;
 
 /** Tests for {@link FirebaseStorage}. */
 @RunWith(RobolectricTestRunner.class)
-@Config(manifest = Config.NONE, sdk = Build.VERSION_CODES.LOLLIPOP_MR1)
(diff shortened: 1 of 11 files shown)


schmidt-sebastian:
Unassigning since there is more cleanup to port (the CL was updated).

schmidt-sebastian:
> Unassigning since there is more cleanup to port (the CL was updated).

The new changes in the CL were already applied to this repo.

firebase-storage/src/test/java/com/google/firebase/storage/PathingTest.java
```diff
@@ -164,7 +164,7 @@ public void dotdotTest() throws Exception {
         FirebaseStorage.getInstance()
             .getReferenceFromUrl("gs://benwu-test1" + ".storage.firebase.com/child/../");
 
-    assertEquals(ref.getName(), "..");
+    assertEquals("..", ref.getName());
```
rsgowman: I have to look up this ordering, every. single. time. (FTR, you've done the correct thing.)

However, if you're going to go to the effort of fixing this, you might consider migrating to assertThat instead. It makes it much more obvious.

Optional.

schmidt-sebastian: Keeping it as is to match the base CL.

PR #487, opened 2019-06-03
Storage: Add TestUtil.await

schmidt-sebastian:
Addresses some leftover feedback from https://github.com/firebase/firebase-android-sdk/pull/458 and adds a shared TestUtil.await() that mimics Task.await() and removes a lot of duplicated code in our unit tests.


firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
```diff
@@ -137,33 +129,25 @@ public void cantUploadToRoot() throws Exception {
...
-          Assert.assertEquals(taskException.get().getCause(), e.getCause());
-          // Task failed.
-        }
-
-        Assert.assertEquals(taskException.get().getCause(), task.getException().getCause());
-
-        return;
-      }
-      Thread.sleep(1);
+    TestUtil.await(task, 300, TimeUnit.SECONDS);
```
schmidt-sebastian: Note: Some of the timeouts are VERY questionable, but I kept them as is to make this purely a refactor.

rsgowman: Yeah, this looks a little insane. However, this test looks to be exercising network failures, so, as an integration test, it might not be unreasonable for this to take multiple minutes to fail. But it's a unit test; so hopefully it doesn't actually take anywhere near that long. (Maybe 5m was used when writing the test against the real network? Then it was converted to a unit test, but the timeout wasn't fixed?)

Regardless, keeping it the same to keep this PR as a refactor sounds reasonable. You could optionally add a TODO to investigate whether this value still makes sense. (If you do, write the todo to apply to all non default timeouts.)

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
```diff
@@ -420,25 +364,19 @@ public void fileUploadWithPauseCancel() throws Exception {
...
-      if (task.isComplete()) {
-        // success!
-        TestUtil.verifyTaskStateChanges("fileUploadWithPauseCancel", task.getResult().toString());
-        Assert.assertTrue(ResumableUploadCancelRequest.CANCEL_CALLED);
-        return;
-      }
-      Thread.sleep(1);
-    }
-    Assert.fail();
+    TestUtil.await(task, 20, TimeUnit.SECONDS);
```
rsgowman: Yeah this ^. This non-default timeout is documented, so we know it's deliberate and meaningful. We should (eventually) do that for all of them. You could even enforce that by changing `TestUtil.await(Task, int, TimeUnit)` to `TestUtil.await(Task, int, TimeUnit, String reasonWhyWereNotUsingTheDefault)` (which is a little odd, but would be pretty effective.) Or, since I see this same reason is used again below, `TestUtil.awaitFairnessBug(task)`

(Aside: 20s timeout in a "unit" test is questionable.)

No action required.

firebase-storage/src/test/java/com/google/firebase/storage/TestUtil.java
```diff
@@ -130,4 +133,31 @@ private static void verifyTaskStateChanges(@Nullable InputStream inputStream, St
...
+    for (int i = 0; i < timeoutMillis; i++) {
+      Robolectric.flushForegroundThreadScheduler();
+      if (task.isComplete()) {
+        // success!
+        return;
+      }
+      Thread.sleep(1);
+    }
+
+    org.junit.Assert.fail();
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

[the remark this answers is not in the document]
schmidt-sebastian: ... and done!

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
schmidt-sebastian: I added TODOs to all timeouts greater than 5 seconds with the exception of those that were already commented.

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
```diff
@@ -137,33 +129,25 @@ public void cantUploadToRoot() throws Exception {
...
-      }
-      Thread.sleep(1);
+    TestUtil.await(task, 300, TimeUnit.SECONDS);
+
+    try {
+      task.getResult();
+      Assert.fail();
+    } catch (RuntimeExecutionException e) {
+      Assert.assertEquals(taskException.get().getCause(), e.getCause().getCause());
+      // Task failed.
```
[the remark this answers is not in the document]
schmidt-sebastian: But.... how will anyone understand this sophisticated logic? :) Removed.

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
schmidt-sebastian: Acknowledged, but left the timeouts and the comment as is for now.

PR #494, opened 2019-06-05
Create `publishChangedToBuildDir` task.

allisonbm92:
This change adds this new task. It leverages the implementation used for
the `checkChanged` and similar tasks.

### DIFF
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
@@ -14,10 +14,12 @@
 
 package com.google.firebase.gradle.plugins.publish
 
+import com.google.firebase.gradle.plugins.ci.AffectedProjectFinder
 import com.google.firebase.gradle.plugins.FirebaseLibraryExtension
 import digital.wup.android_maven_publish.AndroidMavenPublishPlugin
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.tasks.bundling.Jar
 import org.gradle.api.tasks.bundling.Zip
@@ -83,7 +85,11 @@ class PublishingPlugin implements Plugin<Project> {
 
         def publishAllToLocal = project.task('publishAllToLocal')
         def publishAllToBuildDir = project.task('publishAllToBuildDir')
+        def publishChangedToBuildDir = project.task('publishChangedToBuildDir')
         def firebasePublish = project.task('firebasePublish')
+
+        def changedProjects =
+                new AffectedProjectFinder(project, changedPaths(project.rootDir), []).find()
 
         project.getGradle().projectsEvaluated {
             project.subprojects { Project sub ->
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -121,10 +127,27 @@ class PublishingPlugin implements Plugin<Project> {
...
+                    // Add changed projects to publishChangedToBuildDir.
+                    if (changedProjects.contains(sub)) {
+                        publishChangedToBuildDir.dependsOn "$sub.path:publishMavenAarPublicationToBuildDirRepository"
+                        def dependencies = sub.configurations.implementation.dependencies
+                        for (def dep : dependencies) {
+                            if (dep instanceof ProjectDependency) {
+                                def task = dep.getDependencyProject()
+                                        .tasks.findByName("publishMavenAarPublicationToBuildDirRepository")
+                                if (task != null) {
+                                    publishChangedToBuildDir.dependsOn task
```
yifanyang: Why do we need to make `publishChangedToBuildDir` directly depend on `sub`'s dependencies' `publishMavenAarPublicationToBuildDirRepository`? Shouldn't `sub`'s `publishMavenAarPublicationToBuildDirRepository` already depend on it's dependencies' `publishMavenAarPublicationToBuildDirRepository`?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -155,6 +178,13 @@ class PublishingPlugin implements Plugin<Project> {
         }
     }
 
+    private static Set<String> changedPaths(File workDir) {
+        return 'git diff --name-only --submodule=diff HEAD@{0} HEAD@{1}'
```
yifanyang: I would prefer to see if git can be run programmatically. But it is probably fine if you think this task will be invoked only on machines that has git available.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -121,10 +127,27 @@ class PublishingPlugin implements Plugin<Project> {
                             publisher.decorate(sub, it)
                         }
                     }
+
+                    // Add changed projects to publishChangedToBuildDir.
+                    if (changedProjects.contains(sub)) {
+                        publishChangedToBuildDir.dependsOn "$sub.path:publishMavenAarPublicationToBuildDirRepository"
+                        def dependencies = sub.configurations.implementation.dependencies
```
vkryachko: nit: `s/implementation/releaseRuntimeClasspath/`

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -121,10 +127,27 @@ class PublishingPlugin implements Plugin<Project> {
...
+                        for (def dep : dependencies) {
```
vkryachko: What's the intent here? Why do we need to publish dependencies of the project? Also, if it is needed indeed, don't we also want to publish dependencies of dependencies recursively?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -155,6 +178,13 @@ class PublishingPlugin implements Plugin<Project> {
         }
     }
 
+    private static Set<String> changedPaths(File workDir) {
```
vkryachko: afaicr, we have the same git command elsewhere, can we centralize this in one place to avoid duplication?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
allisonbm92: That's not how those tasks are configured. They don't publish their dependencies.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
allisonbm92: It diffs head and head - 1. I copied this from the checkChanged tasks. I think this is primarily intended for Prow, as it might give misleading results locally.

PR #506, opened 2019-06-08
feat: add docs for functions-ktx

rosariopfernandes:
I do realize the functions-ktx library hasn't been released yet, so maybe this should only be merged after its release.


vkryachko:
Hey @rosariopfernandes , thanks for the PR! You're right, we will be able to merge this once we've made at least one release of functions-ktx(should happen within a month or so). So let's hold on to this till then

PR #513, opened 2019-06-11
Firebase Storage: Fix typo in method name

schmidt-sebastian:


firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java
```diff
@@ -110,7 +110,7 @@ public static String getAuthority() {
    * @return Url for the target REST call in string form.
    */
   @NonNull
-  public static String getdefaultURL(@NonNull Uri gsUri) {
+  public static String getDefaultURL(@NonNull Uri gsUri) {
```
schmidt-sebastian: According to Java Style, this should be `getDefaultUrl`, but I wanted to match `getUrl`. Let me know what you think :)

fredzqm: I would prefer `toURL` or `convertToURL`. `get` gave me an impression that this is a getter, but it actually converts a gcs `Uri` to a `string`.

I also don't see `default` anywhere in the method.

PR #515, opened 2019-06-12
Output artifact list during local publishing.

allisonbm92:
This effort replaces #494 by generating a list of artifacts instead of selectively publishing them. The `publishAllToBuildDir` task now produces an extra JSON file containing the names of all published artifacts. It also includes the artifacts that are affected by the most recent commit.

Our compatibility testing infrastructure will be able to use this listing to build two variants of the tests: one against head and the other against the latest release (except for the affected artifacts).

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -37,6 +37,7 @@ dependencies {
     implementation 'org.jsoup:jsoup:1.11.2'
     implementation 'digital.wup:android-maven-publish:3.6.2'
     implementation 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.20'
+    implementation 'org.json:json:20180813'
 
     implementation 'io.opencensus:opencensus-api:0.18.0'
     implementation 'io.opencensus:opencensus-exporter-stats-stackdriver:0.18.0'
@@ -44,7 +45,6 @@ dependencies {
 
     implementation 'com.android.tools.build:gradle:3.2.1'
     testImplementation 'junit:junit:4.12'
-    testImplementation 'org.json:json:20180813'
     testImplementation('org.spockframework:spock-core:1.1-groovy-2.4') {
         exclude group: 'org.codehaus.groovy'
     }
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/AffectedProjectFinder.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/AffectedProjectFinder.groovy
@@ -25,6 +25,10 @@ class AffectedProjectFinder {
     Set<String> changedPaths;
 
     @Builder
+    AffectedProjectFinder(Project project, List<Pattern> ignorePaths) {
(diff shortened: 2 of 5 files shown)


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -121,11 +131,40 @@ class PublishingPlugin implements Plugin<Project> {
...
                         }
                     }
+
                     publishAllToLocal.dependsOn "$sub.path:publishMavenAarPublicationToMavenLocal"
                     publishAllToBuildDir.dependsOn "$sub.path:publishMavenAarPublicationToBuildDirRepository"
 
+                    // Update the changed and all sets each time an artifact is published. This is
+                    // used to build the JSON file of affected targets.
+                    sub.tasks.getByName("publishMavenAarPublicationToBuildDirRepository").doLast {
+                        def artifact = "$sub.group:$sub.name:$sub.version-SNAPSHOT"
```
vkryachko: this will not work with ktx libraries as it's name is just `ktx` instead of `sdk-name-ktx`. One way you could try and solve it is by getting `groupId` and `artifactId` from the `mavenAar` Publication object

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
```diff
@@ -121,11 +131,40 @@ class PublishingPlugin implements Plugin<Project> {
                             publisher.decorate(sub, it)
                         }
                     }
+
                     publishAllToLocal.dependsOn "$sub.path:publishMavenAarPublicationToMavenLocal"
                     publishAllToBuildDir.dependsOn "$sub.path:publishMavenAarPublicationToBuildDirRepository"
 
+                    // Update the changed and all sets each time an artifact is published. This is
+                    // used to build the JSON file of affected targets.
+                    sub.tasks.getByName("publishMavenAarPublicationToBuildDirRepository").doLast {
```
vkryachko: Why does this have to be part of this task? Don't we have all the required info at gradle startup/configuration phase?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
allisonbm92: Good point. It should be fixed now.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/publish/PublishingPlugin.groovy
allisonbm92: I suppose it doesn't. I was thinking of adding each as it finished being built, but I suppose there's no real reason for that.

allisonbm92:
@vkryachko PTAL

PR #516, opened 2019-06-12
Update protos to include IN and ARRAY_CONTAINS_ANY

thebrianchen:


firebase-firestore/src/proto/google/firestore/v1/query.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC.
+// Copyright 2019 Google LLC.
```
mikelehen: Note that typically we don't update the copyright date in existing files when editing them.  But in this case, the official proto (https://github.com/googleapis/googleapis/blob/master/google/firestore/v1beta1/firestore.proto) _does_ have copyright 2019 in it and so once we get back in sync with it, we'll get this change anyway, so it's fine.

PR #519, opened 2019-06-13
Add IN and ARRAY_CONTAINS_ANY (not publicly exposed)

thebrianchen:
Porting over from [Web SDK](https://github.com/firebase/firebase-js-sdk/pull/1843).

Pending adding support to run against Firestore Emulator.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Semaphore;
 import org.junit.After;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -427,7 +428,9 @@ public class QueryTest {
     // much of anything else interesting to test.
   }
 
+  // TODO(in-queries): Re-enable once emulator support is added to travis.
   @Test
+  @Ignore
   public void testQueriesCanUseInFilters() {
     Map<String, Object> docA = map("zip", 98101);
     Map<String, Object> docB = map("zip", 91102);
@@ -448,7 +451,9 @@ public class QueryTest {
     assertEquals(asList(docF), querySnapshotToValues(snapshot));
   }
 
+  // TODO(in-queries): Re-enable once emulator support is added to travis.
   @Test
+  @Ignore
   public void testQueriesCanUseArrayContainsAnyFilters() {
     Map<String, Object> docA = map("array", asList(42L));
     Map<String, Object> docB = map("array", asList("a", 42L, "c"));
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -420,11 +420,56 @@ public class QueryTest {
         testCollectionWithDocs(map("a", docA, "b", docB, "c", docC, "d", docD));
 
     // Search for "array" to contain 42
-    QuerySnapshot snapshot = waitFor(collection.whereArrayContains("array", 42L).get());
+    QuerySnapshot snapshot = waitFor(collection.whereArrayContains("arrays", 42L).get());
     assertEquals(asList(docA, docB, docD), querySnapshotToValues(snapshot));
 
     // NOTE: The backend doesn't currently support null, NaN, objects, or arrays, so there isn't
     // much of anything else interesting to test.
+  }
+
+  @Test
+  public void testQueriesCanUseInFilters() {
+    Map<String, Object> docA = map("zip", 98101);
+    Map<String, Object> docB = map("zip", 91102);
+    Map<String, Object> docC = map("zip", 98103);
+    Map<String, Object> docD = map("zip", asList(98101));
+    Map<String, Object> docE = map("zip", asList("98101", map("zip", 98101)));
+    Map<String, Object> docF = map("zip", map("code", 500));
+    CollectionReference collection =
+        testCollectionWithDocs(
+            map("a", docA, "b", docB, "c", docC, "d", docD, "e", docE, "f", docF));
+
+    // Search for zips matching [98101, 98103].
+    QuerySnapshot snapshot = waitFor(collection.whereIn("zip", asList(98101, 98103)).get());
+    assertEquals(asList(docA, docC), querySnapshotToValues(snapshot));
+
+    // With objects.
+    snapshot = waitFor(collection.whereIn("zip", asList(map("code", 500))).get());
+    assertEquals(asList(docF), querySnapshotToValues(snapshot));
+  }
+
+  @Test
+  public void testQueriesCanUseArrayContainsAnyFilters() {
+    Map<String, Object> docA = map("array", asList(42L));
+    Map<String, Object> docB = map("array", asList("a", 42L, "c"));
+    Map<String, Object> docC = map("array", asList(41.999, "42", map("a", asList(42))));
+    Map<String, Object> docD = map("array", asList(42L), "array2", asList("bingo"));
+    Map<String, Object> docE = map("array", asList(43L));
+    Map<String, Object> docF = map("array", asList(map("a", 42L)));
+    Map<String, Object> docG = map("array", 42L);
+
+    CollectionReference collection =
+        testCollectionWithDocs(
+            map("a", docA, "b", docB, "c", docC, "d", docD, "e", docE, "f", docF));
+
+    // Search for "array" to contain [42, 43].
+    QuerySnapshot snapshot =
+        waitFor(collection.whereArrayContainsAny("array", asList(42L, 43L)).get());
+    assertEquals(asList(docA, docB, docD, docE), querySnapshotToValues(snapshot));
+
+    // With objects.
+    snapshot = waitFor(collection.whereArrayContainsAny("array", asList(map("a", 42L))).get());
+    assertEquals(asList(docF), querySnapshotToValues(snapshot));
   }
 
   @Test
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -420,7 +420,7 @@ public class QueryTest {
         testCollectionWithDocs(map("a", docA, "b", docB, "c", docC, "d", docD));
 
     // Search for "array" to contain 42
-    QuerySnapshot snapshot = waitFor(collection.whereArrayContains("arrays", 42L).get());
+    QuerySnapshot snapshot = waitFor(collection.whereArrayContains("array", 42L).get());
     assertEquals(asList(docA, docB, docD), querySnapshotToValues(snapshot));
 
     // NOTE: The backend doesn't currently support null, NaN, objects, or arrays, so there isn't
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
@@ -650,21 +650,20 @@ public class ValidationTest {
         "Invalid Query. 'array_contains_any' filters support a maximum of 10 elements in the value array.");
 
     expectError(
-            () -> testCollection().whereIn("bar", asList("foo", null)),
-            "Invalid Query. 'in' filters cannot contain 'null' in the value array.");
+        () -> testCollection().whereIn("bar", asList("foo", null)),
+        "Invalid Query. 'in' filters cannot contain 'null' in the value array.");
 
     expectError(
-            () -> testCollection().whereArrayContainsAny("bar", asList("foo", null)),
-            "Invalid Query. 'array_contains_any' filters cannot contain 'null' in the value array.");
+        () -> testCollection().whereArrayContainsAny("bar", asList("foo", null)),
+        "Invalid Query. 'array_contains_any' filters cannot contain 'null' in the value array.");
 
     expectError(
-            () -> testCollection().whereIn("bar", asList("foo", Double.NaN)),
-            "Invalid Query. 'in' filters cannot contain 'NaN' in the value array.");
+        () -> testCollection().whereIn("bar", asList("foo", Double.NaN)),
+        "Invalid Query. 'in' filters cannot contain 'NaN' in the value array.");
 
     expectError(
-            () -> testCollection().whereArrayContainsAny("bar", asList("foo", Float.NaN)),
-            "Invalid Query. 'array_contains_any' filters cannot contain 'NaN' in the value array.");
-
+        () -> testCollection().whereArrayContainsAny("bar", asList("foo", Float.NaN)),
+        "Invalid Query. 'array_contains_any' filters cannot contain 'NaN' in the value array.");
   }
 
   @Test
@@ -703,6 +702,12 @@ public class ValidationTest {
     expectError(() -> collection.whereGreaterThanOrEqualTo(FieldPath.documentId(), 1), reason);
 
     reason =
+        "Invalid query. When querying with FieldPath.documentId() you must provide "
+            + "a valid String or DocumentReference, but it was of type: java.util.Arrays$ArrayList";
+    ;
+    expectError(() -> collection.whereIn(FieldPath.documentId(), asList(1, 2)), reason);
+
+    reason =
         "Invalid query. When querying a collection group by FieldPath.documentId(), the value "
             + "provided must result in a valid document path, but 'foo' is not because it has "
             + "an odd number of segments (1).";
@@ -720,9 +725,6 @@ public class ValidationTest {
         "Invalid query. You can't perform 'array_contains_any' queries on FieldPath.documentId().";
     expectError(
         () -> collection.whereArrayContainsAny(FieldPath.documentId(), asList(1, 2)), reason);
-
-    reason = "Invalid query. You can't perform 'in' queries on FieldPath.documentId().";
-    expectError(() -> collection.whereIn(FieldPath.documentId(), asList(1, 2)), reason);
   }
 
   // Helpers
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
@@ -704,7 +704,6 @@ public class ValidationTest {
     reason =
         "Invalid query. When querying with FieldPath.documentId() you must provide "
             + "a valid String or DocumentReference, but it was of type: java.util.Arrays$ArrayList";
-    ;
     expectError(() -> collection.whereIn(FieldPath.documentId(), asList(1, 2)), reason);
 
     reason =
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
@@ -332,11 +332,7 @@ public class ValidationTest {
     List<String> badFieldPaths =
         asList("foo~bar", "foo*bar", "foo/bar", "foo[1", "foo]1", "foo[1]");
     for (String fieldPath : badFieldPaths) {
-      String reason =
-          "Invalid field path ("
-              + fieldPath
-              + "). Paths must not contain '~', '*', '/', '[', or ']'";
-      verifyFieldPathThrows(fieldPath, reason);
+      verifyFieldPathThrows(fieldPath, "Use FieldPath.of() for field names containing '~*/[]'.");
     }
   }
 
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
@@ -652,6 +652,23 @@ public class ValidationTest {
         () ->
             testCollection().whereArrayContainsAny("bar", asList(1, 2, 3, 4, 5, 6, 7, 8, 9, 9, 9)),
         "Invalid Query. 'array_contains_any' filters support a maximum of 10 elements in the value array.");
+
+    expectError(
+            () -> testCollection().whereIn("bar", asList("foo", null)),
+            "Invalid Query. 'in' filters cannot contain 'null' in the value array.");
+
+    expectError(
+            () -> testCollection().whereArrayContainsAny("bar", asList("foo", null)),
+            "Invalid Query. 'array_contains_any' filters cannot contain 'null' in the value array.");
+
+    expectError(
+            () -> testCollection().whereIn("bar", asList("foo", Double.NaN)),
+            "Invalid Query. 'in' filters cannot contain 'NaN' in the value array.");
+
+    expectError(
+            () -> testCollection().whereArrayContainsAny("bar", asList("foo", Float.NaN)),
+            "Invalid Query. 'array_contains_any' filters cannot contain 'NaN' in the value array.");
+
   }
 
   @Test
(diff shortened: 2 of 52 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -167,16 +167,21 @@ public FieldPath inequalityField() {
     return null;
   }
 
-  public boolean hasArrayContainsFilter() {
+  @Nullable
+  /**
+   * Checks if any of the provided RelationOps are included in the query and returns the first one
```
mikelehen: "RelationOps" => "filter operators"?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -167,16 +167,21 @@ public FieldPath inequalityField() {
...
+   * that is, or null if none are.
+   */
+  public Operator findOperatorFilter(List<Operator> filterOps) {
```
mikelehen: optional: Looking at this code today, "findOperatorFilter()" doesn't seem like the clearest name.  And it's unfortunate that TypeScript and Android end up with different names.  I would be slightly in favor of renaming both to "findFilterOperator()"

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -301,6 +329,72 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
     return whereHelper(fieldPath, Operator.ARRAY_CONTAINS, value);
   }
 
+  /**
+   * Creates and returns a new Query with the additional filter that documents must contain the
+   * specified field, the value must be an array, and that the array must contain any values of the
```
mikelehen: "any values of" => "at least one value from" ?

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -301,6 +329,72 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
...
+   * provided array.
+   *
+   * <p>A Query can have only one whereArrayContainsAny() filter.
+   *
+   * @param field The name of the field containing an array to search
```
mikelehen: period

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -301,6 +329,72 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
     return whereHelper(fieldPath, Operator.ARRAY_CONTAINS, value);
   }
 
+  /**
+   * Creates and returns a new Query with the additional filter that documents must contain the
+   * specified field, the value must be an array, and that the array must contain any values of the
+   * provided array.
+   *
+   * <p>A Query can have only one whereArrayContainsAny() filter.
```
mikelehen: Should we document the other restrictions?  e.g.  "A query can have only one whereArrayContainsAny() filter and it cannot be combined with whereArrayContains() or whereIn()."

And do similar for whereArrayContains() and whereIn().

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -301,6 +329,72 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
...
+   */
+  @NonNull
+  @PublicApi
+  public Query whereArrayContainsAny(@NonNull FieldPath fieldPath, @NonNull List<Object> value) {
+    return whereHelper(fieldPath, Operator.ARRAY_CONTAINS_ANY, value);
+  }
+
+  /**
+   * Creates and returns a new Query with the additional filter that documents must contain the
+   * specified field and that the field must equal any values of the provided array.
```
mikelehen: I'd maybe say, "must contain the specified field and the value must equal one of the values from the provided array."

firebase-firestore/src/main/java/com/google/firebase/firestore/core/RelationFilter.java
```diff
@@ -74,6 +77,17 @@ public boolean matches(Document doc) {
   private boolean matchesValue(FieldValue other) {
     if (operator == Operator.ARRAY_CONTAINS) {
       return other instanceof ArrayValue && ((ArrayValue) other).getInternalValue().contains(value);
+    } else if (operator == Operator.IN) {
+      return value instanceof ArrayValue && ((ArrayValue) value).getInternalValue().contains(other);
```
mikelehen: This instanceof check on `value` is a little misleading because `value` will always be an `ArrayValue` (unless we have a bug in our query construction code), so I would rewrite this as:
```
hardAssert(value instanceof ArrayValue, "IN filter has invalid value: " + value);
if (((ArrayValue)value).getInternalValue()...
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/RelationFilter.java
```diff
@@ -74,6 +77,17 @@ public boolean matches(Document doc) {
...
+    } else if (operator == Operator.ARRAY_CONTAINS_ANY) {
+      if (other instanceof ArrayValue && value instanceof ArrayValue) {
```
mikelehen: this `instanceof` check on `value` should be an assert as well (but not for `other` since `other` is the document field value and may or may not be an array).

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/RemoteSerializerTest.java
```diff
@@ -577,8 +577,13 @@ public void testEncodesMultipleFiltersOnDeeperCollections() {
         Query.atPath(ResourcePath.fromString("rooms/1/messages/10/attachments"))
             .filter(filter("prop", "<", 42))
             .filter(filter("author", "==", "dimond"))
-            .filter(filter("tags", "array-contains", "pending"));
+            .filter(filter("tags", "array-contains", "pending"))
+            .filter(filter("tags", "in", asList("pending", "dimond")));
     Target actual = serializer.encodeTarget(wrapQueryData(q));
+    ArrayValue.Builder array =
```
mikelehen: Can we come up with a name more descriptive than `array`? 😄

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/RemoteSerializerTest.java
```diff
@@ -611,7 +616,80 @@ public void testEncodesMultipleFiltersOnDeeperCollections() {
...
+    assertEquals(serializer.decodeQueryTarget(serializer.encodeQueryTarget(q)), q);
+  }
+
+  @Test
+  public void testEncodesDisjunctiveFiltersOnDeeperCollections() {
+    Query q =
+        Query.atPath(ResourcePath.fromString("rooms/1/messages/10/attachments"))
+            .filter(filter("prop", "<", 42))
+            .filter(filter("author", "==", "dimond"))
+            .filter(filter("tags", "array-contains-any", asList("pending", "dimond")));
```
mikelehen: So...  
1) This test claims to test "disjunctive filters" but I think only tests array-contains-any, not in, right?
2) This test has a ton of unnecessary duplication of the previous test (we don't need to test <, ==, or order-bys again).

I think what happened is Android for whatever reason originally had one ginormous test instead of individual tests.  But now that we're trying to test filters that can't be combined, that's biting us.  So I would try to create more isolated minimal tests, ideally moving us closer to being in-sync with the way the tests work on TypeScript (and probably iOS though I haven't looked).

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -427,6 +427,51 @@ public void testQueriesCanUseArrayContainsFilters() {
     // much of anything else interesting to test.
   }
 
+  @Test
+  public void testQueriesCanUseInFilters() {
```
mikelehen: These tests only work against the emulator, right?  So we need to figure out a way to guard / disable them for now?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
```diff
@@ -408,6 +408,12 @@ public void queriesWithNullOrNaNFiltersOtherThanEqualityFail() {
     expectError(
         () -> collection.whereArrayContains("a", null),
         "Invalid Query. You can only perform equality comparisons on null (via whereEqualTo()).");
+    expectError(
+        () -> collection.whereArrayContainsAny("a", null),
+        "Invalid Query. A non-empty array is required for 'array_contains_any' filters.");
+    expectError(
+        () -> collection.whereIn("a", null),
+        "Invalid Query. A non-empty array is required for 'in' filters.");
```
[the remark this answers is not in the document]
thebrianchen: The signatures require an array, so you can't pass in NaN. Should we check for passing in NaN into the array, in which case I should port this over to web as well?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
thebrianchen: Yeah... Still trying to figure out how to even run the tests against the emulator, but i'll update this after.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
thebrianchen: Good idea. I'll leave this comment unresolved until this PR is close to mergeable and I'll update the TS sdk.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
```diff
@@ -539,10 +545,113 @@ public void queriesWithInequalityDifferentThanFirstOrderByFail() {
...
+    expectError(
+        () ->
+            testCollection()
+                .whereArrayContainsAny("foo", asList(1, 2))
+                .whereArrayContains("foo", 1),
+        "Invalid Query. You cannot use 'array_contains' filters with 'array_contains_any' filters.");
+  }
+
+  @Test
+  public void queriesWithMultipleDisjunctiveFiltersFail() {
```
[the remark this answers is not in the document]
thebrianchen: I think this organization is better since it breaks these into 4 main tests:
- multiple array filters
- multiple disjunctive filters
- verify IN works with ARRAY_CONTAINS
- array input rules

I'll port this over to web once I finish iOS

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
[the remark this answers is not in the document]
rockwotj: FWIW if you do something like:

`whereIn("a", Arrays.asList(null, NaN))` you'll never get any results.

`IN [null] has the same semantics as == null` and the same goes for `NaN`.

You might consider rejecting any arrays that contain null or NaN, I believe that's consistent with ARRAY_CONTAINS, which we have not exposed ARRAY_CONTAINS_NULL or ARRAY_CONTAINS_NAN filters in the backend (so you can't query for an array having null or nan today).

firebase-firestore/src/main/java/com/google/firebase/firestore/core/RelationFilter.java
```diff
@@ -61,8 +62,10 @@ public boolean matches(Document doc) {
       hardAssert(
           refValue instanceof DocumentKey, "Comparing on key, but filter value not a DocumentKey");
       hardAssert(
-          operator != Operator.ARRAY_CONTAINS,
-          "ARRAY_CONTAINS queries don't make sense on document keys.");
+          operator != Operator.ARRAY_CONTAINS
+              && operator != Operator.ARRAY_CONTAINS_ANY
+              && operator != Operator.IN,
+          "'" + operator.toString() + "' queries don't make sense on document keys.");
```
rockwotj: IN makes sense on document keys? Or are we explicitly disallowing batch gets via this API?

/cc @wilhuff @mikelehen

mikelehen: Oh!  Thanks for flagging...  For some reason I just assumed it wouldn't work.  But I guess you can create indexes on document key just like any field in the document?  If so, then we should probably relax this.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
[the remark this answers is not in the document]
mikelehen: Oh!  Yes, good call @rockwotj.  We should reject null and NaN inside the array.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/RelationFilter.java
rockwotj: Well IN on key basically becomes an inequality, so the same restrictions apply as key equality (which is also treated as an inequality)

thebrianchen: @rockwotj  Are document-keys allowed for ARRAY_CONTAINS as well, or is it just allowed for IN and ARRAY_CONTAINS_ANY?

rockwotj: Just `IN`, no `ARRAY_CONTAINS` or `ARRAY_CONTAINS_ANY` (the key can't be an array).

This is a good point, I should add tests to make sure people aren't doing __key__ ARRAY_CONTAINS "foo" queries.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -357,6 +458,30 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
+                  + op.toString()
+                  + "' filters support a maximum of 10 elements in the value array.");
+        }
+        if (((List) value).contains(null)) {
+          throw new IllegalArgumentException(
+                  "Invalid Query. '"
+                          + op.toString()
+                          + "' filters cannot contain 'null' in the value array.");
+        }
+        if (((List) value).contains(Double.NaN) || ((List) value).contains(Float.NaN)) {
```
rockwotj: Does this really work? I thought that contains uses `equals` and `NaN` never equals itself?

thebrianchen: This seems to work based on the unit tests: https://github.com/firebase/firebase-android-sdk/pull/519/files#diff-37f4386134136e6f82e850b93ff16f8fR659

I verified that removing either `Double.NaN` or `Float.NaN` from the `if` will cause the corresponding test to fail.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -301,6 +331,76 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
...
+    return whereHelper(FieldPath.fromDotSeparatedPath(field), Operator.ARRAY_CONTAINS_ANY, value);
+  }
+
+  /**
+   * Creates and returns a new Query with the additional filter that documents must contain the
+   * specified field, the value must be an array, and that the array must contain at least one value
+   * from the provided array.
+   *
+   * <p>A Query can have only one whereArrayContainsAny() filter and it cannot be combined with
+   * whereArrayContains() or whereIn()..
```
rockwotj: extra `.`

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
rockwotj: That seems surprising, does this work with other kinds of NaN?

Here are some sample values:

```

  private static final Double NEGATIVE_NAN = Double.longBitsToDouble(0xFFF1100000000000L);
  private static final Double SIGNALLING_NAN = Double.longBitsToDouble(0x7FF1000000000000L);
  private static final Double NEGATIVE_SIGNALLING_NAN =
      Double.longBitsToDouble(0xFFF1000000000000L);
  private static final Double OTHER_NAN = Double.longBitsToDouble(0x7FF1000000000001L);

  private static final Double NEGATIVE_OTHER_NAN = Double.longBitsToDouble(0xFFF1000000000001L);
```

Maybe this is overkill tho? I don't know the JVM spec exactly for NaNs and if they are normalized or not.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
```diff
@@ -685,6 +701,12 @@ public void queriesFilteredByDocumentIDMustUseStringsOrDocumentReferences() {
             + "a valid String or DocumentReference, but it was of type: java.lang.Integer";
     expectError(() -> collection.whereGreaterThanOrEqualTo(FieldPath.documentId(), 1), reason);
 
+    reason =
+        "Invalid query. When querying with FieldPath.documentId() you must provide "
+            + "a valid String or DocumentReference, but it was of type: java.util.Arrays$ArrayList";
+    ;
```
mikelehen: errant `;`

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -365,11 +369,12 @@ public Query whereArrayContainsAny(@NonNull FieldPath fieldPath, @NonNull List<O
 
   /**
    * Creates and returns a new Query with the additional filter that documents must contain the
-   * specified field and that the field must equal any values of the provided array.
+   * specified field and the value must equal at least one value from the provided array.
```
mikelehen: "at least one value from" is odd phrasing for whereIn, since the only way it would match multiple values is if your array contains duplicate values, which would probably not be an intentional thing to do...  So I would just say "the value must equal one of the values from ..."

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -365,11 +369,12 @@ public Query whereArrayContainsAny(@NonNull FieldPath fieldPath, @NonNull List<O
...
    *
-   * <p>A Query can have only one whereIn() filter.
+   * <p>A Query can have only one whereIn() filter it cannot be combined with
+   * whereArrayContainsAny().
    *
-   * @param field The name of the field containing an array to search
+   * @param field The name of the field containing an array to search.
```
mikelehen: not an array

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -365,11 +369,12 @@ public Query whereArrayContainsAny(@NonNull FieldPath fieldPath, @NonNull List<O
 
   /**
    * Creates and returns a new Query with the additional filter that documents must contain the
-   * specified field and that the field must equal any values of the provided array.
+   * specified field and the value must equal at least one value from the provided array.
    *
-   * <p>A Query can have only one whereIn() filter.
+   * <p>A Query can have only one whereIn() filter it cannot be combined with
```
mikelehen: "it cannot" => "and it cannot"

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -381,11 +386,12 @@ public Query whereIn(@NonNull String field, @NonNull List<Object> value) {
...
+   * whereArrayContainsAny().
    *
-   * @param fieldPath The path of the field containing an array to search
+   * @param fieldPath The path of the field containing an array to search.
```
mikelehen: Please apply above comments to this overload too.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -167,11 +167,11 @@ public FieldPath inequalityField() {
     return null;
   }
 
-  @Nullable
   /**
-   * Checks if any of the provided RelationOps are included in the query and returns the first one
-   * that is, or null if none are.
+   * Checks if any of the provided filter operatorsrelation are included in the query and returns
```
mikelehen: operatorsrelation?

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/RemoteSerializerTest.java
```diff
@@ -646,14 +633,49 @@ public void testEncodesMultipleFiltersOnDeeperCollections() {
...
+    Target expected =
+        Target.newBuilder()
+            .setQuery(queryBuilder)
+            .setTargetId(1)
+            .setResumeToken(ByteString.EMPTY)
+            .build();
+
+    assertEquals(expected, actual);
+    assertEquals(serializer.decodeQueryTarget(serializer.encodeQueryTarget(q)), q);
+  }
```
mikelehen: Any way we can make these tests less boilerplate-y?  Ideally it'd be much closer to the TypeScript one (https://github.com/firebase/firebase-js-sdk/blob/7a52f753c94d326e690ccafed48d82c8959dcf72/packages/firestore/test/unit/remote/node/serializer.test.ts#L786)

Basically, we should try to refactor so that we can test filter serialization in isolation (without a query, target, etc.).  Perhaps we should make `encodeRelationFilter()` public (or package-private?) so that we can test it directly?  Failing that perhaps we could have a test helper method that takes a filter and wraps it up in a dummy query / target in order to serialize / deserialize it.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/RemoteSerializerTest.java
```diff
@@ -710,11 +706,10 @@ public void testEncodesDisjunctiveFiltersOnDeeperCollections() {
...
 
-  // PORTING NOTE: Isolated array-contains, array-contains-any, and in filter test omitted since we
-  // seem to have omitted isolated filter tests on Android (and the encodeRelationFilter() /
-  // decodeRelationFilter() serializer methods are private) in favor of relying on the larger tests.
-  // array-contains, array-contains-any, and in encoding / decoding is covered by
-  // testEncodesMultipleFiltersOnDeeperCollections().
+  // PORTING NOTE: Isolated array-contains filter test omitted since we seem to have omitted
+  // isolated filter tests on Android (and the encodeRelationFilter() / decodeRelationFilter()
+  // serializer methods are private) in favor of relying on the larger tests. array-contains
+  // encoding / decoding is covered by testEncodesMultipleFiltersOnDeeperCollections().
```
mikelehen: FWIW- In the fullness of time, we should try to bring these tests into parity with TypeScript / iOS which have isolated tests for each filter...  So once you make it convenient to have isolated filter tests perhaps change this to:

> TODO(PORTING NOTE): Android currently tests most filter serialization (for equals, greater than, array-contains, etc.) only in testEncodesMultipleFiltersOnDeeperCollections and lacks isolated filter tests like the other platforms have. We should fix this.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
thebrianchen: Gil will probably update this when he ports over https://github.com/firebase/firebase-js-sdk/pull/1894/files

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -301,6 +331,76 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
...
+   * <p>A Query can have only one whereArrayContainsAny() filter and it cannot be combined with
+   * whereArrayContains() or whereIn().
+   *
+   * @param field The name of the field containing an array to search.
+   * @param value The array that contains the values to match.
+   * @return The created Query.
+   */
+  @NonNull
+  @PublicApi
+  public Query whereArrayContainsAny(@NonNull String field, @NonNull List<Object> value) {
```
mikelehen: We need to remove `@PublicApi` and `public` before checking this in (and add a TODO(in-queries) to expose).

PR #522, opened 2019-06-14
Set test type to release only in CI.

vkryachko:
This fixes Android Studio issue, where it is impossible to run
integration tests in debug mode.

Additionally move build type configuration to FirebaseLibraryPlugin to
avoid projects.all configuration in gradle.

### DIFF
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
@@ -20,7 +20,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.firebase.gradle.plugins.ci.device.FirebaseTestServer;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -32,6 +31,25 @@ public void apply(Project project) {
         project.getExtensions().create("firebaseLibrary", FirebaseLibraryExtension.class, project);
 
     LibraryExtension android = project.getExtensions().getByType(LibraryExtension.class);
+    android.buildTypes(
+        types ->
+            types
```
wilhuff: You've lost the comment here about why you're copying the signing configuration from the debug to release build type. Assuming this is still accurate, maybe add this back above `android.buildTypes`?

```
// In the case of and android library signing config only affects instrumentation test APK.	
// We need it signed with default debug credentials in order for FTL to accept the APK.
```

vkryachko: nice catch, added it back

PR #523, opened 2019-06-14
fix functions

VinayGuthal:


firebase-functions/firebase-functions.gradle
```diff
@@ -29,7 +29,7 @@ android {
     compileSdkVersion project.targetSdkVersion
     defaultConfig {
         targetSdkVersion project.targetSdkVersion
-        minSdkVersion project.minSdkVersion
+        minSdkVersion 16
```
vkryachko: Isn't it a breaking change? Do we want to make it?

VinayGuthal: this was supposed to be done at io.. But this didnt go through. https://github.com/FirebasePrivate/firebase-android-sdk/pull/303 for more details

PR #524, opened 2019-06-14
[Firebase Segmentation] Add custom installation id cache layer and tests for it.

diwu-arete:


firebase-segmentation/firebase-segmentation.gradle
```diff
@@ -57,7 +57,7 @@ android {
     compileSdkVersion project.targetSdkVersion
 
     defaultConfig {
-        minSdkVersion project.minSdkVersion
+        minSdkVersion 21
```
vkryachko: All firebase SDKs require at most API level 16, while some 14. Was it a conscious choice to require 21?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,132 @@
...
+          "%s = ? " + "AND " + "%s = ?", GMP_APP_ID_COLUMN_NAME, FIREBASE_APP_NAME_COLUMN_NAME);
+
+  private final SQLiteDatabase localDb;
+
+  CustomInstallationIdCache() {
+    // Since different FirebaseApp in the same Android application should have the same application
+    // context and same dir path, so that use the context of the default FirebaseApp to create/open
+    // the database.
+    localDb =
+        SQLiteDatabase.openOrCreateDatabase(
```
vkryachko: I'd recommend using SqliteOpenHelper as it helps manage the connections, db creation and schema evolution over time. For examples of using it see firestore and transport-runtime in this repo

firebase-segmentation/firebase-segmentation.gradle
diwu-arete: While switch to use SQLiteOpenHelper, we don't need to choose this directory any more.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
diwu-arete: Done

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,208 @@
...
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+import android.database.sqlite.SQLiteOpenHelper;
+import android.os.Build;
+import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
+import com.google.android.gms.common.internal.Preconditions;
+import com.google.firebase.FirebaseApp;
+
+class CustomInstallationIdCache {
```
vkryachko: What are the thread-safety requirements for this class? Will it be used from multiple threads?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,208 @@
...
+    @Override
+    public void onOpen(SQLiteDatabase db) {
+      ensureConfigured(db);
+    }
+  }
+
+  private final OpenHelper openHelper;
+
+  CustomInstallationIdCache() {
+    // Since different FirebaseApp in the same Android application should
```
vkryachko: Why not pass it in as a constructor argument? it should make the class easier to test

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
diwu-arete: I just make methods in this class as synchronized methods, for now.
Probably we could lock at smaller granularity in the next cls when we can see clearer logic on the layer on top of this.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
diwu-arete: Switching to use a shared preference.

I don't think we need to test against the SharePreferences itself, so keep not passing as a argument,.

diwu-arete:
> Actually, taking a step back: it looks like your persistence needs are pretty simple at the moment: e.g. it's one value per firebase App. That said, do you actually need sqlite? alternatively can you just use shared preferences for persistence?
> 
> Related: do you ever plan to support direct boot mode? in which case you might want to persist state in a direct-boot aware way.

As talked with Vlad offline, change to use shared preference.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final String SHARED_PREFS_NAME = "CustomInstallationIdCache";
+
+  private static final String CUSTOM_INSTALLATION_ID_KEY = "Cid";
+  private static final String INSTANCE_ID_KEY = "Iid";
+  private static final String CACHE_STATUS_KEY = "Status";
+
+  private static CustomInstallationIdCache singleton = null;
+  private final SharedPreferences prefs;
+
+  static CustomInstallationIdCache getInstance() {
```
vkryachko: Due to the way SharedPrefs are implemented, this class does not have to be a singleton, as multiple instances of sharedPrefs that point to the same file share the same thread-safe in-memory backing structure. As a side-effect of doing it your interface will become much simpler, as you won't have to pass FirebaseApp around to be able to read/write to the cache, e.g. something along these lines:

```java
class Cache {
  // note: injecting direct deps only https://github.com/google/guice/wiki/InjectOnlyDirectDependencies
  Cache(String persistenceKey, SharedPreferences prefs) {}
  Value read();
  void write(Value value);
}
```

When your SDK initializes it can then construct this cache as `new Cache(app.getPersistenceKey(), context.getSharedPreferences());` Note that Cache is completely decoupled from FirebaseApp, so there is not need to initialize FirebaseApp during tests, which leads to faster more isolated tests. wdyt?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,109 @@
...
+    SharedPreferences.Editor editor = prefs.edit();
+    editor.putString(
+        getSharedPreferencesKey(firebaseApp, CUSTOM_INSTALLATION_ID_KEY),
+        entryValue.getCustomInstallationId());
+    editor.putString(
+        getSharedPreferencesKey(firebaseApp, INSTANCE_ID_KEY), entryValue.getFirebaseInstanceId());
+    editor.putInt(
+        getSharedPreferencesKey(firebaseApp, CACHE_STATUS_KEY),
+        entryValue.getCacheStatus().ordinal());
+    editor.commit();
```
vkryachko: Will this code be called from the UI thread? If so, this will do blocking disk IO, which is generally discouraged. A few options to deal with this:
1. change `commit() -> apply()`, this will apply the change to disk asynchronously(no way to know if it failed)
2. make this method async e.g. returning `Task<Boolean>` and have a backing thread poll for IO operations

wdyt?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  synchronized void clear(FirebaseApp firebaseApp) {
+    SharedPreferences.Editor editor = prefs.edit();
+    editor.remove(getSharedPreferencesKey(firebaseApp, CUSTOM_INSTALLATION_ID_KEY));
+    editor.remove(getSharedPreferencesKey(firebaseApp, INSTANCE_ID_KEY));
+    editor.remove(getSharedPreferencesKey(firebaseApp, CACHE_STATUS_KEY));
```
vkryachko: forgot to `commit()/apply()`?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private CustomInstallationIdCache() {
+    // Since different FirebaseApp in the same Android application should have the same application
+    // context and same dir path, so that use the context of the default FirebaseApp to create the
+    // shared preferences.
+    prefs =
+        FirebaseApp.getInstance()
+            .getApplicationContext()
+            .getSharedPreferences(SHARED_PREFS_NAME, 0); // private mode
```
vkryachko: nit: instead of the comment pls use `Context.MODE_PRIVATE`

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
diwu-arete: I agree with the multi cache instance idea.

But for the constructor arguments, I still prefer Cache(FirebaseApp app),
I don't want the caller to pass in a random string other than a app.getPersistenceKey().
Also the shared pref, I want to hide the shared pref name inside of the cache class.

Initializing FirebaseApp in tests doesn't hurt, which is fine, I think

diwu-arete: 1. For passing in a String argument and SharedPreference argument from the caller, I don't like the idea, because I don't want the caller to pass in a random string or a random shared pref name.

2. For singleton or non-singleton, I don't have a strong opinion. But the singleton mode can centralize management of all FirebaseApps in one shared pref file, which allows to clearAll() for example, and it doesn't hurt the test too much.

vkryachko: I don't feel strongly re 1.

2. you can centralize all FirebaseApps in one shared pref file even if you don't use a singleton, as calling `getSharedPreferences()` with the same name+mode returns the same instance, clearAll can also be implemented as a static method that takes a `Context`. It's ultimately up to your team to decide the best approach here, but I feel like singleton does not provide many benefits due to the nature of how sharedPrefs work. If you do decide to use a singleton, pls make `getInstance()` thread-safe.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
```diff
@@ -0,0 +1,130 @@
...
+
+  synchronized Task<Boolean> clear(FirebaseApp firebaseApp) {
+    SharedPreferences.Editor editor = prefs.edit();
+    editor.remove(getSharedPreferencesKey(firebaseApp, CUSTOM_INSTALLATION_ID_KEY));
+    editor.remove(getSharedPreferencesKey(firebaseApp, INSTANCE_ID_KEY));
+    editor.remove(getSharedPreferencesKey(firebaseApp, CACHE_STATUS_KEY));
+    return commitSharedPreferencesEditAsync(editor);
+  }
+
+  @VisibleForTesting
```
vkryachko: Optionally consider using `@RestrictTo(RestrictTo.Scope.Tests)`, as it will also warn you if you try to call it from within this class.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/CustomInstallationIdCache.java
diwu-arete: Changing to thread-safe singleton.

If we find singleton pattern doesn't work well later, we can always change it.

PR #541, opened 2019-06-19
Simplify FieldValueOptions

schmidt-sebastian:
A code cleanup that removes code. 

This CL ports some of the recent FieldValue changes in Android and moves the special handling of some of the FieldValue classes into DocumentSnapshot. This is purely a refactor.


firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
```diff
@@ -589,8 +601,25 @@ private Object convertValue(FieldValue value, FieldValueOptions options) {
             database.getDatabaseId());
       }
       return new DocumentReference(key, firestore);
+    } else if (value instanceof TimestampValue) {
+      Timestamp timestamp = ((TimestampValue) value).value();
```
wilhuff: Rather than this one giant method, it would be nice to factor these out into little, easy to understand private helpers (as was done in iOS).

schmidt-sebastian:
Hm. `testServerTimestampsCanRetainPreviousValueThroughConsecutiveUpdates()` is still broken. Looking.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
schmidt-sebastian: Helpers added.

PR #545, opened 2019-06-19
Add the state machine of updating custom installation id in the local cache and update to Firebase Segmentation backend. 

diwu-arete:
Close review needed for FirebaseSegmentation.java

CL also contains unit tests. 
The http client is not implemented yet.


diwu-arete:
This is a pretty large PR, but the key part is just FirebaseSegmentation.java which needs a close review.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
```diff
@@ -55,7 +79,161 @@ public static FirebaseSegmentation getInstance(@NonNull FirebaseApp app) {
     return app.get(FirebaseSegmentation.class);
   }
 
-  Task<Void> setCustomInstallationId(String customInstallationId) {
-    return Tasks.forResult(null);
+  Task<Void> setCustomInstallationId(@Nullable String customInstallationId) {
```
vkryachko: should the return type be `@NonNull public Task<Void>` instead?

vkryachko: Also what are the thread-safety/concurrency requirements for this method(including the ones it calls)? I looks like multiple calls to this from different threads could lead to races

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
```diff
@@ -55,7 +79,161 @@ public static FirebaseSegmentation getInstance(@NonNull FirebaseApp app) {
...
+                        CustomInstallationIdCache.CacheStatus.PENDING_UPDATE)));
+
+    // Start requesting backend when first cache update is done.
+    Task<Code> backendRequestResultTask =
+        firstUpdateCacheResultTask.onSuccessTask(
+            firstUpdateCacheResult -> {
+              if (firstUpdateCacheResult) {
+                String iid = instanceIdResultTask.getResult().getId();
+                String iidToken = instanceIdResultTask.getResult().getToken();
+                return backendServiceClient.updateCustomInstallationId(
```
vkryachko: What if the device is offline, do you want to hold on to the new value and sync when device gets connected or do you want to throw?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
diwu-arete: 1)  Yes, definitely it should be nonnull public

2)  Change it to be synchronized

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
diwu-arete: I have thought this case.

If the device is offline, I would
1) the backendServiceClient return a error
2) the cache status will remain in PENDING_UPDATE or PENDING_CLEAR
3) have a separate listener that retry the PENDINGs when device is back online.

WDYT?

vkryachko: Sounds good.

PR #554, opened 2019-06-21
Disable FTL for AB Testing

allisonbm92:
AB Testing does not have device tests.

### DIFF
--- a/firebase-abt/firebase-abt.gradle
+++ b/firebase-abt/firebase-abt.gradle
@@ -17,7 +17,7 @@ plugins {
 }
 
 firebaseLibrary {
-    testLab.enabled = true
+    testLab.enabled = false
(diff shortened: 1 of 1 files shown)


firebase-abt/firebase-abt.gradle
```diff
@@ -17,7 +17,7 @@ plugins {
 }
 
 firebaseLibrary {
-    testLab.enabled = true
+    testLab.enabled = false
```
vkryachko: It's false by default, so should be able to just remove this line, up to you

allisonbm92: I think defaulting to true might be better behavior. This way, it's explicit that the tests are not running (in case they add some in the future).

PR #559, opened 2019-06-21
Performance: Decode documents in background thread

schmidt-sebastian:
This PR reduces the time it takes to execute `getAllDocumentsMatchingQuery` by moving the Proto serialization into Android's background queue.

Note that the only real complication here is that Android's [THREAD_POOL_EXECUTOR](https://developer.android.com/reference/android/os/AsyncTask#THREAD_POOL_EXECUTOR) uses a LinkedBlockingQueue with a fixed size of 120. We can't enqueue more than 120 elements without triggering RejectedExecutionsExceptions. Instead, this PR uses its own queue and up to 4 runners (the core pool size of the THREAD_POOL_EXECUTOR).

On a Nexus 5X, reading 1700 documents from a single collection takes:
- without PR: 3621ms,  3600ms,  3637ms,  3550ms,  3593ms (avg 3605 ms)
- with this PR: 1498ms, 1438ms, 1502ms, 1496ms, 1433ms (avg 1473 ms)

Reading a collection with 10 document 100 times:
- without PR: 2295 ms, 2280 ms, 2319 ms, 2296ms, 2315ms (avg 2301 ms)
- with PR:  1380ms, 1646 ms, 1503 ms, 1191ms, 1451 ms (avg 1416 ms)

Reading a collection with 2 document 1000 times:
- without PR: 8902ms, 8520ms, 7472ms, 7953ms, 10907ms (avg 8751 ms)
- with PR:  7901ms, 8458ms, 9109ms, 8730ms, 778ms1 (avg 8395 ms)

Reading a collection with 1 document 1000 times:
- without PR: 780ms,  763ms,  724ms,  774ms,  727ms (avg 753 ms)
- with PR:  796ms, 752ms,  724ms,  786ms,  722ms (avg 756 ms)

Note that without the special handling of one document collections, the read time for a 1 document collection roughly doubled (from .7 ms to 1.5 ms).


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
```diff
@@ -136,20 +143,38 @@ public MaybeDocument get(DocumentKey documentKey) {
...
-                return;
-              }
-
-              Document doc = (Document) maybeDoc;
-              if (!query.matches(doc)) {
-                return;
-              }
-
-              results.put(doc.getKey(), doc);
+              byte[] rawContents = row.getBlob(1);
```
wilhuff: This one method is now huge, tightly intermingling threading implementation details with the query processing. Most of this could be generic. There are other circumstances where we potentially process large numbers of rows (e.g. the mutation queue) and this could be extended to those places. Probably even more importantly, if we pull this out into something separate we could test it directly :-).

One thing that's preventing this from being generic is the intermediate results are stored in a concurrent map, which ties the results to the query structure. However, we don't need this to be a map of key to document until we're assembling the result. We could just as easily publish to each document into a dequeue.

Additional notes:
  * This implementation leaves time on the table by not starting query matching until all the protos are decoded. The Firestore worker thread that was processing query results can't start assembling the result before the worker threads are done.
  * Also, we're performing query matching serially on the Firestore worker, when we could be performing it in the parallel block.
  * This is a very low number of threads that are all doing significant work per item. We could probably get away with just a simple lock wrapping our output without building an intermediate `ConcurrentHashMap`.

If we were targeting Java 8, I'd suggest that you've essentially reimplemented a subset of `java.util.stream`. There are ports of that to Android API < 24, but that's a lot of code for something that's pretty small.

At a high level this could be phrased as map/reduce, where the map phase is parsing and filtering (first by `instanceof Document`, second by `query.matches`). The reduce phase is the assembly of the `matchingDocuments` map.

I suggest a few refactoring experiments:

  * See if pulling query matching into the worker threads helps
  * See if directly assembling the `ImmutableSortedMap` while holding a lock works
      * If not, see if a regular `Deque` of matching docs while holding a lock is OK
      * Alternatively try a `ConcurrentLinkedQueue`
  * Then try to extract a component that does the parallel processing in a way that's more self contained

WDYT?

schmidt-sebastian: I'll try to make this more generic, but I didn't want this to turn into a 500 line change with multiple new classes. I'll give it another go though.

Re: Query processing. Query matching is not idempotent (because of https://github.com/firebase/firebase-android-sdk/blob/745272a1e939448ca2f92934af2aa73ffb8c427b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java#L54) and If we do want to perform this concurrently, then I would rather tackle this in a follow-up PR.

wilhuff: Does this turn into a problem in practice? What if we access the orderBy in this thread before enqueuing any work?

schmidt-sebastian: @mikelehen mostly convince me that Query matching is not a problem in practice. I also believe that Query matching itself is not a huge bottleneck, but it doesn't hurt to also execute it in the background, especially given that we are already running a task there.

I rewrote the code in this function to use the Tasks API. By that I mean I added APIs to Util that mirror their equivalent methods of the Task API, but with some significant differences:

- My version of `await` can be called on the main thread to support unit testing of the RemoteDocumentStore.
- My version of `whenAllComplete` doesn't execute the continuation on the main queue, which blocks our SpecTests on Robolectric.

If you like this approach, I can write dedicated tests for these helpers. We might also want to move them to a dedicated place (such as TaskUtil).

schmidt-sebastian:
Taking this one back for now, since the Tasks change likely makes single document fetches much slower. It always queues an execution on a background thread.

schmidt-sebastian:
I updated the PR to add a new queue for background tasks, which doesn't rely on Android's Task API to do its management (since that would move all Task management executed to a background thread). The queue is not super general-purpose, but it works for the use case in this PR and should also work if we choose to do something similar in the mutation queue.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Executors.java
```diff
@@ -29,6 +40,39 @@
...
   public static final Executor DIRECT_EXECUTOR = Runnable::run;
 
+  /**
+   * An executor that runs tasks in parallel on Android's AsyncTask.THREAD_POOL_EXECUTOR.
+   *
+   * <p>Unlike the main THREAD_POOL_EXECUTOR, this executor manages its own queue of tasks and can
+   * handle an unbounded number of pending tasks.
+   */
+  public static final Executor BACKGROUND_EXECUTOR =
+      new Executor() {
```
wilhuff: Please don't make major components we're going to reason about like this anonymous classes. Make it a `private static final class` with a name. This will make stack traces much, much saner.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Executors.java
```diff
@@ -29,6 +40,39 @@
   /** An executor that executes the provided runnable immediately on the current thread. */
   public static final Executor DIRECT_EXECUTOR = Runnable::run;
 
+  /**
+   * An executor that runs tasks in parallel on Android's AsyncTask.THREAD_POOL_EXECUTOR.
+   *
+   * <p>Unlike the main THREAD_POOL_EXECUTOR, this executor manages its own queue of tasks and can
+   * handle an unbounded number of pending tasks.
```
wilhuff: Having an unbounded number of pending tasks seems problematic from a memory usage point of view.

Previously we'd unpack one blob, parse it, compare it, and then allow it to be GCed.

Now, if we get significantly ahead of the background threads' ability to clear this queue each queued document will be resident. Previously, ignoring constant factors around the current document we're looking at, our memory usage was essentially O(matching documents), but this increases it to potentially be O(all documents in a collection). This could be a huge difference. We should block admission into the queue if the background threads aren't cleaning up.

A straightforward way to bound this is to just use a semaphore with a fixed number of permits to guard admission. This will have the side effect of fixing the thread-safety issue you've documented below, which makes me uncomfortable.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Executors.java
```diff
@@ -29,6 +40,39 @@
...
+        public void execute(Runnable command) {
+          pendingTasks.add(command);
+
+          if (activeRunnerCount.get() < MAXIMUM_CONCURRENT_BACKGROUND_TASKS) {
+            // Note that the runner count could temporarily exceed
+            // MAXIMUM_CONCURRENT_BACKGROUND_TASKS if this is code path was executed in parallel.
+            // While undesired, this would merely queue another task on THREAD_POOL_EXECUTOR,
+            // and we are unlikely to hit the 120 pending task limit.
+            activeRunnerCount.incrementAndGet();
+            AsyncTask.THREAD_POOL_EXECUTOR.execute(
```
wilhuff: We still need to handle the fact that `AsyncTask.THREAD_POOL_EXECUTOR.execute` can reject adding this task.

We should implement a caller runs policy, rather than trying to block on the executor becoming available. i.e.

```java
// acquire permit
Runnable wrappedCommand = // thing that runs and then releases semaphore
try {
  AsyncTask.THREAD_POOL_EXECUTOR.execute(wrappedCommand);
} catch (RejectedExecutionException ignored) {
  wrappedCommand.run();
} 
```

firebase-firestore/src/main/java/com/google/firebase/firestore/util/TaskQueue.java
```diff
@@ -0,0 +1,120 @@
...
+    acceptsTasks = false;
+
+    List<T> allResults = new ArrayList<>(totalTaskCount);
+    try {
+      completedTasks.acquire(totalTaskCount);
+      while (!taskResults.isEmpty()) {
+        TaskResult<T> currentResult = taskResults.take();
+
+        if (currentResult.exception != null) {
+          throw new ExecutionException("Unhandled exception in task", currentResult.exception);
```
wilhuff: Assuming task result ordering doesn't matter, since your policy here is just to throw on the first exception we encountered you could save all this translation code (and the TaskResult type declaration even) if you made the completedTasks buffer just contain `T` (and then stored the first exception separately).

firebase-firestore/src/main/java/com/google/firebase/firestore/util/TaskQueue.java
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+          allResults.add(currentResult.result);
+        }
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
wilhuff: I worry slightly that this will return partial `allResults` if this method is interrupted. If the caller performs no further I/O or blocking operations, they may not see another exception and may proceed as if the contents of `allResults` are all the actual results. It seems risky to rely on something else to observe the interrupted status to conclude that the results are actually incorrect.

go/java-practices/interruptedexception recommends throwing an alternative exception (after re-interrupting). Should we just propagate InterruptedException here and let the caller figure it out? Alternatively, should we throw the appropriate FirebaseFirestoreException already here? It's going to be hard to describe what we were trying to do, so it seems like propagating is like the right thing.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/TaskQueue.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private TaskResult(int numTask, Exception exception) {
+      this.numTask = numTask;
+      this.result = null;
+      this.exception = exception;
+    }
+
+    @Override
+    public int compareTo(TaskResult<T> other) {
```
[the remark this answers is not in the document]
schmidt-sebastian: I feel this class is a lot more useful if it returns an ordered set of results. It’s true that this is not needed here - but if we simplify it to what’s actual needed we just end up with a task counter and a semaphore (we also never throw checked exceptions). We might also not want to store the intermediate map that often mostly contains null values. 
All that brings me back to the original version of this PR, which was simpler and a bit more explicit about what it did at the cost of increasing method size.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/TaskQueue.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.Semaphore;
+
+/**
+ * A queue for parallel execution of independent tasks.
```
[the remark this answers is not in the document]
schmidt-sebastian: I'll defer addressing this comment in case we end up removing this class again.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/TaskQueue.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * <p>All tasks must return the same result type T and results can be be fetched via
+ * `awaitResults()`. It is an error to enqueue tasks after invoking `awaitResults()`.
+ *
+ * <p>This class is not thread-safe. All public methods must be called from the same thread.
+ */
+public class TaskQueue<T> {
```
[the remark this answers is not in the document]
schmidt-sebastian: I'll defer addressing this comment in case we end up removing this class again.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/TaskQueue.java
schmidt-sebastian: The page you linked to mostly deals with cancelled RPCs, in which case an Interrupt might be an expected event that needs to be treated explicitly. If we follow this advice, then awaitResults needs to re-throw some checked exception (either InterruptedException or FirebaseFirestoreException). This places all the burden on the callsite and again makes me question the usefulness of this class. 

If we do see an InterruptedException here, we can probably just assume that the client will no longer be in a useable state after (`acquire()` and take()` always run on our AsyncQueue). I'm trying to convince myself that throwing an AssertionFailure is the right thing to do here.

wilhuff:
`ThrottledForwardingExecutor` LGTM. Please do add comments and tests :-). I think it fully encapsulates the nastiness of submitting into Android's thread pool.

Regarding `TaskQueue`, I'm torn. I think there is value in trying to separate the concerns of parallel processing/result accumulation, but I can also see your point that there isn't much to this class, especially if we push responsibility for Interruption into the caller.

schmidt-sebastian:
Updated the PR based on offline discussion:

- Added tests/comments for ThrottledForwardingExecutor.
- Made the "TaskQueue" much less generic and only capable of handling the needs of SQLiteRemoteDocumentCache.

schmidt-sebastian:
> Updated the PR based on offline discussion:
> 
> * Added tests/comments for ThrottledForwardingExecutor.
> * Made the "TaskQueue" much less generic and only capable of handling the needs of SQLiteRemoteDocumentCache.

@wilhuff When you have time, can you take another look?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -331,6 +334,35 @@ public void onOpen(SQLiteDatabase db) {
...
   }
 
+  /**
+   * A simple queue that executes tasks in parallel on the Android's AsyncTask.THREAD_POOL_EXECUTOR
+   * and supports blocking on their completion.
+   *
+   * <p>This class is not thread-safe. In particular, `execute()` and `drain()` should not be called
+   * from parallel threads.
+   */
+  static class BackgroundQueue implements Executor {
```
wilhuff: nit: This particular class in which you've placed this class is huge. I'd love to see us move things out of here if possible.

Why not promote this to top-level? It's also totally generic, so you could also dump it in `util`, but leaving it `local` seems fine too.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/ThrottledForwardingExecutorTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    CountingExecutor countingExecutor = new CountingExecutor();
+    ThrottledForwardingExecutor throttledExecutor =
+        new ThrottledForwardingExecutor(maximumConcurrency, countingExecutor);
+
+    // Schedule more than `maximumConcurrency` parallel tasks and wait until all scheduling has
+    // finished.
+    int numTasks = maximumConcurrency + 1;
+    CountDownLatch schedulingLatch = new CountDownLatch(1);
+    for (int i = 0; i < numTasks; ++i) {
```
wilhuff: It doesn't seem like there's any difference between `currentTask` and `i`. Maybe just make the loop use `currentTask`?

Similarly, numTasks isn't used except as the loop bound, maybe just use `maximumConcurrency + 1` as the loop condition? (I bet if you shortened it to `maxConcurrency` it might even all still fit in one line.)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
schmidt-sebastian: Moved to a top-level class in Util. For now, it doesn't have tests though.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/ThrottledForwardingExecutorTest.java
schmidt-sebastian: `currentTask` is the "effectively final" version of `i` and used in the inner class created below.

Furthermore, `numTask` is actually used twice - once in the loop and then below in line 60. For that reason, I left it as is.

PR #560, opened 2019-06-21
Allow IN queries with arrays of documentIds

thebrianchen:
Notes for @mikelehen:
- I'm not a huge fan of `validateAndCalculateReferenceValue`, but I'm not sure how else to refactor it without having almost identical code copied.
- The error messages for IN queries are now slightly off. What is the best way to customize messages for IN queries using documentIds? At this point, if we want different error messages, is it better to just have separate code blocks instead of trying to force the logic into a refactored `validateAndCalculateReferenceValue`?

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
@@ -353,7 +353,7 @@ public class Query {
                 + op.toString()
                 + "' queries on FieldPath.documentId().");
       } else if (op == Operator.IN) {
-        validateDisjunctiveOperatorValueArray(value, op);
+        validateDisjunctiveFilterElements(value, op);
         List<FieldValue> referenceList = new ArrayList<>();
         for (Object arrayValue : (List) value) {
           referenceList.add(parseDocumentIdValue(arrayValue));
@@ -364,7 +364,7 @@ public class Query {
       }
     } else {
       if (op == Operator.IN || op == Operator.ARRAY_CONTAINS_ANY) {
-        validateDisjunctiveOperatorValueArray(value, op);
+        validateDisjunctiveFilterElements(value, op);
       }
       fieldValue = firestore.getDataConverter().parseQueryValue(value);
     }
@@ -387,20 +387,20 @@ public class Query {
    */
   private ReferenceValue parseDocumentIdValue(Object documentIdValue) {
     if (documentIdValue instanceof String) {
-      String documentKey = (String) documentIdValue;
-      if (documentKey.isEmpty()) {
+      String documentId = (String) documentIdValue;
+      if (documentId.isEmpty()) {
         throw new IllegalArgumentException(
             "Invalid query. When querying with FieldPath.documentId() you must provide a valid "
                 + "document ID, but it was an empty string.");
       }
-      if (!query.isCollectionGroupQuery() && documentKey.contains("/")) {
+      if (!query.isCollectionGroupQuery() && documentId.contains("/")) {
         throw new IllegalArgumentException(
             "Invalid query. When querying a collection by FieldPath.documentId() you must "
                 + "provide a plain document ID, but '"
-                + documentKey
+                + documentId
                 + "' contains a '/' character.");
       }
-      ResourcePath path = query.getPath().append(ResourcePath.fromString(documentKey));
+      ResourcePath path = query.getPath().append(ResourcePath.fromString(documentId));
       if (!DocumentKey.isDocumentKey(path)) {
         throw new IllegalArgumentException(
             "Invalid query. When querying a collection group by FieldPath.documentId(), the "
@@ -424,7 +424,7 @@ public class Query {
   }
 
   /** Validates that the value passed into a disjunctive filter satisfies all array requirements. */
-  private void validateDisjunctiveOperatorValueArray(Object value, Operator op) {
+  private void validateDisjunctiveFilterElements(Object value, Operator op) {
     if (!(value instanceof List) || ((List) value).size() == 0) {
       throw new IllegalArgumentException(
           "Invalid Query. A non-empty array is required for '" + op.toString() + "' filters.");
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
@@ -39,6 +39,7 @@ import com.google.firebase.firestore.core.ViewSnapshot;
 import com.google.firebase.firestore.model.Document;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.ResourcePath;
+import com.google.firebase.firestore.model.value.ArrayValue;
 import com.google.firebase.firestore.model.value.FieldValue;
 import com.google.firebase.firestore.model.value.ReferenceValue;
 import com.google.firebase.firestore.model.value.ServerTimestampValue;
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -84,6 +84,77 @@ public FirebaseFirestore getFirestore() {
     return firestore;
   }
 
+  /** Validates the given document ID and returns the corresponding FieldValue. */
+  private FieldValue validateAndCalculateReferenceValue(FieldPath fieldPath, Object value) {
```
mikelehen: We usually organize our code "top down" where helper methods live below the point where they're used...  so I'd move these methods below whereHelper()...

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -422,65 +493,19 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
-        ResourcePath path = query.getPath().append(ResourcePath.fromString(documentKey));
-        if (!DocumentKey.isDocumentKey(path)) {
-          throw new IllegalArgumentException(
-              "Invalid query. When querying a collection group by FieldPath.documentId(), the "
-                  + "value provided must result in a valid document path, but '"
-                  + path
-                  + "' is not because it has an odd number of segments ("
-                  + path.length()
-                  + ").");
+      if (op == Operator.IN) {
```
mikelehen: nit: It doesn't actually matter, but I think `else if` would make the code structure slightly cleaner.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -84,6 +84,77 @@ public FirebaseFirestore getFirestore() {
     return firestore;
   }
 
+  /** Validates the given document ID and returns the corresponding FieldValue. */
+  private FieldValue validateAndCalculateReferenceValue(FieldPath fieldPath, Object value) {
+    com.google.firebase.firestore.model.FieldPath internalPath = fieldPath.getInternalPath();
+    hardAssert(
+        internalPath.isKeyField(),
+        "fieldPath must be a key field to calculate the reference value.");
```
mikelehen: So we don't use `fieldPath` for anything other than to assert that `isKeyField()` is true?  In that case, I think we should just drop the parameter.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
mikelehen: I think I would:
1. Rename this to parseDocumentIdValue().  "parse" is a good verb to use when we're both validating and transforming an input.
2. Rename "value" to "documentIdValue"
3. Change the return type to ReferenceValue which is more specific than FieldValue so makes the code more type-safe and self-documenting.
4. Change the comment to something like "Parses the given documentIdValue into a ReferenceValue, throwing appropriate errors if the value is anything other than a DocumentReference or String or if the string is malformed."

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -84,6 +84,77 @@ public FirebaseFirestore getFirestore() {
...
+        throw new IllegalArgumentException(
+            "Invalid query. When querying a collection group by FieldPath.documentId(), the "
+                + "value provided must result in a valid document path, but '"
+                + path
+                + "' is not because it has an odd number of segments ("
+                + path.length()
+                + ").");
+      }
+      fieldValue =
+          ReferenceValue.valueOf(this.getFirestore().getDatabaseId(), DocumentKey.fromPath(path));
```
mikelehen: You can get rid of "fieldValue" and just `return` here.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -84,6 +84,77 @@ public FirebaseFirestore getFirestore() {
...
+    } else if (value instanceof DocumentReference) {
+      DocumentReference ref = (DocumentReference) value;
+      fieldValue = ReferenceValue.valueOf(this.getFirestore().getDatabaseId(), ref.getKey());
```
mikelehen: just `return`

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -422,65 +493,19 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
-              "Invalid query. When querying a collection group by FieldPath.documentId(), the "
-                  + "value provided must result in a valid document path, but '"
-                  + path
-                  + "' is not because it has an odd number of segments ("
-                  + path.length()
-                  + ").");
+      if (op == Operator.IN) {
+        validateDisjunctiveOperatorValueArray(value, op);
+        List referenceList = new ArrayList();
+        for (Object arrayValue : (List) value) {
```
mikelehen: `arrayValue` is potentially confusing since we have an `ArrayValue` type which itself represents an array.  Perhaps rename to `element`?

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -422,65 +493,19 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
-          throw new IllegalArgumentException(
-              "Invalid query. When querying a collection group by FieldPath.documentId(), the "
-                  + "value provided must result in a valid document path, but '"
-                  + path
-                  + "' is not because it has an odd number of segments ("
-                  + path.length()
-                  + ").");
+      if (op == Operator.IN) {
+        validateDisjunctiveOperatorValueArray(value, op);
+        List referenceList = new ArrayList();
```
mikelehen: Prefer generic versions of `List<>` for type-safety, so this would be `List<ReferenceValue>`.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -422,65 +493,19 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
-        if (!DocumentKey.isDocumentKey(path)) {
-          throw new IllegalArgumentException(
-              "Invalid query. When querying a collection group by FieldPath.documentId(), the "
-                  + "value provided must result in a valid document path, but '"
-                  + path
-                  + "' is not because it has an odd number of segments ("
-                  + path.length()
-                  + ").");
+      if (op == Operator.IN) {
+        validateDisjunctiveOperatorValueArray(value, op);
```
mikelehen: Not wild about this name...  Anything I come up with feels like a meaningless smashing together of words though. 
 :-)  WDYT of `validateDisjunctiveFilterElements()`?  Idunno.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -422,65 +493,19 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
+        List referenceList = new ArrayList();
+        for (Object arrayValue : (List) value) {
+          referenceList.add(validateAndCalculateReferenceValue(fieldPath, arrayValue));
         }
-        fieldValue =
-            ReferenceValue.valueOf(this.getFirestore().getDatabaseId(), DocumentKey.fromPath(path));
-      } else if (value instanceof DocumentReference) {
-        DocumentReference ref = (DocumentReference) value;
-        fieldValue = ReferenceValue.valueOf(this.getFirestore().getDatabaseId(), ref.getKey());
+        fieldValue = firestore.getDataConverter().parseQueryValue(referenceList);
```
mikelehen: I'm not sure this is actually necessary / helpful since we've already thoroughly parsed this into model objects (i.e. `ReferenceValue`). So we could probably just do `ArrayValue.fromList(referenceList)`

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
thebrianchen: `ArrayValue.fromList()` requires `List<FieldValue>`, so I'm using that instead.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
mikelehen: You seem to have resolved this without implementing it or commenting on why you did not... :-)

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
mikelehen: You seem to have resolved this without implementing it or commenting on why you did not... :-)

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
mikelehen: FWIW, if we wanted to make this work, we could change `ArrayValue` to wrap a `List<? extends FieldValue>` instead of `List<FieldValue>` (read https://dzone.com/articles/covariance-and-contravariance for context).  Probably fine to leave as-is though.

thebrianchen:
I think I renamed `validateDisjunctiveFilterElements` but then escaped out of it. I still respect the review, for it is sacred.

PR #561, opened 2019-06-21
Performance: Don't deserialize full document Proto for Query execution

schmidt-sebastian:
The PR adds a lazy-deserialization option to Documents. This allows Query filter evaluation without parsing the entire document Proto.

Speedup of a Query that returns 2 documents out of a collection with 1700 documents:

- With master: 3644ms, 3591ms, 3577ms, 3616ms, 3603ms (avg 3606ms)
- With PR: 3012ms, 2881ms, 2894ms, 2929ms, 2888ms (avg 2920ms)

These numbers, as always, are from my trusted Nexus 5X.


schmidt-sebastian:
Note: This pulls in https://github.com/firebase/firebase-js-sdk/pull/1903 to fix the unit tests since we now always assert that no Document proto has a zero-version, regardless of source.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
```diff
@@ -36,59 +40,107 @@
...
+      RemoteSerializer serializer,
+      com.google.firestore.v1.Document proto,
+      DocumentState documentState) {
+    DocumentKey key = serializer.decodeKey(proto.getName());
+    SnapshotVersion version = serializer.decodeVersion(proto.getUpdateTime());
+
+    hardAssert(
+        !version.equals(SnapshotVersion.NONE), "Found a document Proto with no snapshot version");
+
+    return new Document(key, version, documentState) {
```
wilhuff: This kind of anonymous class makes for some very difficult to understand stack traces when they're reported to us.

A better arrangement is to make named private static classes, e.g. `EagerDocument` and `LazyProtoDocument` that we explicitly instantiate.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/SetMutation.java
```diff
@@ -72,7 +72,8 @@ public MaybeDocument applyToRemoteDocument(
     // accepted the mutation so the precondition must have held.
 
     SnapshotVersion version = mutationResult.getVersion();
-    return new Document(getKey(), version, value, Document.DocumentState.COMMITTED_MUTATIONS);
+    return Document.fromObjectValue(
```
wilhuff: The SetMutation `value` field could be similarly constructed from a proto such that this could use the `fromProto` version. It's possible this is something for a follow-up PR.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
```diff
@@ -36,59 +40,107 @@
...
+          // the requested field path. This speeds up Query execution as query filters can discard
+          // documents based on a single field.
+          Value value = proto.getFieldsMap().get(path.getFirstSegment());
+          for (int i = 1; value != null && i < path.length(); ++i) {
+            if (value.getValueTypeCase() != Value.ValueTypeCase.MAP_VALUE) {
+              return null;
+            }
+            value = value.getMapValue().getFieldsMap().get(path.getSegment(i));
+          }
+          return value != null ? serializer.decodeValue(value) : null;
```
wilhuff: Our `OrderBy` implementation does not implement a Schwartzian transform so we'll be performing these decodes for each comparison. For a large collection this could potentially eat into the gains you're observing here. I wonder if a cache of `FieldPath` to `FieldValue` would hurt here?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
```diff
@@ -36,59 +40,107 @@
...
-   * serialization). Might be null.
+   * Creates a new Document from a Proto representation.
+   *
+   * <p>The Proto is only converted to an ObjectValue if the consumer calls `getData()`.
    */
-  private final com.google.firestore.v1.Document proto;
-
-  public @Nullable com.google.firestore.v1.Document getProto() {
-    return proto;
+  public static Document fromProto(
```
wilhuff: If `LazyDocument` were a public class, this could more naturally just be expressed as a method on the `RemoteSerializer`.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
```diff
@@ -36,59 +40,107 @@
...
     return KEY_COMPARATOR;
   }
 
-  private final ObjectValue data;
-
-  private final DocumentState documentState;
+  /** Creates a new Document from an existing ObjectValue. */
+  public static Document fromObjectValue(
+      DocumentKey key, SnapshotVersion version, ObjectValue data, DocumentState documentState) {
+    return new Document(key, version, documentState) {
```
wilhuff: This new type doesn't really seem like it adds much. It's just the degenerate case where the proto has been pre-parsed. It seems like we could leave Document as a single class that just happens to have two constructors one that takes ObjectiValue as before and one that takes no ObjectValue but a proto and a serializer.

schmidt-sebastian: My goal was to not inflate the current Document class by letting it know about the Serializer. I updated the class to now take a Function that deserializes Proto values.

FWIW, in my ideal world the ObjectValue-backed Document would go away - but that is many sprints from now.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
schmidt-sebastian: Inlined these two classes.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
schmidt-sebastian: I added a cache - but this has to be a ConcurrentHashMap since this function can be called from user land. Please do shout if that is not actually an issue :)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/SetMutation.java
schmidt-sebastian: Yeah, I want to address the mutations afterwards.

firebase-firestore/src/test/java/com/google/firebase/firestore/DocumentSnapshotTest.java
```diff
@@ -45,12 +45,13 @@ public void testEquals() {
     assertNotEquals(base, differentData);
     assertNotEquals(base, fromCache);
 
+    // Note: `base` and `differentData` have the same hash code since we no longer take document
```
wilhuff: This can't be right. `equals` and `hashCode` have to be consistent or all kinds of hell breaks loose when you use these in hash sets or as keys in a map. If two objects are equal for equality purposes, the contract for hashCode requires that the hashCodes be equal too.

We can't just ignore `data` in equality. There are a few cases to consider:

  * When version=`SnapshotVersion.MIN` the documents are just made-up, so data must play a roll
  * When `DocumentState` isn't synced, the documents are potentially dirty even if they have a version

However if we have a non-`MIN` version and the `DocumentState` is `SYNCED` then we could ignore the data for equality and hashing purposes. This would preserve what you're after here which is avoiding rehydrating the data in the common case of loading a value from the remote document cache.

What do you think of implementing both equality and hashCode according to the principle above? I think that would give most (if not all) the benefit without compromising the contract of equals/hashCode.

schmidt-sebastian: Hrm. This is a good comment, and I 100% agree with you, but I don't think I broke the contract here in the way that you stated. I removed `data` from hash code (I am still using document key and version). This might cause more hash collisions, but the invariant that two equal objects have the same hash code should remain intact.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
schmidt-sebastian: I think the impact of this caching will be hard to quantify. I thought about adding this initially, but was only won over by your comments earlier. We do re-run getField() excessively during the computation of our Views.

I changed the code to lazy-init the field value cache. I do want to change the comparator to compare on the Protos directly, at which point we can remove the cache again.

firebase-firestore/src/test/java/com/google/firebase/firestore/DocumentSnapshotTest.java
wilhuff: Ha! Sorry about misreading that. I didn't realize that this was about an expected collision.

An alternative phrasing that might be less confusing is:

> Note: the assertions below that hash codes of different values are not equal is not something that can be guaranteed. In particular `base` and `differentData` have a hash collision because we don't use data in the hashCode.

You might also want to call attention to this in an implementation comment in `DocumentSnapshot.hashCode()`. That way someone won't see that the value is unused in the future and then accidentally "fix" it.

schmidt-sebastian: Good idea about the additional comment. I added it to `Document.hashCode()`.

PR #562, opened 2019-06-22
Update documentation to reflect emulator is default for integration tests.

thebrianchen:


firebase-firestore/README.md
```diff
@@ -41,7 +41,34 @@ the source root.
...
 
-### Run on Local Emulator
+By default, integration tests run against the Firestore emulator.
+
+### Setting up the Firestore Emulator
+
+The integration tests require that the Firestore emulator is running on port
+8080, which is default when running it via CLI.
+
+  * Ensure that `gogole-services.json` is properly configured.
```
wilhuff: What does this mean? Also typo.

thebrianchen: I wish to communicate that the Android SDK is hard-coded to look for the port on 8080, and that port 8080 is the default port that the android emulator will start on when running it via CLI.

wilhuff: I should have been more specific with my feedback: what does it mean to "ensure that `google-services.json` is properly configured"? Do I need to do something special? If this is the standard setup maybe just link to https://github.com/firebase/firebase-android-sdk#project-setup ?

Also, if we're running against the emulator, why do we need the `google-services.json` at all?

thebrianchen: Gotcha. I did not realize that the `google-services.json` isn't needed when running against the emulator.

Issue #491, opened 2019-06-05
AssertionError: INTERNAL ASSERTION FAILED: Encoded field value should not be null.


daviduzan:
* Android Studio version: 3.4.1
  * Firebase Component: Firestore
  * Component version: 19.0.2

Writing **FieldValue.increment(1)** together with **FieldValue.serverTimestamp()** to document while being offline results in a crash the second time:
**Caused by: java.lang.AssertionError: INTERNAL ASSERTION FAILED: Encoded field value should not be null.**

#### Steps to reproduce & Relevant Code:

When internet connection is not available:
'''

        FirebaseFirestore db = FirebaseFirestore.getInstance();
        Map<String, Object> map1 = new HashMap<>();

        map1.put("field1", FieldValue.increment(1));
        map1.put("field2", FieldValue.serverTimestamp());
        
        db.collection("temp").document("temp").set(map1, SetOptions.merge());
        db.collection("temp").document("temp").set(map1, SetOptions.merge());
'''

#### Stack trace:
java.lang.RuntimeException: Internal error in Firestore (19.0.2).
        at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$5(com.google.firebase:firebase-firestore@@19.0.2:379)
        at com.google.firebase.firestore.util.AsyncQueue$$Lambda$5.run(Unknown Source:2)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
     Caused by: java.lang.AssertionError: INTERNAL ASSERTION FAILED: Encoded field value should not be null.
        at com.google.firebase.firestore.util.Assert.fail(com.google.firebase:firebase-firestore@@19.0.2:46)
        at com.google.firebase.firestore.util.Assert.hardAssert(com.google.firebase:firebase-firestore@@19.0.2:31)
        at com.google.firebase.firestore.remote.RemoteSerializer.encodeValue(com.google.firebase:firebase-firestore@@19.0.2:263)
[...]

PR #564, opened 2019-06-24
Store the coerced base value in the BaseMutation

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-android-sdk/issues/491

The PR changes what we store in a BaseMutation. With the changes here, every transform operation computes what it needs to store in the BaseMutation. For Numeric increment, this is either an existing numeric value or the IntegerValue.of(0), replacing all invalid base values at time of creation.


firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
```diff
@@ -123,23 +121,28 @@ public MaybeDocument applyToLocalView(
...
-  public boolean isIdempotent() {
     for (FieldTransform transform : fieldTransforms) {
-      if (!transform.isIdempotent()) {
-        return false;
+      FieldValue previousValue = null;
+      if (maybeDoc instanceof Document) {
+        previousValue = ((Document) maybeDoc).getField(transform.getFieldPath());
+      }
+
+      FieldValue value = transform.getOperation().computeBaseValue(previousValue);
```
mikelehen: We have a lot of "value" stuff going on here: extractBaseValue(), baseValue, previousValue, value, computeBaseValue() and it's a bit misleading and hard to keep track of. I think we're probably overloading "baseValue" a bit. You tried to differentiate via verb (compute vs extract), but it might be better to change the noun...

So I'm wondering if we could rename "baseValue" in "extractBaseValue()" to make it a bit clearer that it's a "composite" value to be used as a patch mutation...  extractBaseMutationValue()?  extractBaseValues()?

Idunno. WDYT?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformOperation.java
```diff
@@ -16,21 +16,34 @@
...
    * Computes a final transform result after the transform has been acknowledged by the server,
    * potentially using the server-provided transformResult.
    */
-  FieldValue applyToRemoteDocument(FieldValue previousValue, FieldValue transformResult);
+  FieldValue applyToRemoteDocument(@Nullable FieldValue previousValue, FieldValue transformResult);
 
-  /** Returns whether this field transform is idempotent. */
-  boolean isIdempotent();
+  /**
+   * If applicable, returns the base value to persist with this transform. If a base value is
```
mikelehen: "with this transform" is a tad misleading.  Maybe "for this transform"

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Unreleased
+- [fixed] Fixed an internal assertion that was triggered when a
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/NumericTransformsTest.java
```diff
@@ -158,4 +159,28 @@ public void multipleDoubleIncrements() throws ExecutionException, InterruptedExc
     snap = accumulator.awaitRemoteEvent();
     assertEquals(0.111D, snap.getDouble("sum"), DOUBLE_EPSILON);
   }
+
+  @Test
+  public void serverTimestampAndIncrement() throws ExecutionException, InterruptedException {
+    // This test stacks two pending transforms (a ServerTimestamp and an Increment transform) and
+    // reproduces an the setup that was reported in
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed. Sorry :/

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ArrayTransformOperation.java
```diff
@@ -39,17 +39,24 @@
...
+      @Nullable FieldValue previousValue, FieldValue transformResult) {
     // The server just sends null as the transform result for array operations, so we have to
     // calculate a result the same as we do for local applications.
     return apply(previousValue);
   }
 
+  @Override
+  @Nullable
+  public FieldValue computeBaseValue(@Nullable FieldValue currentValue) {
+    return null;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -113,6 +114,21 @@ public abstract MaybeDocument applyToRemoteDocument(
   public abstract MaybeDocument applyToLocalView(
       @Nullable MaybeDocument maybeDoc, @Nullable MaybeDocument baseDoc, Timestamp localWriteTime);
 
+  /**
+   * If applicable, returns the base value to persist with this mutation. If a base value is
+   * provided, the mutation is always applied to this base value, even if document has already been
+   * updated.
```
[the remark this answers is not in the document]
schmidt-sebastian: Updated the comment to:

```
 /**
 * If applicable, returns the base value to persist with this mutation. If a base value is
   * provided, the mutation is always applied to this base value, even if document has already been
   * updated.
   *
   * <p>The base value is a sparse objects that consists of only the document fields that for which
   * this mutation contains a non-idempotent transformation (e.g. a numeric increment). The
   * provided value guarantees consistent behavior for non-idempotent transforms and allow us to return
   * the same latency-compensated value even if the backend has already applied the mutation. The
   * base value is null for idempotent mutations, as they can be re-played even if the backend has
   * already applied them.
   *
   * @return a base value to store along with the mutation, or null for idempotent mutations.
   */
```

And regarding the bug you found:

In https://github.com/firebase/firebase-android-sdk/pull/564/commits/bc718892bbadb72e4687eb8ed930de6e515e99e1, I also fixed the test scenario you described. It turns out that we need even more state, and `localTransformResults` needs both a currentValue and the previousValue. Sorry :/

Another way to fix this would be to only provide one Base Mutation per document, but then we have to merge ObjectValue returned from `extractBaseValue` into a single PatchMutation.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -113,6 +114,21 @@ public abstract MaybeDocument applyToRemoteDocument(
...
+   *
+   * <p>Base values provide consistent behavior for non-idempotent transforms and allow us to return
```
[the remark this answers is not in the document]
schmidt-sebastian: See above

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -113,6 +114,21 @@ public abstract MaybeDocument applyToRemoteDocument(
...
+   * the same latency-compensated value even if the backend has already applied the mutation. The
+   * base value is null for idempotent mutations, as they can be re-played even if the backend has
+   * already applied them.
+   *
+   * @return a base value to store along the mutation, or null for idempotent mutations.
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
schmidt-sebastian: baseValue => baseObject, previousValue => existingValue, value => coercedValue (?)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformOperation.java
schmidt-sebastian: Good catch.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ServerTimestampOperation.java
```diff
@@ -30,7 +30,10 @@ public static ServerTimestampOperation getInstance() {
   }
 
   @Override
-  public FieldValue applyToLocalView(@Nullable FieldValue previousValue, Timestamp localWriteTime) {
+  public FieldValue applyToLocalView(
+      @Nullable FieldValue currentValue,
+      @Nullable FieldValue previousValue,
+      Timestamp localWriteTime) {
     return new ServerTimestampValue(localWriteTime, previousValue);
```
mikelehen: So this is the only implementation of applyToLocalView() that uses `previousValue` instead of `currentValue`, and I was surprised that it does...  It seems reasonable to expect that if you do:

```
db.batch()
  .update(ref, "foo", "old")
  .update(ref, "foo", FieldValue.serverTimestamp()
  .commit();
```

then the `ServerTimestampBehavior.PREVIOUS` value would be "old" (i.e. `currentValue`) rather than whatever the value was before the batch...

But I see if you change this to `currentValue` then we have a test that breaks...  But I think it breaks for a bad reason (a deeper bug in our code, possibly in the way `UserDataConverter` splits updates into a `PatchMutation` and a `TransformMutation`)

I would be a fan of:
1) Getting rid of `previousValue` and always using `currentValue`.  This will clean up the code and make it much easier to reason about.  FWIW I had a hard time groking `previousValue` and `currentValue` since they're both "previous" values.  More explicit names might be `preBatchValue` and `preMutationValue`...  This might technically be a breaking change for ServerTimestampBehavior.PREVIOUS in the case I demonstrated above, but I think it's probably subtle enough we don't care.
2) Tracking down / fixing whatever the bug that exposes. :-)  [sorry for the extra homework...]

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
mikelehen: Maaaaaaybe.  The current code includes all of the following in different places: existingValue, coercedValue, previousValue, currentValue, baseObject, baseValue, maybeDoc, baseDoc  ...  which I'm pretty certain is more terms than we have concepts.

Can you see if we can get rid of previousValue in applyToLocalView()?  And then I'll take another look at the naming.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -119,12 +119,14 @@ public abstract MaybeDocument applyToLocalView(
    * provided, the mutation is always applied to this base value, even if document has already been
    * updated.
    *
-   * <p>Base values provide consistent behavior for non-idempotent transforms and allow us to return
-   * the same latency-compensated value even if the backend has already applied the mutation. The
-   * base value is null for idempotent mutations, as they can be re-played even if the backend has
+   * <p>The base value is a sparse objects that consists of only the document fields that for which
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ServerTimestampOperation.java
schmidt-sebastian: The problem the currentValue/previousValue "hack" addresses is actually different from what you suspect. A problematic call could be something simple as:

```
docRef.update(map("nested", map("time", FieldValue.serverTimestamp())
```

This turns into two mutations in a single mutation batch:

- A PatchMutation that replaces any existing value of "nested" with `map()`.
- A Transform mutation that sets `nested.time` to a ServerTimestamp.

By the time the Transform is applied, `maybeDoc` no longer contains the previous field value.

I added a comment to the test that explains this.

I do think we can improve our naming here, but I don't feel that re-working this entire setup is time well spent.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ServerTimestampTest.java
```diff
@@ -181,6 +181,12 @@ public void testServerTimestampsCanReturnPreviousValue() {
     DocumentSnapshot previousSnapshot = accumulator.awaitRemoteEvent();
     verifyTimestampsAreResolved(previousSnapshot);
 
+    // The following update includes an update of tge nested map "deep", which updates it to contain
```
mikelehen: tge => the

You might consider configuring your IDE to make typos more prominent. 😛 
image

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ObjectValue.java
```diff
@@ -57,6 +60,31 @@ public int typeOrder() {
     return TYPE_ORDER_OBJECT;
   }
 
+  /** Recursively extracts the FieldPaths that are set in this ObjectValue. */
+  public FieldMask getFieldMask() {
```
wilhuff: Why did the way mutations calculate field masks need to change?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
```diff
@@ -207,7 +211,12 @@ private Document requireDocument(@Nullable MaybeDocument maybeDoc) {
         previousValue = ((Document) baseDoc).getField(fieldTransform.getFieldPath());
       }
 
-      transformResults.add(transform.applyToLocalView(previousValue, localWriteTime));
+      FieldValue currentValue = null;
```
wilhuff: It seems as if we could use currentValue if available or fall back on previousValue if not currentValue is not available. At that point the signatures for the TransformOperations wouldn't have to change.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ObjectValue.java
schmidt-sebastian: Before, we used the mutation's FieldMask to compute the baseMutation, which caused us to over-select fields. With this PR, each mutation returns the precise set of field it cares about in an ObjectValue, but we still need to provide the field mask to the PatchMutation.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
schmidt-sebastian: Done

firebase-firestore/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 - [fixed] Fixed an internal assertion that was triggered when an
   update with a `FieldValue.serverTimestamp()` and an update with a
   `FieldValue.increment()` were pending for the same document (#491).
+- [changed] Improved performance for queries with filters that only return a
+  small subset of the documents in a collection.
```
mikelehen: I'm not sure this is a good description of this fix.

schmidt-sebastian: This is the merge commit...

mikelehen: Whoops... 😂 SORRY!  When I opened the PR, github took me to https://github.com/firebase/firebase-android-sdk/pull/564/files/700c7c155505a388cc4f902c30e46751bbc6607a..bec9bd64fc3c8be10bb90cdf465820a3766844d4 which shows the added changelog as part of your diff...  I thought github was smarter than that. :-/

PR #566, opened 2019-06-24
Make QueryMatching immutable

schmidt-sebastian:
The goal of this PR is to make Query.matches() thread-safe. It does this by creating an immutable view of the Query state that we use for matching and memoizes the Query's orderBy constraint only once and before the Query's execution.

There are many other ways to achieve this, but this one seemed the most straightforward to me and simplifies the code in the Query class a bit. Apart from creating the new QueryMatcher class, this PR mostly moves code to the new class.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
@@ -330,61 +330,17 @@ public final class Query {
     return memoizedOrderBy;
   }
 
-  private boolean matchesPathAndCollectionGroup(Document doc) {
-    ResourcePath docPath = doc.getKey().getPath();
-    if (collectionGroup != null) {
-      // NOTE: this.path is currently always empty since we don't expose Collection
-      // Group queries rooted at a document path yet.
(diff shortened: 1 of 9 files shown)


mikelehen:
Can you provide more context on this?  In what way is Query not immutable today?

schmidt-sebastian:
> Can you provide more context on this? In what way is Query not immutable today?

Some of the query operations, including query matching when Bounds are provided, create a memoized OrderBy list that modifies the existing Query instance.

mikelehen:
Is that memoization not thread-safe?  It should be!

mikelehen:
(worst case, two threads could theoretically hit the memoization at the same time and we accidentally do it twice; but it should still be 100% safe)

schmidt-sebastian:
We could synchronize access to it ("There are many other ways to achieve this").

schmidt-sebastian:
FWIW, @mikelehen (in offline discussion) pointed out that the issue I am trying to solve (making Query "more" thread-safe by avoiding orderBy-memoization) might not be worth addressing and that all usages of the Query class are already thread-safe. While parallel execution might result in duplicate memoization, we could decide to ignore this.

I'm gonna leave this PR open for comment/thoughts.

PR #567, opened 2019-06-24
Run IN query integration tests only when running against emulator

thebrianchen:


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -428,33 +428,37 @@ public void testQueriesCanUseArrayContainsFilters() {
...
   @Test
-  @Ignore
   public void testQueriesCanUseInFilters() {
-    Map<String, Object> docA = map("zip", 98101);
-    Map<String, Object> docB = map("zip", 91102);
-    Map<String, Object> docC = map("zip", 98103);
-    Map<String, Object> docD = map("zip", asList(98101));
-    Map<String, Object> docE = map("zip", asList("98101", map("zip", 98101)));
-    Map<String, Object> docF = map("zip", map("code", 500));
+    if (!isRunningAgainstEmulator()) {
```
schmidt-sebastian: I know you looked into this - but what is your reason for not using `assumeTrue()`?

https://reflectoring.io/conditional-junit4-junit5-tests/

thebrianchen: The reason I had was that `assumeTrue()` wasn't used anywhere else in the codebase, but I guess that's not a valid reason to avoid using it.

PR #573, opened 2019-06-26
Http client in Firebase Segmentation SDK to call backend service.

diwu-arete:
The proto is defined in https://cs.corp.google.com/piper///depot/google3/google/firebase/segmentation/v1alpha/segmentation_service.proto


firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
```diff
@@ -14,43 +14,182 @@
...
-    ALREADY_EXISTS,
+    SERVER_ERROR,
 
-    PERMISSION_DENIED
+    UNAUTHORIZED,
   }
 
   public SegmentationServiceClient() {
     httpClient = new OkHttpClient();
     httpRequestExecutor = Executors.newFixedThreadPool(4);
```
vkryachko: A meta-point about the implementation so far:

The SDK is already using 6 threads in total, even though all it does is it writes to shared prefs and makes http calls.

To make it more efficient consider doing the following:
1. Expose an async public 3p API
2. Have a **synchronous** internal implementation
3. Have public API (FirebaseSegmentation class) have one thread-pool that it uses to call into the synchronous implementation.

As an additional side-effect of doing it, the come will be come much more simple as it won't have Task chains in it, but rather have straightforward sequential code.

firebase-segmentation/firebase-segmentation.gradle
```diff
@@ -65,8 +32,11 @@ android {
     }
     sourceSets {
```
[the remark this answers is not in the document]
diwu-arete: Done.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
diwu-arete: Done, very good suggestion.

diwu-arete:
Changed okhttpclient to httpsUrlConnection

firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
```diff
@@ -14,43 +14,148 @@
...
+                  resourceName,
+                  apiKey));
+
+      HttpsURLConnection httpsURLConnection = (HttpsURLConnection) url.openConnection();
+      httpsURLConnection.setDoOutput(true);
+      httpsURLConnection.setRequestMethod("PATCH");
+      httpsURLConnection.addRequestProperty(
+          "Authorization", "FIREBASE_INSTALLATIONS_AUTH " + firebaseInstanceIdToken);
+      httpsURLConnection.addRequestProperty("Content-Type", "application/json");
+      OutputStream os = httpsURLConnection.getOutputStream();
```
vkryachko: You need to explicitly close the stream with `try-finally` to avoid a resource leak.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
```diff
@@ -14,43 +14,148 @@
...
+      try {
+        os.write(
```
vkryachko: Additionally you should gzip the stream to save bandwidth, here's an example use of httpurlconnection with resource closing and gzip handling:

https://github.com/firebase/firebase-android-sdk/blob/a7761a4e5a5dbf0656baeeb3a97acf71d99332e6/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L220-L257

firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
diwu-arete: Done.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
diwu-arete: Done.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
```diff
@@ -36,12 +39,14 @@
   private final FirebaseInstanceId firebaseInstanceId;
   private final CustomInstallationIdCache localCache;
   private final SegmentationServiceClient backendServiceClient;
+  private final Executor executor;
 
   FirebaseSegmentation(FirebaseApp firebaseApp) {
```
vkryachko: Could this constructor delegate to the one below?

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
```diff
@@ -107,70 +113,88 @@ public static FirebaseSegmentation getInstance(@NonNull FirebaseApp app) {
...
-    if (cacheEntryValue != null
-        && cacheEntryValue.getCustomInstallationId().equals(customInstallationId)
-        && cacheEntryValue.getCacheStatus() == CustomInstallationIdCache.CacheStatus.SYNCED) {
-      // If the given custom installation id matches up the cached
-      // value, there's no need to update.
-      return Tasks.forResult(null);
-    }
+    TaskCompletionSource<Void> taskCompletionSource = new TaskCompletionSource<>();
+
+    executor.execute(
```
vkryachko: 2 points:
* There is a dedicated `Tasks.call(executor, callable)` to achieve this effect
* Consider extracting the lambda below into its own (private) method that you can annotate with `@WorkerThread`

This applies to the here and throughout as well

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
```diff
@@ -191,50 +215,67 @@ public static FirebaseSegmentation getInstance(@NonNull FirebaseApp app) {
...
-        finalUpdateCacheResult -> {
+
+    TaskCompletionSource<Void> taskCompletionSource = new TaskCompletionSource<>();
+
+    executor.execute(
+        () -> {
+          InstanceIdResult instanceIdResult;
+          try {
+            instanceIdResult = Tasks.await(firebaseInstanceId.getInstanceId());
+          } catch (Exception e) {
```
vkryachko: Is there any particular exception you are anticipating here? as catching `Exception` is very broad: http://go/java-practices/exceptions#catch_exception

firebase-segmentation/src/main/java/com/google/firebase/segmentation/remote/SegmentationServiceClient.java
```diff
@@ -14,43 +14,160 @@
...
+        gzipOutputStream.write(
+            buildUpdateCustomSegmentationDataRequestBody(resourceName, customInstallationId)
+                .toString()
+                .getBytes("UTF-8"));
+      } catch (JSONException e) {
+        throw new IllegalStateException(e);
+      } finally {
+        gzipOutputStream.close();
+      }
+      httpsURLConnection.connect();
```
vkryachko: I don't think you need `connect()` as you've already written in to the connection

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
diwu-arete: Sure.

firebase-segmentation/src/main/java/com/google/firebase/segmentation/FirebaseSegmentation.java
diwu-arete: Very good learning!
Done

PR #578, opened 2019-07-01
Adding an "extras" column to transport databases

ashwinraghav:
1. We add an `extras` column to the `transport_contexts` table.
2. We introduce migrations to the new state for existing clients. Table creates are also modeled as migrations. 
3. Migrations are upgrade only. Downgrades are lossy and tables are dropped and recreated. Here's why
   1. When app developers downgrade their app, our SDKs potentially get downgraded (from say Vn to Vn-1). Since SDKs cannot know in advance what migrations are coming, Vn-1 will not know that (say) renaming a column will restore the database to the state expected by code in Vn-1.
   2. The other option would be to enforce all database migrations to be indefinitely backward compatible going forward. This will allow us to downgrade lossless-ly.
Both options do not yield us enough value, with the potential to introduce bugs.
4. We test migrations using a combination of high level operations on the `SQLiteEventStore` and operations on `SchemaManager`.
5. We no longer need to inject SQL statements, since the behavior of the store can be controlled by injecting only the version of the database. (Sorry! 😞 )


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -17,20 +17,95 @@
...
+
+  private static final String DROP_EVENT_METADATA_SQL = "DROP TABLE event_metadata";
+
+  private static final String DROP_CONTEXTS_SQL = "DROP TABLE transport_contexts";
+
+  static int SCHEMA_VERSION = 2;
+
+  static final SchemaManager.Migration MIGRATE_TO_V1 =
+      (db, fromVersion, toVersion) -> {
+        if (fromVersion < 1 && toVersion >= 1) {
```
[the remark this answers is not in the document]
vkryachko: On second thought, since the order of migrations matters, could we use the implicit order in the migration list as the target migration version?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -127,20 +130,33 @@ private long ensureTransportContext(SQLiteDatabase db, TransportContext transpor
     record.put("backend_name", transportContext.getBackendName());
     record.put("priority", transportContext.getPriority().ordinal());
     record.put("next_request_ms", 0);
+    if (transportContext.getExtras() != null) {
+      record.put("extras", encodeToString(transportContext.getExtras(), DEFAULT));
```
vkryachko: Do we actually have to encode it? could we just store the bytes in a `BLOB` column instead?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -17,20 +17,95 @@
...
+          db.execSQL(CREATE_CONTEXTS_SQL_V1);
+          db.execSQL(CREATE_EVENT_BACKEND_INDEX_V1);
+          db.execSQL(CREATE_CONTEXT_BACKEND_PRIORITY_INDEX_V1);
+        }
+      };
+
+  static final SchemaManager.Migration MIGRATE_TO_V2 =
+      (db, fromVersion, toVersion) -> {
+        if (fromVersion < 2 && toVersion >= 2) {
+          db.execSQL("ALTER TABLE transport_contexts ADD COLUMN extras TEXT");
```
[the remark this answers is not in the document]
ashwinraghav: Unfortunately, the db.query api does not allow us to elegantly query based on blob types.Hence the transformation

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -17,20 +17,95 @@
...
+          db.execSQL(
+              "CREATE UNIQUE INDEX contexts_backend_priority_extras on transport_contexts(backend_name, priority, extras)");
+          db.execSQL("DROP INDEX contexts_backend_priority");
+        }
+      };
+
+  static final List<Migration> INCREMENTAL_MIGRATIONS;
+
+  static {
+    INCREMENTAL_MIGRATIONS = Arrays.asList(MIGRATE_TO_V1, MIGRATE_TO_V2);
```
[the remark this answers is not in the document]
ashwinraghav: I did initially. Thanks :)

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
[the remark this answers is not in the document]
ashwinraghav: Thanks

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -155,12 +147,16 @@ public void onOpen(SQLiteDatabase db) {
   }
 
   private void upgrade(SQLiteDatabase db, int fromVersion, int toVersion) {
-    for (Migration m : INCREMENTAL_MIGRATIONS) {
-      m.upgrade(db, fromVersion, toVersion);
+    int highestVersion = INCREMENTAL_MIGRATIONS.size();
+
+    for (int version = 1; version <= highestVersion; version++) {
```
vkryachko: We might want to be more strict about this logic, e.g. throw an exception if `migrations.size() < toVersion` meaning we are missing some migrations for whatever reason. Additionally we should be able to iterate in the following range instead of from `1 - size()`: `int i = fromVersion; i <= toVersion`, this should also remove the need to do the if check below, wdyt?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: Can you switch to `Base64.encodeToString` and `Base64.DEFAULT` for readability? here and throughout

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
ashwinraghav: Just confirming. I think it should be < and not <=
```
    for (int version = fromVersion; version < toVersion; version++) {
```

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -17,20 +17,87 @@
...
+
+  private static final SchemaManager.Migration MIGRATE_TO_V2 =
+      (db) -> {
+        db.execSQL("ALTER TABLE transport_contexts ADD COLUMN extras BLOB");
+        db.execSQL(
+            "CREATE UNIQUE INDEX contexts_backend_priority_extras on transport_contexts(backend_name, priority, extras)");
+        db.execSQL("DROP INDEX contexts_backend_priority");
+      };
+
+  static final List<Migration> INCREMENTAL_MIGRATIONS = Arrays.asList(MIGRATE_TO_V1, MIGRATE_TO_V2);
```
vkryachko: `private`?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -55,21 +122,47 @@ private void ensureConfigured(SQLiteDatabase db) {
...
 
   @Override
   public void onDowngrade(SQLiteDatabase db, int oldVersion, int newVersion) {
-    ensureConfigured(db);
+    db.execSQL(DROP_EVENTS_SQL);
+    db.execSQL(DROP_EVENT_METADATA_SQL);
+    db.execSQL(DROP_CONTEXTS_SQL);
+    // Indices are dropped automatically when the tables are dropped
+
+    onCreate(db);
```
vkryachko: I think this is wrong, and should instead be `onUpgrade(0, newVersion)`, am I missing something?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -55,21 +122,47 @@ private void ensureConfigured(SQLiteDatabase db) {
...
   }
 
   @Override
   public void onOpen(SQLiteDatabase db) {
     ensureConfigured(db);
   }
+
+  private void upgrade(SQLiteDatabase db, int fromVersion, int toVersion) {
+    if (toVersion > INCREMENTAL_MIGRATIONS.size()) {
+      throw new RuntimeException(
```
vkryachko: Let's use `IllegalArgumentException` as `RuntimeException` is too generic.

transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManagerTest.java
```diff
@@ -71,4 +71,103 @@ public void persist_correctlyRoundTrips() {
...
+    SchemaManager schemaManager = new SchemaManager(RuntimeEnvironment.application, fromVersion);
+    SQLiteEventStore store = new SQLiteEventStore(clock, new UptimeClock(), CONFIG, schemaManager);
+    PersistedEvent event1 = store.persist(CONTEXT1, EVENT1);
+
+    schemaManager.onDowngrade(schemaManager.getWritableDatabase(), toVersion, fromVersion);
+
+    assertThat(store.loadBatch(CONTEXT1)).doesNotContain(event1);
+  }
+
+  @Test(expected = RuntimeException.class)
```
vkryachko: It's [not recommended](go/java-testing/exceptions) to use `@Test(expected)` as there is no way to tell at which point the exception was actually thrown, use `Assert.assertThrows` from junit.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
ashwinraghav: Net effect is the same.

```
@Override
  public void onCreate(SQLiteDatabase db) {
    ensureConfigured(db);
    // schemaVersion here should be the same as the newVersion in onDowngrade.
    upgrade(db, 0, schemaVersion);
  }
```

transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManagerTest.java
ashwinraghav: Took this approach since I saw other SDKs vendoring it in and I was lazy

transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManagerTest.java
```diff
@@ -131,13 +132,17 @@ public void downgradeV2ToV1_withNonEmptyDB_isLossy() {
...
-  @Test(expected = RuntimeException.class)
+  @Test
   public void upgrade_toANonExistentVersion_fails() {
     int oldVersion = 1;
     int nonExistentVersion = 1000;
     SchemaManager schemaManager = new SchemaManager(RuntimeEnvironment.application, oldVersion);
 
-    schemaManager.onUpgrade(schemaManager.getWritableDatabase(), oldVersion, nonExistentVersion);
+    Assert.assertThrows(
+        RuntimeException.class,
```
vkryachko: nit: IllegalArgumentException

PR #579, opened 2019-07-01
Throw last failed attempt's error in failed transactions

thebrianchen:
- Standardize errors based on the spreadsheet.
- Remove local validation for document not found in transactions.
- Checking that the transaction hasn't been committed should be an assert.

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -11,8 +11,8 @@
 - [feature] Added a `@DocumentId` annotation which can be used on a
   `DocumentReference` or `String` property in a POJO to indicate that the SDK
   should automatically populate it with the document's ID.
-- [changed] Failed transactions now throw the failure from the last attempt,
-  instead of `ABORTED.` 
+- [changed] Failed transactions now fail with the exception from the last 
+  attempt instead of an always failing with an exception with code `ABORTED`.
 
 # 20.1.0
 - [changed] SSL and gRPC initialization now happens on a separate thread, which
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -12,7 +12,7 @@
   `DocumentReference` or `String` property in a POJO to indicate that the SDK
   should automatically populate it with the document's ID.
 - [changed] Failed transactions now fail with the exception from the last 
-  attempt instead of an always failing with an exception with code `ABORTED`.
+  attempt instead of always failing with an exception with code `ABORTED`.
 
 # 20.1.0
 - [changed] SSL and gRPC initialization now happens on a separate thread, which
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -11,6 +11,8 @@
 - [feature] Added a `@DocumentId` annotation which can be used on a
   `DocumentReference` or `String` property in a POJO to indicate that the SDK
   should automatically populate it with the document's ID.
+- [changed] Failed transactions now throw the failure from the last attempt,
+  instead of `ABORTED.` 
 
 # 20.1.0
 - [changed] SSL and gRPC initialization now happens on a separate thread, which
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
@@ -435,7 +435,7 @@ public class TransactionTest {
     // assertEquals(1234, snapshot.getDouble("count"));
     // snapshot = waitFor(doc2.get());
     // assertEquals(16, snapshot.getDouble("count"));
-    assertEquals(Code.FAILED_PRECONDITION, ((FirebaseFirestoreException) e).getCode());
+    assertEquals(Code.INVALID_ARGUMENT, ((FirebaseFirestoreException) e).getCode());
     assertEquals("Every document read in a transaction must also be written.", e.getMessage());
   }
 
@@ -486,7 +486,7 @@ public class TransactionTest {
     // We currently require every document read to also be written.
     // TODO: Add this check back once we drop that.
     // assertEquals("bar", snapshot.getString("foo"));
-    assertEquals(Code.FAILED_PRECONDITION, ((FirebaseFirestoreException) e).getCode());
+    assertEquals(Code.INVALID_ARGUMENT, ((FirebaseFirestoreException) e).getCode());
     assertEquals("Every document read in a transaction must also be written.", e.getMessage());
   }
 
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
@@ -55,6 +55,7 @@ public class TransactionTest {
     Exception e = waitForException(firestore.runTransaction(transaction -> transaction.get(doc)));
     // We currently require every document read to also be written.
     // TODO: Fix this check once we drop that requirement.
+    assertEquals(Code.INVALID_ARGUMENT, ((FirebaseFirestoreException) e).getCode());
     assertEquals("Every document read in a transaction must also be written.", e.getMessage());
   }
 
@@ -232,7 +233,6 @@ public class TransactionTest {
     waitForException(transactionTask);
     assertFalse(transactionTask.isSuccessful());
     Exception e = transactionTask.getException();
-    assertTrue(e instanceof FirebaseFirestoreException);
     // This is the error surfaced by the backend.
     assertEquals(Code.INVALID_ARGUMENT, ((FirebaseFirestoreException) e).getCode());
   }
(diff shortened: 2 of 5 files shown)


firebase-firestore/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 - [feature] Added a `@DocumentId` annotation which can be used on a
   `DocumentReference` or `String` property in a POJO to indicate that the SDK
   should automatically populate it with the document's ID.
+- [changed] Failed transactions now throw the failure from the last attempt,
+  instead of `ABORTED.`
```
wilhuff: "throw" and "failure" don't go together that well. Usually you throw exceptions. However, to the end user we don't really throw anyway, since what they see is a `Task`. Tasks fail with exceptions.

Also, "ABORTED" here is _code_ the associated with the exception, not the thing we actually return.

How about:
> Failed transactions now fail with the exception from the last attempt instead of an always failing with an exception with code `ABORTED`.

thebrianchen: Ah, that sounds so much less confusing. I'll update the iOS to reflect this as well.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -480,18 +473,50 @@ public void testCannotReadAfterWriting() {
     assertNotNull(e);
   }
 
+  @Test
+  public void testReadandUpdateNonExistentDocumentWithExternalWrite() {
```
wilhuff: Capitalize "A" in "and" in this method name.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -480,18 +473,50 @@ public void testCannotReadAfterWriting() {
...
+    final FirebaseFirestore firestore = testFirestore();
```
wilhuff: Why `final` for one and not the other local that are referenced in the lambda below?

Note that Java 8 allows for references to variables from the lambda that are "effectively final" without requiring them to be actually final (which is why `counter` works as is). Consider just dropping the `final` here if that works.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -480,18 +473,50 @@ public void testCannotReadAfterWriting() {
...
+    AtomicInteger counter = new AtomicInteger(0);
+
+    // Make a transaction that will fail
+    Task<Void> transactionTask =
+        firestore.runTransaction(
+            transaction -> {
+              int count = counter.incrementAndGet();
+              // Get and update a document that doesn't exist so that the transaction fails.
+              DocumentReference doc = firestore.collection("nonexistent" + count).document();
```
wilhuff: This code: 

```java
DocumentReference doc = firestore.collection("nonexistent").document();
```

(i.e. without the `+ count`) should give you a new random document every time. You shouldn't need to use different collections for each iteration.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -480,18 +473,50 @@ public void testCannotReadAfterWriting() {
...
+              // This should fail, because the document didn't exist at the
+              // start of the transaction.
+              transaction.update(doc, "count", 1234);
+              return null;
+            });
+
+    waitForException(transactionTask);
+    assertFalse(transactionTask.isSuccessful());
+    Exception e = transactionTask.getException();
+    assertTrue(e instanceof FirebaseFirestoreException);
```
wilhuff: You can drop this assertion because:

  * You're already effectively forcing this test in the line below (by casting)
  * If this were to fail this error message would be "`false` is not `true`", which is actually less informative than the `ClassCastException` you'll get on the line below.

More generally, it's important to understand how a test is failing from its error message. If you really did need to assert that a value is an instanceof some class, use a construct that more directly expresses that:

Truth style:
```java
assertThat(e).isInstanceOf(FirebaseFirestoreException.class);
```

If the file you're making the assertion in is still using JUnit/Hamcrest style, it would look like this (after static imports):
```java
assertThat(e, instanceOf(FirebaseFirestoreException.class));
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Transaction.java
```diff
@@ -73,7 +79,7 @@ private void recordVersion(MaybeDocument doc) throws FirebaseFirestoreException
       if (!existingVersion.equals(doc.getVersion())) {
         // This transaction will fail no matter what.
         throw new FirebaseFirestoreException(
-            "Document version changed between two reads.", Code.FAILED_PRECONDITION);
+            "Document version changed between two reads.", Code.INVALID_ARGUMENT);
```
wilhuff: Spreadsheet F7 calls for this to be ABORTED.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Transaction.java
```diff
@@ -133,13 +140,14 @@ private Precondition precondition(DocumentKey key) {
...
    * UpdateOptions.
    */
-  private Precondition preconditionForUpdate(DocumentKey key) {
+  private Precondition preconditionForUpdate(DocumentKey key) throws FirebaseFirestoreException {
     @Nullable SnapshotVersion version = this.readVersions.get(key);
     if (version != null && version.equals(SnapshotVersion.NONE)) {
       // The document to update doesn't exist, so fail the transaction.
-      throw new IllegalStateException("Can't update a document that doesn't exist.");
+      throw new FirebaseFirestoreException(
+          "Can't update a document that doesn't exist.", Code.INVALID_ARGUMENT);
```
wilhuff: Spreadsheet F2 calls for this to be NOT_FOUND.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Transaction.java
```diff
@@ -53,6 +53,12 @@
   private final ArrayList<Mutation> mutations = new ArrayList<>();
   private boolean committed;
 
+  /**
+   * An error that may have occurred as a consequence of a write. If set, the task will fail instead
```
wilhuff: "write" is actually a thing we can do outside of committing (and isn't a method actually in this class) so this comment is somewhat confusing. Additionally the point isn't to capture arbitrary errors so much as usage errors that that can't exposed directly at the call site.

Maybe something like this?

> A deferred usage error that occurred previously in this transaction that will cause the transaction to fail once it actually commits.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Transaction.java
```diff
@@ -133,13 +140,14 @@ private Precondition precondition(DocumentKey key) {
    * Returns the precondition for a document if the operation is an update, based on the provided
    * UpdateOptions.
    */
-  private Precondition preconditionForUpdate(DocumentKey key) {
+  private Precondition preconditionForUpdate(DocumentKey key) throws FirebaseFirestoreException {
     @Nullable SnapshotVersion version = this.readVersions.get(key);
     if (version != null && version.equals(SnapshotVersion.NONE)) {
       // The document to update doesn't exist, so fail the transaction.
```
wilhuff: It's worth expanding on this comment because we were both confused about why we needed to do this. We need to check this locally because we can't express the constraint to the backend.

> The document to update doesn't exist, so fail the transaction.
>
> This has to be validated locally because you can't send a precondition that a document does not exist without changing the semantics of the backend write to be an insert. This is the reverse of what we want, since we want to assert that the document doesn't exist but then send the update and have it fail. Since we can't express that to the backend, we have to validate locally.
>
> Note: this can change once we can send separate verify writes in the transaction.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Transaction.java
thebrianchen: This should stay as INVALID_ARGUMENT b/c the user is performing a get first in the transaction.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -55,9 +55,7 @@ public void testGetDocuments() {
     Exception e = waitForException(firestore.runTransaction(transaction -> transaction.get(doc)));
     // We currently require every document read to also be written.
     // TODO: Fix this check once we drop that requirement.
-    assertEquals("Transaction failed all retries.", e.getMessage());
-    assertEquals(
-        "Every document read in a transaction must also be written.", e.getCause().getMessage());
+    assertEquals("Every document read in a transaction must also be written.", e.getMessage());
```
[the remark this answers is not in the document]
wilhuff: Sorry, I meant here and throughout, where applicable.

PR #582, opened 2019-07-02
Upgrade repo to AGP 3.4.1 and Gradle 5.4.1

vkryachko:


firebase-common/data-collection-tests/data-collection-tests.gradle
```diff
@@ -0,0 +1,52 @@
+// Copyright 2018 Google LLC
```
allisonbm92: What are these data collection tests? Are these related to upgrading Gradle, or are they part of something else?

firebase-database/firebase-database.gradle
```diff
@@ -73,6 +73,7 @@ dependencies {
     implementation project(':firebase-common')
     implementation project(':firebase-database-collection')
 
+    implementation 'androidx.annotation:annotation:1.1.0'
```
allisonbm92: Does the upgrade require us to add this?

vkryachko: Yes, these are needed to satisfy android lint dependencies check. Also these dependencies are correct because database indeed uses these annotations, they just used to come transitively from gms.basement

firebase-common/data-collection-tests/data-collection-tests.gradle
vkryachko: Indeed they are part of the upgrade, agp 3.4 changed the way robolectric initializes, as a side-effect they no longer support the deprecated feature `@Config(manifest="SomeOtherManifest.xml")`, so I had to extract these into a dedicated project to be able to configure a different data collection value in the manifest for tests.

allisonbm92:
@vkryachko Looks like you need to adjust the heap size for Gradle. Gradle 5 decreased the default limit, and check-changed is exceeding it.

vkryachko:
Yep, that's what I've been playing around with, pushed the update

PR #585, opened 2019-07-03
Fix flaky integration tests

schmidt-sebastian:
This fixes the "Running on the wrong queue" assertion that is sometimes triggered when an integration test uses two Firestores (callees of [testCollectionWithDocs()](https://github.com/firebase/firebase-android-sdk/blob/70d4388d08ca08e608a38850716435ebe87cabed/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java#L297)).

I am not 100% certain what leads to this behavior, but the problem is that https://github.com/grpc/grpc-java/blob/f3bf250a46571db3321fc3078b55e74280bdfd87/core/src/main/java/io/grpc/internal/DelayedClientTransport.java#L294 doesn't have the executor sets and then chooses the default executor, which seemingly can be from another instance.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
@@ -77,7 +77,9 @@ public class GrpcCallProvider {
             () -> {
               ManagedChannel channel = initChannel(context, databaseInfo);
               FirestoreGrpc.FirestoreStub firestoreStub =
-                  FirestoreGrpc.newStub(channel).withCallCredentials(firestoreHeaders);
+                  FirestoreGrpc.newStub(channel)
+                      .withCallCredentials(firestoreHeaders)
+                      .withExecutor(asyncQueue.getExecutor());
               callOptions = firestoreStub.getCallOptions();
               return channel;
             });
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
@@ -79,6 +79,9 @@ public class GrpcCallProvider {
               FirestoreGrpc.FirestoreStub firestoreStub =
                   FirestoreGrpc.newStub(channel)
                       .withCallCredentials(firestoreHeaders)
+                      // Ensure all callbacks are issued on the worker queue. If this call is
+                      // removed, all calls need to be audited to make sure they are executed on the
+                      // right thread.
                       .withExecutor(asyncQueue.getExecutor());
               callOptions = firestoreStub.getCallOptions();
               return channel;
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -77,7 +77,9 @@ public static void overrideChannelBuilder(
             () -> {
               ManagedChannel channel = initChannel(context, databaseInfo);
               FirestoreGrpc.FirestoreStub firestoreStub =
-                  FirestoreGrpc.newStub(channel).withCallCredentials(firestoreHeaders);
+                  FirestoreGrpc.newStub(channel)
+                      .withCallCredentials(firestoreHeaders)
+                      .withExecutor(asyncQueue.getExecutor());
```
rsgowman: I don't understand why this isn't a noop? It seems like we already set the executor down on line 122, and my understanding of https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/AbstractStub.html#withExecutor-java.util.concurrent.Executor- would be that the executor set on 122 should apply here.

schmidt-sebastian: The debug view when I hit the codepath from the PR description:

Screen Shot 2019-07-08 at 1 38 55 PM

The executor here is set to the our queue, but only because we don't have an override via CallOptions. It seems relying on a default can cause weirdness if two Firestore clients are active at the same time. If we specify the executor on a per-call level, this weirdness goes away (I have updated the PR to no longer set the executor on the channel itself).

I wish I fully understood what is going on here, but I can't seem to get to the bottom of this. I do however have a 100% failure rate without this PR (for a full run of all integration tests against the emulator) and a 100% success rate with it.

If you have any further insight that could help me debug this, let me know :(

rsgowman: Oh, I think I see it. Maybe.

We build up the stub, but then immediately discard it (first saving the CallOptions.) The CallOptions only saves the options set on the stub, and doesn't take into account the options set on the channel within the stub. (The stub passes it's CallOptions to it's channel when making an rpc.)

Then, we call ManagedChannel::newCall, passing in the calloptions. (The channel itself already has the correct executor, so leaving it null in calloptions *should* work, though apparently doesn't.)

Eventually, we create a https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ClientCallImpl.java, passing in those same CallOptions. Then we call start() on that.

We eventually create the ClientTransport (which is presumably the DelayedClientTransport one that you linked) here: https://github.com/grpc/grpc-java/blob/cc13f74065911dab19c4e757cf559fd84ff3207e/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L257. Note that it gets passed the callOptions, but not the executor.

I don't know if that's a bug in grpc, or is WAI. But while I'm still murky on the details, I think it does explain why your fix works. I think this is worth involving the grpc people to confirm one way or the other.

But for now, your fix lgtm.

schmidt-sebastian: I am not sure if I know have even more questions :) I was never able to fully figure out all the interactions. I will file an issue in the GRPC-Java repo to figure out what the root cause is. Thank you!

schmidt-sebastian: Filed https://github.com/grpc/grpc-java/issues/5966

PR #589, opened 2019-07-08
Prepare `shutdown` to be a public api.

Hui-Wu:
1. FirebaseApp delete calls firestore shutdown.
2. Deregister firestore when shutdown is called, new instance can then be registered.
3. Synchronize AsyncQueue.Executor scheduling. Also make it stop handling most of schedule requests once it is shutdown, only tasks originated from shutdown itself and clearPersistence can proceed.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
@@ -1059,13 +1059,25 @@ public class FirestoreTest {
     assertTrue(instance.getClient().isShutdown());
   }
 
-  @Test(expected = IllegalStateException.class)
+  @Test
   public void testNewOperationThrowsAfterFirestoreShutdown() {
     FirebaseFirestore instance = testFirestore();
-    waitFor(instance.document("abc/123").set(Collections.singletonMap("Field", 100)));
+    final DocumentReference reference = instance.document("abc/123");
+    DocumentReference ref = reference;
+    waitFor(reference.set(Collections.singletonMap("Field", 100)));
 
     instance.shutdown();
 
-    waitForException(instance.document("abc/123").get());
+    final String expectedMessage = "The client has already been shutdown";
+    expectError(() -> waitFor(reference.get()), expectedMessage);
+    expectError(() -> waitFor(reference.update("Field", 1)), expectedMessage);
+    expectError(
+        () -> waitFor(reference.set(Collections.singletonMap("Field", 1))), expectedMessage);
+    expectError(
+        () -> waitFor(instance.runBatch((batch) -> batch.update(reference, "Field", 1))),
+        expectedMessage);
+    expectError(
+        () -> waitFor(instance.runTransaction(transaction -> transaction.get(reference))),
+        expectedMessage);
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
@@ -1062,8 +1062,7 @@ public class FirestoreTest {
   @Test
   public void testNewOperationThrowsAfterFirestoreShutdown() {
     FirebaseFirestore instance = testFirestore();
-    final DocumentReference reference = instance.document("abc/123");
-    DocumentReference ref = reference;
+    DocumentReference reference = instance.document("abc/123");
     waitFor(reference.set(Collections.singletonMap("Field", 100)));
 
     instance.shutdown();
@@ -1079,5 +1078,22 @@ public class FirestoreTest {
     expectError(
         () -> waitFor(instance.runTransaction(transaction -> transaction.get(reference))),
         expectedMessage);
+  }
+
+  @Test
+  public void testShutdownCalledMultipleTimes() {
+    FirebaseFirestore instance = testFirestore();
+    DocumentReference reference = instance.document("abc/123");
+    waitFor(reference.set(Collections.singletonMap("Field", 100)));
+
+    instance.shutdown();
+
+    final String expectedMessage = "The client has already been shutdown";
+    expectError(() -> waitFor(reference.get()), expectedMessage);
+
+    // Calling a second time should go through and change nothing.
+    instance.shutdown();
+
+    expectError(() -> waitFor(reference.get()), expectedMessage);
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
@@ -20,6 +20,7 @@ import static com.google.firebase.firestore.testutil.IntegrationTestUtil.provide
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollection;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollectionWithDocs;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testDocument;
+import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirebaseApp;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitForException;
@@ -30,6 +31,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
@@ -38,6 +40,7 @@ import static org.junit.Assert.fail;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.firebase.FirebaseApp;
 import com.google.firebase.Timestamp;
 import com.google.firebase.firestore.FirebaseFirestoreException.Code;
 import com.google.firebase.firestore.Query.Direction;
@@ -1023,4 +1026,41 @@ public class FirestoreTest {
     FirebaseFirestoreException firestoreException = (FirebaseFirestoreException) e;
     assertEquals(Code.FAILED_PRECONDITION, firestoreException.getCode());
   }
+
+  @Test
+  public void testRestartFirestoreLeadsToNewInstance() {
+    FirebaseApp app = testFirebaseApp();
+    FirebaseFirestore instance = FirebaseFirestore.getInstance(app);
+    FirebaseFirestore sameInstance = FirebaseFirestore.getInstance(app);
+
+    assertSame(instance, sameInstance);
+    waitFor(instance.document("abc/123").set(Collections.singletonMap("field", 100L)));
+
+    instance.shutdown();
+    FirebaseFirestore newInstance = FirebaseFirestore.getInstance(app);
+
+    // Verify new instance works.
+    DocumentSnapshot doc = waitFor(newInstance.document("abc/123").get());
+    assertEquals(doc.get("field"), 100L);
+    waitFor(newInstance.document("abc/123").delete());
+
+    // Verify it is different instance.
+    assertNotSame(instance, newInstance);
+  }
+
+  /*
+  @Test
+  public void testAppDeleteLeadsToFirestoreShutdown() {
+    FirebaseApp app = testFirebaseApp();
+    FirebaseFirestore instance = FirebaseFirestore.getInstance(app);
+    instance.document("abc/123").set(Collections.singletonMap("Field", 100));
+
+    app.delete();
+
+    try {
+      instance.document("abc/123").get();
+    } catch (Exception e) {
+
+    }
+  }*/
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
@@ -1048,19 +1048,24 @@ public class FirestoreTest {
     assertNotSame(instance, newInstance);
   }
 
-  /*
   @Test
   public void testAppDeleteLeadsToFirestoreShutdown() {
     FirebaseApp app = testFirebaseApp();
     FirebaseFirestore instance = FirebaseFirestore.getInstance(app);
-    instance.document("abc/123").set(Collections.singletonMap("Field", 100));
+    waitFor(instance.document("abc/123").set(Collections.singletonMap("Field", 100)));
 
     app.delete();
 
-    try {
-      instance.document("abc/123").get();
-    } catch (Exception e) {
+    assertTrue(instance.getClient().isShutdown());
+  }
 
-    }
-  }*/
+  @Test(expected = IllegalStateException.class)
+  public void testNewOperationThrowsAfterFirestoreShutdown() {
+    FirebaseFirestore instance = testFirestore();
+    waitFor(instance.document("abc/123").set(Collections.singletonMap("Field", 100)));
+
+    instance.shutdown();
+
+    waitForException(instance.document("abc/123").get());
+  }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -117,7 +117,13 @@ public class FirebaseFirestore {
     // so there is no need to include it in the persistence key.
     String persistenceKey = app.getName();
 
-    return new FirebaseFirestore(context, databaseId, persistenceKey, provider, queue, app);
+    FirebaseFirestore firestore =
+        new FirebaseFirestore(context, databaseId, persistenceKey, provider, queue, app);
+    app.addLifecycleEventListener(
+        /* onDeleted */ (firebaseAppName, options) -> {
+          firestore.shutdown(/* fromAppDeletion */ true);
+        });
+    return firestore;
   }
 
   @VisibleForTesting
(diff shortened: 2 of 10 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
```diff
@@ -176,97 +176,180 @@ private void markDone() {
...
+   *   <li>1. Synchronized task scheduling. This is different from function 3, which is about task
+   *       execution in a single thread.
+   *   <li>2. Ability to do soft-shutdown: only critical tasks related to shutting Firestore SDK
+   *       down can be executed once the shutdown process initiated.
+   *   <li>3. Single threaded execution service, no concurrent execution among the `Runnable`s
+   *       scheduled in this Executor.
+   * </ul>
    */
-  private final Thread thread;
+  private class SynchronizedShutdownAwareExecutor implements Executor {
```
wilhuff: This seems like a lot of extra complexity to support the notion of tasks being able to execute post shutdown. Do we really need this?

What if we just made the default behavior lenient such that tasks submitted via the usual executor interface and everything but `enqueueAndForgetEvenAfterShutdown` just ignored?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
```diff
@@ -1023,4 +1026,58 @@ public void testClearPersistenceWhileRunningFails() {
...
+    expectError(() -> waitFor(reference.update("Field", 1)), expectedMessage);
+    expectError(
+        () -> waitFor(reference.set(Collections.singletonMap("Field", 1))), expectedMessage);
+    expectError(
+        () -> waitFor(instance.runBatch((batch) -> batch.update(reference, "Field", 1))),
+        expectedMessage);
+    expectError(
+        () -> waitFor(instance.runTransaction(transaction -> transaction.get(reference))),
+        expectedMessage);
+  }
```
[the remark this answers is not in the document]
Hui-Wu: Added test for multiple calls to shutdown.

`testFirestore` creates an instance without `app`, so that is already tested.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -117,7 +117,13 @@ static FirebaseFirestore newInstance(
     // so there is no need to include it in the persistence key.
     String persistenceKey = app.getName();
 
-    return new FirebaseFirestore(context, databaseId, persistenceKey, provider, queue, app);
+    FirebaseFirestore firestore =
+        new FirebaseFirestore(context, databaseId, persistenceKey, provider, queue, app);
+    app.addLifecycleEventListener(
+        /* onDeleted */ (firebaseAppName, options) -> {
+          firestore.shutdown(/* fromAppDeletion */ true);
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -329,13 +335,43 @@ public WriteBatch batch() {
...
+      FirestoreMultiDbComponent component = this.getApp().get(FirestoreMultiDbComponent.class);
+      component.remove(this.databaseId.getDatabaseId());
+    }
     // The client must be initialized to ensure that all subsequent API usage throws an exception.
     this.ensureClientConfigured();
     return client.shutdown();
   }
 
+  /**
+   * Shuts down this Firestore SDK instance.
```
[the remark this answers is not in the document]
Hui-Wu: Used your version without any change!

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -329,13 +335,43 @@ public WriteBatch batch() {
...
+   * desired in your case.
+   *
+   * <p>Note that {@link #clearPersistence()} is an exception, it is designed to run when the
+   * Firestore SDK instance is shutdown.
+   *
+   * <p>To re-start a Firestore SDK instance, simply create a new instance via {@link
+   * #getInstance()} or {@link #getInstance(FirebaseApp)}.
+   */
+  @VisibleForTesting
+  // TODO: Make this public and remove @VisibleForTesting
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -329,13 +335,43 @@ public WriteBatch batch() {
...
+  Task<Void> shutdown() {
+    return shutdown(/* fromAppDeletion */ false);
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
```diff
@@ -45,12 +45,23 @@
 
   /** Provides instances of Firestore for given database names. */
   @NonNull
-  synchronized FirebaseFirestore get(@NonNull String databaseName) {
-    FirebaseFirestore firestore = instances.get(databaseName);
+  synchronized FirebaseFirestore get(@NonNull String databaseId) {
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -138,21 +137,21 @@ public FirestoreClient(
...
-            this.clientShutdown = true;
+          remoteStore.shutdown();
+          persistence.shutdown();
+          if (lruScheduler != null) {
+            lruScheduler.stop();
           }
         });
   }
 
+  /** Has this client been shutdown. */
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
```diff
@@ -176,97 +176,180 @@ private void markDone() {
   }
 
   /**
-   * The single thread that will be used by the executor. This is created early and managed directly
-   * so that it's possible later to make assertions about executing on the correct thread.
+   * A wrapper around a {@link ScheduledThreadPoolExecutor} class that provides:
+   *
+   * <ul>
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
Hui-Wu: It looks like a lot but most of the code actually already existed. I moved them from `AsyncQueue` to nested class `SynchronizedShutdownAwareExecutor`.

`SynchronizedShutdownAwareExecutor` implements `Executor` and replace the instance `AsyncQueue` holds. 

`SynchronizedShutdownAwareExecutor` delegate most of its work to `internalExecutor` which plays the role of previous `AsyncQueue.executor`. 

`SynchronizedShutdownAwareExecutor` adds two things (the real changes):

1. Check for shutdown state before accepting a `runnable`.
2. Synchronization of task scheduling/accepting. This is required to prevent race conditions around `isShuttingDown`. Otherwise some task might be scheduled after `shutdown` is called.

This does what you said in your comment above.

(Hope i made it clearer.)

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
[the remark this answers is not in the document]
wilhuff: In javadoc this will now read:

```
1. 1. Synchronized ...
2. 2. Ability ...
3. 3. Single ...
```
`<ol>` makes each `<li>` insert a number instead of a bullet.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
wilhuff: I agree that most of the code already existed -- it's not the volume of code I'm objecting to.

The issue I'm trying to express is that there are now multiple executors in play where there used to be one, and the abstraction is incomplete because we're still reaching into the inner executor. We also need to do so carefully to preserve the correct semantics, but at the call site it's not obvious what's happening (see `executor.internalExecutor` used in `enqueueAndForgetEvenAfterShutdown` -- it's not obvious to the reader that this is necessary to bypass the shutdown checks that are otherwise in play).

A similar concern is that the threading primitives in here aren't obvious. AsyncQueue doesn't synchronize on itself, but instead synchronizes on its inner executor. This pattern of nonlocal, non-self-contained synchronization is error prone and hard to evaluate for correctness.

What I'm asking for (and don't entirely understand if it's possible) is can we implement the two concerns you're expressing directly in the AsyncQueue without introducing the additional executor?

Hui-Wu: Modified to hide internalExecutor from outside world as we discussed:
1. Stopped using `AsyncQueue.getExecutorForShutdown` from `GrpcCallProvider`, instead just using `Tasks.await` as you suggested.
2. Pull more implementations to `SynchronizedShutdownAwareExecutor` from `AsyncQueue` such that all `synchronized` operation is contained within SynchronizedShutdownAwareExecutor.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -137,48 +138,53 @@ private ManagedChannel initChannel(Context context, DatabaseInfo databaseInfo) {
...
-                FirestoreChannel.class.getSimpleName(),
-                "Interrupted while shutting down the gRPC Managed Channel");
-            // Preserve interrupt status
-            Thread.currentThread().interrupt();
-          }
-        });
+    ManagedChannel channel = null;
+    try {
+      channel = Tasks.await(channelTask);
+    } catch (ExecutionException | InterruptedException e) {
```
wilhuff: If you catch `InterruptedException` you need to re-interrupt the current thread. That means that you can't handle these cases together.

Also, if there was an `ExecutionException` this may be our only opportunity to log what happened, so we should log the the exception too.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -137,48 +138,53 @@ private ManagedChannel initChannel(Context context, DatabaseInfo databaseInfo) {
...
-
-            // Similar to above, something bad happened, but it's not worth asserting. Just log it.
-            Logger.warn(
-                FirestoreChannel.class.getSimpleName(),
-                "Interrupted while shutting down the gRPC Managed Channel");
-            // Preserve interrupt status
-            Thread.currentThread().interrupt();
-          }
-        });
+    ManagedChannel channel = null;
```
wilhuff: This needs a comment about why we're handling it synchronously because this is otherwise a suspicious idea.

Something like: "Handle shutdown synchronously to avoid re-enqueuing on the AsyncQueue after shutdown has started."

firebase-firestore/src/test/java/com/google/firebase/firestore/util/AsyncQueueTest.java
```diff
@@ -123,4 +123,19 @@ public void canManuallyDrainSpecificDelayedTasksForTesting() throws Exception {
...
+    queue.enqueueAndForget(runnableForStep(1));
+
+    // From this point on, `normal` tasks are not scheduled. Only those who explicitly request to
+    // run after shutdown initiated will run.
+    queue.enqueueAndInitiateShutdown(runnableForStep(2));
+
+    queue.enqueueAndForget(runnableForStep(3));
+    queue.enqueueAndForgetEvenAfterShutdown(runnableForStep(4));
+
+    queue.getExecutor().execute(runnableForStep(5));
```
wilhuff: This test makes no assertions. It seems like this should assert that 1, 2, and 4 completed.

tools/errorprone/src/main/java/com/google/firebase/errorprone/ComponentsAppGetCheck.java
```diff
@@ -87,6 +87,19 @@
...
+   *
+   * <pre>{@code
+   * class Foo {
+   *     void shutdown();
+   * }
+   * </pre>
+   */
+  private static final Matcher<ExpressionTree> WITHIN_SHUTDOWN =
+      enclosingMethod(allOf(methodIsNamed("shutdown")));
+  // methodReturns(isVoidType())));
```
wilhuff: Commented out code?

tools/errorprone/src/main/java/com/google/firebase/errorprone/ComponentsAppGetCheck.java
```diff
@@ -87,6 +87,19 @@
                   isSupertypeOf(
                       state -> ASTHelpers.getType(state.findEnclosing(ClassTree.class))))));
 
+  /**
```
wilhuff: This needs review by someone on the Android Core team.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
Hui-Wu: Done.

tools/errorprone/src/main/java/com/google/firebase/errorprone/ComponentsAppGetCheck.java
Hui-Wu: Removed.

firebase-firestore/src/test/java/com/google/firebase/firestore/util/AsyncQueueTest.java
Hui-Wu: Added `waitForExpectedSteps`, this plus the assertions in `runnableForStep` should be sufficient now.

tools/errorprone/src/main/java/com/google/firebase/errorprone/ComponentsAppGetCheck.java
Hui-Wu: Asked Vlad to review.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -329,13 +335,42 @@ public WriteBatch batch() {
     return batch.commit();
   }
 
-  @VisibleForTesting
-  Task<Void> shutdown() {
+  Task<Void> shutdown(boolean fromAppDeletion) {
+    if (!fromAppDeletion && this.getApp() != null) {
+      FirestoreMultiDbComponent component = this.getApp().get(FirestoreMultiDbComponent.class);
```
vkryachko: You should not modify the errorprone check for "convenience" as this convenience comes at a price:

* You're essentially gaining access to any *undeclared* component dependency. This is something we(acore) want teams to do only if absolutely necessary and only as a temporary measure that they should consider tech debt.
* It makes the code harder to unit test as you're no longer getting everything you need via a constructor but effectively gaining access to services through another object you have access to via a "service locator" pattern(see https://github.com/google/guice/wiki/InjectOnlyDirectDependencies)
* Additionally, in this particular case, you are coupling this class to `FirestoreMultiDbComponent` imo unnecessarily

That said, I'd recommend something along the following lines:

* add a constructor parameter to `FirebaseFirestore` in the form of `Consumer<String> onShutdown`(or equivalent as it is java8 only)
* have `FirestoreMultiDbComponent` inject the `Consumer<String>` into `Firestore` as a simple lambda that removes the instance based on the provided `String` `dbId`
* have firestore call `onShutdown` when appropriate.

Hui-Wu: Basically what you suggested, slight tweak.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
```diff
@@ -18,12 +18,19 @@
...
+import com.google.firebase.FirebaseOptions;
 import com.google.firebase.auth.internal.InternalAuthProvider;
 import java.util.HashMap;
 import java.util.Map;
 
 /** Multi-resource container for Firestore. */
-class FirestoreMultiDbComponent {
+class FirestoreMultiDbComponent implements FirebaseAppLifecycleListener {
+
+  public interface DeregisterCommand {
```
vkryachko: `java.lang.Runnable` instead?

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
```diff
@@ -41,16 +48,39 @@
     this.context = context;
     this.app = app;
     this.authProvider = authProvider;
+    this.app.addLifecycleEventListener(this);
```
vkryachko: Just a consideration: `app.delete()` is not a documented API and no other SDK integrates with it(hence it is effectively almost a noop), I'll defer to your judgement on whether you want to integrate with it

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
wilhuff: That's a good idea.

Alternatively, you could just make an interface that contains `remove(String databaseId)` such that `FirestoreMultiDbComponent` could just implement it. This removes the extra capture.

In any case, the interface describing this behavior should not be nested in `FirestoreMultiDbComponent` because that means `FirebaseFirestore` still ends up having an awareness of that type. Instead nest the interface in `FirebaseFirestore`.

Something like this is what I mean:

```java
class FirebaseFiresore {
  // ...
  interface InstanceRegistry {
    void remove(@NonNull String databaseId);
  }
}
```

FirestoreMultiDbComponent can implement this interface directly.

Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
vkryachko: sgtm

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
```diff
@@ -41,16 +45,38 @@
...
+   *
+   * <p>It is a no-op if there is no instance associated with the given database name.
+   */
+  @Override
+  public synchronized void remove(@NonNull String databaseId) {
+    instances.remove(databaseId);
+  }
+
+  @Override
+  public void onDeleted(String firebaseAppName, FirebaseOptions options) {
```
vkryachko: Shouldn't this method also synchronize its access to the map?

Hui-Wu: Fixed. Thanks.

PR #595, opened 2019-07-10
Expose operation counts from Persistence layer

schmidt-sebastian:
This PR adds a pretty-bare bones way to export document read counts from the RemoteDocumentCache and the MutationQueue. I will use this in more thorough tests once we start optimizing our query execution.

The different ways to construct the Persistence layer are getting a little out of hand. I think a good follow up would be to add Builders for some of the classes here and address the overload overload.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
@@ -156,7 +156,7 @@ final class MemoryMutationQueue implements MutationQueue {
           .addToCollectionParentIndex(mutation.getKey().getPath().popLast());
     }
 
-    statsCollector.recordRowsWritten(MemoryMutationQueue.TAG, 1);
+    statsCollector.recordRowsWritten(STATS_TAG, 1);
 
     return batch;
   }
@@ -164,7 +164,7 @@ final class MemoryMutationQueue implements MutationQueue {
   @Nullable
   @Override
   public MutationBatch lookupMutationBatch(int batchId) {
-    statsCollector.recordRowsRead(MemoryMutationQueue.TAG, 1);
+    statsCollector.recordRowsRead(STATS_TAG, 1);
 
     int index = indexOfBatchId(batchId);
     if (index < 0 || index >= queue.size()) {
@@ -209,7 +209,7 @@ final class MemoryMutationQueue implements MutationQueue {
       result.add(batch);
     }
 
-    statsCollector.recordRowsRead(MemoryMutationQueue.TAG, result.size());
+    statsCollector.recordRowsRead(STATS_TAG, result.size());
 
     return result;
   }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
@@ -69,9 +69,11 @@ final class MemoryMutationQueue implements MutationQueue {
   private ByteString lastStreamToken;
 
   private final MemoryPersistence persistence;
+  private final StatsCollector statsCollector;
 
-  MemoryMutationQueue(MemoryPersistence persistence) {
+  MemoryMutationQueue(MemoryPersistence persistence, StatsCollector statsCollector) {
     this.persistence = persistence;
+    this.statsCollector = statsCollector;
     queue = new ArrayList<>();
 
     batchesByDocumentKey = new ImmutableSortedSet<>(emptyList(), DocumentReference.BY_KEY);
@@ -154,12 +156,16 @@ final class MemoryMutationQueue implements MutationQueue {
           .addToCollectionParentIndex(mutation.getKey().getPath().popLast());
     }
 
+    statsCollector.recordRowsWritten(MemoryMutationQueue.TAG, 1);
+
     return batch;
   }
 
   @Nullable
   @Override
   public MutationBatch lookupMutationBatch(int batchId) {
+    statsCollector.recordRowsRead(MemoryMutationQueue.TAG, 1);
+
     int index = indexOfBatchId(batchId);
     if (index < 0 || index >= queue.size()) {
       return null;
@@ -202,6 +208,8 @@ final class MemoryMutationQueue implements MutationQueue {
       hardAssert(batch != null, "Batches in the index must exist in the main table");
       result.add(batch);
     }
+
+    statsCollector.recordRowsRead(MemoryMutationQueue.TAG, result.size());
 
     return result;
   }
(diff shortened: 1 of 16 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
```diff
@@ -154,12 +156,16 @@ public MutationBatch addMutationBatch(
...
 
+    statsCollector.recordRowsWritten(MemoryMutationQueue.TAG, 1);
+
     return batch;
   }
 
   @Nullable
   @Override
   public MutationBatch lookupMutationBatch(int batchId) {
+    statsCollector.recordRowsRead(MemoryMutationQueue.TAG, 1);
```
wilhuff: Can't this just be `TAG`? Also even if you don't want to name it just that, it's actually `MutationQueue.TAG`.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MutationQueue.java
```diff
@@ -25,6 +25,9 @@
 
 /** A queue of mutations to apply to the remote store. */
 interface MutationQueue {
+  /** The tag used by the StatsCollector. */
+  String TAG = "Mutations";
```
wilhuff: It would be nice if these tags matched the table names we're reading from. This one would be `mutations` and the other one `remote_documents`.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteMutationQueue.java
```diff
@@ -288,13 +307,17 @@ public MutationBatch getNextMutationBatchAfterBatchId(int batchId) {
...
+    int rowsRead[] = new int[] {0};
+
     List<MutationBatch> result = new ArrayList<>();
     Set<Integer> uniqueBatchIds = new HashSet<>();
     while (longQuery.hasMoreSubqueries()) {
       longQuery
           .performNextSubquery()
           .forEach(
               row -> {
+                ++rowsRead[0];
```
wilhuff: Would it make sense to just make this a part of forEach and similar methods (and the return value of the method)?

Just spitballing. Feel free to ignore.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -102,18 +100,38 @@ public void onCommit() {
...
+        StatsCollector.newNoOpStatsCollector(),
+        params);
+  }
+
   public SQLitePersistence(
       Context context,
       String persistenceKey,
       DatabaseId databaseId,
       LocalSerializer serializer,
+      StatsCollector statsProvider,
```
wilhuff: Your class is `StatsCollector` but the variable is `statsProvider`. Elsewhere you called it `statsCollector` too. Make these consistent?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
```diff
@@ -39,17 +39,22 @@
 
   private final SQLitePersistence db;
   private final LocalSerializer serializer;
+  private final StatsCollector statsProvider;
 
-  SQLiteRemoteDocumentCache(SQLitePersistence persistence, LocalSerializer serializer) {
+  SQLiteRemoteDocumentCache(
+      SQLitePersistence persistence, LocalSerializer serializer, StatsCollector statsProvider) {
```
wilhuff: s/statsProvider/statsCollector/ as above

firebase-firestore/src/main/java/com/google/firebase/firestore/local/StatsCollector.java
```diff
@@ -0,0 +1,50 @@
...
+ * Collects the operation count from the persistence layer. Implementing classes can expose this
+ * information to measure the efficiency of persistence operations.
+ *
+ * <p>The only consumer of the operation counts is currently the LocalStoreTestCase (via {@link
+ * com.google.firebase.firestore.local.AccumulatingStatsCollector}). If you are not interested in
+ * the stats, you can use `newNoOpStatsCollector()` to return an empty stats collector.
+ */
+abstract class StatsCollector {
+  /** A stats collector that discards all operation counts. */
+  static class NoOpStatsCollector extends StatsCollector {
```
wilhuff: This could be `private static` if `newNoOpStatsCollector()` had the return type of `StatsCollector`.

Alternatively, just consider making `StatsCollector` non-abstract and `new StatsCollector()` would be enough to get you a no-op instance. That would cut the line count here and reduce the number of classes in production by one.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -849,17 +873,28 @@ public void testCanExecuteMixedCollectionQueries() {
     allocateQuery(query);
     assertTargetId(2);
 
+    localStore.executeQuery(query);
```
wilhuff: This isn't awful but please consider making a separate test that specifically exercises interesting cases rather than making additional assertions as a side-effect of some other scenario. In the current configuration it becomes difficult to understand what part of the test is essential to verifying mixed collection queries vs what's essential to verifying the stats.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/StatsCollector.java
schmidt-sebastian: Are your really gonna make me change this PR from a "sophisticated stats framework" to a dumb counter? Okay then. It's done.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
schmidt-sebastian: I basically wanted to write the exact same test that we have here. I did remove the checks in 877 and 888, which should improve the readability of this test.

I can create a separate test but it would likely contain most lines from this test. What do you think?

wilhuff: The worry I essentially have is that we end up wanting to change this test to verify some additional case and then get tripped up around the interaction with the counts and then it's not clear if changing the counts is the right thing to do.

I couldn't find a reference to it quickly, but some hardline testing guru types advocate for only having one assertion per test case essentially on these grounds.

Most of these tests are copies of each other in some way so I don't think that's much of an issue.

schmidt-sebastian: Updated the PR to add a separate test.

PR #600, opened 2019-07-11
Add missing nullability annotations to Firestore.

vkryachko:
Additionally remove uses of the redundant `@PublicApi` annotation.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -17,13 +17,12 @@ package com.google.firebase.firestore;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import com.google.android.gms.tasks.Task;
-import com.google.firebase.annotations.PublicApi;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.ResourcePath;
 import com.google.firebase.firestore.util.Executors;
 import com.google.firebase.firestore.util.Util;
-import javax.annotation.Nullable;
 
 /**
  * A CollectionReference can be used for adding documents, getting document references, and querying
@@ -33,7 +32,6 @@ import javax.annotation.Nullable;
  * test mocks. Subclassing is not supported in production code and new SDK releases may break code
  * that does so.
  */
-@PublicApi
 public class CollectionReference extends Query {
 
   CollectionReference(ResourcePath path, FirebaseFirestore firestore) {
@@ -50,7 +48,6 @@ public class CollectionReference extends Query {
 
   /** @return The ID of the collection. */
   @NonNull
-  @PublicApi
   public String getId() {
     return query.getPath().getLastSegment();
   }
@@ -63,7 +60,6 @@ public class CollectionReference extends Query {
    *     collection.
    */
   @Nullable
-  @PublicApi
   public DocumentReference getParent() {
     ResourcePath parentPath = query.getPath().popLast();
     if (parentPath.isEmpty()) {
@@ -80,7 +76,6 @@ public class CollectionReference extends Query {
    * @return The path of this collection.
    */
   @NonNull
-  @PublicApi
   public String getPath() {
     return query.getPath().canonicalString();
   }
@@ -92,7 +87,6 @@ public class CollectionReference extends Query {
    * @return A DocumentReference pointing to a new document with an auto-generated ID.
    */
   @NonNull
-  @PublicApi
   public DocumentReference document() {
     return document(Util.autoId());
   }
@@ -105,7 +99,6 @@ public class CollectionReference extends Query {
    * @return The DocumentReference instance.
    */
   @NonNull
-  @PublicApi
   public DocumentReference document(@NonNull String documentPath) {
     checkNotNull(documentPath, "Provided document path must not be null.");
     return DocumentReference.forPath(
@@ -121,7 +114,6 @@ public class CollectionReference extends Query {
    * @return A Task that will be resolved with the DocumentReference of the newly created document.
    */
   @NonNull
-  @PublicApi
   public Task<DocumentReference> add(@NonNull Object data) {
     checkNotNull(data, "Provided data must not be null.");
     final DocumentReference ref = document();
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
@@ -217,9 +217,7 @@ public class DocumentReference {
    */
   @NonNull
   public Task<Void> update(
-      @NonNull FieldPath fieldPath,
-      @Nullable Object value,
-      Object... moreFieldsAndValues) {
+      @NonNull FieldPath fieldPath, @Nullable Object value, Object... moreFieldsAndValues) {
     ParsedUpdateData parsedData =
         firestore
             .getDataConverter()
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
@@ -196,7 +196,7 @@ public class DocumentReference {
    */
   @NonNull
   public Task<Void> update(
-      @NonNull String field, @Nullable Object value, @NonNull Object... moreFieldsAndValues) {
+      @NonNull String field, @Nullable Object value, Object... moreFieldsAndValues) {
     ParsedUpdateData parsedData =
         firestore
             .getDataConverter()
@@ -219,7 +219,7 @@ public class DocumentReference {
   public Task<Void> update(
       @NonNull FieldPath fieldPath,
       @Nullable Object value,
-      @NonNull Object... moreFieldsAndValues) {
+      Object... moreFieldsAndValues) {
     ParsedUpdateData parsedData =
         firestore
             .getDataConverter()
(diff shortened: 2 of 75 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
```diff
@@ -205,9 +195,8 @@ public CollectionReference collection(@NonNull String collectionPath) {
    * @return A Task that will be resolved when the write finishes.
    */
   @NonNull
-  @PublicApi
   public Task<Void> update(
-      @NonNull String field, @Nullable Object value, Object... moreFieldsAndValues) {
+      @NonNull String field, @Nullable Object value, @NonNull Object... moreFieldsAndValues) {
```
wilhuff: This seems incorrect. `null` is a valid value in here. Does the `@NonNull` apply to elements of the argument list or to the argument list itself?

vkryachko: It is for the argument list itself, there is no way to annotate members of the list unfortunately, so the kotlin signature will be
```kotlin
fun update(field: String, value: Any?, vararg moreFieldsAndValues: Any!) // note the !
```

Furthermore I only added this to silence the lint check, as this annotation has no effect on the above signature.

I removed these annotations now and updated the check to ignore varargs parameters.

firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
```diff
@@ -17,13 +17,12 @@
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
```
wilhuff: I'm not a fan of the Nullable change to androidx (but also can't say I care too strongly). Is there any good reason to prefer the android-specific version of this over the standard one other than just being consistent with the source of `NonNull`?

In some cases this change would be harmful. For example, we share our `CustomClassMapper` with the Firestore Java Server SDK. If we had properly annotated methods like [this one](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java#L108) then we really should use `javax.annotation.Nullable` because we can't have android dependencies on the server.

Also, the javax version of this seems to be the only `Nullable` I get with code completion so this is going to rot without some kind of wider announcement to go reconfigure IDEs and add a linter check or something like that to enforce the preference. Is this change really worth all that?

wilhuff:
Aside from the question about which Nullable to use, LGTM

firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
vkryachko: You're right, it has no Kotlin behavior difference compared to `androidx` it's mostly for consistency across all SDKs, so I am ok in principle to revert firestore back to using jsr305 annotations. That said, there is nothing standard about javax annotations, they are part of a JSR that has been dormant since 2006 and is unlikely to become standard at all.

As for `androidx.annotations` being unusable on the server, I don't think that is true, `androidx.annotations` is a jar(not aar) that has no dependencies so nothing prevents them from being used on the server. Whether they have any semantic meaning for java server frameworks is a good question though.

That said, I don't feel too strongly but I have a slight preference towards androidx at least for the public APIs for consistency's sake across SDKs, and I am ok to keep javax for all internal APIs if you like. wdyt?

wilhuff: I'd prefer some outcome where we can set it and forget it as far as Android Studio is concerned. Having different classes for public and private classes is something you can't make automatic--every code-completion import would have to choose and it seems like such a speed bump for something that has no practical value. It also seems low enough value that it's unlikely to get caught in code review.

JSR 305 was never ratified, but it seems effectively standard, especially for Google distributed software. For example, even if we standardize on the androidx one, JSR 305 will be in our classpath because that's what Guava, gRPC, and others use.

I'm more concerned that this standardization isn't going to hold automatically since doing so requires people to change their editor configuration. We need to do something to force the world to track on that: a lint check that disallows `javax.annotation.Nullable` would be sufficient. Absent that, we'll soon have both `androidx` and `javax` flavors in our API and otherwise which I think is worse than mixing the sources of NonNull and Nullable.

To be clear: I'm actually OK with standardizing on the androidx one throughout. The number of classes where we need to interoperate with the server is quite small so we can opt of a lint check in those cases should we implement such a check.

vkryachko: Added lint check, note also that kotlin interop lint checks will provide automatic fix hints(in IDE) with correct(androidx) annotations.

vkryachko: Actually, I am going to disable it for now as it ripples to other SDKs, and this PR is already large enough. Will re-enable in a follow-up PR

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
```diff
@@ -0,0 +1,123 @@
...
+                doVisit(parameter)
+            }
+        }
+
+        override fun visitField(node: UField) {
+            doVisit(node)
+        }
+
+        private fun doVisit(node: UDeclaration) {
+            for (annotation in node.annotations) {
```
rlazo: annotation is a reserved word in kotlin

https://kotlinlang.org/docs/reference/grammar.html#identifiers

https://kotlinlang.org/docs/reference/annotations.html

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
```diff
@@ -0,0 +1,123 @@
...
+                    ensureAndroidNullability(context, annotation)
+                }
+            }
+        }
+
+        private fun ensureAndroidNullability(context: JavaContext, annotation: UAnnotation) {
+            annotation.qualifiedName?.let { name ->
+                val path = name.split('.')
+                val packageName = path.subList(0, path.size - 1).joinToString(".")
+                if (packageName in listOf("androidx.annotation", "android.support.annotation")) {
```
rlazo: I actually prefer the style used in 

https://github.com/JetBrains/kotlin/blob/master/core/descriptors.jvm/src/org/jetbrains/kotlin/load/java/JvmAnnotationNames.kt

where you declare a list of fully qualified nullable & nonnullable annotations. Then it's a bit clearer than checking for the packageName.

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
vkryachko: it's only a context sensitive keyword and is perfectly fine to use(proper kotlin parsers don't even highlight it unlike Github's highlighter)

rlazo: TIL There are hard keywords, soft keywords, and modifier keywords in Kotlin.

https://kotlinlang.org/docs/reference/keyword-reference.html

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
vkryachko: Done

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
```diff
@@ -0,0 +1,131 @@
...
+        private fun doVisit(node: UDeclaration) {
+            for (annotation in node.annotations) {
+                if (isNullabilityAnnotation(annotation)) {
+                    ensureAndroidNullability(context, annotation)
+                }
+            }
+        }
+
+        private fun ensureAndroidNullability(context: JavaContext, annotation: UAnnotation) {
+            annotation.qualifiedName?.let { name ->
```
wilhuff: Though less fancy you might consider returning early if qualifiedName is null to save indentation.

It's not really a problem here since it's just the one condition, but it makes things much easier to understand in the presence of multiple. See also http://go/tott-455.

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
```diff
@@ -0,0 +1,136 @@
...
+                if (name in ANDROIDX_ANNOTATIONS) {
+                    return
+                }
+
+                val simpleName = name.split('.').last()
+                val replacement = if (simpleName in NOT_NULL_ANNOTATIONS) "NonNull" else "Nullable"
```
wilhuff: Reading this I initially thought this was a bug (thinking you might replace all annotations with `Nullable`) until I realized that this method is gated by `isNullabilityAnnotation`.

That method is essentially doing everything you're doing in this method so consider just inlining here as another guard condition.

tools/lint/src/test/kotlin/AndroidxNullabilityTests.kt
```diff
@@ -0,0 +1,130 @@
...
+
+private fun javaxAnnotation(name: String): String {
+    return annotationSource("javax.annotation", name)
+}
+
+private fun androidxAnnotation(name: String): String {
+    return annotationSource("androidx.annotation", name)
+}
+
+private val NULLABLE_CLASS = """
```
wilhuff: You might consider naming this `JAVAX_NULLABLE_CLASS` to contrast more directly against the `ANDROIDX` flavor, especially since the latter is actually preferred.

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
vkryachko: good point done

tools/lint/src/main/kotlin/NonAndroidxNullabilityDetector.kt
vkryachko: done

tools/lint/src/test/kotlin/AndroidxNullabilityTests.kt
vkryachko: done

PR #602, opened 2019-07-11
Add nullability annotations, remove `@PublicApi` annotations for Storage.

vkryachko:


firebase-storage/src/main/java/com/google/firebase/storage/StorageTask.java
```diff
@@ -277,8 +267,8 @@ public boolean isPaused() {
    * @throws IllegalStateException if the Task is not yet complete
    * @throws RuntimeExecutionException if the Task failed with an exception
    */
+  @Nullable
```
schmidt-sebastian: This method does not seem to return null.

PR #603, opened 2019-07-11
 Add nullability annotations, remove `@PublicApi` annotations for RTDB.

vkryachko:


vkryachko:
@schmidt-sebastian done

Issue #372, opened 2019-04-23
Stop hardcoding RTDB version to 3.0.0.


mikelehen:
https://github.com/firebase/firebase-android-sdk/blob/7a6b23a58fc187306895842193e1de927618c4b2/firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java#L41 needs to be updated to use `BuildConfig.VERSION_NAME` similar to https://github.com/firebase/firebase-android-sdk/blob/5b06f42913f0bb97335bb2d69c3a3201301f240e/firebase-database/src/main/java/com/google/firebase/database/DatabaseRegistrar.java#L43

PR #605, opened 2019-07-11
Report the correct version for the RTDB

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-android-sdk/issues/372

### DIFF
--- a/firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
+++ b/firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
@@ -38,7 +38,7 @@ import com.google.firebase.database.core.utilities.Validation;
 @PublicApi
 public class FirebaseDatabase {
 
-  private static final String SDK_VERSION = "3.0.0";
+  private static final String SDK_VERSION = BuildConfig.VERSION_NAME;
 
   private final FirebaseApp app;
   private final RepoInfo repoInfo;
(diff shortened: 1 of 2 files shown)


schmidt-sebastian:
CC @davidair

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -38,7 +38,7 @@
 @PublicApi
 public class FirebaseDatabase {
 
-  private static final String SDK_VERSION = "3.0.0";
+  private static final String SDK_VERSION = BuildConfig.VERSION_NAME;
```
wilhuff: This almost seems like it would be better reviewed by someone on the acore team. How does the version get set in the gradle configuration?

https://github.com/firebase/firebase-android-sdk/blob/40439c3b1dbd53bb96014652a453d44e0a840a49/firebase-database/firebase-database.gradle#L34

schmidt-sebastian: I assigned this to Vlad. FWIW, we already use this version here: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/DatabaseRegistrar.java#L43

wilhuff: Ah. In that case, LGTM.

PR #609, opened 2019-07-12
@Hide abt public classes

rlazo:
A/B Testing doesn't have user visible classes, so they should be  annotated with javadoc's @hide.

### DIFF
--- a/firebase-abt/src/main/java/com/google/firebase/abt/component/package-info.java
+++ b/firebase-abt/src/main/java/com/google/firebase/abt/component/package-info.java
@@ -1,0 +1,4 @@
+/**
+ * @hide
+ */
+package com.google.firebase.abt.component;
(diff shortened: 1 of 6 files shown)


firebase-abt/src/main/java/com/google/firebase/abt/component/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * @hide
```
vkryachko: I think the whole comment needs to be on one line, let's see if google-java-format fails on this.

rlazo: PTAL

vkryachko: there is a merge conflict in both files, pls double check

rlazo: fixed

PR #610, opened 2019-07-12
Legacy Firelog backend impl

ashwinraghav:
1. We create 2 destinations - CCT and LegacyFlg.
2. CctBackendFactory's `create` returns a backend injected with a different destination based on the backend name requested.
3. The Api has a nit. It allows the creation of `CctTransportBackend` with (say) `cct` backend name, but log events to it using the CCTFlgDestination, which is harmless, but the other way around can result in backend errors.
4. I initially implemented throwing the endpoints into the destinations (available in one of the commits), but reversed it since it would require that we store urls per event and live with the Api caveat instead


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CCTDestination.java
```diff
@@ -0,0 +1,48 @@
...
+import androidx.annotation.Nullable;
+import com.google.android.datatransport.runtime.Destination;
+
+public class CCTDestination implements Destination {
+  static final String DESTINATION_NAME = "cct";
+  private static volatile CCTDestination instance;
+
+  private CCTDestination() {}
+
+  public static CCTDestination getInstance() {
```
vkryachko: What's the intention behind this method, why not just have a public field? If you're optimizing for lazy instantiation, imo it may not be worth it as the only reason CCTDestination class itself would be loaded is if someone wants to call getInstance()

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CCTDestination.java
```diff
@@ -0,0 +1,48 @@
...
+          instance = new CCTDestination();
+        }
+      }
+    }
+    return instance;
+  }
+
+  @Nullable
+  @Override
+  public String getName() {
```
vkryachko: Hm, not sure why this is `@Nullable`, should it be changed to `@NonNull` here and in the interface?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -0,0 +1,90 @@
...
+
+import static com.google.android.datatransport.cct.CctBackendFactory.mergeStrings;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.android.datatransport.runtime.Destination;
+import com.google.auto.value.AutoValue;
+import java.io.UnsupportedEncodingException;
+
+@AutoValue
```
vkryachko: I don't think we need this to be auto-value, it only adds size and boilerplate, can you make it a plain class?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -0,0 +1,90 @@
...
+abstract class LegacyFlgDestination implements Destination {
+  static final String DESTINATION_NAME = "lflg";
+  static final String DEFAULT_API_KEY = mergeStrings("AzSCki82AwsLzKd5O8z", "IayckHiZRO1EFl1aGoKo");
+
+  @NonNull
+  public abstract String getAPIKey();
+
+  /** Returns a new builder for {@link LegacyFlgDestination}. */
+  public static Builder builder() {
```
vkryachko: Also a builder seems overkill for a class with 1 member

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -0,0 +1,90 @@
...
+  public abstract static class Builder {
+    public abstract Builder setAPIKey(@NonNull String apiKey);
+
+    public Builder withDefaultAPIKey() {
+      return setAPIKey(DEFAULT_API_KEY);
+    }
+
+    public abstract LegacyFlgDestination build();
+
+    Builder fromExtras(@NonNull byte[] extras) {
```
vkryachko: Can this be a static method on the parent class itself?

vkryachko: also, it looks like the only use for this method is to be able to decode the apikey from bytes, can this functionality be achieved without instantiating this class? e.g. by having a `static String decode(byte[] bytes);`

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CCTDestination.java
ashwinraghav: Can't remember why. Resorting to Nonnull

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
ashwinraghav: I was not sure how much about the "extras" we wanted the backend to know about.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -0,0 +1,72 @@
...
+  public byte[] getExtras() {
+    return encodeString(apiKey);
+  }
+
+  @Nullable
+  static byte[] encodeString(@NonNull String s) {
+    try {
+      return s.getBytes("UTF-8");
+    } catch (UnsupportedEncodingException e) {
+      e.printStackTrace();
```
vkryachko: We don't want to print to stdout, additionally this can never actually throw as utf-8 cannot be unsupported, I'd recommend to make the method `@NonNull` and throw an `AssertionError` or `IllegalStateException` here, wdyt?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return null;
+  }
+
+  @Nullable
+  static String decodeExtras(@NonNull byte[] a) {
+    try {
+      return new String(a, "UTF-8");
+    } catch (UnsupportedEncodingException e) {
+      e.printStackTrace();
```
vkryachko: ditto

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctBackendFactory.java
```diff
@@ -15,26 +15,35 @@
...
     return new CctTransportBackend(
         creationContext.getApplicationContext(),
-        URL,
+        url,
         creationContext.getWallClock(),
         creationContext.getMonotonicClock());
   }
 
-  @VisibleForTesting
   static String mergeStrings(String part1, String part2) {
```
vkryachko: Since this is now used by multiple classes, consider extracting into its own class not to couple every caller to this factory class

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CCTDestination.java
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.android.datatransport.cct;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.android.datatransport.runtime.Destination;
+
+public class CCTDestination implements Destination {
```
vkryachko: nit: `final`

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CCTDestination.java
```diff
@@ -0,0 +1,38 @@
...
+  static final String DESTINATION_NAME = "cct";
+  private static volatile CCTDestination INSTANCE = new CCTDestination();
```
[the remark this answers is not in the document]
ashwinraghav: Sry. Tech debt from the previous impl

ashwinraghav:
> lglg

In my understanding this would not be necessary since it is only the CCT backend factory is the only API that needs to be discovered. It returns either a CCT or LFLG impl based on the creation context. 

Is that not the design you had in mind?

vkryachko:
afaict it still is required. The issue is that LegacyDestination has `getName()` different from `cct`, which is the only name `CctBackendFactory` is registered with, so if you request a factory with `name=lflg`, the backend will not be found and the event will go to `/dev/null`

ashwinraghav:
> afaict it still is required. The issue is that LegacyDestination has `getName()` different from `cct`, which is the only name `CctBackendFactory` is registered with, so if you request a factory with `name=lflg`, the backend will not be found and the event will go to `/dev/null`

Ack. I think I follow. Done. It seemed like the `MetadataBackendRegistry` already supports a comma delimiter. Hope this works

vkryachko:
> It seemed like the `MetadataBackendRegistry` already supports a comma delimiter. Hope this works

Cool, I did not realize it had been implemented

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CCTDestination.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.google.android.datatransport.cct;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.android.datatransport.runtime.Destination;
+
+public final class CCTDestination implements Destination {
+  static final String DESTINATION_NAME = "cct";
+  private static final CCTDestination INSTANCE = new CCTDestination();
```
vkryachko: still needs to be public, otherwise this class is unusable, I can't construct it and can get a hold of this instance.

PR #614, opened 2019-07-15
Index-Free (1/6): Adding LocalStoreTestHelper for NO_CHANGE event

schmidt-sebastian:
This is the first of 6 PRs for "index-free" querying.

This PR just adds a test helper to send a resume token (for LocalStoreTest). As part of this, I had to move TestTargetMetadataProvider to a different TestUtil.java (on a related note, we have too many of those...)


firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -526,6 +545,36 @@ public static ByteString streamToken(String contents) {
     return ByteString.copyFrom(contents, Charsets.UTF_8);
   }
 
+  /**
+   * An implementation of TargetMetadataProvider that provides controlled access to the
+   * `TargetMetadataProvider` callbacks. Any target accessed via these callbacks must be registered
+   * beforehand via `setSyncedKeys()`.
+   */
+  public static class TestTargetMetadataProvider
```
wilhuff: Consider making this just a top-level class (i.e. in its own source file).

firebase-firestore/ktx/src/test/java/com/google/firebase/firestore/TestUtil.java
```diff
@@ -138,36 +138,6 @@ public static QuerySnapshot querySnapshot(
     return new QuerySnapshot(query(path), viewSnapshot, FIRESTORE);
   }
 
-  /**
```
wilhuff: Thanks for doing this. At some point we should probably cut these down to the minimum, given the assumption there's no reasonable way to share with the main Firestore testutil.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -912,26 +908,15 @@ public void testPersistsResumeTokens() {
...
-    ByteString resumeToken = resumeToken(1000);
 
-    QueryData queryData = TestUtil.queryData(targetId, QueryPurpose.LISTEN, "foo/bar");
-    TestTargetMetadataProvider testTargetMetadataProvider = new TestTargetMetadataProvider();
-    testTargetMetadataProvider.setSyncedKeys(queryData, DocumentKey.emptyKeySet());
-
-    WatchChangeAggregator aggregator = new WatchChangeAggregator(testTargetMetadataProvider);
-
-    WatchTargetChange watchChange =
-        new WatchTargetChange(WatchTargetChangeType.Current, asList(targetId), resumeToken);
```
wilhuff: This is a current change, not a no-change. Why are these are substitutable?

firebase-firestore/ktx/src/test/java/com/google/firebase/firestore/TestUtil.java
schmidt-sebastian: FWIW, only half of the methods are currently used, but it doesn't seem trivial to just put them in the test class (so I refrained from doing this for now).

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
schmidt-sebastian: Since works because we persist the resume token for both NO_CHANGE and CURRENT:

https://github.com/firebase/firebase-android-sdk/blob/d9ed980c39c58ddf55ab945baa120c2e0a9fa8cb/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregator.java#L109
https://github.com/firebase/firebase-android-sdk/blob/d9ed980c39c58ddf55ab945baa120c2e0a9fa8cb/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregator.java#L138

firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
schmidt-sebastian: Done

PR #616, opened 2019-07-15
Index-Free (3/6): Persist a Query's Sync State

schmidt-sebastian:
This is the third of six PRs for "index-free" queries.

This PR adds a "consistent" metadata flag to QueryData. A consistent query is a query whose View has has no Limbo documents and doesn't need to be re-filled. 

Queries with Limbo documents are not eligible for index-free query execution since a local document with an older update can match a query even though the it is not longer part of the Query Target Mapping. We can, however, not guarantee that no query see this old document, which can cause inconsistent state between views:

- Query is executed against the local cache. The Query’s Target Mapping says that Doc A and Doc B match.
- SDK raises a snapshot with Doc A and Doc B.
- The backend responds with a snapshot that only includes Doc A. We update the Query’s Target Mapping.
- SDK is shut down/crashes.
- After restart, the query is executed again. Because we changed the Snapshot Time and the Query’s Target Mapping, the query only returns Doc A.
- User issues another query against the same collection. This query can now return Doc A and the outdated Doc B.
- User wonders why Doc B matches one query but not the other, even though their version of DocB should match both.

Similarly, a limit query that needs to be refilled from cache can also not be executed based on the Query Target Mapping. If local mutations cause a document to fall out of the limit, we need to scan the full set of local documents since a document with an older update time might take its spot.

- User issues a limit query. We raise a snapshot with Doc A through Doc M.
- The backend confirms that Doc A through Doc M match the limit. We persist the Query target mapping.
- Doc M is edited and the latency-compensated view longer matches.
- The SDK re-execute the query to find a local document that falls into the limit. Since older documents in the cache can now match, we cannot use the updated target mapping.


wilhuff:
Could you add a bit of text to the description about why queries that aren't current or that have limbo documents are ineligible for index-free treatment?

schmidt-sebastian:
I updated the PR:
- After some thinking, I don't think we need to wait for a query to become CURRENT to make it eligible for Index-Free queries.
- With that said, I decided that "synced" was no longer a good term to use for what it is after (since it used in combination with CURRENT in other places in the SDK). I updated the PR to use the terms consistent/consistenWithLocalView/consistentWithBackend.

I also modified the PR description to include some of the contents from the design doc.

schmidt-sebastian:
Per offline discussion, I will update this PR to use store the last consistent snapshot version rather than a boolean flag. This will let us use index-free queries more often.

schmidt-sebastian:
I updated the PR to only track whether there are limbo documents and to store the last limbo-free snapshot version. I plumbed this PR through to all other PRs and have all tests passing.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalViewChanges.java
```diff
@@ -70,6 +73,14 @@ public int getTargetId() {
     return targetId;
   }
 
+  /**
+   * Returns whether there were any unresolved limbo documents in the corresponding view when this
+   * change was computed.
+   */
+  public boolean hasUnresolvedLimboDocuments() {
```
schmidt-sebastian: Note that I picked this name to suggest that the limbo state of the ViewChange doesn't reflect whether the change itself contains limbo documents.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -469,12 +469,35 @@ public void notifyLocalViewChanges(List<LocalViewChanges> viewChanges) {
...
+            if (!viewChange.hasUnresolvedLimboDocuments()) {
+              QueryData queryData = targetIds.get(targetId);
+              hardAssert(
+                  queryData != null,
+                  "Can't set limbo-free snapshot version for unknown target: %s",
+                  targetId);
+
+              // Advance the last limbo free snapshot version
+              SnapshotVersion lastLimboFreeSnapshotVersion = queryData.getSnapshotVersion();
+              QueryData updatedQueryData =
```
wilhuff: This seems like it would be more readable if we added copy-setters for the things we update, something like `queryData.withLastLimboFreeSnapshotVersion(lastLimboFreeSnapshotVersion)`.

What do you think?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -548,10 +571,20 @@ public void releaseQuery(Query query) {
 
           int targetId = queryData.getTargetId();
           QueryData cachedQueryData = targetIds.get(targetId);
+
+          boolean needsUpdate = false;
```
wilhuff: Given how huge this method already is, consider pulling out this needs-update logic into a helper where you can make the internal decisions with simple returns.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: I will send you a follow-up that updates the code here and elsewhere.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: I went down this route, but since there is already a `shouldPersistQueryData()` in LocalStore, I am afraid that adding an additional method that is named similarly but doesn't quite do what the existing method does will add too much confusion.

PR #617, opened 2019-07-15
Index-Free (4/6): Filter document queries by read time

schmidt-sebastian:
This is the fourth of six PRs for "index-free" queries.

This PR adds the ability to `getAllDocumentsMatchingQuery()` to skip documents older than a given update time. This allows us to only scan for documents that have been modified since the Query's Target Mapping has been persisted.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
@@ -131,15 +131,22 @@ final class LocalDocumentsView {
   // documents in a given collection so that SimpleQueryEngine can do that and then filter in
   // memory.
 
-  /** Performs a query against the local view of all documents. */
-  ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(Query query) {
+  /**
+   * Performs a query against the local view of all documents.
+   *
+   * @param query The query to match documents against.
+   * @param sinceUpdateTime If not set to SnapshotVersion.MIN, return only documents that have been
+   *     modified since this snapshot version (exclusive).
+   */
+  ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
+      Query query, SnapshotVersion sinceUpdateTime) {
     ResourcePath path = query.getPath();
     if (query.isDocumentQuery()) {
       return getDocumentsMatchingDocumentQuery(path);
     } else if (query.isCollectionGroupQuery()) {
-      return getDocumentsMatchingCollectionGroupQuery(query);
+      return getDocumentsMatchingCollectionGroupQuery(query, sinceUpdateTime);
     } else {
-      return getDocumentsMatchingCollectionQuery(query);
+      return getDocumentsMatchingCollectionQuery(query, sinceUpdateTime);
     }
   }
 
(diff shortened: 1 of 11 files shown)


schmidt-sebastian:
Unassigning while I am waiting for PRs that this depends upon.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -131,15 +131,22 @@ private MaybeDocument getDocument(DocumentKey key, List<MutationBatch> inBatches
   // documents in a given collection so that SimpleQueryEngine can do that and then filter in
   // memory.
 
-  /** Performs a query against the local view of all documents. */
-  ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(Query query) {
+  /**
+   * Performs a query against the local view of all documents.
+   *
+   * @param query The query to match documents against.
+   * @param sinceUpdateTime If not set to SnapshotVersion.MIN, return only documents that have been
```
wilhuff: Isn't this read time now?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -112,6 +113,11 @@ public MaybeDocument get(DocumentKey key) {
         continue;
       }
 
+      SnapshotVersion readTime = entry.getValue().second;
+      if (readTime.compareTo(sinceUpdateTime) <= 0) {
```
wilhuff: If there exist documents with a read time of zero and you pass `SnapshotVersion.MIN`, won't this skip them?

wilhuff: The migration in #686 didn't populate read time, so all existing documents in the cache will have a zero read time.

Could you add tests for how this is expected to work against pre-existing documents?

schmidt-sebastian: We don't really have good testing infrastructure (yet) that tests client functionality after migration. We have some tests that pretend to do client like queries before and after a schema migration, but they are all low level. I'm not opposed to adding tests like this, but:
- the MemoryRemoteDocumentCache will not have any existing documents and is not affected by migrations
- the SQLiteRemoteDocumentCache reads all documents with NULL version numbers.

Let me know if you want me to add a test.

wilhuff: At some level we should prove to ourselves that this works and that it's going to stay working. I see the NULL check on the read_time so I think you're right that this is functional.

On some level this could just be a test that validates the migration behaves as we expect. If it's straightforward to do so, it would be useful to stand up a `SQLiteRemoteDocumentCache` in the `SQLiteSchemaTest` and validate that migrated documents always match queries regardless of the update time you pass.

If you also deleted NoDocuments with a zero time in that migration you could validate that they don't exist post as well.

wilhuff: Also, to be clear: If it costs too much time or effort to make it work I'm fine with skipping it but I think it's probably worth trying.

schmidt-sebastian: I hacked together a test. It wasn't as bad as I thought it would be, but in return it is also not very pretty :)

PR #620, opened 2019-07-15
Ignore some proto generated files from coverage.

rlazo:
Files ignored are those class files within a "proto/" directory when generated. Generated files in other 
directories are still being counted.


rlazo:
@yifanyang this only applies to classes that, once processed, have proto as part of their package name. Which is true for `firestore/proto/*`, `transport/../proto/*`, and `firebase-config/../proto/*`. 

While **not all generated protobuf** match this rule, **all files that match this rule** are protobuf generated. So it's a safe, intermediate change.

PR #626, opened 2019-07-17
Replace try-with-resources with try and finally so that we don't have desugaring issues

VinayGuthal:
Don't use try-with-resources instead explicity close it

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigStorageClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigStorageClient.java
@@ -76,8 +76,11 @@ public class ConfigStorageClient {
    */
   public synchronized Void write(ConfigContainer container) throws IOException {
     // TODO(issues/262): Consider using the AtomicFile class instead.
-    try (FileOutputStream outputStream = context.openFileOutput(fileName, Context.MODE_PRIVATE)) {
+    FileOutputStream outputStream = context.openFileOutput(fileName, Context.MODE_PRIVATE);
+    try {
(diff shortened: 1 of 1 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigStorageClient.java
```diff
@@ -76,8 +76,11 @@ private ConfigStorageClient(Context context, String fileName) {
    */
   public synchronized Void write(ConfigContainer container) throws IOException {
     // TODO(issues/262): Consider using the AtomicFile class instead.
-    try (FileOutputStream outputStream = context.openFileOutput(fileName, Context.MODE_PRIVATE)) {
+    FileOutputStream outputStream = context.openFileOutput(fileName, Context.MODE_PRIVATE);
+    try {
       outputStream.write(container.toString().getBytes(JSON_STRING_ENCODING));
+    } finally {
+       if(outputStream != null) outputStream.close();
```
vkryachko: why would it be null? I don't think it is possible

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigStorageClient.java
```diff
@@ -100,6 +104,8 @@ public synchronized ConfigContainer read() throws IOException {
     } catch (JSONException | FileNotFoundException e) {
       // File might not have been written to yet, so this not an irrecoverable error.
       return null;
+    } finally {
+       if(fileInputStream != null ) fileInputStream.close();
```
vkryachko: ditto

PR #628, opened 2019-07-18
Only retry transactions on non-permanent errors

thebrianchen:


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -400,10 +400,12 @@ public void testUpdateTransactionally() {
 
     // Let all of the transactions fetch the old value and stop once.
     waitFor(Tasks.whenAll(readTasks));
-    assertEquals(3, started.intValue());
-    // Let all of the transactions continue and wait for them to finish.
+    // There should be 3 initial transaction runs.
+    assertEquals(3, counter.get());
     barrier.setResult(null);
     waitFor(Tasks.whenAll(transactionTasks));
+    // There should be a total of 3 retries: once for the 2nd update, and twice for the 3rd update.
```
mikelehen: Why are we guaranteed that the 3rd one will retry twice?  This seems timing-dependent.  If the 2nd update happens really fast, the 3rd update might only retry once.

In practice, maybe this is relatively reliable, but I worry it could result in a flakey test...  At bare minimum I would add a comment that this isn't a guarantee and we should revisit if this test shows up as flakey.

Or am I missing something?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -622,6 +624,28 @@ public void testCannotHaveAGetWithoutMutations() {
...
+              count.incrementAndGet();
+              // Get and update a document that doesn't exist so that the transaction fails
+              DocumentSnapshot doc =
+                  transaction.get(firestore.collection("nonexistent").document());
+              transaction.update(doc.getReference(), "foo", "bar");
+              return null;
+            });
+
+    // Let all of the transactions fetch the old value and stop once.
+    waitForException(transactionTask);
```
mikelehen: Can we validate the exception is whatever code we're expecting it to be?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -622,6 +624,28 @@ public void testCannotHaveAGetWithoutMutations() {
...
+    assertEquals(1, count.get());
+    assertFalse(transactionTask.isSuccessful());
```
mikelehen: Is this assert necessary given we already called waitForException()?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -589,4 +595,19 @@ private boolean errorIsInteresting(Status error) {
 
     return false;
   }
+
+  private boolean isRetryableError(Exception e) {
```
mikelehen: Since this is specific to transactions, it should probably be named isRetryableTransactionError() or similar.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -589,4 +595,19 @@ private boolean errorIsInteresting(Status error) {
...
+    if (e instanceof FirebaseFirestoreException) {
+      // In transactions, the backend will fail outdated reads with FAILED_PRECONDITION and
+      // non-matching document versions with ABORTED. These errors should be retried.
+      List<FirebaseFirestoreException.Code> retryableExceptions =
+          Arrays.asList(
+              FirebaseFirestoreException.Code.ABORTED,
+              FirebaseFirestoreException.Code.FAILED_PRECONDITION);
+
+      return retryableExceptions.contains(((FirebaseFirestoreException) e).getCode())
+          || !Datastore.isPermanentError(((FirebaseFirestoreException) e).getCode());
```
mikelehen: I don't care strongly, but unless we expect this list to grow I'd probably just write the simplest code for this:
```
import com.google.firebase.firestore.FirebaseFirestoreException.Code;
...
    if (e instanceof FirebaseFirestoreException) {
      // In transactions, the backend will fail outdated reads with FAILED_PRECONDITION and
      // non-matching document versions with ABORTED. These errors should be retried.
      Code code = ((FirebaseFirestoreException)e).getCode();
      return code == Code.ABORTED || 
          code == Code.FAILED_PRECONDITION || 
          !Datastore.isPermanentError(code);
    }
```

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -221,6 +221,45 @@ public static boolean isPermanentError(Status status) {
     }
   }
 
+  /**
+   * Determines whether the given error code that represents a permanent error when received in
+   * response to a non-write operation.
+   *
+   * @see #isPermanentWriteError for classifying write errors.
+   */
+  public static boolean isPermanentError(FirebaseFirestoreException.Code code) {
```
mikelehen: Let's not duplicate this method. 😄 

You could change the existing method to:

```
  public static boolean isPermanentError(Status status) {
    return isPermanentError(FirebaseFirestoreException.Code.fromValue(status.getCode().value()));
  }
```

Or: 
1. Change this method to accept FirebaseFirestoreException instead of Code (and just call .getCode() to get the code).
2. Remove the old method and change all callers of `Datastore.isPermanentError(status)` to `Datastore.isPermanentError(Util.exceptionFromStatus(status))`

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
thebrianchen: That comment was mostly based on me running the integration test and seeing that the the 3rd transaction consistently retried twice, but duly noted, will update the comment accordingly.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -189,8 +189,18 @@ WriteStream createWriteStream(WriteStream.Callback listener) {
    * @see #isPermanentWriteError for classifying write errors.
    */
   public static boolean isPermanentError(Status status) {
+    return isPermanentError(FirebaseFirestoreException.Code.fromValue(status.getCode().value()));
+  }
+
+  /**
+   * Determines whether the given error code that represents a permanent error when received in
```
mikelehen: "code that represents" => "code represents" ?

PR #641, opened 2019-07-19
Add new config file for smoke tests.

allisonbm92:
This change adds the new configuration, which will soon be used by the
runner. This imcreases flexibility by allowing GitHub changes to
immediately affect how the runner operates on CI.

Additionally, this change tweaks the build file to allow testing against
release builds. This will be live once the new runner is deployed.

### DIFF
--- a/smoke-tests/runner.config
+++ b/smoke-tests/runner.config
@@ -1,0 +1,82 @@
+core {
+  # gcloud
+  ## This option is the path to the Gcloud executable. It may be relative or
+  ## absolute. This is optional, because the default is `/usr/bin/gcloud`.
+
+  # gradle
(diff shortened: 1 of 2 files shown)


smoke-tests/runner.config
```diff
@@ -0,0 +1,82 @@
...
+}
+
+firebase {
+  # buildTask
+  ## This option specifies the Gradle task to execute to build Firebase
+  ## libraries. It is mandatory.
+  buildTask assembleAllForSmokeTests
+
+  # localMavenRepo
+  ## This option specifies the path to a local MAven repository. This repository
```
Corrob: s/MAven/Maven

allisonbm92: Done.

PR #643, opened 2019-07-20
Disallow empty document fields

### LINKED ISSUE #632: Firestore crash forcing an uninstallation of the host app: java.lang.IllegalArgumentException: Invalid field path (path.``). Paths must not be empty, begin with '.', end with '.', or contain '..'
- Android Studio version: 3.4.1

- Firebase Component: Firestore

- Component version: 20.1.0

Writing to firestore with "set" and "SetOptions.merge()" using a map with an empty key ("" string) will cause the app to crash. Even worse, it will cause the app to crash immediately every time it is opened, forcing the user to uninstall the app.

#### Steps to reproduce:
```
FirebaseFirestore db = FirebaseFirestore.getInstance();
        Map<String, Object> map1 = new HashMap<>();
        map1.put("", "temp");
       db.collection("col").document("doc").set(map1, SetOptions.merge())
```

#### Stack trace:
java.lang.RuntimeException: Internal error in Firestore (20.1.0).
        at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$5(com.google.firebase:firebase-firestore@@20.1.0:379)
        at com.google.firebase.firestore.util.AsyncQueue$$Lambda$5.run(Unknown Source:2)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
     Caused by: java.lang.RuntimeException: java.lang.IllegalArgumentException: Invalid field path (``). Paths must not be empty, begin with '.', end with '.', or contain '..'
        at com.google.firebase.firestore.util.AsyncQueue.lambda$enqueue$3(com.google.firebase:firebase-firestore@@20.1.0:290)
        at com.google.firebase.firestore.util.AsyncQueue$$Lambda$3.run(Unknown Source:4)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:458)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at com.google.firebase.firestore.util.AsyncQueue$DelayedStartFactory.run(com.google.firebase:firebase-firestore@@20.1.0:205)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.IllegalArgumentException: Invalid field path (``). Paths must not be empty, begin with '.', end with '.', or contain '..'
        at com.google.firebase.firestore.model.FieldPath.fromServerFormat(com.google.firebase:firebase-firestore@@20.1.0:90)
        at com.google.firebase.firestore.remote.RemoteSerializer.decodeDocumentMask(com.google.firebase:firebase-firestore@@20.1.0:537)
        at com.google.firebase.firestore.remote.RemoteSerializer.decodeMutation(com.google.firebase:firebase-firestore@@20.1.0:470)
        at com.google.firebase.firestore.local.LocalSerializer.decodeMutationBatch(com.google.firebase:firebase-firestore@@20.1.0:190)
        at com.google.firebase.firestore.local.SQLiteMutationQueue.decodeInlineMutationBatch(com.google.firebase:firebase-firestore@@20.1.0:437)
        at com.google.firebase.firestore.local.SQLiteMutationQueue.lambda$getNextMutationBatchAfterBatchId$4(com.google.firebase:firebase-firestore@@20.1.0:236)
        at com.google.firebase.firestore.local.SQLiteMutationQueue$$Lambda$5.apply(Unknown Source:4)
        at com.google.firebase.firestore.local.SQLitePersistence$Query.firstValue(com.google.firebase:firebase-firestore@@20.1.0:501)
        at com.google.firebase.firestore.local.SQLiteMutationQueue.getNextMutationBatchAfterBatchId(com.google.firebase:firebase-firestore@@20.1.0:236)
        at com.google.firebase.firestore.local.LocalStore.getNextMutationBatch(com.google.firebase:firebase-firestore@@20.1.0:497)
        at com.google.firebase.firestore.remote.RemoteStore.fillWritePipeline(com.google.firebase:firebase-firestore@@20.1.0:574)
        at com.google.firebase.firestore.remote.RemoteStore.enableNetwork(com.google.firebase:firebase-firestore@@20.1.0:240)
        at com.google.firebase.firestore.remote.RemoteStore.start(com.google.firebase:firebase-firestore@@20.1.0:289)
        at com.google.firebase.firestore.core.FirestoreClient.initialize(com.google.firebase:firebase-firestore@@20.1.0:271)
        at com.google.firebase.firestore.core.FirestoreClient.lambda$new$2(com.google.firebase:firebase-firestore@@20.1.0:117)
        at com.google.firebase.firestore.core.FirestoreClient$$Lambda$2.run(Unknown Source:8)
        at com.google.firebase.firestore.util.AsyncQueue.lambda$enqueue$4(com.google.firebase:firebase-firestore@@20.1.0:311)
        at com.google.firebase.firestore.util.AsyncQueue$$Lambda$4.call(Unknown Source:2)
2019-07-18 16:12:27.304 29220-29220/co.blazepod.blazepod E/AndroidRuntime:     at co

wilhuff:
Fixes #632.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
@@ -22,6 +22,7 @@ import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCol
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testDocument;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
+import static com.google.firebase.firestore.testutil.TestUtil.assertDoesNotThrow;
 import static com.google.firebase.firestore.testutil.TestUtil.expectError;
 import static com.google.firebase.firestore.testutil.TestUtil.map;
 import static java.util.Arrays.asList;
@@ -239,6 +240,9 @@ public class ValidationTest {
 
   @Test
   public void writesMustNotContainReservedFieldNames() {
+    expectWriteSuccess(map("__bar", 1));
+    expectWriteSuccess(map("bar__", 1));
+
     expectWriteError(
         map("__baz__", 1),
         "Invalid data. Document fields cannot begin or end with \"__\" (found in field __baz__)");
@@ -758,6 +762,72 @@ public class ValidationTest {
   }
 
   // Helpers
+
+  /** Performs a write using each write API and makes sure it succeeds. */
+  private static void expectWriteSuccess(Object data) {
+    expectWriteSuccess(data, /*includeSets=*/ true, /*includeUpdates=*/ true);
+  }
+
+  /** Performs a write using each update API and makes sure it succeeds. */
+  private static void expectUpdateSuccess(Map<String, Object> data) {
+    expectWriteSuccess(data, /*includeSets=*/ false, /*includeUpdates=*/ true);
+  }
+
+  /** Performs a write using each set API and makes sure it succeeds. */
+  private static void expectSetSuccess(Object data) {
+    expectWriteSuccess(data, /*includeSets=*/ true, /*includeUpdates=*/ false);
+  }
+
+  /**
+   * Performs a write using each set and/or update API and makes sure it fails with the expected
+   * reason.
+   */
+  private static void expectWriteSuccess(
+      Object data, boolean includeSets, boolean includeUpdates) {
+    DocumentReference ref = testDocument();
+
+    if (includeSets) {
+      assertDoesNotThrow(() -> ref.set(data));
+      assertDoesNotThrow(() -> ref.getFirestore().batch().set(ref, data));
+    }
+
+    if (includeUpdates) {
+      assertTrue("update() only support Maps.", data instanceof Map);
+      assertDoesNotThrow(
+          () -> {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> updateMap = (Map<String, Object>) data;
+            ref.update(updateMap);
+          });
+      assertDoesNotThrow(
+          () -> {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> updateMap = (Map<String, Object>) data;
+            ref.getFirestore().batch().update(ref, updateMap);
+          });
+    }
+
+    waitFor(
+        ref.getFirestore()
+            .runTransaction(
+                (Function<Void>)
+                    transaction -> {
+                      if (includeSets) {
+                        assertDoesNotThrow(() -> transaction.set(ref, data));
+                      }
+                      if (includeUpdates) {
+                        assertTrue("update() only support Maps.", data instanceof Map);
+                        assertDoesNotThrow(
+                            () -> {
+                              @SuppressWarnings("unchecked")
+                              Map<String, Object> updateMap = (Map<String, Object>) data;
+                              transaction.update(ref, updateMap);
+                            });
+                      }
+
+                      return null;
+                    }));
+  }
 
   /** Performs a write using each write API and makes sure it fails with the expected reason. */
   private static void expectWriteError(Object data, String reason) {
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
@@ -241,21 +241,26 @@ public class ValidationTest {
   public void writesMustNotContainReservedFieldNames() {
     expectWriteError(
         map("__baz__", 1),
-        "Invalid data. Document fields cannot begin and end with __ (found in field __baz__)");
+        "Invalid data. Document fields cannot begin or end with \"__\" (found in field __baz__)");
     expectWriteError(
         map("foo", map("__baz__", 1)),
-        "Invalid data. Document fields cannot begin and end with __ (found in field foo.__baz__)");
+        "Invalid data. Document fields cannot begin or end with \"__\" (found in field foo.__baz__)");
     expectWriteError(
         map("__baz__", map("foo", 1)),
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/core/UserData.java
```diff
@@ -303,9 +301,17 @@ private void validatePath() {
...
     private void validatePathSegment(String segment) {
-      if (isWrite() && reservedFieldRegex.matcher(segment).find()) {
-        throw this.createError("Document fields cannot begin and end with __");
+      if (!isWrite()) return;
+
+      if (segment.isEmpty()) {
+        throw this.createError("Document fields must not be empty");
+      }
+
+      if (segment.startsWith(RESERVED) || segment.endsWith(RESERVED)) {
```
mikelehen: I don't understand this change...  Today the clients and backend allow `__x` and `x__` but not `__x__`.  Why are we making the client more restrictive?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/UserData.java
```diff
@@ -303,9 +301,17 @@ private void validatePath() {
       }
     }
 
+    private static final String RESERVED = "__";
+
     private void validatePathSegment(String segment) {
-      if (isWrite() && reservedFieldRegex.matcher(segment).find()) {
-        throw this.createError("Document fields cannot begin and end with __");
+      if (!isWrite()) return;
```
mikelehen: I think we may want to prevent empty field names even for non-writes...?  E.g. `.whereEqualTo("", "value")` should probably fail.  But we don't want to validate the __.*__ regex since `.whereEqualTo("__key__", "x/y/z")` is valid.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/UserData.java
```diff
@@ -303,9 +301,15 @@ private void validatePath() {
       }
     }
 
+    private static final String RESERVED = "__";
```
mikelehen: I don't feel strongly but is there a reason to not use the regex anymore?  JavaScript at least still uses the regex, so this represents a minor porting difference...  and the regex is a bit more self-explanatory, since RESERVED is only reserved if it occurs at the beginning and the end (if you keep it as-is, perhaps add a comment)...

firebase-firestore/src/main/java/com/google/firebase/firestore/core/UserData.java
```diff
@@ -303,9 +301,15 @@ private void validatePath() {
...
+
     private void validatePathSegment(String segment) {
-      if (isWrite() && reservedFieldRegex.matcher(segment).find()) {
-        throw this.createError("Document fields cannot begin and end with __");
+      if (segment.isEmpty()) {
+        throw this.createError("Document fields must not be empty");
+      }
+
+      if (isWrite() && segment.startsWith(RESERVED) && segment.endsWith(RESERVED)) {
+        throw this.createError("Document fields cannot begin or end with \"__\"");
```
mikelehen: I think the original error message ("begin and end with") is more precise, no?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
```diff
@@ -755,6 +763,72 @@ public void queriesUsingInAndDocumentIdMustHaveProperDocumentReferencesInArray()
...
+
+  /** Performs a write using each update API and makes sure it succeeds. */
+  private static void expectUpdateSuccess(Map<String, Object> data) {
+    expectWriteSuccess(data, /*includeSets=*/ false, /*includeUpdates=*/ true);
+  }
+
+  /** Performs a write using each set API and makes sure it succeeds. */
+  private static void expectSetSuccess(Object data) {
+    expectWriteSuccess(data, /*includeSets=*/ true, /*includeUpdates=*/ false);
+  }
```
mikelehen: Are you intentionally including expectUpdateSuccess() and expectSetSuccess() even though they're not used?  If so, that's fine...  just double-checking.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/UserData.java
wilhuff: Javascript uses the regex, but iOS doesn't: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/firebase/firestore/core/user_data.mm#L178

Renamed to match iOS even more closely: `RESERVED_FIELD_DESIGNATOR`

firebase-firestore/src/main/java/com/google/firebase/firestore/core/UserData.java
wilhuff: Yes. This was part of my think-o. Thanks for persisting :-).

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
wilhuff: I added the equivalent facility for checking success that exists for checking failure on the premise that if it were available we'd use it.

rockwotj:
FYI @wilhuff, the backend supports __type__ in certain contexts as a valid field name in Firestore (usually for Legacy Types) not sure if you want to support them in the SDKs, but they are valid.

PR #649, opened 2019-07-22
javadoc fixups

rsgowman:
Addresses the comments in http://b/137376378.

I additionally attempted to find/fix some other similar issues while I was at it. In all cases, I used automated/semi-automated methods to find/fix rather than actually auditing the files, so there's probably many more things that could be addressed.

Generally, I've used `@code` for self-references (i.e. `{@code Blob}` within Blob.java) and `@link` otherwise.) But I haven't attempted to unify this across the codebase.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -26,8 +26,8 @@ import com.google.firebase.firestore.util.Util;
 import javax.annotation.Nullable;
 
 /**
- * A CollectionReference can be used for adding documents, getting document references, and querying
- * for documents (using the methods inherited from Query).
+ * A {@code CollectionReference} can be used for adding documents, getting document references, and
+ * querying for documents (using the methods inherited from {@link Query}).
  *
  * <p><b>Subclassing Note</b>: Cloud Firestore classes are not meant to be subclassed except for use
  * in test mocks. Subclassing is not supported in production code and new SDK releases may break
@@ -56,10 +56,10 @@ public class CollectionReference extends Query {
   }
 
   /**
-   * Gets a DocumentReference to the document that contains this collection. Only subcollections are
-   * contained in a document. For root collections, returns null.
+   * Gets a {@link DocumentReference} to the document that contains this collection. Only
+   * subcollections are contained in a document. For root collections, returns null.
    *
-   * @return The DocumentReference that contains this collection or null if this is a root
+   * @return The {@link DocumentReference} that contains this collection or null if this is a root
    *     collection.
    */
   @Nullable
@@ -86,10 +86,10 @@ public class CollectionReference extends Query {
   }
 
   /**
-   * Returns a DocumentReference pointing to a new document with an auto-generated ID within this
-   * collection.
+   * Returns a {@link DocumentReference} pointing to a new document with an auto-generated ID within
+   * this collection.
    *
-   * @return A DocumentReference pointing to a new document with an auto-generated ID.
+   * @return A {@link DocumentReference} pointing to a new document with an auto-generated ID.
    */
   @NonNull
   @PublicApi
@@ -98,11 +98,11 @@ public class CollectionReference extends Query {
   }
 
   /**
-   * Gets a DocumentReference instance that refers to the document at the specified path within this
-   * collection.
+   * Gets a {@link DocumentReference} instance that refers to the document at the specified path
+   * within this collection.
    *
    * @param documentPath A slash-separated relative path to a document.
-   * @return The DocumentReference instance.
+   * @return The {@link DocumentReference} instance.
    */
   @NonNull
   @PublicApi
@@ -118,7 +118,8 @@ public class CollectionReference extends Query {
    *
    * @param data The data to write to the document (e.g. a Map or a POJO containing the desired
    *     document contents).
-   * @return A Task that will be resolved with the DocumentReference of the newly created document.
+   * @return A Task that will be resolved with the {@link DocumentReference} of the newly created
+   *     document.
    */
   @NonNull
   @PublicApi
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -59,8 +59,8 @@ public class CollectionReference extends Query {
    * Gets a {@code DocumentReference} to the document that contains this collection. Only
    * subcollections are contained in a document. For root collections, returns {@code null}.
    *
-   * @return The {@code DocumentReference} that contains this collection or {@code null} if this is a root
-   *     collection.
+   * @return The {@code DocumentReference} that contains this collection or {@code null} if this is
+   *     a root collection.
    */
   @Nullable
   @PublicApi
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -57,9 +57,9 @@ public class CollectionReference extends Query {
 
   /**
    * Gets a {@code DocumentReference} to the document that contains this collection. Only
-   * subcollections are contained in a document. For root collections, returns null.
+   * subcollections are contained in a document. For root collections, returns {@code null}.
    *
-   * @return The {@code DocumentReference} that contains this collection or null if this is a root
+   * @return The {@code DocumentReference} that contains this collection or {@code null} if this is a root
    *     collection.
    */
   @Nullable
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
@@ -27,7 +27,7 @@ import javax.annotation.Nullable;
 
 /**
  * A {@code CollectionReference} can be used for adding documents, getting document references, and
- * querying for documents (using the methods inherited from {@link Query}).
+ * querying for documents (using the methods inherited from {@code Query}).
  *
  * <p><b>Subclassing Note</b>: Cloud Firestore classes are not meant to be subclassed except for use
  * in test mocks. Subclassing is not supported in production code and new SDK releases may break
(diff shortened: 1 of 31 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
```diff
@@ -56,10 +56,10 @@ public String getId() {
   }
 
   /**
-   * Gets a DocumentReference to the document that contains this collection. Only subcollections are
-   * contained in a document. For root collections, returns null.
+   * Gets a {@link DocumentReference} to the document that contains this collection. Only
+   * subcollections are contained in a document. For root collections, returns null.
    *
-   * @return The DocumentReference that contains this collection or null if this is a root
+   * @return The {@link DocumentReference} that contains this collection or null if this is a root
```
wilhuff: This will add considerable visual noise for no actual benefit. I think we should mostly undo this, or in cases where we're concretely talking about the type and it's important mark those as `{@code}`. `{@link}` should be used to only to refer to types or methods that aren't otherwise implied by the types in the signature.

Look at the current [public method summary](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/CollectionReference#public-method-summary). Also look at the specific doc for getParent (a direct link to which doesn't seem to work). It's already contextually the case that any type mentioned in the signature is linked.

Here's how this [looks in the JDK](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#insert(int,%20java.lang.CharSequence,%20int,%20int)). In this example `CharSequence` is the type of a parameter of the method so it's merely marked as code, not linked.

By contrast, [this method](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#insert(int,%20boolean)) links to `String.valueOf(boolean)` because that link isn't implied by the types of the method.

In JDK javadoc, they frequently go even further and spell words out when referring to the concept behind the type rather than referring to the type name, e.g. [here they spell "character sequence"](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#append(java.lang.CharSequence) rather than `CharSequence`. This doesn't seem to be applied consistently though.

In any case, adding `{@link}` to all non-self types is overkill and distracting.

Adding `{@code}` to everything is also likely to be overkill though I don't mind that as much. Consider changing to just plain words in most cases when the specific type isn't important.

firebase-firestore/src/main/java/com/google/firebase/firestore/CollectionReference.java
```diff
@@ -56,10 +56,10 @@ public String getId() {
   }
 
   /**
-   * Gets a DocumentReference to the document that contains this collection. Only subcollections are
-   * contained in a document. For root collections, returns null.
+   * Gets a {@link DocumentReference} to the document that contains this collection. Only
+   * subcollections are contained in a document. For root collections, returns null.
```
wilhuff: When mentioning the `null` literal we should mark that with `{@code}`. See e.g.:

https://docs.oracle.com/javase/7/docs/api/java/io/File.html#getParentFile()

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentChange.java
```diff
@@ -110,26 +111,26 @@ public QueryDocumentSnapshot getDocument() {
   }
 
   /**
-   * The index of the changed document in the result set immediately prior to this DocumentChange
-   * (i.e. supposing that all prior DocumentChange objects have been applied). Returns -1 for
-   * 'added' events.
+   * The index of the changed document in the result set immediately prior to this {@code
+   * DocumentChange} (i.e. supposing that all prior {@code DocumentChange} objects have been
```
rachelsaunders: maybe "i.e. supposing" to "assuming"

firebase-firestore/src/main/java/com/google/firebase/firestore/FieldValue.java
```diff
@@ -19,35 +19,38 @@
...
 @PublicApi
 public abstract class FieldValue {
 
   FieldValue() {}
 
   /**
-   * Returns the method name (e.g. "FieldValue.delete") that was used to create this FieldValue
-   * instance, for use in error messages, etc.
+   * Returns the method name (e.g. "FieldValue.delete") that was used to create this {@code
+   * FieldValue} instance, for use in error messages, etc.
```
rachelsaunders: "e.g." to "for example,"

firebase-firestore/src/main/java/com/google/firebase/firestore/SnapshotMetadata.java
```diff
@@ -35,10 +35,11 @@
   }
 
   /**
-   * @return true if the snapshot contains the result of local writes (e.g. set() or update() calls)
-   *     that have not yet been committed to the backend. If your listener has opted into metadata
-   *     updates (via `MetadataChanges.INCLUDE`) you will receive another snapshot with
-   *     `hasPendingWrites()` equal to false once the writes have been committed to the backend.
+   * @return true if the snapshot contains the result of local writes (e.g. {@code set()} or {@code
```
rachelsaunders: "e.g." to "for example,"

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLitePersistence.java
```diff
@@ -144,13 +144,13 @@ public void start() {
...
-              + " Keep in mind that multi-process Android apps execute the code in your"
+          "Failed to gain exclusive lock to the Cloud Firestore client's offline persistence. This"
+              + " generally means you are using Cloud Firestore from multiple processes in your"
+              + " app. Keep in mind that multi-process Android apps execute the code in your"
               + " Application class in all processes, so you may need to avoid initializing"
-              + " Firestore in your Application class. If you are intentionally using Firestore"
-              + " from multiple processes, you can only enable offline persistence (i.e. call"
-              + " setPersistenceEnabled(true)) in one of them.",
+              + " Cloud Firestore in your Application class. If you are intentionally using Cloud"
+              + " Firestore from multiple processes, you can only enable offline persistence (i.e. "
```
rachelsaunders: "i.e." to "that is,"

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -63,9 +63,9 @@
    * part of their app's dependencies.
    */
   static final String SSL_DEPENDENCY_ERROR_MESSAGE =
-      "The Firestore SDK failed to establish a secure connection. This is likely a problem with "
-          + "your app, rather than with Firestore itself. See https://bit.ly/2XFpdma for "
-          + "instructions on how to enable TLS on Android 4.x devices.";
+      "The Cloud Firestore SDK failed to establish a secure connection. This is likely a problem "
```
wilhuff: This is wrong but the correct text is probably too wordy. This isn't a "Cloud Firestore SDK", this is the "Cloud Firestore for Firebase SDK" :-(. In the other error messages we refer to things like this "Cloud Firestore client's offline persistence". Maybe s/SDK/client/?

firebase-firestore/src/main/java/com/google/firebase/firestore/MetadataChanges.java
```diff
@@ -17,7 +17,7 @@
 import com.google.firebase.annotations.PublicApi;
 
 /**
- * Indicates whether metadata-only changes (i.e. only {@code DocumentSnapshot.getMetadata()} or
+ * Indicates whether metadata-only changes (i.e.&nbsp;only {@code DocumentSnapshot.getMetadata()} or
```
wilhuff: Was this intended?

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentChange.java
```diff
@@ -110,26 +111,26 @@ public QueryDocumentSnapshot getDocument() {
...
+   * DocumentChange} (assuming that all prior {@code DocumentChange} objects and the current {@code
+   * DocumentChange} object have been applied). Returns -1 for 'removed' events.
    */
   @PublicApi
   public int getNewIndex() {
     return newIndex;
   }
 
-  /** Creates the list of DocumentChanges from a ViewSnapshot. */
+  /** Creates the list of {@code DocumentChange}s from a {@code ViewSnapshot}. */
```
wilhuff: pluralizing with a "s" outside the `{@code block}` ends up rendering with mixed fonts in a single word and is ugly.

Consider rephrasing to "list of `{@code DocumentChange}` instances" (or "objects") or "list of document changes".

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -419,17 +417,17 @@ public static void setLoggingEnabled(boolean loggingEnabled) {
   /**
    * Clears the persistent storage, including pending writes and cached documents.
    *
-   * <p>Must be called while the FirebaseFirestore instance is not started (after the app is
+   * <p>Must be called while the {@code FirebaseFirestore} instance is not started (after the app is
    * shutdown or when the app is first initialized). On startup, this method must be called before
-   * other methods (other than <code>setFirestoreSettings()</code>). If the FirebaseFirestore
+   * other methods (other than {@link #setFirestoreSettings()}). If the {@code FirebaseFirestore}
    * instance is still running, the <code>Task</code> will fail with an error code of <code>
```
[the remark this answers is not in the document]
rsgowman: done.

firebase-firestore/src/main/java/com/google/firebase/firestore/MetadataChanges.java
rsgowman: Yes. Without this, javadoc cuts the summary fragment after the "i.e." (Javadoc assumes the first period followed by a space ends the summary fragment.) There aren't any good workarounds; adding nbsp tricks javadoc into not thinking there's a space there.

But Rachel's suggestions elsewhere provide another possibility; just spell it out in english. Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentChange.java
rsgowman: Used 'document changes' (since I have to address this one individually.)

PR #650, opened 2019-07-23
Removed unused TestHelper methods

schmidt-sebastian:
This PR removes a bunch of code from the duplicated TestHelper classes that is currently not used in Kotlin. I am not super attached to this PR (it does cause our TestHelper classes to deviate quite a bit), but all my recents PRs had compile time failures in these TestHelper classes that I only caught during the CI run.


schmidt-sebastian:
@vkryachko PTAL

PR #652, opened 2019-07-23
Index-Free (5/6):  Add Index-free Query Engine

schmidt-sebastian:
This is the fifth of six PR for "index-free" queries.

This PR adds a new Query Engine that uses the optimization of index-free queries and first loads the documents from the RemoteDocumentCache that the backend considers matching.

The Query Engine is not used/activated.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
@@ -1,0 +1,149 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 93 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
```diff
@@ -0,0 +1,78 @@
...
+  private final LocalDocumentsView localDocumentsView;
+  private final QueryCache queryCache;
+
+  public IndexFreeQueryEngine(LocalDocumentsView localDocumentsView, QueryCache queryCache) {
+    this.localDocumentsView = localDocumentsView;
+    this.queryCache = queryCache;
+  }
+
+  @Override
+  public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
```
schmidt-sebastian: Note for future me: We should probably update this to parse in the set of RemoteKeys as we are otherwise fetching it twice (once in executeQuery, once here).

PR #659, opened 2019-07-24
Adding http client to call fis backend service

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -0,0 +1,191 @@
...
+
+    NETWORK_ERROR,
+
+    SERVER_ERROR,
+
+    UNAUTHORIZED,
+  }
+
+  @NonNull
+  public Code createFirebaseInstallation(
```
ankitaj224: Here we would have to eventually return the response rather than just the Code. This is the initial structure.

ankitaj224: @diwu-arete  does this way of reading response seem right? cs/"readResponse(HttpsURLConnection conn)"

Another way could be to use OkHttpClient: firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java

diwu-arete: You code actually doesn't read any content from the response, instead, only reads the response code.
So I recommend (Vlad recommended me, hah) to just use the java class HttpsURLConnection

Although it seems that you write more code by using HttpsURLConnection, but for SDK development, a very important goal is to make the SDK smaller in size, I believe okhttpclient(or some other 3P libraries) will make the SDK much bigger, because they have much more functions than our simple use case.

ankitaj224: PTAL. Let me know if this is the right way. Thanks.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -78,7 +78,7 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
   @NonNull
   @Override
   public Task<InstallationTokenResult> getAuthToken(boolean forceRefresh) {
-    return Tasks.forResult(InstallationTokenResult.create("dummy_auth_token", 1000l));
+    return Tasks.forResult(InstallationTokenResult.builder().build());
```
diwu-arete: Can you make the two members nonnull(token and expiration) in InstallationTokenResult? So it'll not be okay to build a InstallationTokenResult with token or expiration

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -0,0 +1,246 @@
...
+
+      int httpResponseCode = httpsURLConnection.getResponseCode();
+      switch (httpResponseCode) {
+        case 200:
+          return readCreateResponse(httpsURLConnection);
+        case 401:
+          throw new FirebaseException("The request did not have the required credentials.");
+        default:
+          throw new FirebaseException("There was an internal server error.");
+      }
```
diwu-arete: Why does deleteFirebaseInstallation return a error code, but createFirebaseInstallation throws exceptions?

I think they should be consistent

diwu-arete: For example, both throw 
FirebaseInstallationNetworkException extends FirebaseException {
    Code code;
    FirebaseInstallationNetworkException(Code code, String message, ....) {
    }
}

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -0,0 +1,246 @@
...
+            installationTokenResult.setTokenExpirationTimestampMillis(reader.nextLong());
+          } else {
+            reader.skipValue();
+          }
+        }
+        builder.setAuthToken(installationTokenResult.build());
+        reader.endObject();
+      } else {
+        reader.skipValue();
+      }
```
diwu-arete: If you want http to respond in json format, you need to add this header to your request.
https://stackoverflow.com/questions/43209924/rest-api-use-the-accept-application-json-http-header

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Added the nonull annotation, but that doesnt block the default instance creation. Leaving it as is for now.

PR #667, opened 2019-07-26
Increase Remote Config's default timeout.

allisonbm92:
The default timeout of 5 seconds is way too short. The network is not
guaranteed to be in a good state, and this may cause apps to crash
unexpectedly. This is the source of some flakiness in tests as well.

This change increases the timeout to 30 seconds. Updates are infrequent
and run in the background, so this should not be a problem.

Edit: The timeout is now changed to 60 seconds to match the iOS implementation.


miraziz:
The iOS network timeout is 60 seconds, so if we're going to increase it, it makes sense to raise it to 60.

Ideally, the tests should set a higher timeout manually, but I don't see a major reason not raise the default value as well.

PR #670, opened 2019-07-29
Fix broken documentation found while reviewing C# docs.

wilhuff:


wilhuff:
PS: this is a comment-only change!

PR #680, opened 2019-08-01
Add support for /?ns= query param in Database URL

schmidt-sebastian:
We need to specify the namespace to connect to the Firestore Emulator from an Android Emulator.

With this change, a connection to the Emulator can be made by specifying a database URL such as "http://10.0.2.2:9000/?ns=foo". I ended up using Android URI parser over Java's since it has built in support for query param.

The new unit tests (some of them are old and moved from the integration test suite) pass with bold the old and the new parser.

Note that this PR drops the path segment encoding since I am no longer roundtripping through URL to get the database path.

### DIFF
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 19.1.0
+- [feature] The SDK adds support for the Firebase Database Emulator. To connect
+  to the emulator, specify "http://<emulatorHost>/?ns=<projectId>" as your
+  Database URL (via `FirebaseDatabase.getInstance(String)`).
+  Note that if you are running the Database Emulator on "localhost" and
+  connecting from an app that is running inside an Android Emulator, the
+  emulator host will be "10.0.2.2" followed by its port.
+
 # 18.0.1
 - [changed] The SDK now reports the correct version number (via
   `FirebaseDatabase.getSdkVersion()`).
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -1,7 +1,7 @@
 # 19.1.0
-- [feature] The SDK added support for the Firebase Database Emulator. To
-  connect to the emulator, specify "http://<emulatorHost>/?ns=<projectId>" as
-  your Database URL (via `FirebaseDatabase.getInstance(String)`).
+- [feature] Added support for the Firebase Database Emulator. To connect to
+  the emulator, specify "http://<emulatorHost>/?ns=<projectId>" as your
+  Database URL (via `FirebaseDatabase.getInstance(String)`).
   Note that if you are running the Database Emulator on "localhost" and
   connecting from an app that is running inside an Android Emulator, the
   emulator host will be "10.0.2.2" followed by its port.
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -1,7 +1,7 @@
 # 19.1.0
-- [feature] The SDK adds support for the Firebase Database Emulator. To connect
-  to the emulator, specify "http://<emulatorHost>/?ns=<projectId>" as your
-  Database URL (via `FirebaseDatabase.getInstance(String)`).
+- [feature] The SDK added support for the Firebase Database Emulator. To
+  connect to the emulator, specify "http://<emulatorHost>/?ns=<projectId>" as
+  your Database URL (via `FirebaseDatabase.getInstance(String)`).
   Note that if you are running the Database Emulator on "localhost" and
   connecting from an app that is running inside an Android Emulator, the
   emulator host will be "10.0.2.2" followed by its port.
--- a/firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
+++ b/firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.database.core.utilities;
 
+import android.net.Uri;
 import android.util.Base64;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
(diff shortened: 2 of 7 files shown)


schmidt-sebastian:
I also ported iOS's amazing decoding capabilities, which means that we now also treat "repo.foo.bar/a%b&c@d/space: /non-ascii:ø" and "repo.foo.bar/a%25b%26c%40d/space%3A%20/non-ascii%3A%C3%B8" the same.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -70,36 +44,68 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
       if (port != -1) {
         repoInfo.secure = scheme.equals("https");
         repoInfo.host += ":" + port;
+      } else if (repoInfo.host.equals("localhost")) {
```
yuchenshi: Why this change? Also what if people use `127.0.0.1` (a perfectly valid thing to do) instead of `localhost`?

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -70,36 +44,68 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
...
+        repoInfo.secure = scheme.equals("https");
       } else {
         repoInfo.secure = true;
       }
-      String[] parts = repoInfo.host.split("\\.", -1);
 
-      repoInfo.namespace = parts[0].toLowerCase();
+      String namespaceParam = uri.getQueryParameter("ns");
+      String[] parts = uri.getHost().split("\\.", -1);
+      if (parts.length == 3 || namespaceParam == null) {
```
yuchenshi: According to the internal doc, this should really test for `.firebaseio.com` instead of number of parts to align with other SDKs. Plus this probably will have some false positives for `my.dev.local` if people wants to put that in their hosts file.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
schmidt-sebastian: We can special-case `127.0.0.1` as well if you feel we should. This code just makes sure that for localhost we don't use SSL even if no port is specified. Without adding support for `127.0.0.1`, users need to specify `127.0.0.1:9000` (so really this only becomes an issue if the emulator runs on port 80 or 443)

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
schmidt-sebastian: The original code is very forgiving (in my opinion much too forgiving), but making it less forgiving might break some users.

yuchenshi: Okay I think this is fine.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
yuchenshi: I think the existing behavior when port is present is fine. For other cases, it should be tied to the same condition as below -- if we decide to use namespace from host, then we force SSL (same behavior as before), or we use `scheme.equals("https")` if we decide to use ns from query. In this way all emulator-specific behavior is at least triggered by the same conditions and we don't have to think about such edge cases.

schmidt-sebastian: As stated in the iOS port, this special case is now a thing of the past.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -70,36 +44,68 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
...
+      } else if (repoInfo.host.equals("localhost")) {
+        repoInfo.secure = scheme.equals("https");
       } else {
         repoInfo.secure = true;
       }
-      String[] parts = repoInfo.host.split("\\.", -1);
 
-      repoInfo.namespace = parts[0].toLowerCase();
+      String namespaceParam = uri.getQueryParameter("ns");
+      String[] parts = uri.getHost().split("\\.", -1);
```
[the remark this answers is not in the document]
schmidt-sebastian: You are correct - but it seems to be generally frowned upon to rely on the default behavior (https://g3doc.corp.google.com/devtools/staticanalysis/errorprone/g3doc/bugpattern/StringSplitter.md?cl=head). I left the code as is.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -70,36 +44,68 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
...
+      if (parts.length == 3 || namespaceParam == null) {
+        // We use the subdomain from the provided host even when a query parameter is provided to
+        // maintain backward compatibility.
```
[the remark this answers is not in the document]
schmidt-sebastian: Yeah. That what the general idea. I scrapped this to simplify the code a bit.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -70,36 +44,68 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
...
-      ArrayList<String> segs = new ArrayList<String>();
-      for (int i = 0; i < src.length(); i += maxFrameSize) {
-        int end = Math.min(i + maxFrameSize, src.length());
-        String seg = src.substring(i, end);
-        segs.add(seg);
+  /**
+   * Extracts the path string from the original URL without changing the encoding (unlike
+   * Uri.getPath()).
+   */
+  private static String extractPathString(String originalUrl) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

schmidt-sebastian:
Coincidentally, I changed my mind and left the original URL decoding behavior as is.

firebase-database/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 19.1.0
+- [feature] The SDK added support for the Firebase Database Emulator. To
```
mikelehen: I still say this should be "Added support ..." to match our prior convention...  (else it sounds like the SDK added the emulator support to itself).

firebase-database/src/test/java/com/google/firebase/database/core/utilities/ParseUrlTest.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Test
+  public void testUrlParsingSslDetection() throws DatabaseException {
+    // Hosts with custom ports are considered non-secure
+    ParsedUrl parsed = Utilities.parseUrl("http://gsoltis.fblocal.com:9000");
+    assertFalse(parsed.repoInfo.secure);
+
+    // Hosts with the default ports are considered secure
+    parsed = Utilities.parseUrl("http://gsoltis.firebaseio.comr");
```
yuchenshi:
```suggestion
    parsed = Utilities.parseUrl("http://gsoltis.firebaseio.com");
```

PR #686, opened 2019-08-05
Index-Free: Using readTime instead of updateTime

schmidt-sebastian:
This PR updates index-free queries to use readTime instead of updateTime, which allows us to pick up documents that matched a given query at the Query's snapshot version but were not part of the query's limit (see spec test: "Resumed limit queries use updated documents")

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
@@ -124,7 +124,7 @@ final class MemoryRemoteDocumentCache implements RemoteDocumentCache {
   }
 
   Iterable<MaybeDocument> getDocuments() {
-    return new DocumentIterator();
+    return new DocumentIterable();
   }
 
   /**
@@ -144,15 +144,17 @@ final class MemoryRemoteDocumentCache implements RemoteDocumentCache {
 
   long getByteSize(LocalSerializer serializer) {
     long count = 0;
-    for (MaybeDocument doc : new DocumentIterator()) {
+    for (MaybeDocument doc : new DocumentIterable()) {
       count += getKeySize(doc.getKey());
       count += serializer.encodeMaybeDocument(doc).getSerializedSize();
     }
     return count;
   }
 
-  /** An Iterator that iterates the current set of documents in the RemoteDocumentCache. */
-  private class DocumentIterator implements Iterable<MaybeDocument> {
+  /**
+   * A proxy that exposes an iterator over the current set of documents in the RemoteDocumentCache.
+   */
+  private class DocumentIterable implements Iterable<MaybeDocument> {
     @NonNull
     @Override
     public Iterator<MaybeDocument> iterator() {
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
@@ -15,9 +15,10 @@
 package com.google.firebase.firestore.local;
 
 import static com.google.firebase.firestore.model.DocumentCollections.emptyDocumentMap;
-import static com.google.firebase.firestore.model.DocumentCollections.emptyMaybeDocumentMap;
 import static com.google.firebase.firestore.util.Assert.hardAssert;
 
+import android.util.Pair;
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.firebase.database.collection.ImmutableSortedMap;
 import com.google.firebase.firestore.core.Query;
@@ -25,6 +26,7 @@ import com.google.firebase.firestore.model.Document;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.MaybeDocument;
 import com.google.firebase.firestore.model.ResourcePath;
+import com.google.firebase.firestore.model.SnapshotVersion;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -15,38 +15,40 @@
...
 import java.util.Iterator;
 import java.util.Map;
 
 /** In-memory cache of remote documents. */
 final class MemoryRemoteDocumentCache implements RemoteDocumentCache {
 
-  /** Underlying cache of documents. */
-  private ImmutableSortedMap<DocumentKey, MaybeDocument> docs;
+  /** Underlying cache of documents and their read times. */
+  private ImmutableSortedMap<DocumentKey, Pair<MaybeDocument, SnapshotVersion>> docs;
```
schmidt-sebastian: Note that the real only benefit of the changes in this file (and the additional memory-overhead) is that we can run the LocalStore tests with IndexFree queries both against the MemoryRemoteDocumentCache and the SQLiteRemoteDocumentCache. We need to match the filter logic to be able to compare row read counts.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/RemoteDocumentCache.java
```diff
@@ -40,8 +41,9 @@
    * for the key, it will be replaced.
    *
    * @param maybeDocument A Document or NoDocument to put in the cache.
+   * @param readTime The time at which the document was read or committed.
    */
-  void add(MaybeDocument maybeDocument);
+  void add(MaybeDocument maybeDocument, SnapshotVersion readTime);
```
wilhuff: Dumb question: why not make `readTime` a part of the `MaybeDocument`?

schmidt-sebastian: That was the road I went down on at first - it generates a significant amount of changes to plumb through a value that is only ever read directly at source.

wilhuff: I'm sort of surprised there are that many changes. Did you happen to commit it so I can see?

For example: all our tests could use the same snapshot for both versions by default and only specify a second timestamp when the difference really matters. That means nearly all tests would be unaffected.

Oh, maybe the issue is that the read_time comes from the TargetChange and is not intrinsically a part of the document?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/RemoteDocumentCacheTestCase.java
```diff
@@ -134,7 +136,7 @@ public void testSetAndReadLotsOfDocuments() {
   public void testSetAndReadDeletedDocument() {
     String path = "a/b";
     NoDocument deletedDoc = deletedDoc(path, 42);
-    add(deletedDoc);
+    add(deletedDoc, version(42));
```
wilhuff: All of these tests use identical versions for the update_time and read_time which means we could have gotten these reversed and won't see it. Could you add a test that verifies read time doesn't get crossed with update time?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -89,12 +92,13 @@ public MaybeDocument get(DocumentKey key) {
     // we need to match the query against.
     ResourcePath queryPath = query.getPath();
     DocumentKey prefix = DocumentKey.fromPath(queryPath.append(""));
-    Iterator<Map.Entry<DocumentKey, MaybeDocument>> iterator = docs.iteratorFrom(prefix);
+    Iterator<Map.Entry<DocumentKey, Pair<MaybeDocument, SnapshotVersion>>> iterator =
```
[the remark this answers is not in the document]
schmidt-sebastian: Coming with Java 10 (by which I "came with Java 10"):

```
var iterator = docs.iteratorFrom(prefix)
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -140,10 +144,31 @@ private static long getKeySize(DocumentKey key) {
...
-      count += serializer.encodeMaybeDocument(entry.getValue()).getSerializedSize();
+    for (MaybeDocument doc : new DocumentIterator()) {
+      count += getKeySize(doc.getKey());
+      count += serializer.encodeMaybeDocument(doc).getSerializedSize();
     }
     return count;
   }
+
+  /** An Iterator that iterates the current set of documents in the RemoteDocumentCache. */
+  private class DocumentIterator implements Iterable<MaybeDocument> {
```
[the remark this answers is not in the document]
schmidt-sebastian: Changed to: 

```
  /** A proxy that exposes an iterator over the current set of documents in the RemoteDocumentCache. */
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/RemoteDocumentCache.java
schmidt-sebastian: There are at least two additional challenges:
- If I add it to MaybeDocument, I have to change it for Document, UnknownDocument and NoDocument. At least for NoDocuments it is throwaway work as these are never written to the cache.
- MaybeDoocument currently uses the generic term `version`. This works well with all subtypes, but it doesn't quite work if introduce another kind of version.

I never actually got as far as to see how we would construct the individual documents. By exposing the read time only as an input to the RemoteDocumentCache there are only two callsites that need changing, which won me over.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/RemoteDocumentCacheTestCase.java
schmidt-sebastian: I think it will become easier to add this test when https://github.com/firebase/firebase-android-sdk/pull/617 lands (which should be the next PR). I added a TODO for now and will resolve it when we gain the ability to exclude documents by read time.

PR #689, opened 2019-08-06
awaitPendingWrites initial revision

Hui-Wu:
Added 'waitForPendingWrites' to firestore.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
@@ -58,6 +58,7 @@ import java.util.concurrent.TimeoutException;
 class MockCredentialsProvider extends EmptyCredentialsProvider {
 
   private static MockCredentialsProvider instance;
+  private Listener<User> listener;
 
   public static MockCredentialsProvider instance() {
     if (MockCredentialsProvider.instance == null) {
@@ -66,9 +67,7 @@ class MockCredentialsProvider extends EmptyCredentialsProvider {
     return MockCredentialsProvider.instance;
   }
 
-  private MockCredentialsProvider() {
-    super();
-  }
+  private MockCredentialsProvider() {}
 
   @Override
   public void setChangeListener(Listener<User> changeListener) {
@@ -79,8 +78,6 @@ class MockCredentialsProvider extends EmptyCredentialsProvider {
   public void changeUserTo(User user) {
     listener.onValue(user);
   }
-
-  private Listener<User> listener;
 }
 
 /** A set of helper methods for tests */
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
@@ -36,11 +36,13 @@ import com.google.firebase.firestore.ListenerRegistration;
 import com.google.firebase.firestore.MetadataChanges;
 import com.google.firebase.firestore.QuerySnapshot;
 import com.google.firebase.firestore.auth.EmptyCredentialsProvider;
+import com.google.firebase.firestore.auth.User;
 import com.google.firebase.firestore.core.DatabaseInfo;
 import com.google.firebase.firestore.local.Persistence;
 import com.google.firebase.firestore.model.DatabaseId;
 import com.google.firebase.firestore.testutil.provider.FirestoreProvider;
 import com.google.firebase.firestore.util.AsyncQueue;
+import com.google.firebase.firestore.util.Listener;
 import com.google.firebase.firestore.util.Logger;
 import com.google.firebase.firestore.util.Logger.Level;
 import java.util.ArrayList;
@@ -52,6 +54,34 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+
+class MockCredentialsProvider extends EmptyCredentialsProvider {
+
+  private static MockCredentialsProvider instance;
+
+  public static MockCredentialsProvider instance() {
+    if (MockCredentialsProvider.instance == null) {
+      MockCredentialsProvider.instance = new MockCredentialsProvider();
+    }
+    return MockCredentialsProvider.instance;
+  }
+
+  private MockCredentialsProvider() {
+    super();
+  }
+
+  @Override
+  public void setChangeListener(Listener<User> changeListener) {
+    super.setChangeListener(changeListener);
+    this.listener = changeListener;
+  }
+
+  public void changeUserTo(User user) {
+    listener.onValue(user);
+  }
+
+  private Listener<User> listener;
+}
 
 /** A set of helper methods for tests */
 public class IntegrationTestUtil {
@@ -239,7 +269,7 @@ public class IntegrationTestUtil {
             context,
             databaseId,
             persistenceKey,
-            new EmptyCredentialsProvider(),
+            MockCredentialsProvider.instance(),
             asyncQueue,
             /*firebaseApp=*/ null,
             /*instanceRegistry=*/ (dbId) -> {});
@@ -408,5 +438,9 @@ public class IntegrationTestUtil {
 
   public static boolean isRunningAgainstEmulator() {
     return CONNECT_TO_EMULATOR;
+  }
+
+  public static void testChangeUserTo(User user) {
+    MockCredentialsProvider.instance().changeUserTo(user);
   }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -363,10 +363,12 @@ public class FirebaseFirestore {
   }
 
   /**
-   * Wait until all pending writes existed at the time of calling are sent to the backend.
+   * Waits until all pending writes that existed at the time of calling have been successfully
+   * written to the server.
    *
-   * @return A {@link Task} which resolves when all pending writes are sent to the backend. If there
-   *     is a Firebase user change, the return {@link Task} will resolve to an exception.
+   * @return A {@code Task} which resolves when all pending writes are written to the backend. If
+   *     there is a Firebase user change, the returned {@code Task} will resolve to an exception. If
+   *     the client is offline, the returned {@code Task} will not resolve.
    */
   Task<Void> waitForPendingWrites() {
     return client.waitForPendingWrites();
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -369,9 +369,9 @@ public class FirebaseFirestore {
    * <p>The returned Task completes immediately if there are no outstanding writes. Otherwise, the
    * Task waits for all previously issued writes (including those written in a previous app
    * session), but it does not wait for writes that were added after the method is called. If you
-   * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
+   * wish to wait for additional writes, you have to call {@code waitForPendingWrites()} again.
    *
-   * <p>Any outstanding `waitForPendingWrites()` Tasks are cancelled during user changes.
+   * <p>Any outstanding {@code waitForPendingWrites()} Tasks are cancelled during user changes.
    *
    * @return A {@code Task} which resolves when all currently pending writes have been acknowledged
    *     by the backend.
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -363,12 +363,18 @@ public class FirebaseFirestore {
   }
 
   /**
-   * Waits until all pending writes that existed at the time of calling have been successfully
-   * written to the server.
+   * Waits until all currently pending writes for the active user have been acknowledged by the
+   * backend.
    *
-   * @return A {@code Task} which resolves when all pending writes are written to the backend. If
-   *     there is a Firebase user change, the returned {@code Task} will resolve to an exception. If
-   *     the client is offline, the returned {@code Task} will not resolve.
+   * <p>The returned Task completes immediately if there are no outstanding writes. Otherwise, the
+   * Task waits for all previously issued writes (including those written in a previous app
+   * session), but it does not wait for writes that were added after the method is called. If you
+   * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
+   *
+   * <p>Any outstanding `waitForPendingWrites()` Tasks are cancelled during user changes.
+   *
+   * @return A {@code Task} which resolves when all currently pending writes have been acknowledged
+   *     by the backend.
    */
   Task<Void> waitForPendingWrites() {
     return client.waitForPendingWrites();
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -370,8 +370,8 @@ public class FirebaseFirestore {
    *
    * @return A {@link Task} which resolves when all pending writes are acknowledged by the server.
    */
-  Task<Void> awaitPendingWrites() {
-    return client.awaitPendingWrites();
+  Task<Void> waitForPendingWrites() {
+    return client.waitForPendingWrites();
   }
 
   @VisibleForTesting
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -363,12 +363,10 @@ public class FirebaseFirestore {
   }
 
   /**
-   * Wait for server acknowledgement for all pending writes existing at the time of calling this
-   * method.
+   * Wait until all pending writes existed at the time of calling are sent to the backend.
    *
-   * <p>Both acceptance and rejection count as server acknowledgement.
-   *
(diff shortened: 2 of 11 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -282,6 +282,14 @@ public LocalWriteResult writeLocally(List<Mutation> mutations) {
         });
   }
 
+  /**
+   * Returns the largest (latest) batch id in mutation queue that is pending server response.
+   * Returns 0 if the queue is empty.
```
schmidt-sebastian: Can we return -1 to clearly differentiate an empty queue and the case where we are writing for the first write? Returning `0` requires some internal knowledge about how we assign batch IDs. What's even better is that we have a constant for this: https://github.com/firebase/firebase-android-sdk/blob/69f29d43f4b4e14bed7f5c44a1dc3883c11cce57/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationBatch.java#L41

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
```diff
@@ -187,6 +187,11 @@ public MutationBatch getNextMutationBatchAfterBatchId(int batchId) {
     return queue.size() > index ? queue.get(index) : null;
   }
 
+  @Override
+  public int getLargestUnacknowledgedBatchId() {
+    return queue.size() == 0 ? 0 : nextBatchId - 1;
```
schmidt-sebastian: Google style guide generally recommends using isEmpty(), since it always runs in O(1).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteMutationQueue.java
```diff
@@ -249,6 +249,16 @@ public MutationBatch getNextMutationBatchAfterBatchId(int batchId) {
         .firstValue(row -> decodeInlineMutationBatch(row.getInt(0), row.getBlob(1)));
   }
 
+  @Override
+  public int getLargestUnacknowledgedBatchId() {
+    if (isEmpty()) {
```
schmidt-sebastian: Nit: In case the mutation queue is not empty, you are running a very similar query twice.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteMutationQueue.java
```diff
@@ -249,6 +249,16 @@ public MutationBatch getNextMutationBatchAfterBatchId(int batchId) {
...
+      return 0;
+    }
+    return db.query("SELECT MAX(batch_id) FROM mutations " + "WHERE uid = ?")
```
schmidt-sebastian: Nit: s/" + "//

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -1143,4 +1143,21 @@ public void testHandlesPatchMutationWithTransformThenRemoteEvent() {
     assertChanged(doc("foo/bar", 1, map("sum", 1), Document.DocumentState.LOCAL_MUTATIONS));
     assertContains(doc("foo/bar", 1, map("sum", 1), Document.DocumentState.LOCAL_MUTATIONS));
   }
+
+  @Test
+  public void testGetHighestUnacknowledgedBatchIdReturnsExpectedResult() {
```
schmidt-sebastian: Nit: Maybe remove `ReturnsExpectedResult`, since it doesn't add too much to the descriptiveness.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -362,6 +362,18 @@ public WriteBatch batch() {
     return shutdownInternal();
   }
 
+  /**
+   * Wait for server acknowledgement for all pending writes existing at the time of calling this
+   * method.
+   *
+   * <p>Both acceptance and rejection count as server acknowledgement.
+   *
+   * @return A {@link Task} which resolves when all pending writes are acknowledged by the server.
```
schmidt-sebastian: We likely want to update the comment a bit:

- We should use concepts that are more widely known ("server acknowledgement for all pending writes" -> "when all writes have been sent to the backend").
- We should mention how this behaves during a user change.
- We should mention that this works even after client restart.

If you want to, we can tweak this comment in the chatroom.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
Hui-Wu: When would you need this information, when is checking if it's -1 or 0 helpful?

If we don't know a case when this will be useful, i'd like to keep the abstraction simple.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -282,6 +282,14 @@ public LocalWriteResult writeLocally(List<Mutation> mutations) {
...
+   */
+  public int getHighestUnacknowledgedBatchId() {
```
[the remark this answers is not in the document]
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteMutationQueue.java
Hui-Wu: Fixed.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteMutationQueue.java
Hui-Wu: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
Hui-Wu: Yeah. I made some change, but not enough. Especially i don't understand your comment about client restart, and how this still works with a restart. Let's chat tmr.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -565,6 +626,7 @@ public void handleCredentialChange(User user) {
       // Notify local store and emit any resulting events from swapping out the mutation queue.
       ImmutableSortedMap<DocumentKey, MaybeDocument> changes = localStore.handleUserChange(user);
       emitNewSnapsAndNotifyLocalStore(changes, /*remoteEvent=*/ null);
+      failOutstandingPendingWritesAwaitingTasks();
```
schmidt-sebastian: Ideally, we would want to fail these Tasks before we raise the new snapshots for the newly signed in user. That way, we handle the callbacks for the old user before invoking any callbacks that affect the new user.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
schmidt-sebastian: One of the reasons that this is a heavily requested feature is that it offers users the ability to monitor the state of writes that were issued during an earlier app session.If a user issues a write and then closes the app before the write is committed, it is not very straightforward to figure out whether the write made it to the backend once the app is opened again. `drain/awaitPendingWrites` solves this problem nicely.

Just for reference: This can be done today by either issuing a dummy write and waiting for its completion (one would have to know that we commit writes sequentially) or by issuing gets for all documents that were written and checking the `hasPendingWrites` flags.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -362,6 +362,16 @@ public WriteBatch batch() {
     return shutdownInternal();
   }
 
+  /**
+   * Wait until all pending writes existed at the time of calling are sent to the backend.
```
wilhuff: A bunch of comments on this:
  * Phrase comments with an implicit "this method" at the beginning
  * missing "that" between "writes existed"
  * This doesn't just wait until writes are sent, but also waits for them to complete

> [This method ...] *Waits* until all pending writes that existed at the time of calling are have been successfully written to the server. The task will will not resolve while the client is offline.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -362,6 +362,16 @@ public WriteBatch batch() {
...
+   *
+   * @return A {@link Task} which resolves when all pending writes are sent to the backend. If there
```
wilhuff: Don't `@link` items that are already in the signature of the method. Use `@link` for anything related that's not part of the signature.

Use `{@code Task}` to highlight that it's a type without the extra visual noise that comes from an actual link.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -362,6 +362,18 @@ public WriteBatch batch() {
...
 
+  /**
+   * Waits until all pending writes that existed at the time of calling have been successfully
+   * written to the server.
+   *
+   * @return A {@code Task} which resolves when all pending writes are written to the backend. If
+   *     there is a Firebase user change, the returned {@code Task} will resolve to an exception. If
+   *     the client is offline, the returned {@code Task} will not resolve.
+   */
+  Task<Void> waitForPendingWrites() {
```
schmidt-sebastian: I still think we should move this comment to a doc (it could be the API proposal doc) and do some collaborative editing: 

- `pending writes that existed at the time of calling` could be clearer.
- `written to the server` is an unusual way of phrasing the right thing.
- I think we should move the comment on the user change to a separate paragraph and move it out of the `@return` section
- Tasks on Android don't resolve, they complete.

@mikelehen usually has a ton of good feedback on these comments, and his feedback tends to be much more valuable than what I can provide. 

Thanks!

wilhuff: The API proposal doc would be a good place to collaborate on this.

Note that the "written to the server" language is used in the iOS documentation already which is why I suggested it: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Public/FIRDocumentReference.h#L117. I'm definitely open to finding a better way to succinctly describe this.

Hui-Wu:
> Question: if there are no pending writes, will the task resolve while offline?
It will resolve in that case. That sounds right to me.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
Hui-Wu: Using Sebastian's version now, looks great.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
```diff
@@ -53,6 +55,34 @@
...
+
+  public static MockCredentialsProvider instance() {
+    if (MockCredentialsProvider.instance == null) {
+      MockCredentialsProvider.instance = new MockCredentialsProvider();
+    }
+    return MockCredentialsProvider.instance;
+  }
+
+  private MockCredentialsProvider() {
+    super();
```
schmidt-sebastian: Nit: This super() call is not needed, is it?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
```diff
@@ -53,6 +55,34 @@
...
+  public void setChangeListener(Listener<User> changeListener) {
+    super.setChangeListener(changeListener);
+    this.listener = changeListener;
+  }
+
+  public void changeUserTo(User user) {
+    listener.onValue(user);
+  }
+
+  private Listener<User> listener;
```
schmidt-sebastian: Nit: Move to top of class.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -362,6 +362,24 @@ public WriteBatch batch() {
...
   }
 
+  /**
+   * Waits until all currently pending writes for the active user have been acknowledged by the
+   * backend.
+   *
+   * <p>The returned Task completes immediately if there are no outstanding writes. Otherwise, the
+   * Task waits for all previously issued writes (including those written in a previous app
+   * session), but it does not wait for writes that were added after the method is called. If you
+   * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
```
schmidt-sebastian: Use `{@code}` for all mentions of waitForPendingWrites.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -427,9 +436,63 @@ public void handleRejectedWrite(int batchId, Status status) {
...
 
+  /**
+   * Takes a snapshot of current mutation queue, and register a user task which will resolve when
+   * all those mutations are either accepted or rejected by the server.
+   */
+  public void registerPendingWritesTask(TaskCompletionSource<Void> userTask) {
+    if (!remoteStore.canUseNetwork()) {
+      Logger.debug(
+          TAG,
+          "The network is disabled. The task returned by 'awaitPendingWrites()' will not"
```
schmidt-sebastian: Super nit: We usually leave the space at the end of the line. This helps to align comments with more than two lines.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -427,9 +436,63 @@ public void handleRejectedWrite(int batchId, Status status) {
...
+    }
+
+    if (pendingWritesCallbacks.containsKey(largestPendingBatchId)) {
+      pendingWritesCallbacks.get(largestPendingBatchId).add(userTask);
+    } else {
+      pendingWritesCallbacks.put(largestPendingBatchId, Lists.newArrayList(userTask));
+    }
+  }
+
+  /** Resolves tasks waiting for this batch id to get acknowledged by server, if there is any. */
```
schmidt-sebastian: s/is/are/

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -1143,4 +1143,21 @@ public void testHandlesPatchMutationWithTransformThenRemoteEvent() {
...
+    assertEquals(1, localStore.getHighestUnacknowledgedBatchId());
+
+    writeMutation(patchMutation("foo/bar", map("abc", 321)));
+    assertEquals(2, localStore.getHighestUnacknowledgedBatchId());
+
+    acknowledgeMutation(1);
+    assertEquals(2, localStore.getHighestUnacknowledgedBatchId());
+
+    rejectMutation();
+    assertEquals(-1, localStore.getHighestUnacknowledgedBatchId());
```
schmidt-sebastian: s/-1/MutationBatch.UNKNOWN/

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
Hui-Wu: It's not.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
Hui-Wu: Done.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
Hui-Wu: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
Hui-Wu: Done.

PR #693, opened 2019-08-07
Add runtime kotlin version detection for platform info purposes.

vkryachko:


firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -668,4 +675,13 @@ public void execute(@NonNull Runnable command) {
       HANDLER.post(command);
     }
   }
+
+  @Nullable
+  private static String detectKotlinVersion() {
+    try {
+      return kotlin.KotlinVersion.CURRENT.toString();
+    } catch (NoClassDefFoundError ex) {
+      return null;
```
ashwinraghav: Can we write this as "no_kotlin_version" or "-1". Could give us a sense of what portion of clients do not have kotlin. I realise this can be derived already. But, seems more convenient

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -668,4 +675,13 @@ public void execute(@NonNull Runnable command) {
       HANDLER.post(command);
     }
   }
+
+  @Nullable
+  private static String detectKotlinVersion() {
+    try {
+      return kotlin.KotlinVersion.CURRENT.toString();
+    } catch (NoClassDefFoundError ex) {
```
ashwinraghav: Based on cursory research, this seems like the only exception that could be caused.
Are there other exceptions we should be watching out for ? ClassNotFoundException ?

vkryachko: Nope, this is the only one that can be thrown

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
vkryachko: I erred on the side of not populating what's not present, e.g. it's consistent with what is done for unity, flutter, cpp, etc. So I think I prefer to do it this way, but @davidair could correct me if that's not the case and/or your suggestion adds convenience.

firebase-common/ktx/src/test/kotlin/com/google/firebase/ktx/Tests.kt
```diff
@@ -42,7 +42,7 @@ class VersionTests {
     fun libraryVersions_shouldBeRegisteredWithRuntime() {
         withApp("ktxTestApp") {
             val uaPublisher = get(UserAgentPublisher::class.java)
-            assertThat(uaPublisher.userAgent).contains("kotlin")
+            assertThat(uaPublisher.userAgent).contains("fire-kotlin")
```
davidair: Please keep as "kotlin" since it's not a Firebase product

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
davidair: We normally only put things into the platform info that "are there" - and they vary between platform. It doesn't make sense to put the Xcode version into the Android string, so I would recommend omitting Kotlin if not present.

PR #695, opened 2019-08-08
Don't persist manufactured documents

schmidt-sebastian:
This PR makes sure we don't write documents to the RemoteDocumentCache that have a version of 0. This allows us to add an assert in the Index-Free query branch and also makes sure that we don't just persist NoDocuments because of RPC failures.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
@@ -28,6 +28,7 @@ import com.google.firebase.firestore.core.TargetIdGenerator;
 import com.google.firebase.firestore.model.Document;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.MaybeDocument;
+import com.google.firebase.firestore.model.NoDocument;
 import com.google.firebase.firestore.model.SnapshotVersion;
 import com.google.firebase.firestore.model.mutation.Mutation;
 import com.google.firebase.firestore.model.mutation.MutationBatch;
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -381,17 +382,18 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
...
 
-            // If a document update isn't authoritative, make sure we don't
-            // apply an old document version to the remote cache. We make an
-            // exception for SnapshotVersion.MIN which can happen for
-            // manufactured events (e.g. in the case of a limbo document
-            // resolution failing).
+            // If a document update isn't authoritative, make sure we don't apply an old document
+            // version to the remote cache. We make an exception for SnapshotVersion.MIN which can
+            // happen for manufactured events (e.g. in the case of a limbo document resolution
+            // failing).
```
mikelehen: As mentioned on chat, this comment is now misleading.  It says we don't write old document versions to the remote cache except we make an exception for SnapshotVersion.MIN (implying we _do_ write it to cache).  With your change, this is no longer true.

Perhaps remove "We make an exception ..." from this comment and add a new comment inside your new else-if case that says something like:

This is a manufactured event (since version==SnapshotVersion.NONE), likely as a result of a failed limbo resolution lookup.  Rather than write the NoDocument to cache with an invalid version, we just remove whatever might currently be in cache.

schmidt-sebastian: I updated it just as you wrote this comment:) Let me know if the new version is a better description. Thanks!

mikelehen: Haha.  Nice!  LGTM.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -381,17 +382,19 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
...
-            // resolution failing).
+            // If a document update isn't authoritative, make sure we don't apply an old document
+            // version to the remote cache.
             if (existingDoc == null
-                || doc.getVersion().equals(SnapshotVersion.NONE)
                 || (authoritativeUpdates.contains(doc.getKey()) && !existingDoc.hasPendingWrites())
                 || doc.getVersion().compareTo(existingDoc.getVersion()) >= 0) {
               remoteDocuments.add(doc);
               changedDocs.put(key, doc);
+              // NoDocuments with SnapshotVersion.MIN are used in manufactured events (e.g. in the
```
wilhuff: This structure make the comment look as if it applies to the block about authoritative updates. Better to put this in the next block (and pull the comment above into this one--above the actions taken).

schmidt-sebastian: Done

PR #697, opened 2019-08-09
Add Index-Free query engine

schmidt-sebastian:
This is the last PR with index-free logic. It, however, doesn't turn it on yet (and is still against the indexfree-master branch).

There are a fair bit of comments in the code to explain how it all works. The only other thing worth pointing out is that I updated the LocalStoreTests to run parameterized with different query engine, which changed some of the plumbing during the test setup.

I verified that the Spec tests pass with the IndexFreeQueryEngine, but also left those running against the SimpleQueryEngine to match the code in production.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
@@ -126,8 +126,7 @@ public final class Query {
         && startAt == null
         && endAt == null
         && (getExplicitOrderBy().isEmpty()
-            || (getExplicitOrderBy().size() == 1
-                && getFirstOrderByField().isKeyField()));
+            || (getExplicitOrderBy().size() == 1 && getFirstOrderByField().isKeyField()));
   }
 
   /** The filters on the documents returned by the query. */
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
@@ -117,6 +117,13 @@ public final class Query {
     return collectionGroup != null;
   }
 
+  /**
+   * Returns true if this query does not specify any query constraints that could remove results.
+   */
+  public boolean matchesAllDocuments() {
+    return filters.isEmpty() && limit == NO_LIMIT && startAt == null && endAt == null;
+  }
+
   /** The filters on the documents returned by the query. */
   public List<Filter> getFilters() {
     return filters;
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
@@ -121,7 +121,11 @@ public final class Query {
    * Returns true if this query does not specify any query constraints that could remove results.
    */
   public boolean matchesAllDocuments() {
-    return filters.isEmpty() && limit == NO_LIMIT && startAt == null && endAt == null;
+    return filters.isEmpty()
+        && getExplicitOrderBy().isEmpty()
+        && limit == NO_LIMIT
+        && startAt == null
+        && endAt == null;
   }
 
   /** The filters on the documents returned by the query. */
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
@@ -122,10 +122,12 @@ public final class Query {
    */
   public boolean matchesAllDocuments() {
     return filters.isEmpty()
-        && getExplicitOrderBy().isEmpty()
         && limit == NO_LIMIT
         && startAt == null
-        && endAt == null;
+        && endAt == null
+        && (getExplicitOrderBy().isEmpty()
+            || (getExplicitOrderBy().size() == 1
+                && getFirstOrderByField().isKeyField()));
   }
 
   /** The filters on the documents returned by the query. */
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
@@ -16,6 +16,7 @@ package com.google.firebase.firestore.local;
 
 import static com.google.firebase.firestore.util.Assert.hardAssert;
 
+import androidx.annotation.Nullable;
 import com.google.firebase.database.collection.ImmutableSortedMap;
 import com.google.firebase.database.collection.ImmutableSortedSet;
 import com.google.firebase.firestore.core.Query;
@@ -25,7 +26,6 @@ import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.MaybeDocument;
 import com.google.firebase.firestore.model.SnapshotVersion;
 import java.util.Map;
-import javax.annotation.Nullable;
 
 /**
  * A query engine that takes advantage of the target document mapping in the QueryCache. The
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
@@ -1,0 +1,151 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+import static com.google.firebase.firestore.util.Assert.hardAssert;
+
+import com.google.firebase.database.collection.ImmutableSortedMap;
+import com.google.firebase.database.collection.ImmutableSortedSet;
+import com.google.firebase.firestore.core.Query;
+import com.google.firebase.firestore.model.Document;
+import com.google.firebase.firestore.model.DocumentCollections;
+import com.google.firebase.firestore.model.DocumentKey;
+import com.google.firebase.firestore.model.MaybeDocument;
+import com.google.firebase.firestore.model.SnapshotVersion;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * A query engine that takes advantage of the target document mapping in the QueryCache. The
+ * IndexFreeQueryEngine optimizes query execution by only reading the documents previously matched a
+ * query plus any documents that were edited after the query was last listened to.
+ *
+ * <p>There are some cases where Index-Free queries are not guaranteed to produce to the same
+ * results as full collection scans. In these case, the IndexFreeQueryEngine falls back to a full
+ * query processing. These cases are:
+ *
+ * <ol>
+ *   <li>Limit queries where a document that matched the query previously no longer matches the
+ *       query. In this case, we have to scan all local documents since a document that was sent to
+ *       us as part of a different query result may now fall into the limit.
+ *   <li>Limit queries that include edits that occurred after the last remote snapshot (both
+ *       latency-compensated or committed). Even if an edited document continues to match the query,
+ *       an edit may cause a document to sort below another document that is in the local cache.
+ *   <li>Queries where the last snapshot contained Limbo documents. Even though a Limbo document is
+ *       not part of the backend result set, we need to include Limbo documents in local views to
+ *       ensure consistency between different Query views. If there exists a previous query snapshot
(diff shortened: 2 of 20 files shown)


schmidt-sebastian:
PTAL :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -561,6 +562,21 @@ public QueryData allocateQuery(Query query) {
...
+   * persisted to the QueryCache.
+   */
+  @VisibleForTesting
+  @Nullable
+  QueryData getQueryData(Query query) {
+    QueryData queryData = queryCache.getQueryData(query);
+    if (queryData == null) {
+      return null;
+    }
+    QueryData updatedQueryData = targetIds.get(queryData.getTargetId());
```
wilhuff: If you asked the map first you could avoid even reading from the query cache most of the time.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -611,7 +627,23 @@ public void releaseQuery(Query query) {
 
   /** Runs the given query against all the documents in the local store and returns the results. */
   public ImmutableSortedMap<DocumentKey, Document> executeQuery(Query query) {
-    return queryEngine.getDocumentsMatchingQuery(query);
+    QueryData queryData = getQueryData(query);
+    if (queryData != null) {
+      ImmutableSortedSet<DocumentKey> remoteKeys =
+          this.queryCache.getMatchingKeysForTargetId(queryData.getTargetId());
```
wilhuff: Not all query engines require this data, so shouldn't this be part of the index-free engine only?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -14,17 +14,26 @@
...
-/** Represents a query engine capable of performing queries over the local document cache. */
-interface QueryEngine {
+/**
+ * Represents a query engine capable of performing queries over the local document cache. You must
+ * call setQueryDataProvider() and setLocalDocumentsView() before using.
+ */
+public interface QueryEngine {
+
+  /** Sets the document view to query against. */
+  void setLocalDocumentsView(LocalDocumentsView localDocuments);
```
wilhuff: I wish there were a way to avoid this.

firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
```diff
@@ -528,4 +529,22 @@ public void testImplicitOrderBy() {
...
         baseQuery.orderBy(orderBy("foo", "desc")).orderBy(orderBy("bar", "asc")).getOrderBy());
   }
+
+  @Test
+  public void testMatchesAllDocuments() {
+    Query baseQuery = Query.atPath(ResourcePath.fromString("collection"));
+    assertTrue(baseQuery.matchesAllDocuments());
+
+    Query query = baseQuery.filter(filter("foo", "==", "bar"));
+    assertFalse(query.matchesAllDocuments());
```
wilhuff: Add a test for OrderBy?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java
```diff
@@ -0,0 +1,281 @@
...
+    MockitoAnnotations.initMocks(this);
+
+    expectIndexFreeExecution = false;
+    existingQueryResults.clear();
+    updatedQueryResults.clear();
+
+    queryEngine = new IndexFreeQueryEngine();
+    queryEngine.setLocalDocumentsView(localDocumentsView);
+
+    doAnswer(
```
wilhuff: It seems like this could be done much more straightforwardly by just using a regular MemoryQueryCache. 

This would make `addExistingResult` call `queryCache.addMatchingKeys(doc.getKey(), TEST_TARGET_ID)` and avoid all these mocks, which don't port easily and end up being very brittle and hard to change.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java
```diff
@@ -0,0 +1,281 @@
...
+  private static final Document NON_MATCHING_DOC_B =
+      doc("coll/b", 1, map("matches", false, "order", 2), Document.DocumentState.SYNCED);
+  private static final Document UPDATED_MATCHING_DOC_B =
+      doc("coll/b", 11, map("matches", true, "order", 2), Document.DocumentState.SYNCED);
+
+  @Mock private LocalDocumentsView localDocumentsView;
+  @Mock private QueryCache queryCache;
+  private QueryEngine queryEngine;
+
+  private final Map<DocumentKey, Document> existingQueryResults = new HashMap<>();
```
wilhuff: I don't entirely understand the difference between these two maps. Could you add comments about what they mean and how they're used by the test?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java
```diff
@@ -0,0 +1,281 @@
...
+        runQuery(query, queryData, /* expectIndexFree= */ true);
+    assertEquals(asList(MATCHING_DOC_A, MATCHING_DOC_B), values(docs));
+  }
+
+  @Test
+  public void filtersNonMatchingInitialResults() {
+    Query query = query("coll").filter(filter("matches", "==", true));
+    QueryData queryData = queryData(query, /* hasLimboFreeSnapshot= */ true);
+
+    addExistingResult(NON_MATCHING_DOC_A);
```
wilhuff: How would this happen? Wouldn't an existing result from the server have to match?

It seems like maybe I don't understand what an existing result is. See comment above.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java
```diff
@@ -0,0 +1,281 @@
...
+  public void usesTargetMappingForInitialView() {
+    Query query = query("coll").filter(filter("matches", "==", true));
+    QueryData queryData = queryData(query, /* hasLimboFreeSnapshot= */ true);
+
+    addExistingResult(MATCHING_DOC_A);
+    addExistingResult(MATCHING_DOC_B);
+
+    ImmutableSortedMap<DocumentKey, Document> docs =
+        runQuery(query, queryData, /* expectIndexFree= */ true);
+    assertEquals(asList(MATCHING_DOC_A, MATCHING_DOC_B), values(docs));
```
wilhuff: You don't ever use `docs` except to pass it to `values()`. Maybe have `runQuery` just return the list of matching documents?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -1007,6 +1033,211 @@ public void testHandlesSetMutationThenAckThenTransformMutationThenAckThenTransfo
     assertChanged(doc("foo/bar", 2, map("sum", 3), Document.DocumentState.LOCAL_MUTATIONS));
   }
 
+  @Test
+  public void testUsesTargetMappingToExecuteQueries() {
+    // This test verifies that once a target mapping has been written, only documents that match
+    // the query are read from the RemoteDocumentCache.
+    assumeTrue(queryEngine instanceof IndexFreeQueryEngine);
```
wilhuff: Nit: none of the other platforms have an equivalent to `assumeTrue`. Is this really better than just `if`/`return`?

It seems like `!garbageCollectorIsEager()` is a similar kind of assumption. Handle them the same way?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -14,14 +14,33 @@
...
 import org.robolectric.annotation.Config;
 
-@RunWith(RobolectricTestRunner.class)
+@RunWith(ParameterizedRobolectricTestRunner.class)
 @Config(manifest = Config.NONE)
 public class SQLiteLocalStoreTest extends LocalStoreTestCase {
 
+  private QueryEngine queryEngine;
+
+  @ParameterizedRobolectricTestRunner.Parameters(name = "QueryEngine = {0}")
```
wilhuff: Probably not for this PR but all the different mechanisms we have for parameterizing the local store tests are getting out of hand.

We should have one way of describing the product of persistence x gc x query-engine.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java
```diff
@@ -0,0 +1,251 @@
...
+    assertEquals(docSet(query.comparator(), MATCHING_DOC_B), docs);
+  }
+
+  @Test
+  public void includesChangesSinceInitialResults() {
+    Query query = query("coll").filter(filter("matches", "==", true));
+    QueryData originalQueryData = queryData(query, /* hasLimboFreeSnapshot= */ true);
+
+    addDocumentToRemoteResult(MATCHING_DOC_A);
+    addDocumentToRemoteResult(NON_MATCHING_DOC_B);
```
wilhuff: Similar to my earlier question: I don't understand how this happens. Is this shorthand for the query result initially matching B, but then an update changed B to not match?

Also, all versions of document B are at version 2 and changing `UPDATED_MATCHING_DOC_B` doesn't change that. I guess this happens to work because `queryData` forces the initial query to version 1, but that means that we never find B except because it's been updated. 

Is this the premise of the test? If so it'd be worth spelling this out in comments.

If so, note that "remote" and "local" don't really capture the difference between these operations that well since both cases are updating the remote document cache. One case is setting which documents match the query and the other is performing an out-of-band update I'd suggest renaming these `initiallyMatched` (or `matchesQuery` or `addToQueryResult`) vs  `otherQueryUpdates` (or `updateSeparately`, `updateByOtherQuery`, `updateOutOfBand`, or something similar).

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
wilhuff: I don't have a strong preference either but it seems like we should at least handle query engine and GC the same way. `assumeFalse(garbageCollectorIsEager)` or `assumeTrue(garbageCollectorNonEager())`?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
wilhuff: Yeah, I guess unless we can make the persistence layer just have a notion of the current queue, this is hard to avoid. LGTM unless you want to try your hand at that (probably separately).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java
```diff
@@ -0,0 +1,150 @@
...
+  private @Nullable ImmutableSortedMap<DocumentKey, Document> executeIndexFreeQuery(
+      Query query, QueryData queryData, ImmutableSortedSet<DocumentKey> remoteKeys) {
+    // Fetch the documents that matched the query at the last snapshot.
+    ImmutableSortedMap<DocumentKey, MaybeDocument> previousResults =
+        localDocumentsView.getDocuments(remoteKeys);
+
+    // Limit queries are not eligible for index-free query execution if any part of the result was
+    // modified after we received the last query snapshot. This makes sure that we re-populate the
+    // view with older documents that may sort before the modified document.
+    if (query.hasLimit()
```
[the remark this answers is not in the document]
wilhuff: That seems pretty reasonable.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -117,6 +117,13 @@ public boolean isCollectionGroupQuery() {
     return collectionGroup != null;
   }
 
+  /**
+   * Returns true if this query does not specify any query constraints that could remove results.
+   */
+  public boolean matchesAllDocuments() {
+    return filters.isEmpty() && limit == NO_LIMIT && startAt == null && endAt == null;
```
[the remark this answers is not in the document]
schmidt-sebastian: I didn't feel like there was much to gain from special-casing this, but since this is now a generalized API in Query itself, it probably makes sense to address this case.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java
schmidt-sebastian: For this test case, I can just use two matching documents in the beginning and overwrite one to be non-matching. There is a similar test case where that does quite achieve the same test coverage, I added a comment here (`doesNotUseInitialResultsForLimitQueryWithDocumentRemoval()`).

Furthermore, I split the addDocumentToRemoteResult() into two calls. We now have `addDocument()` and `persistQueryMapping()` which should hopefully make it much clearer what is going on.

schmidt-sebastian:
Test failure is for a RTDB test (persistedServerValuesAreResolved).

PR #698, opened 2019-08-09
Retrying transactions with backoff

thebrianchen:
Good call on recommending that I try and implement Android first before merging web. This has a been a painful, but educational experience.

A couple things about Android:
- The recursive method seemed like the only viable option here, since we cannot await backoffs in the same thread as the one calling it (something that I realized after a bit too much struggling). That being said, I still think the web approach is cleaner and more readable compared to Android. How do you feel about keeping the implementations different? I'm also not sure how I would port the whole `TaskCompletionSource` strategy used by Android for recursion over to web.
- The `skipDelaysForTimerIds` approach of web doesn't port over to Android b/c it relies on `runDelayedTasksUntil`, which blocks until the delay is completed. This means that we must call delays from outside the AsyncQueue, or else it deadlocks. Overriding the backoff is the only other approach that works aside from calling a `runDelayedTasksUntil` on a 10ms interval on a separate thread. Thoughts on this approach?

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
@@ -250,8 +250,8 @@ public class SyncEngine implements RemoteStore.RemoteStoreCallback {
    * updateFunction, the client can read and write values using the supplied transaction object.
    * After the updateFunction, all changes will be committed. If a retryable error occurs (ex: some
    * other client has changed any of the data referenced), then the updateFunction will be called
-   * again after a backoff. If the updateFunction still fails after the given number of retires,
-   * then the transaction will be rejected.
+   * again after a backoff. If the updateFunction still fails after all retries, then the
+   * transaction will be rejected.
    *
    * <p>The transaction object passed to the updateFunction contains methods for accessing documents
    * and collections. Unlike other datastore access, data accessed with the transaction will not
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
@@ -262,11 +262,7 @@ public class SyncEngine implements RemoteStore.RemoteStoreCallback {
    */
   public <TResult> Task<TResult> transaction(
       AsyncQueue asyncQueue, Function<Transaction, Task<TResult>> updateFunction) {
-    TransactionRunner<TResult> runner =
-        new TransactionRunner<>(asyncQueue, remoteStore, updateFunction, 5);
-
-    runner.runTransaction();
-    return runner.getTask();
+    return new TransactionRunner<TResult>(asyncQueue, remoteStore, updateFunction).run();
   }
 
   /** Called by FirestoreClient to notify us of a new remote event. */
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
@@ -23,7 +23,6 @@ import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
-import com.google.android.gms.tasks.Tasks;
 import com.google.common.base.Function;
 import com.google.common.collect.Lists;
 import com.google.firebase.database.collection.ImmutableSortedMap;
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
@@ -17,10 +17,8 @@ package com.google.firebase.firestore.core;
 import static com.google.firebase.firestore.util.Assert.fail;
 import static com.google.firebase.firestore.util.Assert.hardAssert;
 
-import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
-import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.common.base.Function;
@@ -43,12 +41,10 @@ import com.google.firebase.firestore.model.SnapshotVersion;
 import com.google.firebase.firestore.model.mutation.Mutation;
 import com.google.firebase.firestore.model.mutation.MutationBatch;
 import com.google.firebase.firestore.model.mutation.MutationBatchResult;
-import com.google.firebase.firestore.remote.Datastore;
 import com.google.firebase.firestore.remote.RemoteEvent;
 import com.google.firebase.firestore.remote.RemoteStore;
 import com.google.firebase.firestore.remote.TargetChange;
 import com.google.firebase.firestore.util.AsyncQueue;
-import com.google.firebase.firestore.util.ExponentialBackoff;
 import com.google.firebase.firestore.util.Logger;
 import com.google.firebase.firestore.util.Util;
 import io.grpc.Status;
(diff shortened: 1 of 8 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,119 @@
...
+import com.google.firebase.firestore.FirebaseFirestoreException;
+import com.google.firebase.firestore.remote.Datastore;
+import com.google.firebase.firestore.remote.RemoteStore;
+import com.google.firebase.firestore.util.AsyncQueue;
+import com.google.firebase.firestore.util.AsyncQueue.TimerId;
+import com.google.firebase.firestore.util.ExponentialBackoff;
+
+/**
+ * TransactionRunner encapsulates the logic needed to run and retry transactions so that the caller
+ * does not have to manage the backoff and retry count through recursive calls.
```
mikelehen: "so that the caller does not have to manage the backoff and retry count through recursive calls." probably won't make a lot of sense to random-dev-in-the-future who is unaware of how this code used to look.  I would just omit it or just say "... logic needed to run and retry transactions with backoff." or whatever.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public class TransactionRunner<TResult> {
+  private AsyncQueue asyncQueue;
+  private RemoteStore remoteStore;
+  private Function<Transaction, Task<TResult>> updateFunction;
+  private int retries;
```
mikelehen: retries => retriesLeft ?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private ExponentialBackoff backoff;
+  private TaskCompletionSource<TResult> taskSource = new TaskCompletionSource<>();
+
+  public TransactionRunner(
+      AsyncQueue asyncQueue,
+      RemoteStore remoteStore,
+      Function<Transaction, Task<TResult>> updateFunction,
+      int retries) {
```
mikelehen: Since retries is always passed as 5, I would remove it as an argument and instead define a RETRY_COUNT constant in this class.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,119 @@
...
+                                  }
+                                });
+                      }
+                    }
+                  });
+        });
+  }
+
+  /** Returns the result of the transaction after it has been run. */
+  public Task<TResult> getTask() {
```
mikelehen: Is there a reason for this to be a separate function?  I think I'd just make runTransaction() return the `Task<TResult>`.

This will also probably port better to iOS (`runTransaction()` will just accept a completion callback instead of returning a Task).

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,119 @@
...
+                  new OnCompleteListener<TResult>() {
+                    @Override
+                    public void onComplete(@NonNull Task<TResult> userTask) {
+                      if (!userTask.isSuccessful()) {
+                        if (retries > 0 && isRetryableTransactionError(userTask.getException())) {
+                          retries -= 1;
+                          runTransaction();
+                        } else {
+                          taskSource.setException(userTask.getException());
+                        }
```
mikelehen: The above ~6 lines are essentially duplicated below.  Can we move them into a helper method?  Maybe `handleTransactionError()` or something?

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
```diff
@@ -471,19 +479,28 @@ public void enqueueAndForget(Runnable task) {
...
-        !containsDelayedTask(timerId),
-        "Attempted to schedule multiple operations with timer id %s.",
-        timerId);
-
     DelayedTask delayedTask = createAndScheduleDelayedTask(timerId, delayMs, task);
     delayedTasks.add(delayedTask);
 
+    // Fast-forward delays for timerIds that have been overridden.
+    for (TimerId timerIdToSkip : timerIdsToSkip) {
+      if (delayedTask.timerId == timerIdToSkip) {
```
mikelehen: Is there a reason this can't be

```
if (timerIdsToSkip.contains(delayedTask.timerId)) {
  delayedTask.skipDelay();
}
```
?

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
```diff
@@ -471,19 +479,28 @@ public void enqueueAndForget(Runnable task) {
...
+        delayedTask.skipDelay();
+      }
+    }
     return delayedTask;
   }
 
+  /**
+   * For Tests: Skip all delays for a timer id.
```
mikelehen: Maybe say "Skip all subsequent delays ..." to make it clear that it won't fast-forward already queued delayed operations.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -262,35 +261,12 @@ private void addUserCallback(int batchId, TaskCompletionSource<Void> userTask) {
...
-                        }
-                        return Tasks.forException(e);
-                      });
-            });
+      AsyncQueue asyncQueue, Function<Transaction, Task<TResult>> updateFunction) {
+    TransactionRunner<TResult> runner =
+        new TransactionRunner<>(asyncQueue, remoteStore, updateFunction, 5);
+
+    runner.runTransaction();
+    return runner.getTask();
```
mikelehen: FWIW- I think I would:
1) Make runTransaction() return the Task directly as I suggest below.
2) Rename `runTransaction()` to just `run()`
3) Change this to just:

```
return new TransactionRunner<TResult>(asyncQueue, remoteStore, updateFunction).run();
```

But if you're not a fan, feel free to disregard.

mikelehen:
Oh! By the way, I realized I forgot to answer your question:

> from a code health and maintainability perspective, why is it bad to introduce new test hooks, like I did with removeTransactionBackoffs? I can see how skipDelaysForTimerId is a more elegant solution, but when is it appropriate/preferred to add new test hooks?

Uh, this is pretty subjective.  I guess my general philosophy is that test hooks are bad for a variety of reasons, including:
1) You end up shipping test-only code as part of your product which means it's essentially "dead code."
2) It means your tests are modifying the behavior of the client which necessarily means it's not a fully realistic test so there's risk of masking bugs or confusing future devs working on the code / tests.

And so in general, it's good to have as few test hooks as possible, keep the scope as limited as possible, and make them reusable so you don't have to invent more test hooks later.

In our case, we've already chosen the AsyncQueue as a useful place to allow tests to hook in to deal with time-related issues in tests (i.e. fast-forwarlding time via runDelayedOperationsEarly()).  So adding new test hooks to the AsyncQueue is preferable to a new transaction-specific test hook.  It's more consistent and is more likely to be reusable in the future.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private void runWithBackoff() {
+    backoff.backoffAndRun(
+        () -> {
+          final Transaction transaction = remoteStore.createTransaction();
+          updateFunction
+              .apply(transaction)
+              .addOnCompleteListener(
+                  asyncQueue.getExecutor(),
+                  new OnCompleteListener<TResult>() {
```
mikelehen: Sorry for not mentioning before, but you can replace these anonymous classes with lambda syntax, which looks a bit cleaner:

```
    backoff.backoffAndRun(
        () -> {
          final Transaction transaction = remoteStore.createTransaction();
          updateFunction
              .apply(transaction)
              .addOnCompleteListener(
                  asyncQueue.getExecutor(),
                  (@NonNull Task<TResult> userTask) -> {
                    if (!userTask.isSuccessful()) {
                      handleTransactionError(userTask);
                    } else {
                      transaction
                          .commit()
                          .addOnCompleteListener(
                              asyncQueue.getExecutor(),
                              (@NonNull Task<Void> commitTask) -> {
                                if (commitTask.isSuccessful()) {
                                  taskSource.setResult(userTask.getResult());
                                } else {
                                  handleTransactionError(commitTask);
                                }
                              });
                    }
                  });
        });
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -0,0 +1,116 @@
...
+      // non-matching document versions with ABORTED. These errors should be retried.
+      FirebaseFirestoreException.Code code = ((FirebaseFirestoreException) e).getCode();
+      return code == FirebaseFirestoreException.Code.ABORTED
+          || code == FirebaseFirestoreException.Code.FAILED_PRECONDITION
+          || !Datastore.isPermanentError(((FirebaseFirestoreException) e).getCode());
+    }
+    return false;
+  }
+
+  private void handleTransactionError(Task task) {
```
mikelehen: nit: We typically organize methods "top-down" meaning that this should go above isRetryableTransactionError(), since runWithBackoff() calls handleTransactionError() which calls isRetryableTransactionError().

firebase-firestore/src/main/java/com/google/firebase/firestore/core/TransactionRunner.java
```diff
@@ -27,8 +26,7 @@
 import com.google.firebase.firestore.util.ExponentialBackoff;
 
 /**
- * TransactionRunner encapsulates the logic needed to run and retry transactions so that the caller
- * does not have to manage the backoff and retry count through recursive calls.
+ * TransactionRunner encapsulates the logic needed to run and retry transactions without backoff.
```
mikelehen: without => with ? 😄

PR #703, opened 2019-08-13
getId() implementation with instrumented tests.

ankitaj224:


firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -26,4 +47,99 @@
...
+                .setApplicationId("1:123456789:android:abcdef")
+                .setProjectId("project-id")
+                .setApiKey("api_key")
+                .build());
+    actualCache = new FiidCache(firebaseApp);
+    when(backendClientReturnsOk.createFirebaseInstallation(
+            anyString(), anyString(), anyString(), anyString()))
+        .thenReturn(
+            InstallationResponse.builder()
+                .setName("/projects/123456789/installations/fid")
```
andirayo: Please use a properly formatted FID.
Maybe not necessary right now, but I have seen many tests in the past that had to be updated because a simple format check was added somewhere.
Just use something like "fis-testing-is-awesome" or "cccccccccccccccccccccc" which are both valid FIDs.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -26,4 +47,99 @@
...
+                .setRefreshToken("refresh-token")
```
andirayo: Refresh-Tokens currently start with "1:"
How about
"1:test-refresh-token"

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -26,4 +47,99 @@
...
+                .setAuthToken(
+                    InstallationTokenResult.builder()
+                        .setToken("auth-token")
```
andirayo: FIS Auth-Tokens are currently JWTokens with 2 periods.
How about "fis.auth.token"

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -35,10 +43,23 @@
...
+    this(firebaseApp, new FiidCache(firebaseApp), new FirebaseInstallationServiceClient());
+  }
+
+  FirebaseInstallations(
+      FirebaseApp firebaseApp,
+      FiidCache localCache,
+      FirebaseInstallationServiceClient serviceClient) {
     this.firebaseApp = firebaseApp;
+    this.serviceClient = serviceClient;
+    this.executor = Executors.newFixedThreadPool(6);
```
andirayo: Why is this not a constant?
Why 6?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
     return firebaseApp.getName();
   }
+
+  /**
+   * Create firebase installation id for the {@link FirebaseApp} on FIS Servers and client side
+   * cache.
+   *
+   * <pre>
+   *     The workflow is:
+   *         check if cache empty or cache status is REGISTER_ERROR
```
andirayo: This description is somewhat confusing to me: What happens if the status is REGISTER_ERROR or not?

andirayo: Also what happens if status is UNREGISTERED?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+   *                                 |
+   *                           create random fid
+   *                                 |
+   *               update cache with cache status UNREGISTERED
```
andirayo: is there a status "IN PROGRESS" or something to prevent multiple requests in parallel?
do you get a lock on the database entry or how do you deal with multithreading?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+   *                                 |
+   *                           send http request to backend
+   *                                 |                  |
+   *  on success: set cache entry status to REGISTERED  |
+   *                                 |                  |
+   *                                 |         on failure: set cache entry status to REGISTER_ERROR
+   *                                 |                     |
+   *                                return          throw exception
```
andirayo: return FID

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+   *
+   *
+   *                      else if cached FID exists
+   *                             |            |
+   *     if cache status is UNREGISTERED      |
+   *                  |                   return cached FID
```
andirayo: this line should fork from the "if cache status is UNREGISTERED"

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -26,4 +47,99 @@
...
+  }
+
+  @Test
+  public void testCreateFirebaseInstallation_CacheOk_BackendOk() throws Exception {
+    FirebaseInstallations firebaseInstallations =
+        new FirebaseInstallations(firebaseApp, actualCache, backendClientReturnsOk);
+
+    // No exception, means success.
+    assertThat(Tasks.await(firebaseInstallations.getId())).isNotEmpty();
+    FiidCacheEntryValue entryValue = actualCache.readCacheEntryValue();
```
andirayo: Ok, I have a few problems here:
a) What is Fiid? It's called FID.
b) I'm a bit worried about the term "cache" here.
As a cache I understand something that is temporarily in memory of a system in order to safe multiple database- or filesystem-lookups or requests, but when the system shuts down the cache is usually deleted.

Isn't this the local storage, where the FID (and other data) gets persisted?
If yes, can't we call is something like that, e.g. "FidStorageEntry" or "PersistedFidEntry", I don't know.

Happy to discuss in person.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -26,4 +47,99 @@
...
+  }
+
+  @Test
+  public void testCreateFirebaseInstallation_CacheError_BackendOk() throws InterruptedException {
+    FirebaseInstallations firebaseInstallations =
+        new FirebaseInstallations(firebaseApp, cacheReturnsError, backendClientReturnsOk);
+
+    // Expect exception
+    try {
+      Tasks.await(firebaseInstallations.getId());
```
andirayo: out of curiosity:
Does this framework not have "assertThrows"?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+
+    FiidCacheEntryValue cacheEntryValue = localCache.readCacheEntryValue();
+
+    if (cacheEntryValue == null
+        || cacheEntryValue.getCacheStatus() == FiidCache.CacheStatus.REGISTER_ERROR) {
+      String fid = Utils.createRandomFid();
+
+      boolean firstUpdateCacheResult =
+          localCache.insertOrUpdateCacheEntry(
+              FiidCacheEntryValue.create(fid, FiidCache.CacheStatus.UNREGISTERED, "", "", 0, 0));
```
andirayo: maybe use AutoValue and a Builder here?
Seems like a lot of parameters and it's unclear what , "'', "", 0, 0 does

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -35,10 +43,23 @@
 public class FirebaseInstallations implements FirebaseInstallationsApi {
 
   private final FirebaseApp firebaseApp;
+  private final FirebaseInstallationServiceClient serviceClient;
+  private final FiidCache localCache;
```
andirayo: For the usage of term "cache" see below.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+  /**
+   * Registers the created FID with FIS Servers if the Network is available and update the cache.
+   */
+  private void registerAndSaveFID(String fid) throws FirebaseInstallationsException {
+    try {
+      if (Utils.isNetworkAvailable(firebaseApp.getApplicationContext())) {
+        long creationTime = Utils.getCurrentTimeInSeconds();
+
+        InstallationResponse installationResponse =
+            serviceClient.createFirebaseInstallation(
```
andirayo: I think, I mentioned this before (I know it's pretty nitty):
Can we change the order of the parameters maybe?
like first FID, then API-Key, then all the metadata?
Or first the API-Key, then FID, then all the metadata?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+    }
+    return null;
+  }
+
+  /**
+   * Registers the created FID with FIS Servers if the Network is available and update the cache.
+   */
+  private void registerAndSaveFID(String fid) throws FirebaseInstallationsException {
+    try {
+      if (Utils.isNetworkAvailable(firebaseApp.getApplicationContext())) {
```
andirayo: I would strongly recommend to avoid large if-clauses like this one.

How about the first line of the method is this:
if (! Utils.isNetworkAvailable(firebaseApp.getApplicationContext())) {
  return;
}

andirayo: Also: Why is this if-clause part of the try-catch block, can we pull it before the try-catch block?
Or does #isNetworkAvailable throw an exception?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+
+      return cacheEntryValue.getFirebaseInstallationId();
+    }
+    return null;
+  }
+
+  /**
+   * Registers the created FID with FIS Servers if the Network is available and update the cache.
+   */
+  private void registerAndSaveFID(String fid) throws FirebaseInstallationsException {
```
andirayo: nit: Fid (only the "F" should be uppercase)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+   *  on success: set cache        \
+   *  entry status to REGISTERED   \
+   *                |              \
+   *                |         on failure: set cache entry status to REGISTER_ERROR
+   *                |                     |
+   *              return          throw exception
+   * </pre>
+   */
+  @WorkerThread
+  private String createFirebaseInstallationId() throws FirebaseInstallationsException {
```
andirayo: Is this method supposed to be called "getFirebaseInstallationId" rather than "create..."?
I'm asking because later in the method a locally stored FID might be returned (not generated)...

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +124,105 @@ String getApplicationId() {
...
+
+    FiidCacheEntryValue cacheEntryValue = localCache.readCacheEntryValue();
+
+    if (cacheEntryValue == null
```
andirayo: I recommend to avoid nested if-clauses.
How about outsourcing the different code paths in named methods and refactor this method like this:

private String getFirebaseInstallationId() {
  if (localFidExistsAndRegistered()) {
    return localFid();
  }

  return generateRegisterSaveAndGetNewFid();
}

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Changed it. Its a random number, based on 3 APIs I chose double the number of available threads.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated the description. Let me know, if you want me to change further.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated. Thanks for pointing it out.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Changed it in the parent PR. https://github.com/firebase/firebase-android-sdk/pull/676/commits/34848c6d152bac0a13efdbaef4a341409735000d

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: We currently don't have any IN PROGRESS state. I am looking at ios-sdk to see how its handled. I ll discuss with you in person about this.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -25,8 +31,8 @@
 import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
-import com.google.firebase.installations.local.FiidCache;
-import com.google.firebase.installations.local.FiidCacheEntryValue;
+import com.google.firebase.installations.local.PersistedFid;
+import com.google.firebase.installations.local.PersistedFidEntryValue;
```
andirayo: Probably driving you crazy here, so please feel free to ignore this comment:
How about "PersistedFidEntry" instead of "PersistedFidEntryValue"?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -126,101 +127,133 @@ String getName() {
...
+   *
+   *                    Persisted Fid exists And is Registered
+   *                                  |
+   *                          return the Persisted Fid
+   *
+   *
    *
+   *                  Persisted Fid exists and is in Unregistered state
+   *                                     |
+   *                Register the Fid on FIS backened and update the PersistedFidEntry
```
andirayo: Please replace "backened" with "server" (or "API").

andirayo: I see this a lot at Google, but I will continue preaching:
My take on the term "backend":
The term "backend" only really makes sense in the context of "frontend vs. backend".
Looking at our FIS API, the OnePlatform Service is the frontend (that accepts requests). Our Spanner database is the backend (that stores data).
Looking at our FIS SDK, there is not really a frontend or backend, but if anything, the public methods are the frontend (that can be called by other SDKs or the App), the local storage is the backend (where FIDs and tokens are persisted).
I think in this case the terminology is rather client (our SDK) vs. server (our API).
Thus, I recommend to either use the term "server" or "API".

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ankitaj224: There isnt one. However, I found RC using one by adding Asssert in Test Utils.

https://github.com/firebase/firebase-android-sdk/blob/5301121df4727b357293b3b7de00555362f90170/firebase-config/src/test/java/com/google/firebase/remoteconfig/testutil/Assert.java

Rest of the firebase-android-sdk teams have handled the same way as me.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -126,101 +127,133 @@ String getName() {
...
+    return fid;
   }
 
   /**
-   * Registers the created FID with FIS Servers if the Network is available and update the cache.
+   * Registers the created Fid with FIS Servers if the Network is available and update the shared
+   * prefs.
    */
-  private void registerAndSaveFID(String fid) throws FirebaseInstallationsException {
+  private void registerAndSaveFid(String fid) throws FirebaseInstallationsException {
```
andirayo: This method has to be asynchronous to not block callers to #getId. 
What I'm saying is that if someone calls #getId, the method should immediately return an FID independent of if that FID has to be registered with FIS or not.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -126,101 +127,133 @@ String getName() {
...
+  private String getFirebaseInstallationId() throws FirebaseInstallationsException {
 
-    FiidCacheEntryValue cacheEntryValue = localCache.readCacheEntryValue();
+    PersistedFidEntryValue persistedFidEntryValue = persistedFid.readPersistedFidEntryValue();
+
+    if (persistedFidDoesNotExistsOrInErrorState(persistedFidEntryValue)) {
+      return createFidAndPersistFidEntry();
+    }
+
+    if (persistedFidExistsAndRegistered(persistedFidEntryValue)) {
```
andirayo: Can we please move this case to the top, because this is the standard case.

andirayo: Also, something we added to the Web SDK:
We should check here if the current FIS Auth Token is still valid.
If not, this request should trigger an asynchronous request to FIS to update the local FIS Auth Token.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -126,101 +127,133 @@ String getName() {
...
-    if (cacheEntryValue == null
-        || cacheEntryValue.getCacheStatus() == FiidCache.CacheStatus.REGISTER_ERROR) {
-      String fid = Utils.createRandomFid();
+    return null;
+  }
 
-      boolean firstUpdateCacheResult =
-          localCache.insertOrUpdateCacheEntry(
-              FiidCacheEntryValue.create(fid, FiidCache.CacheStatus.UNREGISTERED, "", "", 0, 0));
+  private static boolean persistedFidDoesNotExistsOrInErrorState(
```
andirayo: persistedFidMissingOrInErrorState

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -126,101 +127,133 @@ String getName() {
...
+              .setRefreshToken(installationResponse.getRefreshToken())
+              .setExpiresInSecs(
+                  installationResponse.getAuthToken().getTokenExpirationTimestampMillis())
+              .setTokenCreationEpochInSecs(creationTime)
+              .build());
 
     } catch (FirebaseInstallationServiceException exception) {
-      localCache.insertOrUpdateCacheEntry(
-          FiidCacheEntryValue.create(fid, FiidCache.CacheStatus.REGISTER_ERROR, "", "", 0, 0));
+      persistedFid.insertOrUpdatePersistedFidEntry(
```
andirayo: nice! looks so much better!

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -126,101 +127,133 @@ String getName() {
...
+          PersistedFidEntryValue.builder()
+              .setFirebaseInstallationId(fid)
+              .setPersistedStatus(PersistedFid.PersistedStatus.REGISTER_ERROR)
+              .setTokenCreationEpochInSecs(0)
```
andirayo: Do we need to set these values?
How about using default values in the Builder?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFidEntryValue.java
```diff
@@ -0,0 +1,76 @@
...
+    public abstract Builder setRefreshToken(@Nullable String value);
+
+    @NonNull
+    public abstract Builder setExpiresInSecs(long value);
+
+    @NonNull
+    public abstract Builder setTokenCreationEpochInSecs(long value);
+
+    @NonNull
+    public abstract PersistedFidEntryValue build();
```
andirayo: Here you can set default values of fields that have not been set yet.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: As discussed offline, will add this when auth token method is implemented.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Done. Please take another look.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -156,53 +156,47 @@ String getName() {
...
 
-    if (persistedFidExistsAndUnregistered(persistedFidEntryValue)) {
-      registerAndSaveFid(persistedFidEntryValue.getFirebaseInstallationId());
-      return persistedFidEntryValue.getFirebaseInstallationId();
+    if (persistedFidExistsAndUnregistered(persistedFidEntry)) {
+      Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry.getFirebaseInstallationId()));
+      return persistedFidEntry.getFirebaseInstallationId();
     }
 
     return null;
```
andirayo: How can this method ever return NULL?

ankitaj224: If delete() is called and there is no network availability, we mark registration status as deleted.  In that case, it returns NULL.

Is that not right?

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ciarand: it was only added in junit5, and this build is using junit4. https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertThrows-java.lang.Class-org.junit.jupiter.api.function.Executable-

firebase-installations/src/androidTest/java/com/google/firebase/installations/local/PersistedFidTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @Test
+  public void testReadCacheEntry_Null() {
+    assertNull(cache0.readPersistedFidEntryValue());
+    assertNull(cache1.readPersistedFidEntryValue());
+  }
+
+  @Test
+  public void testUpdateAndReadCacheEntry() throws Exception {
+    assertTrue(
```
ciarand: prefer truth in cases like this http://github.com/google/truth

firebase-installations/src/androidTest/java/com/google/firebase/installations/local/PersistedFidTest.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  @Test
+  public void testReadCacheEntry_Null() {
+    assertNull(cache0.readPersistedFidEntryValue());
+    assertNull(cache1.readPersistedFidEntryValue());
+  }
+
+  @Test
+  public void testUpdateAndReadCacheEntry() throws Exception {
```
ciarand: can you add some comments explaining what this is testing?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -71,7 +93,7 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
   @NonNull
   @Override
   public Task<String> getId() {
-    return Tasks.forResult("fid-is-better-than-iid");
+    return Tasks.call(executor, () -> getFirebaseInstallationId());
```
ciarand: this would be more cleanly expressed as a method reference (`this::getFirebaseInstallationId`)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +125,129 @@ String getApplicationId() {
   String getName() {
     return firebaseApp.getName();
   }
+
+  /**
+   * Get firebase installation id for the {@link FirebaseApp}. Either from the persisted fid entry
+   * if already exists and is not error state or create a new Fid on the FIS Servers and persist in
+   * the shared prefs.
+   *
+   * <pre>
```
ciarand: I love diagrams, but I'm not sure how much value this is bringing given it seems to be three parallel cases

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +125,129 @@ String getApplicationId() {
...
+    if (persistedFidMissingOrInErrorState(persistedFidEntry)) {
+      return createFidAndPersistFidEntry();
+    }
+
+    if (persistedFidExistsAndUnregistered(persistedFidEntry)) {
+      Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry.getFirebaseInstallationId()));
+      return persistedFidEntry.getFirebaseInstallationId();
+    }
+
+    return null;
```
ciarand: mark it `@Nullable` if it's nullable

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -83,4 +86,15 @@ private static String encodeFidBase64UrlSafe(byte[] rawValue) {
     bb.putLong(uuid.getLeastSignificantBits());
     return bb.array();
   }
+
+  public static long getCurrentTimeInSeconds() {
```
ciarand: imo your testing life is going to be way easier if you make these non-static and inject an instance of Utils. For time stuff there's already com.google.android.gms.common.util.Clock and  com.google.android.gms.common.util.DefaultClock that would be a better choice

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -83,4 +86,15 @@ private static String encodeFidBase64UrlSafe(byte[] rawValue) {
...
+    return System.currentTimeMillis() / 1000;
+  }
+
+  public static boolean isNetworkAvailable(Context context) {
```
ciarand: as above, I'd inject a "NetworkChecker" or similar so that you're not forced to muck around with a real ConnectivityManager in tests

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -99,7 +108,7 @@ public synchronized boolean insertOrUpdateCacheEntry(@NonNull FiidCacheEntryValu
   public synchronized boolean clear() {
     SharedPreferences.Editor editor = prefs.edit();
     editor.remove(getSharedPreferencesKey(FIREBASE_INSTALLATION_ID_KEY));
-    editor.remove(getSharedPreferencesKey(CACHE_STATUS_KEY));
+    editor.remove(getSharedPreferencesKey(PERSISTED_STATUS_KEY));
     editor.remove(getSharedPreferencesKey(AUTH_TOKEN_KEY));
```
ciarand: wdyt about this

```
public synchronized boolean clear() {
    SharedPreferences.Editor editor = prefs.edit();
    for (String k : FIID_PREF_KEYS) {
      editor.remove(getSharedPreferencesKey(k));
    }
    editor.commit();
}
```

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +125,129 @@ String getApplicationId() {
...
+    if (!Utils.isNetworkAvailable(firebaseApp.getApplicationContext())) {
+      return null;
+    }
+
+    try {
+      long creationTime = Utils.getCurrentTimeInSeconds();
+
+      InstallationResponse installationResponse =
+          serviceClient.createFirebaseInstallation(
+              firebaseApp.getOptions().getApiKey(),
```
ciarand: can you add /* paramName= */ style comments here? otherwise it looks really easy to transpose one of the string params into another

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -35,10 +43,24 @@
...
+    this(firebaseApp, new PersistedFid(firebaseApp), new FirebaseInstallationServiceClient());
+  }
+
+  FirebaseInstallations(
+      FirebaseApp firebaseApp,
+      PersistedFid persistedFid,
+      FirebaseInstallationServiceClient serviceClient) {
     this.firebaseApp = firebaseApp;
+    this.serviceClient = serviceClient;
+    this.executor = Executors.newFixedThreadPool(THREAD_COUNT);
```
ciarand: a fixed thread pool never GC's idle threads ("The threads in the pool will exist until it is explicitly shutdown.")

1. are you sure we need 6 threads? that seems like a huuuuge amount
2. can we not inject the executor in this constructor? it's usually an antipattern for the instantiating class to also instantiate its own thread pool

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: @andirayo Can you please check this? I think the issue is, we have introduced a state DELETED as per fis-android-sdk-design doc. But from your comments on my doc, seems like you don't want the DELETED state. 

We can change the above function as :

  @WorkerThread
  private String getFirebaseInstallationId() throws FirebaseInstallationsException {
    PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();

    if (persistedFidExistsAndRegistered(persistedFidEntry)) {
      return persistedFidEntry.getFirebaseInstallationId();
    } else  if (persistedFidExistsAndUnregistered(persistedFidEntry)) {
      Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry.getFirebaseInstallationId()));
      return persistedFidEntry.getFirebaseInstallationId();
    }
      return createFidAndPersistFidEntry();
    }

Wdyt?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
andirayo: Except for the "else if" in there (an else is not necessary after you "return" in the if-statement, I like it better yes.

How about
// not registered with FIS yet
if (fidExistsAndUnregistered(persistedFidEntry)) {
  Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry.getFirebaseInstallationId()));
}
// return FID from in local storage
if (fidExists(persistedFidEntry)) {
  return persistedFidEntry.getFirebaseInstallationId();
}

return createFidAndPersistFidEntry();

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +125,129 @@ String getApplicationId() {
...
+    return persistedFidEntry == null
+        || persistedFidEntry.getPersistedStatus() == PersistedFid.PersistedStatus.REGISTER_ERROR;
+  }
+
+  private static boolean persistedFidExistsAndRegistered(PersistedFidEntry persistedFidEntry) {
+    return persistedFidEntry != null
+        && persistedFidEntry.getPersistedStatus() == PersistedFid.PersistedStatus.REGISTERED;
+  }
+
+  private static boolean persistedFidExistsAndUnregistered(PersistedFidEntry persistedFidEntry) {
```
andirayo: I think, this method name is fine, but how about one of these?

#fidExistsButUnregistered
#fidExistsButNeedsRegistration
#fidExistsButNeedsRegistrationWithFis

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -103,4 +125,129 @@ String getApplicationId() {
...
+    return persistedFidEntry != null
+        && persistedFidEntry.getPersistedStatus() == PersistedFid.PersistedStatus.UNREGISTERED;
+  }
+
+  private String createFidAndPersistFidEntry() throws FirebaseInstallationsException {
```
andirayo: how about #createAndPersistFid

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -21,37 +21,38 @@
...
-    // Cache entry is synced to Firebase backend
+  public enum PersistedStatus {
+    // Persisted Fid entry is synced to Firebase backend
     REGISTERED,
-    // Cache entry is waiting for Firebase backend response or internal network retry
+    // Persisted Fid entry is waiting for Firebase backend response or internal network retry
     UNREGISTERED,
-    // Cache entry is in error state when syncing with Firebase backend
+    // Persisted Fid entry is in error state when syncing with Firebase backend
     REGISTER_ERROR,
```
andirayo: what's the difference between UNREGISTERED AND REGISTER_ERROR?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -21,37 +21,38 @@
...
-  // Status of each cache entry
-  // NOTE: never change the ordinal of the enum values because the enum values are stored in cache
+public class PersistedFid {
+  // Status of each presisted fid entry
+  // NOTE: never change the ordinal of the enum values because the enum values are stored in shared
+  // prefs
   // as their ordinal numbers.
-  public enum CacheStatus {
-    // Cache entry is synced to Firebase backend
+  public enum PersistedStatus {
```
andirayo: seems like this enum should be called "RegistrationStatus"

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -21,37 +21,38 @@
...
     REGISTERED,
-    // Cache entry is waiting for Firebase backend response or internal network retry
+    // Persisted Fid entry is waiting for Firebase backend response or internal network retry
     UNREGISTERED,
-    // Cache entry is in error state when syncing with Firebase backend
+    // Persisted Fid entry is in error state when syncing with Firebase backend
     REGISTER_ERROR,
-    // Cache entry is in delete state before syncing with Firebase backend
+    // Persisted Fid entry is in delete state before syncing with Firebase backend
     DELETED
```
andirayo: As you mentioned below, I dont think we need this.
But doesn't hurt for now

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -21,37 +21,38 @@
...
+  // prefs
   // as their ordinal numbers.
-  public enum CacheStatus {
-    // Cache entry is synced to Firebase backend
+  public enum PersistedStatus {
+    // Persisted Fid entry is synced to Firebase backend
     REGISTERED,
-    // Cache entry is waiting for Firebase backend response or internal network retry
+    // Persisted Fid entry is waiting for Firebase backend response or internal network retry
     UNREGISTERED,
```
andirayo: Don't we need a status "IN_PROGRESS" or "PENDING" for when the SDK is waiting for a request to the FIS server to complete?
Without that, you might have race conditions, where one call to #getId triggers a request to FIS and a second call to #getId triggers another request. The first one succeeds, the second one returns an error...

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -21,37 +21,38 @@
...
     DELETED
   }
 
-  private static final String SHARED_PREFS_NAME = "FiidCache";
+  private static final String SHARED_PREFS_NAME = "PersistedFid";
 
   private static final String FIREBASE_INSTALLATION_ID_KEY = "Fid";
   private static final String AUTH_TOKEN_KEY = "AuthToken";
   private static final String REFRESH_TOKEN_KEY = "RefreshToken";
   private static final String TOKEN_CREATION_TIME_IN_SECONDS_KEY = "TokenCreationEpochInSecs";
```
andirayo: This is usually called a "Timestamp".
Any reason to prefer "Creation Time in Seconds" over "Timestamp"?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -62,30 +63,38 @@ public FiidCache(@NonNull FirebaseApp firebaseApp) {
...
+    String fid = prefs.getString(getSharedPreferencesKey(FIREBASE_INSTALLATION_ID_KEY), null);
+    int status = prefs.getInt(getSharedPreferencesKey(PERSISTED_STATUS_KEY), -1);
     String authToken = prefs.getString(getSharedPreferencesKey(AUTH_TOKEN_KEY), null);
     String refreshToken = prefs.getString(getSharedPreferencesKey(REFRESH_TOKEN_KEY), null);
     long tokenCreationTime =
         prefs.getLong(getSharedPreferencesKey(TOKEN_CREATION_TIME_IN_SECONDS_KEY), 0);
     long expiresIn = prefs.getLong(getSharedPreferencesKey(EXPIRES_IN_SECONDS_KEY), 0);
 
-    if (iid == null || status == -1) {
+    if (fid == null || status == -1) {
```
andirayo: Why is this not a constant?
What does -1 mean?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
andirayo: I like it.
Or also
FIID_PREF_KEYS.map(this::getSharedPreferencesKey).each(editor::remove).

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -14,10 +14,37 @@
...
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.when;
+
+import androidx.test.core.app.ApplicationProvider;
 import androidx.test.runner.AndroidJUnit4;
+import com.google.android.gms.tasks.Tasks;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.FirebaseOptions;
+import com.google.firebase.installations.local.PersistedFid;
+import com.google.firebase.installations.local.PersistedFidEntry;
```
andirayo: What is the difference between PersistedFid and PersistedFidEntry?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -21,37 +21,38 @@
...
 /**
- * A layer that locally caches a few Firebase Installation attributes on top the Firebase
- * Installation backend API.
+ * A layer that locally persists a few Firebase Installation attributes on top the Firebase
+ * Installation API.
  */
-public class FiidCache {
-  // Status of each cache entry
-  // NOTE: never change the ordinal of the enum values because the enum values are stored in cache
+public class PersistedFid {
```
andirayo: How about calling this LocalStorage or FisStorage or FirebaseInstallationsStorage?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFidEntry.java
```diff
@@ -0,0 +1,78 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+
+/**
+ * This class represents a persisted fid entry in {@link PersistedFid}, which contains a few
+ * Firebase Installation attributes and the persisted status of this entry.
+ */
+@AutoValue
+public abstract class PersistedFidEntry {
```
andirayo: Please feel free to ignore this comment.
Maybe, it's easier to discuss in person and just submit this PR with this name.
I'm just wondering:
Why can't we just call this "FirebaseInstallation"?
(or "PersistedFirebaseInstallation" in case you want to make it obvious that it's received from local storage)
Let's discuss in person after this is submitted.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: I was referring to FLoC sdk while implementing the executor. I am not sure whats the best approach to decide the thread pool count.

If I just inject the executor, the user of the sdk has to instantiate the executor class. This might be a wrong question, but, how will they know the right number of threads required? Or, how does this improve the performance?

ciarand: do you actually need multiple threads? a cached executor service w/ a cap of one thread and ~30s idle time should be sufficient, no? You're never actually running more than one thing at a time right?

>If I just inject the executor, the user of the sdk has to instantiate the executor class.

I'm suggesting that you move the instantiation of the executor service into the delegating constructor, which ensures there's at least one constructor where you can inject a test-controllable executor.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: https://help.github.com/en/articles/creating-and-highlighting-code-blocks

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: I will be running more than one task at a time( Line 96 and Line 170 ). Task for getId and another task for registering FID. 

Do you recommend using newCachedThreadPool() or have a custom executor

 new ThreadPoolExecutor(0, 2,
                                      30L, TimeUnit.SECONDS,
                                      new SynchronousQueue<Runnable>());

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: what do you all think about something like this? I haven't checked that it'll compile, but it seems to clearly express what we're trying to do:

1. return the existing fid if it's there
2. create one if it didn't exist
3. trigger a background registration if it wasn't previously registered

```java
public Task<FID> getFid() {
  return getExistingFid()
    .continueWith(orElse(this::createAndPersistNewFid))
    .onSuccessTask(this::registerFidIfNecessary);
}

public static <F, T> Continuation<F, T> orElse(Supplier<T> supplier) {
  return t -> {
    if (t.isSuccessful()) { return t.getResult(); }
    return supplier.get();
  };
}

private FID createAndPersistNewFid() {
  FID fid = Utils.createRandomFid();
  persist(fid);
  return fid;
}

private void registerFidIfNecessary(FID fid) {
  // do network stuff
}
```

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
ciarand: +1

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
ciarand: we pass -1 as a default in the call to prefs.getInt

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: @ciarand I tried incorporating the changes you mentioned below:

```
  @RequiresApi(api = Build.VERSION_CODES.N)
  private Task<String> getFid() {
    return Tasks.call(executor, this::getExistingFid)
        .continueWith(orElse(new NewFidSupplier()))
        .onSuccessTask(this::registerFidIfNecessary);
  }

  private PersistedFidEntry getExistingFid() throws FirebaseInstallationsException {
    PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
    if (persistedFidMissingOrInErrorState(persistedFidEntry)) {
      throw new FirebaseInstallationsException(
          "Failed to get existing fid.", FirebaseInstallationsException.Status.CLIENT_ERROR);
    }
    return persistedFidEntry;
  }

  @RequiresApi(api = Build.VERSION_CODES.N)
  @NonNull
  public static <F, T> Continuation<F, T> orElse(@NonNull Supplier<T> supplier) {
    return t -> {
      if (t.isSuccessful()) {
        return (T) t.getResult();
      }
      return supplier.get();
    };
  }

  @RequiresApi(api = Build.VERSION_CODES.N)
  class NewFidSupplier implements Supplier<PersistedFidEntry> {

    @Override
    public PersistedFidEntry get() {
      return createAndPersistNewFid();
    }
  }

  private PersistedFidEntry createAndPersistNewFid() {
    String fid = Utils.createRandomFid();
    try {
      persistFid(fid);
      PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
      return persistedFidEntry;
    } catch (FirebaseInstallationsException e) {
      // Exception handling
    }
    return PersistedFidEntry.builder().build();
  }

  private void persistFid(String fid) throws FirebaseInstallationsException {
    boolean firstUpdateCacheResult =
        persistedFid.insertOrUpdatePersistedFidEntry(
            PersistedFidEntry.builder()
                .setFirebaseInstallationId(fid)
                .setRegistrationStatus(PersistedFid.RegistrationStatus.UNREGISTERED)
                .build());

    if (!firstUpdateCacheResult) {
      throw new FirebaseInstallationsException(
          "Failed to update client side cache.",
          FirebaseInstallationsException.Status.CLIENT_ERROR);
    }
  }

  private Task<String> registerFidIfNecessary(PersistedFidEntry persistedFidEntry) {
    if (persistedFidEntry.getRegistrationStatus() != PersistedFid.RegistrationStatus.REGISTERED) {
      Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry));
    }
    return Tasks.forResult(persistedFidEntry.getFirebaseInstallationId());
  }

  /** Registers the created Fid with FIS Servers and update the shared prefs. */
  private Void registerAndSaveFid(PersistedFidEntry persistedFidEntry)
      throws FirebaseInstallationsException {
    // network call
  }
```

ankitaj224: I have 2 concerns: 

- Supplier get call requires API level 24, Is this right way of handling it? 
- Exception handling by the Supplier while persisting the newly created fid

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: You're loading two separate jobs onto the same executor, but they don't run in parallel right? The getId call runs first, then the registration call runs asynchronously after the getId call finishes.

A custom executor is probably fine. `new ThreadPoolExecutor(0, 1, 30L, TimeUnit.SECONDS, new SynchronousQueue<>());`

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: Yeah we can't use the standard java.util.functional.Supplier interface. We can just define one in the parent class:

```
interface Supplier<T> {
  T get() throws Exception;
}
```

When you say "Exception handling by the Supplier while persisting the newly created fid" do you mean you're worried that you can't propagate exceptions or you're worried that they will be propagated?

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -26,4 +59,171 @@
...
+    assertWithMessage("getId Task fails.")
+        .that(Tasks.await(firebaseInstallations.getId()))
+        .isNotEmpty();
+    PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
+    assertWithMessage("Persisted Fid doesn't match")
+        .that(entryValue.getFirebaseInstallationId())
+        .isEqualTo(TEST_FID_1);
+    assertWithMessage("Registration status doesn't match")
+        .that(entryValue.getRegistrationStatus())
+        .isEqualTo(PersistedFid.RegistrationStatus.PENDING);
```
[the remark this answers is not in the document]
ankitaj224: Honestly, nothing. I thought about it. But the test runs have been consistent with registration taking some time. Is there a better way to test it or do i skip the status check altogether.

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -78,7 +89,7 @@ public synchronized PersistedFidEntry readPersistedFidEntryValue() {
 
     return PersistedFidEntry.builder()
         .setFirebaseInstallationId(fid)
-        .setPersistedStatus(PersistedStatus.values()[status])
+        .setRegistrationStatus(RegistrationStatus.values()[status])
```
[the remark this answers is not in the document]
ankitaj224: Done. https://github.com/firebase/firebase-android-sdk/pull/703/commits/f8e75fd4bbf0b83b84fa71bc360501c862e3d89c

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Done. https://github.com/firebase/firebase-android-sdk/pull/703/commits/f8e75fd4bbf0b83b84fa71bc360501c862e3d89c

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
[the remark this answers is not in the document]
ankitaj224: @ciarand Your thoughts.

firebase-installations/src/androidTest/java/com/google/firebase/installations/local/PersistedFidTest.java
```diff
@@ -56,60 +56,89 @@ public void setUp() {
...
-    PersistedFidEntry entryValue = cache0.readPersistedFidEntryValue();
-    assertThat(entryValue.getFirebaseInstallationId()).isEqualTo(TEST_FID_1);
-    assertThat(entryValue.getAuthToken()).isEqualTo(TEST_AUTH_TOKEN);
-    assertThat(entryValue.getRefreshToken()).isEqualTo(TEST_REFRESH_TOKEN);
-    assertThat(entryValue.getPersistedStatus())
-        .isEqualTo(PersistedFid.PersistedStatus.UNREGISTERED);
-    assertThat(entryValue.getExpiresInSecs()).isEqualTo(TEST_TOKEN_EXPIRATION_TIMESTAMP);
-    assertThat(entryValue.getTokenCreationEpochInSecs()).isEqualTo(TEST_CREATION_TIMESTAMP_1);
-    assertNull(cache1.readPersistedFidEntryValue());
+    PersistedFidEntry entryValue = persistedFid0.readPersistedFidEntryValue();
```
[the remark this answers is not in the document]
ciarand: **so** much nicer!

PR #705, opened 2019-08-14
Early exit when stopListening after shutdown.

wu-hui:


mikelehen:
LOL.  My comment makes no sense.  I thought this was porting to a different platform.  I see what happened now.

mikelehen:
As a sidenote, I think the early return is unnecessary since the AsyncQueue will silently just not run the code after shutdown, but it's also non harmful to make it more explicit that we want to allow this.

wu-hui:
> As a sidenote, I think the early return is unnecessary since the AsyncQueue will silently just not run the code after shutdown, but it's also non harmful to make it more explicit that we want to allow this.

I want to do this because it makes the intention clear: this method does not call `verifyNotShutdown` like many others.

PR #707, opened 2019-08-14
Schedule active context jobs upon initialization(b/139435527).

vkryachko:


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/WorkInitializer.java
```diff
@@ -0,0 +1,49 @@
...
+    this.guard = guard;
+  }
+
+  public void ensureContextsScheduled() {
+    executor.execute(
+        () ->
+            guard.runCriticalSection(
+                () -> {
+                  for (TransportContext context : store.loadActiveContexts()) {
+                    scheduler.schedule(context, 1);
```
ashwinraghav: I guess the attempt number here is "reset". Worth leaving a comment

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -274,6 +274,28 @@ public void recordNextCallTime(TransportContext transportContext, long timestamp
         });
   }
 
+  @Override
+  public Iterable<TransportContext> loadActiveContexts() {
+    return inTransaction(
```
ashwinraghav: Thinking out loud:
What guarantees for newly updating clients that this query is run after the client migrates to V2 schema ? If it were to run before, it may not find the extras column ?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -286,6 +308,23 @@ public void close() {
     schemaManager.close();
   }
 
+  @VisibleForTesting
```
ashwinraghav: Is this operation only for testing or is only the visibility adjusted ? Can we move it to the tests themselves ?

vkryachko: It's only needed in tests, I replaced it with `@RestrictTo(TESTS)` which is more strict and warns for any use of the method in SDK code. It is problematic to move the code to the test itself as the implementation of the method uses internal state of the `store` that I'd rather not expose as public.

PR #713, opened 2019-08-16
Propagate error for ListAll

schmidt-sebastian:
Turns out that we didn't propagate backend errors for ListAll, instead the Task would just hang.

Fixes https://b.corp.google.com/issues/139533387

### DIFF
--- a/firebase-storage/test-app/src/main/java/com/example/storage/MainActivity.java
+++ b/firebase-storage/test-app/src/main/java/com/example/storage/MainActivity.java
@@ -40,6 +40,7 @@ import com.google.android.gms.common.internal.Preconditions;
 import com.google.android.gms.tasks.Continuation;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.firebase.FirebaseApp;
 import com.google.firebase.auth.FirebaseAuth;
 import com.google.firebase.storage.FirebaseStorage;
 import com.google.firebase.storage.TestCommandHelper;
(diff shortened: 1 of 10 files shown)


firebase-storage/test-app/test-app.gradle
```diff
@@ -67,4 +67,6 @@ dependencies {
 // androidTest integration tests.
 // ==========================================================================
 ext.packageName = "com.example.storage"
-apply from: '../../gradle/googleServices.gradle'
\ No newline at end of file
+apply from: '../../gradle/googleServices.gradle'
+
+apply plugin: "com.google.gms.google-services"
```
schmidt-sebastian: I needed to add this line for the Test App to start. I hope it doesn't beak CI.

schmidt-sebastian: I changed this to call initializeApp because it does break CI.

schmidt-sebastian:
@fredzqm Let me know if you feel comfortable reviewing this.

firebase-storage/test-app/src/main/java/com/example/storage/MainActivity.java
```diff
@@ -104,6 +105,7 @@ protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     setContentView(R.layout.activity_main);
 
+    FirebaseApp.initializeApp(getApplicationContext());
```
fredzqm: OOC, why this line weren't there before.

schmidt-sebastian: This lets us use the test app without the Google Services plugin, which we had to remove as it breaks the CI build.

PR #714, opened 2019-08-19
Download metalava jar before running the command

VinayGuthal:


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -70,7 +73,11 @@ public void apply(Project project) {
               });
     }
     if (System.getenv().containsKey("METALAVA_BINARY_PATH")) {
```
vkryachko: Seems like you no longer need this ENV check, right?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -70,7 +73,11 @@ public void apply(Project project) {
...
-      setupApiInfomrationAnalysis(project, android);
+      try {
+        setupApiInformationAnalysis(project, android);
+      } catch (IOException e) {
```
vkryachko: what's the purpose of this? can you remove the catch? also why does this method need to throw IOException?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/apiinfo/ApiInformationTask.java
```diff
@@ -82,8 +82,10 @@ void execute() {
 
         // Generate api.txt file and store it in the  build directory.
         getProject().exec(spec-> {
-            spec.setCommandLine(Arrays.asList(
-                getMetalavaBinaryPath(),
+            spec.setExecutable("java");
```
vkryachko: this is not portable, pls use `javaExec` instead

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/apiinfo/GetMetalavaJarTask.java
```diff
@@ -0,0 +1,56 @@
...
+  public abstract void setOutputFile(File outputFile);
+
+  @TaskAction
+  void execute() {
+    if (getOutputFile().exists()) {
+      return;
+    }
+    try {
+      URL jarUrl = new URL("https://storage.googleapis.com/android-ci/metalava-full-1.3.0-SNAPSHOT.jar");
+      Files.copy(jarUrl.openStream(), getOutputFile().toPath(), StandardCopyOption.REPLACE_EXISTING);
```
vkryachko: I don't think this will close the input stream once done, pls make sure it is closed.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
VinayGuthal: Done

PR #715, opened 2019-08-19
Index-Free: Add ReadTime backfill

schmidt-sebastian:
This adds the logic needed to backfill the read time for existing documents in the RemoteDocumentCache. The code migrates 100 documents at startup and then stops until the next time the client starts. To persist the current state, we keep a high watermark for the first document that we need to migrate.

It might make sense to run this as a delayed operation in order not to block client startup, but I left that as an exercise for the reader for now. Let me know if you think we should do that.


schmidt-sebastian:
Added periodic scheduling. This turned out to be a lot more work than anticipated since some query tests restart persistence, which then schedules two of the same Timer IDs on the AsyncQueue.

schmidt-sebastian:
Udpated once more to cancel the migration on shutdown. Unfortunately, we call it from the tests, so I removed the DelayedTask.cancel assert that makes sure we only call it from the AsyncQueue

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDataBackfill.java
```diff
@@ -0,0 +1,177 @@
...
+                      LOG_TAG,
+                      "Failed to decode document for key %s during read time backfill",
+                      EncodedPath.decodeFieldPath(encodedCurrentPath));
+                }
+              });
+
+          SQLiteStatement updateStatement =
+              db.compileStatement("UPDATE target_globals SET first_document_without_read_time = ?");
+          if (rowCount[0] == BACKFILL_MIGRATION_SIZE && lastMigratedKey[0] != null) {
+            String nextKey = EncodedPath.prefixSuccessor(lastMigratedKey[0]);
```
mikelehen: I'm a little nervous about this prefixSuccessor stuff.  Is there a reason we couldn't just store the last_migrated_key and then do a `>` query instead of `>=` ?

Maybe the issue is differentiating the "not-started" from "finished" state?  Perhaps we could:
* Store "" initially to indicate that we need to migrate the first doc with key > "".
* Store valid keys while migration is in progress.
* Store NULL once the migration is done.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -373,6 +367,22 @@ private void addReadTime() {
...
+  private void addReadTimeMigrationWatermark() {
+    if (!tableContainsColumn("target_globals", "first_document_without_read_time")) {
+      db.execSQL("ALTER TABLE target_globals ADD COLUMN first_document_without_read_time TEXT");
+    }
+
+    SQLiteStatement insertFirstMissingReadTime =
+        db.compileStatement(
+            "UPDATE target_globals SET first_document_without_read_time = ("
+                + "SELECT path FROM remote_documents "
+                + "WHERE read_time_seconds IS NULL ORDER BY path LIMIT 1)");
```
mikelehen: Per my comment elsewhere, we could probably simplify and just store "" initially.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
```diff
@@ -113,7 +116,6 @@ void skipDelay() {
      * the AsyncQueue) provides a guarantee that the task will not be run.
      */
     public void cancel() {
-      verifyIsCurrentThread();
```
mikelehen: Hrm...  This assert is valid and serves an important purpose (ensures that we can provide the guarantee described in the method comment).  So I am not wild about removing it rather than figuring out a way to fix the tests (If we make SqlitePersistence AsyncQueue-aware then we really should be calling all of its methods on the AsyncQueue).

Worst case if we want a cheap fix maybe we can try/catch the exception from tests?  Else we need some sort of better plumbing around PersistenceTestHelpers so we can recover the AsyncQueue or something...

schmidt-sebastian:
RIP

PR #716, opened 2019-08-20
Use list of files instead of string

VinayGuthal:


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/apiinfo/GenerateApiTxtFileTask.java
```diff
@@ -35,7 +38,7 @@
     abstract File getApiTxt();
 
     @Input
```
vkryachko: I think you want `@InputFiles`, otherwise gradlew will cache by file path, not contents.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/apiinfo/ApiInformationTask.java
```diff
@@ -44,7 +45,7 @@
     abstract File getApiTxt();
 
     @Input
```
vkryachko: Same gere

PR #720, opened 2019-08-20
Don't persist documents that we already have

schmidt-sebastian:
Port of https://github.com/firebase/firebase-js-sdk/pull/2099.

Also pulls in spec test changes from https://github.com/firebase/firebase-js-sdk/pull/2098


firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -395,8 +396,16 @@ public static RemoteEvent noChangeEvent(int targetId, int version, ByteString re
 
   public static RemoteEvent addedRemoteEvent(
       MaybeDocument doc, List<Integer> updatedInTargets, List<Integer> removedFromTargets) {
-    DocumentChange change =
-        new DocumentChange(updatedInTargets, removedFromTargets, doc.getKey(), doc);
+    return addedRemoteEvent(Collections.singletonList(doc), updatedInTargets, removedFromTargets);
```
schmidt-sebastian: Note: the changes here are also part of https://github.com/firebase/firebase-android-sdk/pull/697/files#diff-ac4f307b0804f62b9719e8c7556226c2L396

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -329,50 +329,35 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
     return persistence.runTransaction(
         "Apply remote event",
         () -> {
+          SnapshotVersion remoteVersion = remoteEvent.getSnapshotVersion();
```
var-const: Question: this variable is now much further from its first point of use, and I didn't notice any references to it in-between. Perhaps move it back to the end of this function?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -329,50 +329,35 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
...
-            if (!resumeToken.isEmpty()) {
-              QueryData oldQueryData = queryData;
-              queryData =
-                  queryData.copy(remoteEvent.getSnapshotVersion(), resumeToken, sequenceNumber);
-              targetIds.put(boxedTargetId, queryData);
-
-              if (shouldPersistQueryData(oldQueryData, queryData, change)) {
-                queryCache.updateQueryData(queryData);
+            QueryData oldQueryData = targetIds.get(targetId);
+            if (oldQueryData != null) {
```
var-const: Optional: I personally liked the `continue` approach that was in place before (this function is long, and I feel the increased nestedness hurts readability).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -329,50 +329,35 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
...
+              // Only update the remote keys if the query is still active. This ensures that we can
+              // persist the updated query data along with the updated assignment.
+              queryCache.removeMatchingKeys(change.getRemovedDocuments(), targetId);
+              queryCache.addMatchingKeys(change.getAddedDocuments(), targetId);
+
+              // Update the resume token if the change includes one. Don't clear any preexisting
```
var-const: Question: which code is responsible for `Don't clear any preexisting value` -- is it `shouldPersistQueryData`?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -329,50 +329,35 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
...
           long sequenceNumber = persistence.getReferenceDelegate().getCurrentSequenceNumber();
-          Set<DocumentKey> authoritativeUpdates = new HashSet<>();
 
-          Map<Integer, TargetChange> targetChanges = remoteEvent.getTargetChanges();
           for (Map.Entry<Integer, TargetChange> entry : targetChanges.entrySet()) {
             Integer boxedTargetId = entry.getKey();
             int targetId = boxedTargetId;
             TargetChange change = entry.getValue();
 
-            // Do not ref/unref unassigned targetIds - it may lead to leaks.
```
var-const: Does this comment no longer apply?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -391,10 +376,9 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
             MaybeDocument existingDoc = existingDocs.get(key);
 
             if (existingDoc == null
-                || (authoritativeUpdates.contains(doc.getKey()) && !existingDoc.hasPendingWrites())
-                || doc.getVersion().compareTo(existingDoc.getVersion()) >= 0) {
-              // If a document update isn't authoritative, make sure we don't apply an old document
-              // version to the remote cache.
+                || doc.getVersion().compareTo(existingDoc.getVersion()) > 0
+                || (doc.getVersion().compareTo(existingDoc.getVersion()) == 0
+                    && existingDoc.hasPendingWrites())) {
```
var-const: Hmm, why do we overwrite the existing doc when it has pending writes?

firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -395,8 +396,16 @@ public static RemoteEvent noChangeEvent(int targetId, int version, ByteString re
...
+  }
+
+  public static RemoteEvent addedRemoteEvent(
+      List<MaybeDocument> docs, List<Integer> updatedInTargets, List<Integer> removedFromTargets) {
+    Preconditions.checkArgument(!docs.isEmpty(), "Cannot pass empty docs array");
+
+    ResourcePath collectionPath = docs.get(0).getKey().getPath().popLast();
+    SnapshotVersion version = docs.get(0).getVersion();
```
var-const: Optional: move this declaration to the point right before it's used?

firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -407,11 +416,16 @@ public static RemoteEvent addedRemoteEvent(
 
               @Override
               public QueryData getQueryDataForTarget(int targetId) {
-                return queryData(targetId, QueryPurpose.LISTEN, doc.getKey().toString());
+                return queryData(targetId, QueryPurpose.LISTEN, collectionPath.toString());
```
var-const: Question: what's the reason for this change?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
var-const: This is the only important question, the rest are nitpicks.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: It's used in line 354. I moved it to the top of the function body - let me know if you think that's better. I can otherwise also inline the usage in line 354.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: Sounds good. Reverted.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: Yes. I removed that part of the comment since it is a bit confusing.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: When the write stream sends us an acknowledgment, we persist a document with `hasCommittedMutations`. We then pretend we know the state of the document, but that is not always true (another client may have changed a field we didn't write to, or we supplied an ArrayUnion/ArrayRemove transform that was applied to a set of elements that the client didn't yet know about). By allowing the server to send us the state of a committed document, we overwrite the `hasCommittedMutations` version with the actual version from the backend.

firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
schmidt-sebastian: Done.

firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
schmidt-sebastian: The new test passes multiple docs to `addedRemoteEvent`. The test helper now uses the collection path of the first doc to represent the query.

FWIW I also believe the original implementation should have returned a collection path here.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: I rewrote it below since I didn't quite understand it at first.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
var-const: Thanks for the explanation!

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
var-const: Ah, cool. I also found the original comment confusing.

PR #722, opened 2019-08-20
Migrate to the new AndroidChannelBuilder.usingBuilder API

wilhuff:


wilhuff:
Note this won't work yet because the corresponding API hasn't been released.

PR #724, opened 2019-08-21
Update API text file

schmidt-sebastian:
This should remove the additional messages for every Firestore PR. InstanceRegistry was never officially part of our public API.


schmidt-sebastian:
See https://github.com/firebase/firebase-android-sdk/commit/3057375ccfc848f99e390c4388a96d79b2b5ebbc for context

VinayGuthal:
In the message on the prs. if its intended You can update it by running ./gradlew firebase-firestore:generateApiTxtFile

PR #726, opened 2019-08-22
Index-free: Use different SQL for non-index free queries

schmidt-sebastian:
I blame this one on @mikelehen 

The basic thought process behind is that we don't need to backfill any read times. When a user upgrades, none of the QueryData will have a lastLimboFreeSnapshot version set. Once we have a lastLimboFreeSnapshot, we know that the target mapping can be relied upon. Any changes since the target mapping will have a read time.

Thoughts? Fears?


mikelehen:
LOL...  I just didn't want to have to finish reviewing the migration code.

PR #728, opened 2019-08-22
Whitelist lint warning for missing tools:targetApi attr.

rlazo:


firebase-common/lint-baseline.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="5" by="lint 3.4.1" client="gradle" variant="all" version="3.4.1">
+
```
vkryachko: Pls add a todo here to undo this once bazel supports the attribute

rlazo: Done

PR #730, opened 2019-08-23
Enable Index-Free queries

schmidt-sebastian:


firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
-# Unreleased
+# Unreleased (21.1.0)
+- [changed] With disk persistence enabled, query processing time for most
+  queries is now mainly dependent on the size of the result set rather than the
+  size of the locally cached collection. Queries that use the `limit()` API as
+  well as queries that have not been recently executed may not benefit from
+  this change.
+  If you are updating to v21.1.0, only queries that have been synchronized with
```
mikelehen: I don't think you need "If you are updating to v21.1.0".  Even for a brand-new user, the first time running the query will always be the slow path.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,13 @@
-# Unreleased
+# Unreleased (21.1.0)
+- [changed] With disk persistence enabled, query processing time for most
+  queries is now mainly dependent on the size of the result set rather than the
```
mikelehen: I was worried about over-promising and now you're making an even stronger promise. 😄   I think this needs some caveat like "repeatedly-executed queries" or something, since the first time you do a query it will still be slow.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -264,8 +264,7 @@ protected void specTearDown() throws Exception {
    */
   private void initClient() {
     localPersistence = getPersistence(garbageCollectionEnabled);
-    // TODO(index-free): Update to index-free query engine when it becomes default.
```
mikelehen: There's one other TODO which I think is actually just stale / wrong.

https://github.com/firebase/firebase-android-sdk/blob/209285735903837d12d0af476132f6415edf9e95/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java#L49

mikelehen:
By the way, I wonder if we should consider any sort of "debug-only" API to opt into the old QueryEngine for a while, to help us debug reports in case people report unexpected query results or something.

firebase-firestore/CHANGELOG.md
schmidt-sebastian: I tried walk back on the promise in the second sentence. To make it clearer, I changed it back to also use "repeatedly executed".

firebase-firestore/CHANGELOG.md
schmidt-sebastian: To me, this was part of "have not been recently executed".  I did end up combining it with the previous list of caveats.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
schmidt-sebastian: This was going to go away with my schema backfill PR. But instead that went away...

firebase-firestore/CHANGELOG.md
```diff
@@ -1,11 +1,9 @@
...
-  queries is now mainly dependent on the size of the result set rather than the
-  size of the locally cached collection. Queries that use the `limit()` API as
-  well as queries that have not been recently executed may not benefit from
-  this change.
-  If you are updating to v21.1.0, only queries that have been synchronized with
-  the backend at least once will see improved performance.
+- [changed] With disk persistence enabled, query processing time for repeatedly
+  executed queries is now mainly dependent on the size of the result set rather
+  than the size of the locally cached collection. Queries that use the
+  `limit()` API as well as queries that have never or have not recently been
```
mikelehen: "have never or have not recently been" is probably easier to read as "have not recently (or ever) been"...

And yes, I realize now that "never" is a subset of "not recently", but thank you for being explicit for me. :-)

I'll let Gil re-tear it up now.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-# Unreleased
+# Unreleased (21.1.0)
+- [changed] With disk persistence enabled, query processing time for repeatedly
```
wilhuff: Elsewhere in this doc we refer to this as "on-disk persistence". Make this consistent? However, I wonder: is this even necessary to call out? Does the IndexFreeQueryEngine special case for SQLite?

Elsewhere we also just come out and say "Improved performance of queries" maybe we should structure this similarly?

Also, limits aren't problematic directly--they only cause problems when documents are deleted.

Finally, "query processing time ... now mainly dependent on the size of the result set" is awkward and still inaccurate because our behavior is O(last result set + changes since last execution). I suggest being less specific.

Spitballing:

Improved the performance of repeatedly executed queries. Recently executed queries should see dramatic improvements. This benefit is reduced over time as changes accumulate. Queries with limits where documents are deleted from the results may not benefit.

mikelehen: > Also, limits aren't problematic directly--they only cause problems when documents are deleted.

It's much worse than that.  We have to "refill" the limit if:
1. You delete an item in the query (as you said).
2. You modify an item such that it no longer matches the query (since a new item will get pulled in).
3. You modify an item such that it still matches the query but it's order changed so it may no longer actually be in the top N of the query anymore.

So I think right now (based on discussions with Sebastian), if a document from the target-document mapping has been modified in any way, then we fall back to doing the full collection scan.

schmidt-sebastian: I changed this to:

```
- [changed] Improved performance for local query execution. For queries
  that have recently been synchronized with the backend, execution time now
  depends on the size result set rather than on the size of the locally cached
  collection.
  Queries that use the `limit()` API may not see improved performance if an
  unsynchronized document update changes the order of the result set.
  Furthermore, execution time for all queries degrades if the queried
  collection received frequent document updates while the query was inactive.
```

The technically correct version is:

"Queries that use the `limit()` API may not see improved performance if an
 unsynchronized document update has the potential to change the order of the result set."

mikelehen: Uhhh...  I hate to dwell on this changelog entry too much (ahh, screw it, we're way past that now, let's double-down!), but ...

>   Queries that use the `limit()` API may not see improved performance if an unsynchronized document update changes the order of the result set.

seems like too much information and also...  not correct? Or am I confused?  I assumed what we implemented is "Queries that use limit() won't see improved performance if we've cached a newer version of any document in the results since the last time you did the query."  So for example if you read or write a new version of any document that was previously in the result set, then next time you do the query, we have to do a full scan, right?  Maybe I should look at the code. :-)

I also think we want to avoid making this too detailed / technical in general.  I liked Gil's version.  So maybe just:

> Improved the performance of repeatedly executed queries. Recently executed queries should see dramatic improvements. This benefit is reduced over time as changes accumulate. Queries with limits may not always benefit, depending on the accumulated changes.

schmidt-sebastian: Updated to match basically what you suggested.

And you don't need to look at the code since your understanding is 100% correct (as always :) )

PR #732, opened 2019-08-23
Rename shutdown to terminate and make it public.

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -333,37 +333,36 @@ public WriteBatch batch() {
...
-    return client.shutdown();
+    return client.terminate();
   }
 
   /**
-   * Shuts down this {@code FirebaseFirestore} instance.
+   * Terminates this {@code FirebaseFirestore} instance.
    *
-   * <p>After shutdown only the {@link #clearPersistence()} method may be used. Any other method
+   * <p>After terminate only the {@link #clearPersistence()} method may be used. Any other method
```
mikelehen: I think this should be "termination" or maybe "calling {@code terminate()}".  I think I prefer the latter to be more explicit.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -333,37 +333,36 @@ public WriteBatch batch() {
...
    * will throw an {@link IllegalStateException}.
    *
-   * <p>To restart after shutdown, simply create a new instance of {@code FirebaseFirestore} with
+   * <p>To restart after terminate, simply create a new instance of {@code FirebaseFirestore} with
```
mikelehen: same here

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -333,37 +333,36 @@ public WriteBatch batch() {
...
    * {@link #getInstance()} or {@link #getInstance(FirebaseApp)}.
    *
-   * <p>Shutdown does not cancel any pending writes and any tasks that are awaiting a response from
-   * the server will not be resolved. The next time you start this instance, it will resume
-   * attempting to send these writes to the server.
+   * <p>{@code terminate} does not cancel any pending writes and any tasks that are awaiting a
```
mikelehen: When referencing a method, I'd always include parens (i.e. "terminate()").

firebase-firestore/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 - [changed] Transactions now perform exponential backoff before retrying.
   This means transactions on highly contended documents are more likely to
   succeed.
+- [feature] Added a `terminate` method to `FirebaseFirestore` which can
```
mikelehen: We usually include parens when referencing methods, i.e. "terminate()"...

firebase-firestore/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+  terminate the Firestore SDK instance and release the resources held by
+  the instance.
```
mikelehen: nit: I'd avoid saying "instance" twice and just say "terminate the Firestore SDK instance and release any held resources."

Though I also wonder if we should mention clearPersistence(), since that was the motivation for this method.  Something like:

> Added a `terminate()` method to `FirebaseFirestore` which terminates the instance, releasing any held resources. Once it completes, you can optionally call `clearPersistence()` to wipe persisted Firestore data from disk.

wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
wu-hui: Done.

firebase-firestore/CHANGELOG.md
wu-hui: Done.

PR #734, opened 2019-08-23
Index-free: Drop last limbo free snapshot if re-upgraded

schmidt-sebastian:
If we don't want to backfill the read time, we need to ensure that all documents written after any last limbo free snapshot version need to already have a read time. Since that is not the case if a user downgrades, we should clear all limbo free snapshot versions when the user upgrades again.


schmidt-sebastian:
`check-changed` failure is unrelated.

PR #748, opened 2019-08-28
Migrate apk size table to metrics database.

yifanyang:
This PR is part of the effort of migrating apk size measurement database to the same one used by ios/android project health metrics.

New tables have been created with a different name (with the prefix of "Android"). This change accommodates to the table name change on the database server side. The PR that switches the database to upload is [here](https://team-review.git.corp.google.com/c/firebase-android-core/test-infra/+/537770). These two PRs will need to be submitted together.

Records in the old database/table will later be migrated to the new database/table manually.


allisonbm92:
1. Can you update the broken apksize tests?

2. Can you add some detail to the description (don't leave it blank). Pull requests like this can be confusing to other developers that aren't familiar with what this affects.

yifanyang:
> 1. Can you update the broken apksize tests?
Done.

> 2. Can you add some detail to the description (don't leave it blank). Pull requests like this can be confusing to other developers that aren't familiar with what this affects.
Done. Also updated the description in the related PR: https://team-review.git.corp.google.com/c/firebase-android-core/test-infra/+/537770.

PR #749, opened 2019-08-29
Re-add tools:targetApi to firebase-common AndroidManifest.

rlazo:
This attr had to be removed in https://github.com/firebase/firebase-android-sdk/pull/728 because a bug in the tooling that has been fixed now.


rlazo:
> Why is it now ok to add?

The bug in the tool is now fixed.

PR #757, opened 2019-08-30
Don't double escape list references

schmidt-sebastian:
Fixes b/139750197

This PR makes sure that we always encode query parameters and always use the decoded path if we process them further.

### DIFF
--- a/firebase-storage/src/androidTest/java/com/google/firebase/storage/IntegrationTest.java
+++ b/firebase-storage/src/androidTest/java/com/google/firebase/storage/IntegrationTest.java
@@ -51,6 +51,8 @@ public class IntegrationTest {
 
   private final String randomPrefix = UUID.randomUUID().toString();
 
+  private final String unicodePrefix = "prefix/\\%:😊 ";
+
   @Before
   public void before() throws ExecutionException, InterruptedException {
     if (storageClient == null) {
@@ -60,6 +62,7 @@ public class IntegrationTest {
       Tasks.await(getReference("metadata.dat").putBytes(new byte[0]));
       Tasks.await(getReference("download.dat").putBytes(new byte[LARGE_FILE_SIZE_BYTES]));
       Tasks.await(getReference("prefix/empty.dat").putBytes(new byte[0]));
+      Tasks.await(getReference(unicodePrefix + "/empty.dat").putBytes(new byte[0]));
     }
   }
 
@@ -73,6 +76,15 @@ public class IntegrationTest {
     assertThat(tempFile.exists()).isTrue();
     assertThat(tempFile.length()).isEqualTo(LARGE_FILE_SIZE_BYTES);
     assertThat(fileTask.getBytesTransferred()).isEqualTo(LARGE_FILE_SIZE_BYTES);
+  }
+
+  @Test
+  public void downloadUnicodeFile() throws ExecutionException, InterruptedException, IOException {
+    File tempFile = new File(Environment.getExternalStorageDirectory(), "empty.dat");
+
+    Tasks.await(getReference(unicodePrefix + "/empty.dat").getFile(tempFile));
+
+    assertThat(tempFile.exists()).isTrue();
   }
 
   @Test
@@ -159,6 +171,19 @@ public class IntegrationTest {
     assertThat(listResult.getItems())
         .containsExactly(getReference("metadata.dat"), getReference("download.dat"));
     assertThat(listResult.getPageToken()).isNull();
+  }
+
+  @Test
+  public void listUnicodeFiles() throws ExecutionException, InterruptedException {
+    Task<ListResult> listTask = getReference("prefix").listAll();
+    ListResult listResult = Tasks.await(listTask);
+
+    assertThat(listResult.getPrefixes()).containsExactly(getReference(unicodePrefix));
+
+    listTask = getReference(unicodePrefix).listAll();
+    listResult = Tasks.await(listTask);
+
+    assertThat(listResult.getItems()).containsExactly(getReference(unicodePrefix + "/empty.dat"));
   }
 
   @NonNull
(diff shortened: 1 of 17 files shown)


schmidt-sebastian:
> The public api surface has changed for the subproject firebase-storage:
> error: Removed method com.google.firebase.storage.StorageException.getMessage() [RemovedMethod]
> 
> Please update the api.txt files for the subprojects being affected by this change by running ./gradlew ${subproject}:generateApiTxtFile. Also perform a major/minor bump accordingly.

This is an overridden method that does the same thing as the method of the super class.

schmidt-sebastian:
Most of my confidence with this PR comes from the fact that any encoding change breaks the existing unit tests. With that said, I am handing this over to @rsgowman

firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java
```diff
@@ -116,18 +115,18 @@ public static String getDefaultURL(@NonNull Uri gsUri) {
         + "/b/"
         + gsUri.getAuthority()
         + "/o/"
-        + (pathWithoutBucket != null ? Slashes.unSlashize(pathWithoutBucket) : "");
+        + (pathWithoutBucket != null ? Uri.encode(pathWithoutBucket) : "");
```
rsgowman: Rather than getDefaultURL() returning a string, could it instead return an android.net.Uri? This would allow you to:
* Be able to use Uri.Builder instead of string manipulation in order to construct the uri.
* Not have to worry about whether the returned string is an encoded or decoded string.

The same argument applies throughout this file. Basically, anywhere a uri is created via string manipulation, you could change it to use Uri.Builder. I think that would make it easier to track encoded/decoded status and ensure valid uris are being generated.

At some point, you'd still have to convert it to a java.net.URL though. But I think the conversion is just:
```
Uri uri = ...;
URL url = new URL(uri.toString());
```
(And I *think* that properly takes care of encoding urls? But definitely best to verify that.)

schmidt-sebastian: I played around with your suggestions and I think I do like it better when Uri handles all encoding. It removes a lot of path and query parameter encoding and moves the responsibility to a class that likely has test coverage for all these things. I am a tiny bit afraid that this might introduce an obscure regression, but it is also not unlikely that the original version of this PR did. Overall, I removed a bunch of Uri to String to Uri conversion so everything should be safer now.

firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java
```diff
@@ -108,40 +106,37 @@ public static String getAuthority() {
...
-  @Nullable
-  public static String getPathWithoutBucket(@NonNull Uri gsUri) {
-    String path = gsUri.getEncodedPath();
-    if (path != null && path.startsWith("/")) {
-      // this should always be true.
-      path = path.substring(1);
+  private static String getPathWithoutBucket(@NonNull Uri gsUri) {
+    String path = gsUri.getPath();
+    if (path == null) {
+      return "";
```
schmidt-sebastian: This change allowed me to remove the nullability handling from two callsites.

rsgowman: Another option would be to throw an exception. (It seems like if path is null here, nothing good will result. But I don't know the code base... maybe it's legit? In any case, returning "" maintains the existing behaviour, so might be better to just leave it that way unless you're sure this is always an error case.) No action required.

firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java
```diff
@@ -194,7 +189,7 @@ protected int getOutputRawSize() {
    * @return query parameters in string form.
    */
   @Nullable
-  protected String getQueryParameters() {
+  protected Map<String, String> getQueryParameters() {
```
rsgowman: nit: update javadoc `@return` string

firebase-storage/src/main/java/com/google/firebase/storage/network/ResumableUploadByteRequest.java
```diff
@@ -31,15 +30,12 @@
   public ResumableUploadByteRequest(
       @NonNull Uri gsUri,
       @NonNull FirebaseApp app,
-      @NonNull String uploadURL,
+      @NonNull Uri uploadURL,
```
rsgowman: The Uri class has a number of verification methods (i.e. isAbsolute(), isHeirarchical(), etc) you could use if you want to verify that it seems reasonable. (I'm not certain what it's *supposed* to look like, so I'm not sure which one(s) would apply). Optional. If you use any of these, also apply to the other Resumable*Request.java files.

firebase-storage/src/androidTest/java/com/google/firebase/storage/IntegrationTest.java
```diff
@@ -75,6 +78,15 @@ public void downloadFile() throws ExecutionException, InterruptedException, IOEx
     assertThat(fileTask.getBytesTransferred()).isEqualTo(LARGE_FILE_SIZE_BYTES);
   }
 
+  @Test
+  public void downloadUnicodeFile() throws ExecutionException, InterruptedException, IOException {
```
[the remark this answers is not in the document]
schmidt-sebastian: I actually fell for the double encoding trap the first time I tried to reproduce it. As part of this PR, I looked at the console to see that the file names were correct when we finished uploading the initial state to the backend. 

That being said, I created a fake network trace for a file with a unicode filename and am using that now in a unit test. That should verify that an upload with unicode uses the correct path. And if that works, the list() tests in this file should verify that List also works with unicode filenames.

firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java
schmidt-sebastian: Done

PR #766, opened 2019-09-05
Add debug logging.

vkryachko:


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/JobInfoScheduler.java
```diff
@@ -89,14 +92,20 @@ public void schedule(TransportContext transportContext, int attemptNumber) {
...
     int jobId = getJobId(transportContext);
     // Check if there exists a job scheduled for this backend name.
-    if (isJobServiceOn(jobScheduler, jobId, attemptNumber)) return;
+    if (isJobServiceOn(jobScheduler, jobId, attemptNumber)) {
+      Logging.d(
+          LOG_TAG, "Upload for context %s is already scheduled. Returning...", transportContext);
+      return;
+    }
+
+    long nextCallTime = eventStore.getNextCallTime(transportContext);
```
VinayGuthal: why is this needed?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -0,0 +1,49 @@
...
+  public static void d(String tag, String message, Object arg1, Object arg2) {
+    Log.d(getTag(tag), String.format(message, arg1, arg2));
+  }
+
+  public static void d(String tag, String message, Object... args) {
+    Log.d(getTag(tag), String.format(message, args));
+  }
+
+  public static void i(String tag, String message) {
+    Log.d(getTag(tag), message);
```
VinayGuthal: Log.i maybe?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/JobInfoScheduler.java
vkryachko: It's needed to be able to log it below

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -54,13 +55,11 @@
...
 import java.util.Map;
 import java.util.TimeZone;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import java.util.zip.GZIPOutputStream;
 
 final class CctTransportBackend implements TransportBackend {
 
-  private static final Logger LOGGER = Logger.getLogger(CctTransportBackend.class.getName());
+  private static final String LOG_TAG = "CctTransportBackend";
```
rlazo: Any reason for not preferring the class.getname()?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -243,7 +243,8 @@ private HttpResponse doSend(HttpRequest request) throws IOException {
       }
       channel.write(ByteBuffer.wrap(output.toByteArray()));
       int responseCode = connection.getResponseCode();
-      LOGGER.info("Status Code: " + responseCode);
+      Logging.i(LOG_TAG, "Status Code: " + responseCode);
```
rlazo: Use %s instead of +

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
vkryachko: yes, `class.getName()` will return the name of the class after proguarding, so it could end up being "a", so to avoid it I explicitly preserve the class name as a constant string

rlazo: Makes sense

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
vkryachko: Yeah, so I decided not to add an overload that takes format args and did not want to use String.format() directly here as this case is too simple. As we add more uses of info logging we can revisit this.

PR #767, opened 2019-09-05
Don't double save Query Data

schmidt-sebastian:
This removes the sophisticated logic to determine whether we need to persist query data in `releaseQuery()` as LRU GC bumps the sequence number and always persists the query when the reference is removed.


schmidt-sebastian:
Screen Shot 2019-09-05 at 10 23 28 AM

I DID IT!

PR #769, opened 2019-09-05
FIS getAuthToken implementation.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -169,6 +178,16 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
     };
   }
 
+  @NonNull
+  private static <F, T> Continuation<F, T> call(@NonNull Supplier<T> supplier) {
+    return t -> {
+      if (!t.isComplete()) {
+        Tasks.await(t);
```
ankitaj224: @ciarand  Instrumented tests fail sometimes because of this line. Its not strictly waiting for the prev Task (getId) to complete. Ends up calling refreshAuthTokenIfNecessary with empty PersistedFidEntry ( pointed you to that code). 

Can you please help me to understand what am I missing?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -244,6 +263,79 @@ private Void registerAndSaveFid(PersistedFidEntry persistedFidEntry)
     }
     return null;
   }
+
+  private String refreshAuthTokenIfNecessary(boolean forceRefresh)
+      throws FirebaseInstallationsException {
+
+    PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
+
+    if (persistedFidEntry == null) {
```
ankitaj224: @ciarand : Tests fail here because getId task is not yet complete.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Test failure is alternating.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -38,6 +38,7 @@
 import com.google.firebase.installations.remote.FirebaseInstallationServiceClient;
 import com.google.firebase.installations.remote.FirebaseInstallationServiceException;
 import com.google.firebase.installations.remote.InstallationResponse;
+import com.google.firebase.installations.remote.InstallationTokenResult;
```
ciarand: is this import necessary? I don't see it being used

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -54,6 +55,8 @@
   private final Clock clock;
   private final Utils utils;
 
+  private static final long TOKEN_EXPIRATION_BUFFER = 3600L;
```
ciarand: what does this variable represent? is it a time period?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -115,11 +118,17 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
+   * Returns a valid authentication token for the Firebase installation. Generates a new token if
+   * one doesn't exist, is expired or about to expire.
+   *
+   * <p>Should only be called if the Firebase Installation is registered.
+   */
   @NonNull
   @Override
-  public Task<InstallationTokenResult> getAuthToken(boolean forceRefresh) {
-    return Tasks.forResult(InstallationTokenResult.builder().build());
+  public Task<String> getAuthToken(final boolean forceRefresh) {
```
ciarand: doesn't make sense to mark a primitive parameter final

ciarand: in general, boolean parameters are a code smell. Could we either split the method (forceGetAuthToken?) or change the param to some sort of enum / int flag here? (See https://developer.android.com/reference/android/support/annotation/IntDef)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: I don't think this is correct. You're calling `Tasks.call(executor, this::getId)` but `getId` is already calling `Tasks.call` so you're doubly nesting your async operations.

Instead of this:

```java
    return Tasks.call(executor, this::getId)
        .continueWith(call(() -> refreshAuthTokenIfNecessary(forceRefresh)));
```

I think you mean to do something like this:

```java
    return getId().continueWith((idTask) -> refreshAuthTokenIfNecessary(idTask.getResult(), forceRefresh));
```

NOTE: The call to `getResult` will throw an Exception if the `getId` task failed, which will cause the return value of this method (the `getAuthToken` method) to be a failed Task, which is afaict what we want.

You're also not passing an explicit executor into the .continueWith call though, so the refreshAuthTokenIfNecessary method is getting run on the main thread (!). Given you're not allowed to call Tasks.await on the main thread, that seems like a bug.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ankitaj224: Yes, it is necessary. I moved the InstallationTokenResult file to different package. Hence, the change.

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -24,14 +28,35 @@
  */
 public interface FirebaseInstallationsApi {
 
+  /** Specifies the refresh options to get a FIS AuthToken. */
+  @IntDef({FORCE_REFRESH, DO_NOT_FORCE_REFRESH})
```
ankitaj224: @ciarand  I added the IntDef as per your suggestion. I tried incorporating as much as I understood. I think, there is scope for improvement. PTAL to see if I am correctly using it :) Thank you.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Very good catch. You are right. That was the issue :) Fixed it. About the 'Tasks.await' it wasn't a bug until I called ```Tasks.call(executor, this::getId)``` . After the change recommended by you, I could not await in the Main Thread. I updated the  code to await using the ```executor```. Let me know what you think.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Yes it is. 1 hour buffer time to evaluate if the auth token has expired.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: You are right. That was by mistake. I forgot to remove it. Thanks for catching this.

ankitaj224: I used IntDef. PTAL, if I am using it correctly.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -217,4 +225,22 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
           .isEqualTo(FirebaseInstallationsException.Status.CLIENT_ERROR);
     }
   }
+
+  @Test
+  public void testGetAuthToken_registeredFid_successful() throws Exception {
```
ankitaj224: I just added one test to review the IntDef usage. I will add more tests, once that is reviewed . Thank you.

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
ciarand: The @IntDef part is what I meant (although I think we should make it a bitmap), but the stateful setters aren't quite what I had in mind. Something like this was what I was thinking:

`getAuthToken(@AuthTokenOptions int authTokenOptions);`

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ciarand: got it

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -169,6 +191,15 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
     };
   }
 
+  @NonNull
+  private <F, T> Continuation<F, T> call(@NonNull Supplier<T> supplier) {
+    return t -> {
+      // Waiting for Task that registers FID on the FIS Servers
+      executor.awaitTermination(1000, TimeUnit.MILLISECONDS);
```
ciarand: this is just 1, TimeUnit.SECONDS

ciarand: I don't think `awaitTermination` is right here unless you're shutting down the executor service somewhere else. From the JavaDoc: "Blocks until all tasks have completed execution after a shutdown request, or the timeout occurs, or the current thread is interrupted, whichever happens first."

ciarand: you shouldn't need to do anything like that - then continueWith part ensures that this only gets called when the previous task is completed

ankitaj224: But we have a background task that registers the FID on FIS Servers, which runs independent of the prev task. That's why I wanted to wait for sometime.

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
ankitaj224: Updated. PTAL. Thank you.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated the code using CountDownLatch to await for the background task. 

@ciarand Wdyt?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
diwu-arete: Mention the time unit in the var name? for example secs

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -169,6 +180,15 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
     };
   }
 
+  @NonNull
+  private <F, T> Continuation<F, T> call(@NonNull Supplier<T> supplier) {
```
diwu-arete: give this method a more specific name?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -169,6 +180,15 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
...
+    return t -> {
+      // Waiting for Task that registers FID on the FIS Servers
+      latch.await(10, TimeUnit.SECONDS);
+      return supplier.get();
+    };
+  }
+
   private PersistedFidEntry createAndPersistNewFid() throws FirebaseInstallationsException {
```
diwu-arete: add javadoc or comment? same for the others

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -24,14 +28,32 @@
...
+  int FORCE_REFRESH = 1;
+
   /**
    * Async function that returns a globally unique identifier of this Firebase app installation.
    * This is a url-safe base64 string of a 128-bit integer.
    */
   Task<String> getId();
 
   /** Async function that returns a auth token(public key) of this Firebase app installation. */
-  Task<InstallationTokenResult> getAuthToken(boolean forceRefresh);
```
diwu-arete: Sorry I was not following what you may have changed.  Why do we change to return string instead of InstallationTokenResult?
We decided to return a struct in Firebase API review go/fis-sdk-review

ankitaj224: https://docs.google.com/document/d/1ib_AMf959UAPVyrzcnYMIFsxV2dMCzl5ZPG-8qHTH1M/edit#heading=h.ankqhj32c8xa

ankitaj224: https://github.com/firebase/firebase-js-sdk/blob/master/packages/installations/src/functions/get-token.ts

ios and js sdk returns String too

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -244,6 +269,86 @@ private Void registerAndSaveFid(PersistedFidEntry persistedFidEntry)
...
+          FirebaseInstallationsException.Status.SDK_INTERNAL_ERROR);
+    }
+  }
+
+  private boolean isAuthTokenExpired(PersistedFidEntry persistedFidEntry) {
+    return (persistedFidEntry.getTokenCreationEpochInSecs() + persistedFidEntry.getExpiresInSecs()
+        > currentTime() + AUTH_TOKEN_EXPIRATION_BUFFER);
+  }
+
+  private long currentTime() {
```
diwu-arete: currentTimeInSecs ?

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
diwu-arete: Well, 
1) the Android SDK design doc was written before the SDK review, I should change it but I didn't.
2) the iOS SDK and web SDK work started before the Android SDK review, we decided to do the struct in Android, and later change the iOS / web SDK to use the struct as well.

ankitaj224: Oh, got it. Makes sense. Rayo mentioned to be consistent with js-sdk as its already out.

I ll update it to a struct.

ankitaj224: Updated.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -28,26 +28,23 @@
...
 
-  /** Specifies the refresh options to get a FIS AuthToken. */
-  @IntDef({FORCE_REFRESH, DO_NOT_FORCE_REFRESH})
-  @Retention(SOURCE)
-  @interface RefreshAuthTokenOption {}
   /**
-   * AuthToken is forcefully refreshed on calling the {@link
-   * FirebaseInstallationsApi#getAuthToken()}.
+   * Specifies the options to get a FIS AuthToken. Default value of AuthTokenOption =
+   * DO_NOT_FORCE_REFRESH.
```
ciarand: the default part is describing the method, not this option, so the javadoc should be on the method not this option

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -48,27 +48,16 @@
...
   private final FirebaseInstallationServiceClient serviceClient;
   private final PersistedFid persistedFid;
   private final ExecutorService executor;
   private final Clock clock;
   private final Utils utils;
 
-  private static final long AUTH_TOKEN_EXPIRATION_BUFFER = 3600L; // 1 hour
+  private static final long AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS = 3600L; // 1 hour
+
+  private CountDownLatch latch = new CountDownLatch(1);
```
ciarand: I don't think this is a good solution... It's non-final, it means that this class can't be reused, and I don't think it's sufficient to express the state you're trying to express

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -128,7 +117,7 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
   public Task<String> getId() {
     return Tasks.call(executor, this::getPersistedFid)
         .continueWith(orElse(this::createAndPersistNewFid))
-        .onSuccessTask(this::registerFidIfNecessary);
+        .onSuccessTask(persistedFidEntry -> registerFidIfNecessary(persistedFidEntry));
```
ciarand: this could be a method ref

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -192,14 +188,15 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
   }
 
   @NonNull
-  private <F, T> Continuation<F, T> call(@NonNull Supplier<T> supplier) {
+  private <F, T> Continuation<F, T> awaitFidRegistration(@NonNull Supplier<T> supplier) {
```
ciarand: instead of doing this weird out of band latch thing, we could just copy the Tasks.await implementation that uses a real listener

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -56,7 +55,7 @@
   private final Clock clock;
   private final Utils utils;
 
-  private static final long AUTH_TOKEN_EXPIRATION_BUFFER = 3600L; // 1 hour
+  private static final long AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS = 3600L; // 1 hour
 
   private CountDownLatch latch = new CountDownLatch(1);
```
diwu-arete: Do not make the latch global variable.

If this is only useful for getAuthToken call, then make it a local variable in there.

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
ankitaj224: Is this right now?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Ack.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: I changed this when I was trying listener logic.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -191,13 +229,20 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+   */
+  private Task<String> registerFidIfNecessary(
+      PersistedFidEntry persistedFidEntry, AwaitListener listener) {
     String fid = persistedFidEntry.getFirebaseInstallationId();
 
     // Check if the fid is unregistered
     if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
       updatePersistedFidWithPendingStatus(fid);
-      Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry));
+      Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry))
```
ankitaj224: There are additional check I can add . Ex: If Registered , listener onSuccess. I will do that once, I get this code reviewed. 

Thanks.

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
ciarand: Sorry, I wasn't clear. You had it correct (Javadoc above the `@interface` declaration), I just meant the "Default value of AuthTokenOption = DO_NOT_FORCE_REFRESH" should be above the method that uses it

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -117,7 +116,14 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
   public Task<String> getId() {
     return Tasks.call(executor, this::getPersistedFid)
         .continueWith(orElse(this::createAndPersistNewFid))
-        .onSuccessTask(persistedFidEntry -> registerFidIfNecessary(persistedFidEntry));
+        .onSuccessTask(persistedFidEntry -> registerFidIfNecessary(persistedFidEntry, null));
```
ciarand: isn't this all just getId(null)?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -373,6 +379,29 @@ private boolean isAuthTokenExpired(PersistedFidEntry persistedFidEntry) {
   private long currentTimeInSecs() {
     return TimeUnit.MILLISECONDS.toSeconds(clock.currentTimeMillis());
   }
+
+  interface CombinedListener extends OnCompleteListener<Void> {}
```
ciarand: don't bother including this, just have AwaitListener extend OnCompleteListener directly

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -373,6 +379,29 @@ private boolean isAuthTokenExpired(PersistedFidEntry persistedFidEntry) {
...
+
+  private static final class AwaitListener implements CombinedListener {
+    private final CountDownLatch mLatch = new CountDownLatch(1);
+
+    public void onSuccess() {
+      mLatch.countDown();
+    }
+
```
ciarand: imo don't bother including this (the argumentless await() version), we shouldn't ever block without a timeout

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -188,10 +197,11 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
   }
 
   @NonNull
-  private <F, T> Continuation<F, T> awaitFidRegistration(@NonNull Supplier<T> supplier) {
+  private <F, T> Continuation<F, T> awaitFidRegistration(
```
ciarand: could be static

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -224,19 +234,15 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
     if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
       updatePersistedFidWithPendingStatus(fid);
       Tasks.call(executor, () -> registerAndSaveFid(persistedFidEntry))
-          .addOnCompleteListener(
-              task -> {
-                if (task.isSuccessful()) {
-                  latch.countDown();
-                }
-              });
+          .addOnCompleteListener(listener);
```
ciarand: should this be checking for nullness?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
ankitaj224: Thanks for the clarification. I corrected it.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated the code as per @ciarand comments. PTAL. Thanks.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -191,17 +230,45 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+   * Registers the FID with FIS servers if FID is in UNREGISTERED state.
+   *
+   * <p>Updates FID registration status to PENDING to avoid multiple network calls to FIS Servers.
+   */
+  private Task<String> registerFidIfNecessary(
+      PersistedFidEntry persistedFidEntry, AwaitListener listener) {
     String fid = persistedFidEntry.getFirebaseInstallationId();
 
     // Check if the fid is unregistered
     if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
```
ankitaj224: Yet to implement race conditions between multiple getAuthToken calls at the same time.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -115,23 +114,34 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
   @NonNull
   @Override
   public Task<String> getId() {
+    return getId(null);
+  }
+
+  private Task<String> getId(AwaitListener awaitListener) {
```
diwu-arete: add a comment or javadoc for this method, especially mention when the listener will be notified.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -373,6 +401,23 @@ private boolean isAuthTokenExpired(PersistedFidEntry persistedFidEntry) {
   private long currentTimeInSecs() {
     return TimeUnit.MILLISECONDS.toSeconds(clock.currentTimeMillis());
   }
+
+  private static final class AwaitListener implements OnCompleteListener<Void> {
```
diwu-arete: you may consider to move this(or some other code too) into a separate class. (The current class is already too big)

diwu-arete:
Pls check the api-information — Job failed.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

VinayGuthal:
run ./gradlew :firebase-installations:generateApiTxtFile and commit the changes to fix the api-information bug

VinayGuthal:
same thing with :firebase-installations-interop:apiInformation project as well

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/AwaitListener.java
```diff
@@ -0,0 +1,38 @@
...
+import androidx.annotation.NonNull;
+import com.google.android.gms.tasks.OnCompleteListener;
+import com.google.android.gms.tasks.Task;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+final class AwaitListener implements OnCompleteListener<Void> {
+  private final CountDownLatch mLatch = new CountDownLatch(1);
+
+  public void onSuccess() {
```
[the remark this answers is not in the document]
ankitaj224: Its used. If the status is REGISTERED, instead of waiting for 10s, I am using this method to update the awaitlistener.

FirebaseInstalllations.java line 261

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -169,6 +200,17 @@ private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persi
     };
   }
 
+  @NonNull
+  private static <F, T> Continuation<F, T> awaitFidRegistration(
+      @NonNull Supplier<T> supplier, AwaitListener listener) {
+    return t -> {
+      // Waiting for Task that registers FID on the FIS Servers
+      listener.await(10, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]

firebase-installations/src/main/java/com/google/firebase/installations/AwaitListener.java
```diff
@@ -0,0 +1,38 @@
...
+package com.google.firebase.installations;
+
+import androidx.annotation.NonNull;
+import com.google.android.gms.tasks.OnCompleteListener;
+import com.google.android.gms.tasks.Task;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+final class AwaitListener implements OnCompleteListener<Void> {
+  private final CountDownLatch mLatch = new CountDownLatch(1);
```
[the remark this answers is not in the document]
ankitaj224: Done. Thanks.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -217,4 +275,217 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+        .that(task1.getResult().getToken())
+        .isEqualTo(TEST_AUTH_TOKEN_3);
+    assertWithMessage("Persisted Auth Token doesn't match")
+        .that(task2.getResult().getToken())
+        .isEqualTo(TEST_AUTH_TOKEN_4);
+  }
+
+  class ServiceClient extends FirebaseInstallationServiceClient {
+
+    private boolean secondCall;
```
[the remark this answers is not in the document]
ankitaj224: Thanks for the suggestion. I have tried using thenAnswer. PTAL.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -217,4 +275,217 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+        new FirebaseInstallations(
+            mockClock, executor, firebaseApp, serviceClient, mockPersistedFid, mockUtils);
+
+    // Call getAuthToken multiple times with FORCE_REFRESH option. Also, sleep for 500ms in between
+    // the calls to ensure tasks are called in order.
+    Task<InstallationTokenResult> task1 =
+        firebaseInstallations.getAuthToken(FirebaseInstallationsApi.FORCE_REFRESH);
+    Thread.sleep(500);
+    Task<InstallationTokenResult> task2 =
+        firebaseInstallations.getAuthToken(FirebaseInstallationsApi.FORCE_REFRESH);
```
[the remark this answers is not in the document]
ankitaj224: I wanted the sleep to ensure task executed in order. But I removed it, instead of validating on both expected values

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -217,4 +275,217 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+            firebaseInstallations.getAuthToken(FirebaseInstallationsApi.DO_NOT_FORCE_REFRESH));
+
+    assertWithMessage("Persisted Auth Token doesn't match")
+        .that(installationTokenResult.getToken())
+        .isEqualTo(TEST_AUTH_TOKEN_2);
+  }
+
+  @Test
+  public void testGetAuthToken_unregisteredFid_fetchedNewTokenFromFIS() throws Exception {
+    when(mockPersistedFid.readPersistedFidEntryValue())
```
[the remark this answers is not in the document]
ankitaj224: Done. PTAL.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -217,4 +275,217 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+    // Expect exception
+    try {
+      Tasks.await(
+          firebaseInstallations.getAuthToken(FirebaseInstallationsApi.DO_NOT_FORCE_REFRESH));
+      fail();
+    } catch (ExecutionException expected) {
+      Throwable cause = expected.getCause();
+      assertWithMessage("Exception class doesn't match")
+          .that(cause)
+          .isInstanceOf(FirebaseInstallationsException.class);
```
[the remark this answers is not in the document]
ankitaj224: Done. Did you also want the status to be changed? I tried but dint find a easy way.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -68,12 +78,53 @@
...
+          .setExpiresInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP)
+          .setRegistrationStatus(PersistedFid.RegistrationStatus.REGISTERED)
+          .build();
 
   @Before
   public void setUp() throws FirebaseInstallationServiceException {
     MockitoAnnotations.initMocks(this);
     FirebaseApp.clearInstancesForTest();
-    executor = new ThreadPoolExecutor(0, 2, 10L, TimeUnit.SECONDS, new SynchronousQueue<>());
+    executor = new ThreadPoolExecutor(0, 4, 10L, TimeUnit.SECONDS, new SynchronousQueue<>());
```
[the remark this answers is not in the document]
ankitaj224: Done. PTAL.

firebase-installations-interop/api.txt
```diff
@@ -5,15 +5,15 @@ package com.google.firebase.installations {
...
+    method @NonNull public abstract long getTokenExpirationInSecs();
     method @NonNull public abstract com.google.firebase.installations.InstallationTokenResult.Builder toBuilder();
   }
 
   public abstract static class InstallationTokenResult.Builder {
     ctor public InstallationTokenResult.Builder();
     method @NonNull public abstract com.google.firebase.installations.InstallationTokenResult build();
     method @NonNull public abstract com.google.firebase.installations.InstallationTokenResult.Builder setToken(@NonNull String);
-    method @NonNull public abstract com.google.firebase.installations.InstallationTokenResult.Builder setTokenExpirationTimestampMillis(@NonNull long);
+    method @NonNull public abstract com.google.firebase.installations.InstallationTokenResult.Builder setTokenExpirationInSecs(@NonNull long);
```
ciarand: just noticed this, but longs (lowercase primitive) can't be null

firebase-installations-interop/src/main/java/com/google/firebase/installations/InstallationTokenResult.java
```diff
@@ -46,7 +45,7 @@
     public abstract Builder setToken(@NonNull String value);
 
     @NonNull
-    public abstract Builder setTokenExpirationTimestampMillis(@NonNull long value);
+    public abstract Builder setTokenExpirationInSecs(@NonNull long value);
```
ciarand: I guess this is the source of the generated API txt doc

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -264,11 +266,12 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
     // Expect exception
     try {
       Tasks.await(firebaseInstallations.getId());
-      fail();
+      fail("Could not update local storage.");
     } catch (ExecutionException expected) {
       Throwable cause = expected.getCause();
```
ciarand: now that this local cause variable is only being used in one place, I'd just inline it

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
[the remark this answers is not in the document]
ciarand: You'd have to create a custom truth subject for that and tbh I don't think it's worth it atm

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -399,7 +407,11 @@ public void testGetAuthToken_serverError_failure() throws Exception {
   }
 
   @Test
-  public void testGetAuthToken_multipleCalls_fetchedNewTokenOnce() throws Exception {
+  public void testGetAuthToken_multipleCallsDoNotForceRefresh_fetchedNewTokenOnce()
+      throws Exception {
+    // Using mockPersistedFid to ensure the order of returning persistedFidEntry  to 2 tasks
```
ciarand: nit: double space

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -399,7 +407,11 @@ public void testGetAuthToken_serverError_failure() throws Exception {
...
+    // triggered simultaneously. Task2 waits for Task1 to complete.On Task1 completion, task2 reads
```
ciarand: nit: `complete.On` should be `complete. On`

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -425,67 +437,54 @@ public void testGetAuthToken_multipleCalls_fetchedNewTokenOnce() throws Exceptio
...
 
   @Test
   public void testGetAuthToken_multipleCallsForceRefresh_fetchedNewTokenTwice() throws Exception {
     when(mockPersistedFid.readPersistedFidEntryValue()).thenReturn(REGISTERED_FID_ENTRY);
-    // Use a custom ServiceClient to mock the network calls ensuring task1 is not completed
-    // before task2. Hence, we can test multiple calls to getAUthToken() and verify task2 waits for
-    // task1 to complete.
-    ServiceClient serviceClient = new ServiceClient();
+    // Use a mock ServiceClient for network calls with delay(1000ms) to ensure first task is not
+    // completed before the second task starts. Hence, we can test multiple calls to getAUthToken()
```
ciarand: nit: getAUthToken should be getAuthToken

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -425,67 +437,54 @@ public void testGetAuthToken_multipleCalls_fetchedNewTokenOnce() throws Exceptio
...
+    // and verify one task waits for another task to complete.
+
+    doAnswer(
+            new AnswersWithDelay(
```
ciarand: I'm not 100% sure here, but I think we're supposed to use the AdditionalAnswers class to create these instead of reaching into the mockito internal package: https://static.javadoc.io/org.mockito/mockito-core/2.9.0/org/mockito/AdditionalAnswers.html

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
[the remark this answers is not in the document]
ciarand: Why don't they execute in order at the moment? Is it because the executor has 4 threads? Could we use a single threaded executor to make this deterministic?

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ankitaj224: Thanks for the link. I dint realize it was internal package. Updated.

firebase-installations-interop/src/main/java/com/google/firebase/installations/InstallationTokenResult.java
ankitaj224: Done. Thanks.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -444,19 +440,19 @@ public void testGetAuthToken_multipleCallsDoNotForceRefresh_fetchedNewTokenOnce(
...
     // Use a mock ServiceClient for network calls with delay(1000ms) to ensure first task is not
-    // completed before the second task starts. Hence, we can test multiple calls to getAUthToken()
+    // completed before the second task starts. Hence, we can test multiple calls to getAuthToken()
     // and verify one task waits for another task to complete.
 
     doAnswer(
-            new AnswersWithDelay(
+            AdditionalAnswers.answersWithDelay(
                 1000,
                 new Returns(
```
ciarand: `Returns` is still internal to Mockito. Should be able to use a lambda:

```
doAnswer(AdditionalAnswers.answerWithDelay(1_000, (unused) -> myThing.build()))
```

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -480,10 +476,10 @@ public void testGetAuthToken_multipleCallsForceRefresh_fetchedNewTokenTwice() th
     // As we cannot ensure which task got executed first, verifying with both expected values
     assertWithMessage("Persisted Auth Token doesn't match")
         .that(task1.getResult().getToken())
-        .isAnyOf(TEST_AUTH_TOKEN_3, TEST_AUTH_TOKEN_4);
+        .isEqualTo(TEST_AUTH_TOKEN_3);
```
ciarand: nice

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -61,7 +60,7 @@
   FirebaseInstallations(FirebaseApp firebaseApp) {
     this(
         DefaultClock.getInstance(),
-        new ThreadPoolExecutor(0, 1, 30L, TimeUnit.SECONDS, new SynchronousQueue<>()),
```
ciarand: weren't we going to stick with this constructor but use a linked list for the queue instead? that way the thread's allowed to die - otherwise it just sits there idle wasting memory

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Got it. Updated. Thanks.

PR #773, opened 2019-09-05
Implement firebase-storage-ktx Kotlin library.

rlazo:


firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
```diff
@@ -398,6 +403,11 @@ public Builder setContentEncoding(@Nullable String contentEncoding) {
       return this;
     }
 
+    @Nullable
```
schmidt-sebastian: Please add JavaDoc (here and everywhere). This also needs to go through API review, but an email might be fine here.

schmidt-sebastian: Thanks for sending the review doc. Please do add JavaDoc though.

rlazo: done.

schmidt-sebastian: @rachelsaunders Can you take a quick glance at this? Do we want to use `@return`?

firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
```diff
@@ -398,17 +404,31 @@ public Builder setContentEncoding(@Nullable String contentEncoding) {
...
+    public String getContentEncoding() {
+      return mMetadata.mContentEncoding.getValue();
+    }
+
     /**
-     * @param contentDisposition changes the content disposition for the {@link StorageReference}
+     * Changes the content disposition for the {@link StorageReference}
+     *
+     * @param contentDisposition changes the new content disposition to use.
```
schmidt-sebastian: s/changes /

rlazo: fixed.

firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
rachelsaunders: I'm actually not familiar with the syntax required here.  It's something on my learning TODOs for post-FireConf, though!

Is there a similar situation in another file that might be an exemplar for what you're trying to achieve? If not, we can reach out to some other folks to see if they know the answer.

schmidt-sebastian: I think we can go down three routes here:

```
/** Returns the content encoding for the {@link StorageReference} */
```

```
/** 
   * @return The content encoding for the {@link StorageReference}.
   */
```

```
/** 
   * Returns the content encoding for the {@link StorageReference} 
   *
   * @return The content encoding for the {@link StorageReference}.
   */
```

If Rachel doesn't have a preference, we can merge as is and correct if needed when the docs are generated.

rlazo: @rachelmyers I think sebastian refers to using `@return` tag as he's doing in the rest of the file instead of writing `/** Returns ` This file uses both, so I can go with any style.

rachelsaunders: Yes, I would agree with consistency, unless one syntax results in a different output style than the other -- _and_ those different styles are desired.

What does the rest of the Firestore library use?

FWIW - This is the ktx library, so we currently don't generate Javadocs for it... yet.

schmidt-sebastian: We are very inconsistent :)

rlazo: I've updated the javadocs to use @return, since it's the most popular option in this particular file. PTAL Thanks!

firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
```diff
@@ -443,5 +469,11 @@ public Builder setContentType(@Nullable String contentType) {
       mMetadata.mContentType = MetadataValue.withUserValue(contentType);
       return this;
     }
+
+    /** @return the content Type of this associated {@link StorageReference} */
```
schmidt-sebastian: s/content/Content/

firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
```diff
@@ -398,17 +404,31 @@ public Builder setContentEncoding(@Nullable String contentEncoding) {
...
 
+    /** @return the content disposition for the {@link StorageReference} */
+    @Nullable
+    public String getContentDisposition() {
+      return mMetadata.mContentDisposition.getValue();
+    }
+
     /**
-     * sets the Cache Control for the {@link StorageReference}
+     * Sets the Cache Control for the {@link StorageReference}
```
schmidt-sebastian: s/Control/Control header/

firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
```diff
@@ -418,6 +438,12 @@ public Builder setCacheControl(@Nullable String cacheControl) {
       return this;
     }
 
+    /** @return the Cache Control for the {@link StorageReference} */
```
schmidt-sebastian: s/Control/Control header/

firebase-storage/src/main/java/com/google/firebase/storage/StorageMetadata.java
rlazo: done

PR #775, opened 2019-09-06
Initial change to move components to its own library.

vkryachko:


firebase-components/firebase-components.gradle
```diff
@@ -0,0 +1,54 @@
...
+plugins {
+    id 'firebase-library'
+}
+
+firebaseLibrary {
+    publishSources = true
+}
+
+android {
+
```
rlazo: rm

firebase-common/firebase-common.gradle
```diff
@@ -61,6 +61,7 @@ android {
 }
 
 dependencies {
+    implementation project(':firebase-components')
```
sheepmaster: I think with this:
```suggestion
    api project(':firebase-components')
```
you don't have to add the dependency to all the other SDKs that already depend on this one.

vkryachko: Nice, I did not know you could suggest changes in github!

As for changing to `api`, I explicitly decided not to do it and force SDKs to include a dependency on components so that it propagates into their pom.xml as opposed to transitively depending on it.

firebase-components/firebase-components.gradle
vkryachko: done

firebase-common/firebase-common.gradle
sheepmaster: OK, fair enough.

PR #778, opened 2019-09-09
Adding onSnapshotsInSync (not public)

thebrianchen:
Porting from [web](https://github.com/firebase/firebase-js-sdk/pull/2100).

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -30,12 +30,14 @@ import com.google.firebase.firestore.FirebaseFirestoreException.Code;
 import com.google.firebase.firestore.auth.CredentialsProvider;
 import com.google.firebase.firestore.auth.EmptyCredentialsProvider;
 import com.google.firebase.firestore.auth.FirebaseAuthCredentialsProvider;
+import com.google.firebase.firestore.core.AsyncEventListener;
 import com.google.firebase.firestore.core.DatabaseInfo;
 import com.google.firebase.firestore.core.FirestoreClient;
 import com.google.firebase.firestore.local.SQLitePersistence;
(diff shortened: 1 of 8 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -460,6 +464,93 @@ public static void setLoggingEnabled(boolean loggingEnabled) {
...
+   * @param runnable A callback to be called every time all snapshot listeners are in sync with each
+   *     other.
+   * @return A registration object that can be used to remove the listener.
+   */
+  private ListenerRegistration addSnapshotsInSyncListener(
+      Executor userExecutor, @Nullable Activity activity, @NonNull Runnable runnable) {
+    ensureClientConfigured();
+    EventListener<Void> eventListener =
+        (Void v, FirebaseFirestoreException error) -> {
+          runnable.run();
```
mikelehen: Since we're ignoring `error` maybe make it more explicit that this is okay / intentional with an assert?
```
hardAssert(error == null, "snapshots-in-sync listeners should never get errors.");

File: firebase-firestore/src/main/java/com/google/firebase/firestore/core/EventManager.java [hunk 2]
```diff
@@ -54,12 +59,15 @@
...
 
+  private final Set<EventListener<Void>> snapshotsInSyncListeners;
+
   private OnlineState onlineState = OnlineState.UNKNOWN;
 
   public EventManager(SyncEngine syncEngine) {
     this.syncEngine = syncEngine;
     queries = new HashMap<>();
     syncEngine.setCallback(this);
+    snapshotsInSyncListeners = new HashSet<>();
```
mikelehen: I think you can move this to the `snapshotsInSyncListeners` definition? i.e.

```
private final Set<EventListener<Void>> snapshotsInSyncListeners = new HashSet<>();
```

PR #781, opened 2019-09-10
Factor out component runtime initialization.

vkryachko:
The change move component initialization into firebase-components, which
now has a new package `compat` that contains `ContainerOwner` that takes
over responsibilities of `ComponentRuntime` instance management from
`FirebaseApp`.

`FirebaseApp` also became a component itself with its own
`ComponentRegistrar` and is itself discovered like any other component.


firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -197,25 +170,22 @@ public static FirebaseApp getInstance() {
...
-              "FirebaseApp with name %s doesn't exist. %s", name, availableAppNamesMessage);
-      throw new IllegalStateException(errorMessage);
+    if (instance != null && instance.get(FirebaseApp.class) != null) {
+      return instance.get(FirebaseApp.class);
     }
+
+    List<String> availableAppNames = getAllAppNames();
+    String availableAppNamesMessage;
+    if (availableAppNames.isEmpty()) {
+      availableAppNamesMessage = "";
```
rlazo: Maybe use "No Apps registered" or similar as error message.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -236,21 +206,21 @@ public static FirebaseApp getInstance(@NonNull String name) {
...
-      }
-      return initializeApp(context, firebaseOptions);
+
+    ComponentRuntime instance = ContainerOwner.getInstance(DEFAULT_APP_NAME);
+    if (instance != null) {
+      return getInstance();
     }
+    FirebaseOptions firebaseOptions = FirebaseOptions.fromResource(context);
+    if (firebaseOptions == null) {
+      Log.w(
```
rlazo: This looks a bit more like an error than a warning.

firebase-components/src/main/java/com/google/firebase/components/compat/ContainerOwner.java
```diff
@@ -0,0 +1,183 @@
...
+  public static List<ComponentRuntime> getAll() {
+    synchronized (CONTAINERS) {
+      return new ArrayList<>(CONTAINERS.values());
+    }
+  }
+
+  public static ComponentRuntime initialize(
+      Context applicationContext, String name, Component<?>... additionalComponents) {
+    if (applicationContext.getApplicationContext() != null) {
+      // In shared processes' content providers getApplicationContext() can return null.
```
rlazo: In the case of shared processes content providers, it's OK to use the context directly passed even if its not an applicationContext, right? then I think it will be clearer to call the parameter context.

firebase-components/src/main/java/com/google/firebase/components/compat/ContainerOwner.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private static boolean isDefault(ComponentRuntime componentRuntime) {
+    return DEFAULT_CONTAINER_NAME.equals(componentRuntime.get(ContainerInfo.class).getName());
+  }
+
+  private static boolean isUserUnlocked(Context context) {
+    if (Build.VERSION.SDK_INT >= 24) {
+      return context.getSystemService(UserManager.class).isUserUnlocked();
+    } else {
```
rlazo: No need for else, just the return.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -197,25 +170,22 @@ public static FirebaseApp getInstance() {
...
-      } else {
-        availableAppNamesMessage =
-            "Available app names: " + TextUtils.join(", ", availableAppNames);
-      }
-      String errorMessage =
-          String.format(
-              "FirebaseApp with name %s doesn't exist. %s", name, availableAppNamesMessage);
-      throw new IllegalStateException(errorMessage);
+    if (instance != null && instance.get(FirebaseApp.class) != null) {
+      return instance.get(FirebaseApp.class);
```
sheepmaster: Do we expect this to happen? getApps() does not check for existence of a FirebaseApp when adding it to the list.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -197,25 +170,22 @@ public static FirebaseApp getInstance() {
...
-        availableAppNamesMessage = "";
-      } else {
-        availableAppNamesMessage =
-            "Available app names: " + TextUtils.join(", ", availableAppNames);
-      }
-      String errorMessage =
-          String.format(
-              "FirebaseApp with name %s doesn't exist. %s", name, availableAppNamesMessage);
-      throw new IllegalStateException(errorMessage);
+    if (instance != null && instance.get(FirebaseApp.class) != null) {
```
sheepmaster: Do we expect FirebaseApp to be null? getApps() does not check for existence of a FirebaseApp when adding it to the list.

firebase-components/src/main/java/com/google/firebase/components/compat/ContainerOwner.java
```diff
@@ -0,0 +1,183 @@
...
+
+/** */
+public final class ContainerOwner {
+
+  private static final @NonNull String DEFAULT_CONTAINER_NAME = "[DEFAULT]";
+
+  /** A map of (name, ComponentRuntime) instances. */
+  private static final Map<String, ComponentRuntime> CONTAINERS = new ArrayMap<>();
+
+  @GuardedBy("CONTAINERS")
```
sheepmaster: This probably should annotate CONTAINERS.

firebase-components/src/main/java/com/google/firebase/components/compat/ContainerOwner.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicReference;
+
+/** */
+public final class ContainerOwner {
+
+  private static final @NonNull String DEFAULT_CONTAINER_NAME = "[DEFAULT]";
+
+  /** A map of (name, ComponentRuntime) instances. */
+  private static final Map<String, ComponentRuntime> CONTAINERS = new ArrayMap<>();
```
sheepmaster: So, I know that this is what FirebaseApp used to do, but looking through the Java style guide, it actually says that only deeply immutable data structures should be in UPPER_CASE (https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names).

firebase-components/src/main/java/com/google/firebase/components/compat/ContainerOwner.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicReference;
+
+/** */
+public final class ContainerOwner {
+
+  private static final @NonNull String DEFAULT_CONTAINER_NAME = "[DEFAULT]";
```
sheepmaster: This value is duplicated in FirebaseApp. Make it public here and refer to it in FirebaseApp?

firebase-components/src/main/java/com/google/firebase/components/compat/ContainerOwner.java
```diff
@@ -0,0 +1,183 @@
...
+import com.google.firebase.components.ContainerInfo;
+import com.google.firebase.components.Preconditions;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicReference;
+
+/** */
```
sheepmaster: Flesh out the comment?

PR #786, opened 2019-09-10
add firestore support for heartbeat and user agent string

VinayGuthal:
This pr adds in an interface grpcMetadata which the FirestoreChannel uses to populate its headers. This also adds in the UserAgentInformation and the HeartBeatInformation for usage in the firestore backend.


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -283,6 +288,7 @@ private Metadata requestHeaders() {
     headers.put(X_GOOG_API_CLIENT_HEADER, X_GOOG_API_CLIENT_VALUE);
     // This header is used to improve routing and project isolation by the backend.
     headers.put(RESOURCE_PREFIX_HEADER, this.resourcePrefixValue);
+    grpcMetadata.updateMetadata(headers);
```
VinayGuthal: add nullability check

firebase-firestore/src/main/java/com/google/firebase/firestore/grpc/DefaultGrpcMetadata.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.firebase.heartbeatinfo.HeartBeatInfo;
+import com.google.firebase.platforminfo.UserAgentPublisher;
+import io.grpc.Metadata;
+
+/**
+ * Class provides an implementation of the GrpcMetadata interface.
+ *
+ * <p>This updates the metadata with platformInfo string and the heartBeatInfo code.
+ */
+public class DefaultGrpcMetadata implements GrpcMetadata {
```
wilhuff: `DefaultGrpcMetadata` is not very descriptive. Since this is related to `x-firebase-client*` headers could we use `FirebaseClient` instead of `Default`?

Other alternatives:
  * `HeartBeat`-  -- though this isn't quite right because this handles both the user agent and heartbeat headers.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
```diff
@@ -41,13 +44,17 @@
         Component.builder(FirestoreMultiDbComponent.class)
             .add(Dependency.required(FirebaseApp.class))
             .add(Dependency.required(Context.class))
+            .add(Dependency.required(HeartBeatInfo.class))
```
wilhuff: These should be optional dependencies, shouldn't they?

firebase-firestore/src/main/java/com/google/firebase/firestore/grpc/DefaultGrpcMetadata.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public DefaultGrpcMetadata(
+      @NonNull UserAgentPublisher userAgentPublisher, @NonNull HeartBeatInfo heartBeatInfo) {
+    this.userAgentPublisher = userAgentPublisher;
+    this.heartBeatInfo = heartBeatInfo;
+  }
+
+  @Override
+  public void updateMetadata(@NonNull Metadata metadata) {
+    metadata.put(USER_AGENT_HEADER, userAgentPublisher.getUserAgent());
```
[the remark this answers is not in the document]
VinayGuthal: The value of these headers would dictate the type of heartbeat.
0 - no heartbeat
1 - sdk heartbeat
2- global heartbeat
3 - combined heartbeat.
I can put the condition to put the header only if the value is not 0. Is there any problem in sending the value anyway? 
The backend code will handle the scenarios when the value is 1, 2 or 3 and assume 0 means no heart beat.
. The USER_AGENT_HEADER needs to be added unconditionally

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
VinayGuthal: These should be required? Bcz this assumes we have the firebase dependencies during component initialization. The constructor should have nulllable attributes so that they can be used in the tests. Am I understanding this currently?

cc @vkryachko

firebase-firestore/src/main/java/com/google/firebase/firestore/grpc/DefaultGrpcMetadata.java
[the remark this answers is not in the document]
wilhuff: Let's discuss offline.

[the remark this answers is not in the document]
VinayGuthal: Have updated the heartbeatinfo header to be sent only once a day. Once I discuss with you we can decide as to what to do about userplatforminfo

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
```diff
@@ -45,8 +45,12 @@ public DefaultGrpcMetadata(
   @Override
   public void updateMetadata(@NonNull Metadata metadata) {
     metadata.put(USER_AGENT_HEADER, userAgentPublisher.getUserAgent());
-    metadata.put(
-        HEART_BEAT_HEADER,
-        Integer.toString(heartBeatInfo.getHeartBeatCode(firebaseFirestoreHeartBeatTag).getCode()));
+    int heartBeatCode = heartBeatInfo.getHeartBeatCode(firebaseFirestoreHeartBeatTag).getCode();
+    if (heartBeatCode != 0) {
```
wilhuff: Since we can't link to the design doc for this here, could you add a comment about why `0` is special here?

```
// Non-zero values indicate some kind of heartbeat should be sent
```

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
```diff
@@ -45,8 +45,12 @@ public DefaultGrpcMetadata(
...
+      metadata.put(
+          HEART_BEAT_HEADER,
+          Integer.toString(
+              heartBeatInfo.getHeartBeatCode(firebaseFirestoreHeartBeatTag).getCode()));
```
wilhuff: Re-use `heartBeatCode` here? Otherwise it's very likely that you'll end up sending zero every time here since the first call consumed the beat for the day.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
```diff
@@ -45,8 +45,12 @@ public DefaultGrpcMetadata(
   @Override
   public void updateMetadata(@NonNull Metadata metadata) {
     metadata.put(USER_AGENT_HEADER, userAgentPublisher.getUserAgent());
-    metadata.put(
-        HEART_BEAT_HEADER,
-        Integer.toString(heartBeatInfo.getHeartBeatCode(firebaseFirestoreHeartBeatTag).getCode()));
+    int heartBeatCode = heartBeatInfo.getHeartBeatCode(firebaseFirestoreHeartBeatTag).getCode();
```
wilhuff: How costly is calling `getHeartBeatCode`? Does it perform any I/O? Does it talk to the network?

VinayGuthal: it looks into the file stored in the app to check when the last time the heartbeat was sent. It doesnt talk to network.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
VinayGuthal: reused nice catch!

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
VinayGuthal: yes sure! Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
vkryachko: Logically it seems optional, as it is not critical to the operation of Firestore, however I don't think it matters much as the components in question are part of `firebase-common` so there should not be a case where they would be missing at runtime, this could in theory change in the future though.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
```diff
@@ -41,13 +44,17 @@
...
+            .add(Dependency.required(UserAgentPublisher.class))
             .add(Dependency.optional(InternalAuthProvider.class))
             .factory(
                 c ->
                     new FirestoreMultiDbComponent(
                         c.get(Context.class),
                         c.get(FirebaseApp.class),
-                        c.get(InternalAuthProvider.class)))
+                        c.get(InternalAuthProvider.class),
+                        new FirebaseClientGrpcMetadataProvider(
```
vkryachko: If you choose to make the components optional pls update this code to check for nulls and inject null instead of `FirebaseClientGrpcMetadataProvider`

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
vkryachko: it uses `SharedPreferences` so technically it only reads from disk once and keeps/manipulates state in memory, syncing to disk asynchronously.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
vkryachko: Pls switch to provider dependency, i.e. `requiredProvider()`/`optionalProvider()`, the components are not needed until later, so it's nice to not do more work than necessary(initializing the components) during Firestore initialization.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
```diff
@@ -41,13 +44,17 @@
...
+                            c.get(UserAgentPublisher.class), c.get(HeartBeatInfo.class))))
```
vkryachko: This needs to change to `c.getProvider()`, and also `FirebaseClientGrpcMetadataProvider` needs to change to use `Provider<T>` instead of values directly

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProviderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    metadataProvider.updateMetadata(metadata);
+    assertThat(metadata.keys().size()).isEqualTo(0);
+  }
+
+  @Test
+  public void updateHeaderWhenHBCodeisNoneZero() {
+    Metadata metadata = new Metadata();
+    when(mockUserAgentProvider.get()).thenReturn(mockUserAgent);
+    when(mockHeartBeatProvider.get()).thenReturn(mockHeartBeat);
+    when(mockHeartBeat.getHeartBeatCode(any())).thenReturn(HeartBeatInfo.HeartBeat.GLOBAL);
```
vkryachko: pls add test for `HeartBeat.SDK`

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProviderTest.java
```diff
@@ -0,0 +1,80 @@
...
+      Metadata.Key.of("x-firebase-client-log-type", Metadata.ASCII_STRING_MARSHALLER);
+
+  private static final Metadata.Key<String> USER_AGENT_HEADER =
+      Metadata.Key.of("x-firebase-client", Metadata.ASCII_STRING_MARSHALLER);
+
+  @Test
+  public void noUpdateWhenNullProvider() {
+    Metadata metadata = new Metadata();
+    when(mockUserAgentProvider.get()).thenReturn(null);
+    when(mockHeartBeatProvider.get()).thenReturn(null);
```
vkryachko: pls add 2 tests when only one of the providers is null

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
```diff
@@ -0,0 +1,60 @@
...
+package com.google.firebase.firestore.remote;
+
+import androidx.annotation.NonNull;
+import com.google.firebase.heartbeatinfo.HeartBeatInfo;
+import com.google.firebase.inject.Provider;
+import com.google.firebase.platforminfo.UserAgentPublisher;
+import io.grpc.Metadata;
+
+/**
+ * Class provides an implementation of the GrpcMetadataProvider interface.
```
vkryachko:
```suggestion
 * Provides an implementation of the GrpcMetadataProvider interface.
```

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
```diff
@@ -0,0 +1,60 @@
...
+  public FirebaseClientGrpcMetadataProvider(
+      @NonNull Provider<UserAgentPublisher> userAgentPublisherProvider,
+      @NonNull Provider<HeartBeatInfo> heartBeatInfoProvider) {
+    this.userAgentPublisherProvider = userAgentPublisherProvider;
+    this.heartBeatInfoProvider = heartBeatInfoProvider;
+  }
+
+  @Override
+  public void updateMetadata(@NonNull Metadata metadata) {
+    if (heartBeatInfoProvider.get() == null || userAgentPublisherProvider.get() == null) {
```
vkryachko: would it make sense to still send heartbeat if userAgentPublisher is null?

tools/measurement/apksize/src/inappmessagingdisplay/java/com.google.apksize/InAppMessagingDisplay.java
```diff
@@ -16,16 +16,16 @@
 
 import android.app.Activity;
 import android.content.Context;
+import androidx.annotation.NonNull;
```
vkryachko: Is this file change intentional?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
VinayGuthal: no i dont think so... Bcz with the userAgentPublisher is how we know which sdk is being used.

tools/measurement/apksize/src/inappmessagingdisplay/java/com.google.apksize/InAppMessagingDisplay.java
VinayGuthal: this is a part of googleJavaFormat

VinayGuthal: it just organized the imports

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -85,10 +87,12 @@ public FirestoreClient(
       DatabaseInfo databaseInfo,
       FirebaseFirestoreSettings settings,
       CredentialsProvider credentialsProvider,
-      final AsyncQueue asyncQueue) {
+      final AsyncQueue asyncQueue,
+      @Nullable GrpcMetadataProvider metadata) {
```
wilhuff: `metadataProvider`

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -87,12 +88,15 @@ public Datastore(
       DatabaseInfo databaseInfo,
       AsyncQueue workerQueue,
       CredentialsProvider credentialsProvider,
-      Context context) {
+      Context context,
+      @Nullable GrpcMetadataProvider grpcMetadataProvider) {
```
wilhuff: This could be `metadataProvider` like elsewhere.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -64,12 +64,16 @@
...
 
+  private final GrpcMetadataProvider grpcMetadataProvider;
+
   FirestoreChannel(
       AsyncQueue asyncQueue,
       Context context,
       CredentialsProvider credentialsProvider,
-      DatabaseInfo databaseInfo) {
+      DatabaseInfo databaseInfo,
+      GrpcMetadataProvider metadata) {
```
wilhuff: `metadataProvider` and possibly make `this.grpcMetadataProvider` match.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -109,7 +111,8 @@ static FirebaseFirestore newInstance(
       @NonNull FirebaseApp app,
       @Nullable InternalAuthProvider authProvider,
       @NonNull String database,
-      @NonNull InstanceRegistry instanceRegistry) {
+      @NonNull InstanceRegistry instanceRegistry,
+      @Nullable GrpcMetadataProvider metadata) {
```
vkryachko:
```suggestion
      @Nullable GrpcMetadataProvider metadataProvider) {
```

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -134,7 +137,7 @@ static FirebaseFirestore newInstance(
 
     FirebaseFirestore firestore =
         new FirebaseFirestore(
-            context, databaseId, persistenceKey, provider, queue, app, instanceRegistry);
+            context, databaseId, persistenceKey, provider, queue, app, instanceRegistry, metadata);
```
vkryachko:
```suggestion
            context, databaseId, persistenceKey, provider, queue, app, instanceRegistry, metadataProvider);
```

PR #790, opened 2019-09-12
Simplify IndexFree tests

schmidt-sebastian:
This simplifies the newly added Index-Free tests in LocalStoreTests:

- Instead of using three docs, most tests only need two.
- Removed some setup and asserts that were unrelated to the actual test.

### DIFF
--- a/firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
+++ b/firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
@@ -418,13 +418,15 @@ public class TestUtil {
               }
             });
 
+    SnapshotVersion version = SnapshotVersion.NONE;
+
     for (MaybeDocument doc : docs) {
       DocumentChange change =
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -418,13 +418,15 @@ public QueryData getQueryDataForTarget(int targetId) {
               }
             });
 
+    SnapshotVersion version = SnapshotVersion.NONE;
+
     for (MaybeDocument doc : docs) {
       DocumentChange change =
           new DocumentChange(updatedInTargets, removedFromTargets, doc.getKey(), doc);
       aggregator.handleDocumentChange(change);
+      version = doc.getVersion().compareTo(version) > 0 ? doc.getVersion() : version;
```
schmidt-sebastian: Not sure why this worked before but `testQueriesIncludeDocumentsFromOtherQueries()` needs to use `20` as the snapshot version in order to use a higher read time for the newly added doc.

PR #800, opened 2019-09-13
Remove Index-Free assert

schmidt-sebastian:
This PR removes the assert that causes a crash for one of our customers: 

```
Caused by: java.lang.AssertionError: INTERNAL ASSERTION FAILED: Cannot add document to the RemoteDocumentCache with a read time of zero
        at com.google.firebase.firestore.g.b.a(com.google.firebase:firebase-firestore@@21.1.0:46)
        at com.google.firebase.firestore.g.b.a(com.google.firebase:firebase-firestore@@21.1.0:31)
```

This is the first step in fixing this issue. As a follow up, we have to fix the underlying root cause and clean up all affected data.

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 21.1.1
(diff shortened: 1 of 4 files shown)


schmidt-sebastian:
CL added. PTAL.

PR #802, opened 2019-09-13
Change the order of checks in applyRemoteEvent()

schmidt-sebastian:
If there is a chance that a limbo resolution fails for a document that no longer exists, we could get into a state where we write a document into the RemoteDocumentCache with a snapshot version of zero. By re-ordering the statements, we make sure this does not happen

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
@@ -384,7 +384,7 @@ public final class LocalStore {
             // rejected limbo resolutions (which fabricate NoDocuments with SnapshotVersion.NONE)
             // never add documents to cache.
             if (doc instanceof NoDocument && doc.getVersion().equals(SnapshotVersion.NONE)) {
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -380,18 +380,23 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
             MaybeDocument doc = entry.getValue();
             MaybeDocument existingDoc = existingDocs.get(key);
 
-            if (existingDoc == null
+            // Note: The order of the steps below is important, since we want to ensure that
+            // rejected limbo resolutions (which fabricate NoDocuments with SnapshotVersion.NONE)
+            // never add documents to cache.
+            if (doc instanceof NoDocument && doc.getVersion().equals(SnapshotVersion.NONE)) {
+              // NoDocuments with SnapshotVersion.MIN are used in manufactured events. We remove
```
wilhuff: s/MIN/NONE/?

schmidt-sebastian: Fixed

Issue #706, opened 2019-08-14
Bug-SQLite Exception during removeOrphanedDocuments


highthunder:
### [READ] Step 1: Are you in the right place?

Yes, this is an issue with Cloud Firestore's internal on-device sqlite database and this repository appears to contain all of the code required to manage that database.

### [REQUIRED] Step 2: Describe your environment

Android Studio version: 3.4.2
Firebase Component: Cloud Firestore
Component version: 19.0.2 (via firebase-bom:19.0.0)
Compile/Target SDK version: 28

### [REQUIRED] Step 3: Describe the problem

This is probably the same issue that I reported a few months ago where user's with over ~7000 rows crashed during a migration, just with a different query:
https://github.com/firebase/firebase-android-sdk/issues/370

#### Stack trace:

```
# OS Version: 7.1.1
# Device: Galaxy Tab E 8.0
# RAM Free: 32.9%
# Disk Free: 57.6%

#0. Crashed: main
       at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$5 + 379(AsyncQueue.java:379)
       at com.google.firebase.firestore.util.AsyncQueue$$Lambda$5.run(AsyncQueue.java)
       at android.os.Handler.handleCallback + 751(Handler.java:751)
       at android.os.Handler.dispatchMessage + 95(Handler.java:95)
       at android.os.Looper.loop + 154(Looper.java:154)
       at android.app.ActivityThread.main + 6823(ActivityThread.java:6823)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run + 1557(ZygoteInit.java:1557)
       at com.android.internal.os.ZygoteInit.main + 1445(ZygoteInit.java:1445)

--

Fatal Exception: java.lang.RuntimeException: Internal error in Firestore (19.0.2).
       at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$5 + 379(AsyncQueue.java:379)
       at com.google.firebase.firestore.util.AsyncQueue$$Lambda$5.run(AsyncQueue.java)
       at android.os.Handler.handleCallback + 751(Handler.java:751)
       at android.os.Handler.dispatchMessage + 95(Handler.java:95)
       at android.os.Looper.loop + 154(Looper.java:154)
       at android.app.ActivityThread.main + 6823(ActivityThread.java:6823)
       at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #803, opened 2019-09-14
Split orphaned document GC into batches

schmidt-sebastian:
This applies the same fix as https://github.com/firebase/firebase-android-sdk/pull/374 to orphan document removal. I used the same batch size since both operations just read document kets.

Fixes https://github.com/firebase/firebase-android-sdk/issues/706

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteLruReferenceDelegate.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteLruReferenceDelegate.java
@@ -24,6 +24,14 @@ import com.google.firebase.firestore.util.Consumer;
 
 /** Provides LRU functionality for SQLite persistence. */
 class SQLiteLruReferenceDelegate implements ReferenceDelegate, LruDelegate {
+  /**
+   * The batch size for orphaned document GC in `removeOrphanedDocuments()`.
+   *
+   * <p>This addresses https://github.com/firebase/firebase-android-sdk/issues/706, where a customer
+   * reported that LRU GC hit a CursorWindow size limit during orphaned document removal.
+   */
+  static final int REMOVE_ORPHANED_DOCUMENTS_BATCH_SIZE = 100;
+
   private final SQLitePersistence persistence;
   private ListenSequence listenSequence;
   private long currentSequenceNumber;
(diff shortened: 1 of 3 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/LruGarbageCollectorTestCase.java
```diff
@@ -401,6 +401,30 @@ public void testRemoveOrphanedDocuments() {
         });
   }
 
+  @Test
+  public void testRemoveOrphanedDocumentsExistsEarlyWithNoDocuments() {
```
wu-hui: Exists -> Exits ?

wu-hui: Also it's not testing `EarlyExits` but the case where no documents are collectable? If my understanding is correct, maybe rename the test?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteLruReferenceDelegate.java
```diff
@@ -148,20 +156,29 @@ private void removeSentinel(DocumentKey key) {
   @Override
   public int removeOrphanedDocuments(long upperBound) {
     int[] count = new int[1];
```
wu-hui: Not related to this change, but it'd be nicer if we have a brief implementation comment here to explain why `int[]`, something like:

`// Using int[1] instead of int such that it is update-able from lambdas`.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LruGarbageCollectorTestCase.java
schmidt-sebastian: Fixed both issues as once by renaming this to`testRemoveOrphanedDocumentsWithNoDocuments()`

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteLruReferenceDelegate.java
schmidt-sebastian: I'm gonna leave this as is for now. I think this is a common enough (anti-)pattern that we don't need to point it out explicitly. Thanks for the suggestion though!

PR #805, opened 2019-09-16
Implement AAR size measurement.

rlazo:
The new task `generateAarSizeMeasurements` generates either a table or a JSON report with the size, in bytes, of the release AAR generated by each subproject. It follows the same logic, and has the same flags, as `generateApkSizeMeasurements`.

### DIFF
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/aarsize/GenerateMeasurementsTask.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/aarsize/GenerateMeasurementsTask.groovy
@@ -27,7 +27,7 @@ import org.gradle.api.tasks.TaskAction
  * Generates size measurements after building the release aar's.
  *
  * <p>This task can run in two modes. The first mode, enabled when running the task with the
- * {@code pull_request} flag set, is a dependency of {@linkUploadMeasurementsTask} and generates
+ * {@code pull_request} flag set, is a dependency of {@link UploadMeasurementsTask} and generates
  * a JSON file with measurement information. The second mode, enabled when running the task without
  * flags, outputs a table to standard out with more human-readable information. See the README for
  * more details.
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/aarsize/GenerateMeasurementsTask.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/aarsize/GenerateMeasurementsTask.groovy
@@ -1,0 +1,120 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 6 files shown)


tools/measurement/apksize/default.gradle
```diff
@@ -30,7 +30,7 @@ android {
     defaultConfig {
         applicationId 'com.google.apksize'
         minSdkVersion project.targetSdkVersion
-	    multiDexEnabled true
+            multiDexEnabled true
```
allisonbm92: Should this be aligned with the other lines?

tools/measurement/apksize/sdks_ignore.csv
```diff
@@ -0,0 +1,5 @@
+firebase-database-collection
```
allisonbm92: Why ignore these?

tools/measurement/apksize/default.gradle
rlazo: Done.

tools/measurement/apksize/sdks_ignore.csv
rlazo: @vkryachko can confirm, but I have the impression that these are not actual sdks, but more like helper code or backend, so I wouldn't like to include those in the report.

allisonbm92: It would still be beneficial to know their sizes, don't you think?

tools/measurement/apksize/apksize.gradle
```diff
@@ -42,6 +42,15 @@ task generateApkSizeMeasurements(type: GenerateMeasurementsTask) {
     reportFile = file("$buildDir/size-report.json")
 }
 
+task generateAarSizeMeasurements(type: com.google.firebase.gradle.plugins.measurement.aarsize.GenerateMeasurementsTask) {
```
yifanyang: I'm not an expert of groovy. Does it support import aliases like python or typescript?

tools/measurement/apksize/apksize.gradle
```diff
@@ -25,7 +25,7 @@ apply from: "src/config/config.gradle"
 apply from: "src/database/database.gradle"
 apply from: "src/firestore/firestore.gradle"
 apply from: "src/functions/functions.gradle"
-apply from: "src/inappmessagingdisplay/inappmessaging-display.gradle"
+// apply from: "src/inappmessagingdisplay/inappmessaging-display.gradle"
```
yifanyang: Is this still needed?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/aarsize/GenerateMeasurementsTask.groovy
```diff
@@ -0,0 +1,120 @@
...
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.Optional
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+
+/**
+ * Generates size measurements after building the release aar's.
+ *
+ * <p>This task can run in two modes. The first mode, enabled when running the task with the
+ * {@code pull_request} flag set, is a dependency of {@linkUploadMeasurementsTask} and generates
```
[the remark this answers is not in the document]
rlazo: Good catch!

tools/measurement/apksize/apksize.gradle
rlazo: It is. I commented it out during testing. Fixed.

tools/measurement/apksize/sdks_ignore.csv
rlazo: Done. PTAL.

tools/measurement/apksize/apksize.gradle
rlazo: I didn't know either, but it does. Fixed!

PR #809, opened 2019-09-17
Refresh Auth token if expired during getId call.

ankitaj224:
Addressing Rayo's comment on another PR which was marked as a TODO.

https://github.com/firebase/firebase-android-sdk/pull/703#discussion_r314100359

### DIFF
--- a/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
+++ b/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
@@ -23,7 +23,6 @@ import static com.google.firebase.installations.FisAndroidTestConstants.TEST_AUT
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_AUTH_TOKEN_4;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_2;
-import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_3;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_FID_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_PROJECT_ID;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_REFRESH_TOKEN;
@@ -88,7 +87,7 @@ public class FirebaseInstallationsInstrumentedTest {
           .setFirebaseInstallationId(TEST_FID_1)
           .setAuthToken(TEST_AUTH_TOKEN)
           .setRefreshToken(TEST_REFRESH_TOKEN)
-          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_1)
           .setExpiresInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP)
           .setRegistrationStatus(PersistedFid.RegistrationStatus.REGISTERED)
           .build();
@@ -98,7 +97,7 @@ public class FirebaseInstallationsInstrumentedTest {
           .setFirebaseInstallationId(TEST_FID_1)
           .setAuthToken(TEST_AUTH_TOKEN)
           .setRefreshToken(TEST_REFRESH_TOKEN)
-          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_1)
           .setExpiresInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP_2)
           .setRegistrationStatus(PersistedFid.RegistrationStatus.REGISTERED)
           .build();
@@ -108,7 +107,7 @@ public class FirebaseInstallationsInstrumentedTest {
           .setFirebaseInstallationId(TEST_FID_1)
           .setAuthToken("")
           .setRefreshToken("")
-          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_1)
           .setExpiresInSecs(0)
           .setRegistrationStatus(PersistedFid.RegistrationStatus.UNREGISTERED)
           .build();
@@ -118,7 +117,7 @@ public class FirebaseInstallationsInstrumentedTest {
           .setFirebaseInstallationId(TEST_FID_1)
           .setAuthToken(TEST_AUTH_TOKEN_2)
           .setRefreshToken(TEST_REFRESH_TOKEN)
-          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_3)
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
           .setExpiresInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP)
           .setRegistrationStatus(PersistedFid.RegistrationStatus.REGISTERED)
           .build();
@@ -179,7 +178,7 @@ public class FirebaseInstallationsInstrumentedTest {
             mockClock, executor, firebaseApp, backendClientReturnsOk, persistedFid, mockUtils);
 
     // No exception, means success.
-    assertWithMessage("getId Task fails.")
+    assertWithMessage("getId Task failed")
         .that(Tasks.await(firebaseInstallations.getId()))
         .isNotEmpty();
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
@@ -206,7 +205,7 @@ public class FirebaseInstallationsInstrumentedTest {
             mockClock, executor, firebaseApp, backendClientReturnsOk, persistedFid, mockUtils);
 
     // No exception, means success.
-    assertWithMessage("getId Task fails.")
+    assertWithMessage("getId Task failed")
         .that(Tasks.await(firebaseInstallations.getId()))
         .isNotEmpty();
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
@@ -287,7 +286,7 @@ public class FirebaseInstallationsInstrumentedTest {
         new FirebaseInstallations(
             mockClock, executor, firebaseApp, backendClientReturnsOk, persistedFid, mockUtils);
 
-    assertWithMessage("getId Task fails.")
+    assertWithMessage("getId Task failed")
         .that(Tasks.await(firebaseInstallations.getId()))
         .isNotEmpty();
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
@@ -437,14 +436,13 @@ public class FirebaseInstallationsInstrumentedTest {
   @Test
   public void testGetAuthToken_multipleCallsDoNotForceRefresh_fetchedNewTokenOnce()
       throws Exception {
-    // Using mockPersistedFid to ensure the order of returning persistedFidEntry to 2 tasks
-    // triggered simultaneously. Task2 waits for Task1 to complete. On Task1 completion, task2 reads
-    // the UPDATED_AUTH_TOKEN_FID_ENTRY by Task1 on execution.
-    when(mockPersistedFid.readPersistedFidEntryValue())
-        .thenReturn(EXPIRED_AUTH_TOKEN_ENTRY, UPDATED_AUTH_TOKEN_FID_ENTRY);
+    // Update local storage with fid entry that has auth token expired. When 2 tasks are triggered
+    // simultaneously, Task2 waits for Task1 to complete. On Task1 completion, task2 reads  the
+    // UPDATED_AUTH_TOKEN_FID_ENTRY by Task1 on execution.
+    persistedFid.insertOrUpdatePersistedFidEntry(EXPIRED_AUTH_TOKEN_ENTRY);
     FirebaseInstallations firebaseInstallations =
         new FirebaseInstallations(
-            mockClock, executor, firebaseApp, backendClientReturnsOk, mockPersistedFid, mockUtils);
+            mockClock, executor, firebaseApp, backendClientReturnsOk, persistedFid, mockUtils);
 
     // Call getAuthToken multiple times with DO_NOT_FORCE_REFRESH option
     Task<InstallationTokenResult> task1 =
--- a/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
+++ b/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
@@ -15,6 +15,7 @@
 package com.google.firebase.installations;
 
 import static com.google.common.truth.Truth.assertWithMessage;
+import static com.google.firebase.installations.FisAndroidTestConstants.DEFAULT_PERSISTED_FID_ENTRY;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_API_KEY;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_APP_ID_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_AUTH_TOKEN;
@@ -163,7 +164,8 @@ public class FirebaseInstallationsInstrumentedTest {
             new FirebaseInstallationServiceException(
                 "SDK Error", FirebaseInstallationServiceException.Status.SERVER_ERROR));
     when(persistedFidReturnsError.insertOrUpdatePersistedFidEntry(any())).thenReturn(false);
-    when(persistedFidReturnsError.readPersistedFidEntryValue()).thenReturn(null);
+    when(persistedFidReturnsError.readPersistedFidEntryValue())
+        .thenReturn(DEFAULT_PERSISTED_FID_ENTRY);
     when(mockUtils.createRandomFid()).thenReturn(TEST_FID_1);
     when(mockClock.currentTimeMillis()).thenReturn(TEST_CREATION_TIMESTAMP_1);
     // Mocks success on FIS deletion
@@ -533,7 +535,9 @@ public class FirebaseInstallationsInstrumentedTest {
     Tasks.await(firebaseInstallations.delete());
 
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
-    assertWithMessage("Persisted Fid Entry is not null.").that(entryValue).isNull();
+    assertWithMessage("Persisted Fid Entry is not null.")
+        .that(entryValue)
+        .isEqualTo(DEFAULT_PERSISTED_FID_ENTRY);
     verify(backendClientReturnsOk, times(1))
         .deleteFirebaseInstallation(TEST_API_KEY, TEST_FID_1, TEST_PROJECT_ID, TEST_REFRESH_TOKEN);
   }
@@ -549,7 +553,9 @@ public class FirebaseInstallationsInstrumentedTest {
     Tasks.await(firebaseInstallations.delete());
 
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
-    assertWithMessage("Persisted Fid Entry is not null.").that(entryValue).isNull();
+    assertWithMessage("Persisted Fid Entry is not null.")
+        .that(entryValue)
+        .isEqualTo(DEFAULT_PERSISTED_FID_ENTRY);
     verify(backendClientReturnsOk, never())
         .deleteFirebaseInstallation(TEST_API_KEY, TEST_FID_1, TEST_PROJECT_ID, TEST_REFRESH_TOKEN);
   }
@@ -563,7 +569,9 @@ public class FirebaseInstallationsInstrumentedTest {
     Tasks.await(firebaseInstallations.delete());
 
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
-    assertWithMessage("Persisted Fid Entry is not null.").that(entryValue).isNull();
+    assertWithMessage("Persisted Fid Entry is not null.")
+        .that(entryValue)
+        .isEqualTo(DEFAULT_PERSISTED_FID_ENTRY);
     verify(backendClientReturnsOk, never())
         .deleteFirebaseInstallation(TEST_API_KEY, TEST_FID_1, TEST_PROJECT_ID, TEST_REFRESH_TOKEN);
   }
--- a/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
+++ b/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
@@ -23,6 +23,7 @@ import static com.google.firebase.installations.FisAndroidTestConstants.TEST_AUT
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_AUTH_TOKEN_4;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_2;
+import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_3;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_FID_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_PROJECT_ID;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_REFRESH_TOKEN;
@@ -32,6 +33,7 @@ import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -116,7 +118,7 @@ public class FirebaseInstallationsInstrumentedTest {
           .setFirebaseInstallationId(TEST_FID_1)
           .setAuthToken(TEST_AUTH_TOKEN_2)
           .setRefreshToken(TEST_REFRESH_TOKEN)
-          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_3)
           .setExpiresInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP)
           .setRegistrationStatus(PersistedFid.RegistrationStatus.REGISTERED)
           .build();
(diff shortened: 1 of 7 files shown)


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +235,45 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
       PersistedFidEntry persistedFidEntry, AwaitListener listener) {
     String fid = persistedFidEntry.getFirebaseInstallationId();
 
-    // Check if the fid is unregistered
-    if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
-      updatePersistedFidWithPendingStatus(fid);
-      executeFidRegistration(persistedFidEntry, listener);
-    } else {
-      updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
+    if (!isFidUnregisteredOrAuthTokenExpired(persistedFidEntry)) {
```
diwu-arete: I prefer just putting this code inline here
persistedFidEntry.getRegistrationStatus() != RegistrationStatus.UNREGISTERED
        && !isAuthTokenExpired(persistedFidEntry)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -273,12 +283,20 @@ private void executeFidRegistration(PersistedFidEntry persistedFidEntry, AwaitLi
...
+   * completion.
+   */
+  private void refreshAuthToken(AwaitListener listener) {
+    Task<InstallationTokenResult> task = Tasks.call(executor, () -> refreshAuthTokenIfNecessary(1));
+    if (listener != null) {
+      task.addOnCompleteListener((unused) -> listener.onSuccess());
+    }
+  }
+
+  private void updatePersistedFidWithPendingStatus(PersistedFidEntry persistedFidEntry) {
```
diwu-arete: A general comment : I feel like you are doing functionalization too much in this class.

Functionalization is useful when you want to hide a chunk of detailed logic, or to avoid duplicated code. But for example for updatePersistedFidWithPendingStatus, this function is only used in one place(doesn't avoid duplicated code), and it's not hiding any detailed logic (use one line to replace another line)

A downside of over functionalization is, with too many small method, it's really hard to find the relationship between them.

Above all, for these cases, I prefer making the code inline than making as a function.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -355,7 +373,8 @@ private InstallationTokenResult getValidAuthToken(PersistedFidEntry persistedFid
 
   private boolean isPersistedFidRegistered(PersistedFidEntry persistedFidEntry) {
     return persistedFidEntry != null
-        && persistedFidEntry.getRegistrationStatus() == RegistrationStatus.REGISTERED;
+        && (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.REGISTERED
+            || persistedFidEntry.getRegistrationStatus() == RegistrationStatus.PENDING);
```
diwu-arete: This can be removed now, right?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +235,41 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+      return updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
     }
 
+    updatePersistedFidWithPendingStatus(persistedFidEntry);
+    executeFISCalls(persistedFidEntry, listener);
     return Tasks.forResult(fid);
   }
 
-  private void updateAwaitListenerIfRegisteredFid(
+  private void executeFISCalls(PersistedFidEntry persistedFidEntry, AwaitListener listener) {
```
diwu-arete: Through the method name, it confusing that what the difference is between executeFISCalls and executeFidRegistration

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +235,41 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+
+    if (!isAuthTokenExpired(persistedFidEntry)) {
+      executeFidRegistration(persistedFidEntry, listener);
+      return;
+    }
+    refreshAuthToken(listener);
```
diwu-arete: How about removing this method and putting the code inline here.

It confusing that you have refreshAuthToken and refreshAuthTokenIfNecessary

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
diwu-arete: executeFISCallsAsync?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -273,12 +283,20 @@ private void executeFidRegistration(PersistedFidEntry persistedFidEntry, AwaitLi
     }
```
diwu-arete: This is only used in one place, how about putting as inline code and you can also remove listener != null check.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -277,6 +279,36 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
 
+  @Test
+  public void testGetId_expiredAuthToken_refreshesAuthToken() throws Exception {
+    // Update local storage with fid entry that has auth token expired.
+    persistedFid.insertOrUpdatePersistedFidEntry(EXPIRED_AUTH_TOKEN_ENTRY);
+    FirebaseInstallations firebaseInstallations =
+        new FirebaseInstallations(
+            mockClock, executor, firebaseApp, backendClientReturnsOk, persistedFid, mockUtils);
+
+    assertWithMessage("getId Task fails.")
```
ciarand: s/failed/fails

firebase-installations/src/androidTest/java/com/google/firebase/installations/FisAndroidTestConstants.java
```diff
@@ -36,4 +36,5 @@
 
   public static final long TEST_CREATION_TIMESTAMP_1 = 2000L;
   public static final long TEST_CREATION_TIMESTAMP_2 = 2000L;
```
ciarand: what's the point of having TIMESTAMP_1 and TIMESTAMP_2 if they're the same?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +235,41 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
       PersistedFidEntry persistedFidEntry, AwaitListener listener) {
     String fid = persistedFidEntry.getFirebaseInstallationId();
 
-    // Check if the fid is unregistered
-    if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
-      updatePersistedFidWithPendingStatus(fid);
-      executeFidRegistration(persistedFidEntry, listener);
-    } else {
-      updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
+    if (persistedFidEntry.getRegistrationStatus() != RegistrationStatus.UNREGISTERED
```
ciarand: Can you add a helper for this? `PersistedFidEntry#isRegistered`

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -273,12 +283,20 @@ private void executeFidRegistration(PersistedFidEntry persistedFidEntry, AwaitLi
     }
   }
 
-  private void updatePersistedFidWithPendingStatus(String fid) {
+  /**
+   * Refreshes FID auth token with FIS servers in a background thread and updates the listener on
+   * completion.
+   */
+  private void refreshAuthToken(AwaitListener listener) {
+    Task<InstallationTokenResult> task = Tasks.call(executor, () -> refreshAuthTokenIfNecessary(1));
```
ciarand: never pass the raw value for an `@intdef` constant; treat it like an enum and always use the predefined constants

(Gradle has a check for this afair, but I don't know if this repo is configured to run them)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: imo the name of this method is now misleading

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Thanks for the detailed explanation. Updated the code. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: As discussed offline, we need this check.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: No. We need this check. With this new flow - both getId and getAuthToken need this. 

getAuthToken uses it with persistedFid  in REGISTERED state.
getId uses it with persistedFid  in PENDING state.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +231,50 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+    // token, execute Fid registration.
+    if (!isAuthTokenExpired(persistedFidEntry)) {
       executeFidRegistration(persistedFidEntry, listener);
-    } else {
-      updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
+      return Tasks.forResult(fid);
     }
 
+    // If auth token is expired, refresh FID auth token with FIS servers in a background thread and
+    // updates the listener on
```
diwu-arete: nit : in one line.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -277,6 +279,36 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+        .that(Tasks.await(firebaseInstallations.getId()))
+        .isNotEmpty();
+    PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
+    assertWithMessage("Persisted Fid doesn't match")
+        .that(entryValue.getFirebaseInstallationId())
+        .isEqualTo(TEST_FID_1);
+
+    // Waiting for Task that generates auth token with the FIS Servers
+    executor.awaitTermination(500, TimeUnit.MILLISECONDS);
```
ciarand: really don't like this (feels like a weird hack when the test controls the underlying executorservice), but it's more of a theoretical concern than a pragmatic one

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +235,41 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
-      executeFidRegistration(persistedFidEntry, listener);
-    } else {
-      updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
+    if (persistedFidEntry.getRegistrationStatus() != RegistrationStatus.UNREGISTERED
+        && !isAuthTokenExpired(persistedFidEntry)) {
+      return updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
     }
 
+    updatePersistedFidWithPendingStatus(persistedFidEntry);
+    executeFISCalls(persistedFidEntry, listener);
```
ciarand: I don't understand this method name

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +235,41 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+  private void executeFISCalls(PersistedFidEntry persistedFidEntry, AwaitListener listener) {
+
+    if (!isAuthTokenExpired(persistedFidEntry)) {
+      executeFidRegistration(persistedFidEntry, listener);
+      return;
+    }
+    refreshAuthToken(listener);
+  }
+
+  private Task<String> updateAwaitListenerIfRegisteredFid(
```
ciarand: is this only being used in one place? I agree with Di's comments here - inlining smaller methods like this may improve readability and make the code easier to understand

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,31 +231,50 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+    // If FID registration is in complete/pending state or FID auth token is valid, update the
+    // listener if awaiting and return the persisted fid.
+    if (persistedFidEntry.getRegistrationStatus() != RegistrationStatus.UNREGISTERED
+        && !isAuthTokenExpired(persistedFidEntry)) {
+      return updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
+    }
+
+    // Update FID registration status to PENDING to avoid multiple network calls to FIS Servers.
+    persistedFid.insertOrUpdatePersistedFidEntry(
+        persistedFidEntry.toBuilder().setRegistrationStatus(RegistrationStatus.PENDING).build());
```
diwu-arete: you need to keep "reading cache status" and "updating cache status" in one atomic section, to avoid duplicated network calls.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Good catch. I missed it. Updated.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FisAndroidTestConstants.java
ankitaj224: Good catch. I missed it. Updated.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Thank you so much. Updated. Any examples on how do I configure it?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated. PTAL.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ankitaj224: One way to solve this could be allowing the getID implementation that takes a listener public. For this, we need Android SDK review( I not 100% sure). 

@ciarand  and @diwu-arete  wdyt?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Added, as this method could be reused in getId and delete APIs.

ankitaj224:
/test device-check-changed
--

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ciarand: probably, probably not worth it

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,62 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
-
-    // Check if the fid is unregistered
-    if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
-      updatePersistedFidWithPendingStatus(fid);
-      executeFidRegistration(persistedFidEntry, listener);
-    } else {
-      updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
-    }
+  private Task<String> registerFidIfNecessary(AwaitListener listener) {
+    synchronized (persistedFid) {
```
ciarand: if you're going to start synchronizing these instead of piping the values through a Task chain then you should add @GuardedBy annotations http://errorprone.info/bugpattern/GuardedBy

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,62 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+      PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
```
ciarand: can this value be null? you're checking for null from this call in other places, so there prob should be an explanation of why this instance is safe

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,62 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+      String fid = persistedFidEntry.getFirebaseInstallationId();
+
+      // If FID registration status is PENDING, return the persisted fid. This avoids multiple
+      // network calls to FIS Servers.
+      if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.PENDING) {
```
ciarand: isPending?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -409,7 +419,7 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
 
     PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
 
-    if (isPersistedFidRegistered(persistedFidEntry)) {
+    if (persistedFid.isFidRegistered(persistedFidEntry)) {
```
ciarand: this isn't right, is it? if `persistedFidEntry` is null then the `isFidRegistered` method will return true and then we'll call `persistedFidEntry.getFirebaseInstallationId()` right below which will NPE

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -134,4 +134,10 @@ public boolean clear() {
   private String getSharedPreferencesKey(String key) {
     return String.format("%s|%s", persistenceKey, key);
   }
+
+  @NonNull
+  public boolean isFidRegistered(@Nullable PersistedFidEntry persistedFidEntry) {
```
ciarand: imo this should go on PersistedFidEntry or some sort of static util class... Otherwise it feels like we're just putting it here for convenience

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: if ```persistedFidEntry``` is null then the ```isFidRegistered``` method will return false.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: So, I am locking on persistedFid and not persistedFidEntry( was piping it earlier on this value through task).

Also, I cannot synchronize on persistedFid throughout this class.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Ideally not ( if empty, in the previous we create and persist fid).  However, to avoid any NPE, added null checks.

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
ankitaj224: Updated. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Changes this.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,81 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
-
-    // Check if the fid is unregistered
-    if (persistedFidEntry.getRegistrationStatus() == RegistrationStatus.UNREGISTERED) {
-      updatePersistedFidWithPendingStatus(fid);
-      executeFidRegistration(persistedFidEntry, listener);
-    } else {
-      updateAwaitListenerIfRegisteredFid(persistedFidEntry, listener);
+  private Task<String> registerFidIfNecessary(AwaitListener listener)
+      throws FirebaseInstallationsException {
+    Boolean isNetworkCallRequired = false;
```
ciarand: don't use a `Boolean` when you don't need the concept of nullability

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: Yes, my point was that synchronization isn't necessarily required if you're just piping them through a Task call stack.

Why are we synchronizing on it here if we can't do it consistently? What benefit does it provide here that it doesn't provide elsewhere? Can we document that more explicitly?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,81 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+      PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
+
+      if (persistedFidEntry.isNotGenerated()) {
+        throw new FirebaseInstallationsException(
+            "Local storage has no persisted Fid entry.",
+            FirebaseInstallationsException.Status.CLIENT_ERROR);
+      }
+
+      // If FID registration status is REGISTERED and auth token is expired, or registration status
+      // is UNREGISTERED network call iks required.
```
ciarand: nit: iks

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
diwu-arete: boolean

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
```diff
@@ -85,7 +87,7 @@ public PersistedFidEntry readPersistedFidEntryValue() {
     if (fid == null
         || status == -1
```
ciarand: `status == -1` is unnecessary now

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,81 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
 
-  private void updateAwaitListenerIfRegisteredFid(
-      PersistedFidEntry persistedFidEntry, AwaitListener listener) {
-    if (listener != null
-        && persistedFidEntry.getRegistrationStatus() == RegistrationStatus.REGISTERED) {
-      listener.onSuccess();
+    if (updatedPersistedFidEntry.isNotGenerated()) {
+      throw new FirebaseInstallationsException(
+          "Local storage has no persisted Fid entry.",
+          FirebaseInstallationsException.Status.CLIENT_ERROR);
```
ciarand: is this check just making sure we were able to write to disk? it seems redundant with the one above

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,81 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
 
-  private void updatePersistedFidWithPendingStatus(String fid) {
-    persistedFid.insertOrUpdatePersistedFidEntry(
-        PersistedFidEntry.builder()
-            .setFirebaseInstallationId(fid)
-            .setRegistrationStatus(RegistrationStatus.PENDING)
-            .build());
+    // If fid registration for this firebase installation is incomplete without an expired auth
+    // token, execute Fid registration.
+    if (!isAuthTokenExpired(updatedPersistedFidEntry)) {
```
ciarand: isn't this being checked twice in the same body?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -235,50 +230,81 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
...
+    }
+
+    // If auth token is expired, refresh FID auth token with FIS servers in a background thread
+    // and updates the listener on completion
+    Task<InstallationTokenResult> refreshAuthTokenTask =
+        Tasks.call(executor, () -> refreshAuthTokenIfNecessary(FORCE_REFRESH));
+
+    // Update the listener if awaiting
+    if (listener != null) {
+      refreshAuthTokenTask.addOnCompleteListener((unused) -> listener.onSuccess());
```
ciarand: you're turning an onComplete into an onSuccess, this probably needs an explanation

ankitaj224: changed it to onSuccess. This is to update the listener  asap on success instead of making it wait for 10s.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedFid.java
ankitaj224: Done. Thanks.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: PENDING state was introduced to avoid multiple network calls for getId(). Registration status changes from UNREGISTERED to PENDING before triggering the network call.

Need for synchronization: Reading the existing registration status and updating it to PENDING should be atomic. That's the block i tried to synchronize. 

I think piping wont solve the atomicity issue.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Yes, it is. First time its checked to evaluate if we need network calls. Second check helps to decide what type of network call is required.

Network call should be made : If UNREGISTERED or AuthToken Expired
Before triggering the call, we update the status to PENDING.

So, we rely on  AuthToken Expired check. 

If AuthToken not Expired: REGISTER Fid
else: Refresh Auth Token

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -418,18 +450,13 @@ public void testGetAuthToken_serverError_failure() throws Exception {
...
-    // triggered simultaneously. Task2 waits for Task1 to complete. On Task1 completion, task2 reads
-    // the UPDATED_AUTH_TOKEN_FID_ENTRY by Task1 on execution.
-    when(mockPersistedFid.readPersistedFidEntryValue())
-        .thenReturn(
-            EXPIRED_AUTH_TOKEN_ENTRY,
-            EXPIRED_AUTH_TOKEN_ENTRY,
-            EXPIRED_AUTH_TOKEN_ENTRY,
-            UPDATED_AUTH_TOKEN_FID_ENTRY);
+    // Update local storage with fid entry that has auth token expired. When 2 tasks are triggered
+    // simultaneously, Task2 waits for Task1 to complete. On Task1 completion, task2 reads  the
```
ciarand: nit: "reads  the" has too many (two) spaces

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -508,7 +535,9 @@ public void testDelete_registeredFID_successful() throws Exception {
     Tasks.await(firebaseInstallations.delete());
 
     PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
-    assertWithMessage("Persisted Fid Entry is not null.").that(entryValue).isNull();
+    assertWithMessage("Persisted Fid Entry is not null.")
```
ciarand: the "is not null" message isn't correct any more

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: I'm having a really really hard time understanding the flow of this method. I don't know if it's just that I lack the domain knowledge or familiarity with the terms, but the multiple return points, double checking in and out of the synch block, and sometimes-async kickoffs are breaking my brain.

andirayo (commented):
Removing this from my list of review-requests.

PR #818, opened 2019-09-18
Fix AlarmManagerScheduler timestamp computation.

rlazo:
AlarmManager expects time to be specified as milliseconds since epoch, instead of the ms since when invoked.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerScheduler.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerScheduler.java
@@ -19,7 +19,6 @@ import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
-import android.os.SystemClock;
 import android.util.Base64;
 import androidx.annotation.VisibleForTesting;
 import com.google.android.datatransport.runtime.TransportContext;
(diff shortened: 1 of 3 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerScheduler.java
```diff
@@ -98,7 +106,7 @@ public void schedule(TransportContext transportContext, int attemptNumber) {
 
     long backendTime = eventStore.getNextCallTime(transportContext);
 
-    long scheduleDelay =
+    long scheduleDelay = clock.getTime() +
```
vkryachko: Can you add `clock.getTime()` directly to the `alarmManager.set()` so that logging below does not change?

rlazo: Done.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerScheduler.java
```diff
@@ -44,12 +47,15 @@
...
 
+  private final Clock clock;
+
   public AlarmManagerScheduler(
       Context applicationContext, EventStore eventStore, SchedulerConfig config) {
     this(
         applicationContext,
         eventStore,
         (AlarmManager) applicationContext.getSystemService(Context.ALARM_SERVICE),
+        new WallTimeClock(),
```
vkryachko: Also, for consistency, can you inject this clock via the `SchedulingModule`?

And actually, since we are using ELAPSED_REALTIME, you want the `@Monotonic Clock` not the Wall Time one.

rlazo: PTAL.

PR #822, opened 2019-09-18
Backport executeQuery() API changes

schmidt-sebastian:
From Index-Free feedback on Web (https://github.com/firebase/firebase-js-sdk/pull/2169)

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
@@ -598,25 +598,29 @@ public final class LocalStore {
         });
   }
 
-  /** Runs the given query against all the documents in the local store and returns the results. */
-  public ImmutableSortedMap<DocumentKey, Document> executeQuery(Query query) {
-    QueryData queryData = getQueryData(query);
-    if (queryData != null) {
-      ImmutableSortedSet<DocumentKey> remoteKeys =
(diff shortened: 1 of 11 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -598,25 +598,29 @@ public void releaseQuery(Query query) {
...
-   * advantage of the provided query data and the set of remote document keys.
-   */
-  public ImmutableSortedMap<DocumentKey, Document> executeQuery(
-      Query query, @Nullable QueryData queryData, ImmutableSortedSet<DocumentKey> remoteKeys) {
-    return queryEngine.getDocumentsMatchingQuery(query, queryData, remoteKeys);
+    ImmutableSortedMap<DocumentKey, Document> documents =
+        queryEngine.getDocumentsMatchingQuery(
+            query,
+            usePreviousResults ? lastLimboFreeSnapshotVersion : SnapshotVersion.NONE,
+            usePreviousResults ? remoteKeys : DocumentKey.emptyKeySet());
```
mikelehen: Feel free to ignore since you'd need to back-back-port, but it seems like maybe we should just have two overloads of QueryEngine.GetDocumnetsMatchingQuery() so we don't need to pass dummy values in.

schmidt-sebastian: Ignored.

:)

PR #823, opened 2019-09-19
Add Query to Target ID Map

schmidt-sebastian:
Ports: https://github.com/firebase/firebase-js-sdk/pull/2179/files


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -357,7 +361,7 @@ public SnapshotVersion getLastRemoteSnapshotVersion() {
                   oldQueryData
                       .withResumeToken(resumeToken, remoteEvent.getSnapshotVersion())
                       .withSequenceNumber(sequenceNumber);
-              targetIds.put(boxedTargetId, newQueryData);
+              queryDataByTarget.put(boxedTargetId, newQueryData);
```
mikelehen: preexisting nit: this will have to un-box so you might as well pass `targetId`

schmidt-sebastian: Fixed

PR #827, opened 2019-09-19
Enabling FIS network call timeout.

ankitaj224:
There is no timeout for HttpsURLConnection. Rayo mentioned other FIS SDKs  timeout after 10s.

### DIFF
--- a/firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
+++ b/firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
@@ -59,6 +59,8 @@ public class FirebaseInstallationServiceClient {
   private static final String X_ANDROID_PACKAGE_HEADER_KEY = "X-Android-Package";
   private static final String X_ANDROID_CERT_HEADER_KEY = "X-Android-Cert";
 
+  private static final int TIMEOUT_IN_MILLI_SECS = 10000;
+
   private final Context context;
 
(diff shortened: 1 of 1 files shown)


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -90,6 +92,8 @@ public InstallationResponse createFirebaseInstallation(
                   apiKey));
 
       HttpsURLConnection httpsURLConnection = (HttpsURLConnection) url.openConnection();
+      httpsURLConnection.setConnectTimeout(TIMEOUT_IN_MILLI_SECS);
+      httpsURLConnection.setReadTimeout(TIMEOUT_IN_MILLI_SECS);
```
ankitaj224: I want to verify if this is the right approach. I will then add it to all the network calls.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -59,6 +59,8 @@
   private static final String X_ANDROID_PACKAGE_HEADER_KEY = "X-Android-Package";
   private static final String X_ANDROID_CERT_HEADER_KEY = "X-Android-Cert";
 
+  private static final int TIMEOUT_IN_MILLI_SECS = 10000;
```
diwu-arete: NETWORK_TIMEOUT_MILLIS

PR #838, opened 2019-09-23
Add Firebase Dynamic Links

davidmotson:


firebase-dynamic-links/firebase-dynamic-links.gradle
```diff
@@ -0,0 +1,65 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+plugins {
+    id 'firebase-library'
+}
+
+firebaseLibrary {
```
allisonbm92: I would explicitly add `testLab.enabled = false` here.

It's the default, but I think it's best to make it explicit so we don't forget to set it to true after adding the tests.

PR #843, opened 2019-09-24
Support extra URL parameter in LegacyFlgDestination.

rlazo:


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -282,14 +282,18 @@ public BackendResponse send(BackendRequest request) {
...
+    if (destinationEndPoint != null && !destinationEndPoint.isEmpty()) {
+      actualEndPoint = parseUrlOrThrow(destination.getEndPoint());
+    }
 
     try {
       HttpResponse response =
           retry(
               5,
-              new HttpRequest(endPoint, requestBody, apiKey),
+              new HttpRequest(actualEndPoint, requestBody, destination.getAPIKey()),
```
vkryachko: since this code is shared between lflg and cct backends, doesn't it mean that we will end up always sending the API key to cct even though it's not needed there?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -46,22 +61,43 @@ public String getName() {
   @NonNull
   @Override
   public byte[] getExtras() {
-    return encodeString(apiKey);
+    return encodeAsExtras();
   }
 
   @NonNull
-  static byte[] encodeString(@NonNull String s) {
+  public static LegacyFlgDestination parseFromExtras(@Nullable byte[] a) {
```
vkryachko: Consider checking for null before calling into this method, especially given that the only time it is expected to be null is if called in the context of the `cct` backend, where the `DEFAULT_INSTANCE` might not make sense

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -46,22 +61,43 @@ public String getName() {
...
+    if (a == null) {
+      return LegacyFlgDestination.DEFAULT_INSTANCE;
+    }
     try {
-      return s.getBytes("UTF-8");
+      String encodedExtra = new String(a, "UTF-8");
+      if (!encodedExtra.startsWith(EXTRAS_VERSION_MARKER)) {
+        throw new IllegalStateException("Version marker missing from extras");
```
vkryachko: We need to make sure we don't propagate exceptions outside of the backend class, otherwise it will crash runtime

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
rlazo: You're right.  Fixed.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
rlazo: This was suppose to be an `EMPTY_INSTANCE`. I'll go with your suggestion and check before.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
rlazo: I'd made the `CctTransportBackend` handle the exception by logging the error and not trying to send the event. Does that sounds as a reasonable approach?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -282,14 +282,29 @@ public BackendResponse send(BackendRequest request) {
...
+        LegacyFlgDestination destination =
+            LegacyFlgDestination.parseFromExtras(request.getExtras());
+        apiKey = destination.getAPIKey();
+        String destinationEndPoint = destination.getEndPoint();
+        if (destinationEndPoint != null && !destinationEndPoint.isEmpty()) {
+          actualEndPoint = parseUrlOrThrow(destination.getEndPoint());
+        }
+      } catch (IllegalStateException e) {
+        Logging.e(LOG_TAG, "Couldn't parse extras", e);
+        return null;
```
vkryachko: I don't think we can return null from this method. I think it's appropriate to return `BackendResponse.fatal()` instead.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -46,22 +61,40 @@ public String getName() {
...
   }
 
   @NonNull
-  static byte[] encodeString(@NonNull String s) {
+  public static LegacyFlgDestination parseFromExtras(@NonNull byte[] a) {
     try {
-      return s.getBytes("UTF-8");
+      String encodedExtra = new String(a, "UTF-8");
+      if (!encodedExtra.startsWith(EXTRAS_VERSION_MARKER)) {
+        throw new IllegalStateException("Version marker missing from extras");
```
vkryachko: I think `IllegalArgumentException` is more appropriate as it's not the state of the Destination that is invalid, but rather the input is invalid.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -46,22 +61,40 @@ public String getName() {
...
+      }
+      encodedExtra = encodedExtra.substring(EXTRAS_VERSION_MARKER.length());
+      @SuppressWarnings("StringSplitter")
+      String[] fields = encodedExtra.split(EXTRAS_DELIMITER);
```
vkryachko: Instead of suppressing the warning pls use 
```suggestion
      String[] fields = encodedExtra.split(EXTRAS_DELIMITER, 2);
```

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -15,28 +15,43 @@
...
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import com.google.android.datatransport.runtime.Destination;
 import java.io.UnsupportedEncodingException;
 
 public class LegacyFlgDestination implements Destination {
   static final String DESTINATION_NAME = "lflg";
   private static final String DEFAULT_API_KEY =
       StringMerger.mergeStrings("AzSCki82AwsLzKd5O8zo", "IayckHiZRO1EFl1aGoK");
+  private static final String EXTRAS_VERSION_MARKER = "lfdv1$";
```
vkryachko: might make sense to drop the prefix?
```suggestion
  private static final String EXTRAS_VERSION_MARKER = "1$";
```

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/LegacyFlgDestination.java
```diff
@@ -15,28 +15,43 @@
...
+  private static final String EXTRAS_DELIMITER = "@";
```
vkryachko: Actually `@` is a perfectly valid symbol in URLs even though it's unlikely in our case. So I'd suggest using something more unique

PR #844, opened 2019-09-24
Merge Feature Rollouts functionality into Feature Rollouts branch

clp93:


transport/transport-backend-cct/gradle.properties
```diff
@@ -1,17 +1,3 @@
-# Copyright 2019 Google LLC
```
clp93: revert changes to this file.

PR #847, opened 2019-09-25
Simplifying FirebaseInstallations class by adding listeners.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,66 +159,86 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
-        return (T) t.getResult();
+  private void triggerOnStateReached(PersistedFidEntry persistedFidEntry) {
+    synchronized (lock) {
+      Iterator<StateListener> it = listeners.iterator();
+      while (it.hasNext()) {
+        StateListener l = it.next();
+        boolean doneListening = l.onStateReached(persistedFidEntry);
+        if (doneListening) {
+          it.remove();
+          break;
```
ciarand: I'm not sure whether we should keep the `break` or not. If we don't, listeners will finish earlier than they would normally which seems reasonable to me. WDYT?

firebase-installations/src/main/java/com/google/firebase/installations/StateListener.java
```diff
@@ -0,0 +1,79 @@
...
+package com.google.firebase.installations;
+
+import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.firebase.installations.local.PersistedFidEntry;
+
+interface StateListener {
+  boolean onStateReached(PersistedFidEntry persistedFidEntry);
+}
+
+class GetIdListener implements StateListener {
```
ciarand: Don't include multiple classes in a single file. This file should have one class / interface (StateListener) and the others should either be inner classes or separate files.

ciarand: https://google.github.io/styleguide/javaguide.html#s3.4.1-one-top-level-class

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,66 +159,86 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
+          String fid = utils.createRandomFid();
+          persistFid(fid);
+          persistedFidEntry = persistedFid.readPersistedFidEntryValue();
+        }
+
+        // GetIdListener will be notified as authTokenOption is always DO_NOT_FORCE_REFRESH.
+        // GetAuthTokenListener should only be notified if FORCE_REFRESH is not required.
+        if (authTokenOption != FORCE_REFRESH) {
+          triggerOnStateReached(persistedFidEntry);
+        }
```
ciarand: is this conditional necessary? couldn't we rely on the final triggerOnStateReached to kick this off?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: I dint think of the ```break;``` until my ```testGetAuthToken_multipleCallsForceRefresh_fetchedNewTokenTwice()``` test failed. 

Reason: Multiple getAuthToken(FOREC_REFRESH) calls - if first call is completed it triggers onStateReached for both calls that are in the list . 

GetAuthTokenListener onStateReached condition will pass (Registered & Auth Toke is valid) and remove both listeners from the list. 

However, the second call execution continues, generates new auth token and updates the shared pref with new values - resulting in discrepancy in what we report and we persist in FIS DB & local storage.

firebase-installations/src/main/java/com/google/firebase/installations/StateListener.java
ankitaj224: Thanks for the link. I thought about it but wasn't sure what the guidelines was. I will keep in mind for future changes.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: This check is needed to meet the purpose of FORCE_REFRESH. 

If clients request getAuthToken(FORCE_REFRESH) where persisted fid is Reistered and has valid AuthToken, we are still expected to generate a new Auth Token( forcefully refresh it).

Without the check 
``` if (authTokenOption != FORCE_REFRESH) { triggerOnStateReached(persistedFidEntry);} ```, we will notify the client with older auth token, remove the listener from the list and then refresh the auth token and persist it.

File: firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java (hunk 1 above)
Comment (ciarand) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: ignoring what the code actually does atm, what's the expected behavior of calling getAuthToken(FORCE_REFRESH) twice? Are we expecting the first Task to be completed with a soon-to-be invalid auth token and the second task will be completed with the newly-valid token? Or do we want both tasks to be completed with the same auth token?

Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: AFIK  we are expected to return the first Task on completion with a soon-to-be invalid auth token and the second task will be completed with the newly-valid token.  I verified this flow with Di and yet to verify it with Rayo.

IMO it would be good to have both tasks be completed with the same auth token. But wont we complicate the logic ( like Force Refresh behaves differently when one call is made vs multiple calls, or when multiple calls with do-not-force-refresh and force-refresh options are called).

Your thoughts.

Comment (ciarand) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: If that's the expected and desired outcome then I guess this is fine. I think it's a little error prone, but that's a product decision for your team and not a technical one that should be resolved here.

I think it would be clearer if we completed both tasks with the same result where possible, but to do that we'd have to keep the current auth token in memory and read it as part of our listener creation routine. I don't think that's prohibitively complicated (and it means we could simplify elsewhere by removing the break here and the conditional below), but that's a decision for your team.

Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: This is one corner case. I am yet to discuss and decide the outcome with the team. 

However, I tried incorporating your suggestion. Can you PTAL if the changes are correct now.

File: firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java [hunk 4]
```diff
@@ -159,66 +159,91 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
+
+        // Don't notify the listeners at this point; we might as well make ure the auth token is up
+        // to date before letting them know.
+
+        // Refresh Auth token if:
+        // 1. Invalid auth token
+        // 2. If FORCE_REFRESH with a pending listener
+        synchronized (lock) {
+          if ((authTokenOption == FORCE_REFRESH && !listeners.isEmpty())
```
ciarand: we can't hold the synchronized lock while doing network I/O because it'll block arbitrary threads that are calling getId / getAuthToken

ciarand: I get why we have this `!listeners.isEmpty()` check, but it feels like it's a symptom of what we're trying to check (is anyone still waiting for a state change?) instead of the actual thing we want to check (whether our current auth token needs to be refreshed)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,66 +159,91 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
-  private static boolean persistedFidMissingOrInErrorState(PersistedFidEntry persistedFidEntry) {
-    return persistedFidEntry == null
-        || persistedFidEntry.getRegistrationStatus() == RegistrationStatus.REGISTER_ERROR;
+  private Task<InstallationTokenResult> addGetAuthTokenListener(int authTokenOption) {
+    TaskCompletionSource<InstallationTokenResult> taskCompletionSource =
+        new TaskCompletionSource<>();
+    String oldAuthToken =
+        persistedFid.readPersistedFidEntryValue().getAuthToken() == null
+            ? ""
+            : persistedFid.readPersistedFidEntryValue().getAuthToken();
```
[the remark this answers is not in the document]
ciarand: this doesn't even have to be ints, it could probably just be a boolean ("shouldRefresh"?) that's accessed in the same synchronized block as the listeners additions

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -290,30 +267,106 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+        .isNotEmpty();
+    PersistedFidEntry entryValue = persistedFid.readPersistedFidEntryValue();
+    assertThat(entryValue).hasFid(TEST_FID_1);
+
+    // Waiting for Task that registers FID on the FIS Servers
+    executor.awaitTermination(500, TimeUnit.MILLISECONDS);
+
+    // Validate that registration is complete with updated auth token
+    PersistedFidEntry updatedFidEntry = persistedFid.readPersistedFidEntryValue();
+    assertThat(updatedFidEntry).isEqualTo(UPDATED_AUTH_TOKEN_FID_ENTRY);
```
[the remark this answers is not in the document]
ankitaj224: Updated it to assert on the updated auth token value. Please review to see if it makes sense now. Else  I ll try to be more descriptive.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: PTAL.

ankitaj224: Makes sense. Updated. PTAL. Thanks

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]
ankitaj224: About the file I/O : shouldn't it be okay even if its called by any thread? As long as our onStateReached conditional can handle it correctly. I tested it with my test cases and it seemed to work fine. However, you are right about https://github.com/firebase/firebase-android-sdk/pull/847#discussion_r329233840 .

I tried implementing your suggestion based on my understanding of the comment. Also, I have tested the changes. I will be honest, I dint understand this part 

> ("shouldRefresh"?) that's accessed in the same synchronized block as the listeners additions

please let me know wdyt.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -51,31 +54,30 @@
   private final FirebaseInstallationServiceClient serviceClient;
   private final PersistedFid persistedFid;
   private final ExecutorService executor;
-  private final Clock clock;
   private final Utils utils;
+  private final Object lock = new Object();
 
-  private static final long AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS = 3600L; // 1 hour
-  private static final long AWAIT_TIMEOUT_IN_SECS = 10L;
+  private boolean shouldRefreshAuthToken;
```
ciarand: also needs to be @GuardedBy

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -106,25 +108,28 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
-   * Returns a globally unique identifier of this Firebase app installation.Also, updates the {@link
-   * AwaitListener} when the FID registration is complete.
-   */
-  private Task<String> getId(AwaitListener awaitListener) {
-    return Tasks.call(executor, this::getPersistedFid)
-        .continueWith(orElse(this::createAndPersistNewFid))
-        .onSuccessTask(
-            persistedFidEntry -> registerFidIfNecessary(persistedFidEntry, awaitListener));
+    Task<String> task = addGetIdListener();
+    executor.execute(doRegistration());
```
ciarand: we can make the runnable that doRegistration returns a field that just gets resubmitted; no need to instantiate a new runnable every time if all the state is contained in the parent

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -140,12 +145,12 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
   @Override
   public synchronized Task<InstallationTokenResult> getAuthToken(
```
ciarand: we don't wanna synchronize the getAuthToken method, just the `addGetAuthTokenListener` method. This is also synchronized on `this` which doesn't help us; we need to synchronize on `lock`

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Done. PTAL

ankitaj224:
> tried to demo suggestions in code: https://github.com/firebase/firebase-android-sdk/pull/860/files

Thanks for the demo. I picked up few suggestions from your demo.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -290,30 +258,106 @@ public void testGetId_PersistedFidError_BackendOk() throws InterruptedException
...
+
+    // Validate that registration status is REGISTER_ERROR
+    PersistedFidEntry updatedFidEntry = persistedFid.readPersistedFidEntryValue();
+    assertThat(updatedFidEntry).hasFid(TEST_FID_1);
+    assertThat(updatedFidEntry).hasRegistrationStatus(RegistrationStatus.REGISTER_ERROR);
+  }
+
+  @Test
+  public void testGetId_expiredAuthToken_refreshesAuthToken() throws Exception {
+    when(mockUtils.isAuthTokenExpired(any())).thenReturn(true);
```
ciarand: why does this happen first thing in this test but like 3rd in the previous test?

ankitaj224: No specific reason. Does the order matter?

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -138,34 +131,29 @@ public void setUp() throws FirebaseInstallationServiceException {
...
         .thenThrow(
             new FirebaseInstallationServiceException(
                 "SDK Error", FirebaseInstallationServiceException.Status.SERVER_ERROR));
-    when(persistedFidReturnsError.insertOrUpdatePersistedFidEntry(any())).thenReturn(false);
-    when(persistedFidReturnsError.readPersistedFidEntryValue()).thenReturn(null);
+
     when(mockUtils.createRandomFid()).thenReturn(TEST_FID_1);
-    when(mockClock.currentTimeMillis()).thenReturn(TEST_CREATION_TIMESTAMP_1);
+    when(mockUtils.currentTimeInSecs()).thenReturn(TEST_CREATION_TIMESTAMP_2);
+    when(mockUtils.isAuthTokenExpired(any())).thenReturn(false);
```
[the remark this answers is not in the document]
ankitaj224: Updated where appropriate for tests to be clear.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -451,33 +516,34 @@ public void testGetAuthToken_multipleCallsDoNotForceRefresh_fetchedNewTokenOnce(
...
-                1000,
+                500,
                 (unused) ->
                     InstallationTokenResult.builder()
                         .setToken(TEST_AUTH_TOKEN_4)
                         .setTokenExpirationInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP)
                         .build()))
         .when(backendClientReturnsOk)
         .generateAuthToken(anyString(), anyString(), anyString(), anyString());
+    when(mockUtils.isAuthTokenExpired(any())).thenReturn(false);
```
ankitaj224: I skipped updating ```any()``` to a specific FID entry for ```mockUtils.isAuthTokenExpired()``` only here. This was to hide the details that don't improve the test. Hope that's fine.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,66 +163,101 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
   }
 
-  @NonNull
-  private static <F, T> Continuation<F, T> awaitFidRegistration(
-      @NonNull Supplier<T> supplier, AwaitListener listener) {
-    return t -> {
-      // Waiting for Task that registers FID on the FIS Servers
-      listener.await(AWAIT_TIMEOUT_IN_SECS, TimeUnit.SECONDS);
-      return supplier.get();
+  private final Runnable doRegistration() {
```
ciarand: I wouldn't bother wrapping this in a method like this. Either inline it at the field declaration site (`Runnable doRegistration = () -> { /* code here */ }` or make `doRegistration` the actual runnable and just chuck a method reference on the executor (`execute(this::doRegistration)`).

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,66 +163,101 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
+        PersistedFidEntry persistedFidEntry = persistedFid.readPersistedFidEntryValue();
+
+        // New FID needs to be created
+        if (persistedFidEntry.isErrored() || persistedFidEntry.isNotGenerated()) {
+          String fid = utils.createRandomFid();
+          persistFid(fid);
+          persistedFidEntry = persistedFid.readPersistedFidEntryValue();
+        }
+
+        // Notify the listeners only after force refreshing auth token if shouldRefreshAuthToken is
```
ciarand: trouble is that this means the getId listeners won't be notified until registration happens, which is something we wanted to avoid.

The initial draft version didn't run into this because it called triggerOnStateReached(NOT_REGISTERED) instead of triggerOnStateReached(persistedFidEntry) which meant the listeners were simpler because each step of the state machine always gets called. Could we change it so that it's calling `triggerOnStateReached(newState, persistedFidEntry)` in order to get the step-by-step simplification benefits of the state machine?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,66 +163,101 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
+            needRefresh = shouldRefreshAuthToken;
+          }
+        }
+
+        // Refresh Auth token if needed
+        if (needRefresh) {
+          synchronized (lock) {
+            shouldRefreshAuthToken = false;
+          }
+          fetchAuthTokenFromServer(persistedFidEntry);
```
ciarand: why do the fetch after setting `shouldRefreshAuthToken` to false instead of in reverse?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: shouldRefreshAuthToken will be false by default for getId. Hence, getId listener will be notified before registration happens. (Matches the initial draft version).

It is a matter of doing the checks here vs doing the checks in listeners. 

With ```triggerOnStateReached(newState, persistedFidEntry)``` we have to change the definition of both the listeners (because interface changes). ```GetIdListener``` doesn't need newState.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ciarand: that's fine, I think the rest of the cases have benefited from the change

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -38,6 +42,26 @@
   /** Length of new-format FIDs as introduced in 2019. */
   public static final int FID_LENGTH = 22;
 
+  private static final long AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS = 3600L; // 1 hour
```
ciarand: nit: imo TimeUnit.HOURS.toSeconds(1) is clearer

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -38,6 +42,26 @@
...
+    this.clock = clock;
+  }
+
+  /**
+   * Checks if the FIS Auth token is expired or going to expire in next 1 hour
+   * (AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS).
+   */
+  public boolean isAuthTokenExpired(PersistedFidEntry persistedFidEntry) {
+    return (persistedFidEntry.getTokenCreationEpochInSecs() + persistedFidEntry.getExpiresInSecs()
+        > currentTimeInSecs() + AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS);
```
ciarand: nit: are the outer parens needed?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -38,6 +42,26 @@
...
 
+  private static final long AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS = 3600L; // 1 hour
+
+  Utils(Clock clock) {
+    this.clock = clock;
+  }
+
+  /**
+   * Checks if the FIS Auth token is expired or going to expire in next 1 hour
+   * (AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS).
```
ciarand: nit: use a {@link} tag if you're going to ref symbols in a javadoc

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -159,64 +161,97 @@ public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
...
-  @NonNull
-  private static <F, T> Continuation<F, T> awaitFidRegistration(
-      @NonNull Supplier<T> supplier, AwaitListener listener) {
-    return t -> {
-      // Waiting for Task that registers FID on the FIS Servers
-      listener.await(AWAIT_TIMEOUT_IN_SECS, TimeUnit.SECONDS);
-      return supplier.get();
-    };
-  }
+      // Always notify the GetIdListeners. For GetAuthTokenListeners, only notify if force
```
ciarand: I am concerned that now a getId call won't complete as early as it could if there's a pending force-refresh call. Is that something we should be worried about?

ankitaj224: We need not be worries. getId call will never be called with force-refresh.

ankitaj224: Thanks for clarifying offline. I understand your concern now. I referred to your demo, made some modifications and updated the code. PTAL and let me wdyt.

Thank you.

PR #848, opened 2019-09-25
Fixing FISClient to correctly parse expiration timestamp.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -329,4 +328,15 @@ private String getFingerprintHashForPackage() {
...
+
+  /**
+   * Returns parsed token expiration timestamp in seconds.
+   *
+   * @param expiresIn is expiration timestamp in String format: 604800s
+   */
+  private static long parseTokenExpirationTimestamp(String expiresIn) {
+    return (expiresIn == null || expiresIn.length() == 0)
+        ? 0L
+        : Long.parseLong(expiresIn.substring(0, expiresIn.length() - 1));
```
diwu-arete: I recommend to do a regex check on the string, which pattern is [0-9]+s

ankitaj224: Updated. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -329,4 +332,17 @@ private String getFingerprintHashForPackage() {
...
   }
+
+  /**
+   * Returns parsed token expiration timestamp in seconds.
+   *
+   * @param expiresIn is expiration timestamp in String format: 604800s
+   */
+  private static long parseTokenExpirationTimestamp(String expiresIn) {
+    checkArgument(
+        EXPIRATION_TIMESTAMP_PATTERN.matcher(expiresIn).find(), "Invalid Expiration Timestamp.");
```
diwu-arete: Does find mean finding a sub string?

I guess you expect a exact match? not just find?

PR #849, opened 2019-09-26
feat: add FirebaseFirestoreSettings builder ktx

rosariopfernandes:
Add a Kotlin Extension for building `FirebaseFirestoreSettings` objects.

### DIFF
--- a/docs/ktx/firestore.md
+++ b/docs/ktx/firestore.md
@@ -68,3 +68,27 @@ val objectList = snapshot.toObjects(MyClass::class.java)
 val snapshot: QuerySnapshot = ...
 val objectList = snapshot.toObjects<MyClass>()
 ```
+
+### Setup Firestore with a local emulator
+
+**Kotlin**
+```kotlin
+val settings = FirebaseFirestoreSettings.Builder()
+                   .setHost("10.0.2.2:8080")
+                   .setSslEnabled(false)
+                   .setPersistenceEnabled(false)
+                   .build()
+
+FirebaseFirestore.getInstance().setFirestoreSettings(settings)
+```
+
+**Kotlin + KTX**
+```kotlin
+val settings = firestoreSettings {
+    setHost("10.0.2.2:8080")
+    setSslEnabled(false)
+    setPersistenceEnabled(false)
+}
+
+Firebase.firestore.setFirestoreSettings(settings)
+```
--- a/docs/ktx/firestore.md
+++ b/docs/ktx/firestore.md
@@ -79,16 +79,31 @@ val settings = FirebaseFirestoreSettings.Builder()
                    .setPersistenceEnabled(false)
                    .build()
 
-FirebaseFirestore.getInstance().setFirestoreSettings(settings)
+val firestore = FirebaseFirestore.getInstance()
+firestore.setFirestoreSettings(settings)
+
+val anotherFirestore = FirebaseFirestore.getInstance(app)
+anotherFirestore.setFirestoreSettings(settings)
 ```
 
 **Kotlin + KTX**
 ```kotlin
-val settings = firestoreSettings {
+val firestore = Firebase.firestore {
     setHost("10.0.2.2:8080")
     setSslEnabled(false)
     setPersistenceEnabled(false)
 }
 
-Firebase.firestore.setFirestoreSettings(settings)
+val anotherFirestore = Firebase.firestore(app) {
+    setHost("10.0.2.2:8080")
+    setSslEnabled(false)
+    setPersistenceEnabled(false)
+}
+
+// Alternatively
+Firebase.firestore.setSettings {
+    setHost("10.0.2.2:8080")
+    setSslEnabled(false)
+    setPersistenceEnabled(false)
+}
 ```
--- a/docs/ktx/firestore.md
+++ b/docs/ktx/firestore.md
@@ -89,11 +89,4 @@ firestore.firestoreSettings = firestoreSettings {
     isSslEnabled = false
     isPersistenceEnabled = false
 }
-
-// Alternatively
-Firebase.firestore.setSettings {
-    host = "http://10.0.0.2:8080"
-    isSslEnabled = false
-    isPersistenceEnabled = false
-}
 ```
--- a/docs/ktx/firestore.md
+++ b/docs/ktx/firestore.md
@@ -79,31 +79,21 @@ val settings = FirebaseFirestoreSettings.Builder()
                    .setPersistenceEnabled(false)
                    .build()
 
-val firestore = FirebaseFirestore.getInstance()
 firestore.setFirestoreSettings(settings)
-
-val anotherFirestore = FirebaseFirestore.getInstance(app)
-anotherFirestore.setFirestoreSettings(settings)
 ```
 
 **Kotlin + KTX**
 ```kotlin
-val firestore = Firebase.firestore {
-    setHost("10.0.2.2:8080")
-    setSslEnabled(false)
-    setPersistenceEnabled(false)
-}
-
-val anotherFirestore = Firebase.firestore(app) {
-    setHost("10.0.2.2:8080")
-    setSslEnabled(false)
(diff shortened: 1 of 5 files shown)


docs/ktx/firestore.md
```diff
@@ -67,4 +67,28 @@ val objectList = snapshot.toObjects(MyClass::class.java)
...
+
+**Kotlin + KTX**
+```kotlin
+val settings = firestoreSettings {
+    setHost("10.0.2.2:8080")
+    setSslEnabled(false)
+    setPersistenceEnabled(false)
+}
+
+Firebase.firestore.setFirestoreSettings(settings)
```
vkryachko: I'd suggest two alternatives/additions to consider:

**1. Pass settings when initializing firestore**
Context: it's only possible to set setting on firestore before it is used, so it may be a good idea to make it impossible/hard to do differently.

Namely by passing settings when initializing firestore, that way you only get to set the settings when you're creating it.

```kotlin
val firestore = Firebase.firestore {
  host = "10.0.2.2:8080"
  sslEnabled = false
}
val anotherFirestore = Firebase.firestore(app) {
  host = "10.0.2.2:8080"
  sslEnabled = false
}
```

It might make sense to do the same for the Java SDK, but that's a different story

**2. Make setter an extension to `FirebaseFirestore` to make things more concise**

```kotlin
Firebase.firestore.setSettings {
  host = "10.0.2.2:8080"
  sslEnabled = false
}
```

rosariopfernandes: @vkryachko I really like both additions. I've just added them.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -147,6 +148,33 @@ inline fun <reified T : Any> QuerySnapshot.toObjects(): List<T> = toObjects(T::c
 inline fun <reified T : Any> QuerySnapshot.toObjects(serverTimestampBehavior: DocumentSnapshot.ServerTimestampBehavior): List<T> =
         toObjects(T::class.java, serverTimestampBehavior)
```
rlazo:
```
fun FirebaseFirestore.setSettings(init: FirebaseFirestoreSettings.Builder.() -> Unit) {
    val builder = FirebaseFirestoreSettings.Builder()
    builder.init()
    firestoreSettings = builder.build()
}
```

docs/ktx/firestore.md
```diff
@@ -67,4 +67,43 @@ val objectList = snapshot.toObjects(MyClass::class.java)
...
+    setHost("10.0.2.2:8080")
+    setSslEnabled(false)
+    setPersistenceEnabled(false)
+}
+
+// Alternatively
+Firebase.firestore.setSettings {
+    setHost("10.0.2.2:8080")
+    setSslEnabled(false)
+    setPersistenceEnabled(false)
```
rlazo:
```
Firebase.firestore.firestoreSettings = firestoreSettings {
    host ="10.0.2.2:8080"
    sslEnabled = false
    persistenceEnabled = false
}
```

rosariopfernandes:
@rlazo I've added the missing java getters and changed the extension functions. Please check

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
```diff
@@ -145,6 +145,22 @@ public Builder setCacheSizeBytes(long value) {
       return this;
     }
```
rlazo: `@Nullable`

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
```diff
@@ -145,6 +145,22 @@ public Builder setCacheSizeBytes(long value) {
...
+      return sslEnabled;
+    }
+
+    public boolean isPersistenceEnabled() {
+      return persistenceEnabled;
+    }
+
+    public long getCacheSizeBytes() {
+      return cacheSizeBytes;
+    }
```
rlazo: Please, add comments for this methods, specially the default values (because they return primitive types, it cannot be null).

docs/ktx/firestore.md
```diff
@@ -79,31 +79,21 @@ val settings = FirebaseFirestoreSettings.Builder()
...
-    setSslEnabled(false)
-    setPersistenceEnabled(false)
+firestore.firestoreSettings = firestoreSettings {
+    host = "http://10.0.0.2:8080"
+    isSslEnabled = false
+    isPersistenceEnabled = false
 }
 
 // Alternatively
 Firebase.firestore.setSettings {
```
rlazo: Remove the alternative. Enabling the assignment `=` as the "preferred" way makes it a bit clearer that we are using the builder pattern, and that have a getter and a setter in java for `firestoreSettings`.

rosariopfernandes: @rlazo should I remove it entirely or only remove from the docs?

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
```diff
@@ -145,6 +145,30 @@ public Builder setCacheSizeBytes(long value) {
       return this;
     }
 
+    /** @return the host of the Cloud Firestore backend. */
+    @Nullable
+    public String getHost() {
+      return host;
+    }
+
+    /** @return boolean indicating whether SSL is enabled or not. The default is to use SSL. */
```
rlazo:
```suggestion
    /** @return boolean indicating whether SSL is enabled or not. */
```

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
```diff
@@ -145,6 +145,30 @@ public Builder setCacheSizeBytes(long value) {
...
+    public boolean isSslEnabled() {
+      return sslEnabled;
+    }
+
+    /**
+     * @return boolean indicating whether local persistent storage is enabled or not.
+     * The default is to use local persistent storage.
```
rlazo:
```suggestion
```

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
```diff
@@ -145,6 +145,30 @@ public Builder setCacheSizeBytes(long value) {
...
+     */
+    public boolean isPersistenceEnabled() {
+      return persistenceEnabled;
+    }
+
+    /** @return cache size for on-disk data. The default is 100 MB */
```
rlazo:
```suggestion
    /** @return cache size for on-disk data. */
```

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
rlazo: I'm sorry for the unnecessary back-and-forth, I missed my early cup of copy and it shows. As there's always a valid value returned, we can get rid of the comment with the default value. I've suggested the changes to simplify this.

docs/ktx/firestore.md
rlazo: Please, do remove it entirely.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
rosariopfernandes: @rlazo No problems! :) All done.

rlazo:
Thanks a lot for being so accommodating @rosariopfernandes And once again, thanks for contributing to Firebase :tada:

PR #853, opened 2019-09-26
Added unit test

davidmotson:


firebase-dynamic-links/firebase-dynamic-links.gradle
```diff
@@ -17,7 +17,7 @@ plugins {
 }
 
 firebaseLibrary {
-    testLab.enabled = false
+    testLab.enabled = true
```
vkryachko: This should still be `false` as there are no instrumentation tests here afaict

firebase-dynamic-links/src/test/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
...
+          package="com.google.firebase.dynamiclink.test"
+          android:versionCode="1"
+          android:versionName="1">
+
+    <uses-sdk android:minSdkVersion="16"/>
+
+    <application>
+        <service android:name="com.google.firebase.components.ComponentDiscoveryService">
+            <meta-data
+               android:name="com.google.firebase.components:com.google.firebase.analytics.connector.FakeConnectorComponent"
```
vkryachko: I can't see this class in your PR, is it needed?
If not, you can delete this manifest, no?

PR #854, opened 2019-09-27
Upload AAR size when uploading apk size too.

rlazo:


tools/measurement/apksize/src/transport/java/com.google.apksize/TransportSample.java
```diff
@@ -29,7 +28,8 @@ public void runSample(Context context) {
 
     TransportRuntime.initialize(context);
 
-    TransportFactory f = TransportRuntime.getInstance().newFactory(LegacyFlgDestination.DEFAULT_INSTANCE);
```
allisonbm92: Is this related to the AAR size collection?

PR #856, opened 2019-09-27
Propagating the exceptions to the clients.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/GetAuthTokenListener.java
```diff
@@ -38,12 +38,13 @@ public boolean onStateReached(PersistedFidEntry persistedFidEntry) {
               .build());
       return true;
     }
+    return false;
+  }
 
+  @Override
+  public boolean onException(PersistedFidEntry persistedFidEntry, Exception exception) {
     if (persistedFidEntry.isErrored()) {
```
ciarand: when would it be the case that we'd get an exception without the persistedFidEntry being in the error state?

firebase-installations/src/main/java/com/google/firebase/installations/GetIdListener.java
```diff
@@ -27,15 +27,16 @@ public GetIdListener(TaskCompletionSource<String> taskCompletionSource) {
...
+  }
 
+  @Override
+  public boolean onException(PersistedFidEntry persistedFidEntry, Exception exception) {
     if (persistedFidEntry.isErrored()) {
-      taskCompletionSource.setException(
-          new FirebaseInstallationsException(
-              "Failed to update client side cache.",
-              FirebaseInstallationsException.Status.CLIENT_ERROR));
+      taskCompletionSource.trySetException(exception);
```
ciarand: given this is the same in each of the listener impls so far we could move it into a default interface method

firebase-installations/src/main/java/com/google/firebase/installations/GetAuthTokenListener.java
ankitaj224: Interrupted Exception 
Exception thrown while persisting. 
Thread failure

firebase-installations/src/main/java/com/google/firebase/installations/GetIdListener.java
ankitaj224: error: [NoStaticOrDefaultMethodsInInterfaces] Avoid static/default methods in interfaces: We currently desugar SDKs with retrolambda which has limited support for those.

Default methods require API level 24. Current min is 14

ciarand: oh, sad! I thought we had moved past retrolambda

firebase-installations/src/main/java/com/google/firebase/installations/StateListener.java
```diff
@@ -18,4 +18,6 @@
 
 interface StateListener {
   boolean onStateReached(PersistedFidEntry persistedFidEntry);
+
+  boolean onException(PersistedFidEntry persistedFidEntry, Exception exception);
```
ciarand: might be good to document the meaning behind the boolean return value

ankitaj224: Updated. Thanks.

PR #857, opened 2019-09-27
Add support for `HIGHEST` priority to transport Events.

vkryachko:


transport/transport-api/src/main/java/com/google/android/datatransport/Priority.java
```diff
@@ -14,10 +14,41 @@
...
   /** Quality of the service is within an hour. */
-  DEFAULT,
+  DEFAULT(0),
+
+  /** Events delivered at most daily, on unmetered network. */
+  VERY_LOW(1),
+
+  /** Events delivered as soon as possible. */
+  HIGHEST(2),
+  ;
```
rlazo: remove the `;`

transport/transport-api/src/main/java/com/google/android/datatransport/Priority.java
```diff
@@ -14,10 +14,41 @@
...
+  }
+
+  public int getValue() {
+    return value;
+  }
 
-  /** Events delivered at most daily, on unmetered network */
-  VERY_LOW,
+  public static Priority forValue(int value) {
+    if (value < 0 || value >= PRIORITY_MAP.length) {
```
rlazo: For this to work, values would have to increment monotonically.  Either comment that in the `enum`, or use check the last value of the array as max.

transport/transport-api/src/main/java/com/google/android/datatransport/Priority.java
vkryachko: It's needed syntactically actually, note the comma one line above. I added it to make these lines friendlier to future diffs

rlazo: ack

transport/transport-api/src/main/java/com/google/android/datatransport/Priority.java
vkryachko: I fixed the underlying issue that you pointed out. Note that I also removed the int -> priority logic into transport-runtime.

PR #862, opened 2019-09-28
Implement firebase-database coroutine extension

msasikanth:
Doesn't support generics like List<> and Map<> unless suspended function is inlined and type is reified

### DIFF
--- a/firebase-database/ktx/ktx.gradle
+++ b/firebase-database/ktx/ktx.gradle
@@ -49,6 +49,7 @@ dependencies {
     implementation project(':firebase-database')
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'com.google.android.gms:play-services-tasks:17.0.0'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.1'
 
     androidTestImplementation 'junit:junit:4.12'
     androidTestImplementation "com.google.truth:truth:$googleTruthVersion"
--- a/firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
+++ b/firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
@@ -16,14 +16,19 @@ package com.google.firebase.database.ktx
 
 import androidx.annotation.Keep
 import com.google.firebase.FirebaseApp
-import com.google.firebase.database.DataSnapshot
-import com.google.firebase.database.GenericTypeIndicator
-import com.google.firebase.database.FirebaseDatabase
 import com.google.firebase.components.Component
 import com.google.firebase.components.ComponentRegistrar
-
+import com.google.firebase.database.DataSnapshot
+import com.google.firebase.database.DatabaseError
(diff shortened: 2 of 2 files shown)


rosariopfernandes:
@msasikanth any specific reason why we can't inline the function and use reified type? It would be nice if it could support generics.

msasikanth:
> @msasikanth any specific reason why we can't inline the function and use reified type? It would be nice if it could support generics.

Inline calls like this may increase byte code at call site. But I maybe wrong, if it's fine I can make this function inline

rlazo:
Thanks a lot for your contribution @msasikanth ! I'll get back to you after the changes have been reviewed internally.

msasikanth:
> Thanks a lot for your contribution @msasikanth ! I'll get back to you after the changes have been reviewed internally.

Awesome. One test failed because of ':firebase-database:ktx:generateLicenses', what this supposed to happen?

rlazo:
> > Thanks a lot for your contribution @msasikanth ! I'll get back to you after the changes have been reviewed internally.
> 
> Awesome. One test failed because of ':firebase-database:ktx:generateLicenses', was this supposed to happen?

It's most likely a temporal issue. I've re-run the test to confirm.

rlazo:
> > > Thanks a lot for your contribution @msasikanth ! I'll get back to you after the changes have been reviewed internally.
> > 
> > 
> > Awesome. One test failed because of ':firebase-database:ktx:generateLicenses', was this supposed to happen?
> 
> It's most likely a temporal issue. I've re-run the test to confirm.

This is actually caused by a bug in our license fetcher. I'll fix it in a separate PR.

rlazo:
> > > > Thanks a lot for your contribution @msasikanth ! I'll get back to you after the changes have been reviewed internally.
> > > 
> > > 
> > > Awesome. One test failed because of ':firebase-database:ktx:generateLicenses', was this supposed to happen?
> > 
> > 
> > It's most likely a temporal issue. I've re-run the test to confirm.
> 
> This is actually caused by a bug in our license fetcher. I'll fix it in a separate PR.

https://github.com/firebase/firebase-android-sdk/pull/868 addresses the issue.

msasikanth:
> > > > > Thanks a lot for your contribution @msasikanth ! I'll get back to you after the changes have been reviewed internally.
> > > > 
> > > > 
> > > > Awesome. One test failed because of ':firebase-database:ktx:generateLicenses', was this supposed to happen?
> > > 
> > > 
> > > It's most likely a temporal issue. I've re-run the test to confirm.
> > 
> > 
> > This is actually caused by a bug in our license fetcher. I'll fix it in a separate PR.
> 
> #868 addresses the issue.

I accidentally forgot and pushed the latest changes from firebase/firebase-android-sdk. Should I open another PR with just the coroutines commit?

rlazo:
Thanks for your contribution @msasikanth! We've reviewed it internally and while we think it provides ease of use, it differs from the general approach we are using in firebase-database. We are evaluating what's the best approach to use throughout all SDKs, and we'd like to keep the API consistent in the meanwhile. Once again, thanks for your contribution to Firebase!

PR #863, opened 2019-09-30
Add log source name support to cct backend.

vkryachko:


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -194,6 +193,14 @@ private BatchedLogRequest getRequestBody(BackendRequest backendRequest) {
                               .setFingerprint(firstEvent.get(KEY_FINGERPRINT))
                               .build())
                       .build());
+
+      // set log source to either its numeric value or its name.
+      try {
+        requestBuilder.setLogSource(Integer.valueOf(entry.getKey()));
+      } catch (NumberFormatException ex) {
+        requestBuilder.setLogSourceName(entry.getKey());
```
rlazo: Would it be useful to log this case?

vkryachko: Not sure, but I am leaning towards "not useful", it's not an exceptional case but rather an implementation detail that allows us to support both ints and names. wdyt?

rlazo: LGTM

PR #867, opened 2019-10-01
Migrate org.mockito.Matchers to org.mockito.ArgumentMatchers

TimvdLippe:
The former is deprecated in favor of the latter. This matches the
internal change cl/272200869


firebase-inappmessaging/firebase-inappmessaging.gradle
```diff
@@ -135,7 +135,7 @@ dependencies {
     annotationProcessor 'com.google.auto.value:auto-value:1.6.2'
     annotationProcessor 'com.ryanharter.auto.value:auto-value-parcel:0.2.6'
 
-    testImplementation 'org.mockito:mockito-core:1.10.19'
```
TimvdLippe: This was the only project using 1.10.19 while all the others (including the `androidTestImplementation` was using 2.25.0)

PR #868, opened 2019-10-01
Add support for multiple URLs in RemoteLicenseFetcher

rlazo:
I've modified the fetchers to include a `LicenseName` for debugging purposes, instead of using the URL, because the mechanism now tries to download the license from any of the available sources.


rlazo:
@vkryachko PTAL

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/LicenseResolverPlugin.groovy
```diff
@@ -52,7 +51,6 @@ class LicenseResolverPlugin implements Plugin<Project> {
             [new AndroidSdkTermsFetcher(),
              new Apache2LicenseFetcher(),
              new AnotherApache2LicenseFetcher(),
```
vkryachko: What about this one? should we remove it too?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
```diff
@@ -44,14 +50,15 @@ abstract class RemoteLicenseFetcher implements Serializable {
...
+    List<String> urls = remoteUrls
 
     for (int i = 0; i < 3; ++i) {
       try {
         if (i > 0) {
           Thread.sleep(i * 1000)
         }
 
-        return getTextAttempt()
+        return getTextAttempt(urls[i % urls.size()])
```
vkryachko: hm, looks like there is no guarantee that the license will be fetched from the url that is specified in the pom, is there? If not, this may be a problem

@ashwinraghav wdyt?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/LicenseResolverPlugin.groovy
rlazo: It comes from a site that requires a different method to extract the license text, so the fetchers cannot be merged together.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/license/RemoteLicenseFetcher.groovy
rlazo: True. If this is a real concern, a middle ground could be make fetchers deal with either `http` or `https` as equals, which will reduce somewhat the amount of fetchers without actual merging different sources.

PR #869, opened 2019-10-01
Index-Free: Drop last limbo free snapshot version for schema v9

schmidt-sebastian:
This PR re-adds the asserts we reverted out in v21.10.1 and drops the last limbo free Snapshot Version for any query data that has been written with v21.10.1.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
@@ -49,7 +49,7 @@ class SQLiteSchema {
    * The version of the schema. Increase this by one for each migration added to runMigrations
    * below.
    */
-  static final int VERSION = 9;
+  static final int VERSION = 10;
 
   // Remove this constant and increment VERSION to enable indexing support
   static final int INDEXING_SUPPORT_VERSION = VERSION + 1;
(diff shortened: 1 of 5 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -131,13 +131,16 @@ void runMigrations(int fromVersion, int toVersion) {
       createV8CollectionParentsIndex();
     }
 
-    if (fromVersion < 9 && toVersion >= 9) {
+    if ((fromVersion < 9 && toVersion >= 9) || (fromVersion < 10 && toVersion >= 10)) {
```
wilhuff: This seems off.

If a user is currently running 8 or below, `fromVersion` will satisfy both branches. If someone's on 9, then just `fromVersion < 10` will match. Also, now that we're shipping schema 10, toVersion will never be 9.

This seems to imply that it's sufficient to just make this `if (fromVersion < 10 && toVersion >= 10)` and add a comment above that indicates schema version 9 was shipped, dropped and replayed as 10.

If you really want to be able to run this once as version 9 and again as 10 you could make this `if (fromVersion < 10 && toVersion >= 9)`.

schmidt-sebastian: I wrote this code so that it runs only once f a user migrates from an older version either version 9 or 10. "if (fromVersion < 10 && toVersion >= 9)" is actually slightly different as it also drops the last limbo free snapshot version if a user migrates from version 9 to version 9. 
It would still be cleaner if we just dropped support for schema version 9, but that would make it harder to write a unit test for this new code path.

For now, I cleaned this up by special-casing the migration from 9 to 10, which should make this code a little easier to follow.

PR #870, opened 2019-10-01
feat: add builder ktx for FirebaseRemoteConfigSettings

rosariopfernandes:
Add Kotlin Extensions:
1. Building `FirebaseRemoteConfigSettings` objects 
```kotlin
val configSettings = remoteConfigSettings {
    setMinimumFetchIntervalInSeconds(3600)
    setFetchTimeoutInSeconds(60)
}
```

2. Setting Config Settings on a FirebaseRemoteConfig object
```kotlin
remoteConfig.setConfigSettingsAsync {
    setMinimumFetchIntervalInSeconds(3600)
    setFetchTimeoutInSeconds(60)
}
```

### DIFF
--- a/firebase-config/ktx/ktx.gradle
+++ b/firebase-config/ktx/ktx.gradle
@@ -49,6 +49,7 @@ dependencies {
     implementation project(':firebase-config')
     implementation project (':firebase-abt')
     implementation 'androidx.annotation:annotation:1.1.0'
+    implementation 'com.google.android.gms:play-services-tasks:17.0.0'
 
     testImplementation "org.robolectric:robolectric:$robolectricVersion"
     testImplementation 'junit:junit:4.12'
--- a/firebase-config/ktx/ktx.gradle
+++ b/firebase-config/ktx/ktx.gradle
@@ -49,7 +49,6 @@ dependencies {
     implementation project(':firebase-config')
     implementation project (':firebase-abt')
     implementation 'androidx.annotation:annotation:1.1.0'
-    implementation 'com.google.android.gms:play-services-tasks:17.0.0'
 
     testImplementation "org.robolectric:robolectric:$robolectricVersion"
     testImplementation 'junit:junit:4.12'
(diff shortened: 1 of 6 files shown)


rlazo:
Thanks for your contribution @rosariopfernandes! I like the proposal to extend the `FirebaseRemoteConfigSettings.Builder`. We could also add the missing getters in Java to use properties. 

I'm not totally sold on the second proposal. I think we could get most of the gains from just implementing the first part.

```
remoteConfig.setConfigSettingsAsync(remoteConfigSettings { 
    minimumFetchIntervalInSeconds = 3600
    fetchTimeoutInSeconds = 60
})
```

rosariopfernandes:
Thanks for the review @rlazo . I've removed the second proposal and added the missing Java getters.

firebase-config/ktx/ktx.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     implementation project(':firebase-config')
     implementation project (':firebase-abt')
     implementation 'androidx.annotation:annotation:1.1.0'
+    implementation 'com.google.android.gms:play-services-tasks:17.0.0'
```
rlazo: Is this dependency needed? The only new import is in `RemoteConfig.kt` and doesn't seem related to that library.

rosariopfernandes: It was being used for the second proposal. I forgot to remove the dependency when I removed the function. But it's done now, thanks

rlazo:
Hi @rosariopfernandes, thanks a lot for your contribution to Firebase! We've reviewed your proposal internally and it's been approved as-is. It will be merged into `master` as soon as the pending checks pass.

rosariopfernandes:
@rlazo Thanks!

I wonder if the changes from #876 are somehow affecting the `new-smoke-tests` job here.

rlazo:
> @rlazo Thanks!
> 
> I wonder if the changes from #876 are somehow affecting the `new-smoke-tests` job here.

I've seen that error before, so I'm not sure whether it's related. We're looking into it 😄

rlazo:
As the failure isn't cause by the PR, we're merging it. Thanks again @rosariopfernandes !

PR #879, opened 2019-10-04
Remote Config Fetch event telemetry prototype

jacquetd:
Implements Fetch client-side logging for Remote Config.

Make use of `Firebase-datatransport` as a Component in the Android SDK to inject a Transport object that handles sending of logs to Firebase servers.
Log hasn't been fully configured yet, so a placeholder value is used for TRANSPORT_FINAL/transportName.


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -71,9 +77,11 @@
  * @author Lucas Png
  */
 public class ConfigFetchHttpClient {
+
   private static final String API_KEY_HEADER = "X-Goog-Api-Key";
   private static final String ETAG_HEADER = "ETag";
   private static final String IF_NONE_MATCH_HEADER = "If-None-Match";
+  private static final String TRANSPORT_FINAL = "-1"; //(TODO) Replace with actual logSource int
```
erikeldridge: Is this a header? If so, add the "_HEADER" suffix for consistency. If not, move the declaration down and keep the headers grouped together

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -86,30 +94,42 @@
...
+      long readTimeoutInSeconds,
+      TransportFactory transportFactory) {
     this.context = context;
     this.appId = appId;
     this.apiKey = apiKey;
     this.projectNumber = extractProjectNumberFromAppId(appId);
     this.namespace = namespace;
     this.connectTimeoutInSeconds = connectTimeoutInSeconds;
     this.readTimeoutInSeconds = readTimeoutInSeconds;
+    this.transport =
```
erikeldridge: Since transport instantiation is orthogonal to fetch logic, can we inject a `Transport` instance, rather than injecting the factory?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -207,9 +230,30 @@ FetchResponse fetch(
...
+
+    long endTime = System.currentTimeMillis();
+    ClientLogEvent clientLogEvent = generateClientLogEvent(instanceId, startTime, endTime);
+    transport.send(Event.ofTelemetry(clientLogEvent));
+
     return FetchResponse.forBackendUpdatesFetched(fetchedConfigs, fetchResponseETag);
   }
 
+  @VisibleForTesting
+  public ClientLogEvent generateClientLogEvent(String instanceId, long startTime, long endTime) {
```
erikeldridge: Following on the ideas of differentiating `Transport` config and injecting the `Transport` instance, I think we could encapsulate logging logic in a class and inject an instance of that class, eg
```
class RemoteConfigLogger {
  public RemoteConfigLogger (transport) { ... }

  // A static creation method that initializes a RemoteConfigLogger with a Transport instance might be helpful too

  logFetchEvent(latency){ ... }
}
...
class ConfigFetchHttpClient {
  public ConfigFetchHttpClient ( ... remoteConfigLogger) { ... }
  ...
  FetchResponse fetch(...) {
    ... remoteConfigLogger.logFetchEvent(endTime - startTime)
  }
```

As a side benefit, encapsulating log logic in a class give us one place to look when debugging logging logic, comparable to https://github.com/firebase/firebase-js-sdk/blob/master/packages/remote-config/src/errors.ts

Related, here's a (copy of a) previous project that used the same pattern, and also enabled external devs to subscribe to the same events, which we might consider if folks want to use their own analytics: https://github.com/teju/Example_Code/blob/master/digits-android-master/digits/src/main/java/com/digits/sdk/android/DigitsEventCollector.java

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
jacquetd: It is not a header. I'll move it!

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
jacquetd: I've decided to inject a ConfigLogger class instead that uses the Transport instance.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
jacquetd: I've followed your advice and moved the logic into another class called ConfigLogger that takes in an instance of Transport, since it makes it a lot cleaner in terms of tests as well.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -86,21 +87,25 @@
   private final long connectTimeoutInSeconds;
   private final long readTimeoutInSeconds;
 
+  private ConfigLogger configLogger;
```
clp93: can this be `private final ConfigLogger`?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -207,6 +215,15 @@ FetchResponse fetch(
     }
 
     ConfigContainer fetchedConfigs = extractConfigs(fetchResponse, currentTime);
+
+    long endTime = System.currentTimeMillis();
+    configLogger.logFetchEvent(
+        appId,
+        namespace,
+        instanceId,
+        /* timestampMillis = */ System.currentTimeMillis(),
```
clp93: `endTime`?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,63 @@
...
+package com.google.firebase.remoteconfig.internal;
+
+import com.google.android.datatransport.Event;
+import com.google.android.datatransport.Transport;
+import com.google.firebase.remoteconfig.BuildConfig;
+import com.google.firebase.remoteconfig.proto.ClientMetrics.ClientLogEvent;
+import com.google.firebase.remoteconfig.proto.ClientMetrics.ClientLogEvent.EventType;
+import com.google.firebase.remoteconfig.proto.ClientMetrics.FetchEvent;
+
+public class ConfigLogger {
```
clp93: javadocs in this class please. Thanks!

e.g.

/**  Lightweight client for logging client side metrics for Firebase Remote Config. */

clp93: does this have to be public? Try to reduce visibility as much as possible. Thanks!

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private final Transport<ClientLogEvent> transport;
+
+  public ConfigLogger(Transport<ClientLogEvent> transport) {
```
clp93: ditto

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,63 @@
...
+    this.transport = transport;
+  }
+
+  public void logFetchEvent(
```
clp93: ditto

firebase-config/src/proto/com/google/android/gms/config/proto/client_metrics.proto
```diff
@@ -0,0 +1,64 @@
...
+
+package com.google.firebase.remoteconfig.proto;
+
+option java_package = "com.google.firebase.remoteconfig.proto";
+option java_outer_classname = "ClientMetrics";
+
+// Single unit of metrics data collected from Firebase integrated apps using the
+// Remote Config SDK.
+// Every clientside metrics related event that is logged will correspond to
+// this object.
```
clp93: one of these enums.

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -145,15 +129,14 @@ protected RemoteConfigComponent(
...
     this.firebaseAbt = firebaseAbt;
-    this.transportFactory = transportFactory;
 
     Transport<ClientLogEvent> transport =
         transportFactory.getTransport(
-            TRANSPORT_FINAL,
-            ClientLogEvent.class,
-            ClientLogEvent::toByteArray);
-
+            /* name = */ TRANSPORT_FINAL,
```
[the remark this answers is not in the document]
clp93: please find and replace throughout. Thanks!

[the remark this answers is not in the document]
jacquetd: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
jacquetd: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
jacquetd: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
jacquetd: Done.

jacquetd: Looks like I'd have to keep it public unless I move ConfigLogger out of the /remoteconfig/internal folder into the parent /remoteconfig folder since I instantiate a ConfigLogger object in the RemoteConfigComponent class. What do you think about keeping it public vs. moving the code to its parent folder?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
jacquetd: Same issue as above.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
jacquetd: Modified to exclude public.

firebase-config/src/proto/com/google/android/gms/config/proto/client_metrics.proto
jacquetd: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
clp93: Sounds good, thats fine then! anything in /internal is hidden (i.e., not visible to developers) anyways

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -29,12 +34,12 @@ public ConfigLogger(Transport<ClientLogEvent> transport) {
     this.transport = transport;
   }
 
-  public void logFetchEvent(
+  void logFetchEvent(
```
clp93: javadoc?

jacquetd: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -34,6 +34,10 @@ public ConfigLogger(Transport<ClientLogEvent> transport) {
     this.transport = transport;
   }
 
+  /**
+   * Creates a {@link ClientLogEvent} with Fetch-specific metrics {@link FetchEvent}
+   * and sends the Event via {@link Transport} to log to the backend.
```
clp93: How about "Create and log a {@link ClientLogEvent} that contains metrics related to a {@link FetchEvent}."?

jacquetd: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -207,6 +215,15 @@ FetchResponse fetch(
     }
 
     ConfigContainer fetchedConfigs = extractConfigs(fetchResponse, currentTime);
+
+    long endTime = System.currentTimeMillis();
+    configLogger.logFetchEvent(
+        appId,
+        namespace,
+        instanceId,
+        /* timestampMillis= */ endTime,
```
erikeldridge: Minor: the event timestamp is specific to log logic, so I think it'd make more sense encapsulated in `ConfigLogger`. In other words, we happen to have a timestamp in this case, so it isn't tedious here, but it'd be tedious to require a timestamp every time we log an event.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,79 @@
...
+   * Creates and log a {@link ClientLogEvent} that contains metrics related to a {@link FetchEvent}
+   */
+  void logFetchEvent(
+      String appId, String namespace, String fid, long timestampMillis, long networkLatencyMillis) {
+
+    ClientLogEvent.Builder clientLogEventBuilder =
+        createClientLogEventBuilder(
+            appId, namespace, fid, timestampMillis, /* sdkVersion= */ BuildConfig.VERSION_NAME);
+
+    ClientLogEvent fetchEvent = createFetchEvent(clientLogEventBuilder, networkLatencyMillis);
```
erikeldridge: Minor: mutating `clientLogEventBuilder` in a method called `createFetchEvent` is hard to follow, and given `createFetchEvent` just appends a couple fields, I'd argue it doesn't add enough value to justify the indirection, esp since we have a first-class builder that provides convenient setters.

I think something like this might be easier to follow:
```
void logFetchEvent(...){
  ClientLogEvent clientLogEvent = createClientLogEventBuilder(...).
    .setEventType(EventType.FETCH)
    .setFetchEvent(FetchEvent.newBuilder().setNetworkLatencyMillis(networkLatencyMillis).build())
        .build();
 
  transport.send(Event.ofData(fetchEvent));
}
```

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,79 @@
...
+   * metrics.
+   */
+  private ClientLogEvent.Builder createClientLogEventBuilder(
+      String appId, String namespace, String fid, long timestampMillis, String sdkVersion) {
+
+    return ClientLogEvent.newBuilder()
+        .setAppId(appId)
+        .setNamespaceId(namespace)
+        .setFid(fid)
+        .setTimestampMillis(timestampMillis)
```
erikeldridge: Minor: I think we could improve convenience by generating the timestamp here

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Returns a {@link ClientLogEvent.Builder} that instantiates general fields for client-side
+   * metrics.
+   */
+  private ClientLogEvent.Builder createClientLogEventBuilder(
+      String appId, String namespace, String fid, long timestampMillis, String sdkVersion) {
+
+    return ClientLogEvent.newBuilder()
+        .setAppId(appId)
```
erikeldridge: Minor: given `appId`, `namespace` and `sdkVersion` are constant, I think we could improve convenience making them instance fields.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigLoggerTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @Before
+  public void setup() {
+    initMocks(this);
+    configLogger = new ConfigLogger(mockTransport);
+  }
+
+  @Test
+  public void fetch_newValues_logsAppId() {
+    configLogger.logFetchEvent(
```
erikeldridge: Minor: it's a bit hard to follow these are all testing a single field in the output of the same method, as opposed to a single test, eg 
```
  @Test
  public void logFetchEvent_populatesEventCorrectly() {
    configLogger.logFetchEvent(
        FAKE_APP_ID,
        DEFAULT_NAMESPACE,
        FAKE_FID,
        /* timestampMillis= */ 100,
        /* networkLatencyMillis= */ 10);

    verify(mockTransport).send(logEventCaptor.capture());
   
    ClientLogEvent clientLogEvent = logEventCaptor.getValue().getPayload();

    assertThat(clientLogEvent.getAppId()).isEqualTo(FAKE_APP_ID);
    assertThat(clientLogEvent.getNamespaceId()).isEqualTo(DEFAULT_NAMESPACE);
    assertThat(clientLogEvent.getFid()).isEqualTo(FAKE_FID);
    ...
  }
```

It might make sense to define a separate test for `createEventLogBuilder`, esp if we populate the builder from instance fields and/or generate the timestamp in that method, but I'd guess the value of doing this will become more clear as we add events and it become tedious to assert common values.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigLoggerTest.java
```diff
@@ -0,0 +1,148 @@
...
+  private ConfigLogger configLogger;
+
+  @Before
+  public void setup() {
+    initMocks(this);
+    configLogger = new ConfigLogger(mockTransport);
+  }
+
+  @Test
+  public void fetch_newValues_logsAppId() {
```
erikeldridge: I think `logFetchEvent_...` would be more appropriate since it's testing a method called `logFetchEvent`.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
jacquetd: Fixed.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
jacquetd: I wasn't quite sure how to proceed with this so thank you for the advice! I adjusted it according to your example.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
jacquetd: Fixed and adjusted tests accordingly.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
jacquetd: I've injected `FirebaseInstanceId` into the `ConfigLogger` constructor and retrieve the ID every time `logFetchEvent` is called, as well as inject `appId` and `namespace` into the constructor as well.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigLoggerTest.java
jacquetd: I've decided to compile the tests that check if the general `ClientLogEvent` fields are sent properly in `logFetchEvent`, and another test to validate `FetchLatency` specific fields.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigLoggerTest.java
jacquetd: Fixed.

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -119,10 +129,17 @@ protected RemoteConfigComponent(
...
     this.firebaseInstanceId = firebaseInstanceId;
     this.firebaseAbt = firebaseAbt;
+
     this.analyticsConnector = analyticsConnector;
 
     this.appId = firebaseApp.getOptions().getApplicationId();
 
+    this.transport =
+        transportFactory.getTransport(
+            /* name = */ TRANSPORT_FINAL,
```
clp93: nit: name= */

jacquetd: Done.

PR #883, opened 2019-10-07
Disabling javadoc on common ktx

ashwinraghav:
Seems to preserve the -javadoc.jar


vkryachko:
iiuc, with this flag disabled, the javadoc.jar produced should be empty, can you verify that it is not the case?

PR #888, opened 2019-10-08
Extract FID from FIS createInstallation response

ankitaj224:


firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -117,6 +118,16 @@
           .setRegistrationStatus(PersistedFid.RegistrationStatus.UNREGISTERED)
           .build();
 
+  private static final PersistedFidEntry INVALID_FID_ENTRY =
```
andirayo: I have the feeling we talked about this multiple times already and thus please feel free to ignore:
Why is this not called PersistedInstallationEntry or something.
Since the data stored identifies the Firebase Installation as a whole, i.e. contains the Installation's FID, registration-status, it's Refresh-Token and latest Auth-Token...

Please ignore if we have discussed this already in the past.
I'm also happy to talk in person if you prefer.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -117,6 +118,16 @@
...
+      PersistedFidEntry.builder()
+          .setFirebaseInstallationId(INVALID_TEST_FID)
+          .setAuthToken("")
+          .setRefreshToken("")
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_1)
+          .setExpiresInSecs(0)
```
andirayo: Assuming that this is an object (or a proto), can we create a class (or proto) for Auth-Token that has the fields "token", "creationTimestamp", "expirationTimestamp" or "expiresIn", so that the fields that belong together are part of the same object?

firebase-installations/api.txt
```diff
@@ -94,6 +94,7 @@ package com.google.firebase.installations.remote {
     ctor public InstallationResponse();
     method @NonNull public static com.google.firebase.installations.remote.InstallationResponse.Builder builder();
     method @NonNull public abstract InstallationTokenResult getAuthToken();
+    method @NonNull public abstract String getFid();
     method @NonNull public abstract String getName();
```
andirayo: I recommend calling this "uri" instead of "name" since that is a more precise description than "name".
On the server-side we have to call this field "name" due to API design restrictions, but the value of the field is the unique resource identifier (URI) of this Firebase Installation.

ankitaj224: Done. Thanks.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ankitaj224: Yes, we have. I acknowledged the comments and changed to PersistedFidEntry. I can change it to 'PersistedInstallationEntry' once more as per your suggestion. But will do that as part of another PR to contain the scope of the current PR.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
ankitaj224: I can change but IMO dont see much benefit. This object is used to represent persisted fid & metadata - simpler the class, easier it is to read and write. 

Objects that are exposed to users have been separated out into a class (or proto) for Auth-Token & a class for firebase installation. This addresses your comment & is useful to segregate the details to the clients.

PR #889, opened 2019-10-09
Use auto-generated apks for size measurement.

rlazo:
By using gradle to generate the dependencies instead of a manual approach, the maintenance cost goes down significantly.


rlazo:
Here's a sample output https://gist.github.com/rlazo/4d96a574bd5c41e769eaa8c08975c7cf

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/apksize/GenerateVariantPlugin.groovy
```diff
@@ -0,0 +1,47 @@
...
+            firebaseProjectsNames.each {
+                sourceSets.getByName(it) { sourceSet ->
+                    sourceSet.java.srcDirs += ['src/empty/java']
+                }
+            }
+        }
+        firebaseProjectsNames.each {
+            project.dependencies.add("${it}Implementation", project.rootProject.project(":${it}"))
+        }
+    }
```
allisonbm92: I think it would be clearer to have one loop containing the blocks instead of placing multiple loops within the configuration blocks. What do you think?

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/apksize/GenerateVariantPlugin.groovy
```diff
@@ -0,0 +1,47 @@
...
+            firebaseProjectsNames.each {
+                productFlavors.create("$it") {
+                    dimension "apkSize"
+                }
+            }
+            firebaseProjectsNames.each {
+                sourceSets.getByName(it) { sourceSet ->
+                    sourceSet.java.srcDirs += ['src/empty/java']
+                }
+            }
```
allisonbm92: Is this needed? Empty doesn't contain anything of value and can probably be deleted as well.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/apksize/GenerateVariantPlugin.groovy
rlazo: Done. At one of the many iterations I needed it to get it to work, and then I didn't check back to see if it was necessary.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/apksize/GenerateVariantPlugin.groovy
rlazo: Empty is the base "flavor" we use to measure all the others. I like the idea of having it as a separate flavor to make this clear. I include it's source code in all the others in case we would need to make a change in it down the road. But it's not strictly necessary. I'd remove it if you feel strongly about it.

allisonbm92: I agree that we should have it as a separate flavor, but it should be completely empty. Now that we've eliminated the samples, it doesn't need the SampleCode class. At this point, there's no need to include the source in the other flavors, because it should always be empty.

If you want to keep the source directory in each flavor to prevent future errors, perhaps we should rename it to `base` or something like that. That would better describe its purpose anyway.

rlazo: I've removed the source dependency and `SampleCode`, and I've renamed it to base as it's a much better name to what it actually is. Thanks!

PR #895, opened 2019-10-10
Implementing retry once for FIS Client.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -53,8 +55,6 @@
   private static final String CONTENT_ENCODING_HEADER_KEY = "Content-Encoding";
   private static final String GZIP_CONTENT_ENCODING = "gzip";
 
-  private static final String UNAUTHORIZED_ERROR_MESSAGE =
-      "The request did not have the required credentials.";
   private static final String INTERNAL_SERVER_ERROR_MESSAGE = "There was an internal server error.";
   private static final String NETWORK_ERROR_MESSAGE = "The server returned an unexpected error:";
```
ciarand: missing a space between the colon and the error when you concat using +. It would be slightly less error prone to use either a format string here or have a dedicated method for generating these errors (or both)

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -218,45 +214,45 @@ public InstallationTokenResult generateAuthToken(
...
-          new URL(
-              String.format(
-                  "https://%s/%s/%s?key=%s",
-                  FIREBASE_INSTALLATIONS_API_DOMAIN,
-                  FIREBASE_INSTALLATIONS_API_VERSION,
-                  resourceName,
-                  apiKey));
+      int retryCount = 0;
+      while (retryCount <= MAX_RETRIES) {
+        URL url =
```
ciarand: let's move this out of the loop because it doesn't change

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -218,45 +214,45 @@ public InstallationTokenResult generateAuthToken(
...
-      HttpsURLConnection httpsURLConnection = (HttpsURLConnection) url.openConnection();
-      httpsURLConnection.setConnectTimeout(NETWORK_TIMEOUT_MILLIS);
-      httpsURLConnection.setReadTimeout(NETWORK_TIMEOUT_MILLIS);
-      httpsURLConnection.setDoOutput(true);
-      httpsURLConnection.setRequestMethod("POST");
-      httpsURLConnection.addRequestProperty("Authorization", "FIS_v2 " + refreshToken);
-      httpsURLConnection.addRequestProperty(CONTENT_TYPE_HEADER_KEY, JSON_CONTENT_TYPE);
-      httpsURLConnection.addRequestProperty(ACCEPT_HEADER_KEY, JSON_CONTENT_TYPE);
-      httpsURLConnection.addRequestProperty(CONTENT_ENCODING_HEADER_KEY, GZIP_CONTENT_ENCODING);
+        HttpsURLConnection httpsURLConnection = (HttpsURLConnection) url.openConnection();
```
ciarand: there's a lot of overlap between this method and the other one that sets up a POST request. Can we maybe consolidate some of the shared logic?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -218,45 +214,45 @@ public InstallationTokenResult generateAuthToken(
...
+        httpsURLConnection.setConnectTimeout(NETWORK_TIMEOUT_MILLIS);
+        httpsURLConnection.setReadTimeout(NETWORK_TIMEOUT_MILLIS);
+        httpsURLConnection.setDoOutput(true);
```
ciarand: I don't see where we're writing a request (output) body. Do we still need this setDoOutput call?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -318,6 +314,19 @@ private InstallationTokenResult readGenerateAuthTokenResponse(HttpsURLConnection
     return builder.build();
   }
 
+  // Read the error message from the response.
+  private String readErrorResponse(HttpsURLConnection conn) throws IOException {
+    BufferedReader reader =
+        new BufferedReader(new InputStreamReader(conn.getErrorStream(), Charset.defaultCharset()));
```
ciarand: shouldn't we use UTF8 here instead? are we expecting the backend to return a default-charset encoded response or a utf8 one?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -318,6 +314,19 @@ private InstallationTokenResult readGenerateAuthTokenResponse(HttpsURLConnection
...
+    StringBuilder response = new StringBuilder();
+    for (String input = reader.readLine(); input != null; input = reader.readLine()) {
+      response.append(input);
+    }
+    return String.format(
+        "The server responded with an error. HTTP response:[%d %s %s]",
```
ciarand: nit: space after colon

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -318,6 +314,19 @@ private InstallationTokenResult readGenerateAuthTokenResponse(HttpsURLConnection
...
   }
 
+  // Read the error message from the response.
+  private String readErrorResponse(HttpsURLConnection conn) throws IOException {
+    BufferedReader reader =
+        new BufferedReader(new InputStreamReader(conn.getErrorStream(), Charset.defaultCharset()));
+    StringBuilder response = new StringBuilder();
+    for (String input = reader.readLine(); input != null; input = reader.readLine()) {
+      response.append(input);
+    }
```
ciarand: Hmm, based off my reading of the docs BufferedReader#readLine will return a string (or null) with no newlines, so we're turning this:

```
foo
bar
baz
```
into this:
```
foobarbaz
```

Is that intended?

ankitaj224: Good catch. Thanks. Updated. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Updated. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -97,46 +101,36 @@ public InstallationResponse createFirebaseInstallation(
...
+          default:
+            throw new FirebaseException(readErrorResponse(httpsURLConnection));
+        }
       }
+      throw new FirebaseException(INTERNAL_SERVER_ERROR_MESSAGE);
     } catch (IOException e) {
-      throw new FirebaseInstallationServiceException(
-          NETWORK_ERROR_MESSAGE + e.getMessage(),
-          FirebaseInstallationServiceException.Status.NETWORK_ERROR);
+      throw new FirebaseException(String.format(NETWORK_ERROR_MESSAGE, e.getMessage()));
```
ciarand: does it make sense to embed the IOException as a cause?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ciarand: I wouldn't bother with the system-dependent newline, just use '\n' (Android will never switch this and it's clearer and slightly more efficient)

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: IOException wraps many exceptions : MalformedException, ProtocolException, UnknownServiceException etc. My reasoning to add this as a cause for users to know what went wrong and handle accordingly.

PR #898, opened 2019-10-11
Generalize ComponentDiscovery to any android Service.

vkryachko:
This allows users to declare their own component graphs in their own
dedicated Service.


firebase-components/src/main/java/com/google/firebase/components/ComponentDiscovery.java
```diff
@@ -56,8 +57,29 @@
   private final T context;
   private final RegistrarNameRetriever<T> retriever;
 
+  /**
+   * @deprecated It's unsafe to use when firebase_common is not available. Use {@link
+   *     #forContext(Context, Class)} instead.
+   * @throws IllegalStateException if firebase_common is not available.
+   */
+  @Deprecated
   public static ComponentDiscovery<Context> forContext(Context context) {
```
sheepmaster: When would someone call this deprecated method?

vkryachko: I guess my thinking was that the old firebase-common would call it, although on second thought, old firebase-common will not work with firebase-components anyway as they would have duplicate symbols. Will delete this method then

PR #900, opened 2019-10-11
Update Serializer condition to the check for null.

rlazo:
`k` is a String variable, so `(k instanceof String)` can only be false if `k` is `null`. Updated the condition to reflect this.


bklimt:
I bet this was just copy-pasted from the `Map` case above.

PR #903, opened 2019-10-12
Fix race condition when fieldValueCache is null

### LINKED ISSUE #902: DocumentSnapshot.get throws NullPointerException
* Android Studio version: 3.5
  * Firebase Component: Firestore
  * Component version: 21.1.1

#### Steps to reproduce:

We have been seeing some crashes caused by NPE when calling `DocumentSnapshot.get`:

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.Object java.util.Map.put(java.lang.Object, java.lang.Object)' on a null object reference
       at com.google.firebase.firestore.model.Document.getField(Document.java:141)
       at com.google.firebase.firestore.DocumentSnapshot.getInternal(DocumentSnapshot.java:627)
       at com.google.firebase.firestore.DocumentSnapshot.get(DocumentSnapshot.java:286)
       at com.google.firebase.firestore.DocumentSnapshot.get(DocumentSnapshot.java:244)
       at com.fitifyapps.fitify.data.FirebaseManager.onUserSnapshot(FirebaseManager.java:193)
       at com.fitifyapps.fitify.data.FirebaseManager.access$onUserSnapshot(FirebaseManager.java:34)
       at com.fitifyapps.fitify.data.FirebaseManager$addUserListener$1.onEvent(FirebaseManager.java:99)
       at com.fitifyapps.fitify.data.FirebaseManager$addUserListener$1.onEvent(FirebaseManager.java:34)
       at com.google.firebase.firestore.DocumentReference.lambda$addSnapshotListenerInternal$2(DocumentReference.java:504)
       at com.google.firebase.firestore.DocumentReference$$Lambda$3.a(DocumentReference.java:6)
       at com.google.firebase.firestore.core.AsyncEventListener.lambda$onEvent$0(AsyncEventListener.java:42)
       at com.google.firebase.firestore.core.AsyncEventListener$$Lambda$1.run(AsyncEventListener.java:6)
       at android.os.Handler.handleCallback(Handler.java:873)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:193)
       at android.app.ActivityThread.main(ActivityThread.java:6692)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

#### Relevant Code:

After examining the source code, it seems that this is a race condition that can happen when `Document.getData` and `Document.getField` are called concurrently on different threads. First, `getField` makes sure that `fieldValueCache` is not null. However, `getData` then sets it to null, and `getField` crashes when trying to call the `put` method.

MattSkala:
Fixes #902

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
@@ -117,10 +117,12 @@ public final class Document extends MaybeDocument {
     } else {
       hardAssert(proto != null && converter != null, "Expected proto and converter to be non-null");
 
+      Map<FieldPath, FieldValue> fieldValueCache = this.fieldValueCache;
       if (fieldValueCache == null) {
         // TODO(b/136090445): Remove the cache when `getField` is no longer called during Query
         // ordering.
         fieldValueCache = new ConcurrentHashMap<>();
+        this.fieldValueCache = fieldValueCache;
       }
 
       FieldValue fieldValue = fieldValueCache.get(path);


var-const:
@wilhuff If I'm reading the relevant code correctly, it uses `ConcurrentHashMap`, which is thread-safe, but also relies on reassigning the variable itself, which isn't. One alternative that comes to mind is creating the map non-lazily and clearing it instead of setting it to `null`. Let me know what you think.

/ok-to-test

MattSkala:
@var-const I agree making `fieldValueCache` non-null is a better solution, should I change it?

wilhuff:
`ConcurrentHashMap` is pretty expensive and most users get the whole document contents rather than just an individual field. The `fieldValueCache` is lazily instantiated to avoid adding this overhead to every Document unconditionally. The cache exists at all because internally, when sorting results, we can access the same field multiple times. (The TODO in the code is really indicating that we should apply a Schwartzian transform before sorting at which point we can eliminate the cache altogether.)

This fix seems OK as is.

PR #907, opened 2019-10-14
keep sNetworkRequestUrl public

chandlerding:
Since there are no option to set a custom hostname for firebasestorage api, please keep NetworkRequest.sNetworkRequestUrl public, so that developers can override it with their own reverse proxy, this is particularly useful if *.googleapis.com being blocked.


chandlerding:
@googlebot I signed it!

Issue #712, opened 2019-08-16
Update dagger to 2.24


hameno:
## What feature would you like to see?

Newer dagger version (2.24+) so that we can upgrade our app. This currently fails with:

    dagger.android.support.DaggerApplication_MembersInjector: can't find referenced method 'void injectBroadcastReceiverInjector(dagger.android.DaggerApplication,dagger.android.DispatchingAndroidInjector)' in program class dagger.android.DaggerApplication_MembersInjector

## How would you use it?

n/a

ashwinraghav:
Thanks for notifying us.
Seems like this needs to change across inappmessaging and transport runtime!

prakhar1989:
test

PR #909, opened 2019-10-14
Update FIAM version of Dagger to address #712

JasonAHeron:
Update FIAM version of Dagger to address #712


firebase-inappmessaging-display/firebase-inappmessaging-display.gradle
```diff
@@ -80,10 +80,10 @@ dependencies {
     implementation "com.squareup.picasso:picasso:2.71828"
     implementation "com.squareup.okhttp:okhttp:2.7.5"
 
-    implementation "com.google.dagger:dagger:2.22"
-    implementation "com.google.dagger:dagger-android-support:2.22"
+    implementation "com.google.dagger:dagger:2.24"
+    implementation "com.google.dagger:dagger-android-support:2.24"
```
ashwinraghav: It seems like this dep is never really used. Can we get rid of it?

JasonAHeron: gone now 🗡

PR #911, opened 2019-10-14
FIS error handling

ankitaj224:
For FirebaseInstallationServiceClient#createFirebaseInstallation
1. 5xx errors, **do not update** the PersistedInstallationEntry status to **REGISTER_ERROR** 
2. Store detailed 4xx server exceptions in the local storage. In the succeeding getId calls, return the error message to the clients


firebase-installations/api.txt
```diff
@@ -44,6 +44,7 @@ package com.google.firebase.installations.local {
     method @NonNull public static com.google.firebase.installations.local.PersistedInstallationEntry.Builder builder();
     method @Nullable public abstract String getAuthToken();
     method public abstract long getExpiresInSecs();
+    method @Nullable public abstract String getFISException();
```
andirayo: I think this should be #getFisException

Compare:
http://go/java-style#s5.3-camel-case

andirayo:
Regarding the PR's description:
"""Incase of 500 errors, mark PersistedInstallationEntry status UNREGISTERED"""

I am confused by this sentence.
5XX errors are server-errors that the server usually recovers from after a retry.
I may be missing context, but usually the SDK should retry one time a request that it got a 5XX response for and then give up, but not change the status of the Installation.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -291,6 +292,31 @@ public void testGetId_PersistedInstallationOk_BackendError() throws Exception {
     assertThat(updatedInstallationEntry).hasRegistrationStatus(RegistrationStatus.REGISTER_ERROR);
   }
 
+  @Test
+  public void testGetId_500ServerError_UnregisteredFID() throws Exception {
+    when(backendClientReturnsOk.createFirebaseInstallation(
```
ciarand: where does this test stub a 500 error?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -296,15 +303,17 @@ private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallation
...
-      persistedInstallation.insertOrUpdatePersistedInstallationEntry(
-          PersistedInstallationEntry.builder()
-              .setFirebaseInstallationId(installationResponse.getFid())
-              .setRegistrationStatus(RegistrationStatus.REGISTERED)
-              .setAuthToken(installationResponse.getAuthToken().getToken())
-              .setRefreshToken(installationResponse.getRefreshToken())
-              .setExpiresInSecs(installationResponse.getAuthToken().getTokenExpirationTimestamp())
-              .setTokenCreationEpochInSecs(creationTime)
-              .build());
+      if (installationResponse.getFid() != null) {
```
andirayo: fyi:
If a #createInstallation request gets a successful response, the field "fid" should always be set.

What happens here if the field "fid" is not set? What state will the SDK end up in?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallationEntry.java
```diff
@@ -41,6 +41,9 @@
 
   public abstract long getTokenCreationEpochInSecs();
 
+  @Nullable
+  public abstract String getFISException();
```
ciarand: are we exposing this to clients? Having exception == String is weird. How about just calling it error message?

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallationEntry.java
```diff
@@ -90,6 +93,9 @@ public abstract Builder setRegistrationStatus(
     @NonNull
     public abstract Builder setTokenCreationEpochInSecs(long value);
 
+    @NonNull
+    public abstract Builder setFISException(@Nullable String value);
```
andirayo: why can this be @Nullable?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -128,7 +128,8 @@ public InstallationResponse createFirebaseInstallation(
             throw new FirebaseException(readErrorResponse(httpsURLConnection));
```
andirayo: referring to the 4 lines above:
```case 500: retryCount++;```

Is it possible for us to treat all server-errors (i.e. 5XX) like this?

Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java]: Done. PTAL.

File: firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallationEntry.java (hunk 5 above)
Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallationEntry.java]: When PersistedInstallationEntry stores a successfully registered entry.

File: firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java (hunk 3 above)
Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: This is to differentiate between 200, 5xx & 4xx errors.

200 - createInstallation successful response - fid is not empty
5xx - retry once - empty response & fid
4xx - throws exception

File: firebase-installations/api.txt (hunk 1 above)
Comment (ankitaj224) [reply firebase-installations/api.txt]: Done.thanks.

File: firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallationEntry.java (hunk 4 above)
Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallationEntry.java]: Yes, the stored error message. Changed it. PTAL.

Idea: getId is non-blocking call. After we return the id, we encounter some error during registration. We persist the error message & return it to the clients next time getId is called.

File: firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java (hunk 2 above)
Comment (ankitaj224) [reply firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java]: Sorry for the confusion. Added a comment & changed the test name. Does this makes sense?

**Comment by ankitaj224** [comment]
> Regarding the PR's description:
> """Incase of 500 errors, mark PersistedInstallationEntry status UNREGISTERED"""
> 
> I am confused by this sentence.
> 5XX errors are server-errors that the server usually recovers from after a retry.
> I may be missing context, but usually the SDK should retry one time a request that it got a 5XX response for and then give up, but not change the status of the Installation.

Background: 
My previous impl of FIS Android SDK would retry once on 500, then throw an exception. That would result in persisted installation entry status marked as REGISTER_ERROR. 

After our discussion: On 5xx, we retry once & leave the persisted installation entry status as UNREGISTERED. As a result, no exception will be thrown on subsequent getId calls.

File: firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java (hunk 3 above)
Comment (andirayo) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: If that is what you are doing, then I would recommend your code to express that.
In this case, I would recommend that you check for the 200 HTTP-status instead of the "fid" field.

Comment (ankitaj224) [reply firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java]: For that: I have to include Response Code as a field in InstallationResponse. Is that what you recommend?

File: firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java [hunk 7]
```diff
@@ -118,17 +118,17 @@ public InstallationResponse createFirebaseInstallation(
...
-          default:
-            throw new FirebaseException(readErrorResponse(httpsURLConnection));
+
+        if (httpResponseCode == 200) {
+          return readCreateResponse(httpsURLConnection);
+        } else if (httpResponseCode >= 500 && httpResponseCode < 600) {
+          retryCount++;
+        } else {
+          throw new FirebaseException(readErrorResponse(httpsURLConnection));
         }
```
andirayo: I recommend rewriting this code like this to add a bit more code clarity:

```
  return readCreateResponse(httpsURLConnection);
}
// usually the FIS server recovers from errors: We retry one time before giving up.
if (httpResponseCode >= 500 && httpResponseCode < 600) {
  retryCount++;
  continue;
}

// unrecoverable server response or unknown error
throw new FirebaseException(readErrorResponse(httpsURLConnection));

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ciarand: since y'all control both sides of this exchange we should be able to use and trust the status codes. imo having canonical status codes as a classification for installation responses is a good idea.

ankitaj224: Ack. PTAL.

Issue #482, opened 2019-06-02
Memory leak in FirebaseInAppMessagingDisplay


Thomas-Vos:
### [READ] Step 1: Are you in the right place?
Yes

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.5 Beta 3
  * Firebase Component: firebase-inappmessaging-display
  * Component version: 17.1.1

#### Steps to reproduce:

Install LeakCanary and Firebase In App Messaging in the project. Launch the app. Wait for the leak notification to appear.

LeakCanary is reporting a memory leak for FirebaseInAppMessagingDisplay. Here is part of the leak report:
```
 ┬
 ├─ android.provider.FontsContract
 │    Leaking: NO (it's a GC root and a class is never leaking)
 │    ↓ static FontsContract.sContext
 ├─ com.example.app.App
 │    Leaking: NO (Application is a singleton)
 │    ↓ App.mActivityLifecycleCallbacks
 │          ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ├─ java.util.ArrayList
 │    Leaking: UNKNOWN
 │    ↓ ArrayList.elementData
 │                ~~~~~~~~~~~
 ├─ java.lang.Object[]
 │    Leaking: UNKNOWN
 │    ↓ array Object[].[3]
 │                     ~~~
 ├─ com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay
 │    Leaking: UNKNOWN
 │    ↓ FirebaseInAppMessagingDisplay.firebaseInAppMessagingDisplay
 │                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ├─ com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay$1
 │    Leaking: UNKNOWN
 │    Anonymous class implementing com.google.firebase.inappmessaging.FirebaseInAppMessagingDisplay
 │    ↓ FirebaseInAppMessagingDisplay$1.val$activity
 │                                      ~~~~~~~~~~~~
 ╰→ com.example.app.ui.main.MainActivity
 ​     Leaking: YES (RefWatcher was watching this and MainActivity#mDestroyed is true)
 (removed the rest of the leak dump)
```
[...]

PR #914, opened 2019-10-14
Attempt to fix memory leak referenced in #482

JasonAHeron:
#482 mentions a memory leak because we store a new instance of a class but don't clear it. It seems we don't need to store it at all. So let's not do that then.

### DIFF
--- a/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
+++ b/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
@@ -100,8 +100,6 @@ public class FirebaseInAppMessagingDisplay extends FirebaseInAppMessagingDisplay
   private FiamListener fiamListener;
   private InAppMessage inAppMessage;
   private FirebaseInAppMessagingDisplayCallbacks callbacks;
-  private com.google.firebase.inappmessaging.FirebaseInAppMessagingDisplay
-      firebaseInAppMessagingDisplay;
 
   @Inject
   FirebaseInAppMessagingDisplay(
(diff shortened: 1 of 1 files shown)


firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -185,22 +183,18 @@ public void clearFiamListener() {
   public void onActivityStarted(final Activity activity) {
```
MeghaB: just to confirm - there isn't any additional cleanup we want to take advantage of on other lifecycle events, ex: onActivityCreated vs started or in onActivityDestroyed?

JasonAHeron: I think those already do the correct cleanup as far as I know. May not be enough but I think it's correct in the scope of this change

firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -185,22 +183,18 @@ public void clearFiamListener() {
...
     super.onActivityStarted(activity);
     // Register FIAM listener with the headless sdk.
-    firebaseInAppMessagingDisplay =
-        new com.google.firebase.inappmessaging.FirebaseInAppMessagingDisplay() {
-          @Override
-          public void displayMessage(InAppMessage iam, FirebaseInAppMessagingDisplayCallbacks cb) {
```
ashwinraghav: Was this intended to be a breaking change?

Issue #875, opened 2019-10-02
Firestore Internal Error (21.1.1)


ccjernigan:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.5
  * Firebase Component: Firestore (Database, Firestore, Storage, Functions, etc)
  * Component version: 21.1.1

### [REQUIRED] Step 3: Describe the problem

Firestore is crashing sporadically in production.  I'm seeing the following stack traces together in Crashlytics.

```
Fatal Exception: java.lang.RuntimeException: Internal error in Cloud Firestore (21.1.1).
       at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$3(AsyncQueue.java:529)
       at com.google.firebase.firestore.util.AsyncQueue$$Lambda$3.run(AsyncQueue.java:2000)
       at android.os.Handler.handleCallback(Handler.java:873)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7073)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
```

```
com.google.firebase.firestore.core.SyncEngine.handleCredentialChange (SyncEngine.java:599)
com.google.firebase.firestore.core.FirestoreClient.lambda$new$0 (FirestoreClient.java:105)
com.google.firebase.firestore.core.FirestoreClient$$Lambda$15.run (FirestoreClient.java:3599)
com.google.firebase.firestore.util.AsyncQueue.lambda$enqueue$2 (AsyncQueue.java:431)
[...]

PR #919, opened 2019-10-15
Fix race in credentialsProvider.setChangeListener

schmidt-sebastian:
If the credentials change listeners fires twice before we enqueue the client initialization on the AsyncQueue, `syncEngine` will be null in the credentials callback. This PR re-orders our initialization to make sure that the first thing we execute on the client is always `FirestoreClient.initialize()`.

Fixes https://github.com/firebase/firebase-android-sdk/issues/875

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
@@ -96,19 +96,6 @@ public final class FirestoreClient implements RemoteStore.RemoteStoreCallback {
 
     TaskCompletionSource<User> firstUser = new TaskCompletionSource<>();
     final AtomicBoolean initialized = new AtomicBoolean(false);
-    credentialsProvider.setChangeListener(
-        (User user) -> {
-          if (initialized.compareAndSet(false, true)) {
-            hardAssert(!firstUser.getTask().isComplete(), "Already fulfilled first user task");
-            firstUser.setResult(user);
-          } else {
-            asyncQueue.enqueueAndForget(
-                () -> {
-                  Logger.debug(LOG_TAG, "Credential changed. Current user: %s", user.getUid());
-                  syncEngine.handleCredentialChange(user);
-                });
-          }
-        });
 
     // Defer initialization until we get the current user from the changeListener. This is
     // guaranteed to be synchronously dispatched onto our worker queue, so we will be initialized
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -127,6 +114,21 @@ public FirestoreClient(
...
+        (User user) -> {
+          if (initialized.compareAndSet(false, true)) {
+            hardAssert(!firstUser.getTask().isComplete(), "Already fulfilled first user task");
+            firstUser.setResult(user);
+          } else {
+            // this fires twoce and sync engine isnt' initialized
+            asyncQueue.enqueueAndForget(
+                () -> {
+                  Logger.debug(LOG_TAG, "Credential changed. Current user: %s", user.getUid());
+                  syncEngine.handleCredentialChange(user);
```
var-const: Optional: perhaps assert that `syncEngine` is not null? Or would this be trying too hard to protect against past problems?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -127,6 +114,21 @@ public FirestoreClient(
...
           }
         });
+
+    credentialsProvider.setChangeListener(
+        (User user) -> {
+          if (initialized.compareAndSet(false, true)) {
+            hardAssert(!firstUser.getTask().isComplete(), "Already fulfilled first user task");
+            firstUser.setResult(user);
+          } else {
+            // this fires twoce and sync engine isnt' initialized
```
MattSkala: twoce => twice, isnt' => isn't

Btw does it even make sense to leave the comment here? After the fix, sync engine should be initialized even if this fires twice.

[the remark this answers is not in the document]
schmidt-sebastian: Please ignore this random comment :/ Removed.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
schmidt-sebastian: I don't think it will hurt. Done.

wu-hui:
Will this happen to other SDKs?

schmidt-sebastian:
It shouldn't happen on Web, but it may happen on iOS. iOS PR coming. Thanks for being vigilant!

PR #922, opened 2019-10-16
Add smoke tests for Smart Reply.

yifanyang:


allisonbm92:
Note, the smoke tests are currently broken due to #909.

PR #924, opened 2019-10-16
Reuse existing Iid as Fid

ankitaj224:
Reading iid from "com.google.android.gms.appid" shared pref

### DIFF
--- a/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
+++ b/firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
@@ -28,7 +28,9 @@ import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CRE
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_CREATION_TIMESTAMP_2;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_FID_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_INSTALLATION_RESPONSE;
+import static com.google.firebase.installations.FisAndroidTestConstants.TEST_INSTALLATION_RESPONSE_WITH_IID;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_INSTALLATION_TOKEN_RESULT;
+import static com.google.firebase.installations.FisAndroidTestConstants.TEST_INSTANCE_ID_1;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_PROJECT_ID;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_REFRESH_TOKEN;
 import static com.google.firebase.installations.FisAndroidTestConstants.TEST_TOKEN_EXPIRATION_TIMESTAMP;
@@ -52,6 +54,7 @@ import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseException;
 import com.google.firebase.FirebaseOptions;
+import com.google.firebase.installations.local.IidStore;
 import com.google.firebase.installations.local.PersistedInstallation;
 import com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus;
 import com.google.firebase.installations.local.PersistedInstallationEntry;
@@ -88,10 +91,21 @@ public class FirebaseInstallationsInstrumentedTest {
   @Mock private Utils mockUtils;
   @Mock private PersistedInstallation mockPersistedInstallation;
   @Mock private FirebaseInstallationServiceClient mockClient;
+  @Mock private IidStore mockIidStore;
 
   private static final PersistedInstallationEntry REGISTERED_INSTALLATION_ENTRY =
       PersistedInstallationEntry.builder()
           .setFirebaseInstallationId(TEST_FID_1)
+          .setAuthToken(TEST_AUTH_TOKEN)
+          .setRefreshToken(TEST_REFRESH_TOKEN)
+          .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
+          .setExpiresInSecs(TEST_TOKEN_EXPIRATION_TIMESTAMP)
+          .setRegistrationStatus(PersistedInstallation.RegistrationStatus.REGISTERED)
+          .build();
+
+  private static final PersistedInstallationEntry REGISTERED_IID_ENTRY =
+      PersistedInstallationEntry.builder()
+          .setFirebaseInstallationId(TEST_INSTANCE_ID_1)
           .setAuthToken(TEST_AUTH_TOKEN)
           .setRefreshToken(TEST_REFRESH_TOKEN)
           .setTokenCreationEpochInSecs(TEST_CREATION_TIMESTAMP_2)
@@ -182,6 +196,8 @@ public class FirebaseInstallationsInstrumentedTest {
     doThrow(new FirebaseException("Server Error"))
         .when(backendClientReturnsError)
         .deleteFirebaseInstallation(anyString(), anyString(), anyString(), anyString());
+
+    when(mockIidStore.readIid()).thenReturn(null);
   }
 
   @After
@@ -191,10 +207,15 @@ public class FirebaseInstallationsInstrumentedTest {
 
   @Test
   public void testGetId_PersistedInstallationOk_BackendOk() throws Exception {
-    when(mockUtils.isAuthTokenExpired(REGISTERED_INSTALLATION_ENTRY)).thenReturn(false);
+    when(mockUtils.isAuthTokenExpired(REGISTERED_IID_ENTRY)).thenReturn(false);
     FirebaseInstallations firebaseInstallations =
         new FirebaseInstallations(
-            executor, firebaseApp, backendClientReturnsOk, persistedInstallation, mockUtils);
+            executor,
+            firebaseApp,
+            backendClientReturnsOk,
+            persistedInstallation,
+            mockUtils,
+            mockIidStore);
 
     // No exception, means success.
     assertWithMessage("getId Task failed.")
@@ -214,6 +235,39 @@ public class FirebaseInstallationsInstrumentedTest {
   }
 
   @Test
+  public void testGetId_migrateIid_successful() throws Exception {
+    when(mockIidStore.readIid()).thenReturn(TEST_INSTANCE_ID_1);
+    when(mockUtils.isAuthTokenExpired(REGISTERED_INSTALLATION_ENTRY)).thenReturn(false);
+    when(backendClientReturnsOk.createFirebaseInstallation(
+            anyString(), anyString(), anyString(), anyString()))
+        .thenReturn(TEST_INSTALLATION_RESPONSE_WITH_IID);
+    FirebaseInstallations firebaseInstallations =
+        new FirebaseInstallations(
+            executor,
+            firebaseApp,
+            backendClientReturnsOk,
+            persistedInstallation,
+            mockUtils,
+            mockIidStore);
+
+    // No exception, means success.
+    assertWithMessage("getId Task failed.")
+        .that(Tasks.await(firebaseInstallations.getId()))
+        .isNotEmpty();
+    PersistedInstallationEntry entryValue =
+        persistedInstallation.readPersistedInstallationEntryValue();
+    assertThat(entryValue).hasFid(TEST_INSTANCE_ID_1);
+
+    // Waiting for Task that registers FID on the FIS Servers
+    executor.awaitTermination(500, TimeUnit.MILLISECONDS);
+
+    PersistedInstallationEntry updatedInstallationEntry =
+        persistedInstallation.readPersistedInstallationEntryValue();
+    assertThat(updatedInstallationEntry).hasFid(TEST_INSTANCE_ID_1);
+    assertThat(updatedInstallationEntry).hasRegistrationStatus(RegistrationStatus.REGISTERED);
+  }
+
+  @Test
   public void testGetId_multipleCalls_sameFIDReturned() throws Exception {
     when(mockUtils.isAuthTokenExpired(REGISTERED_INSTALLATION_ENTRY)).thenReturn(false);
     when(backendClientReturnsOk.createFirebaseInstallation(
@@ -221,7 +275,12 @@ public class FirebaseInstallationsInstrumentedTest {
         .thenReturn(TEST_INSTALLATION_RESPONSE);
     FirebaseInstallations firebaseInstallations =
         new FirebaseInstallations(
-            executor, firebaseApp, backendClientReturnsOk, persistedInstallation, mockUtils);
+            executor,
+            firebaseApp,
+            backendClientReturnsOk,
+            persistedInstallation,
+            mockUtils,
+            mockIidStore);
 
     // Call getId multiple times
     Task<String> task1 = firebaseInstallations.getId();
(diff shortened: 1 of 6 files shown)


firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -182,6 +196,8 @@ public void setUp() throws FirebaseException {
     doThrow(new FirebaseException("Server Error"))
         .when(backendClientReturnsError)
         .deleteFirebaseInstallation(anyString(), anyString(), anyString(), anyString());
+
+    when(mockIidStore.readIid()).thenReturn(null);
```
ciarand: this should be the default for any mock

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -710,7 +841,12 @@ public void testDelete_serverError_failure() throws Exception {
     persistedInstallation.insertOrUpdatePersistedInstallationEntry(REGISTERED_INSTALLATION_ENTRY);
     FirebaseInstallations firebaseInstallations =
         new FirebaseInstallations(
-            executor, firebaseApp, backendClientReturnsError, persistedInstallation, mockUtils);
+            executor,
+            firebaseApp,
+            backendClientReturnsError,
+            persistedInstallation,
+            mockUtils,
+            mockIidStore);
```
ciarand: maybe time to move this to a helper

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,108 @@
...
+  private static final String IID_SHARED_PREFS_NAME = "com.google.android.gms.appid";
+  private static final String STORE_KEY_PUB = "|S||P|";
+  private static final String STORE_KEY_ID = "|S|id";
+
+  @GuardedBy("iidPrefs")
+  private final SharedPreferences iidPrefs;
+
+  public IidStore(@NonNull FirebaseApp firebaseApp) {
+    // Different FirebaseApp in the same Android application should have the same application
+    // context and same dir path
```
ciarand: this comment is true, but it's not obvious why it's here? are we doing something differently because of this fact?

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,108 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.PublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.X509EncodedKeySpec;
+
+public class IidStore {
+  private static final String IID_SHARED_PREFS_NAME = "com.google.android.gms.appid";
+  private static final String STORE_KEY_PUB = "|S||P|";
+  private static final String STORE_KEY_ID = "|S|id";
```
ciarand: both of these are missing the subtype part of the key format so they'll lose any subtype'd IDs

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,108 @@
...
+      String id = iidPrefs.getString(STORE_KEY_ID, /* defaultValue= */ null);
+
+      if (id != null) {
+        return id;
+      }
+
+      // Get id from the public key
+      String base64PublicKey = iidPrefs.getString(STORE_KEY_PUB, /* defaultValue= */ null);
+      if (base64PublicKey != null) {
+        return getIdFromPublicKey(parseKey(base64PublicKey));
```
ciarand: need to deal with null

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: Can you please provide some context on subtype. 

I saw that ```SUBTYPE_DEFAULT = "";``` in 
FirebaseInstanceId.java and this value is used for getting id from store.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: if getIdFromPublicKey returns null, i want to return the same. 

Am I missing something?

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: Updated. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ciarand: I don't honestly know much about this topic. Rayo will have more info on subtypes and whether they're important, I just want to know that we're explicitly ok with dropping non-default subtypes

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ciarand: parseKey can return null too, which will cause a NPE not a returned null

ankitaj224: Damn. I missed that. Tysm. Good catch :)

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: @andirayo Can you please take a look at this.

Thanks for pointing it out @ciarand

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Nullable
+  public String readIid() {
+    synchronized (iidPrefs) {
+      // Reading iid from shared preferences
+      String id = iidPrefs.getString(STORE_KEY_ID, /* defaultValue= */ null);
+
+      if (id != null) {
+        return id;
+      }
```
andirayo: (Feel free to ignore this)
How about outsourcing this into its own method, even though its only one line:

private Optional\<String> readInstanceIdFromLocalStorage() {
  return Optional.ofNullable(iidPrefs.getString(STORE_KEY_ID, /* defaultValue= */ null));
}

(after reading my comments below)
You can then rewrite this method like this:

return readInstanceIdFromLocalStorage().orElse(this:: readPublicKeyFromLocalStorageAndCalculateInstanceIdFromIt);

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -220,7 +225,12 @@ private final void doRegistration() {
 
       // New FID needs to be created
       if (persistedInstallationEntry.isNotGenerated()) {
-        String fid = utils.createRandomFid();
+
+        // Read the existing iid for this firebase installation
+        String fid = iidStore.readIid();
+        if (fid == null) {
+          fid = utils.createRandomFid();
+        }
```
andirayo: (please see my comments at the bottom)
Assuming you switch the IidStore class to avoid NULL values, you can rewrite this as
String fid = iidStroe.readIid().orElse(utils::createRandomFid);

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
```diff
@@ -213,15 +235,39 @@ public void testGetId_PersistedInstallationOk_BackendOk() throws Exception {
...
+    // No exception, means success.
+    assertWithMessage("getId Task failed.")
+        .that(Tasks.await(firebaseInstallations.getId()))
+        .isNotEmpty();
+    PersistedInstallationEntry entryValue =
+        persistedInstallation.readPersistedInstallationEntryValue();
+    assertThat(entryValue).hasFid(TEST_INSTANCE_ID_1);
+
+    // Waiting for Task that registers FID on the FIS Servers
+    executor.awaitTermination(500, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
ankitaj224: This test is for getId(). As you know, getId() return fid immediately but registers fid asynchronously. Tests have to wait for half a second while we mock registration. We dont send an actual request to FIS in tests.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: When I looked at the IID Android code: this was the subtype default ``` SUBTYPE_DEFAULT = "";``` Hence, I skipped adding subtype.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,110 @@
...
+    // Different FirebaseApp in the same Android application should have the same application
+    // context and same dir path. We only read existing Iids for the default firebase application.
+    iidPrefs =
+        firebaseApp
+            .getApplicationContext()
+            .getSharedPreferences(IID_SHARED_PREFS_NAME, Context.MODE_PRIVATE);
+  }
+
+  @Nullable
+  public String readIid() {
```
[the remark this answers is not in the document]
ankitaj224: minSDK requirements to use Optional: 24 and firebase-android-sdk minSDK is 14.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,110 @@
...
+          digest, 0, 8, Base64.URL_SAFE | Base64.NO_WRAP | Base64.NO_PADDING);
+    } catch (NoSuchAlgorithmException e) {
+      Log.w(TAG, "Unexpected error, device missing required algorithms");
+    }
+    return null;
+  }
+
+  /** Parse the public key from stored data. */
+  @Nullable
+  private PublicKey parseKey(String base64PublicKey) {
```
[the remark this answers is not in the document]
ankitaj224: minSDK requirements to use Optional: 24 and firebase-android-sdk minSDK is 14.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,110 @@
...
+          return getIdFromPublicKey(publicKey);
+        }
+      }
+
+      return null;
+    }
+  }
+
+  @Nullable
+  private static String getIdFromPublicKey(@NonNull PublicKey publicKey) {
```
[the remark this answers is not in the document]
ankitaj224: minSDK requirements to use Optional: 24 and firebase-android-sdk minSDK is 14.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,110 @@
...
+      // Get id from the public key
+      String base64PublicKey = iidPrefs.getString(STORE_KEY_PUB, /* defaultValue= */ null);
+      if (base64PublicKey != null) {
+        PublicKey publicKey = parseKey(base64PublicKey);
+        if (publicKey != null) {
+          return getIdFromPublicKey(publicKey);
+        }
+      }
+
+      return null;
```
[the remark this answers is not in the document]
ankitaj224: - Ack A and updated the code. PTAL.

- B) minSDK requirements to use Optional: 24 and firebase-android-sdk minSDK is 14.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,110 @@
...
+  private final SharedPreferences iidPrefs;
+
+  public IidStore(@NonNull FirebaseApp firebaseApp) {
+    // Different FirebaseApp in the same Android application should have the same application
+    // context and same dir path. We only read existing Iids for the default firebase application.
+    iidPrefs =
+        firebaseApp
+            .getApplicationContext()
+            .getSharedPreferences(IID_SHARED_PREFS_NAME, Context.MODE_PRIVATE);
+  }
```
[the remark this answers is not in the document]
ankitaj224: We only consider default instance of IidStore. Based on what you said, Iid Android shouldn't even have multiple IIDs right?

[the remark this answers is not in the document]
ankitaj224: I wasn't very clear on the multiple FirebaseApp situation. On understanding on multiple apps work, I agree with you on disallowing instances of Iidstore. 

This is how I plan to handle it: for default app - if no fid in local storage, read iid store
For other apps - if no fid, create new fid ( skip reading iid store)

Does this makes sense to you?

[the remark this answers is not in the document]
ankitaj224: Resolved. PTAL.

firebase-installations/src/androidTest/java/com/google/firebase/installations/FirebaseInstallationsInstrumentedTest.java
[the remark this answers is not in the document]
andirayo: I would recommend to add this explanation to the comment in the code.
Also, 500ms sounds excessive, but can't hurt maybe... :)

firebase-installations/src/androidTest/java/com/google/firebase/installations/FisAndroidTestConstants.java
```diff
@@ -59,6 +61,20 @@
           .setResponseCode(ResponseCode.OK)
           .build();
 
+  public static final InstallationResponse TEST_INSTALLATION_RESPONSE_WITH_IID =
+      InstallationResponse.builder()
+          .setUri("/projects/" + TEST_PROJECT_ID + "/installations/" + TEST_INSTANCE_ID_1)
```
andirayo: you could use a constant for this as well:
FIS_URI = "/projects/%s/installations/%s";

but not important.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class IidStore {
+  private static final String IID_SHARED_PREFS_NAME = "com.google.android.gms.appid";
+  private static final String STORE_KEY_PUB = "|S||P|";
+  private static final String STORE_KEY_ID = "|S|id";
+
+  @GuardedBy("iidPrefs")
+  private final SharedPreferences iidPrefs;
+
+  public IidStore() {
```
andirayo: please add a comment to this method and the entire class, that this should only be called for the [default] instance of FirebaseApp (i.e. the Firebase Core SDK).

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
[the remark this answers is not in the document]
andirayo: Not using Optional here makes me sad...
Still I would recommend to implement the other recommendations from above:

=================================
// If this App-Instance did not store its Instance-ID in storage,
// we may be able to find its Public-Key in order to calculate the App-Instance's Instance-ID from.
return readPublicKeyAndCalculateInstanceIdFromIt();

=================================
private String readPublicKeyFromLocalStorageAndCalculateInstanceIdFromIt() {
  String base64PublicKey = iidPrefs.getString(STORE_KEY_PUB, /* defaultValue= */ null);
  if (base64PublicKey == null) {
    return null;
  }
  
  PublicKey publicKey = parseKey(base64PublicKey);
  if (publicKey == null) {
    return null;
  }
  
  return getIdFromPublicKey(publicKey);
}

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: Done.

PR #925, opened 2019-10-17
Migrate fiam to lflg.

vkryachko:


MeghaB:
This seems like more than just pulling out the constant -> if I'm reading this correctly, the new value isn't 'cct', but something that includes a bit more of the path? How should we test this to confirm this new value works? Cheers!

JasonAHeron:
Let me to an e2e test and wait for the events to come through

PR #927, opened 2019-10-17
Move changes to the correct release in release notes.

wu-hui:


allisonbm92:
Note, the smoke-tests are broken. They should be fixed in a couple of hours.

schmidt-sebastian:
Feel free to merge as is, since this PR only affects the Changelog.

PR #940, opened 2019-10-24
Modify ConfigFetchHttpClient to log all successful Fetch calls.

jacquetd:
Modify ConfigFetchHttpClient so that metrics will be logged in Fetch even when the backend doesn't have an update.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
@@ -210,14 +210,14 @@ public class ConfigFetchHttpClient {
       }
     }
 
+    long endTime = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -210,15 +210,15 @@ FetchResponse fetch(
       }
     }
 
+    long endTime = System.currentTimeMillis();
```
clp93: Can we add a corresponding test to https://github.com/firebase/firebase-android-sdk/blob/rc-telemetry/firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java ? Thanks!

jacquetd: Done and ran test successfully.

PR #943, opened 2019-10-25
Update release version information for firestore.

rlazo:


rosariopfernandes:
I'm not sure if the `api.txt` file has any influence on releases, but if it does, we might want to consider running `./gradlew firebase-firestore:generateApiTxtFile` before release. Because #938 made `FirebaseFirestoreSettings.Builder.getHost()` NonNull, but its entry on `api.txt` is still Nullable:

https://github.com/firebase/firebase-android-sdk/blob/4aac4cf949e95b6d69628d47ede55c146810f850/firebase-firestore/api.txt#L222

rlazo:
Thanks for bringing this up @rosariopfernandes ! The last release of Firestore (21.2.1) does *not* include the changes in the Builder, so it's not affected. That being said, we do need to update `api.txt` before next release. Thanks!

PR #939, opened 2019-10-24
Make IN queries publicly available 

thebrianchen:

PR #946, opened 2019-10-28
Add wildcard for List and varargs support for IN queries

thebrianchen:


firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -278,10 +301,29 @@ Query whereArrayContainsAny(@NonNull String field, @NonNull List<Object> value)
...
+   * combined with {@code whereArrayContains()} or {@code whereIn()}.
+   *
+   * @param fieldPath The path of the field containing an array to search.
+   * @param first The first value of the array that contains the values to match.
+   * @param rest The other values of the array that contain the values to match.
+   * @return The created {@code Query}.
+   */
+  // TODO(in-queries): Expose to public once backend is ready.
+  @NonNull
+  Query whereArrayContainsAny(@NonNull FieldPath fieldPath, Object first, Object... rest) {
```
schmidt-sebastian: Maybe rename first argument to "value" and "rest" to "moreValues" (which matches our update() varargs method).

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -260,10 +261,32 @@ public Query whereArrayContains(@NonNull FieldPath fieldPath, @NonNull Object va
...
+   * combined with {@code whereArrayContains()} or {@code whereIn()}.
+   *
+   * @param field The name of the field containing an array to search.
+   * @param value The first value to match against.
+   * @param moreValues Additional values to match against.
+   * @return The created {@code Query}.
+   */
+  // TODO(in-queries): Expose to public once backend is ready.
+  @NonNull
+  Query whereArrayContainsAny(@NonNull String field, @NonNull Object value, Object... moreValues) {
```
wilhuff: Isn't `null` a valid value for `value`? Seems like that should be `@Nullable`.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -295,10 +337,29 @@ Query whereArrayContainsAny(@NonNull FieldPath fieldPath, @NonNull List<Object>
...
+   * {@code whereArrayContainsAny()}.
+   *
+   * @param field The name of the field to search.
+   * @param value The first value to match against.
+   * @param moreValues Additional values to match against.
+   * @return The created {@code Query}.
+   */
+  // TODO(in-queries): Expose to public once backend is ready.
+  @NonNull
+  Query whereIn(@NonNull String field, Object value, Object... moreValues) {
```
wilhuff: missing nullability specifier on `value`. Here and below.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
```diff
@@ -570,6 +570,10 @@ public void testQueriesAreValidatedOnClient() {
     query.whereArrayContainsAny("y", Arrays.asList("cat"));
     // In on different field works.
     query.whereIn("y", Arrays.asList("cat"));
+    // Array contains any with varargs works.
+    collection.whereArrayContainsAny("a", "b", "c", 42);
```
wilhuff: Could you add a test that verifies `collection.whereArrayContainsAny("field", null)` actually works (not just compiles)?

thebrianchen:
After discussing offline, we've decided to punt adding varargs until later. I'll add the wildcard element the main PR (#939).

PR #951, opened 2019-10-30
Remove fiam pointless null checks

JasonAHeron:


JasonAHeron:
verified works on test app

PR #961, opened 2019-11-05
getAuthToken error handling for 401 & 404 response code

ankitaj224:
- Clear the local storage;  throw an exception with a message to call getId() [ This results in recreation of valid fid & auth token].

### DIFF
--- a/firebase-installations/api.txt
+++ b/firebase-installations/api.txt
@@ -17,6 +17,7 @@ package com.google.firebase.installations {
   }
 
   public enum FirebaseInstallationsException.Status {
+    enum_constant public static final com.google.firebase.installations.FirebaseInstallationsException.Status AUTHENTICATION_ERROR;
     enum_constant public static final com.google.firebase.installations.FirebaseInstallationsException.Status CLIENT_ERROR;
     enum_constant public static final com.google.firebase.installations.FirebaseInstallationsException.Status SDK_INTERNAL_ERROR;
   }
@@ -38,7 +39,6 @@ package com.google.firebase.installations.local {
   }
 
   public enum PersistedInstallation.RegistrationStatus {
-    enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus AUTHENTICATION_TOKEN_ERROR;
     enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus NOT_GENERATED;
     enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus REGISTERED;
     enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus REGISTER_ERROR;
@@ -55,7 +55,6 @@ package com.google.firebase.installations.local {
     method @Nullable public abstract String getRefreshToken();
     method @NonNull public abstract com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus getRegistrationStatus();
     method public abstract long getTokenCreationEpochInSecs();
-    method public boolean isAuthTokenErrored();
     method public boolean isErrored();
     method public boolean isNotGenerated();
     method public boolean isRegistered();
@@ -118,6 +117,8 @@ package com.google.firebase.installations.remote {
     method @Nullable public abstract com.google.firebase.installations.remote.TokenResult.ResponseCode getResponseCode();
     method @Nullable public abstract String getToken();
     method @NonNull public abstract long getTokenExpirationTimestamp();
+    method public boolean isErrored();
+    method public boolean isSuccessful();
     method @NonNull public abstract com.google.firebase.installations.remote.TokenResult.Builder toBuilder();
   }
 
--- a/firebase-installations/api.txt
+++ b/firebase-installations/api.txt
@@ -88,7 +88,7 @@ package com.google.firebase.installations.remote {
   public abstract class InstallationResponse {
     ctor public InstallationResponse();
     method @NonNull public static com.google.firebase.installations.remote.InstallationResponse.Builder builder();
-    method @Nullable public abstract InstallationTokenResult getAuthToken();
+    method @Nullable public abstract com.google.firebase.installations.remote.TokenResult getAuthToken();
     method @Nullable public abstract String getFid();
     method @Nullable public abstract String getRefreshToken();
     method @Nullable public abstract com.google.firebase.installations.remote.InstallationResponse.ResponseCode getResponseCode();
@@ -99,7 +99,7 @@ package com.google.firebase.installations.remote {
   public abstract static class InstallationResponse.Builder {
     ctor public InstallationResponse.Builder();
     method @NonNull public abstract com.google.firebase.installations.remote.InstallationResponse build();
-    method @NonNull public abstract com.google.firebase.installations.remote.InstallationResponse.Builder setAuthToken(@NonNull InstallationTokenResult);
+    method @NonNull public abstract com.google.firebase.installations.remote.InstallationResponse.Builder setAuthToken(@NonNull com.google.firebase.installations.remote.TokenResult);
     method @NonNull public abstract com.google.firebase.installations.remote.InstallationResponse.Builder setFid(@NonNull String);
     method @NonNull public abstract com.google.firebase.installations.remote.InstallationResponse.Builder setRefreshToken(@NonNull String);
     method @NonNull public abstract com.google.firebase.installations.remote.InstallationResponse.Builder setResponseCode(@NonNull com.google.firebase.installations.remote.InstallationResponse.ResponseCode);
--- a/firebase-installations/api.txt
+++ b/firebase-installations/api.txt
@@ -117,7 +117,6 @@ package com.google.firebase.installations.remote {
     method @Nullable public abstract com.google.firebase.installations.remote.TokenResult.ResponseCode getResponseCode();
     method @Nullable public abstract String getToken();
     method @NonNull public abstract long getTokenExpirationTimestamp();
-    method public boolean isAuthTokenErrored();
     method public boolean isSuccessful();
     method @NonNull public abstract com.google.firebase.installations.remote.TokenResult.Builder toBuilder();
   }
--- a/firebase-installations/api.txt
+++ b/firebase-installations/api.txt
@@ -117,7 +117,7 @@ package com.google.firebase.installations.remote {
     method @Nullable public abstract com.google.firebase.installations.remote.TokenResult.ResponseCode getResponseCode();
     method @Nullable public abstract String getToken();
     method @NonNull public abstract long getTokenExpirationTimestamp();
-    method public boolean isErrored();
+    method public boolean isAuthTokenErrored();
     method public boolean isSuccessful();
     method @NonNull public abstract com.google.firebase.installations.remote.TokenResult.Builder toBuilder();
   }
--- a/firebase-installations/api.txt
+++ b/firebase-installations/api.txt
@@ -38,6 +38,7 @@ package com.google.firebase.installations.local {
   }
 
   public enum PersistedInstallation.RegistrationStatus {
+    enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus AUTHENTICATION_TOKEN_ERROR;
     enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus NOT_GENERATED;
     enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus REGISTERED;
     enum_constant public static final com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus REGISTER_ERROR;
@@ -54,6 +55,7 @@ package com.google.firebase.installations.local {
     method @Nullable public abstract String getRefreshToken();
     method @NonNull public abstract com.google.firebase.installations.local.PersistedInstallation.RegistrationStatus getRegistrationStatus();
     method public abstract long getTokenCreationEpochInSecs();
+    method public boolean isAuthTokenErrored();
     method public boolean isErrored();
     method public boolean isNotGenerated();
     method public boolean isRegistered();
@@ -81,7 +83,7 @@ package com.google.firebase.installations.remote {
     ctor public FirebaseInstallationServiceClient(@NonNull Context);
     method @NonNull public com.google.firebase.installations.remote.InstallationResponse createFirebaseInstallation(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
     method @NonNull public void deleteFirebaseInstallation(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
-    method @NonNull public InstallationTokenResult generateAuthToken(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
+    method @NonNull public com.google.firebase.installations.remote.TokenResult generateAuthToken(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
   }
 
   public abstract class InstallationResponse {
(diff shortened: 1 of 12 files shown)


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -270,6 +271,19 @@ private final void doRegistration() {
...
 
+      // If persisted fid is in AUTH_TOKEN_ERROR state, clear the stored fid and update the
+      // listener with an exception.
+      if (persistedInstallationEntry.isAuthTokenErrored()) {
+        persistedInstallation.clear();
+        triggerOnException(
+            persistedInstallationEntry,
+            new FirebaseInstallationsException(
+                "Failed to generate auth token for this Firebase Installation. Call getId() "
+                    + "to recreate a new Fid and a valid auth token.",
```
diwu-arete: You should not only mention this important note in the exception msg.

It needs to be documented in the FirebaseInstallationsException class, basically I think there should be a specific status which indicates that the developer has to call getId() again.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -345,28 +359,37 @@ private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallation
...
+      if (tokenResult.getResponseCode() == TokenResult.ResponseCode.OK) {
+        persistedInstallation.insertOrUpdatePersistedInstallationEntry(
+            persistedInstallationEntry
+                .toBuilder()
+                .setRegistrationStatus(RegistrationStatus.REGISTERED)
+                .setAuthToken(tokenResult.getToken())
+                .setExpiresInSecs(tokenResult.getTokenExpirationTimestamp())
+                .setTokenCreationEpochInSecs(creationTime)
+                .build());
+      }
```
diwu-arete: Pls use
if ()
else if ()
else

for completeness

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -345,28 +359,37 @@ private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallation
...
-              .build());
-
-      return tokenResult;
+      if (tokenResult.getResponseCode() == TokenResult.ResponseCode.FID_ERROR
+          || tokenResult.getResponseCode() == TokenResult.ResponseCode.REFRESH_TOKEN_ERROR) {
+        persistedInstallation.insertOrUpdatePersistedInstallationEntry(
+            persistedInstallationEntry
+                .toBuilder()
+                .setRegistrationStatus(RegistrationStatus.AUTHENTICATION_TOKEN_ERROR)
+                .build());
```
diwu-arete: In the auth error case, if anyways you will clear the cache, why do you update it to AUTHENTICATION_TOKEN_ERROR and then clear it in doRegistration()?

An alternative is to clear it here, and like your old code, still return TokenResult from this method.

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallation.java
```diff
@@ -53,6 +53,7 @@
      * locally before registering with FIS servers.
      */
     REGISTER_ERROR,
+    AUTHENTICATION_TOKEN_ERROR,
```
diwu-arete: How about mentioning the difference between REGISTER_ERROR and AUTHENTICATION_TOKEN_ERROR?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -31,6 +31,7 @@
 import com.google.firebase.installations.remote.FirebaseInstallationServiceClient;
 import com.google.firebase.installations.remote.InstallationResponse;
 import com.google.firebase.installations.remote.InstallationResponse.ResponseCode;
+import com.google.firebase.installations.remote.TokenResult;
```
diwu-arete: Github doesn't allow me to leave comment for unchanged lines, so I leave a comment here for Line#61 in this file:

How about using a more specific name for the variable 'lock' to indicate what the lock is used for?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Makes sense. Changed it. However, I retained the return type as void because it is unused by the caller of the method.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: It is a generic lock used in this class for state listeners and boolean variable. Idk what would be the right way to capture the purpose aptly.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -345,28 +358,30 @@ private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallation
...
+            persistedInstallationEntry
+                .toBuilder()
+                .setRegistrationStatus(RegistrationStatus.REGISTERED)
+                .setAuthToken(tokenResult.getToken())
+                .setExpiresInSecs(tokenResult.getTokenExpirationTimestamp())
+                .setTokenCreationEpochInSecs(creationTime)
+                .build());
+      } else if (tokenResult.isErrored()) {
+        persistedInstallation.clear();
+      }
```
diwu-arete: add a comment saying what's the *else" after this, and why you do a non-op for it.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallationsException.java
```diff
@@ -20,11 +20,12 @@
 /** The class for all Exceptions thrown by {@link FirebaseInstallations}. */
 public class FirebaseInstallationsException extends FirebaseException {
 
-  // TODO(ankitagj): Improve clear exception handling.
   public enum Status {
```
diwu-arete: I think we need clear javadoc for those enum, what do they mean and what does developer need to do,

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
```diff
@@ -0,0 +1,80 @@
...
+
+package com.google.firebase.installations.remote;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+
+/** This class represents a set of values describing a FIS Auth Token Result. */
+@AutoValue
+public abstract class TokenResult {
```
diwu-arete: As you added this TokenResult, should we hide the proto class InstallationTokenResult only in the remote package?

Because I noticed that FirebaseInstallations class uses both.

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public boolean isSuccessful() {
+    return getResponseCode() == ResponseCode.OK;
+  }
+
+  public boolean isErrored() {
+    return getResponseCode() == ResponseCode.FID_ERROR
+        || getResponseCode() == ResponseCode.REFRESH_TOKEN_ERROR;
+  }
+
```
diwu-arete: Will this always be a bi-state? If yes, then I prefer only having isSuccessful()

otherwise when people read the code
if (isSuccessful)
else if (isErrored)
it's a bit confusing that is there another status?

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
ankitaj224: Both classes are needed in FirebaseInstallations. 

InstallationTokenResult as the response of the FIS SDK generateAUthToken API.
TokenResult as the response of FIS Service Client.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -228,6 +228,17 @@ public InstallationTokenResult generateAuthToken(
...
+
+        if (httpResponseCode == 401) {
+          return TokenResult.builder()
+              .setResponseCode(TokenResult.ResponseCode.REFRESH_TOKEN_ERROR)
+              .build();
+        }
+
+        if (httpResponseCode == 404) {
+          return TokenResult.builder().setResponseCode(TokenResult.ResponseCode.FID_ERROR).build();
+        }
```
andirayo: question: why does this not trigger a deletion of the PersistedInstallationEntry?
and if that is done somewhere else, could you please add a comment here, that the FID_ERROR will trigger a reset of the FIS SDK elsewhere?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -228,6 +228,17 @@ public InstallationTokenResult generateAuthToken(
         if (httpResponseCode == 200) {
           return readGenerateAuthTokenResponse(httpsURLConnection);
         }
+
+        if (httpResponseCode == 401) {
+          return TokenResult.builder()
+              .setResponseCode(TokenResult.ResponseCode.REFRESH_TOKEN_ERROR)
+              .build();
+        }
```
andirayo: So, the request was not authenticated, what is happening next?
Are we deleting the PersistedInstallationEntry?
If yes, where is this done?
Can you please add a comment what REFRES_TOKEN_ERROR will trigger elsewhere?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -228,6 +228,17 @@ public InstallationTokenResult generateAuthToken(
         if (httpResponseCode == 200) {
           return readGenerateAuthTokenResponse(httpsURLConnection);
```
andirayo: is there a reason to not just return the response?

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
```diff
@@ -0,0 +1,80 @@
...
+  public boolean isSuccessful() {
+    return getResponseCode() == ResponseCode.OK;
+  }
+
+  public boolean isErrored() {
+    return getResponseCode() == ResponseCode.FID_ERROR
+        || getResponseCode() == ResponseCode.REFRESH_TOKEN_ERROR;
+  }
```
andirayo: why is this not the following?
return !isSuccessful();

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
andirayo: +1

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /** A new FIS Auth-Token, created for this Firebase Installation. */
+  @Nullable
+  public abstract String getToken();
+  /**
+   * The amount of time, in seconds, before the auth-token expires for this Firebase Installation.
+   */
+  @NonNull
+  public abstract long getTokenExpirationTimestamp();
```
andirayo: Here is a contradiction:
The JavaDoc indicates this is a duration (a.k.a. "expires_in"), but the name of the variable indicates it is a timestamp (i.e. "expiration_date").
Can you please be more precise and fix either JavaDoc or variable name

firebase-installations/api.txt
```diff
@@ -110,5 +111,30 @@ package com.google.firebase.installations.remote {
...
+    method @NonNull public abstract com.google.firebase.installations.remote.TokenResult build();
+    method @NonNull public abstract com.google.firebase.installations.remote.TokenResult.Builder setResponseCode(@NonNull com.google.firebase.installations.remote.TokenResult.ResponseCode);
+    method @NonNull public abstract com.google.firebase.installations.remote.TokenResult.Builder setToken(@NonNull String);
+    method @NonNull public abstract com.google.firebase.installations.remote.TokenResult.Builder setTokenExpirationTimestamp(long);
+  }
+
+  public enum TokenResult.ResponseCode {
+    enum_constant public static final com.google.firebase.installations.remote.TokenResult.ResponseCode FID_ERROR;
+    enum_constant public static final com.google.firebase.installations.remote.TokenResult.ResponseCode OK;
+    enum_constant public static final com.google.firebase.installations.remote.TokenResult.ResponseCode REFRESH_TOKEN_ERROR;
```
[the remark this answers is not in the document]
ankitaj224: This file is auto-generated.

firebase-installations/api.txt
```diff
@@ -110,5 +111,30 @@ package com.google.firebase.installations.remote {
...
+  }
+
```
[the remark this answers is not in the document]
ankitaj224: This is auto-generated api.txt file.

firebase-installations/api.txt
```diff
@@ -81,7 +82,7 @@ package com.google.firebase.installations.remote {
     ctor public FirebaseInstallationServiceClient(@NonNull Context);
     method @NonNull public com.google.firebase.installations.remote.InstallationResponse createFirebaseInstallation(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
     method @NonNull public void deleteFirebaseInstallation(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
-    method @NonNull public InstallationTokenResult generateAuthToken(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
+    method @NonNull public com.google.firebase.installations.remote.TokenResult generateAuthToken(@NonNull String, @NonNull String, @NonNull String, @NonNull String);
```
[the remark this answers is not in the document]
ankitaj224: This is auto-generated api.txt file.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallationsException.java
ankitaj224: Added a TODO.

firebase-installations/src/main/java/com/google/firebase/installations/remote/TokenResult.java
ankitaj224: As discussed offline, not exactly a bi-state. Updated the method name.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -345,28 +358,30 @@ private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallation
...
 
   /** Calls the FIS servers to generate an auth token for this Firebase installation. */
-  private InstallationTokenResult fetchAuthTokenFromServer(
-      PersistedInstallationEntry persistedInstallationEntry) throws FirebaseInstallationsException {
+  private void fetchAuthTokenFromServer(PersistedInstallationEntry persistedInstallationEntry)
+      throws FirebaseInstallationsException {
     try {
       long creationTime = utils.currentTimeInSecs();
-      InstallationTokenResult tokenResult =
+      TokenResult tokenResult =
```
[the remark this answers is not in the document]
ankitaj224: Acked.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -345,28 +358,30 @@ private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallation
...
+                .setRegistrationStatus(RegistrationStatus.REGISTERED)
+                .setAuthToken(tokenResult.getToken())
+                .setExpiresInSecs(tokenResult.getTokenExpirationTimestamp())
+                .setTokenCreationEpochInSecs(creationTime)
+                .build());
+      } else if (tokenResult.isErrored()) {
+        persistedInstallation.clear();
+      }
+
     } catch (FirebaseException exception) {
```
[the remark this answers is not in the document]
ankitaj224: Added a TODO to improve Exception handling in general.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -270,6 +271,18 @@ private final void doRegistration() {
         }
       }
 
+      // If persisted installation entry is in NOT_GENERATED state, it was cleared due to
+      // authentication error during auth token generation.
+      if (persistedInstallationEntry.isNotGenerated()) {
```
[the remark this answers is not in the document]

PR #963, opened 2019-11-07
[1/3] Split Query and Target and make RemoteStore and LocalStore work with Targets

wu-hui:
First PR to implement limitToLast on Android.

Corresponding JS PR: https://github.com/firebase/firebase-js-sdk/pull/2254

Future PR will be:

[2/3] Make SyncEngine able to handle Query/Target Mapping.

[3/3] Implement limitToLast

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
@@ -24,16 +24,16 @@ import java.util.List;
 
 /**
  * A Target represents the WatchTarget representation of a Query, which is used by the LocalStore
- * and the RemoteStore to keep track of and to execute backend queries. While multiple Query can map
- * to the same Target, each Targets maps to a single WatchTarget in RemoteStore and a single
+ * and the RemoteStore to keep track of and to execute backend queries. While multiple Queries can
+ * map to the same Target, each Target maps to a single WatchTarget in RemoteStore and a single
  * TargetData entry in persistence.
  */
 public final class Target {
   public static final long NO_LIMIT = -1;
 
-  private @Nullable String memorizedCannonicalId;
+  private @Nullable String memoizedCannonicalId;
 
-  private final List<OrderBy> explicitSortOrder;
+  private final List<OrderBy> orderBy;
   private final List<Filter> filters;
 
   private final ResourcePath path;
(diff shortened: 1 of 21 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -0,0 +1,240 @@
...
+
+  private final @Nullable Bound startAt;
+  private final @Nullable Bound endAt;
+
+  /**
+   * Initializes a Target with a path and additional query constraints. Path must currently be empty
+   * if this is a collection group query.
+   *
+   * <p>NOTE: you should always construct Target from {@code Query.toTarget} instead of using this
+   * constructor, because Query provides an implicit {@code orderBy} property.
```
mikelehen: I left a related comment below, but this seems a little weird to me.  The reason Query has the implicit vs explicit thing is because we let users omit some of the necessary orderBy clauses and then we implicitly add them (e.g. whenever you do an inequality filter you must also order by that field).  But when we construct a Target directly, we should probably be creating it with the correct orderBy's directly.

I don't feel super strongly though.  If we think this is safer / cleaner, I'm fine with it.

wu-hui: When I wrote this in JS, I was not thinking about the case of de-serialization. My plan is to make `Target` constructor only callable from the same package..but TS does not support package-private access?

I forgot to port this here though..now I made `Target` constructor `default` access. It does make de-serialization a little awkward, but now at least people would not be able to construct a `Target` directly. I think this is slightly better.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java
```diff
@@ -807,7 +807,7 @@ public Query decodeQueryTarget(QueryTarget target) {
       endAt = decodeBound(query.getEndAt());
     }
 
-    return new Query(path, collectionGroup, filterBy, orderBy, limit, startAt, endAt);
+    return new Query(path, collectionGroup, filterBy, orderBy, limit, startAt, endAt).toTarget();
```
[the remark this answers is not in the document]
wu-hui: See my reply in `Target`.

PR #969, opened 2019-11-11
[2.9/3] Implement limitToLast and add unit/integration tests.

wu-hui:
This PR implements limitToLast and adds unit and integration tests to test limitToLast and mirror queries.

We still have one more PR to do, which is make spec test work with both query and target and add a spec tests where backend rejects mirror queries.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -20,6 +20,7 @@ import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCol
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollectionWithDocs;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
+import static com.google.firebase.firestore.testutil.TestUtil.expectError;
 import static com.google.firebase.firestore.testutil.TestUtil.map;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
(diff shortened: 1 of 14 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -581,12 +581,19 @@ QueryData getQueryData(Target target) {
     int targetId;
   }
 
-  /** Unpin all the documents associated with the given target. */
+  /**
+   * Unpin all the documents associated with the given target.
+   *
+   * <p>Releasing a non-existing target is a no-op.
```
mikelehen: Hrm...  why do we allow this now?  It seems like we should not be double-releasing targets (if we do, it suggests we released it too early). But I may be confused or missing some context.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -323,13 +323,16 @@ public void handleCredentialChange() {
 
   // Watch Stream
 
-  /** Listens to the target identified by the given QueryData. */
+  /**
+   * Listens to the target identified by the given QueryData.
+   *
+   * <p>It is a no-op if the target of the given query data is already being listened to.
```
mikelehen: Why do we make it a no-op to double-listen or double-unlisten from queries?  It seems like the calling code should still be doing the proper accounting (if we double-unlisten that suggests that we unlistened too early and it's a bug)...  So I'd be inclined to keep this strict, in the hopes of preventing potential bugs.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
wu-hui: Sorry, I should have included more context in the PR.

We decided to do this (make LocalStore.allocateTarget|releaseTarget and RemoteStore.listen|unlisten) mainly because multi-tab in JS make it really difficult to track the state of targets in local store and remote store. The amount of code I need to put into SyncEngine to track each target to avoid double-listen or double-allocate/release made it hard to understand.

I suppose with Android and iOS we could have avoided this and make sure sync engine never double listen. I still do it  to mainly keep consistent with JS.

But I also feel like the old way was kind of leaking implementation details upwards,  in the sense that sync engine must be aware and check for double listen for local/remote store. The fact that 1:N target query mapping + multi-tab made this really obvious and painful. So maybe it's not a bad idea to make all of those method re-entrant?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
wu-hui: See above.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -80,6 +81,283 @@ public void testLimitQueriesUsingDescendingSortOrder() {
...
+    CollectionReference collection =
+        testCollectionWithDocs(
+            map(
+                "a", map("k", "a", "sort", 0),
+                "b", map("k", "b", "sort", 1),
+                "c", map("k", "c", "sort", 1),
+                "d", map("k", "d", "sort", 2)));
+
+    List<QuerySnapshot> snapshots = new ArrayList<>();
+    Semaphore testCounter = new Semaphore(0);
```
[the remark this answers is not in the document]
wu-hui: Done.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -80,6 +81,283 @@ public void testLimitQueriesUsingDescendingSortOrder() {
...
+    assertEquals(asList(map("k", "e", "sort", -1L), map("k", "a", "sort", 0L)), data);
+
+    waitFor(limitToLastTestCounter);
+    assertEquals(2, limitToLastSnapshots.size());
+    data = querySnapshotToValues(limitToLastSnapshots.get(1));
+    assertEquals(asList(map("k", "a", "sort", 0L), map("k", "e", "sort", -1L)), data);
+  }
+
+  @Test
+  public void testUnlistenFromMirrorQueries() {
```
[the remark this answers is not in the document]
wu-hui: The only reason is if one of them failed, it might be easier to figure out where it failed since some of them cover smaller scope. 

That said, I don't think this justifies the amount of duplication here..I merged them into one.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -80,6 +81,283 @@ public void testLimitQueriesUsingDescendingSortOrder() {
...
+
+    waitFor(limitTestCounter);
+    waitFor(limitToLastTestCounter);
+
+    assertEquals(1, limitSnapshots.size());
+    assertEquals(2, limitToLastSnapshots.size());
+  }
+
+  @Test
+  public void testRelistenToMirrorQueries() {
```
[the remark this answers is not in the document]
wu-hui: Combined.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
mikelehen: Thanks for the explanation.

I'm not sure I want to push back too hard, but I do want to clarify my understanding.  It seems like the calling code still needs to do the appropriate tracking / accounting in order to avoid double-calling unlisten or releaseTarget (I'm less concerned about double-calling listen or allocateTarget).  If we double-call unlisten, doesn't that mean we likely unlistened too early and therefore potentially missed some server updates for the target?  Similarly, double-calling releaseTarget likely means we released the target too early, potentially making it eligible for garbage collection when it shouldn't be.

Or am I confused and there are cases when it's normal / safe for the SDK code to e.g. unlisten to the same target multiple times?

In my mind the tracking / accounting we were doing was unavoidable and so we made listen/unlisten strict about it just to sanity-check that we were doing it right.

Given this is just porting the JS logic, I'll go ahead and approve for now, but maybe we can sync up tomorrow at standup?

mikelehen: And by "tomorrow at standup" I mean Monday or whenever you're back in the office. 😂

mikelehen: FYI- I tried adding the strict checking back to unlisten() and releaseTarget() in the JS SDK and all the tests still seem to pass (which is a relief since I still think if they didn't pass that would indicate a bug).  I'm not concerned about listen() and allocateTarget().  I think it's relatively safe to allow calling them repeatedly.

So I think the question is whether we think it's valuable to prevent double calling unlisten() and releaseTarget() (and whether it's an okay contract to allow double-listen but not double-unlisten).  Let's sync up when you're back!

PR #980, opened 2019-11-14
Authentication IID Migration to FIS

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -223,12 +223,20 @@ private final void doRegistration() {
       PersistedInstallationEntry persistedInstallationEntry =
           persistedInstallation.readPersistedInstallationEntryValue();
 
+      // Default value of iidToken
```
andirayo: you did not change that in this CL, but this method is another example for a try-catch block that covers the entire method.
That smells to me that this method should throw that exception.
Otherwise, can we maybe move the try-catch block around the part that actually can cause the exception?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -223,12 +223,20 @@ private final void doRegistration() {
...
       // New FID needs to be created
       if (persistedInstallationEntry.isNotGenerated()) {
 
         // For a default firebase installation read the existing iid. For other custom firebase
         // installations create a new fid
         String fid = readExistingIidOrCreateFid();
+
+        // For a default firebase installation read the stored star scoped iid token. This token
+        // will be used for authenticating the iid on FIS server.
+        iidToken = iidStore.readToken();
```
andirayo: Please call this instanceIdMigrationAuth (or something similar) and only describe in the comment that this is an IID-Token.
The reason is that semantically this is the authentication secret for the migration step and that is what is important for us.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -243,7 +251,7 @@ private final void doRegistration() {
 
       // FID needs to be registered
       if (persistedInstallationEntry.isUnregistered()) {
-        registerAndSaveFid(persistedInstallationEntry);
+        registerAndSaveFid(persistedInstallationEntry, iidToken);
```
andirayo: does this mean we are always sending the IID-Token even if its not a migration of an Instance-ID?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -243,7 +251,7 @@ private final void doRegistration() {
 
       // FID needs to be registered
       if (persistedInstallationEntry.isUnregistered()) {
-        registerAndSaveFid(persistedInstallationEntry);
```
andirayo: This is regarding the throwing of FirebaseInstallationException above in line 244.
That code is an anti-pattern, because the Exception is thrown and caught by you at the end of the method.
You should never throw an exception that you plan to catch in the same library, because that is misusing exceptions for control flow (i.e. as GOTO statements).
Happy to discuss more in person.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -107,6 +114,11 @@ public InstallationResponse createFirebaseInstallation(
         httpsURLConnection.setRequestMethod("POST");
         httpsURLConnection.setDoOutput(true);
 
+        // Set the iid token header for authenticating the iid migration to FIS.
+        if (iidToken != null) {
+          httpsURLConnection.addRequestProperty(X_ANDROID_IID_MIGRATION_KEY, iidToken);
+        }
```
ankitaj224: @andirayo The check when setting the iid-token header.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: No, just for simplifying the code change I am passing the IID-Token (null or  the token read from iid shared pref) .

When calling FIS createInstallation I check if IID-Token is null, I skip setting the header.
Refer to:
https://github.com/firebase/firebase-android-sdk/pull/980/files#r347046883

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: I need to pass the exceptions to the caller of the SDK API asynchronously. Hence the need to have a try-catch block covering the entire method.

It doesn't let me highlight the block that hasn't changed. I can explain you what I mean by this in-person.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: This can definitely be improved. I agree with you. To contain the scope, I ll send the changes in another PR.

ankitaj224:
> I will finish my review later.
> A similar comment to the other PR:
> I think this PR should have broken up in multiple PRs.
> One that introduces the new method #readToken (no users), one that adds the parameter to the #register method, and one that actually adds the business logic.

Makes sense. I will be extra careful next time to break the PR's correctly. Thanks for patiently reviewing a large PR :)

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
andirayo: I think, the check should be the other way around:
If we do not have an IID-Token and it's an Instance-ID then we don't even need to send the request, haha.
I'm fine with it, I just think the logic is a bit backwards.
You are always sending an IID-Token if available even if the identifier is an FID. Vice versa you might send migration requests for Instance-IDs without IID-Token.
I don't know if this is worth fixing, but I wanted to mention it.
If you don't fix it, please document
a) that this is the case
b) and why you decided to implement it this way

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -314,7 +322,8 @@ private void persistFid(String fid) throws FirebaseInstallationsException {
   }
 
   /** Registers the created Fid with FIS servers and update the shared prefs. */
-  private Void registerAndSaveFid(PersistedInstallationEntry persistedInstallationEntry)
+  private Void registerAndSaveFid(
+      PersistedInstallationEntry persistedInstallationEntry, String iidToken)
```
andirayo: instanceIdMigrationAuth?

Please make a brief comment int he JavaDoc, that the instanceIdMigrationAuth is a wildcard-scoped ("*") IID-Token created by IID/FCM's endpoint /register3.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -38,17 +40,23 @@
   private static final String IID_SHARED_PREFS_NAME = "com.google.android.gms.appid";
   private static final String STORE_KEY_PUB = "|S||P|";
   private static final String STORE_KEY_ID = "|S|id";
+  private static final String STORE_KEY_TOKEN = "|T|";
+  private static final String DEFAULT_SCOPE = "|*";
+  private static final String JSON_TOKEN_KEY = "token";
```
andirayo: Can you please add some comments here:
What do all these values mean? Where do the values come from? What is a scope? Where do you use JSON_TOKEN_KEY and so on...

Example:
// The Instance-ID SDK creates a wildcard-scoped ("*") Registration-IID-Token by default.
// The vertical bar is used to separate information identifying Registrations and IID-Tokens.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+    // Check for an explicit sender id
+    String senderId = app.getOptions().getGcmSenderId();
+    if (senderId != null) {
+      return senderId;
+    }
+    String appId = app.getOptions().getApplicationId();
+    if (!appId.startsWith("1:")) {
+      // Not v1, server should be updated to accept the full app ID now
+      return appId;
+    } else {
```
andirayo: please remove this else clause it does not add anything.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+      // For v1 app IDs, fall back to parsing the project ID out
```
andirayo: a) I don't follow: v2 App-IDs also contain the Project-Number that can be parsed.
b) Do you mean Project-ID or Project-Number here?

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+  private static String getDefaultSenderId(FirebaseApp app) {
+    // Check for an explicit sender id
+    String senderId = app.getOptions().getGcmSenderId();
+    if (senderId != null) {
+      return senderId;
+    }
+    String appId = app.getOptions().getApplicationId();
+    if (!appId.startsWith("1:")) {
+      // Not v1, server should be updated to accept the full app ID now
+      return appId;
```
andirayo: I don't understand this clause.
Why should the App-ID start with something different than ":1"?
What does this comment mean? "server should be updated to accept the full app ID now"

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+    } else {
+      // For v1 app IDs, fall back to parsing the project ID out
+      @SuppressWarnings("StringSplitter")
+      String[] parts = appId.split(":");
+      if (parts.length < 2) {
```
andirayo: Why 2? An App-ID has 4 parts.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+        return null; // Invalid format
+      }
+      String projectId = parts[1];
```
andirayo: This is NOT a Project-ID, this is a Project-Number.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+      if (projectId.isEmpty()) {
+        return null; // No project ID
+      }
+      return projectId;
+    }
+  }
+
+  private String createTokenKey(@NonNull String audience) {
```
andirayo: what is an "audience" here? Could you please add an explanatory @param Javadoc?

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
andirayo: I have a feeling that you want to create some constant "TOKEN_STORAGE_DELIMITER" or something and add that String between all these constant values.
Maybe also add a constant "TOKEN_STORAGE_KEY_FORMAT", something like this:
"%s|%s|%s" or whatever the right format is and then later fill the placeholders with the actual values.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+
+  @Nullable
+  public String readToken() {
+    synchronized (iidPrefs) {
+      String token = iidPrefs.getString(createTokenKey(defaultSenderId), null);
+      if (token.isEmpty()) {
+        return null;
+      }
+
+      if (token.startsWith("{")) {
```
andirayo: constant
actually this should be outsourced in a method #isJsonEncoded

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -132,4 +140,55 @@ private PublicKey parseKey(String base64PublicKey) {
...
+        // Encoded as JSON
+        try {
+          JSONObject json = new JSONObject(token);
+          return json.getString(JSON_TOKEN_KEY);
+        } catch (JSONException e) {
+          return null;
```
andirayo: this should be a separate method: #parseIidTokenFromJson

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -61,6 +62,7 @@
 
   private static final String X_ANDROID_PACKAGE_HEADER_KEY = "X-Android-Package";
   private static final String X_ANDROID_CERT_HEADER_KEY = "X-Android-Cert";
+  private static final String X_ANDROID_IID_MIGRATION_KEY = "X-Goog-Fis-Android-Iid-Migration-Auth";
```
andirayo: I kinda like the lowercase version, but this is ok.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -61,6 +62,7 @@
 
   private static final String X_ANDROID_PACKAGE_HEADER_KEY = "X-Android-Package";
   private static final String X_ANDROID_CERT_HEADER_KEY = "X-Android-Cert";
```
andirayo: just because I see this here:
that's great.
have you already implemented sending the App restrictions header?

ankitaj224: Yes, I have

PR #991, opened 2019-11-20
Add ServerValue.increment() methods.

inlined:
* New (currently) package protected method for serverside increments
* Updated ServerValues.resolveFoo to take an existing snapshot
* Added integration tests for new behavior

NOTE: I think I need help undertsanding why I see an integration failure. If my new
tests run then other tests (In EventTest an KeepSyncedTest) fail. It's not clear if
this is because I temporarily go offline or if it's because I've filled the server
queue with an operation the database doesn't yet support and the client cannot handle
the error.

### DIFF
--- a/firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
+++ b/firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
@@ -24,10 +24,10 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import androidx.test.runner.AndroidJUnit4;
-
 import com.google.firebase.database.core.DatabaseConfig;
 import com.google.firebase.database.core.Path;
 import com.google.firebase.database.core.RepoManager;
+import com.google.firebase.database.core.ServerValues;
 import com.google.firebase.database.core.view.Event;
 import com.google.firebase.database.future.ReadFuture;
 import com.google.firebase.database.future.WriteFuture;
@@ -45,7 +45,6 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
-
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Rule;
@@ -1262,7 +1261,9 @@ public class DataTest {
             new MapBuilder().put("/a", "t").put("a/b", "t").build(),
             new MapBuilder().put("/a/b", "t").put("a", "t").build(),
             new MapBuilder().put("/a/b", "t").put("/a/b/.priority", 1.0).build(),
-            new MapBuilder().put("/a/b/.sv", "timestamp").build(),
+            new MapBuilder()
+                .put("/a/b/" + ServerValues.NAME_SUBKEY_SERVERVALUE, ServerValues.NAME_OP_TIMESTAMP)
+                .build(),
             new MapBuilder().put("/a/b/.value", "t").build(),
             new MapBuilder().put("/a/b/.priority", new MapBuilder().put("x", "y").build()).build());
     for (Map<String, Object> badUpdate : badUpdates) {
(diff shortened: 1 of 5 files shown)


firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
```diff
@@ -2740,6 +2744,186 @@ public void onComplete(DatabaseError error, boolean committed, DataSnapshot curr
         < 2000);
   }
 
+  @Test
+  public void testServerIncrementOverwritesExistingData() throws DatabaseException, TimeoutException, InterruptedException {
```
schmidt-sebastian: This code doesn't seem to be formatted correctly. There is a Gradle command you can run:

`./gradlew :firbase-database:googleJavaFormat`

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
```diff
@@ -2740,6 +2744,186 @@ public void onComplete(DatabaseError error, boolean committed, DataSnapshot curr
...
+      latch.bulkRegister(2);
+      ref.setValue(Long.MIN_VALUE + 1);
+      ref.setValue(ServerValue.increment(-2));
+      expectedValues.add(Long.MIN_VALUE + 1);
+      expectedValues.add((double)Long.MIN_VALUE - 1.0);
+
+      latch.awaitAdvanceInterruptibly(0, IntegrationTestValues.getTimeout(), MILLISECONDS);
+
+      // Checking types first makes failures much more obvious
+      List<Class> expectedTypes = expectedValues.stream().map(Object::getClass).collect(toList());
```
schmidt-sebastian: I don't think you can use the streaming APIs, since we can only use Java 7 APIs.

firebase-database/src/main/java/com/google/firebase/database/ServerValue.java
```diff
@@ -30,11 +32,49 @@
...
+   *
+   * This must be an long or a double value. If the ref is not an integer or double, or if the ref
+   * does not yet exist, the transformation will set the ref to the given value. If either of the
+   * given value or the current field value are doubles, both values will be interpreted as doubles.
+   * Double arithmetic and representation of double values follow IEEE 754 semantics. If there is
+   * positive/negative integer overflow, the field is resolved to a double.
+   * @param delta the amount to modify the ref's current value atomically.
+   * @return a placeholder value for modifying a ref atomically server-side.
+   */
+  static final Map<String, Map<String, Object>> increment(long delta) {
```
schmidt-sebastian: Is this return type used anywhere? The type seems very strange.

Would it be possible to return an Object here? You could create an private subclass that you can use internally to allow for `instanceof` checks.

firebase-database/src/main/java/com/google/firebase/database/ServerValue.java
```diff
@@ -30,11 +32,49 @@
...
+  /**
+   * Adds the given delta to the ref's current value.
+   *
+   * This must be an long or a double value. If the ref is not an integer or double, or if the ref
+   * does not yet exist, the transformation will set the ref to the given value. If either of the
+   * given value or the current field value are doubles, both values will be interpreted as doubles.
+   * Double arithmetic and representation of double values follow IEEE 754 semantics. If there is
+   * positive/negative integer overflow, the field is resolved to a double.
+   * @param delta the amount to modify the ref's current value atomically.
+   * @return a placeholder value for modifying a ref atomically server-side.
```
schmidt-sebastian: Please update with changes from iOS.

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -15,11 +15,8 @@
 package com.google.firebase.database.core;
 
 import com.google.firebase.database.core.utilities.Clock;
-import com.google.firebase.database.snapshot.ChildKey;
-import com.google.firebase.database.snapshot.ChildrenNode;
-import com.google.firebase.database.snapshot.Node;
-import com.google.firebase.database.snapshot.NodeUtilities;
-import com.google.firebase.database.snapshot.PriorityUtilities;
+import com.google.firebase.database.snapshot.*;
```
schmidt-sebastian: Please make sure to import every class explicitly.

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -32,47 +29,81 @@
...
+  static Object resolveScalarDeferredValue(String op, Node existing, Map<String, Object> serverValues) {
+      if ("timestamp".equals(op) && serverValues.containsKey(op)) {
+        return serverValues.get(op);
       }
+      return null;
+  }
+
+  static Object resolveComplexDeferredValue(Map<String, Object> op, Node existing, Map<String, Object> serverValues) {
+    // Only supported complex op so far
+    if (!op.containsKey("increment")) {
```
schmidt-sebastian: Nit: Could we define constants and share them between core/ServerValues.java and the public ServerValues.java? I am going to guess that this leaks into the public API (which we don't want to do), but it doesn't hurt to ask :)

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -32,47 +29,81 @@
...
+    }
+
+    Number increment = (Number)incrObject;
+
+    if (!(existing.isLeafNode() && existing.getValue() instanceof Number)) {
+      return increment;
+    }
+
+    Number existingVal = (Number) existing.getValue();
+    if (longCanHold(increment) && longCanHold(existingVal)) {
```
schmidt-sebastian: Same comment as in iOS port regarding naming.

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -32,47 +29,81 @@
...
+      return overflowAwareAdd(increment.longValue(), existingVal.longValue());
```
schmidt-sebastian: Please inline.

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -101,13 +132,30 @@ public void visitChild(ChildKey name, Node child) {
...
           write.addWrite(
-              entry.getKey(), resolveDeferredValueSnapshot(entry.getValue(), serverValues));
+              entry.getKey(), resolveDeferredValueSnapshot(entry.getValue(), existing.getChild(entry.getKey()), serverValues));
     }
     return write;
   }
+
+  // Ignoring types we should never see: BigFoo, AtomicFoo, FooAdder, and FooAccumulator
+  private static boolean longCanHold(Number x) {
+    return !(x instanceof Double || x instanceof Float);
```
schmidt-sebastian: Nit: Move comment into method.

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
inlined: k

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
inlined: Even in unit tests?

firebase-database/src/main/java/com/google/firebase/database/ServerValue.java
inlined: This was intended for consistency with TIMESTAMP which is also a transparent return type. Can change.

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
inlined: Done

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
inlined: done

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
inlined: done

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
inlined: done

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
inlined: done

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
schmidt-sebastian: I asked in the Android Contrib Chatroom for clarification:

- We can theoretically use Java 8 APIs in unit tests.
- We should not use them in Integration Tests, as they run on device. We currently run our tests only on Pixel 2, which is why they pass with this API reference. Ideally, we should also run them on a device that only meets the minimum requirements (see https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-runtime/transport-runtime.gradle#L21-L34 for an example of how we can configure that).

`DataTest` is an integration test (it's under androidTest and runs on device). We should therefore rewrite this part. I also think our codebase will be a little saner if we just avoid using Java 7 APIs consistently

firebase-database/src/main/java/com/google/firebase/database/ServerValue.java
schmidt-sebastian: I would prefer is this just returned `Object`. `Map<String, String>` is still a little more meaningful that `Map< String, Map<String, Object>>`

inlined: Done. The helper type is still a Map, but the public methods return Object.

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
inlined: Fixed

firebase-database/src/main/java/com/google/firebase/database/ServerValue.java
inlined: Done

inlined:
Looks like I'm hitting a quota error.
/retest

PR #996, opened 2019-11-22
Implement FirebaseABTesting.getAllExperiments() to surface ABT experiments for a given origin service

welkinlan:
- Refactor FirebaseABTesting unit test.
- Implement functionalities for converting ConditionUserProperty from/to AbtExperimentInfo.
- Format code according to Google java code style.


firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
```diff
@@ -140,6 +150,27 @@ public void removeAllExperiments() throws AbtException {
     removeExperiments(getAllExperimentsInAnalytics());
   }
 
+  /**
+   * Gets the origin service's list of experiments in the app.
+   *
+   * <p>Note: This is a blocking call and therefore should be called from a worker thread.
```
JasonAHeron: is this a blocking call? should it be annotated with @WorkerThread?

welkinlan: great call! updated the @workerthread annotation

PR #998, opened 2019-11-22
Add support for bytes in Encoders

rlazo:


encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
```diff
@@ -78,6 +79,14 @@ public JsonValueObjectEncoderContext add(@NonNull String name, boolean value)
     return add(value);
   }
 
+  @NonNull
+  @Override
+  public ObjectEncoderContext add(@NonNull String name, byte[] value)
```
vkryachko: we already have primitive array handling? why not add byte[] logic there?

rlazo: Byte arrays are encoded as strings, not as arrays, so I though that expose this was useful. Now that I think about it, it's better to have a simpler API. Merged.

PR #1000, opened 2019-11-24
Support for JSON based LogResponse

rlazo:
Optimally, the code to parse the JSON would use a similar mechanism to the one found in `encoders`, but for the time being we can get away with manual parsing.


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
```diff
@@ -0,0 +1,78 @@
...
+public abstract class LogResponse {
+  /** Client should wait for next_request_wait_millis before sending the next log request. */
+  public abstract long getNextRequestAwaitMillis();
+
+  /**
+   * Quality of Service tiers enforced by server for overriding client qos_tier setting in event
+   * logging.
+   */
+  @Nullable
+  public abstract QosTiersOverride getQosTiersOverride();
```
vkryachko: We don't need these, the only thing that we implement at this point is `next_request_wait_millis`, so it's worth removing the getter and the classes

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
```diff
@@ -0,0 +1,78 @@
...
+import android.util.JsonToken;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+import java.io.IOException;
+import java.io.StringReader;
+
+@AutoValue
+public abstract class LogResponse {
+  /** Client should wait for next_request_wait_millis before sending the next log request. */
+  public abstract long getNextRequestAwaitMillis();
```
vkryachko:
```suggestion
  public abstract long getNextRequestWaitMillis();
```

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
rlazo: Done. PTAL

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
```diff
@@ -0,0 +1,59 @@
...
+public abstract class LogResponse {
+  /** Client should wait for next_request_wait_millis before sending the next log request. */
+  public abstract long getNextRequestWaitMillis();
+
+  static LogResponse create(long nextRequestWaitMillis) {
+    return new AutoValue_LogResponse(nextRequestWaitMillis);
+  }
+
+  @Nullable
+  public static LogResponse fromJson(@Nullable String input) {
```
vkryachko: Since the plan is to read it off network, can this take `InputStream`/`InputStreamReader`/ instead and propagate IOException?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
```diff
@@ -0,0 +1,59 @@
...
+    if (input == null) {
+      return null;
+    }
+
+    JsonReader jsonReader = new JsonReader(new StringReader(input));
+    try {
+      jsonReader.beginObject();
+      while (jsonReader.hasNext()) {
+        String name = jsonReader.nextName();
+        if (name.equals("next_request_wait_millis")) {
```
vkryachko: to confirm, but iiuc the name in the response will be `nextRequestWaitMillis`

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Nullable
+  public static LogResponse fromJson(@Nullable String input) {
+    if (input == null) {
+      return null;
+    }
+
+    JsonReader jsonReader = new JsonReader(new StringReader(input));
+    try {
+      jsonReader.beginObject();
```
vkryachko: If a generic 404/403/502 html page is returned iiuc this will throw `IllegalStateException`, should we catch it and return null?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
rlazo: Done

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
rlazo: Done.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogResponse.java
rlazo: Done.

PR #1004, opened 2019-11-25
Add support for AutoValue in encoders.

rlazo:


encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
```diff
@@ -156,14 +156,23 @@ JsonValueObjectEncoderContext add(@Nullable Object o) throws IOException, Encodi
...
+      objectEncoders.put(o.getClass(), objectEncoders.get(o.getClass().getSuperclass()));
+      return add(o);
     }
     @SuppressWarnings("unchecked") // safe because get the encoder by checking the object's type.
     ValueEncoder<Object> valueEncoder = (ValueEncoder<Object>) valueEncoders.get(o.getClass());
     if (valueEncoder != null) {
       valueEncoder.encode(o, this);
       return this;
+      // Check for AutoValue concrete class. The encoder for the abstract class will work for it.
+    } else if (o.getClass().getSimpleName().startsWith("AutoValue_")
```
vkryachko: I have mixed feeling about this:
1. This uses reflection, hence can be slow.
2. This won't work with proguard enabled as the name will be obfuscated to being unrecognizable.

I wonder if it's best to have users register the `AutoValue_` classes explicitly until we come up with another solution for it? wdyt

rlazo: 1. Yes, reflection is slow, but it's a one time cost. We are registering the new pair the first time we see it, so this shouldn't cause too much issue.
2. That's a harder problem to fix. An option is to do this regardless of the name. Because not finding an encoder throws an exception, a non-autovalue class would still fail.

Unless we expect the `DataEncoder` to be re-created several times, and handling a lot of different classes, I think we can move forward with this approach.

vkryachko: Makes sense, I think we should not check for class name, a few considerations:

* Mutating encoder sets during serialization has thread-safety implications for the encoder. All accesses to the sets must be synchronized now, and since these sets are shared among all contexts, this will decrease performance of parallel serialization. So this is something to consider carefully and preferably try to avoid mutation completely(by trying to do it upfront if possible?)
* Corner case: what if I register an objectEncoder for `base` class, and a value encoder for `derived`? With the current implementation it looks like it will not work as "expected"
* If we choose to support encoding `derived` via a `base` encoder, should we walk the hierarchy all the way up(excluding `java.lang.Object`), not just one level up?

Given all of the above, it seems that some design work is needed here, wdyt?

rlazo: Makes sense, let's design this more carefully.

PR #1005, opened 2019-11-25
Add encoding support multiple types

vkryachko:
* primitive longs
* Enums with ability to register custom encoders if desired
* All java.lang.Number subclasses

Note that longs and all Numbers are represented as numeric values, not
strings. While this could be a problem when parsed with a JavaScript parser,
it is not an issue for us since JS is not involved on the client or the
server.

### DIFF
--- a/encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
+++ b/encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
@@ -73,6 +73,14 @@ final class JsonValueObjectEncoderContext implements ObjectEncoderContext, Value
 
   @NonNull
   @Override
+  public JsonValueObjectEncoderContext add(@NonNull String name, long value)
+      throws IOException, EncodingException {
+    jsonWriter.name(name);
+    return add(value);
+  }
+
+  @NonNull
+  @Override
   public JsonValueObjectEncoderContext add(@NonNull String name, boolean value)
       throws IOException, EncodingException {
     jsonWriter.name(name);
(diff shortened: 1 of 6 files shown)


encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
```diff
@@ -101,6 +109,13 @@ public JsonValueObjectEncoderContext add(int value) throws IOException, Encoding
     return this;
   }
 
+  @NonNull
+  @Override
+  public JsonValueObjectEncoderContext add(long value) throws IOException, EncodingException {
+    jsonWriter.value(value);
```
rlazo: I think we want long values to be encoded as strings always.

encoders/firebase-encoders/src/test/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContextTest.java
```diff
@@ -60,10 +60,10 @@ public void testEncodingPrimitiveTypes() throws IOException, EncodingException {
     ObjectEncoder<DummyClass> objectEncoder =
         (o, ctx) -> {
           ctx.add("String", "string")
-              .add("Integer", 2)
-              .add("Double", 2.2d)
-              .add("Boolean", false)
-              .add("Null", null);
+              .add("int", 2)
+              .add("double", 2.2d)
+              .add("boolean", false)
```
rlazo: Add long here too.

encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
vkryachko: I don't think we have to, see my PR comment above. In the future we may have to do it, but we can have explicit logic for only encoding values `> 2^52-1` as strings. wdyt?

rlazo: sgtm

encoders/firebase-encoders/src/test/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContextTest.java
vkryachko: Done

PR #1014, opened 2019-11-27
Encode json payload as string instead of byte[]

rlazo:


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogEvent.java
```diff
@@ -33,7 +33,7 @@
 
   @SuppressWarnings("mutable")
   @Nullable
-  public abstract byte[] getSourceExtensionJsonProto3Bytes();
+  public abstract String getSourceExtensionJsonProto3Bytes();
```
vkryachko: Not: remove the Bytes suffix? Here and in the setter. Lgtm otherwise

PR #1017, opened 2019-11-28
Add `Map<String, Object>` support to JsonEncoder.

vkryachko:


encoders/firebase-encoders/src/test/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContextTest.java
```diff
@@ -25,12 +26,12 @@
 import com.google.firebase.encoders.EncodingException;
 import com.google.firebase.encoders.ObjectEncoder;
 import com.google.firebase.encoders.ValueEncoder;
+import edu.emory.mathcs.backport.java.util.Collections;
```
rlazo: Is this the right import?

encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
```diff
@@ -192,6 +192,25 @@ JsonValueObjectEncoderContext add(@Nullable Object o) throws IOException, Encodi
...
+      @SuppressWarnings("unchecked")
+      Map<Object, Object> map = (Map<Object, Object>) o;
+      jsonWriter.beginObject();
+      for (Map.Entry<Object, Object> entry : map.entrySet()) {
+        Object key = entry.getKey();
+        try {
+          add((String) key, entry.getValue());
+        } catch (ClassCastException ex) {
+          throw new EncodingException(
+              String.format(
```
rlazo: Rephrase as: "Only String keys are currently supported" to suggest that, while there's a way to encode other types to be map keys (https://developers.google.com/protocol-buffers/docs/proto3#json) it's not supported.

PR #1018, opened 2019-11-28
Implement encoder context nesting.

vkryachko:


encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/ObjectEncoderContext.java
```diff
@@ -68,4 +68,28 @@ ObjectEncoderContext add(@NonNull String name, double value)
...
+   * ctx.add("key", "value");
+   * ObjectEncoderContext nested = ctx.nested("nested");
+   * nested.add("key", "value");
+   *
+   * // After this call the above nested context is invalid.
+   * ctx.add("anotherKey", 1);
+   * }</pre>
+   */
+  @NonNull
+  ObjectEncoderContext nested(@NonNull String name) throws IOException;
```
rlazo: if `nested` received a "runnable" as second parameter we could do something like

```
ctx.add("key", "value")
   .nested("key", (nctx) -> {
        nctx.add("nested_key", "nested_value")})
    .add("second_key", "second_value"));
````
we could get rid of keeping track of whether there's a nested context to close or not. wdyt?

vkryachko: I generally like your interface much better as it makes it "impossible to use the type incorrectly (c)".

However I think we considered it at some point but the problem with lambdas is that they will result in class bloat which will count towards dex limit(as each lambda is effectively an anonymous class), whereas returning a nested context does not have this problem, so no matter how much I "nest" I end up with just one class(encoder that I am implementing itself). I am not certain it is a big deal though

rlazo: You're right, I forgot we discussed this earlier. Let's run with this and see how it goes :-).

PR #1019, opened 2019-11-29
Add polymorphic encoder registration support.

vkryachko:
This allows clients to configure DataEncoders in a way that does not
couple them to a particular output format, e.g. json.

This mechanism will be used for auto-generated encoders in the future.

### DIFF
--- a/encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonDataEncoderBuilder.java
+++ b/encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonDataEncoderBuilder.java
@@ -21,6 +21,8 @@ import com.google.firebase.encoders.EncodingException;
 import com.google.firebase.encoders.ObjectEncoder;
 import com.google.firebase.encoders.ValueEncoder;
 import com.google.firebase.encoders.ValueEncoderContext;
+import com.google.firebase.encoders.config.Configurator;
+import com.google.firebase.encoders.config.EncoderConfig;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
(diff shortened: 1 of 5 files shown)


encoders/firebase-encoders/src/main/java/com/google/firebase/encoders/json/JsonDataEncoderBuilder.java
```diff
@@ -32,7 +34,7 @@
 import java.util.Map;
 import java.util.TimeZone;
 
-public final class JsonDataEncoderBuilder {
+public final class JsonDataEncoderBuilder implements EncoderConfig<JsonDataEncoderBuilder> {
```
rlazo: I'm not sure this class **is** a configuration, but rather, it **has** a configuration. We can encapsulate the object & value encoders below into a configuration object and pass that to the configurator instead of the whole builder being a config.

rlazo: wdyt?

vkryachko: > I'm not sure this class is a configuration, but rather, it has a configuration

I guess one could argue either way but in my view when I use a builder to create an object I am configuring/describing the properties I want the object to have, so in this sense I feel like builder is a configuration.

I guess we could have the builder contain a config object inside, but it seems that then the builder will not do anything other than delegate to the contained config object, i.e. it will have `registerEncoder` and `configureWith` methods which will just call into the config, which will effectively mean that the builder has all the methods that `EncoderConfig` requires but does not implement `EncoderConfig` itself.

rlazo: I think the builder would need to receive the `EncoderConfig` object the client would generate itself. The Builder wouldn't need to expose the methods itself. That being said I'm not strongly attached to that design. I'll approve the PR and you can go ahead with the changes.

PR #1022, opened 2019-12-02
Integrate with codecov.io to display coverage changes on pull requests.

yifanyang:


tools/measurement/coverage/coverage.gradle
```diff
@@ -33,3 +33,48 @@ task uploadCoverageMeasurements(type: UploadMeasurementsTask) {
...
+    environment "VCS_COMMIT_ID", "$commit"
+    environment "VCS_BRANCH_NAME", "$branch"
+    environment "VCS_PULL_REQUEST", "$pr"
+    environment "VCS_SLUG", "firebase/firebase-android-sdk"
+    environment "VCS_TAG", ""
+    environment "CI_BUILD_URL", ""
+    environment "CI_BUILD_ID", "$build"
+    environment "CI_JOB_ID", "$job"
+
+    commandLine 'bash', '-c', 'bash <(curl -s https://codecov.io/bash)'
```
rlazo: I know this is the [recommended way](https://docs.codecov.io/v4.3.0/docs/about-the-codecov-bash-uploader) from codecov.io but I'm not 100% sure about running code from the internet without verifying it first. @vkryachko wdyt?

yifanyang:
In order to verify this actually works, I manually refreshed prow config (which is restored now) with content in this [PR](https://team-review.git.corp.google.com/c/firebase-android-core/test-infra/+/610474), and opened another test [PR](https://github.com/firebase/firebase-android-sdk/pull/1023) that has the same diff as this one. Summary of coverage and links to detailed report have been posted to that test PR. Looks like it is able to work.

tools/measurement/coverage/coverage.gradle
yifanyang: Updated the PR. Planning to save a copy of the bash uploader and only executes from this local copy every time. Will this be an acceptable approach?

yifanyang: Making the bash uploader part of the CI image ([PR](https://team-review.git.corp.google.com/c/firebase-android-core/test-infra/+/611942)), and invoke from there.

PR #1029, opened 2019-12-03
Adds code generation via `@Encodable` annotation.

vkryachko:
Supported features:

* Codegen is based on public java-bean style getters, fields are not encoded
  directly
* Full support for generics
* Supports `java.util.Optional` via `value.getOrElse(null)`
* Fields can be marked as `@Encodable.Ignore` to skip them during encoding
* Fields can be marked as `@Encodable.Field(name="foo")` for custom encoded name

### DIFF
--- a/encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/annotations/Encodable.java
+++ b/encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/annotations/Encodable.java
@@ -1,0 +1,58 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.encoders.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates to automatically generate an encoder capable to encode a given type.
+ *
+ * <p>All fields that need to be encoded <strong>must</strong>must have Java-bean compliant getters,
+ * otherwise they will not be encoded.
+ *
+ * Using the generated encoder:
+ * <pre>{@code
+ * @Encodable
+ * public class MyType {
+ *   public String getField() { return "hello"; }
+ * }
+ *
+ * DataEncoder encoder = new JsonDataEncoderBuilder()
+ *     .configureWith(AutoMyTypeEncoder.CONFIG)
+ *     .build();
+ *
+ * String json = encoder.encode(new MyType);
+ *
+ * }<pre>
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface Encodable {
+
+  /** Specifies a custom field name for a given property of a type. */
+  @Target(ElementType.METHOD)
+  @Retention(RetentionPolicy.CLASS)
+  @interface Field {
+    String name() default "";
+  }
+
+  /** Indicates the code generator to ignore a given property of a type. */
+  @Target(ElementType.METHOD)
+  @Retention(RetentionPolicy.CLASS)
+  @interface Ignore {}
+}
--- a/encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/EncodableProcessor.java
+++ b/encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/EncodableProcessor.java
@@ -1,0 +1,205 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.encoders.processor;
+
+import com.google.auto.service.AutoService;
+import com.google.firebase.encoders.annotations.Encodable;
+import com.google.firebase.encoders.processor.getters.Getter;
+import com.google.firebase.encoders.processor.getters.GetterFactory;
+import com.squareup.javapoet.AnnotationSpec;
+import com.squareup.javapoet.ClassName;
+import com.squareup.javapoet.FieldSpec;
+import com.squareup.javapoet.JavaFile;
+import com.squareup.javapoet.MethodSpec;
+import com.squareup.javapoet.ParameterizedTypeName;
+import com.squareup.javapoet.TypeName;
+import com.squareup.javapoet.TypeSpec;
+import com.squareup.javapoet.WildcardTypeName;
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.processing.AbstractProcessor;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.annotation.processing.Processor;
+import javax.annotation.processing.RoundEnvironment;
+import javax.annotation.processing.SupportedAnnotationTypes;
+import javax.annotation.processing.SupportedSourceVersion;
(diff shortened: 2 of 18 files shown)


encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/Names.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Returns the class name that corresponds to a given element.
+   *
+   * <p>For top level classes returns the class name unchanged. For nested classes returns {@code
+   * ParentNested} for {@code Parent$Nested} classes.
+   */
+  static String generatedClassName(Element element) {
+    StringBuilder sb = new StringBuilder(element.getSimpleName().toString());
+    Element enclosingElement = element.getEnclosingElement();
+    while (!(enclosingElement instanceof PackageElement)) {
```
rlazo: Could you add a separator like AutoValue's `_`?

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/VisitResult.java
```diff
@@ -0,0 +1,41 @@
...
+
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+import java.util.Collections;
+import java.util.Set;
+import javax.lang.model.type.TypeMirror;
+
+@AutoValue
+abstract class VisitResult<T> {
+  abstract Set<TypeMirror> moreToVisit();
```
rlazo: what about `pendingToVisit`

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
```diff
@@ -0,0 +1,169 @@
...
+    TypeMirror map = types.erasure(elements.getTypeElement("java.util.Map").asType());
+    if (types.isAssignable(returnType, map)) {
+      TypeMirror keyType = ((DeclaredType) returnType).getTypeArguments().get(0);
+      if (!types.isSameType(keyType, elements.getTypeElement("java.lang.String").asType())) {
+        messager.printMessage(
+            Diagnostic.Kind.ERROR,
+            "Cannot encode Maps with non-String keys.",
+            ((DeclaredType) returnType).asElement());
+      }
+    }
```
rlazo: Wouldn't it be better to leave this on the encoders? If we ever support more than String for keys, we would have to remember to change 2 places.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
```diff
@@ -0,0 +1,169 @@
...
+    if (types.isAssignable(
+        returnType, types.erasure(elements.getTypeElement("java.util.Optional").asType()))) {
+      returnType = ((DeclaredType) returnType).getTypeArguments().get(0);
+      getterExpression = getterExpression + ".orElse(null)";
+    }
```
rlazo: How likely is that we need to support optional? I though it wasn't available in android in anything less than API24

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
```diff
@@ -0,0 +1,169 @@
...
+
+    return Optional.of(Getter.create(fieldName.get(), getterExpression, returnType));
+  }
+
+  // ownerType = Bar<Optional<Integer>, Boolean>
+  // ownerElement = Bar<T, U>
+  // genericType = U
+  private TypeMirror resolveTypeArguments(DeclaredType ownerType, TypeMirror genericType) {
```
rlazo: comment?

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
```diff
@@ -0,0 +1,169 @@
...
+        && methodName.length() != 2
+        && method.getReturnType().getKind() == TypeKind.BOOLEAN) {
+      return Optional.of(Character.toLowerCase(methodName.charAt(2)) + methodName.substring(3));
+    }
+
+    if (methodName.startsWith("get")
+        && methodName.length() != 3
+        && method.getReturnType().getKind() != TypeKind.VOID) {
+      return Optional.of(Character.toLowerCase(methodName.charAt(3)) + methodName.substring(4));
+    }
```
rlazo: This is perhaps something good to note in the comments.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/EncodableProcessor.java
```diff
@@ -0,0 +1,197 @@
...
+    Set<Encoder> encoders = TypeTraversal.traverse(element.asType(), new GetterVisitor());
+
+    MethodSpec.Builder configureMethod =
+        MethodSpec.methodBuilder("configure")
+            .addParameter(
+                ParameterizedTypeName.get(
+                    ClassName.get("com.google.firebase.encoders.config", "EncoderConfig"),
+                    WildcardTypeName.subtypeOf(Object.class)),
+                "cfg")
+            .addModifiers(Modifier.PUBLIC)
```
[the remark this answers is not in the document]
vkryachko: afaict the lint runs on source only, but added a hide for completeness.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/Names.java
vkryachko: I am not a fan of `_`, also fwiw dagger2 does not use them either, so unless you feel strongly I'd prefer not to have them.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
vkryachko: I think until we do, it's better DevX to fail at compile time than at runtime, when/if we add support for non-strings we can always just remove this code, wdyt?

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
vkryachko: It's available in google3 and will be supported by agp 4 desugaring.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
vkryachko: removed, just notes I used when implementing this thing :)

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
vkryachko: Good point, updated `@Encodable`s javadoc  to address it.

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/annotations/Encodable.java
```diff
@@ -0,0 +1,58 @@
...
+ * Indicates to automatically generate an encoder capable to encode a given type.
+ *
+ * <p>All fields that need to be encoded <strong>must</strong>must have Java-bean compliant getters,
+ * otherwise they will not be encoded.
+ *
+ * Using the generated encoder:
+ * <pre>{@code
+ * @Encodable
+ * public class MyType {
+ *   public String getField() { return "hello"; }
```
rlazo:
```suggestion
 *   public String getField() { return "hello"; }
 *   public boolean isConditional() { return false; }
```

rlazo: Just for completeness.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
rlazo: Sure, let's add a comment to point that out then.

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
rlazo: lgtm

encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/getters/GetterFactory.java
rlazo: Thanks!

root-project.gradle
```diff
@@ -80,15 +80,15 @@ configure(subprojects) {
     }
     tasks.googleJavaFormat {
         source '.'
-        include '**/*.java'
+        include '**/java/**/*.java'
```
rlazo: why is this needed?

rlazo: nvm I saw the changes in the other PR

PR #1030, opened 2019-12-03
Set firebase-database to -Werror

inlined:
As a bar-raising experiment, I added -Werror to our compilation options. This broke the build; fixes are also included in this change. With this update, compiler warnings (most importantly, errorprone warnings) will not be allowed into future pull requests. Hopefully this will catch a bug or two before our customers do.


firebase-database/firebase-database.gradle
```diff
@@ -84,14 +84,14 @@ dependencies {
...
     androidTestImplementation 'androidx.test:runner:1.2.0'
+    androidTestImplementation 'androidx.test.ext:junit:1.1.1'
     androidTestImplementation "com.google.truth:truth:$googleTruthVersion"
     androidTestImplementation 'com.fasterxml.jackson.core:jackson-core:2.9.8'
     androidTestImplementation 'com.fasterxml.jackson.core:jackson-databind:2.9.8'
     androidTestImplementation 'junit:junit:4.12'
+    androidTestImplementation 'org.hamcrest:hamcrest:2.2'
+    androidTestImplementation 'org.hamcrest:hamcrest-library:2.2'
     androidTestImplementation 'net.java:quickcheck:0.6'
-    androidTestAnnotationProcessor 'net.java:quickcheck-src-generator:0.6'
```
inlined: These annotations were not used and caused conflicts because their maximum version is 1.6

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
```diff
@@ -710,7 +712,7 @@ public boolean isComplete(List<EventRecord> events) {
                   try {
                     ref.child("two").removeValue();
                   } catch (DatabaseException e) {
-                    fail("Should not fail");
+                    throw new AssertionError("Should not fail", e);
```
inlined: errorprone calls catch-then-fail an anti-pattern. AssertionError is an unchecked exception that can wrap any other throwable to preserve cause and stack traces.

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
```diff
@@ -2346,9 +2348,8 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
     assertEquals(snap.getPriority().getClass(), Double.class);
     assertEquals(snap.getPriority(), snap.child("b").getPriority());
     assertEquals(snap.child("a").getValue(), snap.child("b").getValue());
-    assert (Math.abs(
```
inlined: assert is a production method, not a test method

firebase-database/src/androidTest/java/com/google/firebase/database/EventTest.java
```diff
@@ -623,15 +626,15 @@ public void onDataChange(DataSnapshot snapshot) {
               try {
                 IntegrationTestHelpers.waitFor(blockSem);
               } catch (InterruptedException e) {
-                e.printStackTrace();
+                LOGGER.log(WARNING, "unexpected error", e);
```
inlined: I'll admit I don't fully understand why `printStackTrace()` is considered an anti-pattern. I'm assuming it's because we should be printing to logcat rather than stdout.

firebase-database/src/androidTest/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -20,8 +20,8 @@
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.fail;
 
-import androidx.test.InstrumentationRegistry;
-import androidx.test.runner.AndroidJUnit4;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
```
inlined: These libraries moved

firebase-database/src/androidTest/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -20,8 +20,8 @@
...
+import androidx.test.platform.app.InstrumentationRegistry;
```
inlined: This library has a breaking change; you now need to access a singleton with `getInstrumentation()` before getting contexts.

firebase-database/src/androidTest/java/com/google/firebase/database/ObjectMapTest.java
```diff
@@ -61,6 +61,11 @@ public String toString() {
     public boolean equals(Object o) {
       return (o instanceof Author) && ((Author) o).name.equals(name) && ((Author) o).id == id;
     }
+
+    @Override
```
inlined: Classes that override equals must override hashCode

firebase-database/src/androidTest/java/com/google/firebase/database/OrderTest.java
```diff
@@ -170,10 +170,9 @@ public void pushDataWithExponentialPriorityAndCheckOrder()
     DatabaseReference reader = refs.get(1);
 
     for (int i = 0; i < 9; ++i) {
-      writer.push().setValue(i, 111111111111111111111111111111.0 / Math.pow(10, i));
```
inlined: Precision loss in float literals had to be fixed.

firebase-database/src/androidTest/java/com/google/firebase/database/RetryRule.java
```diff
@@ -25,6 +25,7 @@ public RetryRule(int retryCount) {
     this.retryCount = retryCount;
   }
 
+  @Override
```
inlined: Method overrides must be annotated as such.

firebase-database/src/androidTest/java/com/google/firebase/database/TransactionTest.java
```diff
@@ -568,6 +567,7 @@ public void onComplete(
 
     // Note that the set actually raises two events, one overlaid on top of the original transaction
     // value, and a second one with the re-run value from the third transaction
+    assertEquals(nodeFooSnaps.size(), 4); // three transactions with one redo
```
inlined: This variable was being written to but never read from. I added an arbitrary test though we could also just remove its pushes.

firebase-database/src/androidTest/java/com/google/firebase/database/core/AndroidPlatformTest.java
```diff
@@ -30,11 +31,11 @@ public void userAgentHasCorrectParts() {
     Context cfg = IntegrationTestHelpers.getContext(0);
     cfg.freeze();
     String userAgent = cfg.getUserAgent();
-    String[] parts = userAgent.split("/");
+    Object[] parts = Splitter.on('/').splitToList(userAgent).toArray();
```
inlined: errorprone says `String::split` behaves in unexpected ways.

firebase-database/src/androidTest/java/com/google/firebase/database/core/persistence/KeepSyncedTest.java
```diff
@@ -57,12 +57,10 @@ private void assertIsKeptSynced(Query query) throws Exception {
 
     new ReadFuture(
             query,
-            new ReadFuture.CompletionCondition() {
-              public boolean isComplete(List<EventRecord> events) {
```
inlined: `isComplete` needs an `@Override` annotation or we could alternatively just use lambda syntax.

firebase-database/src/androidTest/java/com/google/firebase/database/core/persistence/PersistenceTest.java
```diff
@@ -881,7 +883,8 @@ public void ackForServerValueYieldsResolvedValueInCache() throws Throwable {
     new WriteFuture(ref, ServerValue.TIMESTAMP).timedGet();
 
     Node node = manager.serverCache(query).getNode();
-    assertTrue(Math.abs(System.currentTimeMillis() - (Long) node.getValue()) < 1000);
```
inlined: note: here I changed the maximum drift to 2s to match the rest of the codebase and reduce flakiness

firebase-database/src/androidTest/java/com/google/firebase/database/core/persistence/PruningTest.java
```diff
@@ -52,7 +52,7 @@
   private static final PruneForestBuilder PF = new PruneForestBuilder();
   private TestEnv env;
 
-  private class TestEnv {
+  private static class TestEnv {
```
inlined: private internal classes should be static

firebase-database/src/main/java/com/google/firebase/database/core/CompoundWrite.java
```diff
@@ -31,7 +31,7 @@
  * is only allowed to be one write modifying that path. Any write to an existing path or shadowing
  * an existing path will modify that existing write to reflect the write added.
  */
-public class CompoundWrite implements Iterable<Map.Entry<Path, Node>> {
+public final class CompoundWrite implements Iterable<Map.Entry<Path, Node>> {
```
inlined: `equals` cannot check class equality of the parameter (isa CompoundWrite); they should use an instanceOf check or make the class final.

firebase-database/src/main/java/com/google/firebase/database/core/UserWriteRecord.java
```diff
@@ -88,7 +88,7 @@ public boolean equals(Object o) {
     if (!(this.writeId == record.writeId)) {
       return false;
     }
-    if (!(this.path.equals(record.path))) {
```
inlined: unnecessary parens are warnings & now errors

firebase-database/src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -663,7 +669,7 @@ public static void setValue1(String value1) {
     }
   }
 
-  private static enum Enum {
```
inlined: Name collision was disallowed

firebase-database/src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -1243,8 +1245,7 @@ public void beanMapsMustHaveStringKeys() {
 
   @Test
   public void serializeStringBean() {
-    StringBean bean = new StringBean();
-    bean.value = "foo";
```
inlined: Direct member access was disallowed. I had the choice of adding a setValue or just a constructor that accepted a value.

firebase-database/src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -1380,54 +1380,28 @@ public void serializingListsWorks() {
   @Test
   public void serializingMapsWorks() {
     MapBean bean = new MapBean();
-    bean.values =
```
inlined: Parameter list initialization is disallowed because it leads to memory issues. We can use the ImmutableFoo.of methods in unit tests or the ImmutableFoo builders in code with a lower target JRE version.

firebase-database/src/test/java/com/google/firebase/database/util/ComparatorTester.java
```diff
@@ -151,8 +150,8 @@ private final void doTestEquivalanceGroupOrdering() {
         testClassCast(reference);
         for (int otherIndex = 0; otherIndex < equalityGroups.size(); otherIndex++) {
           for (Object other : equalityGroups.get(otherIndex)) {
-            assertThat(Integer.signum(compare(reference, other)))
-                .named("compare(%s, %s)", reference, other)
```
inlined: The `.named` method is deprecated. We're instead supposed to `assertWithMessage`.

firebase-database/src/test/java/com/google/firebase/database/util/ComparatorTester.java
```diff
@@ -165,11 +164,10 @@ private final void doTestEqualsCompatibility() {
       for (Object reference : referenceGroup) {
         for (List<Object> otherGroup : equalityGroups) {
           for (Object other : otherGroup) {
-            assertWithMessage(
-                    "Testing equals() for compatibility with compare()/compareTo(), "
-                        + "add a call to doNotRequireEqualsCompatibility() if this is not required")
+            // Testing equals() for compatibility with compare()/compareTo(),
```
inlined: Per above, I had to use only assertWithMessage, not `.named` I kept the printf text formerly in `.named` and moved the other message to a comment to avoid losing information.

firebase-database/src/testUtil/java/com/google/firebase/database/DeepEquals.java
```diff
@@ -99,7 +100,7 @@ private static boolean weakTypeMatch(Object a, Object b) {
   }
 
   private static boolean deepEquals(Object a, Object b, Set visited) {
-    LinkedList<DualKey> stack = new LinkedList<DualKey>();
+    Deque<DualKey> stack = new ArrayDeque<>();
```
inlined: `LinkedList` is generally prohibited because it almost never outperforms an `ArrayList` (there's a great C++ talk from Meyers on this; std::linked_list performs worse than std::vector for random operations and the gap actually increases as the list size increases. Memory locality is more important than O-notation)

firebase-database/src/test/java/com/google/firebase/database/util/ComparatorTester.java
```diff
@@ -165,11 +164,10 @@ private final void doTestEqualsCompatibility() {
...
+            // add a call to doNotRequireEqualsCompatibility() if this is not required
+            assertWithMessage("%s.equals(%s)", reference, other)
```
schmidt-sebastian: Nit: Should the message have quotes around `%s`?

schmidt-sebastian:
> Annotations for the first case of every type of warning/error I fixed

Awesome :) Thank you!

firebase-database/src/test/java/com/google/firebase/database/util/ComparatorTester.java
inlined: LOL. you just couldn't let me have a clean review, eh?

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
```diff
@@ -2350,9 +2352,8 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
-            System.currentTimeMillis() - Long.parseLong(snap.child("a").getValue().toString()))
-        < 2000);
+    Long drift = System.currentTimeMillis() - Long.parseLong(snap.child("a").getValue().toString());
```
PromanSEW:
```suggestion
    long drift = System.currentTimeMillis() - Long.parseLong(snap.child("a").getValue().toString());
```

firebase-database/src/main/java/com/google/firebase/database/core/utilities/ParsedUrl.java
```diff
@@ -32,4 +32,11 @@ public boolean equals(Object o) {
     if (!repoInfo.equals(parsedUrl.repoInfo)) return false;
     return path.equals(parsedUrl.path);
   }
+
+  @Override
+  public int hashCode() {
+    int result = repoInfo.hashCode();
+    result = 31 * result + path.hashCode();
+    return result;
```
PromanSEW:
```suggestion
    return 31 * repoInfo.hashCode() + path.hashCode();
```

firebase-database/src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -37,11 +37,17 @@
 @org.junit.runner.RunWith(RobolectricTestRunner.class)
 @Config(manifest = Config.NONE)
 public class MapperTest {
-  private static final double EPISLON = 0.00001f;
+  private static final double EPISLON = 0.00025f;
```
PromanSEW:
```suggestion
  private static final double EPSILON = 0.00025f;
```

firebase-database/src/test/java/com/google/firebase/database/DatabaseRegistrarTest.java
```diff
@@ -41,7 +41,7 @@ public void getComponents_publishesLibVersionComponent() {
 
   private static FirebaseApp appForDatabaseUrl(String url, String name) {
     return FirebaseApp.initializeApp(
-        RuntimeEnvironment.application.getApplicationContext(),
+        getApplicationContext(),
```
PromanSEW:
```suggestion
        ApplicationProvider.getApplicationContext(),
```

firebase-database/src/test/java/com/google/firebase/database/UnitTestHelpers.java
```diff
@@ -289,7 +289,7 @@ public static void ensureAppInitialized() {
     if (!appInitialized) {
       appInitialized = true;
       FirebaseApp.initializeApp(
-          RuntimeEnvironment.application.getApplicationContext(),
+          getApplicationContext(),
```
PromanSEW:
```suggestion
          ApplicationProvider.getApplicationContext(),
```

firebase-database/src/test/java/com/google/firebase/database/apitest/FirebaseDatabaseApiTest.java
```diff
@@ -59,7 +59,7 @@ public void getDifferentInstanceForAppWithUrl() {
 
   private static FirebaseApp appForDatabaseUrl(String url, String name) {
     return FirebaseApp.initializeApp(
-        RuntimeEnvironment.application.getApplicationContext(),
+        getApplicationContext(),
```
PromanSEW:
```suggestion
        ApplicationProvider.getApplicationContext(),
```

firebase-database/src/main/java/com/google/firebase/database/core/utilities/ParsedUrl.java
inlined: It's been too many years since I've been in school to remember the name of the notation but writing
AX^N + BX^(N-1) + CX^(N-2)...YX^1 + Z as
((((A)X + B)X + C)X + D)... 
is a much very common factorization for hashes and checksums because it forms a pattern that works incrementally as the input increase. Keeping the pattern.

firebase-database/src/test/java/com/google/firebase/database/DatabaseRegistrarTest.java
inlined: The public Google style guide only states that static methods should be referenced from a class and not a variable. The Google employee style guide is stricter and enforces this by requiring static methods to be statically imported.

firebase-database/src/test/java/com/google/firebase/database/MapperTest.java
inlined: Good typo catch.

inlined:
Ignoring the `api-information` test failure. It seems like all pull requests have failed `:firebase-common:ktx:apiInformation` since #1041.

PR #1040, opened 2019-12-05
Promote warnings to errors in encoders

rlazo:


encoders/firebase-encoders-json/firebase-encoders-json.gradle
```diff
@@ -56,3 +56,9 @@ dependencies {
     testImplementation 'org.mockito:mockito-core:2.25.0'
 
 }
+
+gradle.projectsEvaluated {
```
vkryachko: you don't need `gradle.projectsEvaluated`. calling `tasks.withType()` registers your callback on all existing and *future* tasks with a given type.

rlazo: done.

PR #1041, opened 2019-12-05
Fix api files.

rlazo:
Fix  outdated api.txt files.


inlined:
It looks like this has broken `:firebase-common:ktx:apiInformation`. I can't get a clean build anymore on my PR nor on master. Mind if we roll back?

rlazo:
Actually, this PR partially fixes the issue. We believe the issue with apiInfo was actually caused by #1033 and we're actively investigating on it. Feel free to ignore failures in the test.

PR #1043, opened 2019-12-06
Add inline encoding support.

vkryachko:
Encoders may choose to encode nested object inline instead of a nested
sub-object.


encoders/firebase-encoders-processor/src/test/java/com/google/firebase/encoders/processor/EncodableProcessorTest.java
```diff
@@ -320,7 +320,7 @@ public void compile_withMapOfListOfCustomType_shouldRegisterEncoderForThatType()
   public void compile_withEnum_shouldNotGenerateEnumEncoder() {
     Compilation result =
         javac()
-            .withProcessors(new AutoValueProcessor(), new EncodableProcessor())
+            .withProcessors(new EncodableProcessor())
```
davidmotson: I'm not sure I understand why this PR makes autovalueprocessor no longer necessary here.

vkryachko: It was never necessary, just when I was adding a new test below, I noticed that and removed.

PR #1044, opened 2019-12-06
Initial implementation of reflective encoder.

vkryachko:
The change introduces the concept of a "fallback" encoder that is used
for any type that does not have an explicitly registered encoder.

The reflective encoder itself is currently implemented in terms of
public getters and does not look at class fields, this can be made
configurable in the future.

### DIFF
--- a/encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoder.java
+++ b/encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoder.java
@@ -1,0 +1,69 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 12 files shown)


rlazo:
Is this PR suppose to merge into inline_encode?

vkryachko:
@rlazo no, it just depends on inline encoder changes, I will repoint once `vk.inline_encode` is merged

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoder.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @SuppressWarnings("unchecked")
+    ObjectEncoder<Object> encoder = getEncoder((Class<Object>) obj.getClass());
+    encoder.encode(obj, ctx);
+  }
+
+  private <T> ObjectEncoder<T> getEncoder(Class<T> type) {
+
+    ObjectEncoder<T> encoder = provider.get(type);
+    ;
```
rlazo: nit: remove

PR #937, opened 2019-10-23
Fix Android connectivity issues

thebrianchen:

PR #1045, opened 2019-12-06
Fix Android Connectivity Monitor (v2)

thebrianchen:
Continuation of #937 but instead of binding connectivity attempts to the WatchStream, this version bases connectivity attempts on grpc's ConnectivityState. 

Note that this implementation also depends on two [ManagedChannel APIs](http://google3/third_party/java_src/grpc/api/src/main/java/io/grpc/ManagedChannel.java?l=84&rcl=245107346) that are labeled as `@ExperimentalApi` and require a minimum grpc version `v1.6.1.`: `ManagedChannel.getState()` and `ManagedChannel.notifyWhenStateChanged()`.


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +215,48 @@ void shutdown() {
...
+   * reset the listener for the next state change.
+   */
+  private void listenToConnectivityState(ManagedChannel channel) {
+    ConnectivityState newState = channel.getState(false);
+    // Check that the new state is online, then cancel timer.
+    if (newState == ConnectivityState.READY) {
+      clearConnectivityTimer();
+    } else {
+      channel.notifyWhenStateChanged(newState, () -> listenToConnectivityState(channel));
+    }
```
mikelehen: Would it be possible for us to start the timer when the channel enters ConnectivityState.CONNECTING, and stop the timer when the channel enters ConnectivityState.READY?  And if the timer elapses can we somehow just "close" the channel (instead of marking it idle) in such a way that any pending streams / RPCs will just fail (in which case our existing retry behavior should kick in, but gRPC should use a new connection hopefully)?

Basically this would have two advantages:
1. It would be nice if this connection-timeout logic was 100% based on gRPC APIs (i.e. monitoring the ConnectivityState and resetting the channel when necessary) and it didn't bleed into the rest of our code (runBidiStreamingRpc, etc.).  That's cleaner and if/when gRPC improves their connection behavior, it would be trivial to just remove this timer logic (it could basically be a single method).
2. As-is, we'll only trigger this connection timeout if you're doing streaming RPCs.  If you're doing a non-streaming RPC (i.e. a transaction) then we'll still have the same problem with really long connection timeouts.

That said, I'm not sure how possible this is.  It looks like the two gRPC methods we might be able to play with are shutdownNow() [but then we'd presumably need to re-create the ManagedChannel?] or resetConnectBackoff() [but I don't know if that actually does anything useful in the 100% packet loss scenario].

On this note, if you haven't already, it would probably be worth adding some logging to listenToConnectivityState() and run your 100% packet loss scenario and verify what state(s) the gRPC connection goes through.  Is it just stuck CONNECTING the whole time?  Or does it do more than that?

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
```diff
@@ -74,7 +74,13 @@
...
      * multiple of these may be in the queue at a given time.
      */
-    RETRY_TRANSACTION
+    RETRY_TRANSACTION,
+    /**
+     * A timer used in RemoteStore to monitor when a connection attempt is unsuccessful and retry
+     * accordingly. While `ONLINE_STATE_TIMEOUT` is used to transition from UNKNOWN to OFFLINE,
+     * `CONNECTIVITY_ATTEMPT_TIMER` is used for tracking and retrying connectivity attempts.
+     */
+    CONNECTIVITY_ATTEMPT_TIMER
```
mikelehen: stale

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
```diff
@@ -290,18 +294,26 @@ private void close(State finalState, Status status) {
...
     } else if (code == Code.UNAUTHENTICATED) {
       // "unauthenticated" error means the token was rejected. Try force refreshing it in case it
       // just expired.
       firestoreChannel.invalidateToken();
+    } else if (code == Code.UNAVAILABLE) {
+      // These exceptions are thrown when the gRPC stream is closed with an connection error. For
+      // these cases, we need to use a new connection for the next connection attempt, which is
+      // done by marking the underlying channel as idle.
+      if (forceNewConnection
+          || status.getCause() instanceof java.net.ConnectException
```
mikelehen: I'm not 100% sure this is okay to do...

It looks like ConnectException usually refers to "Connection Refused" which means the backend is potentially down or intentionally trying to shed traffic (e.g. to mitigate a DoS attack).  So we should probably stick with our usual backoff delay.

Were you seeing this exception in your testing?  Or have some other specific reason to include it?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
```diff
@@ -174,6 +174,9 @@ public void run() {
   /** The time a stream stays open after it is marked idle. */
   private static final long IDLE_TIMEOUT_MS = TimeUnit.MINUTES.toMillis(1);
 
+  /** Maximum backoff time when reconnecting. */
```
mikelehen: "when reconnecting" doesn't really do anything to distinguish this from the existing BACKOFF_MAX_DELAY_MS which is also used when reconnecting.

I think the intention is that this max backoff delay is used when we're 100% sure that the connection failed on the client-side and didn't even reach the server.  In practice, I think the only case we can be sure of that is DNS failures (see my other comment about ConnectException).  If that's the case, then maybe we can call this DNS_FAILURE_BACKOFF_MAX_DELAY_MS or something...

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
thebrianchen: In my testing, I noticed that we get this message sometimes when the app starts in an 100% packet loss state, or when grpc shuts down the connection after running its own backoff (which technically shouldn't happen anymore after we add the connectivity timer). 

The main reason to check for ConnectException is to help the user reconnect faster if they start their app in a 100% packet loss state. Without it, if the user hits the maximum backoff delay before regaining connectivity, it could take up to DEFAULT_BACKOFF_MAX_DELAY (1 min) before making the next connection attempt.

However, protecting against DoS attacks seems like a much more important feature than this, so I'll remove the `java.net.ConnectException` check and only keep the `UnknownHostException` one.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
thebrianchen: > Would it be possible for us to start the timer when the channel enters ConnectivityState.CONNECTING, and stop the timer when the channel enters ConnectivityState.READY? 

Yes, I will update to do so. It doesn't make much of a difference in practice in the current state though, since it shifts from IDLE to CONNECTING almost immediately. However, if we end up implementing the timer fully independent of the WatchStream, I can see why that makes sense.

> And if the timer elapses can we somehow just "close" the channel (instead of marking it idle) in such a way that any pending streams / RPCs will just fail (in which case our existing retry behavior should kick in, but gRPC should use a new connection hopefully)?

I originally thought about closing the underlying channel and calling `GrpcCallProvider.initChannel()` again, but that seemed like more overhead, since you would have to tear down and rebuild the channel each time. `enterIdle()` to reset the connection seemed like a cleaner approach that would only require initializing the channel a single time, rather than on each timeout. 

>On this note, if you haven't already, it would probably be worth adding some logging to listenToConnectivityState() and run your 100% packet loss scenario and verify what state(s) the gRPC connection goes through. Is it just stuck CONNECTING the whole time? Or does it do more than that?

From what I've observed, it starts in IDLE, immediately transitions to CONNECTING, and stays stuck in CONNECTING until it goes to TRANSIENT_FAILURE after a minute or two.


From here, I guess the main issue is whether or not to use `enterIdle()` or completely shutdown and reinitialize the channel. If we do a complete tear down, we can keep all the connectivity logic in the grpc provider. However, in the current state, if we don't tear the channel down, we need to re-run the runBidiStreamingRpc to trigger a new connection.

thebrianchen:
Did another pass through, moved logic exclusively to gRPC side and removed the markChannelIdle() code.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +200,71 @@ void shutdown() {
       Thread.currentThread().interrupt();
     }
   }
+
+  /**
+   * Cancels the connectivityStateTimer if the new state indicates gRPC is online. Starts the timer
+   * if the new state indicates gRPC is connecting.
+   */
+  private void onConnectivityStateChange(ManagedChannel channel) {
```
mikelehen: It might be worth adding some logging to all of your changes so that we can tell from the logs when gRPC changes state and when we start / stop the timer, reset the channel, etc.  

In the short-term you can possibly use this logging to help validate these changes are behaving as expected (play around with an app and trigger different kinds of connection problems and make sure the logs make sense).  Longer-term I think it could be extremely helpful when debugging future connection-related complaints from customers.

FWIW- One thing I'm specifically curious about is if onConnectivityStateChange() will be called in response to us calling shutdownNow().  If it _does_ get called we need to make sure it doesn't cause any problems (e.g. creating an extra connection timer or something).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +200,71 @@ void shutdown() {
...
+    // Re-listen for next state change.
+    channel.notifyWhenStateChanged(
+        newState, () -> asyncQueue.enqueueAndForget(() -> onConnectivityStateChange(channel)));
+  }
+
+  /**
+   * Shuts down and re-initializes the channel.
+   *
+   * <p>This is used when the connectivity attempt timer elapses and we need to reset the gRPC
+   * channel to re-establish connectivity.
```
mikelehen: nits: reinitialize and reestablish aren't typically hyphenated.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
mikelehen: still stale

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/OnlineStateTracker.java
```diff
@@ -143,8 +143,8 @@ void handleWatchStreamFailure(Status status) {
         logClientOfflineWarningIfNecessary(
             String.format(
                 Locale.ENGLISH,
-                "Connection failed %d times. Most recent error: %s",
-                MAX_WATCH_STREAM_FAILURES,
+                "Backend didn't respond within %d seconds. Most recent error: %s\n",
+                ONLINE_STATE_TIMEOUT_MS / 1000,
```
mikelehen: Is this a stale change?  Else I don't understand it at all.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +200,71 @@ void shutdown() {
...
+    asyncQueue.enqueueAndForget(
+        () -> {
+          channel.shutdownNow();
+          initChannelTask();
+        });
+  }
+
+  private void initChannelTask() {
+    // We execute network initialization on a separate thread to not block operations that depend on
+    // the AsyncQueue.
```
mikelehen: One potential concern with recreating the channel is it could be expensive.  We're already doing it on a background thread for performance reasons, so it's possible this adds some meaningful delay.

As part of your logging, I'd recommend logging the start/finish of this so that we can get timing data and see how long this takes each time.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +200,71 @@ void shutdown() {
...
+
+  /**
+   * Cancels the connectivityStateTimer if the new state indicates gRPC is online. Starts the timer
+   * if the new state indicates gRPC is connecting.
+   */
+  private void onConnectivityStateChange(ManagedChannel channel) {
+    ConnectivityState newState = channel.getState(false);
+    // Check that the new state is online, then cancel timer.
+    if (newState == ConnectivityState.READY) {
+      clearConnectivityTimer();
```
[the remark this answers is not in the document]
thebrianchen: Good catch!

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/OnlineStateTracker.java
thebrianchen: I thought the original error message did not give the user much, since it would just say that the "Connection failed 1 times" along with the most recent error. Saying that the backend didn't respond for 10 seconds, tells the user more about why the state is being set to offline.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
thebrianchen: What needs to change here? I thought I updated it to reflect that this time is for gRPC attempts.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/OnlineStateTracker.java
mikelehen: But that's not why the state is being set to offline...

If the 10-second timer elapses, then we give them an error message saying so here: https://github.com/firebase/firebase-android-sdk/blob/a5dba09385851ed0dea304cb77564dad0b8d64f3/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/OnlineStateTracker.java#L114

But _here_ we are setting the state to offline because our connection failed and so we are giving them the failure error message.  The 10-second timer has _not_ elapsed at this point.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/AsyncQueue.java
mikelehen: Oops!  Sorry.  I thought this was something else.  My bad.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
thebrianchen: It's taking around 15-40ms to reset the channel from what I've seen based on the logs.

mikelehen: Hrm.  That is longer than I hoped.  I don't know if it's a problem or not.  I suggest getting input from Gil and gRPC folks (and perhaps point out this delay to gRPC and see if they have other suggestions for implementing a connection timeout).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
wilhuff: I don't think the conclusion here is right.

The origin of a status with an exception like this is always local, so if we get a `ConnectException` we're failing to connect _to the GFE_. Any failure past that point would result in a remote status like a bad gateway or similar HTTP response, which gRPC surfaces differently.

While we want to be generous with avoiding fast retries against the Firestore backend, the likelihood that the GFE would be spilling our load this way is so vanishingly small that we can just treat this as a certainty that the local network is to blame. Or in other words, I can't see any case where retrying faster after seeing a ConnectException would cause Firestore any distress.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
wilhuff: How about `CLIENT_NETWORK_FAILURE_BACKOFF_MAX_DELAY_MS`?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -53,6 +55,11 @@
   private static final String X_GOOG_API_CLIENT_VALUE =
       "gl-java/ fire/" + BuildConfig.VERSION_NAME + " grpc/";
 
+  // This timeout is used when attempting to establish a connection. If a connection attempt
+  // does not succeed in time, we close the stream and restart the connection, rather than having
```
wilhuff: The timer callback does not seem to "close the stream" in any way that I can tell. Based on comments in markChannelIdle, the first stream continues? Do this actually close the stream, or is this a side effect?

From what I can tell, the connectivityAttemptTimer merely starts a new clientCall (through a ~recursive invocation of runBidiStreamingRpc). What prevents the observer from getting callbacks from the old stream once that fails ~2 minutes later?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -50,6 +52,10 @@
 
   private CallOptions callOptions;
 
+  // A timer that elapses after CONNECTIVTY_ATTEMPT_TIMEOUT_MS, at which point we close the
```
wilhuff: Same comments as on CONNECTIVITY_ATTEMPT_TIMEOUT_MS; we don't seem to be closing the stream.

I suggest investing in a good description of exactly how this works in one location and then referring to that with something like "See $location for more details about how this works."

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
wilhuff: This was marked resolved without describing the resolution.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +215,48 @@ void shutdown() {
...
+    // Check that the new state is online, then cancel timer.
+    if (newState == ConnectivityState.READY) {
+      clearConnectivityTimer();
+    } else {
+      channel.notifyWhenStateChanged(newState, () -> listenToConnectivityState(channel));
+    }
+  }
+
+  /** Clears the connectivity timer if it exists. */
+  void clearConnectivityTimer() {
```
wilhuff: This is named inconsistently with setConnectivityAttemptTimer, though they seem related. Also, they could likely be grouped together.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/OnlineStateTracker.java
wilhuff: This was marked resolved without seeming to take action on the concern expressed here.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ExponentialBackoff.java
```diff
@@ -151,6 +161,9 @@ public void backoffAndRun(Runnable task) {
     } else if (currentBaseMs > maxDelayMs) {
       currentBaseMs = maxDelayMs;
     }
+
+    // Reset max delay to the default.
+    maxDelayMs = DEFAULT_BACKOFF_MAX_DELAY_MS;
```
wilhuff: This ignored the comments I made on the last version of this PR.

The max delay was configured in the constructor and may not actually be `DEFAULT_BACKOFF_MAX_DELAY_MS`. You can't read this constant here.

Instead, save two max delay fields in the constructor. One should be final, as maxDelayMs is today and it should be used here as the value to which to reset. The second value should be the one you actually bounce around based on which kind of error is in effect.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
mikelehen: I admit I don't have any specific knowledge of GFE load shedding behavior, etc. and am happy to defer to you.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
mikelehen: I just noticed that BACKOFF should maybe be first (to group this with the other backoff-related constants).  So: `BACKOFF_CLIENT_NETWORK_FAILURE_MAX_DELAY_MS`?

mikelehen: FWIW- I don't feel strongly.  Just trying to figure out how to make the name parse more sensibly...

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
mikelehen: I think you're looking at an old version of the PR? 😬   This code is gone in the latest version, right?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
thebrianchen: Reverted back to catch `ConnectException` as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java
thebrianchen: Changed to Michael's suggestion.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/OnlineStateTracker.java
thebrianchen: I believe this was already changed back to the original error message in my last commit.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
thebrianchen: Yes, that's the outdated version. The new one can be found [here](https://github.com/firebase/firebase-android-sdk/pull/1045/files#diff-03f10e7f8aea06522d1afa268c07d2cfR58).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
thebrianchen: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
thebrianchen: Resolution:

The connectivity timer starts when the channel enters a CONNECTING state, and if the timer elapses, the ManagedChannel will be torn down and reinitialized. However, if the channel enters any other state, the timer is cleared, and the state listener is set again.

Although it takes ~15-40ms to reset the channel, it's still a big enough win to justify the cost ([discussion](https://github.com/firebase/firebase-android-sdk/pull/1045#discussion_r356871348)).

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ExponentialBackoff.java
thebrianchen: Done (attempted).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
thebrianchen: Removed private method comments and consolidated the description in `onConnectivityStateChange`

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +201,84 @@ void shutdown() {
...
     }
   }
+
+  /**
+   * Monitors the connectivity state of the gRPC channel and resets the channel when gRPC fails to
+   * connect.
+   *
+   * <p>We currently cannot configure timeouts in connection attempts for gRPC
+   * (https://github.com/grpc/grpc-java/issues/1943), and until they support doing so, the gRPC
+   * connection can stay open for up to 2+ minutes before shutting down.
```
wilhuff: Optional: this would be slightly clearer if we made the deficiency we're working around about the notification of shut down rather than the shut down itself. Suggestion: 

> ... open for up to 2+ minutes before notifying us that it has shut down.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -198,4 +201,84 @@ void shutdown() {
...
+   */
+  private void onConnectivityStateChange(ManagedChannel channel) {
+    ConnectivityState newState = channel.getState(true);
+    Logger.debug(LOG_TAG, "Current gRPC connectivity state: " + newState);
+    // Check that the new state is online, then cancel timer.
+    if (newState == ConnectivityState.CONNECTING) {
+      Logger.debug(LOG_TAG, "Setting the connectivityAttemptTimer");
+      hardAssert(
+          connectivityAttemptTimer == null,
+          "connectivityAttemptTimer should be null when setting a new timer.");
```
mikelehen: How confident are you in this assert? 😁 

The docs for [notifyWhenStateChanged](https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannel.html#notifyWhenStateChanged-io.grpc.ConnectivityState-java.lang.Runnable-) say:

> There is an inherent race between the notification to callback and any call to getState(). There is a similar race between getState() and a call to notifyWhenStateChanged(). The state can change during those races, so there is not a way to see every state transition. "Transitions" to the same state are possible, because intermediate states may not have been observed. The API is only reliable in tracking the current state.

So I think that means we could end up seeing CONNECTING twice in a row, and this assert would fail and the SDK would crash.

So in this case we probably need to be defensive (just clear the existing timer if it exists) rather than strict (throwing an error if the timer already exists).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
mikelehen: 👍

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
thebrianchen: Not very after reading those docs! Thanks for catching this!

PR #1049, opened 2019-12-10
Update metalava api.txt, fix metalava task.

vkryachko:


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -101,55 +105,104 @@ private static void setupApiInformationAnalysis(Project project, LibraryExtensio
...
+                                      .getIncoming()
+                                      .artifactView(
+                                          config ->
+                                              config.attributes(
+                                                  container ->
+                                                      container.attribute(
+                                                          Attribute.of(
+                                                              "artifactType", String.class),
+                                                          "jar")));
+                              task.setClassPath(artifactView.getArtifacts().getArtifactFiles());
```
rlazo: I tried to come up with a witty comment about how nested this is, but I couldn't :-) If you can come up with a way of reduce the nesting level a bit, it would be great, but I don't have a very strong opinion in this particular case.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -101,55 +105,104 @@ private static void setupApiInformationAnalysis(Project project, LibraryExtensio
...
+            .register(
+                "generateApiTxtFile",
+                GenerateApiTxtFileTask.class,
+                task -> {
+                  task.setApiTxt(project.file("api.txt"));
+                  task.setMetalavaJarPath(metalavaOutputJarFile.getAbsolutePath());
+                  task.setSourceSet(mainSourceSet);
+                  task.setBaselineFile(project.file("baseline.txt"));
+                  if (project.hasProperty("updateBaseline")) {
+                    task.setUpdateBaseline(true);
```
rlazo:
```suggestion
                    task.setUpdateBaseline(project.hasProperty("updateBaseline"));
```

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
vkryachko: the nesting is gone now, it was unneeded code

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -141,7 +141,7 @@ private static void setupApiInformationAnalysis(Project project, LibraryExtensio
                   task.setSourceSet(mainSourceSet);
                   task.setBaselineFile(project.file("baseline.txt"));
                   if (project.hasProperty("updateBaseline")) {
```
rlazo: My mistake, I cannot suggest multiline changes in Github. With this change, you don't need the enclosing if/else.

vkryachko: done

PR #1057, opened 2019-12-14
Fis sdk

fredquintana:
Converted the FirebaseInstallations instrumentation tests into unit tests.


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -337,7 +337,7 @@ private PersistedInstallationEntry getPrefsWithGeneratedIdMultiProcessSafe() {
   }
 
   /** Use file locking to acquire a lock that will also block other processes. */
-  private FileLock getCrossProcessLock() {
+  FileLock getCrossProcessLock() {
```
diwu-arete: Why remove the private here?

I don't see any cross-class references added in this PR.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -351,7 +351,7 @@ private FileLock getCrossProcessLock() {
   }
 
   /** Release a previously acquired lock. */
-  private void releaseCrossProcessLock(FileLock fileLock) {
+  void releaseCrossProcessLock(FileLock fileLock) {
```
diwu-arete: Ditto

PR #1058, opened 2019-12-14
[FLoC] Change FLoC instrumental tests to be unit tests.

diwu-arete:


firebase-segmentation/src/test/java/com/google/firebase/segmentation/local/CustomInstallationIdCacheTest.java
```diff
@@ -19,16 +19,16 @@
...
-import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
 
 /** Instrumented tests for {@link CustomInstallationIdCache} */
```
ankitaj224: nit: Update the javadocs.

firebase-segmentation/src/test/java/com/google/firebase/segmentation/FirebaseSegmentationTest.java
```diff
@@ -184,7 +199,11 @@ public void testUpdateCustomInstallationId_CacheError_BackendOk() throws Interru
 
     // Expect exception
     try {
-      Tasks.await(firebaseSegmentation.setCustomInstallationId(CUSTOM_INSTALLATION_ID));
+      TestOnCompleteListener<Void> onCompleteListener = new TestOnCompleteListener<>();
```
ankitaj224: Optional: We can move this into an initialization method.

firebase-segmentation/src/test/java/com/google/firebase/segmentation/FirebaseSegmentationTest.java
```diff
@@ -184,7 +199,11 @@ public void testUpdateCustomInstallationId_CacheError_BackendOk() throws Interru
...
+      firebaseSegmentation
+          .setCustomInstallationId(CUSTOM_INSTALLATION_ID)
+          .addOnCompleteListener(taskExecutor, onCompleteListener);
+      onCompleteListener.await();
```
ankitaj224: Optional: We can move this into an initialization method.

diwu-arete: Hmmm, first, I don't think we can put this entire block into initialization() since it's not always the same in every test case.

If you meant that move TestOnCompleteListener<Void> onCompleteListener = new TestOnCompleteListener<>() into initialization, I also think putting here is better since it's a whole logic block with the two lines below.  Also, if a test case uses  onCompleteListener more than once, then putting it in initialization is also not perfect.

PR #1079, opened 2019-12-23
Change the default FTL bucket to "fireescape-ftl-test-results".

yifanyang:
FTL tests started from local will upload artifacts/logs to the new
bucket. This avoids clutter in the bucket "android-ci".

### DIFF
--- a/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
+++ b/buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
@@ -22,7 +22,7 @@ import org.gradle.api.Project
 
 
 class FirebaseTestServer extends TestServer {
-    private static final String DEFAULT_BUCKET_NAME = 'android-ci'
+    private static final String DEFAULT_BUCKET_NAME = 'fireescape-ftl-test-results'
     final Project project
     final FirebaseTestLabExtension extension
     final Random random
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
```diff
@@ -68,7 +67,10 @@ class FirebaseTestServer extends TestServer {
         Optional<String> resultsBucket = Optional.ofNullable(System.getenv('FTL_RESULTS_BUCKET')).map(Environment.&expand)
         Optional<String> resultsDir = Optional.ofNullable(System.getenv('FTL_RESULTS_DIR')).map(Environment.&expand)
 
-        List<String> args = ['--results-bucket', resultsBucket.orElse(DEFAULT_BUCKET_NAME)]
```
yifanyang: Discussed offline. Changing to uploading to FTL public bucket if the environment variable for bucket is not supplied.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.groovy
```diff
@@ -68,7 +67,10 @@ class FirebaseTestServer extends TestServer {
...
+        List<String> args = []
+        if (resultsBucket.isPresent()) {
+            args += ['--results-bucket', resultsBucket.get()]
+        }
         if (resultsDir.isPresent()) {
             args += ['--results-dir', Paths.get(resultsDir.get(), "${project.path}_${random.nextLong()}")]
```
yifanyang: `nextLong` can give negative numbers. I don't know if you think that is a problem.

vkryachko: It has not been a problem so far, but I would not be opposed to `Math.abs()` on it, but totally optional.

PR #1087, opened 2020-01-02
Rename QueryData to TargetData, and QueryCache to TargetCache.

wu-hui:
Bulk renaming mostly done by Android Studio.


firebase-firestore/mapping.txt
```diff
@@ -1282,10 +1282,10 @@ com.google.firebase.firestore.local.LocalSerializer -> com.google.firebase.fires
     2101:2103:com.google.firebase.firestore.model.MaybeDocument decodeMaybeDocument(com.google.firebase.firestore.proto.MaybeDocument):55 -> zza
```
schmidt-sebastian: I think this file is no longer used/maintained. Let's find out if we need it, but if we don't it is probably better to delete this file rather than to change the ProGuard mappings used in the past.

Update: Vlad says it is ok to remove it.

PR #1095, opened 2020-01-07
Reading IID Token from IID Shared Prefs.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -38,17 +40,79 @@
...
+  }
+
+  private static String getDefaultSenderId(FirebaseApp app) {
+    // Check for an explicit sender id
+    String senderId = app.getOptions().getGcmSenderId();
+    if (senderId != null) {
+      return senderId;
+    }
+    String appId = app.getOptions().getApplicationId();
+    if (!appId.startsWith("1:")) {
```
andirayo: There is also version 2: which works exactly the same as 1:

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -38,17 +40,79 @@
...
+      // Not v1, server should be updated to accept the full app ID now
```
andirayo: I don't understand the purpose of this comments.
I recommend replacing it with this:
"If applicationId does not contain a (GMP-)App-ID, it contains a Sender identifier"

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -38,17 +40,79 @@
...
+      return appId;
+    }
+    // For v1 app IDs, fall back to parsing the project number out
+    @SuppressWarnings("StringSplitter")
+    String[] parts = appId.split(":");
+    if (parts.length < 2) {
```
andirayo: if method #split, splits all parts, then the result should contain exactly 4 parts...

if (parts.length != 4)

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
andirayo: if (!appId.startsWith("1:") && !appId.startsWith("2:")) {

ankitaj224: Done.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: Done.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
ankitaj224: Done.

PR #1099, opened 2020-01-07
Pass the iid token to authenticate Instance-ID migration to FIS.

ankitaj224:


firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
```diff
@@ -90,6 +90,8 @@
 
   public static final String TEST_INSTANCE_ID_1 = "ccccccccccc";
 
+  public static final String TEST_INSTANCE_ID_TOKEN_1 = "iid.token";
```
andirayo: iid:token    (with a colon)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -349,13 +349,23 @@ private String readExistingIidOrCreateFid(PersistedInstallationEntry prefs) {
   /** Registers the created Fid with FIS servers and update the persisted state. */
   private PersistedInstallationEntry registerFidWithServer(PersistedInstallationEntry prefs)
       throws IOException {
+
+    // Note: Default value of instanceIdMigrationAuth: null
+    String iidToken = null;
+
+    if (prefs.getFirebaseInstallationId().equals(iidStore.readIid())) {
```
andirayo: Isn't reading from iidStore a complicated action?
Maybe checking the length of the identifier (e.g. == 11) is enough?
But in the end I don't care, just sharing my thoughts.

PR #1100, opened 2020-01-08
Get the IID token before fetching Remote Config.

danasilver:
Remote Config requires an IID token to evaluate certain conditions, but does not currently wait for the token from `FirebaseInstanceId#getToken()` before fetching. This returns the default condition value in some cases (ex. percentage conditions) rather than the evaluated result for an app instance.

This change replaces the use of the deprecated `#getToken()` with `#getInstanceId()`, which returns an async `Task` we can wait to complete before fetching, without otherwise blocking.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
@@ -36,6 +36,7 @@ import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.analytics.connector.AnalyticsConnector;
 import com.google.firebase.iid.FirebaseInstanceId;
+import com.google.firebase.iid.InstanceIdResult;
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigClientException;
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigException;
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigFetchThrottledException;
@@ -188,7 +189,18 @@ public class ConfigFetchHandler {
                   createThrottledMessage(backoffEndTime.getTime() - currentTime.getTime()),
                   backoffEndTime.getTime()));
     } else {
-      fetchResponseTask = fetchFromBackendAndCacheResponse(currentTime);
+      Task<InstanceIdResult> instanceIdTask = firebaseInstanceId.getInstanceId();
+      fetchResponseTask = instanceIdTask
+              .continueWithTask(executor, (completedIidTask) -> {
+                if (!instanceIdTask.isSuccessful()) {
+                  return Tasks.forException(new FirebaseRemoteConfigClientException(
+                          "Failed to get Firebase Instance ID token for fetch.",
+                          instanceIdTask.getException()));
+                }
+
+                InstanceIdResult instanceIdResult = completedIidTask.getResult();
+                return fetchFromBackendAndCacheResponse(instanceIdResult, currentTime);
+              });
     }
 
     return fetchResponseTask.continueWithTask(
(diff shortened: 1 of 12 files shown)


clp93:
Does this change slow down first fetch? Would adding a call to `getInstanceId()` in `FirebaseRemoteConfig#ensureInitialized()` or somewhere else in `FirebaseRemoteConfig` be appropriate?

danasilver:
It would slow down first fetch in the case where the token is not yet available (first fetch after installing). Agreed that adding a call to `FirebaseRemoteConfig#ensureInitialized()` seems appropriate since the IID token is required for fetch now.

We discussed possibly kicking off a call to `getInstanceId()` when the `FirebaseApp` is initialized, but I'm not sure that's feasible purely within the Remote Config SDK.

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -196,6 +205,7 @@ public void setUp() throws Exception {
   public void ensureInitialized_notInitialized_isNotComplete() {
     loadCacheWithConfig(mockFetchedCache, /*container=*/ null);
     loadCacheWithConfig(mockDefaultsCache, /*container=*/ null);
+    when(mockFirebaseInstanceId.getInstanceId()).thenReturn(Tasks.forResult(INSTANCE_ID_RESULT));
```
clp93: nit: preserve order?

clp93: in previous android changes tests, we have used methods like `loadInstanceIdAndToken()` (which you changed below). WDYT about doing something similar in this file? It does make it slightly more readable, but at this point, i'm just infavor of consistency :)  https://osscs.corp.google.com/firebase/firebase-android-sdk/+/master:firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandlerTest.java;l=868-871;drc=5301121df4727b357293b3b7de00555362f90170

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -270,15 +286,16 @@ private String createThrottledMessage(long throttledDurationInMillis) {
...
+  private FetchResponse fetchFromBackend(InstanceIdResult instanceId, Date currentTime)
+      throws FirebaseRemoteConfigException {
     try {
       HttpURLConnection urlConnection = frcBackendApiClient.createHttpURLConnection();
 
       FetchResponse response =
           frcBackendApiClient.fetch(
               urlConnection,
-              firebaseInstanceId.getId(),
-              firebaseInstanceId.getToken(),
```
erikeldridge: Thinking out loud: this continues to be an instance var. Weird how it has `getInstanceId` and `getId` methods.

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
danasilver: I was ordering by complete then incomplete tasks, but I can use the order of tasks in `ensureInitialized` instead

danasilver: Sure, I like the consistency. I do think just having the mock in the method is easier to read, though.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -188,7 +189,21 @@ public ConfigFetchHandler(
                   createThrottledMessage(backoffEndTime.getTime() - currentTime.getTime()),
                   backoffEndTime.getTime()));
     } else {
-      fetchResponseTask = fetchFromBackendAndCacheResponse(currentTime);
+      Task<InstanceIdResult> instanceIdTask = firebaseInstanceId.getInstanceId();
+      fetchResponseTask =
+          instanceIdTask.continueWithTask(
+              executor,
+              (completedIidTask) -> {
```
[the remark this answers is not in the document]
erikeldridge: 👍

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
erikeldridge: I was just noting the similarity in naming between `firebaseInstanceId.getInstanceId` and `instanceId.getId`. I guess this is highlighted by this review, where we can see both side by side.  To be clear, I don't think any action is required.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -188,7 +189,21 @@ public ConfigFetchHandler(
...
+                if (!completedIidTask.isSuccessful()) {
+                  return Tasks.forException(
+                      new FirebaseRemoteConfigClientException(
+                          "Failed to get Firebase Instance ID token for fetch.",
+                          instanceIdTask.getException()));
```
ciarand: nit: just grab the exception from the callback arg

PR #1104, opened 2020-01-08
Revert "Add ServerValue.increment() methods."

schmidt-sebastian:
Reverts firebase/firebase-android-sdk#991

Reverting this change as it likely has performance implications (as shown in firebase/firebase-js-sdk#2487)


schmidt-sebastian:
cc @inlined

schmidt-sebastian:
Closing this for now to reduce churn, especially since we haven't seen evidence of performance regressions on Android. Let's try to fix JS first and if there are any changes, we can backport them.

PR #1106, opened 2020-01-09
Configure kotlindoc generation per SDK.

vkryachko:
This makes it easier to generate docs for a subset of SDKs, additionally
it produces a set of separate _toc.yaml files as opposed to a global one.


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/Dokka.java
```diff
@@ -0,0 +1,179 @@
+// Copyright 2019 Google LLC
```
rlazo: 2020 maybe :-)

vkryachko: Done

PR #1107, opened 2020-01-09
Add verify support

thebrianchen:
Port from [JS](https://github.com/firebase/firebase-js-sdk/pull/2436).


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -513,38 +499,32 @@ public void testUpdatePOJOTransactionally() {
   }
 
   @Test
-  public void testHandleReadingOneDocAndWritingAnother() {
+  public void retriesWhenDocumentThatWasReadWithoutBeingWrittenChanges() {
```
wilhuff: Though not required, conventionally we've written these methods with a "test" prefix.

PR #1108, opened 2020-01-09
Using isDefaultApp method to check for default firebase app during IID migration.

ankitaj224:
Adressing the cyclic dependency bug: b/147156677


firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -50,15 +50,13 @@
 
   private final String defaultSenderId;
 
-  public IidStore() {
-    // Different FirebaseApp in the same Android application should have the same application
-    // context and same dir path. We only read existing Iids for the default firebase application.
+  public IidStore(@NonNull FirebaseApp firebaseApp) {
```
ankitaj224: I am unsure if the IidStore constructor will fail to initialize for non-default firebase app when getting the shred prefs.
```
private static final String IID_SHARED_PREFS_NAME = "com.google.android.gms.appid";
iidPrefs =
        firebaseApp
            .getApplicationContext()
            .getSharedPreferences(IID_SHARED_PREFS_NAME, Context.MODE_PRIVATE);
```

@ciarand or @gsakakihara  Can you please confirm?

PR #1113, opened 2020-01-10
refactor: add getters to the fields of FirebaseCrashlytics

thatfiredev:
This PR should add getters to `FirebaseCrashlytics` to provide better Kotlin patterns.
With the new getters, `setCrashlyticsCollectionEnabled(boolean)` and `setUserId(String)` can be called as properties in Kotlin:

```kotlin
FirebaseCrashlytics.getInstance().crashlyticsCollectionEnabled = true
FirebaseCrashlytics.getInstance().userId = "uid-1"
```


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -154,6 +154,13 @@ public void setUserId(@NonNull String identifier) {
     core.setUserId(identifier);
   }
 
+  /**
+   * @returns the unique identifier for the current user or empty string if it was not specified.
+   */
+  public @NonNull String getUserId() {
```
rosariopfernandes: Note to reviewer: this should be `@Nullable`, but the reason why I added the `@NonNull` annotation is kotlin requires the getter to have the same annotation as the setter's parameter (hence the empty string being returned).

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -154,6 +154,13 @@ public void setUserId(@NonNull String identifier) {
...
+    return core.getUserId() == null ? "" : core.getUserId();
```
rosariopfernandes: Note to reviewer: this should be `@Nullable`, but the reason why I added the `@NonNull` annotation is kotlin requires the getter to have the same annotation as the setter's parameter (hence the empty string being returned).

mrwillis21:
Hi @rosariopfernandes, thanks for the PR! We'll take a look at this after we finalize the first release. :)

rosariopfernandes:
@mrwillis21 Do you mind having a look at this now that Firebase Crashlytics is GA? :)

It would be nice if this was released with Crashlytics KTX ( #1503 )

mrwillis21:
@rosariopfernandes Hi Rosario! Thanks for the ping. We're currently working through a backlog of issues now that we're GA, and this is definitely on my radar. :)

PR #1115, opened 2020-01-10
Disconnecting HttpsURLConnection.

ankitaj224:
Also, improving the GZIPOutStream initialization and exception handling.


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -142,22 +142,25 @@ public InstallationResponse createFirebaseInstallation(
         httpsURLConnection.addRequestProperty(X_ANDROID_IID_MIGRATION_KEY, iidToken);
       }
 
-      GZIPOutputStream gzipOutputStream =
-          new GZIPOutputStream(httpsURLConnection.getOutputStream());
+      GZIPOutputStream gzipOutputStream = null;
       try {
+        gzipOutputStream = new GZIPOutputStream(httpsURLConnection.getOutputStream());
```
andirayo: This does not seem to throw a JSONException...
Why did you pull it in the try-catch block?
You can still keep the finally block without moving it in, I think.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -142,22 +142,25 @@ public InstallationResponse createFirebaseInstallation(
...
-        gzipOutputStream.close();
+        if (gzipOutputStream != null) {
+          gzipOutputStream.close();
+        }
       }
 
       int httpResponseCode = httpsURLConnection.getResponseCode();
 
       if (httpResponseCode == 200) {
         return readCreateResponse(httpsURLConnection);
```
fredquintana: I take it readCreateResponse() calls close itself?

fredquintana: rather, .disconnect()

fredquintana: I think it would be cleaner to have the reader method not disconnect, then you could do it in one place, maybe in a finally.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: You are right. Good catch :)

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Makes sense. Updated. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -134,43 +134,53 @@ public InstallationResponse createFirebaseInstallation(
...
+        return InstallationResponse.builder().setResponseCode(ResponseCode.BAD_CONFIG).build();
+      } finally {
+        httpsURLConnection.disconnect();
+      }
     }
 
     throw new IOException();
   }
 
+  private void writeToOutputStream(
```
fredquintana: Please make the name describe what is being written to the output stream.

Also, why did you factor it out?

ankitaj224: Updated.

PR #1130, opened 2020-01-15
Change coverage report format to be by product.

yifanyang:
Refactored coverage tasks setup.


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/ci/CheckCoveragePlugin.groovy
```diff
@@ -0,0 +1,164 @@
...
+    private def upload(task) {
+        if (System.getenv().containsKey("FIREBASE_CI")) {
+            def flag = convert(task.project.path)
+            def report = task.reports.xml.destination
+
+            if (System.getenv().containsKey("PROW_JOB_ID")) {
+                task.logger.quiet("Prow CI detected.")
+                uploadFromProwJobs(task.project, report, flag)
+            } else {
+                uploadFromCodecovSupportedEnvironment(task.project, report, flag)
```
rlazo: Just out of curiosity, is this something that happens right now or is it to support future use-cases?

yifanyang: Aha, I have been testing the report with my [forked repo](https://github.com/yifanyang/firebase-android-sdk), which is hooked up with Travis CI. This is to make the coverage report upload also work there. I can remove it if you think it causes confusions.

PR #1131, opened 2020-01-15
Handle incompatible IID SDK version with FIS SDK.

ankitaj224:


firebase-installations/firebase-installations.gradle
```diff
@@ -47,6 +47,10 @@ dependencies {
 
 
     compileOnly "com.google.auto.value:auto-value-annotations:1.6.5"
+    compileOnly ("com.google.firebase:firebase-iid:20.0.2"){
+        exclude group: "com.google.firebase", module: "firebase-common"
+        exclude group: "com.google.firebase", module: "firebase-components"
+    }
```
andirayo: I don't understand this.
Can you please explain to me in person?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -73,6 +78,11 @@
   /* file used for process-level syncronization of generating and persisting fids */
   private static final String LOCKFILE_NAME_GENERATE_FID = "generatefid.lock";
 
+  private static final String IID_SDK_PROP_FILE = "/firebase-iid.properties";
```
andirayo: Please add a code comment why you need this.
Something like
// Values that are needed to prevent incompatible versions of Instance-ID SDK to be used in parallel

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -73,6 +78,11 @@
...
+  private static final String IID_SDK_VERSION_BEFORE_FIS = "20.0.2";
```
andirayo: why only this explicit version?
is this the version of IID since I/O 2019?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -133,8 +143,17 @@ public static FirebaseInstallations getInstance() {
...
+  public static FirebaseInstallations getInstance(@NonNull FirebaseApp app)
+      throws FirebaseInstallationsException {
     Preconditions.checkArgument(app != null, "Null is not a valid value of FirebaseApp.");
+
+    if (!isIIDVersionCompatible()) {
+      throw new FirebaseInstallationsException(
+          "FirebaseInstallations will not work correctly with current version of Firebase "
+              + "Instance ID. Please update your Firebase Instance ID version.",
+          Status.INCOMPATIBLE_VERSIONS);
+    }
```
andirayo: Please outsource in dedicated method!

#preventParallelUsageOfIncompatibleVersionOfInstanceIdSDk

andirayo: let the other method throw the exception!

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -451,4 +470,34 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
...
+      if (TextUtils.isEmpty(iidVersion)
+          || Utils.compareVersion(IID_SDK_VERSION_BEFORE_FIS, iidVersion) >= 0) {
+        return false;
+      }
+    } catch (NoClassDefFoundError exception) {
+      // There is no IID SDK at all.
+      return true;
+    }
+    return true;
+  }
```
andirayo: Please have this method throw an Exception

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -451,4 +470,34 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
...
+    try {
+      final Class<FirebaseInstanceId> firebaseInstanceIdClass = FirebaseInstanceId.class;
+      String iidVersion = getIIDVersion();
+      if (TextUtils.isEmpty(iidVersion)
+          || Utils.compareVersion(IID_SDK_VERSION_BEFORE_FIS, iidVersion) >= 0) {
+        return false;
+      }
+    } catch (NoClassDefFoundError exception) {
+      // There is no IID SDK at all.
+      return true;
```
andirayo: code duplication of `return true`

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -451,4 +470,34 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
...
+    }
+    return true;
+  }
+
+  private static String getIIDVersion() {
```
andirayo: @Nullable

andirayo: How about using Optional?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -451,4 +470,34 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
...
+    Properties props = new Properties();
+    try {
+      String propertiesFileName = IID_SDK_PROP_FILE;
+      InputStream fs = FirebaseInstallations.class.getResourceAsStream(propertiesFileName);
+      if (fs != null) {
+        props.load(fs);
+        return props.getProperty("version", /* defaultValue */ null);
+      }
+    } catch (IOException e) {
```
andirayo: is this the only exception that can be thrown here?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -47,4 +47,28 @@ public boolean isAuthTokenExpired(PersistedInstallationEntry entry) {
   public long currentTimeInSecs() {
     return TimeUnit.MILLISECONDS.toSeconds(calendar.getTimeInMillis());
   }
+
+  /**
+   * Compares two version strings.
+   *
+   * @note It assumes version is in the "1.0.1" format.
```
andirayo: I don't like it

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -47,4 +47,28 @@ public boolean isAuthTokenExpired(PersistedInstallationEntry entry) {
...
+    String[] version1Parts = version1.split("\\.");
+    @SuppressWarnings("StringSplitter")
+    String[] version2Parts = version2.split("\\.");
+    for (int i = 0; i < Math.min(version1Parts.length, version2Parts.length); i++) {
+      int comparison = version1Parts[i].compareTo(version2Parts[i]);
+      if (comparison != 0) {
+        return comparison;
+      }
+    }
+    return Integer.compare(version1Parts.length, version2Parts.length);
```
andirayo: This seems way too complicated to me...
How about we just check for the two last versions...
or all versions since last I/O and disallow those?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: This is the latest version of IID with no on dependency on FIS SDK.

https://mvnrepository.com/artifact/com.google.firebase/firebase-iid

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: Done.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Yes, its the only expected exception.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Added the annotation. Can't add Optional - min SDK requirement for Optional is 24 and firebase sdk's are at 14

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Can't. I need to handle the NoClassDefFoundError exception and return true. Its the best case when there is no IID SDK in gradle depedency.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated the code better readability. Please review again.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -451,4 +477,35 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
     persistedInstallation.insertOrUpdatePersistedInstallationEntry(entry.withNoGeneratedFid());
     return null;
   }
+
+  private static boolean isIIDVersionCompatible() {
+    try {
+      final Class<FirebaseInstanceId> firebaseInstanceIdClass = FirebaseInstanceId.class;
```
vkryachko: Proguard may end up removing this line due to the `code/removal/variable` optimization, I would suggest you test with this optimization as well as with `code/removal/advanced`, `code/removal/simple` optimizations.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -121,7 +135,7 @@ private void preConditionChecks() {
    * @return a {@link FirebaseInstallationsApi} instance
    */
   @NonNull
-  public static FirebaseInstallations getInstance() {
+  public static FirebaseInstallations getInstance() throws FirebaseInstallationsException {
```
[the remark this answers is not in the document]
andirayo: Using an incompatible version of the Instance-ID SDK in parallel can break Firebase Targeting, as the same App-Instance will get 2 different identifiers. Ensuring that we do not break Firebase Targeting had the highest priority when discussing the design of FIS SDKs. There is a rather short summary in go/fis-iid-incompatible-versions.
In that sense, don't you think it is desirable to prevent this situation with a breaking (checked) exception?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -133,11 +147,23 @@ public static FirebaseInstallations getInstance() {
    * @return a {@link FirebaseInstallations} instance
    */
   @NonNull
-  public static FirebaseInstallations getInstance(@NonNull FirebaseApp app) {
+  public static FirebaseInstallations getInstance(@NonNull FirebaseApp app)
+      throws FirebaseInstallationsException {
     Preconditions.checkArgument(app != null, "Null is not a valid value of FirebaseApp.");
+    preventParallelUsageOfIncompatibleVersionOfInstanceIdSDk();
```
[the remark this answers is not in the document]
I think, we only need to check once in the private constructor of FirebaseInstallations.
(and maybe we only need to check for one FirebaseApp, e.g. the [default] app)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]
vkryachko: I guess one could argue either way, but imo the severity of the condition that causes an exception to be thrown is orthogonal to the type of exception to throw(`checked/unchecked`).

To quote effective java(go/ej3e-70):
```
 use checked exceptions for conditions from which the caller
can reasonably be expected to recover.
...By confronting the user with a checked exception, the API designer presents a
mandate to recover from the condition
```

With this in mind, it does not sound like something that can be recovered from at runtime, and the only way to recover is to rebuild the app with a different version of iid, hence my suggestion to make it unchecked. But if there are reasonable ways to recover from this state then sure, it could be checked.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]
vkryachko: On second thought, I think additionally you may want to move the check into the constructor of `FirebaseInstallations`, not the getInstance() method.
Motivation:
* Any other SDK that depends on FIS interop would respect the check automatically, as component dependency resolution does not normally go through `getInstance()` methods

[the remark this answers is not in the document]
vkryachko: re my last comment: seems like this is exactly what Rayo suggested above :)

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -72,6 +78,14 @@
   private final Object lockGenerateFid = new Object();
   /* file used for process-level syncronization of generating and persisting fids */
   private static final String LOCKFILE_NAME_GENERATE_FID = "generatefid.lock";
+  /* File used to read the version of Instance-ID(IID) SDK if it exists in the dependency tree.
+  This is needed to prevent incompatible versions of IID SDK to be used in parallel with FIS SDK. */
+  private static final String IID_SDK_PROP_FILE = "/firebase-iid.properties";
+  /* List of all Instance-ID versions that are incompatible with FIS SDK.*/
+  private static final List<String> IID_SDK_VERSIONS_INCOMPATIBLE_WITH_FIS =
+      Arrays.asList("20.0.2", "20.0.1", "20.0.0", "19.0.1", "19.0.0", "18.0.0");
```
vkryachko: What if I use an even older version? like 17.1.2?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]
ankitaj224: @vkryachko Accepted your suggestion and updated the code. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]
ankitaj224: @vkryachko Moved the check to the constructor. Do you think we still need to make it async?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Changed it to check for any older version than 20.1.1 ( IID with FIS)

ankitaj224:
@vkryachko  and @andirayo 

Updated this PR to handle runtime check for incompatible IID versions. Addressed the older comments. Please review the changes again.

I plan on implementing the lint checks as suggested by vlad in another PR.

Thanks.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -124,6 +134,51 @@ public Thread newThread(Runnable r) {
     this.utils = utils;
     this.iidStore = iidStore;
     this.fidGenerator = fidGenerator;
+    preventParallelUsageOfIncompatibleVersionOfInstanceIdSDk();
```
vkryachko: Why do it in a blocking way? The least we can do is do it on a thread and post the exception to the caller thread(or UI thread) through its Looper

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -124,6 +134,51 @@ public Thread newThread(Runnable r) {
...
+    try {
+      final Class<FirebaseInstanceId> firebaseInstanceIdClass = FirebaseInstanceId.class;
+    } catch (NoClassDefFoundError exception) {
+      // It is OK if there is no IID SDK at all.
+      return true;
+    }
+    String iidVersion = getIIDVersion();
+    // IID SDK version is incompatible if it is older than the IID SDK version with FIS
+    // dependency(i.e 20.1.1).
+    if (TextUtils.isEmpty(iidVersion)
```
vkryachko: afaict this check is not future-proof as iid won't have a properties file once released from github(something @VinayGuthal is working on)

fyi @ashwinraghav

ankitaj224: I wasn't aware of this. Clients moving from IID to directly depend on FIS have been requesting such a check to avoid running into multiple firebase identifiers. Without such a check in place we cant be sure that there is only one identifier.

vkryachko:
> I plan on implementing the lint checks as suggested by vlad in another PR.

With this check planned, do you think it may be sufficient so that the runtime check is not needed? (as lint will trigger on all gradle builds making runtime failure very unlikely to happen)

firebase-installations/firebase-installations.gradle
andirayo: open question

ankitaj224:
> > I plan on implementing the lint checks as suggested by vlad in another PR.
> 
> With this check planned, do you think it may be sufficient so that the runtime check is not needed? (as lint will trigger on all gradle builds making runtime failure very unlikely to happen)

The lint check definitely precedes over the runtime check. However, I am trying to understand how to implement the lint check which will take sometime on my end. Meanwhile, the clients migrating to FIS had the concern about no incompatibility check at all. The runtime check was already in place will unblock the clients, I thought of moving forward with this PR before the lint check.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -27,6 +28,7 @@
 import com.google.firebase.FirebaseException;
 import com.google.firebase.FirebaseOptions;
 import com.google.firebase.heartbeatinfo.HeartBeatInfo;
+import com.google.firebase.iid.FirebaseInstanceId;
```
andirayo: How can this work if there is no Instance ID available?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -76,6 +80,12 @@
   private static final String LOCKFILE_NAME_GENERATE_FID = "generatefid.lock";
   private static final String CHIME_FIREBASE_APP_NAME = "CHIME_ANDROID_SDK";
 
+  /* File used to read the version of Instance-ID(IID) SDK if it exists in the dependency tree.
+  This is needed to prevent incompatible versions of IID SDK to be used in parallel with FIS SDK. */
+  private static final String IID_SDK_PROP_FILE = "/firebase-iid.properties";
+  private static final String IID_VERSION_WITH_FIS = "20.1.1";
```
andirayo: Actually IID v20.1.0 already depended on FIS, but I assume we want to make sure they use the hotfixed version? right?

If that is the case, please add an appropriate comment, example:

// FirebaseInstanceId v20.1.0 already depended on FirebaseInstallations for identifier creation, but had an issue of possible silent failures. Thus, v20.1.1 or later is recommended.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -124,6 +134,51 @@ public Thread newThread(Runnable r) {
...
     this.iidStore = iidStore;
     this.fidGenerator = fidGenerator;
+    preventParallelUsageOfIncompatibleVersionOfInstanceIdSDk();
+  }
+
+  private void preventParallelUsageOfIncompatibleVersionOfInstanceIdSDk() {
+    if (!isIIDVersionCompatible()) {
+      throw new IllegalStateException(
+          "FirebaseInstallations will not work correctly with current version of Firebase "
+              + "Instance ID SDK. Please update your Firebase Instance ID version.");
```
andirayo: These version of FirebaseInstanceId and FirebaseInstallations can not be used in parallel.
Please update your Firebase Instance ID to a version 20.1.0 or later.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -124,6 +134,51 @@ public Thread newThread(Runnable r) {
...
+    }
+  }
+
+  /**
+   * Returns TRUE iff IID SDK is not used along with FIS SDK or IID SDK version is &gt than 20.1.1,
+   * otherwise returns FALSE.
+   */
+  private boolean isIIDVersionCompatible() {
+    try {
+      final Class<FirebaseInstanceId> firebaseInstanceIdClass = FirebaseInstanceId.class;
```
andirayo: How about using the fully qualified class path here?
com.google.firebase.iid.FirebaseInstanceId.class

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
andirayo: Question:
Why are we making changes to FirebaseInstanceId, when the plan is to deprecate this SDK as soon as possible?
Can't we just leave FirebaseInstanceId the way it is right now and focus on moving all Firebase SDKs to directly depend on FIS?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -124,6 +134,51 @@ public Thread newThread(Runnable r) {
...
+      return true;
+    }
+    String iidVersion = getIIDVersion();
+    // IID SDK version is incompatible if it is older than the IID SDK version with FIS
+    // dependency(i.e 20.1.1).
+    if (TextUtils.isEmpty(iidVersion)
+        && utils.compareVersions(IID_VERSION_WITH_FIS, iidVersion) >= 1) {
+      return false;
+    }
+    return true;
```
andirayo: if (TextUtils.isEmpty(iidVersion)) {
  return false;
}

return 1 > utils.compareVersions(IID_VERSION_WITH_FIS, iidVersion);

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Check the isIIDVersionCompatible(). If there is no Instance ID it catches NoClassDefFoundError and returns compatibility true.

vkryachko:
fyi I've prototyped a lint check in #1346 , ptal

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
vkryachko: Could you explain how it was done? Can we have an integ test for this?

PR #1132, opened 2020-01-16
Upgrade Firestore's GRPC to 1.26

schmidt-sebastian:


f0rev:
@schmidt-sebastian when do you plan merge this?

schmidt-sebastian:
> @schmidt-sebastian when do you plan merge this?

Unfortunately, this turns out to be a breaking change for our SDKs as it creates incompatibilities between the different Firebase products. It will likely be a while before we can release this change

easytoo12:
You say "a while" do you have an estimate?

schmidt-sebastian:
> You say "a while" do you have an estimate?

I chose my words carefully because I don't. Our next breaking change release is slated for May, but we can potentially move this up if there is enough demand.

easytoo12:
Sooner would be a great help

CWftw:
We have not been able to update to the latest versions of Firebase due to incompatibilities between protobuf java, java-lite and lite. Our best way forward seems to be waiting for Firebase to move to java-lite, and migrate our legacy code to java-lite. Therefore I agree that the sooner the better as I would like to stay up to date with Firebase releases.

thaidn:
+1 if you can make this happen sooner. A couple of Tink users are being blocked by this.

easytoo12:
Do you have an update on this Schmidt-Sebastian?

schmidt-sebastian:
> Do you have an update on this Schmidt-Sebastian?

We don't have much in terms of good news here. At this point it looks like we may bundle this change with other breaking changes in or around May.

easytoo12:
Hey Schmidt-sebastian is this still in the works for May?

schmidt-sebastian:
> Hey Schmidt-sebastian is this still in the works for May?

Yep!

schmidt-sebastian:
@ashwinraghav Can you re-assign this to someone that can resolve the remaining symbol conflicts?

ZacSweers:
Hey folks, this change is fixing a significant compatibility issue between firebase and any app using protobufs. It seems to have stalled activity-wise, are there plans to resume?

thaidn:
FWIW we've decided to shade our copy of protobuf-javalite in Tink, starting from Tink 1.4.0-RC2 (which is coming next week).

I recommend Firebase doing the same. https://github.com/google/tink/commit/ae241770f030fb82977a64c982f870de87231635 is our implementation, should be straight-forward to copy to Firebase.

schmidt-sebastian:
We are still working on this, but have moved active development to a new PR: https://github.com/firebase/firebase-android-sdk/pull/1534/files

PR #1140, opened 2020-01-16
Check for mandatory Firebase Options.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -107,12 +107,21 @@
   }
 
   /**
-   * Perform pre-condition checks to make sure {@link FirebaseOptions#getApiKey()} and {@link
-   * FirebaseOptions#getProjectId()} are not null.
+   * Perform pre-condition checks to make sure {@link FirebaseOptions#getApiKey()} and ({@link
+   * FirebaseOptions#getProjectId()} or {@link FirebaseOptions#getGcmSenderId()} ()} are not null.
```
andirayo: I recommend avoiding to mention Sender-ID in public documentation, because it is only an internal fall-back mechanism for us.
Feel free to describe in (Java) code-comments.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -107,12 +107,21 @@
...
    */
   private void preConditionChecks() {
+    Preconditions.checkNotNull(getApplicationId());
+    Preconditions.checkNotNull(getProjectIdentifier());
```
andirayo: How about checkNotEmpty?
Reason: If the value is set to "" it doesn't help us either.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -107,12 +107,21 @@
...
     Preconditions.checkNotNull(firebaseApp.getOptions().getApiKey());
-    Preconditions.checkNotNull(firebaseApp.getOptions().getProjectId());
+  }
+
+  /** Returns the Project Id or Project Number for the Google Cloud Project. */
+  @Nullable
+  String getProjectIdentifier() {
+    return firebaseApp.getOptions().getProjectId() == null
```
andirayo: Can you use #isEmpty()?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -107,12 +107,21 @@
   }
 
   /**
-   * Perform pre-condition checks to make sure {@link FirebaseOptions#getApiKey()} and {@link
-   * FirebaseOptions#getProjectId()} are not null.
+   * Perform pre-condition checks to make sure {@link FirebaseOptions#getApiKey()} and ({@link
+   * FirebaseOptions#getProjectId()} or {@link FirebaseOptions#getGcmSenderId()} ()} are not null.
    */
   private void preConditionChecks() {
+    Preconditions.checkNotNull(getApplicationId());
```
andirayo: Did you define the method #getApplicationId?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -107,12 +107,21 @@
...
+    Preconditions.checkNotNull(getProjectIdentifier());
     Preconditions.checkNotNull(firebaseApp.getOptions().getApiKey());
```
andirayo: Wanna create a method for firebaseApp.getOptions().getApiKey()?
maybe #getApiKey?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated.PTAL

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: It was defined previously. git only shows the changes - you can see the method: https://github.com/firebase/firebase-android-sdk/blob/cbb626b2776de344076d37b5b4fb03d6643c30bc/firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java#L143

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -14,6 +14,7 @@
 
 package com.google.firebase.installations;
 
+import android.text.TextUtils;
```
andirayo: I hope I did not make you add another factor that could make our SDK fail, haha

PR #1146, opened 2020-01-21
Add changelog for verify and connectivity

thebrianchen:


firebase-firestore/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Unreleased (21.3.1)
 - [feature] Added `Query.limitToLast(n: long)`, which returns the last `n`
   documents as the result.
+- [changed] Transactions previously required that every document read in a
```
wu-hui: 21.3.1 is released. 

Can you move `21.3.1` to below and create `21.3.2` instead?

thebrianchen: Done. Thanks!

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-# Unreleased (21.3.1)
+# Unreleased (21.3.2)
+- [changed] Transactions previously required that every document read in a
+  transaction must also be written to. This requirement has been removed, such
```
wilhuff: "Written to" on the end of the sentence is a little clunky. Consider "that every document read in a transaction must also be written".

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-# Unreleased (21.3.1)
+# Unreleased (21.3.2)
+- [changed] Transactions previously required that every document read in a
```
wilhuff: "Transactions previously required ... in a transaction" seems redundant. How about "Firestore previously required ..."?

firebase-firestore/CHANGELOG.md
wilhuff: Avoid complicated or legalistic phrasing like "such that". It's better to use plain language. Consider:

> This required has been removed and you can now ...

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
...
+  transaction must also be written to. This requirement has been removed, such
+  that you can now read a document in transaction without writing to it.
+- [changed] Firestore now recovers connectivity more quickly in situations
```
wilhuff: Other entries in this changelog use "Firestore now recovers more quickly" (i.e. without "connectivity"). 

You might also consider rephrasing to avoid "Firestore now recovers ... where Firestore". Something like:

> Firestore now recovers more quickly once connections suffering packet loss return to normal.

... or something like that.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-# Unreleased (21.3.1)
+# Unreleased (21.3.2)
+- [changed] Firestore previously required that every document read in a
```
schmidt-sebastian: FWIW You could make this a feature if you think it is more than a bug fix. You would then bump the version to 21.4.

wilhuff: Going point. I think it's worth flagging this as a `feature` and bumping the minor version.

PR #1149, opened 2020-01-21
Remove proto dep from ABT. Make the POJO public and use that instead

JasonAHeron:
This is to help keep RC proto dep free


JasonAHeron:
I think we're good, the functionality is still the same, just removing one way to make the object

PR #1150, opened 2020-01-21
Add an ObjectValue builder

schmidt-sebastian:
This PR is meant to pave the way for a mutable ObjectValue that is based on a Value.Builder. It proves that we only need `set()` and `delete()` (and not `equals()` and `compareTo()`) on the mutable type since most parts of the code still deal with immutable ObjectValues.

This is meant for an feature branch until we finish the Proto conversion as ObjectValue.Builder().set() is somewhat inefficient for nested field updates.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
@@ -117,7 +117,7 @@ public final class UserDataConverter {
 
     ParseAccumulator accumulator = new ParseAccumulator(UserData.Source.Update);
     ParseContext context = accumulator.rootContext();
-    ObjectValue.Builder updateData = ObjectValue.Builder.emptyBuilder();
+    ObjectValue.Builder updateData = ObjectValue.newBuilder();
 
     for (Entry<String, Object> entry : data.entrySet()) {
       FieldPath fieldPath =
@@ -155,7 +155,7 @@ public final class UserDataConverter {
 
     ParseAccumulator accumulator = new ParseAccumulator(UserData.Source.Update);
     ParseContext context = accumulator.rootContext();
-    ObjectValue.Builder updateData = ObjectValue.Builder.emptyBuilder();
+    ObjectValue.Builder updateData = ObjectValue.newBuilder();
 
     Iterator<Object> iterator = fieldsAndValues.iterator();
     while (iterator.hasNext()) {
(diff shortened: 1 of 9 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
```diff
@@ -117,7 +117,7 @@ public ParsedUpdateData parseUpdateData(Map<String, Object> data) {
 
     ParseAccumulator accumulator = new ParseAccumulator(UserData.Source.Update);
     ParseContext context = accumulator.rootContext();
-    ObjectValue updateData = ObjectValue.emptyObject();
+    ObjectValue.Builder updateData = ObjectValue.Builder.emptyBuilder();
```
wilhuff: This is pretty verbose. How about `ObjectValue.newBuilder()`?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
```diff
@@ -227,11 +228,12 @@ private ObjectValue transformObject(ObjectValue objectValue, List<FieldValue> tr
     hardAssert(
         transformResults.size() == fieldTransforms.size(), "Transform results length mismatch.");
 
+    ObjectValue.Builder builder = objectValue.toBuilder();
     for (int i = 0; i < fieldTransforms.size(); i++) {
       FieldTransform fieldTransform = fieldTransforms.get(i);
       FieldPath fieldPath = fieldTransform.getFieldPath();
-      objectValue = objectValue.set(fieldPath, transformResults.get(i));
+      builder.set(fieldPath, transformResults.get(i)).build();
```
wilhuff: Trailing `.build()` here seems wrong.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/FieldValueTest.java
```diff
@@ -356,4 +374,12 @@ public void testValueOrdering() {
         .addEqualityGroup(wrapObject(map("foo", "0")))
         .testCompare();
   }
+
+  private ObjectValue setField(ObjectValue empty, String a, FieldValue mod) {
```
wilhuff: The caller doesn't necessarily have to pass an empty object, do they? Seems like `empty` could be `objectValue`, as it is below.

firebase-firestore/src/main/java/com/google/firebase/firestore/UserDataConverter.java
schmidt-sebastian: Works for me. I got inspired by the rest of the Protobuf API.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/TransformMutation.java
schmidt-sebastian: Removed

firebase-firestore/src/test/java/com/google/firebase/firestore/model/FieldValueTest.java
schmidt-sebastian: Yeah, it should have been objectValue.

PR #1154, opened 2020-01-22
Use FetchTimeout value in Remote Config Connection and Read Timeouts

clp93:
At the moment, `FirebaseRemoteConfigSettings#FetchTimeoutInSeconds` is used only to set the HTTP Connection timeout, and there's no way for a developer to set the Read timeout (Definitions in http://stackoverflow.com/q/3069382)

To match RC iOS behavior, we use the FetchTimeout value for both the Connection and Read timeouts.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
@@ -232,8 +232,8 @@ public class RemoteConfigComponent {
         appId,
         apiKey,
         namespace,
-        metadataClient.getFetchTimeoutInSeconds(),
-        NETWORK_CONNECTION_TIMEOUT_IN_SECONDS);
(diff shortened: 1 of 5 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -232,8 +232,8 @@ ConfigFetchHttpClient getFrcBackendApiClient(
         appId,
         apiKey,
         namespace,
-        metadataClient.getFetchTimeoutInSeconds(),
-        NETWORK_CONNECTION_TIMEOUT_IN_SECONDS);
+        /* connectTimeoutInSeconds= */ metadataClient.getFetchTimeoutInSeconds(),
```
dmandar: is this the only change?

clp93: yep

PR #1157, opened 2020-01-23
Adding Proto-based equality and comparison

schmidt-sebastian:
This will be used (and tested) in the follow-up CL that adds the FieldValue tpes.

I am also removing one of the the two "compareInt" functions.

For reference: https://github.com/firebase/firebase-android-sdk/pull/1156

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/Blob.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/Blob.java
@@ -82,18 +82,6 @@ public class Blob implements Comparable<Blob> {
 
   @Override
   public int compareTo(@NonNull Blob other) {
-    int size = Math.min(bytes.size(), other.bytes.size());
-    for (int i = 0; i < size; i++) {
-      // Make sure the bytes are unsigned
-      int thisByte = bytes.byteAt(i) & 0xff;
-      int otherByte = other.bytes.byteAt(i) & 0xff;
-      if (thisByte < otherByte) {
-        return -1;
-      } else if (thisByte > otherByte) {
-        return 1;
-      }
-      // Byte values are equal, continue with comparison
-    }
-    return Util.compareIntegers(bytes.size(), other.bytes.size());
+    return Util.compareByteString(bytes, other.bytes);
   }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/Blob.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/Blob.java
@@ -82,6 +82,6 @@ public class Blob implements Comparable<Blob> {
 
   @Override
   public int compareTo(@NonNull Blob other) {
-    return Util.compareByteString(bytes, other.bytes);
+    return Util.compareByteStrings(bytes, other.bytes);
   }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
@@ -12,13 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.firebase.firestore.model;
+package com.google.firebase.firestore.model.value;
 
 import static com.google.firebase.firestore.util.Assert.fail;
 
 import androidx.annotation.Nullable;
 import com.google.common.base.Splitter;
-import com.google.firebase.firestore.model.value.FieldValue;
 import com.google.firebase.firestore.util.Util;
 import com.google.firestore.v1.ArrayValue;
 import com.google.firestore.v1.MapValue;
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
@@ -22,6 +22,8 @@ import com.google.firebase.firestore.util.Util;
 import com.google.firestore.v1.ArrayValue;
 import com.google.firestore.v1.MapValue;
 import com.google.firestore.v1.Value;
+import com.google.protobuf.Timestamp;
+import com.google.type.LatLng;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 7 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
```diff
@@ -0,0 +1,266 @@
...
+import com.google.firebase.firestore.model.value.FieldValue;
+import com.google.firestore.v1.ArrayValue;
+import com.google.firestore.v1.MapValue;
+import com.google.firestore.v1.Value;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class ProtoValueUtil {
```
wilhuff: Same comment about naming: how about `ProtoValues`?

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
```diff
@@ -0,0 +1,266 @@
...
+
+  public static int typeOrder(Value value) {
+
+    switch (value.getValueTypeCase()) {
+      case NULL_VALUE:
+        return FieldValue.TYPE_ORDER_NULL;
```
wilhuff: Putting this class in the `util` package creates a cyclic reference between packages that isn't going to work in C++. Since we're using these utilities in the service of building models, might as well just put this in the `model` package.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
```diff
@@ -0,0 +1,266 @@
...
+    int leftType = typeOrder(left);
+    int rightType = typeOrder(right);
+    if (leftType != rightType) {
+      return false;
+    }
+
+    switch (leftType) {
+      case FieldValue.TYPE_ORDER_ARRAY:
+        return arrayEquals(left, right);
+      case FieldValue.TYPE_ORDER_NUMBER:
```
wilhuff: Nit: consider ordering the switch cases consistently so that it's easier to audit/reason about differences (in this case, that would mean number, array, object).

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
```diff
@@ -0,0 +1,266 @@
...
+        return numberEquals(left, right);
+      case FieldValue.TYPE_ORDER_OBJECT:
+        return objectEquals(left, right);
+      default:
+        return left.equals(right);
+    }
+  }
+
+  private static boolean objectEquals(Value left, Value right) {
```
wilhuff: Nit: consider ordering these methods in the file in the order you're using them in `equals`, above.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
```diff
@@ -0,0 +1,266 @@
...
+      case FieldValue.TYPE_ORDER_ARRAY:
+        return compareArrays(left, right);
+      case FieldValue.TYPE_ORDER_OBJECT:
+        return compareMaps(left, right);
+      default:
+        throw fail("Invalid value type: " + leftType);
+    }
+  }
+
+  private static int compareMaps(Value left, Value right) {
```
wilhuff: Nit: consider ordering these to match their usage above.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
```diff
@@ -0,0 +1,266 @@
...
+    return Long.signum(
+        left.getTimestampValue().getSeconds() - right.getTimestampValue().getSeconds());
+  }
+
+  private static int compareNumbers(Value left, Value right) {
+    if (left.getValueTypeCase() == Value.ValueTypeCase.DOUBLE_VALUE) {
+      double thisDouble = left.getDoubleValue();
+      if (right.getValueTypeCase() == Value.ValueTypeCase.DOUBLE_VALUE) {
+        return Util.compareDoubles(thisDouble, right.getDoubleValue());
+      } else {
```
wilhuff: You could express this with less code if you handled all the unexpected cases in one place, below:

```
if (left is double) {
  if (right is double) {
    return ...;
  } else if (right is int) {
    return ...;
  }
} else if (left is int) {
  if (right is double) {
    return ...;
  } else if (right is int) {
    return ...;
  }
}

hardAssert("Unexpected types %s and %s, left, right);
```

firebase-firestore/src/main/java/com/google/firebase/firestore/Blob.java
```diff
@@ -82,18 +82,6 @@ public int hashCode() {
...
-      int otherByte = other.bytes.byteAt(i) & 0xff;
-      if (thisByte < otherByte) {
-        return -1;
-      } else if (thisByte > otherByte) {
-        return 1;
-      }
-      // Byte values are equal, continue with comparison
-    }
-    return Util.compareIntegers(bytes.size(), other.bytes.size());
+    return Util.compareByteString(bytes, other.bytes);
```
wilhuff: Other compare methods are plural (e.g. compareIntegers). `compareByteStrings`?

schmidt-sebastian: Fixed

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
schmidt-sebastian: Renamed

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
schmidt-sebastian: Done. I believe that means that this class can then be package-private. I did not yet make it so since my next PR temporarily adds a new package for the new ObjectValue to avoid breaking all existing code. Added a TODO to change the visibility.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
schmidt-sebastian: Done (here and everywhere). I am following the type order constants.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ProtoValueUtil.java
schmidt-sebastian: Great suggestion. Updated.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -166,39 +166,46 @@ public static int compare(Value left, Value right) {
...
-    // Only equal if both iterators are exhausted.
-    return Util.compareBooleans(iterator1.hasNext(), iterator2.hasNext());
+    throw fail("Unexpected values: %s vs %s", left, right);
   }
 
-  private static int compareArrays(Value left, Value right) {
-    int minLength =
-        Math.min(left.getArrayValue().getValuesCount(), right.getArrayValue().getValuesCount());
+  private static int compareTimestamps(Value left, Value right) {
+    if (left.getTimestampValue().getSeconds() == right.getTimestampValue().getSeconds()) {
```
wilhuff: Elsewhere we compare and then if non-zero do the second one. Consider rephrasing as:

```java
int cmp = Util.compareIntegers(left.getTimestampValue().getSeconds(), right.getTimestampValue().getSeconds());
if (cmp != 0) return cmp;

return Util.compareIntegers(left.getTimestampValue().getNanos(), right.getTimestampValue().getNanos());
```

This has two benefits:
  * it makes the hierarchy of comparison linear; and
  * it avoids using subtraction for comparison, which is specious in the general case due to the possibility of overflow. I think it's the case that timestamp values are defined in such a way that subtraction can't possibly overflow, but there's no reason not to reuse the comparator we already have, in which case we don't have to think about it.

schmidt-sebastian: This code is actually based on the original code. I do think the comparisons in the other functions are easier to follow, so I updated it.

As part of this, I changed the method to take TimestampValues (along with the rest). This avoids a bunch of extra linebreaks since it removes the calls to getTimestampValue() from here.

PR #1171, opened 2020-01-29
Add test to verify canonical IDs are stable

schmidt-sebastian:
There is a risk in the Protobuf FieldValue rewrite, so let's add a test to make sure we don't break the IDs.

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
@@ -29,6 +29,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
+import com.google.firebase.Timestamp;
 import com.google.firebase.firestore.Blob;
 import com.google.firebase.firestore.GeoPoint;
 import com.google.firebase.firestore.model.Document;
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
```diff
@@ -553,4 +559,62 @@ public void testMatchesAllDocuments() {
...
+
+  @Test
+  public void testCanonicalIdsAreStable() {
+    // This test aims to ensure that we do not break canonical IDs, as they are used as keys in
+    // the TargetCache.
+
+    Query baseQuery = Query.atPath(ResourcePath.fromString("collection"));
+
+    Map<Query, String> queryToCanoncialIdMap = new LinkedHashMap<>();
+    queryToCanoncialIdMap.put(baseQuery, "collection|f:|ob:__name__asc|lt:LIMIT_TO_FIRST");
```
wilhuff: Consider adding an assertion helper instead. Something like

```java
private void assertCanonicalId(String canonicalId, Query query) {
  assertEquals(canonicalId, query.getCanonicalId());
}
```
These end up being about the same size at the call site but are easier to step through/debug/see exactly what failed just from exception line numbers.

schmidt-sebastian: Done

firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
```diff
@@ -553,4 +558,67 @@ public void testMatchesAllDocuments() {
     query = baseQuery.endAt(new Bound(Collections.emptyList(), true));
     assertFalse(query.matchesAllDocuments());
   }
+
+  @Test
+  public void testCanonicalIdsAreStable() {
+    // This test aims to ensure that we do not break canonical IDs, as they are used as keys in
```
wu-hui: Sorry
It's the `Target`'s canonical ID that is used in Target cache. `Query` canonical ID is only used in `SpecTestCase` and I have a TODO to remove it all together.

schmidt-sebastian: Done. It makes me wish we had a test class for Target though.

wu-hui: Yeah, we probably should..

PR #1174, opened 2020-01-29
Fixing a bunch of lint errors from G3 while copying

wu-hui:


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -549,7 +549,6 @@ public void testQueriesCanUseArrayContainsAnyFilters() {
     Map<String, Object> docD = map("array", asList(42L), "array2", asList("bingo"));
     Map<String, Object> docE = map("array", asList(43L));
     Map<String, Object> docF = map("array", asList(map("a", 42L)));
-    Map<String, Object> docG = map("array", 42L);
 
     CollectionReference collection =
         testCollectionWithDocs(
```
thebrianchen: I recently changed tests on master to include `docG` in order to match the other SDKs:
https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java

Can you include docG here? That should make the error go away too.

wu-hui: Done.

PR #1175, opened 2020-01-29
Add support for ignoring nulls in JsonDataEncoder

mrwillis21:


encoders/firebase-encoders-json/src/json/java/com/google/firebase/encoders/json/JsonValueObjectEncoderContext.java
```diff
@@ -37,36 +37,37 @@
...
   @NonNull
   @Override
   public JsonValueObjectEncoderContext add(@NonNull String name, @Nullable Object o)
       throws IOException, EncodingException {
-    maybeUnNest();
-    jsonWriter.name(name);
-    if (o == null) {
-      jsonWriter.nullValue();
-      return this;
+    if (ignoreNullValues) {
```
rlazo: I think you can go further simplify by just adding here

```
if (ignoreNullValues && o == null) return this;
```

and the rest of the code can stay the same.

mrwillis21: The reason I didn't go with that approach is that, in the case `ignoreNullValues` is `true`, and `o` is _not_ `null`, it will be `null`-checked a second time unnecessarily.

rlazo: Actually, both methods `internalAdd` and `internalAddIgnoreNullValues` have a `if (o == null)` condition, so the second check still happens.

mrwillis21: Yes, in either case, `o` gets `null`-checked _once_. But if I just added that line to the original method, `o` would get `null`-checked _again_ after `jsonWriter.name(name);`

rlazo: TL;DR; LGTM

Long form: Actually, the conditional will short-circuit if `ignoreNullValues==false`, so the "first check" will *only* happen when `ignoreNullValues=true` and `o != null`, which most like is going to be **most common case** :-)

mrwillis21: I think this is essentially what I was saying before. :)

> The reason I didn't go with that approach is that, in the case `ignoreNullValues` is `true`, and `o` is _not_ `null`, it will be `null`-checked a second time unnecessarily.

PR #1178, opened 2020-01-30
Add explicit FieldValue canonicalization

schmidt-sebastian:
Adds a new format for canonical IDs of field values that is meant to be verifiable and stable.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
@@ -24,10 +24,11 @@ import com.google.firestore.v1.MapValue;
 import com.google.firestore.v1.Value;
 import com.google.protobuf.Timestamp;
 import com.google.type.LatLng;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.SortedMap;
 import java.util.TreeMap;
 
 // TODO(mrschmidt): Make package-private
@@ -254,12 +255,12 @@ public class ProtoValues {
   /** Generate the canonical ID for the provided field value (as used in Target serialization). */
   public static String canonicalId(Value value) {
     StringBuilder builder = new StringBuilder();
-    stringifyValue(builder, value);
+    canonifyValue(builder, value);
     return builder.toString();
   }
 
   // TODO(mrschmidt): Use in target serialization and migrate all existing TargetData
-  private static void stringifyValue(StringBuilder builder, Value value) {
+  private static void canonifyValue(StringBuilder builder, Value value) {
     switch (value.getValueTypeCase()) {
       case NULL_VALUE:
         builder.append("null");
@@ -274,7 +275,7 @@ public class ProtoValues {
         builder.append(value.getDoubleValue());
         break;
       case TIMESTAMP_VALUE:
-        stringifyTimestamp(builder, value.getTimestampValue());
+        canonifyTimestamp(builder, value.getTimestampValue());
         break;
       case STRING_VALUE:
         builder.append(value.getStringValue());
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -249,4 +249,87 @@ private static int compareMaps(MapValue left, MapValue right) {
...
+  }
+
+  private static void buildGeoPointCanonicalId(StringBuilder builder, LatLng latLng) {
+    builder.append(String.format("{lat:%s,lng:%s}", latLng.getLatitude(), latLng.getLongitude()));
+  }
+
+  private static void buildObjectCanonicalId(StringBuilder builder, MapValue mapValue) {
+    builder.append("{");
+    boolean first = true;
+    for (Map.Entry<String, Value> entry : mapValue.getFieldsMap().entrySet()) {
```
wilhuff: Are you intending to make it an invariant that MapValues are always sorted by key? I don't recall seeing this in your other PR adding `protovalue.ObjectValue.Builder`.

If not, we probably need to sort the strings here :-(. I think maintaining the invariant is probably better though.

Note that server-supplied MapValues already conform to this invariant.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -249,4 +249,87 @@ private static int compareMaps(MapValue left, MapValue right) {
...
+        break;
+      case MAP_VALUE:
+        buildObjectCanonicalId(builder, value.getMapValue());
+        break;
+      default:
+        throw fail("Invalid value type: " + value.getValueTypeCase());
+    }
+  }
+
+  private static void buildTimestampCanonicalId(StringBuilder builder, Timestamp timestamp) {
```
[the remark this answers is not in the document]
wilhuff: "stringify" doesn't really tie into canonical ID though and I think that's probably more important. I don't have any better suggestions though, so feel free to put it back to makeCanonicalId or whatever it was.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -249,4 +250,92 @@ private static int compareMaps(MapValue left, MapValue right) {
...
+
+  private static void stringifyGeoPoint(StringBuilder builder, LatLng latLng) {
+    builder.append(String.format("geo(%s,%s)", latLng.getLatitude(), latLng.getLongitude()));
+  }
+
+  private static void stringifyObject(StringBuilder builder, MapValue mapValue) {
+    // Even though MapValue are likely sorted correctly based on their insertion order (e.g. when
+    // received from the backend), local modifications can bring elements out of order. We need to
+    // re-sort the elements to ensure that canonical IDs are independent of insertion order.
+    SortedMap<String, Value> sortedMap = new TreeMap<>(mapValue.getFieldsMap());
```
wilhuff: Building a binary tree to sort the entries is pretty wasteful in terms of memory allocation. It's also pessimistic because building the tree is always going to cost O(n lg(n)). An alternative would be to collect the keys into an ArrayList and sort that. That has the benefit of allocating less memory and also runs in O(n) if the list is already sorted.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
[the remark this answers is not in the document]
schmidt-sebastian: I will rename these methods to use "canonify" instead. Apparently other people are using this terminology too: https://docs.cfengine.com/docs/3.12/reference-functions-canonify.html

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
schmidt-sebastian: I went through different implementations for optimize for both runtime complexity and readability. While I still think that using large Maps here is kind of an anti-pattern (especially given our lack of truncation support),  adding an extra line here to pre-sort the keys (and performing a key-based lookup on the HashMap) might be a worthy tradeoff. Updated.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -249,4 +251,93 @@ private static int compareMaps(MapValue left, MapValue right) {
...
+  private static void canonifyGeoPoint(StringBuilder builder, LatLng latLng) {
+    builder.append(String.format("geo(%s,%s)", latLng.getLatitude(), latLng.getLongitude()));
+  }
+
+  private static void canonifyObject(StringBuilder builder, MapValue mapValue) {
+    // Even though MapValue are likely sorted correctly based on their insertion order (e.g. when
+    // received from the backend), local modifications can bring elements out of order. We need to
+    // re-sort the elements to ensure that canonical IDs are independent of insertion order.
+    List<String> keys = new ArrayList<>(mapValue.getFieldsMap().keySet());
+    Collections.sort(keys);
```
wilhuff: You can make this `keys.sort()` if you want.

schmidt-sebastian: Done

schmidt-sebastian: TIL: List.sort() is API Level 24.

PR #1182, opened 2020-01-30
Canonical ID Schema Migration

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -530,6 +537,26 @@ private boolean tableContainsColumn(String table, String column) {
     return columns;
   }
 
+  private void rewriteCanonicalIds() {
+    new SQLitePersistence.Query(db, "SELECT target_id, target_proto FROM targets")
```
wilhuff: Dumb question: this file doesn't appear to employ a transaction anywhere. Doesn't `db.execSQL` auto-commit if you don't? It seems like it would be bad if this partially committed.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -530,6 +537,26 @@ private boolean tableContainsColumn(String table, String column) {
...
+            cursor -> {
+              int targetId = cursor.getInt(0);
+              byte[] targetProtoBytes = cursor.getBlob(1);
+
+              try {
+                Target targetProto = Target.parseFrom(targetProtoBytes);
+                TargetData targetData = serializer.decodeTargetData(targetProto);
+                String updatedCanonicalId = targetData.getTarget().getCanonicalId();
+                db.execSQL(
+                    "UPDATE targets SET canonical_id  = ? WHERE target_id = ?",
```
wilhuff: This migration can't be rolled back and has the potential to cause problems because `target_id` must be guaranteed unique (it's the primary key of the table).

If a user does roll back, will this actually cause a problem with duplicate target ids? I think probably not because the next target id is in target_globals and we're not changing that here, right?

One alternative that makes this slightly more rollback friendly would be to put the new target id in a new column and create an additional index on that column. I can't say I'm super excited about this prospect and likely would prefer to just document that this release includes a migration that can't be rolled-back from.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: The transaction is opened here: https://cs.corp.google.com/aosp-master/frameworks/base/core/java/android/database/sqlite/SQLiteOpenHelper.java?q=f:SQLiteopenhelper.java&dr&l=409

It will only get committed if line 421 executes.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: Good catch. The fix for this is actually pretty straightforward as we can always rewrite canonical IDs with the current format. The target IDs don't change and as as long the format in the schema matches the expected format at the current client version we should be safe. I modified the schema converter to support downgrades and updated the code to run this particular migration on both up- and downgrades.

schmidt-sebastian: Hm. I just realized that the new version of the schema converter is pretty dumb. It assumes that my shiny new code is shipped to old SDKs. What do you think about submitting the PR in its original form and letting LRU GC delete stale QueryData if the user chooses to downgrade?

wilhuff: Letting LRU GC do its thing SGTM.

schmidt-sebastian: I reverted the commit that pretended to add support for downgrades.

PR #1190, opened 2020-02-03
Chime Hack: Allow Chime to migrate Insatnce_IDs.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -363,13 +364,14 @@ private PersistedInstallationEntry getPrefsWithGeneratedIdMultiProcessSafe() {
...
+    // Check if this firebase app is the default (first initialized) instance or is a chime app
+    if (!firebaseApp.getName().equals(CHIME_FIREBASE_APP_NAME)
+        && (!firebaseApp.isDefaultApp() || !prefs.shouldAttemptMigration())) {
       return fidGenerator.createRandomFid();
     }
-    // For a default firebase installation, read the existing iid from shared prefs
+    // For a default/chime firebase installation, read the existing iid from shared prefs
     String fid = iidStore.readIid();
-    if (fid == null) {
+    if (TextUtils.isEmpty(fid)) {
```
andirayo: shouldn't this be a separate PR?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -363,13 +364,14 @@ private PersistedInstallationEntry getPrefsWithGeneratedIdMultiProcessSafe() {
   }
 
   private String readExistingIidOrCreateFid(PersistedInstallationEntry prefs) {
-    // Check if this firebase app is the default (first initialized) instance
-    if (!firebaseApp.isDefaultApp() || !prefs.shouldAttemptMigration()) {
+    // Check if this firebase app is the default (first initialized) instance or is a chime app
+    if (!firebaseApp.getName().equals(CHIME_FIREBASE_APP_NAME)
+        && (!firebaseApp.isDefaultApp() || !prefs.shouldAttemptMigration())) {
```
andirayo: Not sure if the bracketing is right here...
I don't know what #shouldAttemptMigration does (so I might be wrong here)...
But it seems, that this should be changed to:

    if ((!firebaseApp.getName().equals(CHIME_FIREBASE_APP_NAME) && !firebaseApp.isDefaultApp())
       || !prefs.shouldAttemptMigration()) {

ankitaj224:
> 2 comments

Addressed.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Small change and seemed relevant as the change the same method. Will do it next time.

PR #1206, opened 2020-02-05
Drop Array, Reference, Number, Integer and DoubleValue

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -157,8 +164,8 @@ private static boolean objectEquals(Value left, Value right) {
   }
 
   /** Returns true if the Value list contains the specified element. */
-  public static boolean contains(List<Value> haystack, Value needle) {
-    for (Value haystackEl : haystack) {
+  public static boolean contains(ArrayValueOrBuilder haystack, Value needle) {
```
wilhuff: Nearly all callers check the array type before calling this. It seems like it would be worth adding an alternative like this:

```java
public static boolean arrayContains(Value haystack, Value needle) {
  return isArray(haystack) && contains(haystack.getArrayValue(), needle);
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Bound.java
```diff
@@ -64,8 +65,8 @@ public String canonicalString() {
     } else {
       builder.append("a:");
     }
-    for (FieldValue indexComponent : position) {
-      builder.append(indexComponent.toString());
+    for (Value indexComponent : position) {
+      builder.append(ProtoValues.canonicalId(indexComponent));
```
wilhuff: While we're fixing the format here, should we separate these values with something?

(Optional; may not be worth it. Just a thought because it may make these more readable.)

firebase-firestore/src/main/java/com/google/firebase/firestore/core/InFilter.java
```diff
@@ -14,23 +14,24 @@
...
 public class InFilter extends FieldFilter {
-  InFilter(FieldPath field, ArrayValue value) {
+  InFilter(FieldPath field, Value value) {
     super(field, Operator.IN, value);
+    hardAssert(ProtoValues.isArray(value), "InFilter expects an ArrayValue");
   }
 
   @Override
   public boolean matches(Document doc) {
     FieldValue other = doc.getField(getField());
```
wilhuff: `doc.getFieldProto(getField())`?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/KeyFieldFilter.java
```diff
@@ -14,20 +14,25 @@
...
+  KeyFieldFilter(FieldPath field, Operator operator, Value value) {
     super(field, operator, value);
+    hardAssert(ProtoValues.isReferenceValue(value), "KeyFieldFilter expects a ReferenceValue");
   }
 
   @Override
   public boolean matches(Document doc) {
-    ReferenceValue referenceValue = (ReferenceValue) getValue();
-    int comparator = doc.getKey().compareTo(referenceValue.getKey());
+    DocumentKey referencedKey = DocumentKey.fromName(getValue().getReferenceValue());
```
wilhuff: This seems like something we could store as a field. It doesn't change for every document we match.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/KeyFieldInFilter.java
```diff
@@ -17,26 +17,27 @@
...
     }
   }
 
   @Override
   public boolean matches(Document doc) {
-    ArrayValue arrayValue = (ArrayValue) getValue();
-    for (FieldValue refValue : arrayValue.getValues()) {
-      if (doc.getKey().equals(((ReferenceValue) refValue).getKey())) {
+    for (Value refValue : getValue().getArrayValue().getValuesList()) {
+      DocumentKey referencedKey = DocumentKey.fromName(refValue.getReferenceValue());
```
wilhuff: Similarly we could do this conversion once in the constructor and keep a list of document keys as a field.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -157,8 +164,8 @@ private static boolean objectEquals(Value left, Value right) {
...
+    for (Value haystackEl : haystack.getValuesList()) {
```
wilhuff: Nit: Java doesn't tend to abbreviate like this. Consider naming this just `element`, or `haystackElement` if you want to preserve the binding to the haystack.

firebase-firestore/ktx/src/test/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -36,12 +36,12 @@ public static FieldValue wrap(Object value) {
     UserDataReader dataReader = new UserDataReader(databaseId);
     // HACK: We use parseQueryValue() since it accepts scalars as well as arrays / objects, and
     // our tests currently use wrap() pretty generically so we don't know the intent.
-    return dataReader.parseQueryValue(value);
+    return new FieldValue(dataReader.parseQueryValue(value));
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed.  This conversion should go away entirely in a follow-up PR too.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
schmidt-sebastian: Only ArrayContainsFilter does the check right at the callsite. All other filters verify the value in the constructor, which is IMHO the better thing to do since the assert will include a stacktrace that should indicate where the incorrect value was accessed.

Furthermore, ArrayTransformOperation doesn't have a Value, it only operates on ArrayValue.Builder. For that reason, I left this as is for now.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Bound.java
schmidt-sebastian: Added comma separation.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/InFilter.java
schmidt-sebastian: Yup! Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/KeyFieldFilter.java
schmidt-sebastian: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/KeyFieldInFilter.java
schmidt-sebastian: Done. This simplifies the code a bit since I can now use List.contains().

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
schmidt-sebastian: Used `haystackElement`

PR #1207, opened 2020-02-06
Drop remaining primitive FieldValue types

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
```diff
@@ -41,34 +41,47 @@
   public static final Value NULL_VALUE =
       Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build();
 
+  /** The order of types in Firestore; this order is defined by the backend. */
+  private static final int TYPE_ORDER_NULL = 0;
```
wilhuff: public methods that return private values make little sense. `typeOrder` should either be private or these constants should be public.

firebase-firestore/src/test/java/com/google/firebase/firestore/UserDataWriterTest.java
```diff
@@ -291,4 +287,8 @@ public void testRejectsJavaArrays() {
   private Object convertValue(FieldValue value) {
     return writer.convertValue(value.getProto());
   }
+
+  private void assertValueType(FieldValue value, Value.ValueTypeCase booleanValue) {
```
wilhuff: JUnit tends to put the expected value as the first argument. Consider reversing these to match.

firebase-firestore/src/test/java/com/google/firebase/firestore/UserDataWriterTest.java
```diff
@@ -259,12 +257,10 @@ private static ObjectValue fromMap(Object... entries) {
...
   public void testConvertsNestedObjects() {
     FieldValue actual = wrapObject("a", map("b", map("c", "foo"), "d", true));
-    ObjectValue expected =
-        fromMap(
-            "a",
-            fromMap(
-                "b", fromMap("c", StringValue.valueOf("foo")), "d", BooleanValue.valueOf(true)));
-    assertEquals(expected, actual);
+    ObjectValue.Builder expected = ObjectValue.newBuilder();
+    expected.set(field("a.b.c"), valueOf("foo"));
```
wilhuff: Optional: You could chain these (`ObjectValue expected = newBuilder.set.set.build()`)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ProtoValues.java
schmidt-sebastian: Made them public.

firebase-firestore/src/test/java/com/google/firebase/firestore/UserDataWriterTest.java
schmidt-sebastian: Changed this. Note that I have a similar helper for canonical IDs, which I kept the same as the callsites look much cleaner in reverse argument order.

firebase-firestore/src/test/java/com/google/firebase/firestore/UserDataWriterTest.java
schmidt-sebastian: It felt cleaner to spread these assignments over multiple lines. Kept as is.

PR #1208, opened 2020-02-06
Make FieldValue a test-only class

schmidt-sebastian:
With all the prior PRs done, we now no longer need FieldValue. I moved it to a testHelpers for now to keep the diff smaller. The next step would be to remove it altogether and to return Value types from `wrap`.


firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ServerTimestampValue.java
```diff
@@ -37,7 +37,7 @@
  * <li>They sort after all TimestampValues. With respect to other ServerTimestampValues, they sort
  *     by their localWriteTime.
  */
-public final class ServerTimestampValue extends FieldValue {
+public final class ServerTimestampValue {
```
wilhuff: I'm surprised this works. Is the only usage of this in `UserDataWriter`?

schmidt-sebastian: The class is only used in `UserDataWriter` and `ProtoValues`. `isServerTimestamp()` is used in a couple other places.

PR #1212, opened 2020-02-06
Dropping guava and appcompat dependency from FIS SDK.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -78,7 +78,15 @@
...
 
   private static final ThreadFactory threadFactory =
-      new ThreadFactoryBuilder().setNameFormat("firebase-installations-executor-%d").build();
+      new ThreadFactory() {
+        private final AtomicInteger mCount = new AtomicInteger(1);
+
+        @Override
+        public Thread newThread(Runnable r) {
+          return new Thread(
+              r, String.format("firebase-installations-executor-%d", mCount.getAndIncrement()));
```
andirayo: Can we make this uppercase this time?
I know that would mean I have to submit a new CL in Google3...
Ok, as discussed offline:
Let's not do it.
We can discuss after the successful release.

PR #1214, opened 2020-02-06
Add new CrashlyticsReport data model

mrwillis21:
New data model is an AutoValue data class representing the
hierarchical Crashlytics report data structure, to be used for
passing report data to the new Transport API. It is serialized
via the firebase-encoders-json `@Encodable` annotation.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
@@ -17,16 +17,21 @@ package com.google.firebase.crashlytics.internal.model;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.auto.value.AutoValue;
-import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Application.Organization;
 import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Event;
-import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Event.Application;
 import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Event.Application.Execution.Thread.Frame;
 import com.google.firebase.encoders.annotations.Encodable;
 import com.google.firebase.encoders.annotations.Encodable.Field;
 import com.google.firebase.encoders.annotations.Encodable.Ignore;
 import java.nio.charset.Charset;
-import java.util.List;
 
+/**
+ * This class represents the data captured by and reported to Crashlytics.
+ *
+ * <p>It is an immutable value class implemented by AutoValue.
+ *
+ * @see <a
+ *     href="https://github.com/google/auto/tree/master/value">https://github.com/google/auto/tree/master/value</a>
+ */
 @Encodable
 @AutoValue
 public abstract class CrashlyticsReport {
@@ -58,11 +63,22 @@ public abstract class CrashlyticsReport {
   @NonNull
   public abstract Session getSession();
 
+  /**
+   * Augment an existing {@link CrashlyticsReport} with a given list of events.
+   *
+   * @return a new {@link CrashlyticsReport} with its events list set to the given list of events.
+   */
   @NonNull
-  public CrashlyticsReport withEvents(@NonNull List<Event> events) {
+  public CrashlyticsReport withEvents(@NonNull ImmutableList<Event> events) {
     return toBuilder().setSession(getSession().withEvents(events)).build();
   }
 
+  /**
+   * Augment an existing {@link CrashlyticsReport} with a given organization ID.
+   *
+   * @return a new {@link CrashlyticsReport} with its Session.Application.Organization object
+   *     containing the given organization ID.
+   */
   @NonNull
   public CrashlyticsReport withOrganizationId(@NonNull String organizationId) {
     return toBuilder().setSession(getSession().withOrganizationId(organizationId)).build();
@@ -127,11 +143,11 @@ public abstract class CrashlyticsReport {
 
     public abstract long getStartedAt();
 
-    @Nullable
-    public abstract User getUser();
+    @NonNull
+    public abstract Application getApp();
 
     @Nullable
-    public abstract Application getApp();
+    public abstract User getUser();
 
     @Nullable
     public abstract OperatingSystem getOs();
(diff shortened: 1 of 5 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
```diff
@@ -0,0 +1,854 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.crashlytics.internal.model;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Application.Organization;
```
bklimt: I'm confused. Isn't that already in this file?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
```diff
@@ -0,0 +1,854 @@
...
+import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Event.Application.Execution.Thread.Frame;
+import com.google.firebase.encoders.annotations.Encodable;
+import com.google.firebase.encoders.annotations.Encodable.Field;
+import com.google.firebase.encoders.annotations.Encodable.Ignore;
+import java.nio.charset.Charset;
+import java.util.List;
+
+@Encodable
+@AutoValue
+public abstract class CrashlyticsReport {
```
bklimt: This top-level class really need a javadoc comment explaining how it works. It looks like the rest of the classes are mostly AutoValue representations of the report model, so maybe it's not necessary to have comments for all of them, but definitely at least one at the top.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
```diff
@@ -0,0 +1,854 @@
...
+  public abstract String getBuildVersion();
+
+  @NonNull
+  public abstract String getDisplayVersion();
+
+  @NonNull
+  public abstract Session getSession();
+
+  @NonNull
+  public CrashlyticsReport withEvents(@NonNull List<Event> events) {
```
bklimt: It's kinda confusing how this lets you build reports but not using the builder itself. Why wouldn't callers just use the builder directly?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/ImmutableList.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.RandomAccess;
+
+/** Immutable list implementation for use with AutoValue */
+public final class ImmutableList<E> implements List<E>, RandomAccess {
```
bklimt: Did you write this from scratch just for this use case?

Why not just use Collections.unmodifiableList directly and call it good enough?

mrwillis21: `Collections.unmodifiableList` does not actually expose a type -- it just throws `RuntimeException`s if you try to modify it. At compile-time, it's just a `List`. For AutoValue's setter's, I needed an actual type that would guarantee immutability: https://github.com/google/auto/blob/master/value/userguide/howto.md#mutable_property

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
mrwillis21: This allows you to augment an existing report object with a list of events while maintaining all of the other values already set on it. https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#withers

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
mrwillis21: Agreed. Will fix.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/ImmutableList.java
bklimt: But this class implements List<E> and has:

```
  public E set(int index, @NonNull E element) {
    return immutableList.set(index, element);
  }
```

So how is that any different than using `Collections.unmodifiableList` directly?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
mrwillis21: It is. The `import` is necessary in order to use the `Organization` type inside `Session` (for `withOrganizationId`. Without this import I'd have to fully qualify its usage there.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/ImmutableList.java
mrwillis21: This type guarantees that the `List` passed into any of the `CrashlyticsReport` object setters is immutable.

Otherwise the list setters would look like (e.g.): `setEvents(List<Event>)` and would rely on callers to be sure and wrap their mutable lists with `Collections.unmodifiableList`. If they did not, the data object would accidentally be mutable.

mrwillis21: Essentially the goal was to have a simple `ImmutableList` concrete type without pulling in Guava.

bklimt: I see. I guess that makes sense. And I see the other uses of `ImmutableList` in Firebase are pulled in from Guava for tests, so there doesn't appear to be anything we can easily re-use. So, did you write this class from scratch?

mrwillis21: Not really. Just created a new `ImmutableList<T>` with an underlying `Collections.unmodifiableList` property, then used `Generate->Delegate methods` in Android Studio. :)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
mrwillis21: I'll add comments to these `with` methods to clarify that.

PR #1219, opened 2020-02-07
Changelog for FieldValue change

schmidt-sebastian:
Changelog PRs are the most fun.

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,9 @@
-# Unreleased (21.4.0)
+# Unreleased 
+- [changed] Internal changes to the in-memory representation of Firestore
+  documents. You should see an overall decrease in object allocations and 
+  a performance improvement for `DocumentSnapshot.getData()` calls.
+
+# 21.4.0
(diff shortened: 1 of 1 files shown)


firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-# Unreleased (21.4.0)
+# Unreleased 
+- [changed] Internal changes to the in-memory representation of Firestore
```
wilhuff: This should be a past-tense sentence. Something like:

> Changed the in-memory representation of Firestore documents to reduce memory allocation and improve performance. Users of `DocumentSnapshot.getData()` and `DocumentSnapshot.toObject()` will see the biggest improvement.

WDYT?

schmidt-sebastian: Changed to:

```
Changed the in-memory representation of Firestore documents to
  reduce memory allocations and improve performance. Calls to 
  `DocumentSnapshot.getData()` and `DocumentSnapshot.toObject()` will see
  the biggest improvement.
```

I would think that all our users use these methods, so I changed "users of" to "calls to".

PR #1221, opened 2020-02-07
Fix unit test running from within Android Studio

wilhuff:
Somewhat counterintuitively, robolectric has a dependency on protobuf-java, and occasionally our tests will all fail with messages about unable to find methods like `makeImmutable()` on our messages. This only happens with Android Studio's JUnit runner, not in Gradle.

We don't use any features that require this, so just exclude that dependency.

Also fix the runner definition to actually find tests.


schmidt-sebastian:
FWIW, I see this only in AS 3.5. In 3.4.2, I can run the tests.

wilhuff:
Storage tests are flaking and are unrelated to this change. Submitting.

PR #1223, opened 2020-02-07
Add CrashlyticsReport JSON de/serialization

mrwillis21:
Uses firebase-encoders-json DataEncoder for serialization and
uses JsonReader manually for deserialization, as there is not yet
a firebase-decoders-json.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/serialization/CrashlyticsReportJsonTransform.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/serialization/CrashlyticsReportJsonTransform.java
@@ -1,0 +1,565 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.crashlytics.internal.model.serialization;
+
+import android.util.Base64;
+import android.util.JsonReader;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/serialization/CrashlyticsReportJsonTransform.java
```diff
@@ -0,0 +1,565 @@
...
+    return CRASHLYTICS_REPORT_JSON_ENCODER.encode(event);
+  }
+
+  @Nullable
+  public CrashlyticsReport reportFromJson(@NonNull String json) {
+    try (JsonReader jsonReader = new JsonReader(new StringReader(json))) {
+      return parseReport(jsonReader);
+    } catch (IOException e) {
+      // TODO: Handle the IOException
+      return null;
```
bklimt: I don't love this, but I won't push back too hard on it. It's just harder to make sure you deal with null upstream than to deal with an Exception.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/serialization/CrashlyticsReportJsonTransform.java
```diff
@@ -0,0 +1,565 @@
...
+          jsonReader.skipValue();
+          break;
+      }
+    }
+    jsonReader.endObject();
+
+    return builder.build();
+  }
+
+  private static CrashlyticsReport.Session.OperatingSystem parseOs(JsonReader jsonReader) throws IOException {
```
bklimt: "Os" you're killing me. I'd argue that in a line with `IOException`, we should be able to call this `parseOS`. :-)

But I guess you want to match it match the name in the object, so whatever.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/serialization/CrashlyticsReportJsonTransform.java
mrwillis21: Agreed, hence the TODO.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/serialization/CrashlyticsReportJsonTransform.java
mrwillis21: Yeah. :man_shrugging:

PR #1224, opened 2020-02-07
Add CrashlyticsReport persistence

mrwillis21:
Serializes to JSON files on disk


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistenceTest.java
```diff
@@ -0,0 +1,261 @@
...
+                .setProximityOn(true)
+                .setRamUsed(10000000)
+                .build())
+        .build();
+  }
+
+  private static ImmutableList<Frame> makeTestFrames() {
+    return ImmutableList.from(
+        Frame.builder()
+            .setPc(0)
```
bklimt: It's probably a better test if these aren't all 0.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -0,0 +1,236 @@
...
+      while ((read = fileInput.read(readBuffer)) > 0) {
+        bos.write(readBuffer, 0, read);
+      }
+      return new String(bos.toByteArray(), UTF_8);
+    } catch (IOException e) {
+      return null;
+    }
+  }
+
+  private static void recursiveDelete(File file) {
```
bklimt: These methods already deleted elsewhere in the SDK, correct? So is the plan to delete the other versions?

mrwillis21: Ultimately, yes. My goal in the long term is to reduce "Utils" classes, so for now I went with the Go proverb "A little copying is better than a little dependency."

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistenceTest.java
mrwillis21: Good call, I'll update this.

PR #1225, opened 2020-02-07
Remove value package, rename ProtoValues

schmidt-sebastian:
FYI: Copyright dates were adjusted by Android Studio plugin.


schmidt-sebastian:
> I think you can revert the android studio copyright changes. It's just looking at whether or not the file is new without considering that it was moved from the other location.

Done. I debated keeping the changes since the files are basically completely rewritten.

PR #1227, opened 2020-02-07
Send CrashlyticsReports through DataTransport

mrwillis21:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSender.java
```diff
@@ -0,0 +1,91 @@
...
+  private static final String CRASHLYTICS_ENDPOINT =
+      mergeStrings("hts/cahyiseot-agolai.o/1frlglgc/aclg", "tp:/rsltcrprsp.ogepscmv/ieo/eaybtho");
+  private static final String CRASHLYTICS_API_KEY =
+      mergeStrings("AzSBpY4F0rHiHFdinTvM", "IayrSTFL9eJ69YeSUO2");
+  private static final String CRASHLYTICS_TRANSPORT_NAME = "FIREBASE_CRASHLYTICS_REPORT";
+
+  private final Transport<CrashlyticsReport> transport;
+  private final SendCallback<CrashlyticsReport> sendCallback;
+
+  static DataTransportCrashlyticsReportSender create(
```
bklimt: Is there a good reason to make this new callback type here instead of using a `Task<CrashlyticsReport>`?

It seems like this is the perfect use case for a Promise -- a result or error that gets resolved at most once.

mrwillis21: I'd originally written this to map to the underlying callback, but I'll take another look. I think you're right.

mrwillis21: Okay, talked about this in person - going to hold off on this for now, but may move the loop up a level and have a `sendReport` method that returns a single `Task`. Will revisit after all the "thin thread" changes are in.

PR #1229, opened 2020-02-11
feat(common-ktx): add FirebaseOptions builder KTX

rosariopfernandes:
This PR should add a FirebaseOptions builder extension function, so that FirebaseOptions can be created like:

```kotlin
val options = firebaseOptions {
            apiKey = "API_KEY"
            applicationId = "APP_ID"
            databaseUrl = "DB_URL"
            gcmSenderId = "GCM_ID"
            projectId = "123"
            storageBucket = "my-bucket.appspot.com"
}
```

Please note that this PR also adds the missing java getters to `FirebaseOptions.Builder`.

### DIFF
--- a/firebase-common/api.txt
+++ b/firebase-common/api.txt
@@ -33,6 +33,12 @@ package com.google.firebase {
     ctor public FirebaseOptions.Builder();
     ctor public FirebaseOptions.Builder(@NonNull com.google.firebase.FirebaseOptions);
     method @NonNull public com.google.firebase.FirebaseOptions build();
+    method @NonNull public String getApiKey();
+    method @NonNull public String getApplicationId();
+    method @Nullable public String getDatabaseUrl();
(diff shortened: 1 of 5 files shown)


rlazo:
Hi @rosariopfernandes Thanks for your proposal. AFAIR, and @vkryachko can correct me if I'm wrong, we did discuss about adding support for this in the past, but we ended up not implementing it because of friction. In this case, the additional friction to create FirebaseOptions is by design. Creating your own options is something better let to cases in which devs really know the possible consequences, since it can introduce tricky bugs to debug.

rosariopfernandes:
@rlazo totally makes sense to me.

This came to mind because I was using this builder to write some automated tests. But later on I noticed this is not really commonly used (you don't see it in many open source projects or blog posts) and wanted to close the PR, but wasn't going to do it before getting some feedback first. Thanks!

PR #1231, opened 2020-02-11
Do not set test experiments as active experiments

JasonAHeron:


firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -317,8 +317,11 @@ private boolean shouldIgnoreCache(String event) {
       campaignId = content.getExperimentalPayload().getCampaignId();
       campaignName = content.getExperimentalPayload().getCampaignName();
       // At this point we set the experiment to become active in analytics.
-      abtIntegrationHelper.setExperimentActive(
-          content.getExperimentalPayload().getExperimentPayload());
+      // As long as it's not a test experiment.
+      if (!content.getIsTestCampaign()) {
+        abtIntegrationHelper.setExperimentActive(
```
MeghaB: can we add a debug log if it's a test one but not active? could be useful for devs debugging their experiment setup?

JasonAHeron: A test experiment will never be active and never be running. As far as FIAM is concerned nothing about test experiments will ever be reported to the abt sdk. The test device helper already does log a message indicating that a test campaign was received and the device is now set in test mode. Should I add an extra message or does the current logging sound good?

PR #1233, opened 2020-02-12
Add Crashlytics data capture class

mrwillis21:
Handles capturing appropriate data from the underlying OS, build
information, and exception information and constructing
CrashlyticsReport and Event objects from it.

Moved AppData and BatteryState utility classes into the "internal"
package, since it was necessary to make them public anyway.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
@@ -1,0 +1,397 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.crashlytics.internal.common;
+
+import android.app.ActivityManager;
+import android.app.ActivityManager.RunningAppProcessInfo;
+import android.content.Context;
+import android.os.Build;
+import android.os.Build.VERSION;
+import android.os.Environment;
+import android.os.StatFs;
+import android.text.TextUtils;
+import com.google.firebase.crashlytics.BuildConfig;
+import com.google.firebase.crashlytics.internal.model.CrashlyticsReport;
+import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Architecture;
+import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Event;
+import com.google.firebase.crashlytics.internal.model.CrashlyticsReport.Session.Event.Application.Execution;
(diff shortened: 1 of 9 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
```diff
@@ -0,0 +1,432 @@
...
+        String modelClass);
+
+    void endSessionDeviceDataCapture();
+
+    void endSessionDataCapture();
+
+    void endReportDataCapture();
+  }
+
+  /** Defines methods for handling data for a given exception event. */
```
bklimt: It seems like there should be some lifecycle that we should document. I mean, I assume that handlers can assume that a start comes before and end for the same event. But is there more to. Are there other assumptions about the order of events that we should make explicit?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
```diff
@@ -0,0 +1,432 @@
...
+ * parsing them, and passing them to handlers implemented by callers to use the captured information
+ * as they see fit.
+ */
+public class CrashlyticsReportDataCapture {
+
+  /**
+   * Defines methods for handling static report data gathered from the device, as well as system and
+   * build properties.
+   */
+  public interface ReportDataHandler {
```
bklimt: I find this model extremely odd. Normally when you have a data structure, you have something like a struct with nested fields, and you have one method that creates or populates the object, and then you pass that object to methods that do something with it. But here, instead of creating the object, we call a handler for every field in the object that would be set. Why does this need to be so complicated? Can't we just populate a Report object and hand it back, and let the other code manipulate it like any other object?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/Architecture.java
```diff
@@ -0,0 +1,65 @@
...
+
+package com.google.firebase.crashlytics.internal.model;
+
+import android.os.Build;
+import android.text.TextUtils;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+public enum Architecture {
```
bklimt: Android style is generally to use `IntDef`s instead of `enum`s, right? Personally, I prefer enums, but thought I'd point it out. If it were part of the public API, it would be a bigger deal.

mrwillis21:
All good feedback. Going to take a different approach here.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/Architecture.java
mrwillis21: Hmm, I guess it depends on which camp you're in (perfmatters/enummatters). :stuck_out_tongue_closed_eyes: But in this case I think an `IntDef` actually makes more sense.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
mrwillis21: Discussed this offline and decided to take a different approach.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
mrwillis21: New approach will eliminate this concern.

PR #1243, opened 2020-02-13
Return task from sendReport call

mrwillis21:
Rather than taking a callback in the constructor of the report
sender, just have the sendReport method return a task and let the
caller handle looping over multiple reports.

### DIFF
--- a/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSenderTest.java
+++ b/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSenderTest.java
@@ -14,82 +14,96 @@
 
 package com.google.firebase.crashlytics.internal.send;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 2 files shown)


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSenderTest.java
```diff
@@ -14,82 +14,96 @@
...
-    final List<CrashlyticsReport> reports = new ArrayList<>();
-    reports.add(report1);
-    reports.add(report2);
+    final Task<CrashlyticsReport> send1 = reportSender.sendReport(report1);
+    final Task<CrashlyticsReport> send2 = reportSender.sendReport(report2);
 
-    reportSender.sendReports(reports);
-    verify(mockCallback).onSendComplete(report1, null);
-    verify(mockCallback).onSendComplete(report2, null);
+    assertTrue(send1.isSuccessful());
```
bklimt: It would be better to call `Tasks.await(send1);` before checking the result of the task. Even if the test happens to resolve the task synchronously right now, there are lots of reasons why it might become async in the future, and that way the tests will keep working as expected. Same for similar usages below.

mrwillis21: Good call. I'll add that. Thanks!

PR #1246, opened 2020-02-14
Add FirebaseCrashlyticsReportManager

mrwillis21:
Handles the interaction between Crashlytics lifecycle events and
crash reports (capture, persistence, and sending).

### DIFF
--- a/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
+++ b/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
@@ -23,17 +23,17 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.android.gms.tasks.Task;
+import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.crashlytics.internal.model.CrashlyticsReport;
 import com.google.firebase.crashlytics.internal.persistence.CrashlyticsReportPersistence;
 import com.google.firebase.crashlytics.internal.send.DataTransportCrashlyticsReportSender;
 import com.google.firebase.crashlytics.internal.settings.model.AppSettingsData;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 import org.junit.Before;
 import org.junit.Test;
-import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
@@ -67,7 +67,7 @@ public class FirebaseCrashlyticsReportManagerTest {
     final CrashlyticsReport mockReport = mock(CrashlyticsReport.class);
     when(dataCapture.captureReportData(anyString(), anyLong())).thenReturn(mockReport);
 
-    reportManager.onSessionBegin(sessionId);
+    reportManager.onBeginSession(sessionId);
 
     verify(dataCapture).captureReportData(sessionId, timestampSeconds);
     verify(reportPersistence).persistReport(mockReport);
@@ -88,7 +88,7 @@ public class FirebaseCrashlyticsReportManagerTest {
             any(Throwable.class), any(Thread.class), anyString(), anyLong(), anyInt(), anyInt()))
         .thenReturn(mockEvent);
 
-    reportManager.onSessionBegin(sessionId);
+    reportManager.onBeginSession(sessionId);
     reportManager.onFatalEvent(exceptionEvent, eventThread);
 
     verify(dataCapture)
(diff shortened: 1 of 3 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public void onSessionEnd() {
+    currentSessionId = null;
+  }
+
+  public void onSessionsFinalize() {
+    reportPersistence.finalizeReports(currentSessionId);
+  }
+
+  public void onReportSend(AppSettingsData appSettingsData) {
```
bklimt: Since this class isn't an interface, and doesn't implement an interface, it's a little weird for all of these methods to be named like they are events. For example, why is this `onReportSend` instead of `sendReport`?

If you are planning on making this into an interface later, then maybe it's okay, but still something to think about.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -0,0 +1,102 @@
...
+    final List<CrashlyticsReport> reportsToSend = reportPersistence.loadFinalizedReports();
+    for (CrashlyticsReport report : reportsToSend) {
+      reportsSender
+          .sendReport(report.withOrganizationId(appSettingsData.organizationId))
+          .addOnCompleteListener(this::onReportSendComplete);
```
bklimt: It's usually kind of a smell to use `addOnCompleteListener` instead of `continueWith`, but it's not obviously wrong here. The implications are:
1. All of the sends will run in parallel.
2. This caller won't ever know if they succeeded or failed (or even finished).
3. The callback will happen on the UI Thread.

Implication #2 seems intentional, and #3 seems unintentional. I'm not sure about #1. It might be a good idea to add a comment about the threading expectations of this method.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: I actually had planned on making it an interface later. :)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: Hmm, I was just following the instructions here: https://developers.google.com/android/guides/tasks

Definitely good points. I'll follow up with you on this.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
```diff
@@ -0,0 +1,186 @@
...
+    return mockTask(result, null, completionListener);
+  }
+
+  private Task<CrashlyticsReport> mockFailedTask(
+      Exception exception, ArgumentCaptor<OnCompleteListener> completionListener) {
+    return mockTask(null, exception, completionListener);
+  }
+
+  @SuppressWarnings("unchecked")
+  private Task<CrashlyticsReport> mockTask(
```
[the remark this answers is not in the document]
mrwillis21: Because I'm not super familiar with all the ins and outs of the Tasks API. :) Thanks for letting me know about these! Depending on what I do with the completion callback, these may change anyhow.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -0,0 +1,102 @@
...
+            MAX_CHAINED_EXCEPTION_DEPTH);
+
+    reportPersistence.persistEvent(capturedEvent, currentSessionId);
+  }
+
+  public void onSessionEnd() {
+    currentSessionId = null;
+  }
+
+  public void onSessionsFinalize() {
```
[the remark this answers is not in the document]
mrwillis21: :+1:

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
```diff
@@ -0,0 +1,156 @@
...
+    when(reportSender.sendReport(mockReport1)).thenReturn(successfulTask);
+    when(reportSender.sendReport(mockReport2)).thenReturn(failedTask);
+
+    reportManager.onSendReports(appSettings);
+
+    verify(reportSender).sendReport(mockReport1);
+    verify(reportSender).sendReport(mockReport2);
+
+    try {
+      Tasks.await(successfulTask);
```
bklimt: This can't be right. `successfulTask` was already declared to be successful, so this isn't going to do anything. I know I said before that it's important to wait on a task to know the test is ready to proceed, but this task is created right here in this method, and there's no work that it waits on, so this really isn't doing anything. So, if there's something that needs to happen before the test continues, make sure to wait on that, but this seems like a no-op as it is.

mrwillis21: Cool, I actually wasn't sure about this point, so I just cargo-culted it from my other tests, but  I did think it was a bit odd. I'll remove it. :)

PR #1247, opened 2020-02-14
Remove D/FirebaseCrashlytics prefix from logs

mrwillis21:
Previously, this could be overridden by end users, but the
override has been removed, rendering this prefix superfluous.


mrwillis21:
Chatted about the changelog offline, decided against it.

PR #1249, opened 2020-02-15
Sample PR for skipping the IOException in CrossProcessLock. 

ankitaj224:
We need to rethink what should CrossProcessLock.acquire return when there is an exception.


firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -40,7 +41,8 @@ static CrossProcessLock acquire(Context appContext, String lockName) {
       FileLock lock = channel.lock();
       return new CrossProcessLock(channel, lock);
     } catch (IOException e) {
-      throw new IllegalStateException("exception while using file locks, should never happen", e);
+      Log.e("CrossProcessLock", e.getMessage());
+      return null;
```
andirayo: Won't this fail with a NullPointerException somewhere else?

Issue #1251, opened 2020-02-17
Add coroutines Flow extension functions for Firestore


martinbonnin:
## What feature would you like to see?

Firestore has [DocumentReference.addSnapshotListener](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentReference.html#addSnapshotListener) and  [Query.addSnapshotListener](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/Query.html#addSnapshotListener).

Both of these are callback based and it would be nice to have coroutines equivalents. Something like: 

```kotlin
fun DocumentReference.toFlow(): Flow<DocumentSnapShot> { 
      TODO()
}

fun Query.toFlow(): Flow<QuerySnapShot> { 
      TODO()
}
```

## How would you use it?

This would be useful to leverage coroutines to make the code look nicer. Also it can be used together with coroutines scopes to bind Firestore queries to the Android lifecycle.

aguatno:
Thank you for sending this feature request. We will keep this FR open and update it as we prioritize new features. Thank you.

svenjacobs:
Also looking forward to this! I'm using my own implementations of `toFlow()` for `DocumentReference` and `Query` but would really like to use some official solution :)

vkryachko:
This was added in firebase-firestore-ktx version 24.3.0

PR #1252, opened 2020-02-18
Add 'toFlow()' extensions to DocumentSnapshot and Query

michgauz:
This PR intends to satisfy this Feature Request : 
https://github.com/firebase/firebase-android-sdk/issues/1251


michgauz:
@googlebot I signed it!

samtstern:
@michgauz thanks for sending this!  I saw your discussion about this on Android Study Group.

Just to set some expectation, all public API changes go through our internal API review process, so if @vkryachko or one of our other Kotlin enthusiasts else decides to push this forward with this it could take some time for them to write up the internal document and schedule the review meeting.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -163,3 +161,35 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+
+/**
+ * Attach a snapshotListener to a DocumentReference and use it as a coroutine flow
+ */
+fun DocumentReference.toFlow() = callbackFlow {
+    val listener = addSnapshotListener { value, error ->
+        if (value != null && value.exists()) {
+            offer(value)
+        } else if (error != null) {
+            Logger.warn("DocumentReference:flow", error.message)
```
vkryachko: It's not clear to me what the best course of action here would be:
1. i.e. just logging the error is not general enough, as developers may want to react to them and don't want them to be swallowed
2. otoh `close()`ing the flow is not consistent with current Firestore's behavior, as it looks like listeners are not "de-registered" upon exceptions

cc @schmidt-sebastian any thoughts?

wilhuff: If the listen stream encounters an error we do currently close the stream. For example, if you have data-dependent rules in effect on the server like a group membership test and a write is performed that removes you from the group, this error is the one that will tell you that permission is denied and that should terminate the stream.

The flow should have the same behavior so this should propagate the error, not swallow it.

vkryachko: Thanks Gil,

@michgauz then we definitely want to call `close(error)` here.

Also, as @samtstern mentioned, we will have to have an internal API review for this before it can be merged.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -18,15 +18,13 @@ import androidx.annotation.Keep
 import com.google.firebase.FirebaseApp
 import com.google.firebase.components.Component
 import com.google.firebase.components.ComponentRegistrar
-import com.google.firebase.firestore.DocumentSnapshot
-import com.google.firebase.firestore.FieldPath
-import com.google.firebase.firestore.FirebaseFirestore
-import com.google.firebase.firestore.QueryDocumentSnapshot
-import com.google.firebase.firestore.QuerySnapshot
-import com.google.firebase.firestore.FirebaseFirestoreSettings
+import com.google.firebase.firestore.*
```
schmidt-sebastian: We generally avoid collapsing imports.

schmidt-sebastian:
@wu-hui Would you be willing to draft the API proposal document for this change?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
michgauz: Thanks for your review.
I've just taken into account your feedbacks and close the channel when receiving an error 👍

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
michgauz: I have reverted the collapsing imports

martinbonnin:
I have just been aware of this:https://github.com/Kotlin/kotlinx.coroutines/issues/974

Long story short, channel.offer might throw and we might need to catch any exception thrown. 
It's not 100% clear to me if it's required here since the firestore callbacks are run on the main thread so maybe that changes things but I'd still feel safer if we could catch them.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -163,3 +167,35 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
     override fun getComponents(): List<Component<*>> =
             listOf(LibraryVersionComponent.create(LIBRARY_NAME, BuildConfig.VERSION_NAME))
 }
+
+/**
+ * Attach a snapshotListener to a DocumentReference and use it as a coroutine flow
+ */
+fun DocumentReference.toFlow() = callbackFlow {
+    val listener = addSnapshotListener { value, error ->
```
martinbonnin: We should allow the user to specify the `metadataChanges`parameter:

```kotlin
fun DocumentReference.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE) = callbackFlow {
    val listener = addSnapshotListener(metadataChanges) { value, error ->
```

I'm unsure about the `executor` parameter. This would typically be handled by coroutine Dispatchers but maybe there's a use case ?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -163,3 +167,35 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+    awaitClose {
+        listener.remove()
+    }
+}
+
+/**
+ * Attach a snapshotListener to a Query and use it as a coroutine flow
+ */
+fun Query.toFlow() = callbackFlow {
+    val listener = addSnapshotListener { value, error ->
```
martinbonnin: Same as above, we should allow to pass the `metadataChanges` parameter

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
michgauz: I have just added the `MetadataChanges` parameter.
Thanks 👍

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
michgauz: 👍

rlazo:
> I have just been aware of this:[Kotlin/kotlinx.coroutines#974](https://github.com/Kotlin/kotlinx.coroutines/issues/974)
> 
> Long story short, channel.offer might throw and we might need to catch any exception thrown.
> It's not 100% clear to me if it's strictly required here since the firestore callbacks are run on the main thread so maybe that changes things but I'd still feel safer if we could catch them.

+1 The comments on that Issue make a very compelling case for dealing with the exception.

rlazo:
Another thing to consider it that the current implementation uses `callbackFlow` which is marked as `@ExperimentalCoroutinesApi` ([ref](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/callback-flow.html))  @samtstern @rosariopfernandes @vkryachko thoughts?

For reference, [androidx.lifecycle](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle) usage of flow is limited to Flow's stable API in non-test code.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -163,3 +168,51 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
             listOf(LibraryVersionComponent.create(LIBRARY_NAME, BuildConfig.VERSION_NAME))
 }
+
+/**
+ * Attach a snapshotListener to a DocumentReference and use it as a coroutine flow
+ * @param metadataChanges Indicates whether metadata-only changes
+ */
+fun DocumentReference.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE) = callbackFlow {
+    val listener = addSnapshotListener(metadataChanges) { value, error ->
+        if (value != null && value.exists()) {
```
svenjacobs, line 178: Why the check for `exists()`? What if I deliberately add a snapshot listener to a document that does not yet exist or if I want to be notified about deletions? Shouldn't I receive all updates so that I can handle them in my application accordingly?

michgauz: Yes you're right. I think I did it following the main use case (receive updates about an existing document reference), but I agree this could be useful too 👍

rlazo:
> Another thing to consider it that the current implementation uses `callbackFlow` which is marked as `@ExperimentalCoroutinesApi` ([ref](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/callback-flow.html)) @samtstern @rosariopfernandes @vkryachko thoughts?
> 
> For reference, [androidx.lifecycle](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle) usage of flow is limited to Flow's stable API in non-test code.

@michgauz thanks for the PR and sorry for the long wait! after talking with the team we've agreed that while `callbackFlow` is the way to go we should restrict ourselves to Flow's stable API.

Until `callbackFlow` is marked as stable, the pattern to use in Firebase to implement flows should be the same as the one used in [androidx.lifecycle.FlowLiveData LiveData.asFlow()](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle/lifecycle-livedata-ktx/src/main/java/androidx/lifecycle/FlowLiveData.kt#86) In short, using `Channel` and `Observer` instead.

rosariopfernandes:
@rlazo Sorry I missed your previous comment. I agree that sticking to the Stable APIs is better.

@michgauz digging into @rlazo 's suggestion of using `Channel` instead, I came up with the code below. PTAL and feel free to use it in your PR:
```kotlin
fun Query.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE): Flow<QuerySnapshot> = flow {
    val channel = Channel<QuerySnapshot>(Channel.CONFLATED)
    var listener: ListenerRegistration? = null
    withContext(Dispatchers.Main.immediate) {
        listener = addSnapshotListener(metadataChanges) { value, error ->
            value?.let { channel.offer(it) }
            error?.let { channel.close(it) }
        }
    }
    try {
        for (value in channel) {
            emit(value)
        }
    } finally {
        GlobalScope.launch {
            listener?.remove()
        }
    }
}
```

svenjacobs:
@rosariopfernandes You should really use `callbackFlow`. Your solution references `GlobalScope` which should be avoided. `callbackFlow` uses `Channel` internally. So you're basically reinventing the wheel 😉

LouisCAD:
`GloblaScope.launch` can be removed and just get the listener removal inside `withContext(Dispatchers.Main.immediate + NonCancellable)`. I think that's an acceptable solution until `callbackFlow` stabilizes.

svenjacobs:
Sorry, I didn't notice that the usage of `callbackFlow` is a problem. But instead of trying to work around it, I would just mark the function as experimental, too, until `callbackFlow` becomes stable.

rosariopfernandes:
@svenjacobs ah, I see. Thanks for pointing that out!

@LouisCAD sounds good! Thanks

michgauz:
Hello :) Thanks for your feedbacks. I update the PR with the @rosariopfernandes feat @LouisCAD solution

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -163,3 +173,51 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+fun Query.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE): Flow<QuerySnapshot> = flow {
+    val channel = Channel<QuerySnapshot>(Channel.CONFLATED)
+    var listener: ListenerRegistration? = null
+    withContext(Dispatchers.Main.immediate) {
+        listener = addSnapshotListener(metadataChanges) { value, error ->
+            value?.let { channel.offer(it) }
+            error?.let { channel.close(it) }
+        }
+    }
+    try {
```
LouisCAD, line 214:
```suggestion
    try {
        withContext(Dispatchers.Main.immediate) {
            listener = addSnapshotListener(metadataChanges) { value, error ->
                value?.let { channel.offer(it) }
                error?.let { channel.close(it) }
            }
        }
```
Fixes a leak of the listener that would occur in case of cancellation of the coroutine on exiting the `withContext` block.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -163,3 +173,51 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+fun DocumentReference.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE) = flow {
+    val channel = Channel<DocumentSnapshot>(Channel.CONFLATED)
+    var listener: ListenerRegistration? = null
+    withContext(Dispatchers.Main.immediate) {
+        listener = addSnapshotListener(metadataChanges) { value, error ->
+            value?.let { channel.offer(it) }
+            error?.let { channel.close(it) }
+        }
+    }
+    try {
```
LouisCAD, line 190:
```suggestion
    try {
        withContext(Dispatchers.Main.immediate) {
            listener = addSnapshotListener(metadataChanges) { value, error ->
                value?.let { channel.offer(it) }
                error?.let { channel.close(it) }
            }
        }
```
Fixes a leak of the listener that would occur in case of cancellation of the coroutine on exiting the `withContext` block.

rosariopfernandes:
Good news, everyone! :smiley: 
`callbackFlow` is stable in [Coroutines 1.5.0](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.5.0) :partying_face:

martinbonnin:
Any chance to get this merged ? 

For the record, my current implementation using the now stable `callbackFlow` is

```kotlin
/**
 * Attach a snapshotListener to a [DocumentReference] and use it as a coroutine [Flow]
 * @param metadataChanges Indicates whether metadata-only changes
 */
fun DocumentReference.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE) = callbackFlow {
    val listener = addSnapshotListener(metadataChanges) { value, error ->
        if (value != null) {
            trySendBlocking(value)
        }
        if (error != null) {
            cancel(CancellationException("Error getting DocumentReference snapshot", error))
        }
    }

    awaitClose { listener.remove() }
}

/**
 * Attach a snapshotListener to a [Query] and use it as a coroutine [Flow]
 * @param metadataChanges Indicates whether metadata-only changes
 */
fun Query.toFlow(metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE): Flow<QuerySnapshot> = callbackFlow {
    val listener = addSnapshotListener(metadataChanges) { value, error ->
        if (value != null) {
            trySendBlocking(value)
        }
        if (error != null) {
            cancel(CancellationException("Error getting Query snapshot", error))
        }
    }

    awaitClose { listener.remove() }
}
```

LouisCAD:
@martinbonnin Why not just use `trySend` instead of risking blocking the thread, which is likely to be the main one?

martinbonnin:
>@martinbonnin Why not just use trySend instead of risking blocking the thread, which is likely to be the main one?

Good question. I'm not sure TBH. I took inspiration from https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/callback-flow.html

svenjacobs:
Can we get this merged, please? 🙏🏼

thatfiredev:
I think all the necessary APIs are stable now.
But this will require us to update to Kotlin 1.6, which requires Gradle 6.1.1
This project is currently using Gradle 5.4, so I think we'll need to first update it while addressing all of the [breaking changes](https://docs.gradle.org/current/userguide/upgrading_version_5.html#changes_5.5), and only then we can get this merged.

martinbonnin:
I can try helping on the Gradle migration but I'm not super familiar with the Groovy `build.gradle`. Would you be ok moving the build scripts to `build.gradle.kts`?

svenjacobs:
> @martinbonnin Why not just use `trySend` instead of risking blocking the thread, which is likely to be the main one?

@LouisCAD I'm currently investigating problems reported by customers which claim they see outdated data. I have the feeling it has something to do with `trySend`. My current implementation of `toFlow` looks mostly like @martinbonnin version above, except I use `trySend` instead of `trySendBlocking`. When using `trySend`, couldn't the Flow lose values if the consumer is too slow (back-pressure)? Firestore snapshot listeners usually return the cached, offline version of a document first and return the synced, online version afterwards. If the consumer is still busy rendering the first result it would never receive the second result and the customer sees old data. Wouldn't it be more safe to use `trySendBlocking` here even if we risk blocking the main thread? Alternatively, could we use `.buffer(Channel.CONFLATED)` on the resulting Flow and keep using `trySend` to ensure that the Flow always emits the latest data?

svenjacobs:
I just noticed that the current implementation of this PR doesn't use `callbackFlow`. Since `callbackFlow` is stable now I guess we could revert the implementation to one of the initial versions that uses `callbackFlow`?

thatfiredev:
@svenjacobs That's right. We can now go back to the original `callbackFlow` implementation now that it is stable.
Also, the SDK is now using Kotlin 1.6.20 (after #3762) so I think there's no technical issues currently blocking this PR. 

@svenjacobs  Out of curiosity: What's the importance of using `buffer()` here? I've never heard of it before (probably because I don't use Kotlin Flows very often).

---

@michgauz Can you please update the PR to make use of `callbackFlow`? I think @svenjacobs has come up with a good proposal in [this comment](https://github.com/firebase/firebase-android-sdk/pull/1252#issuecomment-1134593926).

---

@wu-hui IIUC the internal proposal for this is now outdated. Do we need to get it through the API Review again? If so, I'd be happy to help writing the proposal.

svenjacobs:
> @svenjacobs  Out of curiosity: What's the importance of using `buffer()` here? I've never heard of it before (probably because I don't use Kotlin Flows very often).

I applied the `buffer` operator according to the recommendations of the API documentation of `callbackFlow` in regards to back-pressure. Please also see my [thoughts](https://github.com/firebase/firebase-android-sdk/pull/1252#issuecomment-1125707316) above. I'm curious what @LouisCAD thinks about this?

vkryachko:
> Also, this repo is now using Kotlin 1.6.20

Note that we had to revert the kotlin upgrade for now.

However I don't think that fact should block this PR, namely because: even if this PR is using experimental api's, we know that they are already stable in 1.6, which is the kotlin version devs will end up using in their apps anyway. So unless there have been any breaking changes since the kotlin version we use now and 1.6, it's should be safe to merge and release as is. wdyt? @thatfiredev

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: @schmidt-sebastian could that configuration be commited in the repo (.idea folder or something else?)

LouisCAD:
@martinbonnin Maybe it would make sense to merge the latest master branch into the branch of this PR to get whatever the currently used Kotlin version is?

LouisCAD:
BTW, I agree with the default to CONFLATED that can still be changed, it can be helpful to have control on this, especially for debugging.

martinbonnin:
> Maybe it would make sense to merge the latest master branch into the branch of this PR to get whatever the currently used Kotlin version is?

Damn, you're right, I somehow managed to rebase against an older `master` branch. I just rebased again but it's still `1.3.72` there apparently.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +173,71 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+ * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
+ */
+fun DocumentReference.toFlow(
+    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE,
+    bufferCapacity: Int? = Channel.CONFLATED
+): Flow<DocumentSnapshot?> {
+    val flow = callbackFlow {
+        val registration = addSnapshotListener(metadataChanges) { snapshot, exception ->
+            if (exception != null) {
+                cancel(CancellationException("Error getting DocumentReference snapshot", exception))
```
svenjacobs, line 194: The version of `cancel()` with the `message` and `exception` parameters that has been used in a previous commit just wraps the message in a `CancellationException`. I personally think that version is more readable but in the end it is just a matter of taste, I guess. Any reason why it was changed?

martinbonnin:
@svenjacobs thanks for the ping. Looks like Kotlin [was updated to 1.4.32 recently](https://github.com/firebase/firebase-android-sdk/pull/3855). Let me rebase this PR

martinbonnin:
@thatfiredev I think it's all good now. Tha PR uses coroutines `1.5.2` because it can be read with Kotlin `1.4` and has `callbackFlow`. It's still marked experimental in `1.5.2` but we know it's promoted to stable in `1.6` so it should be safe to use.

The only thing that fails now is smoke tests with:

```
com.google.firebase.database.DatabaseException: Firebase Database error: Permission denied
```

I'm not sure how to fix this?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,71 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+ * Transforms a [DocumentReference] into a coroutine [Flow]
+ *
+ * **Backpressure handling**: by default this method conflates items. If the consumer isn't fast enough,
+ * it might miss some values but is always guaranteed to get the latest value. Use [bufferCapacity]
+ * to change that behaviour
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
+ */
+fun DocumentReference.toFlow(
```
thatfiredev, line 186: We are considering calling this `DocumentReference.documentSnapshot` instead, to better describe the return type:

```kotlin
docRef.documentSnapshot.collect { snapshot ->
    // handle returned document
}
```

I would like to hear y'all thoughts.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,71 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+ * Transforms a [Query] into a coroutine [Flow]
+ *
+ * **Backpressure handling**: by default this method conflates items. If the consumer isn't fast enough,
+ * it might miss some values but is always guaranteed to get the latest value. Use [bufferCapacity]
+ * to change that behaviour
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
+ */
+fun Query.toFlow(
```
thatfiredev, line 220: Same comment as above: we're considering calling this `Query.querySnapshot`.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,71 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE,
+    bufferCapacity: Int? = Channel.CONFLATED
+): Flow<QuerySnapshot?> {
```
thatfiredev, line 223:
```suggestion
): Flow<QuerySnapshot> {
```

Does it really need to be nullable? At the moment, `trySend()` is only called with non-null snapshots.

thatfiredev (commented):
@martinbonnin Thanks for rebasing!
Don't worry about the smoke-tests, we'll look into it when we leave a final review.
[ @vkryachko I wonder why the experimental GH Actions are not running here. Are they not enabled for ktx modules? Or do we need to merge the main branch into this one? (instead of a rebase) ]

**Next Steps:**
Last week I brought back the internal proposal for this and I'm waiting for a final review. It usually takes 2-3 weeks. I'll come back with an update once it's approved.


It's been 2 years since this was first opened but we're finally close to getting it merged 🤞

vkryachko:
@thatfiredev I think it had to do with the following limitation https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks , I just approved the workflows.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: Not sure what the best practices are. It'd be worth investigating what other libs are doing. In all cases, we need to account for `metadataChanges` and `bufferCapacity` parameters so it can't really be a `val`. Maybe this instead?

```kotlin
docRef.snapshots().collect { snapshot ->
  // do something with snapshot
}
```

And if customization is required, this becomes:

```kotlin
docRef.snapshots(metadataChanges, bufferCapacity).collect { snapshot ->
  // do something with snapshot
}
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: Good point 👍 . Doc says

```
It's guaranteed that exactly one of value or error will be non-{@code null}.
```

So 👍 for making everything non-nullable, I'll push the change in a bit.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
svenjacobs: I'm with @martinbonnin here. It should be `snapshots()` or `documentSnapshots()`. Plural, because a `Flow` usually returns multiple values. The `document` in `documentSnapshots()` is redundant because it should be clear that a `DocumentReference` returns `DocumentSnapshot`. So `snapshots()` is my favorite candidate.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
svenjacobs: Same as [here](https://github.com/firebase/firebase-android-sdk/pull/1252#discussion_r938064128). I'm for `snapshots()`.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
thatfiredev: I agree it should be plural. 👍 
Sorry for the `val` confusion, that was a mistake on my side, I forgot to add parenthesis.

firebase-firestore/firebase-firestore.gradle
```diff
@@ -144,7 +144,7 @@ dependencies {
     implementation "io.grpc:grpc-protobuf-lite:$grpcVersion"
     implementation "io.grpc:grpc-okhttp:$grpcVersion"
     implementation "io.grpc:grpc-android:$grpcVersion"
-    implementation 'com.google.android.gms:play-services-basement:18.0.0'
+    api 'com.google.android.gms:play-services-basement:18.0.0'
```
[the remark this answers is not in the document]
thatfiredev: Sorry, I didn't understand the reason for this change and how it relates to the Flows. Are you  hoping to expose `FirebaseException`?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,67 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+ * Transforms a [DocumentReference] into a coroutine [Flow]
+ *
+ * **Backpressure handling**: by default this method conflates items. If the consumer isn't fast enough,
+ * it might miss some values but is always guaranteed to get the latest value. Use [bufferCapacity]
+ * to change that behaviour
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
+ */
+fun DocumentReference.toFlow(
```
thatfiredev, line 186:
```suggestion
fun DocumentReference.documentSnapshots(
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,67 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+ * Transforms a [Query] into a coroutine [Flow]
+ *
+ * **Backpressure handling**: by default this method conflates items. If the consumer isn't fast enough,
+ * it might miss some values but is always guaranteed to get the latest value. Use [bufferCapacity]
+ * to change that behaviour
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
+ */
+fun Query.toFlow(
```
thatfiredev, line 218:
```suggestion
fun Query.querySnapshots(
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,67 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+/**
+ * Transforms a [Query] into a coroutine [Flow]
+ *
+ * **Backpressure handling**: by default this method conflates items. If the consumer isn't fast enough,
+ * it might miss some values but is always guaranteed to get the latest value. Use [bufferCapacity]
+ * to change that behaviour
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
+ */
```
thatfiredev, line 217: I have a humble suggestion to improve the ref docs here, but it might not be final - I will ask a technical writer to review it.

```suggestion
/**
 * Starts listening to this query with the given options and emits its values via a [Flow].
 *
 * When a collection of the returned flow starts, an EventListener will be attached.
 * When a collection completes, the listener will be removed.
 *
 * Backpressure: by default the returned flow is conflated. If the consumer isn't fast enough,
 * it might miss some values but is always guaranteed to get the latest value. Use the [bufferCapacity] parameter
 * to change that behaviour.
 *
 * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
 * @param bufferCapacity the buffer capacity as in [Flow.buffer] or null to not buffer at all
 */
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +172,67 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
     override fun getComponents(): List<Component<*>> =
             listOf(LibraryVersionComponent.create(LIBRARY_NAME, BuildConfig.VERSION_NAME))
 }
+
+/**
+ * Transforms a [DocumentReference] into a coroutine [Flow]
+ *
+ * **Backpressure handling**: by default this method conflates items. If the consumer isn't fast enough,
+ * it might miss some values but is always guaranteed to get the latest value. Use [bufferCapacity]
+ * to change that behaviour
```
thatfiredev, line 181:
```suggestion
 * Starts listening to the document referenced by this `DocumentReference` with the given options and emits its values via a [Flow].
 *
 * When a collection of the returned flow starts, an EventListener will be attached.
 * When a collection completes, the listener will be removed.
 *
 * Backpressure: by default the returned flow is conflated. If the consumer isn't fast enough,
 * it might miss some values but is always guaranteed to get the latest value. Use the [bufferCapacity] parameter
 * to change that behaviour.
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -14,18 +14,28 @@
 
 package com.google.firebase.firestore.ktx
 
+import android.support.multidex.BuildConfig
```
thatfiredev, line 17: I understand you needed this change to build it locally, but its not the correct `BuildConfig` class. If I'm not wrong we rely on a `BuildConfig` class generated at build time under the same package name (`com.google.firebase.firestore.ktx`) which is why we don't have an import for it.

Can you please remove this import line?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
thatfiredev: Don't forget to run `./gradlew :firebase-firestore:ktx:generateApiTxtFile` after making these changes. That's the gradle task that generates our `api.txt` file.

thatfiredev (commented):
Good news, everyone: the internal proposal for this has been approved. 🥳
There are a couple of changes that need to be made before we merge it - I've left them as review comments bellow.

firebase-firestore/firebase-firestore.gradle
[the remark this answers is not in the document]
martinbonnin: The `firebase-firestore` module exposes `FirebaseException` publicly from `EventListener`:

```kotlin
public interface EventListener<T> {

  /**
   * {@code onEvent} will be called with the new value or the error if an error occurred. It's
   * guaranteed that exactly one of value or error will be non-{@code null}.
   *
   * @param value The value of the event. {@code null} if there was an error.
   * @param error The error if there was error. {@code null} otherwise.
   */
  void onEvent(@Nullable T value, @Nullable FirebaseFirestoreException error);
}
```
Without this, I'm getting this error:

Screenshot 2022-08-09 at 20 18 52

We could also add `implementation 'com.google.android.gms:play-services-basement:18.0.0'` to the `:firebase-firestore:kts` module but I believe the proposed change above is better as it will fix it for all consumers, even the non-ktx ones.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: I went with just `snapshots()` in [this commit](https://github.com/firebase/firebase-android-sdk/pull/1252/commits/288ae9aed3ea4691aac5efab5bcbdddd999e5af8). As @svenjacobs [said](https://github.com/firebase/firebase-android-sdk/pull/1252#discussion_r938064128), `documentSnapshots()` is more verbose and the "document" part is already in the "documentRef". Let me know if you want me to change it or feel free to amend.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: I'd skip mentioning `EventListener` as it's an implementation detail that's not relevant to the Kotlin caller.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: Good catch, that's done ✅  Turns out it wasn't even required for building.

firebase-firestore/firebase-firestore.gradle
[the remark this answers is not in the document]
thatfiredev: @martinbonnin Ah, I see. Thanks for explaining. 
I would prefer adding that implementation to firebase-firestore-ktx instead of exposing it on the base module.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
thatfiredev: My idea behind mentioning EventListener is in case a developer is trying to troubleshoot some unexpected behavior on the Flow collection. They can refer to EventListener in our documentation.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
thatfiredev: Sounds good to me.

firebase-firestore/firebase-firestore.gradle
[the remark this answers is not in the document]
martinbonnin: done in https://github.com/firebase/firebase-android-sdk/pull/1252/commits/f6e1aebb9f2d980fd77759fcdcd6e3d4a2bcfb6b

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: Works for me 👍 . 

PS: I believe "edits from maintainers" are enabled on this PR? Feel free to push to the branch directly or open a new from it if you want to do changes.

thatfiredev (approved):
@martinbonnin Thank you very much for being so patient during this review process.

The code changes look good to me. We'll just need a final review from an engineer + one from a tech writer and then we'll be ready to merge this.

---

@vkryachko can you please review this and approve the CI workflows?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +171,73 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+fun DocumentReference.snapshots(
+    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE,
+    bufferCapacity: Int? = Channel.CONFLATED
+): Flow<DocumentSnapshot> {
+    val flow = callbackFlow {
+        val registration = addSnapshotListener(metadataChanges) { snapshot, exception ->
+            if (exception != null) {
+                cancel(message = "Error getting DocumentReference snapshot", cause = exception)
+            } else if (snapshot != null) {
+                trySend(snapshot)
```
vkryachko, line 197: shouldn't it be trySendBlocking instead? here and below

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +171,73 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+            }
+        }
+        awaitClose { registration.remove() }
+    }
+
+    return if (bufferCapacity != null) {
+        flow.buffer(bufferCapacity)
```
vkryachko, line 204: Do we need to have an opinion on this or should we just let the developer do their own buffering?

svenjacobs: Please see my comment [here](https://github.com/firebase/firebase-android-sdk/pull/1252#issuecomment-1143800315) and previous comments. `buffer` is essential for back-pressure handling.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
vkryachko: see my response re `buffer`, as for the comment you linked:

> @martinbonnin Why not just use trySend instead of risking blocking the thread, which is likely to be the main one?

imo the fact that Firestore delivers events on the UI thread is an unfortunate(but necessary) historic artifact as it optimized for devEx before coroutines were available and switching to the UI thread was non-trivial, and most devs want to render UI when data changes so UI thread was chosen. However Firestore provides overloads for addSnapshotListener that takes an executor. Now in the coroutine and flow world, devs have explicit control of what thread they are collecting on, so we can afford to deliver events to the flow on any thread we like.

So I think it's a safer choice to be as unopinionated as possible and use a safe executor + trySendBlocking, i.e. AsyncTask.THREAD_POOL_EXECUTOR or `com.google.firebase.firestore.util.Executors.BACKGROUND_EXECUTOR`(which is more limited).

This will make our code safe by default, i.e. we will not "drop" events with CONFLATED, and given direct control to developers. wdyt?

cc @wu-hui @ehsannas

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
svenjacobs: If fine with this if it is documentation as you suggested.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
svenjacobs: Sounds good to me

martinbonnin: I've pushed a change to use `BACKGROUND_EXECUTOR`. 

2 questions though:

1. Isn't this the equivalent of `buffer(UNLIMITED)` in practice? It's not using the coroutines `buffer()` but I think it will enqueue `Runnable`s in the executors. So all in all it's also doing a choice for the user.
2. Is there any executor that could avoid a thread switch (immediate maybe?) ? I'm not familiar with the Firestore internals but if the code is running in a background thread already, feels like there's no need to dispatch?

vkryachko: 1. Yeah, I guess you're right, it may be suboptimal in this case. so direct executor sounds more appealing.
2. That's a good question, on the one hand I like the idea of using direct executor but I am not sure where the execution happen in that case. grpc io pool maybe? what is the risk of starving those threads in that case? it may not actually be an issue since the developer have full control and can avoid congestion of that pool.

Would love to hear thoughts from the firestore folks on this.

vkryachko: I chatted with Firestore folks offline, given that Firestore is single-threaded internally, they really don't want to run arbitrary user code on their internal thread, so BACKGROUND_EXECUTOR is preferred.

@martinbonnin thanks for accommodating all the changes

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
vkryachko: nit: this is not fixed apparently

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: Sounds good 👍

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
thatfiredev: I think it got fixed initially, but reverted in a follow-up commit.

martinbonnin: Oh gosh, it's back... I just removed it again

martinbonnin: Yea, I'll do a follow up PR to commit the codeStyle so that AS stops doing this automagically. I "just" need to figure out how the .gitignore syntax works...

martinbonnin: Here it is: https://github.com/firebase/firebase-android-sdk/pull/3984

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +171,49 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+/**
+ * Starts listening to the document referenced by this `DocumentReference` with the given options and emits its values via a [Flow].
+ *
+ * - When the returned flow starts being collected, an [EventListener] will be attached.
+ * - When the flow completes, the listener will be removed.
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ */
+fun DocumentReference.snapshots(
+    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE,
```
thatfiredev, line 184:
```suggestion
    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
```diff
@@ -162,3 +171,49 @@ class FirebaseFirestoreKtxRegistrar : ComponentRegistrar {
...
+/**
+ * Starts listening to this query with the given options and emits its values via a [Flow].
+ *
+ * - When the returned flow starts being collected, an [EventListener] will be attached.
+ * - When the flow completes, the listener will be removed.
+ *
+ * @param metadataChanges controls metadata-only changes. Default: [MetadataChanges.EXCLUDE]
+ */
+fun Query.snapshots(
+    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE,
```
thatfiredev, line 207:
```suggestion
    metadataChanges: MetadataChanges = MetadataChanges.EXCLUDE
```

thatfiredev (commented):
CI is failing because Kotlin 1.4 doesn't support trailling commas.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/Firestore.kt
martinbonnin: Yup, sorry, just realized that. It's fixed in https://github.com/firebase/firebase-android-sdk/pull/1252/commits/692114cc815a109667461a817d501ed7872bd3d5.

martinbonnin: Also interestingly this seemed to be a lint-only issue as Kotlin 1.4 is perfectly happy with the trailing coma

vkryachko: yeah, I think our version of ktlint is pretty ancient, we are planning to upgrade it eventually

thatfiredev: @vkryachko I think all comments have been addressed. Can you please approve the pending CI workflows and review the PR once again?

vkryachko (approved):

svenjacobs:
Amazing! Thanks to all contributors, especially @martinbonnin who did all the hard work in the end 👏🏼

thatfiredev:
Thank you very much @michgauz @martinbonnin @svenjacobs @LouisCAD for all the work done here! ❤️ 
I know it was a very long process and I'm glad you all patiently went through it with us.

I can't promise any dates, but this is going to be released on the next version of `firestore-ktx` and we'll be working on adding `Flow`s to more Firebase APIs. 🎉

ColtonIdle:
someone shud update Flow docs on android as they reference fbase https://developer.android.com/kotlin/flow#callback

thatfiredev:
This feature has been released in [firestore-ktx:24.3.0](https://firebase.google.com/support/release-notes/android#firestore-ktx_v24-3-0). 🎉 

-----

Thanks for the heads up @ColtonIdle :) I'll get that page updated.

PR #1253, opened 2020-02-18
Add support for non-fatal events

mrwillis21:


mrwillis21:
I know why the test is flaky, going to fix it in a follow up PR

PR #1255, opened 2020-02-19
Cap events-per-session in Crashlytics persistence

mrwillis21:
Replaces the concept of fatal/nonfatal in the persistence layer
with a persistence-specific concept of normal/high priority events.

This will be used to implement keeping a maximum number of logged
exceptions per session, without leaking information about specific
event types into the persistence layer.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
@@ -43,17 +43,19 @@ public class CrashlyticsReportPersistence {
 
   private static final Charset UTF_8 = Charset.forName("UTF-8");
 
-  private static final String WORKING_DIRECTORY_NAME = "fl";
+  private static final String WORKING_DIRECTORY_NAME = "report-persistence";
   private static final String OPEN_SESSIONS_DIRECTORY_NAME = "sessions";
-  private static final String FATAL_DIRECTORY_NAME = "fatal";
-  private static final String NON_FATAL_DIRECTORY_NAME = "non-fatal";
+  private static final String PRIORITY_REPORTS_DIRECTORY = "priority-reports";
+  private static final String REPORTS_DIRECTORY = "reports";
 
-  private static final String REPORT_FILE_NAME = "report.json";
+  private static final String REPORT_FILE_NAME = "report";
   private static final String EVENT_FILE_NAME_PREFIX = "event";
-  private static final String EVENT_FILE_NAME_FORMAT = EVENT_FILE_NAME_PREFIX + "%s.json";
-  private static final String EVENT_COUNTER_FORMAT = "%010d";
-
-  private static final String EVENT_TYPE_FATAL = "crash";
+  private static final int EVENT_COUNTER_WIDTH = 10; // String width of maximum positive int value
+  private static final String EVENT_COUNTER_FORMAT = "%0" + EVENT_COUNTER_WIDTH + "d";
+  private static final int EVENT_NAME_LENGTH =
+      EVENT_FILE_NAME_PREFIX.length() + EVENT_COUNTER_WIDTH;
+  private static final String PRIORITY_EVENT_SUFFIX = "_";
+  private static final String NORMAL_EVENT_SUFFIX = "";
 
   private static final CrashlyticsReportJsonTransform TRANSFORM =
       new CrashlyticsReportJsonTransform();
@@ -61,18 +63,21 @@ public class CrashlyticsReportPersistence {
   private final AtomicInteger eventCounter = new AtomicInteger(0);
 
   // Storage for sessions that are still being written to
-  private File openSessionsDirectory;
+  private final File openSessionsDirectory;
 
   // Storage for finalized reports
-  // Keep finalized reports organized by whether or not they contain a fatal event.
-  private File fatalReportsDirectory;
-  private File nonFatalReportsDirectory;
+  private final File priorityReportsDirectory;
+  private final File reportsDirectory;
 
-  public CrashlyticsReportPersistence(File rootDirectory) {
+  // TODO: Add settings override
+  private final int defaultMaxEventsToKeep;
+
+  public CrashlyticsReportPersistence(File rootDirectory, int defaultMaxEventsToKeep) {
     final File workingDirectory = new File(rootDirectory, WORKING_DIRECTORY_NAME);
     openSessionsDirectory = new File(workingDirectory, OPEN_SESSIONS_DIRECTORY_NAME);
-    fatalReportsDirectory = new File(workingDirectory, FATAL_DIRECTORY_NAME);
-    nonFatalReportsDirectory = new File(workingDirectory, NON_FATAL_DIRECTORY_NAME);
+    priorityReportsDirectory = new File(workingDirectory, PRIORITY_REPORTS_DIRECTORY);
+    reportsDirectory = new File(workingDirectory, REPORTS_DIRECTORY);
+    this.defaultMaxEventsToKeep = defaultMaxEventsToKeep;
   }
 
   public void persistReport(CrashlyticsReport report) {
@@ -82,7 +87,31 @@ public class CrashlyticsReportPersistence {
     writeTextFile(new File(sessionDirectory, REPORT_FILE_NAME), json);
   }
 
+  /**
+   * Persist an event for a given session with normal priority.
+   *
+   * <p>Only a certain number of normal priority events are stored per-session. When this maximum is
+   * reached, the oldest events will be dropped.
+   *
+   * @param event
+   * @param sessionId
+   */
   public void persistEvent(CrashlyticsReport.Session.Event event, String sessionId) {
+    persistEvent(event, sessionId, false);
+  }
+
+  /**
+   * Persist an event for a given session, specifying whether or not it is high priority.
+   *
+   * <p>Only a certain number of normal priority events are stored per-session. When this maximum is
+   * reached, the oldest events will be dropped. High priority events are not subject to this limit.
+   *
+   * @param event
+   * @param sessionId
+   * @param isHighPriority
+   */
+  public void persistEvent(
+      CrashlyticsReport.Session.Event event, String sessionId, boolean isHighPriority) {
     final File sessionDirectory = getSessionDirectoryById(sessionId);
     if (!sessionDirectory.isDirectory()) {
       // No open session for this ID
@@ -90,18 +119,16 @@ public class CrashlyticsReportPersistence {
       return;
     }
     final String json = TRANSFORM.eventToJson(event);
-    final String eventNumber =
-        String.format(Locale.US, EVENT_COUNTER_FORMAT, eventCounter.getAndIncrement());
-    final String fileName = String.format(EVENT_FILE_NAME_FORMAT, eventNumber);
+    final String fileName = generateEventFilename(isHighPriority);
     writeTextFile(new File(sessionDirectory, fileName), json);
+    trimEvents(sessionDirectory, defaultMaxEventsToKeep);
   }
 
   public void deleteFinalizedReport(String sessionId) {
     final List<File> reportFiles = new ArrayList<>();
     final FilenameFilter filter = (d, f) -> f.startsWith(sessionId);
-    // Could be in either fatal reports or non-fatal reports
-    reportFiles.addAll(getFilesInDirectory(fatalReportsDirectory, filter));
-    reportFiles.addAll(getFilesInDirectory(nonFatalReportsDirectory, filter));
+    reportFiles.addAll(getFilesInDirectory(priorityReportsDirectory, filter));
+    reportFiles.addAll(getFilesInDirectory(reportsDirectory, filter));
     for (File reportFile : reportFiles) {
       reportFile.delete();
     }
@@ -133,15 +160,15 @@ public class CrashlyticsReportPersistence {
             TRANSFORM.reportFromJson(readTextFile(new File(sessionDirectory, REPORT_FILE_NAME)));
         final String sessionId = report.getSession().getIdentifier();
         final List<Event> events = new ArrayList<>();
-        boolean hasFatal = false;
+        boolean isHighPriorityReport = false;
         for (File eventFile : eventFiles) {
           final Event event = TRANSFORM.eventFromJson(readTextFile(eventFile));
-          hasFatal = hasFatal || event.getType().equals(EVENT_TYPE_FATAL);
+          isHighPriorityReport = isHighPriorityReport || isHighPriorityEvent(eventFile.getName());
           events.add(event);
         }
         // FIXME: If we fail to parse the events, we'll need to bail.
         final File outputDirectory =
-            prepareDirectory(hasFatal ? fatalReportsDirectory : nonFatalReportsDirectory);
+            prepareDirectory(isHighPriorityReport ? priorityReportsDirectory : reportsDirectory);
         writeTextFile(
             new File(outputDirectory, sessionId),
             TRANSFORM.reportToJson(report.withEvents(ImmutableList.from(events))));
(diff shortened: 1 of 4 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -43,36 +43,41 @@
...
+  private static final String EVENT_COUNTER_FORMAT = "%0" + EVENT_COUNTER_WIDTH + "d";
+  private static final int EVENT_NAME_LENGTH =
+      EVENT_FILE_NAME_PREFIX.length() + EVENT_COUNTER_WIDTH;
+  private static final String PRIORITY_EVENT_SUFFIX = "_";
+  private static final String NORMAL_EVENT_SUFFIX = "";
 
   private static final CrashlyticsReportJsonTransform TRANSFORM =
       new CrashlyticsReportJsonTransform();
 
   private final AtomicInteger eventCounter = new AtomicInteger(0);
```
jakeouellette: What resets this back to zero? Is this "Persistence" class recreated before it gets sufficiently large?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -151,22 +178,55 @@ public void finalizeReports(String currentSessionId) {
...
+    return allReports;
+  }
+
+  private static boolean isHighPriorityEvent(String eventFileName) {
+    return eventFileName.endsWith(PRIORITY_EVENT_SUFFIX);
+  }
+
+  private String generateEventFilename(boolean isHighPriority) {
+    final String eventNumber =
+        String.format(Locale.US, EVENT_COUNTER_FORMAT, eventCounter.getAndIncrement());
```
jakeouellette: It is interesting to me that this method is side-effecting and returns a value. Would it be more obvious to the reader if this method were static, and you passed  `generateEventFilename(eventCounter.getAndIncrement(), isHighPriority)` in?

jakeouellette: (I guess you'd probably want to rename it if you did that. I could go either way really, just wanted to ask!)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -151,22 +178,55 @@ public void finalizeReports(String currentSessionId) {
...
+    Collections.sort(normalPriorityEventFiles, CrashlyticsReportPersistence::oldestEventFileFirst);
+    return capFilesCount(normalPriorityEventFiles, maximum);
+  }
+
+  private static String eventNameOnly(String eventFileName) {
+    return eventFileName.substring(0, EVENT_NAME_LENGTH);
+  }
+
+  private static boolean isNormalPriorityEventFile(File dir, String name) {
+    return name.startsWith(EVENT_FILE_NAME_PREFIX) && !isHighPriorityEvent(name);
```
jakeouellette: Possibly a dumb optimization, but if !isHighPriorityEvent is more variable, and EVENT_FILE_NAME_PREFIX is expected 99% of the time, it might be worth putting the second check first to short circuit early more frequently.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -151,22 +178,55 @@ public void finalizeReports(String currentSessionId) {
...
+  }
+
+  private static int oldestEventFileFirst(File f1, File f2) {
```
jakeouellette: Is it worth having a unit test for this?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -222,6 +282,25 @@ private static String readTextFile(File file) {
...
   }
 
+  /**
+   * Deletes files from the list until the list size is equal to the maximum. If list is already
+   * correctly sized, no files are deleted. List should be sorted in the order in which files should
+   * be deleted.
+   *
+   * @return the number of files retained on disk
+   */
+  private static int capFilesCount(List<File> files, int maximum) {
```
jakeouellette: Is it worth having a unit test on this method?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -151,22 +178,55 @@ public void finalizeReports(String currentSessionId) {
...
+        String.format(Locale.US, EVENT_COUNTER_FORMAT, eventCounter.getAndIncrement());
+    final String prioritySuffix = isHighPriority ? PRIORITY_EVENT_SUFFIX : NORMAL_EVENT_SUFFIX;
+    return EVENT_FILE_NAME_PREFIX + eventNumber + prioritySuffix;
   }
 
   private File getSessionDirectoryById(String sessionId) {
     return new File(openSessionsDirectory, sessionId);
   }
 
+  private static int trimEvents(File sessionDirectory, int maximum) {
```
jakeouellette: does persistEvents instrument this trim in any unit tests (possibly they weren't changed by this PR)

jakeouellette: why does this return an int if we never use it

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
jakeouellette: Why do we return an int if we never use it?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -151,22 +178,55 @@ public void finalizeReports(String currentSessionId) {
...
+    final List<File> normalPriorityEventFiles =
+        getFilesInDirectory(
+            sessionDirectory, CrashlyticsReportPersistence::isNormalPriorityEventFile);
+    Collections.sort(normalPriorityEventFiles, CrashlyticsReportPersistence::oldestEventFileFirst);
```
jakeouellette: getFilesInDirectory makes an array internally, did we consider just working with the array class? (e.g., by doing Arrays.sort)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: In practice, yes. If you've got a situation where you have greater than `Integer.MAX_VALUE` events coming in for a single session, something is terribly wrong, and the information isn't going to be particularly useful to you anyway.

See also: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java#L241

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -43,36 +43,41 @@
...
+  private final File openSessionsDirectory;
 
   // Storage for finalized reports
-  // Keep finalized reports organized by whether or not they contain a fatal event.
-  private File fatalReportsDirectory;
-  private File nonFatalReportsDirectory;
+  private final File priorityReportsDirectory;
+  private final File reportsDirectory;
+
+  // TODO: Add settings override
```
jakeouellette: do we use // TODO(name): style?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: Solid point. I don't think it would neccessitate a rename.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: :+1:

jakeouellette:
Do we ever keep more than one high priority event in a session, anyway?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
jakeouellette: Why do we return an int if we never use it?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: It does: https://github.com/firebase/firebase-android-sdk/pull/1255/files#diff-611035c62a50caf0141b0f8ff764617bR158

mrwillis21: Future planned usage.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: Not in OSS

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -151,22 +178,55 @@ public void finalizeReports(String currentSessionId) {
...
+  private static String eventNameOnly(String eventFileName) {
+    return eventFileName.substring(0, EVENT_NAME_LENGTH);
+  }
+
+  private static boolean isNormalPriorityEventFile(File dir, String name) {
+    return name.startsWith(EVENT_FILE_NAME_PREFIX) && !isHighPriorityEvent(name);
+  }
+
+  private static int oldestEventFileFirst(File f1, File f2) {
+    final String name1 = eventNameOnly(f1.getName());
```
jakeouellette: This was a headscratcher for me at first, if a name is [path][name-part][number-part][priority-part], I wasn't realizing eventNameOnly returned [name-part][number-part].

mrwillis21:
> Do we ever keep more than one high priority event in a session, anyway?

No, but that's not something the persistence layer needs to know about.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: Open to other name ideas. :)

jakeouellette: maybe a comment or a slightly adjusted name (e.g., getEventNameAndEventNumber?)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: This is a method reference used as a comparator, which must return an int.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: I think, in general, working with lists is simpler than working with arrays, and in this case I'm not sure there's much performance benefit either way. Happy to discuss it further if you feel strongly about it, though.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: Actually, after looking back at this - we need to filter out any non-event files *first*, so this needs to stay as it is.

mrwillis21: Rewrote these two methods though, and paired them, so they'd be easier to reason about.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: This is a comparator.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: Potentially, but I'll come back to it in another PR.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: :+1:

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
jakeouellette: thoughts on the unit test question?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -190,21 +191,24 @@ public void finalizeReports(String currentSessionId) {
...
+    return fileName.startsWith(EVENT_FILE_NAME_PREFIX) && fileName.endsWith(PRIORITY_EVENT_SUFFIX);
   }
 
-  private String generateEventFilename(boolean isHighPriority) {
-    final String eventNumber =
-        String.format(Locale.US, EVENT_COUNTER_FORMAT, eventCounter.getAndIncrement());
-    final String prioritySuffix = isHighPriority ? PRIORITY_EVENT_SUFFIX : NORMAL_EVENT_SUFFIX;
-    return EVENT_FILE_NAME_PREFIX + eventNumber + prioritySuffix;
+  private static boolean isNormalPriorityEventFile(File dir, String name) {
+    return name.startsWith(EVENT_FILE_NAME_PREFIX) && !name.endsWith(PRIORITY_EVENT_SUFFIX);
```
jakeouellette: I'm prefer startsWith and endsWith here, though I could also see some folks preferring a regex match

mrwillis21: Pretty certain `startsWith` and `endsWith` are more efficient than a regex would be, which is why I used them.

PR #1257, opened 2020-02-20
Make FIAM correctly fetch on first open

JasonAHeron:


firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/FirebaseInAppMessaging.java
```diff
@@ -161,53 +163,33 @@ public boolean areMessagesSuppressed() {
...
 
   /**
-   * Unregisters a listener to in app message display events.
+   * Unregisters a fiamDisplay to in app message display events.
    *
    * @hide
    */
   @Keep
   @KeepForSdk
   public void clearDisplayListener() {
```
prakhar1989: why do we need this method?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -233,7 +234,7 @@ private boolean shouldIgnoreCache(String event) {
 
               Function<CampaignImpressionList, Maybe<FetchEligibleCampaignsResponse>> serviceFetch =
                   impressions ->
-                      Maybe.fromCallable(() -> apiClient.getFiams(impressions))
+                      taskToMaybe(apiClient.getFiams(impressions))
```
prakhar1989: should this be task to single? http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Single.html

JasonAHeron: next cl

PR #1260, opened 2020-02-20
Remove guava usage from firestore.

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/util/Preconditions.java
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+package com.google.firebase.firestore.util;
+
+import static com.google.firebase.firestore.util.Strings.lenientFormat;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class Preconditions {
```
schmidt-sebastian: Please mention the source of this file somewhere (preferably a GitHub link to the Guava sources)

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Strings.java
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * @param template a string containing zero or more {@code "%s"} placeholder sequences. {@code
+   *     null} is treated as the four-character string {@code "null"}.
+   * @param args the arguments to be substituted into the message template. The first argument
+   *     specified is substituted for the first occurrence of {@code "%s"} in the template, and so
+   *     forth. A {@code null} argument is converted to the four-character string {@code "null"};
+   *     non-null values are converted to strings using {@link Object#toString()}.
+   * @since 25.1
+   */
+  public static String lenientFormat(@Nullable String template, @Nullable Object... args) {
```
schmidt-sebastian: If I read this comment correctly, we should be able to drop this method and replace its usage by String.format, reducing the size of this PR. The risk that String.format crashes seems minuscule given our usage.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -461,11 +460,12 @@ public void registerPendingWritesTask(TaskCompletionSource<Void> userTask) {
       return;
     }
 
-    if (pendingWritesCallbacks.containsKey(largestPendingBatchId)) {
-      pendingWritesCallbacks.get(largestPendingBatchId).add(userTask);
-    } else {
-      pendingWritesCallbacks.put(largestPendingBatchId, Lists.newArrayList(userTask));
+    if (!pendingWritesCallbacks.containsKey(largestPendingBatchId)) {
+      List<TaskCompletionSource<Void>> tasks = new ArrayList();
+      pendingWritesCallbacks.put(largestPendingBatchId, tasks);
```
schmidt-sebastian: Nit: You can just inline the call to `new ArrayList`, which would remove all those generic parameters.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
```diff
@@ -264,8 +265,11 @@ private static int compareTimestamps(Timestamp left, Timestamp right) {
   }
 
   private static int compareReferences(String leftPath, String rightPath) {
-    List<String> leftSegments = Splitter.on('/').splitToList(leftPath);
-    List<String> rightSegments = Splitter.on('/').splitToList(rightPath);
+    checkNotNull(leftPath);
+    checkNotNull(rightPath);
+    List<String> leftSegments = Lists.newArrayList(leftPath.split("/"));
```
schmidt-sebastian: Nit: This is a private method. I don't think we need to do null-checks here. `rightPath` and `leftPath` would only be null if the Proto is invalid, in which case `leftPath.split()` would throw the same NPE that you are checking for manually.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Preconditions.java
```diff
@@ -0,0 +1,89 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.util;
+
+import static com.google.firebase.firestore.util.Strings.lenientFormat;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
```
schmidt-sebastian: Consider copying the class JavaDoc.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Function.java
```diff
@@ -0,0 +1,74 @@
...
+ * <p>To use an existing function (say, named {@code function}) in a context where the <i>other
+ * type</i> of function is expected, use the method reference {@code function::apply}. A future
+ * version of {@code com.google.firebase.firestore.util.Function} will be made to <i>extend</i>
+ * {@code java.util.function.Function}, making conversion code necessary only in one direction. At
+ * that time, this interface will be officially discouraged.
+ *
+ * @author Kevin Bourrillion
+ * @since 2.0
+ */
+public interface Function<F, T> {
```
schmidt-sebastian: Please mention somewhere that this is a copy.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/MoreObjects.java
```diff
@@ -0,0 +1,358 @@
...
+ * java.util.Objects}.
+ *
+ * <p>See the Guava User Guide on <a
+ * href="https://github.com/google/guava/wiki/CommonObjectUtilitiesExplained">writing {@code Object}
+ * methods with {@code MoreObjects}</a>.
+ *
+ * @author Laurence Gonsalves
+ * @since 18.0 (since 2.0 as {@code Objects})
+ */
+public final class MoreObjects {
```
schmidt-sebastian: Please replace the only usage of this class with a custom toString() implementation (as we have done in almost all other places of the code base).

samtstern:
:heart: :heart: :heart:

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
wilhuff: For background: `checkNotNull` is only useful if you're retaining an argument in a method but not actually using it there. The best example is if you're taking a constructor argument and assigning it to an instance variable without using it. The idea is to force the NPE to happen at the point where the caller is apparent.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Strings.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Preconditions.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/MoreObjects.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
```diff
@@ -221,12 +220,17 @@ public int hashCode() {
...
-        .toString();
+    StringBuilder builder = new StringBuilder("FirebaseFirestoreSettings{");
+
+    builder.append(String.format("host=%s, ", host));
+    builder.append(String.format("sslEnabled=%s, ", sslEnabled));
+    builder.append(String.format("persistenceEnabled=%s, ", persistenceEnabled));
+    builder.append(
+        String.format("timestampsInSnapshotsEnabled=%s, ", timestampsInSnapshotsEnabled));
+    builder.append(String.format("cacheSizeBytes=%s", cacheSizeBytes));
+
```
schmidt-sebastian: Nit: This is a somewhat strange combination of StringBuilder and String.format. Do you mind using the pattern used here: https://github.com/firebase/firebase-android-sdk/blob/8409ac490902afb840d5adbe52b86216546921be/firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java#L108

Please also take a look here:

https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

TL/DR: The Java compiler will generate the same code for you.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Function.java
```diff
@@ -17,7 +17,10 @@
 import javax.annotation.Nullable;
 
 /**
- * Determines an output value based on an input value; a pre-Java-8 version of {@link
+ * This is a copy of
```
schmidt-sebastian: Please move this outside the class JavaDoc, as you likely don't want this to be the summary of this API. I would recommend just moving this above in `//` style comment.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Preconditions.java
```diff
@@ -14,11 +14,87 @@
...
 package com.google.firebase.firestore.util;
 
-import static com.google.firebase.firestore.util.Strings.lenientFormat;
-
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+/**
+ * This is an adaptation of
+ * https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java.
```
schmidt-sebastian: Same comment regarding the placement of this.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java
wu-hui: Done. Thanks.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Function.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Preconditions.java
wu-hui: Done.

PR #1261, opened 2020-02-20
Add captured logs to Crashlytics reports

mrwillis21:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -75,6 +80,11 @@ public void onNonFatalEvent(Throwable event, Thread thread) {
     onEvent(event, thread, EVENT_TYPE_LOGGED, false);
   }
 
+  @Override
+  public void onLog(long timestamp, String log) {
```
jakeouellette: This makes sense, but I want to talk out a naming thing that is going on here: You designed this manager as a "listener" as events. `onBeginSession`, `onFatalEvent`, `onEndSession` all make sense to me as lifecycle events because they're outside of the operation of this collaborator, and crashlytics is acting as *listener* of them. `onLog` (and to some degree,`onNonFatalEvent`) are a little different because the user explicitly calls it, right? In this case, we're not listening to a log occuring, we're listening to the app logging the information. I find it a little weird that `onLog` results in *logging*, but 🤷‍♂ I can't really think of a different way to name this.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -99,19 +109,31 @@ private void onEvent(
...
+                event,
+                thread,
+                eventType,
+                timestamp,
+                EVENT_THREAD_IMPORTANCE,
+                MAX_CHAINED_EXCEPTION_DEPTH,
+                includeAllThreads)
+            .toBuilder();
+
+    final String content = logFileManager.getLogString();
```
jakeouellette: out of curiosity, what's the threading model of the method this code is in? is it always on a single thread?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/log/LogFileManager.java
```diff
@@ -142,6 +146,11 @@ public void writeToLog(long timestamp, String msg) {}
       return null;
     }
 
+    @Override
+    public String getLogAsString() {
+      return null;
```
jakeouellette: I'm not sure I understand why this is null

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
```diff
@@ -84,6 +84,13 @@ public static Builder builder() {
   @NonNull
   public abstract Session getSession();
 
+  // TODO
```
jakeouellette: Can we add a comment as to why this is a TODO?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: For now, yes. It's all synchronous. My goal is to push threading as far out to the callers as possible.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: In my mind, the "event" happening here is a "user requests a log to be written" event, which is one event that can happen during the lifecycle of a session. (I'm pretty certain I'm going to rename the interface to `SessionLifecycleEvents`, as it happens).

Similar to a "button press" event, it's user-initiated, but still an event.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/log/LogFileManager.java
mrwillis21: Look at the context - this is a no-op implementation.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
mrwillis21: Because we haven't implemented NDK yet. I'll add that to the TODO.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: That is, it isn't intended to be thread-safe (yet).

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
jakeouellette: It isn't the end user requesting that log though, it's likely some internal collaborator class. If you were implementing the code for onButtonPress, you would probably have code in a controller that called a class with a method that looked like `sendForm(...)`, not one that looked like `onFormSent(...)` The collaborator is requesting the sendForm action being taken, in that example. The expectation is a form gets sent.

I think the lifecycle idea works for session start / session ends, because you're abstracting away the meaning of them, but if we're planning to add lots of methods like onCustomKeys(...) it makes me wonder how much the lifecycle naming works.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/log/LogFileManager.java
jakeouellette: ahh, should it be returning "" then? (Same story with the bytes one -- should it be returning an empty array?) (Should we document the expected outputs in the FileLogStore interface?)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
jakeouellette: 👍

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: Talked about this offline, decided there's a good opportunity for some better interfacing here, but we'll break that out in a separate PR.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/log/LogFileManager.java
mrwillis21: Good opportunities for improvement. I'll add a TODO.

mrwillis21: Also, to answer your question, no - the intent is that `null` means there is no data, separate from a potentially empty log line `""`.

PR #1262, opened 2020-02-20
update versions 

VinayGuthal:


firebase-firestore/ktx/gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=21.4.0
-latestReleasedVersion=21.3.1
+version=21.4.1
```
vkryachko: The version properties are not meaningful for ktx libraries, instead ktx libs use the version of their corresponding Java SDK. pls remove `version` and `latestReleasedVersion` and make sure firebase-firestore has the correct version properties.

vkryachko: looks like you did not remove the properties from this file

VinayGuthal: Done

PR #1263, opened 2020-02-21
Quick Fix: FIS SDK "current time" doesn't get update.

ChaoqunCHEN:
FIS SDK is not aware of the real system time due to the "current time" was set once when the calendar instance got initialized (at app launches in our case). This bug makes FIS SDK not being able to fetch new AuthToken on time which causes 1M+ authentication fails on VerifyToken RPC call.

Note: this is a quick fix. Some affected tests removed should be brought back when @ankitaj224  is back to work.

### DIFF
--- a/firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
+++ b/firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
@@ -127,14 +127,13 @@ public class FirebaseInstallationsTest {
 
   private FirebaseInstallations firebaseInstallations;
   private Utils utils;
-  private FakeCalendar fakeCalendar;
 
   @Before
   public void setUp() {
     MockitoAnnotations.initMocks(this);
     FirebaseApp.clearInstancesForTest();
     executor = new ThreadPoolExecutor(0, 1, 30L, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
-    fakeCalendar = new FakeCalendar(5000000L);
+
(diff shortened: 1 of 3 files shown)


ChaoqunCHEN:
> It would be nice to at least have a test where we create store an auth token in the database with an expiration date that is 30 minutes old and check that FIS auth token fetches a new token, but not critical.

Added test `testGetAuthToken_expiredAuthToken_fetchedNewTokenFromFIS`

firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
```diff
@@ -639,13 +554,13 @@ public void testGetAuthToken_expiredAuthToken_fetchedNewTokenFromFIS() throws Ex
         PersistedInstallationEntry.INSTANCE.withRegisteredFid(
             TEST_FID_1,
             TEST_REFRESH_TOKEN,
-            utils.currentTimeInSecs(),
+            // Make experation time to be 30 min ago
+            utils.currentTimeInSecs()
+                - TEST_TOKEN_EXPIRATION_TIMESTAMP
+                - TimeUnit.MINUTES.toSeconds(30),
```
andirayo: make this + TimeUnit.MINUTES.toSeconds(30) please
FIS SDK should update a FIS auth token even if it is still valid for another hour...
(see grace period)

ChaoqunCHEN: done

PR #1265, opened 2020-02-21
Crashlytics update Update endpoint from api. => update.

samedson:
New DNS to facilitate backend migration


mrwillis21:
Tested this out, and was able to onboard a fresh app and register an updated version. LGTM!

PR #1266, opened 2020-02-21
Add custom keys to Crashlytics reports

mrwillis21:


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
```diff
@@ -156,19 +167,123 @@ public void testOnFatalEvent_addsLogsToEvent() {
...
+    verify(mockEventBuilder).setApp(mockEventApp);
+    verify(mockEventBuilder).build();
+    verify(logFileManager).clearLog();
+  }
+
+  @Test
+  public void testOnFatalEvent_addsNoKeysToEventWhenNoneAvailable() {
+    mockEventInteractions(System.currentTimeMillis());
+
+    final Map<String, String> attributes = new HashMap<>();
```
jakeouellette: How do you feel about `Collections.emptyMap()` (or `ImmutableMap.of()`)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/UserMetadata.java
```diff
@@ -37,6 +38,7 @@ public void setUserId(String identifier) {
     userId = sanitizeAttribute(identifier);
   }
 
+  @NonNull
   public Map<String, String> getCustomKeys() {
     return Collections.unmodifiableMap(attributes);
```
jakeouellette: Not sure it matters, but just a note here that unmodifiable maps can still change, just not by the user of the returned value.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -148,4 +174,23 @@ private boolean onReportSendComplete(Task<CrashlyticsReport> task) {
     // TODO: Something went wrong. Log? Throw?
     return false;
   }
+
+  private static List<CustomAttribute> getSortedCustomAttributes(Map<String, String> attributes) {
+    if (attributes == null || attributes.isEmpty()) {
+      return null;
+    }
+
+    List<CustomAttribute> attributesList = new ArrayList<>();
```
jakeouellette: Since this is a copy, you might consider setting a capacity, e.g., `new ArrayList<>(attributes.size())`?

jakeouellette:
My largest concern is whether or not the `UserMetadata`'s `attributes` field can be "set" while another class uses its reference. You might consider moving the sorted-copy inside the `UserMetadata` class (e.g., add a getSortedAttributes(Comparator...) and protecting the class from modification while it is copied.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: :+1:

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/UserMetadata.java
mrwillis21: :+1: Discussed this offline - the `ConcurrentHashMap` implementation guarantees this is safe, but we should definitely revisit it later.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
mrwillis21: :+1:

PR #1270, opened 2020-02-24
Add user ID to Crashlytics reports

mrwillis21:


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
```diff
@@ -288,6 +288,21 @@ public void testOnFatalEvent_addsNoKeysToEventWhenNoneAvailable() {
     verify(logFileManager).clearLog();
   }
 
+  @Test
+  public void testOnEndSession_addsUserIdToReport() {
+    final String userId = "testUser";
+    final String sessionId = "testSessionId";
+    final long timestamp = System.currentTimeMillis();
```
jakeouellette: This is not problematic, per-se, but consider replacing with a constant time value to make the test deterministic

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -98,8 +98,14 @@ public void onCustomKey(String key, String value) {
...
   }
 
+  @Override
+  public void onUserId(String userId) {
+    reportMetadata.setUserId(userId);
+  }
+
   @Override
   public void onEndSession() {
+    reportPersistence.persistUserIdForSession(reportMetadata.getUserId(), currentSessionId);
```
jakeouellette: These two lines are fine! -- but I think this is where there's an opportunity to instead of calling reportPersistence and passing in a currentSessionId, making a per session-id object that gets persisted by these calls.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManagerTest.java
mrwillis21: The value of this timestamp isn't important for this test, we just need one.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
mrwillis21: Yup, let's chat more about this.

PR #1273, opened 2020-02-24
Add new Crashlytics report upload variant setting

mrwillis21:


mrwillis21:
> Would it be possible to make constant ints to represent the semantic behavior of these different integers?

I think that would be appropriate for the consumers of the setting (and I do intend to do so there), but this code only cares about taking the value it's handed and creating an object out of it. `0` here is simply the default value of an int, with no semantic meaning.

PR #1280, opened 2020-02-25
Add upload variant switch for DataTransport

mrwillis21:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FirebaseCrashlyticsReportManager.java
```diff
@@ -116,6 +119,11 @@ public void onFinalizeSessions() {
 
   @Override
   public void onSendReports(AppSettingsData appSettingsData) {
+    if (appSettingsData.reportUploadVariant != REPORT_UPLOAD_VARIANT_DATATRANSPORT) {
```
jakeouellette: what's the default on this value? Any chance we could lose reports due to deleting it with this not set during startup or when folks first upgrade their SDK, or something?

mrwillis21: Good call; the default value is 0, and we should have that go through reports endpoint -- a fallback which isn't currently implemented.

mrwillis21: I'll take that as a follow-up PR though.

PR #1281, opened 2020-02-26
Silently fail if disk is full or cannot be read for other reasons

andirayo:
This has been done to temporarily quick-fix apps from crash-looping on devices with full disk.
Presumably the SDK will throw error (instead of an exception) later when it cannot persist the FID and data.
b/149571942

### DIFF
--- a/firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
+++ b/firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
@@ -39,6 +39,10 @@ class CrossProcessLock {
       // This method blocks until it can retrieve the lock.
       FileLock lock = channel.lock();
       return new CrossProcessLock(channel, lock);
+    } catch (FileNotFoundException e) {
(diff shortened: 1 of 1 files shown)


fredquintana:
You need to be sure that the release path doesn't fail. E.g. check that lock and file are not null.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -39,6 +39,10 @@ static CrossProcessLock acquire(Context appContext, String lockName) {
       // This method blocks until it can retrieve the lock.
       FileLock lock = channel.lock();
       return new CrossProcessLock(channel, lock);
+    } catch (FileNotFoundException e) {
+      // Firebase Installations will silently fail if the disk is full or can not be accessed for other reasons.
+      // This has been done to temporarily quick-fix apps from crash-looping on devices with full disk.
+      // Presumably the SDK will throw error (instead of an exception) later when it cannot persist the FID and data.
```
fredquintana: you need to return a new CrossProcessLock here

PR #1282, opened 2020-02-26
Make CrossProcessLock resilient to disk issues

fredquintana:
If the CrossProcessLock fails due to disk issues it will fail instead of throwing an exception.

As this should only happen due to issues like disk full or messed up permissions the FirebaseInstallations code will log and ignore any failure to get the log and move forward.

### DIFF
--- a/firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
+++ b/firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
@@ -15,6 +15,7 @@
 package com.google.firebase.installations;
 
 import android.content.Context;
+import android.util.Log;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
@@ -23,6 +24,7 @@ import java.nio.channels.FileLock;
 
 /** Use file locking to acquire a lock that will also block other processes. */
 class CrossProcessLock {
+  private final static String TAG = "CrossProcessLock";
   private final FileChannel channel;
   private final FileLock lock;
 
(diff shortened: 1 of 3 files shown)


firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -31,16 +33,46 @@ private CrossProcessLock(FileChannel channel, FileLock lock) {
...
-      FileChannel channel = new RandomAccessFile(file, "rw").getChannel();
+      channel = new RandomAccessFile(file, "rw").getChannel();
       // Use the file channel to create a lock on the file.
       // This method blocks until it can retrieve the lock.
-      FileLock lock = channel.lock();
-      return new CrossProcessLock(channel, lock);
+      lock = channel.lock();
+      CrossProcessLock crossProcessLock = new CrossProcessLock(channel, lock);
+      channel = null;
+      lock = null;
```
andirayo: Do we need this?
Can't we just do this:

      return crossProcessLock;
    } catch (IOException e) {
      Log.e(TAG, "encountered error while creating and acquiring the lock, ignoring", e);
      if (channel == null) {
        return null;
      }
      try {
        channel.close();
      } catch (IOException e) {
        // nothing to do here
      }
      if (lock != null) {
        try {
          lock.release();
        } catch (IOException e) {
          // nothing to do here
        }
      }
      return null;
    }

fredquintana: That would also work, do you prefer it?

PR #1288, opened 2020-02-28
Use Locale#toLanguageTag to properly encode IETF BCP 47 language tags.

danasilver:
[`Locale#toLanguageTag`](https://developer.android.com/reference/java/util/Locale#toLanguageTag()) is preferred to [`Locale#toString`](https://developer.android.com/reference/java/util/Locale#toString()) for interchange purposes.

The Firebase backends expect a BCP-47 encoded language tag, which is produced by `Locale#toLanguageTag`, and not in all cases by `Locale#toString`.

### DIFF
--- a/firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java
+++ b/firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java
@@ -199,13 +199,27 @@ public class ConfigFetchHttpClientTest {
     assertThat(requestBody.get(APP_ID)).isEqualTo(FAKE_APP_ID);
     Locale locale = context.getResources().getConfiguration().locale;
     assertThat(requestBody.get(COUNTRY_CODE)).isEqualTo(locale.getCountry());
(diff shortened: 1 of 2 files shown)


firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java
```diff
@@ -208,6 +208,20 @@ public void fetch_setsAllElementsOfRequestBody_sendsRequestBodyToServer() throws
         .isEqualTo(new JSONObject(userProperties).toString());
   }
 
+  @Test
+  public void fetch_requestEncodesLanguageSubtags() throws Exception {
```
clp93: Can we add a similar test for a device that's below API Level 21? Thanks!

danasilver: Good call!

danasilver:
Tested end to end locally and got the expected result for a RC condition with the `zh-Hant-TW` language tag.

PR #1289, opened 2020-02-28
Crashlytics refactor report manager

mrwillis21:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
```diff
@@ -34,7 +34,7 @@
  * This class handles Crashlytics lifecycle events and manages capture, persistence, and sending of
```
jakeouellette: Should we use any language of "SessionReportingCoordinator"? e.g., "This Class coordinates Crashlytics lifecycle events and manages the capture, persistence, and sending of sessions as reports to Firebase Crashlytics"?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
```diff
@@ -34,7 +34,7 @@
...
  * reports to Firebase Crashlytics.
  */
-public class FirebaseCrashlyticsReportManager implements CrashlyticsLifecycleEvents {
+public class SessionReportingCoordinator implements CrashlyticsLifecycleEvents {
```
jakeouellette: Should we rename CrashlyticsLifecycleEvents to something less specific than SessionReportingCoordinator, e.g.,  SessionCoordinator, or maybe call the interface SessionReportingCoordinator and rename this class to something more specific about how it does it?

mrwillis21: In the long term, that's the plan, but this is one step in that direction. We'll come back to this.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
mrwillis21: Good point.

mrwillis21: Done.

PR #1290, opened 2020-02-28
Move variant handling out of SessionReportingCoordinator

mrwillis21:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
```diff
@@ -36,19 +36,21 @@
  */
 public class SessionReportingCoordinator implements CrashlyticsLifecycleEvents {
 
+  public interface SendCheck {
+    boolean shouldSendViaDataTransport(int variant);
```
jakeouellette: Observation: This makes visible that the logic is determined by the variant passed into the interface. I think that it isn't necessary to make that visible because the SessionReportingCooridnator shouldn't be responsible for passing settings around.

There's an opportunity to hide that extra complexity. One way you could do it is by renaming this as `SendStrategy`, and make its interface not take any parameters. Then when constructing this, pass in the value. (If the settings were stateful and ever needed to be updated, you would pass in a new strategy alongside the settings.)

mrwillis21: Done.

PR #1300, opened 2020-03-02
Add a limit to the number of currently open sessions

jakeouellette:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -211,8 +208,9 @@ public void finalizeReports(String currentSessionId) {
             TRANSFORM.reportToJson(report.withEvents(ImmutableList.from(events))));
       }
       recursiveDelete(sessionDirectory);
-      capFinalizedReports();
     }
+
+    capFinalizedReports();
```
jakeouellette: this was just an unnecessary call to cap all the reports after each report.

PR #1301, opened 2020-03-02
Catching NPE and throwing IOE instead.

andirayo:
Youtube observed this error on some devices when rolling out a new version. b/150526702
We were not able to reproduce this Exception, but found implementations of (Http)URLConnection that might return NULL for #getOutputStream.
The new code is preventing NPE and following the default error scenario that a connection to the FIS server API can not be established.

### DIFF
--- a/firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
+++ b/firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
@@ -179,7 +179,7 @@ public class FirebaseInstallationServiceClient {
       throws IOException {
     OutputStream outputStream = httpURLConnection.getOutputStream();
     if (outputStream == null) {
-      throw new ObjectStreamException("Cannot send request to FIS. No OutputStream available.");
+      throw new ObjectStreamException("Cannot send CreateInstallation request to FIS. No OutputStream available.");
     }
     
     GZIPOutputStream gzipOutputStream = new GZIPOutputStream(outputStream);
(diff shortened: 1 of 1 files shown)


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -175,7 +177,12 @@ public InstallationResponse createFirebaseInstallation(
   private void writeFIDCreateRequestBodyToOutputStream(
       HttpURLConnection httpURLConnection, @NonNull String fid, @NonNull String appId)
       throws IOException {
-    GZIPOutputStream gzipOutputStream = new GZIPOutputStream(httpURLConnection.getOutputStream());
+    OutputStream outputStream = httpURLConnection.getOutputStream();
+    if (outputStream == null) {
+      throw new ObjectStreamException("Cannot send request to FIS. No OutputStream available.");
```
fredquintana: please change to:

 Cannot send create request

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -198,7 +205,12 @@ private static JSONObject buildCreateFirebaseInstallationRequestBody(String fid,
 
   private void writeGenerateAuthTokenRequestBodyToOutputStream(HttpURLConnection httpURLConnection)
       throws IOException {
-    GZIPOutputStream gzipOutputStream = new GZIPOutputStream(httpURLConnection.getOutputStream());
+    OutputStream outputStream = httpURLConnection.getOutputStream();
+    if (outputStream == null) {
+      throw new ObjectStreamException("Cannot send request to FIS. No OutputStream available.");
```
fredquintana: please change to:

 Cannot send authtoken request

PR #1304, opened 2020-03-02
Add support for other Firebase products to integrate with Remote Config.

ashwinraghav:


firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -773,6 +791,41 @@ void updateAbtWithActivatedExperiments(@NonNull JSONArray abtExperiments) {
...
+      executor.execute(() -> firebasePersonalization.pullArm(key));
+    }
+  }
+
+  /**
+   * Notifies the Firebase Personalization SDK about activated metadata.
+   *
+   * @hide
+   */
+  void updatePersonalizationWithActivatedMetadata(@NonNull ConfigContainer configs) {
```
ashwinraghav: It is not clear why the personalization SDK needs to know anything at this stage. That seems like an optimization

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -358,6 +368,7 @@ public boolean activateFetched() {
    */
   @NonNull
   public String getString(@NonNull String key) {
+    pullArm(key);
```
ashwinraghav: We are essentially doing a write-on-read. Developers are used to treating the get*() APIs as idempotent functions. This will no longer be true. Are you ok if GA counts these as independent events.

ashwinraghav: It seems like we are going to build a variety of products built on top of config that would require this feedback loop to be established. I feel like a filter pattern or a listener pattern is a good candidate to consider here, at the risk of optimizing too early.

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -187,6 +195,7 @@ synchronized FirebaseRemoteConfig get(
               firebaseApp,
               firebaseInstanceId,
               isAbtSupported(firebaseApp, namespace) ? firebaseAbt : null,
+              isAbtSupported(firebaseApp, namespace) ? firebasePersonalization : null,
```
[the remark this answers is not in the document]
vic-flair: Moved it out to getPersonalization().

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigRegistrar.java
```diff
@@ -45,6 +46,7 @@
             .add(Dependency.required(FirebaseApp.class))
             .add(Dependency.required(FirebaseInstanceId.class))
             .add(Dependency.required(AbtComponent.class))
+            .add(Dependency.required(PersonalizationComponent.class))
```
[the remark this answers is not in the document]
vic-flair: Removed the separate PersonalizationComponent and just put everything into RC itself.

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
vic-flair: Yeah, I guess it is. Removed.

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
vic-flair: Yeah, we're OK with GA counting these as separate events. Implemented the listener pattern.

firebase-config/src/main/java/com/google/firebase/remoteconfig/BiConsumer.java
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.remoteconfig;
+
+import androidx.annotation.Nullable;
+
+/** A port of {@link java.util.function.BiConsumer}, a generic consumer of values. */
+@FunctionalInterface
+public interface BiConsumer<T, U> {
```
danasilver, line 21: Can we move this interface to the internal package?

vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -0,0 +1,63 @@
...
+   * Called when an arm is pulled, and uses Google Analytics for Firebase to log it if it's a
+   * Personalization parameter.
+   *
+   * @param key Remote Config parameter
+   * @param configContainer JSON of {@link ConfigContainer}
+   */
+  public void logArmActive(@NonNull String key, @NonNull JSONObject configContainer) {
+    JSONObject ids = configContainer.optJSONObject(PERSONALIZATION_METADATA_KEY);
+    JSONObject values = configContainer.optJSONObject(CONFIGS_KEY);
+    if (ids == null || values == null) {
```
erikeldridge, line 49: As above, if this is an error case, I'd recommend logging something to help customers debug unexpected behavior.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -400,6 +401,16 @@ private static ConfigContainer extractConfigs(JSONObject fetchResponse, Date fet
         containerBuilder.withAbtExperiments(experimentDescriptions);
       }
 
+      JSONObject personalizationMetadata = null;
+      try {
+        personalizationMetadata = fetchResponse.getJSONObject(PERSONALIZATION_METADATA);
+      } catch (JSONException e) {
+        // Do nothing if personalizationMetadata does not exist.
```
erikeldridge, line 408: I'd recommend logging something anywhere we suppress an error. Silent failures limit the details customers can provide in support requests.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -325,6 +337,33 @@ public FirebaseRemoteConfigValue getValue(String key) {
...
+   *
+   * @param key parameter key that was retrieved
+   * @param container the container the key was retrieved from
+   */
+  private void callListeners(String key, ConfigContainer container) {
+    JSONObject configs = container.getContainer();
+
+    synchronized (listeners) {
+      for (BiConsumer<String, JSONObject> listener : listeners) {
+        executor.execute(() -> listener.accept(key, configs));
```
erikeldridge, line 362: Thinking: I like this enforcement of the listener being called on a separate thread, since this is running in the low-latency `get` layer

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -92,14 +102,14 @@ public ConfigGetParameterHandler(
    * @param key A Firebase Remote Config parameter key.
    */
   public String getString(String key) {
-    String activatedString = getStringFromCache(activatedConfigsCache, key);
-    if (activatedString != null) {
-      return activatedString;
+    ConfigAndValue configAndValue = getStringFromCache(activatedConfigsCache, key);
```
erikeldridge, line 105: Given the `config` is available as `activatedConfigsCache` it's unclear to me why we need the `ConfigAndValue` tuple, eg as opposed to:

```
public String getString(String key) {
  String activatedString = getStringFromCache(activatedConfigsCache, key);
  if (activatedString != null) {
    ConfigContainer cachedContainer = getConfigsFromCache(activatedConfigsCache);
    callListeners( ... cachedContainer);
    return activatedString;
  }
  ...
```

erikeldridge, line 105: Thinking: customers can also get params via `getAll` ( https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfig#getAll() ), which won't log. It feels relatively straightforward to say that's an unsupported advanced use-case, but it is an inconsistency.

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -153,6 +154,14 @@ public synchronized FirebaseRemoteConfig get(String namespace) {
     ConfigCacheClient activatedCacheClient = getCacheClient(namespace, ACTIVATE_FILE_NAME);
     ConfigCacheClient defaultsCacheClient = getCacheClient(namespace, DEFAULTS_FILE_NAME);
     ConfigMetadataClient metadataClient = getMetadataClient(context, appId, namespace);
+
+    ConfigGetParameterHandler getHandler = getGetHandler(activatedCacheClient, defaultsCacheClient);
+    Personalization personalization =
+        getPersonalization(firebaseApp, namespace, analyticsConnector);
+    if (personalization != null) {
+      getHandler.addListener(personalization::logArmActive);
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/BiConsumer.java
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.remoteconfig.internal;
+
+import androidx.annotation.Nullable;
+
+/** A port of {@link java.util.function.BiConsumer}, a generic consumer of values. */
```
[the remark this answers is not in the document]
vic-flair: Turns out they also have this in com.google.android.gms, so used that one instead.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
vic-flair: That works too. I figured we're already calling getConfigsFromCache() in getStringFromCache(), so it was easier to just return it, but I guess it is ugly.

vic-flair: getAll() relies on getValue(), which does log, so this case is supported.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -92,14 +102,14 @@ public ConfigGetParameterHandler(
...
+    if (configAndValue == null) {
+      configAndValue = getStringFromCache(defaultConfigsCache, key);
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
erikeldridge: Thx

Thinking: `getJSONObject` throws if the key doesn't exist, which will be true for anyone not using P13n. We could check for the key first w `has`, but since this code only runs during fetch, I don't think it'll be spammy enough to justify that.

danasilver: Some other thoughts: we follow this `getJSONObject` pattern for other field above but don't log if there's nothing to get. RC had bug reports of spammy logging when we were logging warnings in the expected case for performance monitoring, so using at least the `INFO` level seems appropriate. I might even use `DEBUG` here to make it easier to hide this log message for users not using p13n.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
```diff
@@ -80,7 +89,12 @@ static ConfigContainer copyOf(JSONObject containerJson) throws JSONException {
     return new ConfigContainer(
         containerJson.getJSONObject(CONFIGS_KEY),
         new Date(containerJson.getLong(FETCH_TIME_KEY)),
-        containerJson.getJSONArray(ABT_EXPERIMENTS_KEY));
+        containerJson.getJSONArray(ABT_EXPERIMENTS_KEY),
+        containerJson.getJSONObject(PERSONALIZATION_METADATA_KEY));
+  }
+
+  JSONObject getContainer() {
```
danasilver, line 96: Is this used anymore?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Gets the Personalization metadata associated with {@code key}.
+   *
+   * @param key Remote Config parameter
+   * @param configContainer cache of {@link ConfigContainer}
+   */
+  @Nullable
+  public static JSONObject getMetadata(
```
danasilver, line 63: What do you think of calling this `getPersonalizationMetadata` since it's used statically without the `Personalization` class as context?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -325,6 +347,31 @@ public FirebaseRemoteConfigValue getValue(String key) {
...
+  public void addListener(BiConsumer<String, JSONObject> listener) {
+    synchronized (listeners) {
+      listeners.add(listener);
+    }
+  }
+
+  /**
+   * Calls all listeners in {@link #listeners}.
+   *
+   * @param value parameter value that was retrieved
```
danasilver, line 364: Wonder if there's a more precise word than retrieved here? I think it refers to the value being retrieved by the developer via a `get()` call, right?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
vic-flair: Nope.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: Prefixed them with `Personalization.` so there's context.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
vic-flair: Reworded.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Called when an arm is pulled, and uses Google Analytics for Firebase to log it if it's a
+   * Personalization parameter.
+   *
+   * @param key Remote Config parameter
+   * @param configContainer JSON of {@link ConfigContainer}
+   */
+  public void logArmActive(@NonNull String key, @NonNull JSONObject configContainer) {
```
vkryachko, line 49: It's not blocking but might be worth thinking about long-term: why is json passed here? it does not look type-safe, given that there are multiple checks for certain keys that are expected to be in this json object. Would it make sense to create a more type-safe interface to the config?

vkryachko (commented):
Overall lgtm, left one question below

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: Good point. Switched it to use ConfigContainer instead.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -325,6 +344,35 @@ public FirebaseRemoteConfigValue getValue(String key) {
...
+    synchronized (listeners) {
+      listeners.add(listener);
+    }
+  }
+
+  /**
+   * Calls all listeners in {@link #listeners}.
+   *
+   * @param key Remote Config parameter key
+   * @param container cache of {@link ConfigContainer}
```
danasilver, line 362: The `ConfigContainer` is a local cache of the config, but I'm not sure it makes sense to call it a cache in this context. Maybe `{@link ConfigContainer} containing {@code key}`

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -0,0 +1,70 @@
...
+  public Personalization(@NonNull AnalyticsConnector analyticsConnector) {
+    this.analyticsConnector = analyticsConnector;
+  }
+
+  /**
+   * Called when an arm is pulled, and uses Google Analytics for Firebase to log it if it's a
+   * Personalization parameter.
+   *
+   * @param key Remote Config parameter
+   * @param configContainer cache of {@link ConfigContainer}
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -325,6 +344,35 @@ public FirebaseRemoteConfigValue getValue(String key) {
...
+   * @param listener function that takes in the parameter key and the {@link ConfigContainer} JSON
+   */
+  public void addListener(BiConsumer<String, ConfigContainer> listener) {
+    synchronized (listeners) {
+      listeners.add(listener);
+    }
+  }
+
+  /**
+   * Calls all listeners in {@link #listeners}.
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -325,6 +344,35 @@ public FirebaseRemoteConfigValue getValue(String key) {
     return allConfigs;
   }
 
+  /**
+   * Adds a listener that will be called whenever one of the get methods is called.
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java
```diff
@@ -325,6 +344,35 @@ public FirebaseRemoteConfigValue getValue(String key) {
...
+   *
+   * @param listener function that takes in the parameter key and the {@link ConfigContainer} JSON
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -0,0 +1,70 @@
...
+  /** The app's Firebase Analytics client. */
+  private final AnalyticsConnector analyticsConnector;
+
+  /** Creates an instance of {@code Personalization}. */
+  public Personalization(@NonNull AnalyticsConnector analyticsConnector) {
+    this.analyticsConnector = analyticsConnector;
+  }
+
+  /**
+   * Called when an arm is pulled, and uses Google Analytics for Firebase to log it if it's a
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandlerTest.java
```diff
@@ -83,7 +84,9 @@
   public void setUp() {
     MockitoAnnotations.initMocks(this);
 
-    getHandler = new ConfigGetParameterHandler(mockActivatedCache, mockDefaultsCache);
+    getHandler =
+        new ConfigGetParameterHandler(
+            MoreExecutors.directExecutor(), mockActivatedCache, mockDefaultsCache);
```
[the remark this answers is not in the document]
vic-flair: Done.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -0,0 +1,213 @@
...
+    assertThat(FAKE_LOGS.get(0).toString()).isEqualTo(params1.toString());
+
+    Bundle params2 = new Bundle();
+    params2.putString(ARM_KEY, "id2");
+    params2.putString(ARM_VALUE, "value2");
+    assertThat(FAKE_LOGS.get(1).toString()).isEqualTo(params2.toString());
+  }
+
+  @Test
+  public void logArmActive_RemoteConfigIntegration_successful() {
```
danasilver, line 153: Maybe this should go with the FirebaseRemoteConfig tests since it's testing the functionality of the `get` methods? Then the class wouldn't need `@VisibleForTesting`.

vic-flair: Done.

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -189,6 +205,22 @@ public void setUp() throws Exception {
                 mockFireperfGetHandler,
                 RemoteConfigComponent.getMetadataClient(context, APP_ID, FIREPERF_NAMESPACE));
 
+    personalizationFrc =
+        FirebaseApp.getInstance()
+            .get(RemoteConfigComponent.class)
+            .get(
+                firebaseApp,
+                RemoteConfigComponent.DEFAULT_NAMESPACE,
```
danasilver, line 213: Do we need a different instance of RC for personalization? Using the default namespace will get the same RC instance as `frc` above.

vic-flair: We need it to use a real `ConfigGetParameterHandler` so we can test that it's doing things correctly.

danasilver (approved):

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
erikeldridge: In general, if there's an error and it's actionable, I think we'd be wise to log a warning. In this case, it sounds like the common case will be to throw, so it's borderline not an error, and only actionable for folks using P13n, so it should not be `WARN` level. I guess in this case, we can't differentiate between an error case of missing metadata and the common case of not using the feature. `DEBUG` probably makes sense since it's a suppressed error, which folks would be looking for when debugging.

PR #1308, opened 2020-03-03
check for all allowable iidtoken scopes, not just *

fredquintana:
The code was hardcoded to only check for the wildcard scope when looking for the iidtoken when migrating an iid to a fid.

Since scopes FCM and GCM are also allowed, check for those, too.


firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -44,9 +44,9 @@
   private static final String STORE_KEY_PUB = "|S||P|";
   private static final String STORE_KEY_ID = "|S|id";
   private static final String STORE_KEY_TOKEN = "|T|";
-  private static final String DEFAULT_SCOPE = "|*";
   private static final String JSON_TOKEN_KEY = "token";
   private static final String JSON_ENCODED_PREFIX = "{";
+  private static final String [] ALLOWABLE_SCOPES = new String[]{"*", "FCM", "GCM"};
```
andirayo: Is the space on purpose between String and []?

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -86,14 +86,23 @@ private static String getDefaultSenderId(FirebaseApp app) {
     return projectNumber;
   }
 
-  private String createTokenKey(@NonNull String senderId) {
-    return STORE_KEY_TOKEN + senderId + DEFAULT_SCOPE;
+  private String createTokenKey(@NonNull String senderId, @NonNull String scope) {
+    return STORE_KEY_TOKEN + senderId + " | " + scope;
```
andirayo: constant for the pipe character "|"

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -86,14 +86,23 @@ private static String getDefaultSenderId(FirebaseApp app) {
...
   }
 
   @Nullable
   public String readToken() {
     synchronized (iidPrefs) {
-      String token = iidPrefs.getString(createTokenKey(defaultSenderId), null);
+      Log.d("Rayo", "readToken: defaultSenderId: " + defaultSenderId);
```
andirayo: I know you would have removed it, but still annoying you with a comment, haha

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -86,14 +86,23 @@ private static String getDefaultSenderId(FirebaseApp app) {
...
+      String token = null;
+
+      for (String scope : ALLOWABLE_SCOPES) {
+        token = iidPrefs.getString(createTokenKey(defaultSenderId, scope), null);
+        if (!TextUtils.isEmpty(token)) {
+          break;
```
andirayo: How about we replace `break;` by this line:
```
return (token.startsWith(JSON_ENCODED_PREFIX)) ? parseIidTokenFromJson(token) : token;
```


And below instead of
```
if (TextUtils.isEmpty(token)) {
  return null;
}
```
We just do `return null;`

fredquintana: done

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
fredquintana: done

fredquintana:
I also added unit tests for the different supported scopes as well as the empty scope and the json formatted tokens.

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -44,9 +45,9 @@
   private static final String STORE_KEY_PUB = "|S||P|";
   private static final String STORE_KEY_ID = "|S|id";
   private static final String STORE_KEY_TOKEN = "|T|";
-  private static final String DEFAULT_SCOPE = "|*";
   private static final String JSON_TOKEN_KEY = "token";
   private static final String JSON_ENCODED_PREFIX = "{";
+  private static final String [] ALLOWABLE_SCOPES = new String[]{"*", "FCM", "GCM", ""};
```
andirayo: Is the space on purpose between String and []?

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -86,23 +93,22 @@ private static String getDefaultSenderId(FirebaseApp app) {
     return projectNumber;
   }
 
-  private String createTokenKey(@NonNull String senderId) {
-    return STORE_KEY_TOKEN + senderId + DEFAULT_SCOPE;
+  private String createTokenKey(@NonNull String senderId, @NonNull String scope) {
+    return STORE_KEY_TOKEN + senderId + "|" + scope;
```
andirayo: Maybe a constant for the pipe "|"?

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
fredquintana: done

firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
fredquintana: no, done

PR #1314, opened 2020-03-04
Minor refactor of Crashlytics realtime analytics event processing.

mrichards:
Remove the one-off ExecutorService and CountdownLatch that were
responsible for blocking to wait for the receipt of the app
exception event. They are replaced with a TaskCompletionSource
that improves readability and has less runtime overhead.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
@@ -84,6 +84,7 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
```diff
@@ -1809,83 +1805,47 @@ private void sendSessionReports(AppSettingsData appSettings, boolean dataCollect
...
+    final TaskCompletionSource<Void> source = new TaskCompletionSource<>();
+    // This listener completes the TaskCompletionSource when it receives any app exception event
+    // that originated from Crashlytics. The implementation assumes there can be at most ONE app
+    // exception event being processed and waited upon at a time.
+    analyticsReceiver.setCrashlyticsOriginEventListener(
+        (id, extras) -> {
+          String eventName = extras.getString(AnalyticsConnectorReceiver.EVENT_NAME_KEY);
+          if (AnalyticsConnectorReceiver.APP_EXCEPTION_EVENT_NAME.equals(eventName)) {
+            Logger.getLogger().d(Logger.TAG, "Received app exception event callback from FA");
+            source.setResult(null);
```
mrichards: One concern I have here: This blocks indefinitely, so if for some reason we never get the analytics callback, the app will hang rather than terminate. Ideas?

mrwillis21: This should ultimately be handled by the timeout for the uncaught exception handler chain, no?

mrwillis21: Also, you should use `trySetResult` to avoid the possibility of throwing an exception here.

mrichards: The callback issue turned out to be due to a pre-existing issue in the awaitEvenIfOnMainThread impl, which is fixed in the newest commit on this PR.

PR #1319, opened 2020-03-06
Make binary size report format compatible with the metrics service.

yifanyang:


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.groovy
```diff
@@ -0,0 +1,72 @@
...
+            return
+        }
+
+        def owner = null
+        def repo = null
+        def baseCommit = null
+        def headCommit = null
+        def pullRequest = null
+
+        if (System.getenv().containsKey("PROW_JOB_ID")) {
```
rlazo: Reverse the logic to reduce nesting
```
if (!system.getenv...) {
project.logger.quiet
return ?
} 
...
```
the continue with the rest of the code

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.groovy
```diff
@@ -0,0 +1,72 @@
...
+            owner = System.getenv("REPO_OWNER")
+            repo = System.getenv("REPO_NAME")
+            baseCommit = System.getenv("PULL_BASE_SHA")
+            headCommit = System.getenv("PULL_PULL_SHA") ?: baseCommit
+            pullRequest = System.getenv("PULL_NUMBER")
+        } else {
+            project.logger.quiet "CI other than Prow is not supported currently."
+        }
+
+        post(project, report, owner, repo, headCommit, baseCommit, pullRequest)
```
rlazo: this post happens even if the CI is not prow

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.groovy
```diff
@@ -0,0 +1,72 @@
...
+        if (baseCommit && pullRequest) {
+            endpoint += "?base_commit=${baseCommit}&pull_request=${pullRequest}"
+        }
+
+        def request = "curl ${post} ${headerAuth} ${headerContentType} ${body} \"${endpoint}\""
+
+        project.logger.quiet "Making post request: ${request} ..."
+
+        project.exec {
+            commandLine 'bash', '-c', request
```
rlazo: I'm not 100% sure that relying in curl for this is the best option, but I'll not object it.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/TestLogFinder.groovy
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.gradle.plugins.measurement
+
+/** A utility class that finds test log links on different CI systems. */
+class TestLogFinder {
+
+    /** Returns the link to the current running test. */
+    static String getCurrentLogLink() {
```
rlazo: nit: `generateCurrentLogLink` sounds like a better fit since you are generating the url, but it's a matter of style.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.groovy
yifanyang: Done.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.groovy
yifanyang: Fixed. Previously in case of non-Prow CI, the post request will be sent and fail silently; now it will not be sent at all.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.groovy
yifanyang: I agree with your concern. The best option is probably to sent the request programatically.

My thinking is that this request needs an authentication header, which comes from running `gcloud auth print-identity-token`. There doesn't seem to be a way to run gcloud command programatically, so seems like some shell command is unavoidable. Also this part of code will only be run on CI (it will be skipped when running locally), and we have control over the container image, then it should be fine.

I can change to use some java http client if you prefer that way.

buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/TestLogFinder.groovy
yifanyang: Done. Changed to `generateCurrentLogLink`.

PR #1322, opened 2020-03-06
Make Logging consistent across Crashlytics

jakeouellette:
- Remove tags from internal logger, all tags are FirebaseCrashlytics, and we don't want to accidentally use something else.
- Make NDK use this internal logger

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsCore.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsCore.java
@@ -129,7 +129,7 @@ public class CrashlyticsCore {
     // before starting the crash detector make sure that this was built with our build
     // tools.
     final String mappingFileId = CommonUtils.getMappingFileId(context);
-    Logger.getLogger().d(Logger.TAG, "Mapping file id is: " + mappingFileId);
+    Logger.getLogger().d("Mapping file ID is: " + mappingFileId);
 
     // Throw an exception and halt the app if the build ID is required and not present.
     // TODO: This flag is no longer supported and should be removed, as part of a larger refactor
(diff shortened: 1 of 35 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CommonUtils.java
```diff
@@ -430,7 +424,7 @@ public static void logControlledError(Context context, String msg, Throwable tr)
    */
   public static void logControlled(Context context, int level, String tag, String msg) {
```
jakeouellette: This is an interesting one: the tag was never used, and it still isn't.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/Logger.java
```diff
@@ -33,65 +35,65 @@ public static Logger getLogger() {
     return DEFAULT_LOGGER;
   }
 
-  public boolean isLoggable(String tag, int level) {
+  private boolean canLog(int level) {
```
jakeouellette: This name had to change because there is a heuristic lint check, `LogTagMismatch` that was getting confused by the lack of tag

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsCore.java
```diff
@@ -129,7 +129,7 @@ public boolean onPreExecute(SettingsDataProvider settingsProvider) {
     // before starting the crash detector make sure that this was built with our build
     // tools.
     final String mappingFileId = CommonUtils.getMappingFileId(context);
-    Logger.getLogger().d(Logger.TAG, "Mapping file id is: " + mappingFileId);
+    Logger.getLogger().d("Mapping file id is: " + mappingFileId);
```
mrwillis21: Can we just fix this to be "ID"

jakeouellette: Done!

PR #1323, opened 2020-03-06
Simplify finalizing native reports

mrwillis21:
Collapse finalizing previous native report into existing session
finalize operation.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
@@ -861,8 +861,7 @@ class CrashlyticsController {
       // data when we aren't including current.
       finalizePreviousNativeSession(mostRecentSessionIdToClose);
       if (!nativeComponent.finalizeSession(mostRecentSessionIdToClose)) {
-        Logger.getLogger()
-            .d("Could not finalize native session: " + mostRecentSessionIdToClose);
+        Logger.getLogger().d("Could not finalize native session: " + mostRecentSessionIdToClose);
       }
     }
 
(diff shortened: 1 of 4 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
```diff
@@ -858,6 +858,9 @@ private void doCloseSessions(int maxCustomExceptionEvents, boolean includeCurren
 
     if (includeCurrent) {
       reportingCoordinator.onEndSession();
+    } else if (nativeComponent.hasCrashDataForSession(mostRecentSessionIdToClose)) {
```
jakeouellette: can you add some justification for why this is an else if? why do we only do native stuff if includeCurrent = false? Why do we not do it if includeCurrent = false?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
```diff
@@ -858,6 +858,9 @@ private void doCloseSessions(int maxCustomExceptionEvents, boolean includeCurren
...
+      finalizePreviousNativeSession(mostRecentSessionIdToClose);
+      nativeComponent.finalizeSession(mostRecentSessionIdToClose);
```
jakeouellette: before we were returning a boolean and logging based on it, do we need to do an equivalent?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
mrwillis21: Yeah, this is one of those "the code needs more cleaning" issues. `includeCurrent` implies this is being run at Java crash time. We only want to run this code for the startup-time initialize.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
mrwillis21: Yeah, no harm in that.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/core/CrashlyticsController.java
mrwillis21: Added a comment.

PR #1326, opened 2020-03-09
Update the "Getting Started" instructions in README.md.

dconeybe:


dconeybe:
Note that the 10-minute wait is an unavoidable step that will occur regardless of the contents of the `.idea` folder.  No development can occur in Android Studio until the Gradle Sync task is completed.  This is only a one-time cost though and is only incurred once per computer (a re-import will not suffer from this long wait time) because Gradle's caches are shared across projects.

I think our best solution is to delete the contents of `.idea/runConfigurations` from the repository and instead supply instructions to add these run configurations manually via the Android Studio UI.  The steps are very easy.  As an added benefit, non-firestore development does not have to see the firestore run configurations in the drop-down menu.

dconeybe:
I'll create another PR for the change to delete `.idea/runConfigurations` and simply document how to add them manually.

I also forgot to mention that there doesn't appear to be a way to "import" and "export" run configurations from the Android Studio UI.  We *could* tell people to copy files into the `.idea` folder but IMO that is more work than simply adding them via the UI.

dconeybe:
Thank you for your thoughtful feedback, @wilhuff.  However, I am going to close this PR and not merge it.  The main motivation for the documentation update was to call out the annoying steps of deleting the .idea folder, etc, which, as of PR #1330, are no longer required.  Therefore, this documentation update is now obsolete.

PR #1328, opened 2020-03-09
Update Firestore run configurations to work in Android Studio 3.6.1

dconeybe:


.idea/runConfigurations/Firestore_Integration_Tests__Firestore_Emulator_.xml
```diff
@@ -1,19 +1,18 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Firestore Integration Tests (Firestore Emulator)" type="AndroidTestRunConfigurationType" factoryName="Android Instrumented Tests">
     <module name="firebase-firestore" />
-    <option name="TESTING_TYPE" value="1" />
+    <option name="TESTING_TYPE" value="0" />
     <option name="INSTRUMENTATION_RUNNER_CLASS" value="" />
     <option name="METHOD_NAME" value="" />
     <option name="CLASS_NAME" value="" />
-    <option name="PACKAGE_NAME" value="com.google.firebase.firestore" />
```
wilhuff: Removing the package qualification seems wrong (especially since you're adding it below).

.idea/runConfigurations/Firestore_Unit_Tests.xml
```diff
@@ -1,14 +1,15 @@
...
   <configuration default="false" name="Firestore Unit Tests" type="AndroidJUnit" factoryName="Android JUnit">
     <module name="firebase-firestore" />
+    <option name="PACKAGE_NAME" value="com.google.firebase.firestore" />
     <option name="MAIN_CLASS_NAME" value="" />
     <option name="METHOD_NAME" value="" />
     <option name="TEST_OBJECT" value="directory" />
     <option name="PARAMETERS" value="" />
-    <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/firebase-firestore" />
-    <dir value="$PROJECT_DIR$/firebase-firestore/src/test/java/" />
+    <option name="WORKING_DIRECTORY" value="$MODULE_DIR$" />
```
wilhuff: This is now inconsistent with the line below it.

.idea/runConfigurations/Firestore_Integration_Tests__Firestore_Emulator_.xml
dconeybe: The change of `TESTING_TYPE` from `1` to `0` above reflects changing the "Testing Type" from "All in package" to "All in module" in the Android Studio UI.  The "All in Module" value will run all tests in the `androidTest` folder regardless of their Java package name.  As a result, there is no need to specify the Java package of the tests to execute, and hence the removal of this line.  To me, this change makes the configuration slightly more robust: since we want to run all tests I think it's better to do that than to specify to run all tests in a given Java package, which just happens to contain all of the tests.

dconeybe: Also, in response to your comment about what this "fixes"... the change to this file does not fix anything.  It is merely a clean-up.

dconeybe:
As it turns out, the real culprit for the Android Studio import not working was the absence of `build.gradle`. This was fixed by PR #1330. Therefore, these changes to the run configurations are not required to fix the import.  As a result, I will close this PR.

PR #1330, opened 2020-03-09
Rename root-project.gradle to build.gradle

### LINKED ISSUE #1329: Android Studio Import Fails

  * Android Studio version: 3.6.1
  * Firebase Component: all
  * Component version: n/a

Importing the project into Android Studio 3.6.1 fails.  This appears to be a new issue in this latest version of Android Studio.  It appears to be due to the absence of `build.gradle` from the root directory of the repository.  Creating an empty `build.gradle` file prior to the import fixes the import.

#### Steps to reproduce:

*Steps to Reproduce*
1. Run `git clean -dfx` in your local clone of the Git repository.
1. Launch Android Studio.
1. Click "Import project (Gradle, Eclipse ADT, etc.)"
1. Select the root directory of the Git repository and click "Ok"

*Expected Results*
Android Studio successfully imports the project.  Gradle Sync completes successfully, then the "project" view shows all of the sub-projects.

*Actual Results*
Android Studio appears to successfully import the project; however, the "Project" view just shows the list of files in the root directory of the Git repository.  None of the code navigation works.  After Android Studio is restarted, then things appear to be better, but they are not: the "Project" view can now be expanded to navigate subdirectories but the sub-projects are not detected to be Gradle modules. Also, there is a notice at the bottom that reads "Frameworks Detected, Android framework is detected. Configure".  Clicking "Configure" leads to some very weird, non-functional project layout.

dconeybe:
Fixes #1329


vkryachko:
Awesome! thanks for the upgrade

/cc @rlazo

PR #1332, opened 2020-03-09
Integrate FIS into FIAM Headless

JasonAHeron:


firebase-inappmessaging/src/androidTest/java/com/google/firebase/inappmessaging/FirebaseInAppMessagingFlowableTest.java
```diff
@@ -223,22 +225,31 @@ public void setUp() {
...
         spy((Application) InstrumentationRegistry.getTargetContext().getApplicationContext());
-    String id = FirebaseInstanceId.getInstance().getId();
-    when(instanceId.getId()).thenReturn(id);
-    when(instanceId.getInstanceId())
+    when(firebaseInstallations.getId()).thenReturn(Tasks.forResult(INSTANCE_ID));
+    when(firebaseInstallations.getToken(false))
         .thenReturn(
             Tasks.forResult(
-                new InstanceIdResult() {
+                new InstallationTokenResult() {
```
vkryachko: Why anonymous subclass? How about `InstallationTokenResult.builder()...build()`?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/AbtIntegrationHelper.java
```diff
@@ -17,16 +17,21 @@
...
 import java.util.Date;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import javax.inject.Inject;
 
 /** @hide */
+@FirebaseAppScope
 public class AbtIntegrationHelper {
-  private static FirebaseABTesting abTesting;
+  private FirebaseABTesting abTesting;
```
vkryachko:
```suggestion
  private final FirebaseABTesting abTesting;
```
?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
```diff
@@ -89,28 +89,35 @@ static FetchEligibleCampaignsResponse createCacheExpiringResponse() {
...
-                              .setProjectNumber(firebaseApp.getOptions().getGcmSenderId())
-                              .addAllAlreadySeenCampaigns(
-                                  impressionList.getAlreadySeenCampaignsList())
-                              .setClientSignals(getClientSignals())
-                              .setRequestingClientApp(getClientAppInfo(instanceIdResult))
-                              .build()));
-            });
+    return firebaseInstallations
+        .getId()
+        .continueWithTask(
```
vkryachko: Why are you using continueWithTask? do you want to proceed even if this task is cancelled?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
```diff
@@ -89,28 +89,35 @@ static FetchEligibleCampaignsResponse createCacheExpiringResponse() {
...
+            idTask ->
+                firebaseInstallations
+                    .getToken(false)
```
vkryachko: It looks like this method can happen in parallel with the above `getId()` any reason you are doing them sequentially?, i.e.  could you try doing `Tasks.whenAll(idTask, tokenTask).continueWith()`?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/DataCollectionHelper.java
```diff
@@ -43,16 +42,10 @@
   public DataCollectionHelper(
       FirebaseApp firebaseApp,
       SharedPreferencesUtils sharedPreferencesUtils,
-      FirebaseInstanceId firebaseInstanceId,
       Subscriber firebaseEventsSubscriber) {
     this.sharedPreferencesUtils = sharedPreferencesUtils;
     isGlobalAutomaticDataCollectionEnabled =
         new AtomicBoolean(firebaseApp.isDataCollectionDefaultEnabled());
-    if (isAutomaticDataCollectionEnabled()) {
-      // Trigger this as early as possible, to minimize any latencies on returning the token
```
vkryachko: Is this no longer desirable? Or did FIS change in a way that makes it a non-issue?

firebase-inappmessaging/src/test/java/com/google/firebase/inappmessaging/FirebaseInAppMessagingTest.java
```diff
@@ -167,11 +165,34 @@ public void setup() {
 
     when(campaignCacheClient.get()).thenReturn(Maybe.just(campaignsResponse));
 
-    when(firebaseInstanceId.getId()).thenReturn(INSTANCE_ID);
-    when(firebaseInstanceId.getToken()).thenReturn(INSTANCE_TOKEN);
+    when(firebaseInstallations.getId()).thenReturn(Tasks.forResult(INSTANCE_ID));
+    when(firebaseInstallations.getToken(false))
+        .thenReturn(
+            Tasks.forResult(
+                new InstallationTokenResult() {
```
vkryachko: Builder here as well?

firebase-inappmessaging/src/test/java/com/google/firebase/inappmessaging/FirebaseInAppMessagingTest.java
```diff
@@ -307,11 +336,11 @@ public void messageClicked(InAppMessage inAppMessage, Action action) {
...
   }
 
-  @Implements(FirebaseInstanceId.class)
-  public static class ShadowFirebaseInstanceId {
-    @Implementation
-    public static FirebaseInstanceId getInstance() {
-      return firebaseInstanceId;
-    }
-  }
+  //  @Implements(FirebaseInstanceId.class)
```
vkryachko: Could this be deleted?

firebase-inappmessaging/src/test/java/com/google/firebase/inappmessaging/internal/ApiClientTest.java
```diff
@@ -118,26 +118,37 @@ public void setup() throws NameNotFoundException {
...
             providerInstaller);
     when(application.getPackageName()).thenReturn(PACKAGE_NAME);
     when(packageManager.getPackageInfo(PACKAGE_NAME, 0)).thenReturn(packageInfo);
-    when(firebaseInstanceId.getInstanceId())
+    when(firebaseInstallations.getId()).thenReturn(Tasks.forResult(INSTANCE_ID));
+    when(firebaseInstallations.getToken(false))
         .thenReturn(
             Tasks.forResult(
-                new InstanceIdResult() {
+                new InstallationTokenResult() {
```
vkryachko: Builder here and throughout

firebase-inappmessaging/src/androidTest/java/com/google/firebase/inappmessaging/FirebaseInAppMessagingFlowableTest.java
JasonAHeron: Brilliant, I didn't notice this was a builder!

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
JasonAHeron: Nice! done

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
JasonAHeron: so now I changed it to continueWith and I check both results against null. Am I not handling error cases correctly?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/DataCollectionHelper.java
JasonAHeron: This just felt like a premature optimization and it seemed like this would be a good time to understand if it was actually necessary

firebase-inappmessaging/src/test/java/com/google/firebase/inappmessaging/FirebaseInAppMessagingTest.java
JasonAHeron: done

firebase-inappmessaging/src/test/java/com/google/firebase/inappmessaging/FirebaseInAppMessagingTest.java
JasonAHeron: yep!

firebase-inappmessaging/src/test/java/com/google/firebase/inappmessaging/internal/ApiClientTest.java
JasonAHeron: yep

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
vkryachko: I am actually not sure that null can be returned at all from those tasks, if an error happened `getResult()` will throw that error so I think it should be fine in terms of error handling.

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
```diff
@@ -89,13 +89,15 @@ static FetchEligibleCampaignsResponse createCacheExpiringResponse() {
...
+    return Tasks.whenAll(idTask, tokenTask)
         .continueWith(
-            instanceIdResultTask -> {
-              InstanceIdResult instanceIdResult = instanceIdResultTask.getResult();
-              if (instanceIdResult == null) {
-                Logging.logw("InstanceID is null, not calling backend");
+            (unused) -> {
+              String idResult = idTask.getResult();
+              InstallationTokenResult tokenResult = tokenTask.getResult();
+              if (tokenResult == null || idResult == null) {
```
ankitaj224: None of the tasks will return null. Its on me for not adding the annotations on the APIs. I will update that on my end.

The Tasks would throw exceptions which you might want to handle based on completion state.

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/MetricsLoggerClient.java
```diff
@@ -83,78 +83,103 @@ public MetricsLoggerClient(
...
+                    createDismissEntry(message, id, dismissTransform.get(dismissType))
+                        .toByteArray());
+                logEventAsync(message, AnalyticsConstants.ANALYTICS_DISMISS_EVENT, false);
+              });
     }
   }
 
-  private CampaignAnalytics createEventEntry(InAppMessage message, EventType eventType) {
-    return createCampaignAnalyticsBuilder(message).setEventType(eventType).build();
+  private CampaignAnalytics createEventEntry(InAppMessage message, String id, EventType eventType) {
```
ankitaj224: nit: call it fid or instanceId to differentiate from any other type of id used in FIAM. Idk if there is one, just think it would avoid any confusion.

firebase-inappmessaging/firebase-inappmessaging.gradle
```diff
@@ -116,17 +116,17 @@ dependencies {
     implementation project(':protolite-well-known-types')
     implementation project(':transport:transport-api')
     implementation project(':firebase-datatransport')
+    implementation project(':firebase-installations-interop')
```
ankitaj224: Wondering if you should depend on the released version vs the HEAD? Idk whats the standard way in firebase.

PR #1336, opened 2020-03-10
Consolidate remaining 'core' classes into 'internal/common'

mrwillis21:


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/CommonUtilsTest.java
```diff
@@ -288,6 +295,71 @@ public void testResolveBuildId_bothIds() throws Exception {
...
+      oldFiles.add(createEmptyClsFile(testDir));
+
+      // This should have no effect - we're not over the limit
+      Utils.capFileCount(
+          testDir, clsFilter, maxFiles, CrashlyticsController.SMALLEST_FILE_NAME_FIRST);
+      assertEquals(oldFiles.size(), testDir.listFiles(clsFilter).length);
+    }
+    // The filesystem only has 1sec precision, so we need to sleep long enough that
+    // the timestamps in the 2nd set of files are later than the ones in the first.
+    Thread.sleep(1500);
```
jakeouellette: Thread.sleep behaviors are not ideal in tests because it can slow your overall test suite. Are there any alternative ways we could do this (e.g., abstract / mock more systems, or loop and check for state existing within some time bounds, and exit early if state becomes present?

PR #1337, opened 2020-03-10
Remove IID from FIAM Display Dependencies

JasonAHeron:


JasonAHeron:
Hey Rayo, this actually just removes an unused IID dep from FIAM Display. This is not to integrate FIS. The FIAM FIS integration CL is already in master

firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -75,8 +75,8 @@
  *   <li>send usage metrics to the Firebase backend.
  * </ul>
  *
- * <p>To delete the Instance ID and the data associated with it, see {@link
- * com.google.firebase.iid.FirebaseInstanceId#deleteInstanceId}.
+ * <p>To delete the Instance ID and the data associated with it, see
+ * FirebaseInstallationsApi#delete.
```
prakhar1989: I think we should use `@link` and provide the full [package name](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java#L15)

JasonAHeron: If we do that then we need to import the package...

PR #1338, opened 2020-03-10
Annotate FirebaseInstalaltionsApi class

ankitaj224:
To avoid the confusion for clients integrating with FIS around null checks.


firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -28,15 +29,18 @@
    * Async function that returns a globally unique identifier of this Firebase app installation.
    * This is a url-safe base64 string of a 128-bit integer.
    */
+  @NonNull
   Task<String> getId();
 
   /** Async function that returns a auth token(public key) of this Firebase app installation. */
```
andirayo: Please extend the JavaDoc with @param forceRefresh:
Example:
@param forceRefresh If set to TRUE this method will trigger a server request every time when called. Otherwise it will return a locally cached token if available and valid. It is strongly recommended to set this field to FALSE unless a previously returned token turned out to be expired (which in itself is an indicator for time synchronization issues).

PR #1343, opened 2020-03-11
Remove ABT Experiment Cleanup from FIAM fetch path

JasonAHeron:
Additionally, cleanup the way Installation IDs are handled


firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -243,7 +243,6 @@ private boolean shouldIgnoreCache(String event) {
                                           "Successfully fetched %d messages from backend",
                                           resp.getMessagesList().size())))
                           .doOnSuccess(analyticsEventsManager::updateContextualTriggers)
-                          .doOnSuccess(abtIntegrationHelper::updateRunningExperiments)
```
prakhar1989: Why has this been removed?

JasonAHeron: this is the whole point of this PR. This updateRunningExperiments removes the active experiments which are not currently running

PR #1345, opened 2020-03-11
Improving FISServiceClient by closing all acquired connection resources

ankitaj224:
Also, cleaning up code duplication as per the b/150658215


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -176,20 +181,23 @@ public InstallationResponse createFirebaseInstallation(
...
     } catch (JSONException e) {
       throw new IllegalStateException(e);
-    } finally {
-      gzipOutputStream.close();
+    }
+  }
+
+  private static void writeRequestBodyToOutputStream(
+      HttpURLConnection httpURLConnection, JSONObject jsonObject) throws IOException {
+    try (OutputStream outputStream = httpURLConnection.getOutputStream()) {
```
andirayo: I'm not too familiar with this pattern.
Are OutputStreams always AutoCloseable?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -176,20 +181,23 @@ public InstallationResponse createFirebaseInstallation(
...
+      if (outputStream == null) {
+        throw new IOException("Cannot send request to FIS servers. No OutputStream available.");
+      }
+      try (GZIPOutputStream gzipOutputStream = new GZIPOutputStream(outputStream)) {
+        gzipOutputStream.write(jsonObject.toString().getBytes("UTF-8"));
```
andirayo: Please move the calls to .toString().getBytes("UTF-8") outside of this method.
This method should have one "single responsibility":
Writing some content to a stream (and not do transformations on that content).

===============

If you want to remove code duplication for these methods, 
you should name this method appropriately, example"
#writeJsonToOpenConnection()

Even better would be to create a new method with the above name
which does the transformation and then calls this existing method that writes to the given UrlConnection.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -176,20 +181,23 @@ public InstallationResponse createFirebaseInstallation(
...
+          httpURLConnection, buildCreateFirebaseInstallationRequestBody(fid, appId));
     } catch (JSONException e) {
       throw new IllegalStateException(e);
-    } finally {
-      gzipOutputStream.close();
+    }
+  }
+
+  private static void writeRequestBodyToOutputStream(
+      HttpURLConnection httpURLConnection, JSONObject jsonObject) throws IOException {
```
andirayo: String content instead of JSONObject jsonObject

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -401,13 +406,14 @@ private InstallationResponse readCreateResponse(HttpURLConnection conn) throws I
...
-
+    inputStream.close();
     return builder.setResponseCode(ResponseCode.OK).build();
   }
 
   // Read the response from the generateAuthToken FirebaseInstallation API.
   private TokenResult readGenerateAuthTokenResponse(HttpURLConnection conn) throws IOException {
-    JsonReader reader = new JsonReader(new InputStreamReader(conn.getInputStream(), UTF_8));
+    InputStream inputStream = conn.getInputStream();
+    JsonReader reader = new JsonReader(new InputStreamReader(inputStream, UTF_8));
```
andirayo: in a future CL we should probably remove this code duplication

ankitaj224: Ack.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: That's right. OutputStream implements Closeable

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Done.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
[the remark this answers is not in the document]
ankitaj224: Done.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -167,6 +167,11 @@ public InstallationResponse createFirebaseInstallation(
         return InstallationResponse.builder().setResponseCode(ResponseCode.BAD_CONFIG).build();
       } finally {
         httpURLConnection.disconnect();
+        try {
+          httpURLConnection.getInputStream().close();
+        } catch (IOException unused) {
```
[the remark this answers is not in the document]
ankitaj224: Removed this as we dont have any open input streams at this point. It was just an experiement.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -177,20 +180,27 @@ public InstallationResponse createFirebaseInstallation(
...
+    }
+  }
+
+  private static byte[] getJsonBytes(JSONObject jsonObject) throws UnsupportedEncodingException {
+    return jsonObject.toString().getBytes("UTF-8");
+  }
+
+  private static void writeRequestBodyToOutputStream(URLConnection urlConnection, byte[] jsonBytes)
+      throws IOException {
+    try (OutputStream outputStream = urlConnection.getOutputStream()) {
```
andirayo, line 197: Probably this is not an issue, but I just want to make sure that we are not running into an issue here:
If outputStream is NULL this will not cause an error when Java is trying to close the stream, right?
Did you test this?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -177,20 +180,27 @@ public InstallationResponse createFirebaseInstallation(
...
+      writeRequestBodyToOutputStream(
+          httpURLConnection, getJsonBytes(buildCreateFirebaseInstallationRequestBody(fid, appId)));
     } catch (JSONException e) {
       throw new IllegalStateException(e);
-    } finally {
-      gzipOutputStream.close();
+    }
+  }
+
+  private static byte[] getJsonBytes(JSONObject jsonObject) throws UnsupportedEncodingException {
```
andirayo, line 191: I wonder if this creates more confusion than clarity. What do you think?

Maybe it's better to just say `throws IOException` and explain in the JavaDoc (or even an inline comment) that it is a `UnsupportedEncodingException`?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -177,20 +180,27 @@ public InstallationResponse createFirebaseInstallation(
...
-    }
-
-    GZIPOutputStream gzipOutputStream = new GZIPOutputStream(outputStream);
     try {
-      gzipOutputStream.write(
-          buildCreateFirebaseInstallationRequestBody(fid, appId).toString().getBytes("UTF-8"));
+      writeRequestBodyToOutputStream(
+          httpURLConnection, getJsonBytes(buildCreateFirebaseInstallationRequestBody(fid, appId)));
     } catch (JSONException e) {
       throw new IllegalStateException(e);
```
andirayo, line 187: Where is the `JSONException` thrown? How about catching this `JSONException` where it is thrown and throw an IllegalStateException instead. That way this method even becomes cleaner and we have less code duplication (of catching this exception).
If you do, please add in JavaDoc of this comment that it might throw an `IllegalStateException` in case there is a problem with the JSON.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
andirayo: Saw this again. Happy to chat offline.

ankitaj224: We can discuss offline and if there are changes, they will be part of another PR.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java

PR #1346, opened 2020-03-12
Prototype a lint check for fis.

vkryachko:


vkryachko:
@ankitaj224 the check will run but it will not find iid among dependencies, hence it will not fail

ankitaj224:
> @ankitaj224 the check will run but it will not find iid among dependencies, hence it will not fail

In case of FIAM- that's correct. But if some other teams have iid dependency I can assume it will correctly identify that.

PR #1347, opened 2020-03-12
Lint jar for IID Version incompatabiltiy

ankitaj224:
Patched and improved https://github.com/firebase/firebase-android-sdk/pull/1346

@vkryachko  please merge your PR and I ll update the my branch  accordingly.


firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    private fun isIncompatibleVersion(coordinates: MavenCoordinates): Boolean {
+        // firebase-iid v20.1.0 already depended on FirebaseInstallations for identifier creation,
+        // but had an issue of possible silent failures. Thus, v20.1.1 or later is recommended.
+        val validIidVersionWithFis = "20.1.1"
```
andirayo: I don't think developers need this kind of context.
Just tell them which version to use.

If this comment is meant for us, then I would recommend to rephrase:
"firebase-iid v20.1.0 is compatible with FirebaseInstallations, but contains a bug that was fixed in v20.1.1"

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+        )
+
+        @JvmField
+        val INCOMPATIBLE_IID_VERSION = Issue.create(
+                id = "IncompatibleIidVersion",
+                briefDescription = "FirebaseInstanceId version is not compatible with FirebaseInstallations",
+
+                explanation = """
+           These versions of FirebaseInstanceId and FirebaseInstallations can not be used in parallel.
+           Please update your Firebase Instance ID to a version 20.1.0 or later.
```
andirayo: There seems to be an inconsistent Inconsistent use of versions!
Sometimes it's 20.1.0, sometimes 20.1.1 (see below).
I assume that you want to prevent usage of <20.1.0 and recommend usage of v20.1.1...

I think, you would make your life easier by just marking everything before v20.1.1 as incompatible and keep it consistent.

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        if (versionAComponents.size == versionBComponents.size) {
+            return 0
+        }
+        if (versionAComponents.size < versionBComponents.size) {
+            return -1
+        } else {
+            return 1
+        }
+    }
```
andirayo: This method seems unnecessarily complicated.
Can't we just check for the version that we are interested in?
kind of like this (pseudo code):

versionComponents = version.split('.')
// check if correct SemVer format
return FALSE if 3 != versionComponents.size
// major version before v20? => incompatible
return FALSE if 20 > versionComponants[0].asInt
// major version after v21? => compatible
return TRUE if 21 <= versionComponants[0].asInt
// major version is v20 and minor version major version after v20.1? => compatible
return (1 <= versionComponants[1].asInt)

vkryachko:
@ankitaj224 ptal at #1346

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
ankitaj224: Agreed. Thanks for catching this.

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
ankitaj224: Updated.

vkryachko:
@ankitaj224 a few points:
* we should try to support as many android gradle plugin versions as possible, right now it's >= 3.4 but we can make it 3.2+ I believe and this change should do it: https://github.com/firebase/firebase-android-sdk/commit/0e779e943fbba97e5bc27d5fe0c7442211a8b5c9
* ask lint folks for feedback on the implementation before/after merging this change

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
vkryachko: This explanation is not the part that developers are most likely to see, instead this is the error they will see in their build:

```
/Users/vkryachko/Code/firebase/firebase-android-sdk/firebase-inappmessaging-display: Error: Incompatible IID version found in variant debug: com.google.firebase:firebase-iid:20.0.1@aar [IncompatibleIidVersion]
```

With that in mind I think it could be improved in a few ways:
1. Most likely developers don't have a direct dependency on iid and don't even know what iid is, it might make sense mention that the dependency may be transitive and how it most likely ended up in their build
2. "@aar" part may confuse developers, consider dropping it
3. Consider adding a call for action here like how exactly the problem can be solved with a code example to copy paste into their build file.

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -14,22 +14,15 @@
...
-
 import com.android.builder.model.MavenCoordinates
-import com.android.tools.lint.detector.api.Category
-import com.android.tools.lint.detector.api.Context
-import com.android.tools.lint.detector.api.Implementation
-import com.android.tools.lint.detector.api.Issue
-import com.android.tools.lint.detector.api.Location
-import com.android.tools.lint.detector.api.Scope
-import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.*
```
andirayo: This is against Java readability rules in Google I think. Not sure how this is handled here?

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -59,15 +53,29 @@ class IncompatibleIidVersionDetector : Detector() {
...
     private fun isIncompatibleVersion(coordinates: MavenCoordinates): Boolean {
-        // TODO: implement version check
-        return coordinates.version == "0.0.0"
+        val versionComponents = coordinates.version.split('.').toTypedArray()
+        // check if correct version format
+        if (3 != versionComponents.size) {
+            return true
+        }
+        // Incompatible if major version is before v20
+        if (20 > versionComponents.get(0).toInt()) {
```
andirayo: Can #toInt throw an exception?
Do we care?

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -59,15 +53,29 @@ class IncompatibleIidVersionDetector : Detector() {
...
                     }
                 }
             }
         }
     }
 
     private fun isIncompatibleVersion(coordinates: MavenCoordinates): Boolean {
-        // TODO: implement version check
-        return coordinates.version == "0.0.0"
+        val versionComponents = coordinates.version.split('.').toTypedArray()
```
andirayo: Can you tell split to split this into maximum 3 components?
If yes, please do so.

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
```diff
@@ -59,15 +53,29 @@ class IncompatibleIidVersionDetector : Detector() {
...
-                                "Incompatible IID version found in variant ${variant.name}: ${lib.name}")
+                                "Incompatible IID version found in variant ${variant.name}: ${lib.name.removeSuffix("@aar")}.\n" +
+                                        "To resolve this issue, consider updating the build file to the latest firebase-iid version.")
                     }
                 }
             }
         }
     }
 
     private fun isIncompatibleVersion(coordinates: MavenCoordinates): Boolean {
```
andirayo: Generally it's not a good idea to use negative words in method or variable names.
Thus, I would recommend to call this #isCompatibleInstanceIdVersion

firebase-installations/customer-lint-checks/src/main/java/com/google/firebase/installations/lint/IncompatibleIidVersionDetector.kt
ankitaj224: Done.

PR #1349, opened 2020-03-13
Nudge customers to update if on an old Firebase Analytics

mrichards:
We can infer the customer is on an outdated version of FA
if we cannot attach a listener with the "clx" origin. If
that is the case, use the old "crash" origin and write a
logcat warning with a nudge to upgrade.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
@@ -69,11 +69,7 @@ public class AnalyticsConnectorReceiver implements AnalyticsConnectorListener, A
 
     if (analyticsConnectorHandle == null) {
       Logger.getLogger()
-          .d(
-              "Could not register AnalyticsConnectorListener with origin "
-                  + "\""
-                  + CRASHLYTICS_ORIGIN
-                  + "\"");
+          .d("Could not register AnalyticsConnectorListener with Crashlytics origin.");
       // Older versions of FA don't support CRASHLYTICS_ORIGIN. We can try using the old Firebase
       // Crash Reporting origin
       analyticsConnectorHandle =
@@ -84,9 +80,9 @@ public class AnalyticsConnectorReceiver implements AnalyticsConnectorListener, A
       if (analyticsConnectorHandle != null) {
         Logger.getLogger()
             .w(
-                "Outdated version of Firebase Analytics detected. For improved "
-                    + "performance and compatibility with Crashlytics, please update to the latest "
-                    + "version of Firebase Analytics.");
+                "A new version of the Google Analytics for Firebase SDK is now available. "
+                    + "For improved performance and compatibility with Crashlytics, please "
+                    + "update to the latest version.");
       }
     }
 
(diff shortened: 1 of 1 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
```diff
@@ -66,6 +67,29 @@ public boolean register() {
...
+      // Crash Reporting origin
+      analyticsConnectorHandle =
+          analyticsConnector.registerAnalyticsConnectorListener(LEGACY_CRASH_ORIGIN, this);
+
+      // If FA allows us to connect with the legacy origin, but not the new one, nudge customers
+      // to update their FA version.
+      if (analyticsConnectorHandle != null) {
+        Logger.getLogger()
+            .w(
+                "Outdated version of Firebase Analytics detected. For improved "
```
mrichards: I'm going to coordinate with the Analytics team on the specific wording, which is likely to change. I created the PR to get feedback on the general approach; don't worry about the warning message just yet.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
```diff
@@ -66,6 +67,29 @@ public boolean register() {
     analyticsConnectorHandle =
         analyticsConnector.registerAnalyticsConnectorListener(CRASHLYTICS_ORIGIN, this);
 
+    if (analyticsConnectorHandle == null) {
+      Logger.getLogger()
+          .d(
+              "Could not register AnalyticsConnectorListener with origin "
+                  + "\""
+                  + CRASHLYTICS_ORIGIN
```
mrwillis21: I don't think we want to expose this implementation detail to the user. Maybe something more generic like "Could not register analytics connector with Crashlytics origin"?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
```diff
@@ -66,6 +67,29 @@ public boolean register() {
...
+                  + "\"");
+      // Older versions of FA don't support CRASHLYTICS_ORIGIN. We can try using the old Firebase
+      // Crash Reporting origin
+      analyticsConnectorHandle =
+          analyticsConnector.registerAnalyticsConnectorListener(LEGACY_CRASH_ORIGIN, this);
+
+      // If FA allows us to connect with the legacy origin, but not the new one, nudge customers
+      // to update their FA version.
+      if (analyticsConnectorHandle != null) {
```
mrwillis21: Would it be worth consolidating this with the `return` value below?

mrwillis21:
A couple small notes. Overall though I'm happy with the approach.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
mrichards: Sure, NBD. In general I'm not opposed to exposing inconsequential (to the customer) implementation details in debug logs, but it doesn't really matter what the message is here, so long as we know how to interpret it. :)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/AnalyticsConnectorReceiver.java
mrichards: I couldn't come up with a way to do that which *helped readability* and the performance impact is so close to nil that it isn't worth optimizing. We could have fewer null checks but we'd need to sprinkle return statements throughout the method. So I (mildly) prefer it as-is.

mrwillis21: 👍

PR #1353, opened 2020-03-16
C++ API for Crashlytics NDK 

mrichards:
Update the crashlytics.h file to remove the legacy C API and add a new C++ API in the `firebase::crashlytics` namespace.

The 3rd commit is only formatting changes, so that a diff with the first two will show the relevant content changes.


firebase-crashlytics-ndk/src/main/jni/libcrashlytics/include/crashlytics/external/crashlytics.h
```diff
@@ -14,85 +14,151 @@
...
+    static inline __crashlytics_context_t* __crashlytics_init() __CRASHLYTICS_DECORATED;
+    static inline void                     __crashlytics_free(__crashlytics_context_t **context) __CRASHLYTICS_DECORATED;
+
+    __crashlytics_context_t *__context;
+
+    bool VerifyCrashlytics() {
+        if (__context) {
+            return true;
+        }
+        return false;
```
mrwillis21: Is the intent to log an error here, as you mentioned in the refdoc?

mrichards: It was, but then I removed it in the most recent commit so we wouldn't need to include any android headers, per kmandrika's suggestion. I'll update the comment to match reality.

mrichards:
> This all seems reasonable to me, though I'd love to get @kmandrika's sign-off on the syntax.

He did a stealthy backchannel review, resulting in commit #4 (5e1df4fd1f44133755509a346d4c29897095c038).

firebase-crashlytics-ndk/src/main/jni/libcrashlytics/include/crashlytics/external/crashlytics.h
```diff
@@ -14,85 +14,152 @@
...
+    }
+
+    void Initialize() {
```
kmandrika: All definitions within this file should be declared as inline, otherwise, you are going to run into a "multiple definition" compiler error. You can test this by including this file into two separate translation units and compiling.

PR #1354, opened 2020-03-16
Add datatransport support to Crashlytics NDK layer

jakeouellette:
Add support for uploading native reports via datatransport layer


jakeouellette:
Moved this code back to the mainline branch

Issue #1339, opened 2020-03-11
App crashing due to IllegalStateException


sudsingh438:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.6 RC 3
  * Firebase Component:  firebase-perf
  * Component version: 19.0.5

### [REQUIRED] Step 3: Describe the problem
Crashes are being reported on play console with below stack trace:

java.lang.IllegalStateException: 
  at com.google.firebase.iid.FirebaseInstanceId.zzl (FirebaseInstanceId.java:75)
  at com.google.firebase.iid.FirebaseInstanceId.getId (FirebaseInstanceId.java:49)
  at com.google.firebase.perf.internal.zzf.zzbu (zzf.java:181)
  at com.google.firebase.perf.internal.zzf.zzbt (zzf.java:44)
  at com.google.firebase.perf.internal.zzf.zzc (zzf.java:195)
  at com.google.firebase.perf.internal.zze.run (zze.java:2)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:764)

#### Steps to reproduce:

We are not able to reproduce it locally.

andirayo:
@sudsingh438 :

#### Questions
1. Which version of the Firebase Instance ID are you using?
1. Are you able to share the complete `IllegalStateException` stack trace that might show the cause?
1. When initializing Firebase (i.e. class `FirebaseApp`) are you initializing with a full set of Firebase options (e.g. using a `google-services.json` from the Firebase Console)?

#### Background
If the initialization of Firebase (i.e. instance of `FirebaseApp`) is missing one of the following required parameters, this might explain the issue: Application ID, Project ID, API key.

The latest version of the Instance ID SDK depends on the new Firebase Installations SDK which is a change that might cause issues like this.

(@ankitaj224 : for visibility)

sudsingh438:
andirayo:
1) Firebase instance id version : 20.1.1 
2) The one that I shared is the description is the complete stack trace from play console.
<img width="912" alt="Screenshot 2020-03-12 at 1 41 45 AM" src="https://user-images.githubusercontent.com/10254300/76459718-ae960180-6402-11ea-8a40-e939dd72a8d9.png">

This stack trace is the one we got from Sentry 
[...]

PR #1355, opened 2020-03-16
Moving FIS network calls onto a separate executor

ankitaj224:
https://github.com/firebase/firebase-android-sdk/issues/1339


ankitaj224:
Leaving the delete network call onto the main thread as we don't have any callbacks in the delete api.
@fredquintana  wdyt?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -307,7 +310,11 @@ private final void doRegistrationInternal(boolean forceRefresh) {
     }
 
     triggerOnStateReached(prefs);
+    networkExecutor.execute(() -> run(forceRefresh));
+  }
 
+  private void run(boolean forceRefresh) {
```
fredquintana: please give this a better name

firebase-installations/gradle.properties
```diff
@@ -1,2 +1,2 @@
-version=16.0.1
+version=16.0.2
 latestReleasedVersion=16.0.0
```
andirayo: I am not about the meaning of these, but does this have to be `v16.0.1`?
```
latestReleasedVersion=16.0.1
```

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -307,7 +310,11 @@ private final void doRegistrationInternal(boolean forceRefresh) {
     }
 
     triggerOnStateReached(prefs);
+    networkExecutor.execute(() -> run(forceRefresh));
```
fredquintana: just realized, we need to be a little careful, here.

If the subsequent call has a different value for "forceRefresh" then we need to be sure that it gets enqueued, otherwise the flag value can get lost.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

How about this:
#sendRequestToFirebaseInstallationsServerApi
#syncWithFirebaseInstallationsServerApi 
#updateLocalAuthToken

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -66,6 +67,7 @@
   private final IidStore iidStore;
   private final RandomFidGenerator fidGenerator;
   private final Object lock = new Object();
+  private final ExecutorService networkExecutor;
```
andirayo: Please cluster both the ExecutorServices below each other.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -66,6 +67,7 @@
   private final IidStore iidStore;
```
andirayo: How about changing the name of the existing threadpool?
private final ExecutorService backgroundExecutor;

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -124,6 +126,7 @@ public Thread newThread(Runnable r) {
     this.utils = utils;
     this.iidStore = iidStore;
     this.fidGenerator = fidGenerator;
+    this.networkExecutor = Executors.newSingleThreadExecutor();
```
andirayo: Q1: Why is this not injected like all the other variables and created by the other constructor method?
Q2: Is there a reason that we have different ways how backgroundExecutor and networkExecutor are initialized?

firebase-installations/gradle.properties
ankitaj224: Good catch. I am not sure about the variable either but seems right to update it.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Updated this. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: the force refresh values come from the main thread and will be added on to the network executor. I dont see how the flag gets lost/ overwritten.

@fredquintana wdyt can go wrong?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Q2: Idea was to simplify the creation but changed it now to replicate the backgroundExecutor. Reasoning: we had a bug reported by some internal team that checks for thread tags and has whitelisted only firebase-installations-executor. 

Q1: The network executor is an impl detail specific to FIS. Adding multiple executors in the constructor could lead to the confusion. I dont see how we can benefit from initializing through injection.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Done

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
andirayo: If possible I would recommend to always start a method name with a verb.

Also please remove the "fis" prefix:
a) "fis" is not a public term (so rather use "firebaseInstallations")
b) it's noise (we are in a class that is called FirebaseInstallations)

If you prefer your current name over the suggested ones, then I recommend to update it like this:
#sendNetworkCall
#makeNetworkCall

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -119,11 +120,14 @@ public Thread newThread(Runnable r) {
...
     this.serviceClient = serviceClient;
-    this.executor = executor;
     this.persistedInstallation = persistedInstallation;
     this.utils = utils;
     this.iidStore = iidStore;
     this.fidGenerator = fidGenerator;
+    this.backgroundExecutor = backgroundExecutor;
+    this.networkExecutor =
+        new ThreadPoolExecutor(
+            1, 1, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>(), threadFactory);
```
fredquintana: Executors.newCachedThreadPool() would be good here.

 If you choose to explicitly create it, please document the constants.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -307,7 +311,13 @@ private final void doRegistrationInternal(boolean forceRefresh) {
     }
 
     triggerOnStateReached(prefs);
+    // Execute network calls (CreateInstallations or GenerateAuthToken) to the FIS Servers on
+    // a separate executor i.e networkExecutor
+    networkExecutor.execute(() -> fisNetworkCall(forceRefresh));
+  }
 
+  private void fisNetworkCall(boolean forceRefresh) {
```
fredquintana: I think fis is redundant here. How about "doNetworkCall".

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
andirayo: Do I read this correctly?
0L Milliseconds is the timeout time?

Questions and Notes:
1) Can we use the same numbers that we use for the backgroundExecutor?
2) Please create constants and use them for both executors
3) Why is the core pool size different to the other executor?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
andirayo: +1.
I already made suggestions in the other thread (see below).
But #doNetWorkCall sounds totally reasonable.

#sendNetworkCall
#makeNetworkCall
#sendRequestToFirebaseInstallationsServerApi
#syncWithFirebaseInstallationsServerApi
#updateLocalAuthToken

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: 0L is the keep alive time - when the number of threads is greater than the core, this is the maximum time that excess idle threads will wait for new tasks before terminating.

Created constants that are shared between the two executors.

ankitaj224: As mentioned earlier, reason for explicitly creating is some 1p apps restrict creating unnamed thread pools. We got the ThreadFactory created in FIS whitelisted.

Refer to b/148952570

I documented the constants.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -119,11 +122,19 @@ public Thread newThread(Runnable r) {
...
     this.persistedInstallation = persistedInstallation;
     this.utils = utils;
     this.iidStore = iidStore;
     this.fidGenerator = fidGenerator;
+    this.backgroundExecutor = backgroundExecutor;
+    this.networkExecutor =
+        new ThreadPoolExecutor(
+            /* corePoolSize= */ CORE_POOL_SIZE,
+            /* maximumPoolSize= */ MAXIMUM_POOL_SIZE,
+            /* keepAliveTime= */ 0L,
```
fredquintana: did you intend this to be KEEP_ALIVE_TIME_IN_SECONDS?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -119,11 +122,19 @@ public Thread newThread(Runnable r) {
...
+            TimeUnit.MILLISECONDS,
```
fredquintana: change to SECONDS

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -94,12 +97,12 @@ public Thread newThread(Runnable r) {
...
-            /* corePoolSize= */ 0,
-            /* maximumPoolSize= */ 1,
-            /* keepAliveTime= */ 30L,
+            /* corePoolSize= */ CORE_POOL_SIZE,
+            /* maximumPoolSize= */ MAXIMUM_POOL_SIZE,
+            /* keepAliveTime= */ KEEP_ALIVE_TIME_IN_SECONDS,
             TimeUnit.SECONDS,
             /* workQueue= */ new LinkedBlockingQueue<>(),
-            /* threadFactory= */ threadFactory),
+            /* threadFactory= */ THREAD_FACTORY),
```
andirayo: No comments needed since you are using constants that explain the purpose of the parameters.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -119,11 +122,19 @@ public Thread newThread(Runnable r) {
...
     this.fidGenerator = fidGenerator;
+    this.backgroundExecutor = backgroundExecutor;
+    this.networkExecutor =
+        new ThreadPoolExecutor(
+            /* corePoolSize= */ CORE_POOL_SIZE,
+            /* maximumPoolSize= */ MAXIMUM_POOL_SIZE,
+            /* keepAliveTime= */ 0L,
+            TimeUnit.MILLISECONDS,
+            /* workQueue= */ new LinkedBlockingQueue<>(),
+            /* threadFactory= */ THREAD_FACTORY);
```
andirayo: No comments needed since you are using constants that explain the purpose of the parameters.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
andirayo: Why is keepAliveTime 0 seconds instead of 30 seconds?

PR #1356, opened 2020-03-17
Crashlytics ndk datatransport

jakeouellette:
* Add datatransport connection for NDK


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FileBackedNativeSessionFile.java
```diff
@@ -0,0 +1,77 @@
...
+            .setContents(asBytes())
+            .setFilename(name)
+            .build()
+        : null;
+  }
+
+  private byte[] asBytes() {
+    final byte[] readBuffer = new byte[8192];
+    final ByteArrayOutputStream bos = new ByteArrayOutputStream();
+    try (InputStream stream = new FileInputStream(f)) {
```
mrwillis21: Shouldn't we re-use `getStream` here?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/NativeSessionFileGzipper.java
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+  }
+
+  private static void gzipInputStream(@Nullable InputStream input, @NonNull File output)
+      throws IOException {
+    if (input == null) {
+      return;
+    }
+    byte[] buffer = new byte[1024];
```
mrwillis21: We should probably use 8192 here, as above.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
```diff
@@ -130,6 +132,27 @@ public void persistNonFatalEvent(Throwable event, Thread thread, long timestamp)
     persistEvent(event, thread, EVENT_TYPE_LOGGED, timestamp, false);
   }
 
+  public void finalizeNativeEvent(
```
mrwillis21: Minor stylistic request: Can we call this `finalizeSessionWithNativeEvent`, passing the session ID and the native files?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
```diff
@@ -130,6 +132,27 @@ public void persistNonFatalEvent(Throwable event, Thread thread, long timestamp)
...
+      for (NativeSessionFile nativeSessionFile : nativeSessionFiles) {
+        FilesPayload.File filePayload = nativeSessionFile.asFilePayload();
+        if (filePayload != null) {
+          nativeFiles.add(filePayload);
+        }
+      }
+
+      filesPayloadBuilder.setFiles(ImmutableList.from(nativeFiles));
+
+      reportPersistence.finalizeNativeEvent(
```
mrwillis21: Same as above, can we call this `finalizeSessionWithNativeEvent(sessionID, reportData, filesPayload)`

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
```diff
@@ -98,6 +97,10 @@ public static Builder builder() {
    */
   @NonNull
   public CrashlyticsReport withEvents(@NonNull ImmutableList<Event> events) {
+    if (this.getNdkPayload() != null) {
```
mrwillis21: Is there any concern that we might have a situation where this is null, but session is as well?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FileBackedNativeSessionFile.java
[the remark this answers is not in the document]
jakeouellette: It was originally this, but then we have to null check the stream because getStream() returns a nullable.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FileBackedNativeSessionFile.java
```diff
@@ -0,0 +1,77 @@
...
+      int read;
+      while ((read = stream.read(readBuffer)) > 0) {
+        bos.write(readBuffer, 0, read);
+      }
+      return new String(bos.toByteArray(), UTF_8).getBytes();
```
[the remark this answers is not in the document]
jakeouellette: This code was copied from https://github.com/firebase/firebase-android-sdk/blob/master/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java#L370 , any idea why that one doesn't?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/NativeSessionFile.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /** Representation of the NativeSessionFile as a stream */
+  @Nullable
+  InputStream getStream();
+
+  /**
+   * Representation of the NativeSessionFile as {@link CrashlyticsReport.FilesPayload.File} object.
+   */
+  @Nullable
+  CrashlyticsReport.FilesPayload.File asFilePayload();
```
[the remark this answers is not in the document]
jakeouellette: Short answer: Eventually getStream() will go away because this class will not have the old codepath, and this will just **be** the FilesPayloadSerializer.

Long answer:  
Yeah, I went back and forth on this a little bit myself. I didn't want to build the datatransport code on getStream() because it results in a double copy of a byte stream into a new byte stream.

(wearing OOP nerd hat) There's actually a way to solve this where you separate out the data (bytes, files) and behavior (gzip from stream of bytes, gzip from filestream, datatransport from bytes of files or bytes of bytes).

The way is multiple dispatch. On java though, multiple dispatch has to be implemented via visitor pattern (implement a stream converter that visits the bytes / files, then converts to a stream, implement a filepayloadconverter that visits the bytes / files, and converts them to bytes). Doing so is pretty verbose. Since the API for getStream() is going away, I decided it was okay to just flatten them into this api, but if you wanted, I can switch to visitor.

This whole multiple dispatch challenge, for what it is worth, is called the "expression problem" https://eli.thegreenplace.net/2016/the-expression-problem-and-its-solutions/

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/NativeSessionFileGzipper.java
jakeouellette: This was copypasta of the other place where we were zipping, happy to switch it (I assumed each was set for domain-specific reasons of the objects they were serializing)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
jakeouellette: sure

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/SessionReportingCoordinator.java
jakeouellette: sure

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
jakeouellette: I see what you mean, so check for getSession() == null and do illegal state then? sounds good

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FileBackedNativeSessionFile.java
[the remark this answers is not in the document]
mrwillis21: Because that's returning the read-in bytes _as a UTF-8 String_, you just need to return the read-in bytes here, no?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/NativeSessionFile.java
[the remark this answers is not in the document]
mrwillis21: 🤓 alright, leave it the way it is. 😝

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/NativeSessionFileGzipper.java
mrwillis21: Understood, yeah, no good reason for it.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
mrwillis21: 👍

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FileBackedNativeSessionFile.java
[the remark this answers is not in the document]
mrwillis21: I think it would be more consistent, since `getStream` is the underlying API. Besides, you're already re-catching the `FNFException` in this method.

If your concern is the `try-with-resources`, it works with nullable `AutoCloseable`s, you'd just need to check that `stream` is non-null inside the `try` block.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/FileBackedNativeSessionFile.java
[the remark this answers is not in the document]
jakeouellette: Updated this one to be just toByteArray!

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataTransportState.java
```diff
@@ -0,0 +1,46 @@
...
+    if (!dataTransportState) {
+      return NONE;
+    }
+    if (!dataTransportNativeState) {
+      return JAVA_ONLY;
+    }
+    return ALL;
+  }
+
+  static DataTransportState getState(AppSettingsData appSettingsData) {
```
mrwillis21: Add a test here?

jakeouellette:
last 2 commits LGTM

PR #1358, opened 2020-03-17
Add session end fields to Crashlytics reports

mrwillis21:


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistenceTest.java
```diff
@@ -199,7 +232,10 @@ public void testFinalizeReports_capsReports() throws IOException {
     for (int i = 0; i < 10; i++) {
       persistReportWithEvent(reportPersistence, "testSession" + i, true);
     }
-    reportPersistence.finalizeReports("skippedSession");
+
+    final long endedAt = System.currentTimeMillis();
+
+    reportPersistence.finalizeReports("skippedSession", endedAt);
```
jakeouellette: if it is not relevant to the test, I'm happy with inlining, e.g., 0L for all of these

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -192,17 +192,20 @@ public void finalizeReports(String currentSessionId) {
...
             TRANSFORM.reportFromJson(readTextFile(new File(sessionDirectory, REPORT_FILE_NAME)));
-        final String sessionId = report.getSession().getIdentifier();
 
-        if (userId != null) {
-          report = report.withUserId(userId);
-        }
+        // TODO: Handle null case
+
+        if (report != null) {
+          report = report.withSessionEndFields(sessionEndTime, isHighPriorityReport, userId);
```
jakeouellette: this finalizeReports call is pretty nested. Suggestion:

Refactor to make it so this method is much shorter, like:
```
writeReport(...);
recursiveDelete(sessionDirectory);
```
Then in writeReport(...), flip the two nested ifs into early terminating statements:

```
if (eventFiles.isEmpty()) {
  return;
}
...
if (report == null) {
  // TODO: Handle null case
  return
}```

File: firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java [hunk 3]
```diff
@@ -860,7 +860,7 @@ private void doCloseSessions(int maxCustomExceptionEvents, boolean includeCurren
 
     closeOpenSessions(sessionBeginFiles, offset, maxCustomExceptionEvents);
 
-    reportingCoordinator.finalizeSessions();
+    reportingCoordinator.finalizeSessions(new Date().getTime() / 1000);
```
jakeouellette: This didn't break any CrashlyticsController tests. Do we not mock reportingCoordinator in those tests to check it was called in unit?

... one pattern I often see (and implement) is a TimeProvider interface that would do new Date().getTime() / 1000, what's nice about that pattern is then you could pass it in and mock it to make the tests not break (if they had)

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistenceTest.java
mrwillis21: 👍

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
mrwillis21: I considered that pattern, and in fact we actually have a `SystemCurrentTimeProvider`, but I wasn't sure whether to pass it in here or into the constructor.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: 👍

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
jakeouellette: Cool, there's probably a wider discussion about that which I'm happy to punt on, but we can chat about when you're interested.

Given that this change didn't break any tests, do you need to add any tests that validate this method?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -270,6 +234,61 @@ private void capFinalizedReports() {
...
+    final List<Event> events = new ArrayList<>();
+    boolean isHighPriorityReport = false;
+    for (File eventFile : eventFiles) {
+      final String eventJson = readTextFile(eventFile);
+      if (eventJson == null) {
+        // TODO: Handle null case
+        continue;
+      }
+      final Event event = TRANSFORM.eventFromJson(eventJson);
+      isHighPriorityReport = isHighPriorityReport || isHighPriorityEventFile(eventFile.getName());
```
jakeouellette:
```suggestion
      isHighPriorityReport |= isHighPriorityEventFile(eventFile.getName());
```

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -270,6 +234,61 @@ private void capFinalizedReports() {
...
+  private void synthesizeReportFile(File sessionDirectory, long sessionEndTime) {
+    final List<File> eventFiles =
+        getFilesInDirectory(sessionDirectory, (f, name) -> name.startsWith(EVENT_FILE_NAME_PREFIX));
+
+    // Only process the session if it has associated events
+    if (eventFiles.isEmpty()) {
+      return;
+    }
+
+    // TODO: Handle all nullable return values
```
jakeouellette: which values are the `return values` in this comment?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -270,6 +234,61 @@ private void capFinalizedReports() {
         getAllFilesInDirectory(priorityReportsDirectory), getAllFilesInDirectory(reportsDirectory));
   }
 
+  private File getSessionDirectoryById(String sessionId) {
+    return new File(openSessionsDirectory, sessionId);
+  }
+
+  private void synthesizeReportFile(File sessionDirectory, long sessionEndTime) {
+    final List<File> eventFiles =
+        getFilesInDirectory(sessionDirectory, (f, name) -> name.startsWith(EVENT_FILE_NAME_PREFIX));
```
jakeouellette: is it worth pulling `(f, name) -> name.startsWith(EVENT_FILE_NAME_PREFIX)` into a static constant?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
jakeouellette: nice!!

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrwillis21: Personally I find this less clear, and would prefer we keep it as is if it's all the same to you.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
mrwillis21: I think we can do a separate pass and add tests for validating all of the `SessionReportingCoordinator` calls in `CrashlyticsController`.

PR #1359, opened 2020-03-17
Update minimum required fiid version for Remote Config

clp93:
Bump Remote Config's IID dependency to what should be the next version of IID that's going out.

### DIFF
--- a/firebase-config/firebase-config.gradle
+++ b/firebase-config/firebase-config.gradle
@@ -85,7 +85,7 @@ dependencies {
     implementation project(':firebase-common')
     implementation project (':firebase-abt')
     implementation project(':firebase-components')
-    implementation ('com.google.firebase:firebase-iid:20.0.1') {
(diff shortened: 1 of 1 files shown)


firebase-config/firebase-config.gradle
```diff
@@ -85,7 +85,7 @@ dependencies {
     implementation project(':firebase-common')
     implementation project (':firebase-abt')
     implementation project(':firebase-components')
-    implementation ('com.google.firebase:firebase-iid:20.0.1') {
+    implementation ('com.google.firebase:firebase-iid:20.1.3') {
```
ankitaj224: 20.1.4

clp93:
Hey @ankitaj224 : I don't see that version on Maven https://mvnrepository.com/artifact/com.google.firebase/firebase-iid.

Also @danasilver will be driving this change :)

PR #1367, opened 2020-03-18
Add support for -DexclusiveSpecTest=<TestName> to SpecTestCase.

dconeybe:


firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -132,6 +133,14 @@
   // this tag and they'll all be run (but all others won't).
   private static final String EXCLUSIVE_TAG = "exclusive";
 
+  // The name of a Java system property ({@link System#getProperty(String)}) whose value is the name
+  // of the sole spec test to execute. This is an alternative to setting the {@link #EXCLUSIVE_TAG}
+  // tag, which requires modifying the JSON file. To use this property, specify
+  // -DexclusiveSpecTest=<TestName> to the Java runtime, replacing <TestName> with the name of the
+  // test to execute exclusively.  The <TestName> value is the result of appending the "itName" of
+  // the test to its "describeName", separated by a space character.
+  private static final String EXCLUSIVE_PROPERTY = "exclusiveSpecTest";
```
schmidt-sebastian: What do you think of making this a filter and allowing Regular Expressions? It should maybe add one or two lines to this PR, but make this feature much easier to use (we don't have to specify the full test name) and more flexible (we could run a group of tests).

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -1097,6 +1106,12 @@ public void testSpecTests() throws Exception {
       parsedSpecFiles.add(new Pair<>(f.getName(), fileJSON));
     }
 
+    String exclusiveTestNameFromSystemProperty =
+        emptyToNull(System.getProperty(EXCLUSIVE_PROPERTY));
```
schmidt-sebastian: If you go with regular expression approach, this could be `System.getProperty(EXCLUSIVE_PROPERTY, ".*")`, in which case you can always apply the filter.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -14,6 +14,7 @@
 
 package com.google.firebase.firestore.spec;
 
+import static com.google.common.base.Strings.emptyToNull;
```
schmidt-sebastian: We should try no to put in more Guava dependencies.

dconeybe: In my opinion, and exception for tests makes a lot of sense since there are no downstream dependencies on the test artifacts that could result in dependency mismatch conflicts.  Moreover, this class already imports `com.google.common.collect.Sets` from guava, so this is not a net new dependency on guava.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
dconeybe: SGTM.  I'll make this change.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
wu-hui: I am OK to import Guava in tests. 

And thanks for adding this, will be really handy to try to debug a specific test.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -1115,7 +1137,19 @@ public void testSpecTests() throws Exception {
...
-        boolean runTest = shouldRunTest(tags) && (!exclusiveMode || tags.contains(EXCLUSIVE_TAG));
+        boolean runTest;
+        if (!shouldRunTest(tags)) {
+          runTest = false;
+        } else if (!exclusiveMode) {
+          runTest = true;
+        } else if (tags.contains(EXCLUSIVE_TAG)) {
+          runTest = true;
+        } else if (testNameFilter != null) {
+          runTest = testNameFilter.matcher(name).find();
```
schmidt-sebastian: Optional: This could just be 


``
} else {
 runTest = testNameMatcher.matches(find);
}
```

File: firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java [hunk 5]
```diff
@@ -1097,6 +1109,16 @@ public void testSpecTests() throws Exception {
       parsedSpecFiles.add(new Pair<>(f.getName(), fileJSON));
     }
 
+    String testNameFilterFromSystemProperty =
+        emptyToNull(System.getProperty(TEST_FILTER_PROPERTY));
+    Pattern testNameFilter;
```
schmidt-sebastian: You should turn this into a Matcher right away, like such:

``` 
 if (testNameFilterFromSystemProperty == null) {
      testNameFilterFromSystemProperty = ".*";
 } else {
      exclusiveMode = true;
 }
 Matcher testNameMatcher = Pattern.compile(testNameFilterFromSystemProperty).matcher();
```

dconeybe: A Matcher object must be created with a string against which to match.  So we must create a new Matcher object for each test name.
https://developer.android.com/reference/java/util/regex/Pattern#matcher(java.lang.CharSequence)

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
dconeybe: This won't work though because I need to be able to distinguish between the property being set or not being set.  The reason is that if it is set then `exclusiveMode` must be set to `true` and if it is not then it must be left alone.  By using a default value I cannot know if .* was specified explicitly or was the default value.

dconeybe:
Thank you for the review, @schmidt-sebastian!

PR #1369, opened 2020-03-18
fix(iam): check if message body is not empty string

### LINKED ISSUE #1362: App crashes in In-App Messaging card type with empty message body

  * Android Studio version: 3.5.3
  * Firebase Component: In-App Messaging
  * Component version: 19.0.2

#### Steps to reproduce:

#### Relevant Code:

```
1. Compose a campaign with Card Type layout
2. Set optional message body to empty
3. App crashes
```

rosariopfernandes:
Should fix #1362 

In proto3, if a String value has not been set, it will default to empty String. (See https://developers.google.com/protocol-buffers/docs/proto3#default)

Unfortunately, proto3 doesn't allow us to explicitly specify a default value. To work around that, we could check if the body is an empty String when decoding the message.


JasonAHeron:
See comments on issue #1362

Thank you

PR #1371, opened 2020-03-19
Update release note for 21.4.2

wu-hui:


firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-# Unreleased 
+# Unreleased
+
+# 21.4.2
+- [fixed] Remove Guava dependency from the SDK. This should eliminate crashes caused
+  by user App depending on a wrong flavor of Guava (#1125).
```
schmidt-sebastian: I thought GRPC still depended on Guava and the problem still exists, just at a different layer?

wu-hui: Changed to a better description. Rachel wants us to provide more information than `internal improvements`.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 # Unreleased
 
 # 21.4.2
-- [fixed] Remove Guava dependency from the SDK. This should eliminate crashes caused
-  by user App depending on a wrong flavor of Guava (#1125).
+- [changed] Removed Guava dependency from the SDK. This change is the first
```
wilhuff: The "first step" in here makes it seem like there will be an active project to remove Guava from gRPC, but that's essentially a nonstarter, isn't it?

PR #1377, opened 2020-03-20
Improve coverage report format in GitHub pull requests.

yifanyang:
Calculate and upload coverage reports to our own Metrics Service instead
of Codecov.


vkryachko:
Could the parsing logic be done in `fireci`?

The motivation for `fireci` was to:
keep any ci related postprocessing logic separate from the gradle build with the intent to avoid dependency creep in the main build which, in addition to build slowdown, can cause classpath issues.

yifanyang:
I'm not sure I understand. What do you mean by doing it in `fireci`?

vkryachko:
You can extend https://github.com/firebase/firebase-android-sdk/tree/master/ci/fireci with custom commands, via `fireciplugins` directory.

So in your case you could invoke gradle and then do postprocessing on the expected output of the build.

ci/fireci/fireciplugins/coverage.py
```diff
@@ -0,0 +1,122 @@
...
+
+  print('Posting to the metrics service ...')
+  print(f'Request endpoint: {endpoint}')
+  print(f'Request data: {data}')
+  result = requests.post(f'{metrics_service_url}{endpoint}', data=data, headers=headers)
+  print(f'Response: {result.text}')
+
+
+def construct_request_header():
+  result = subprocess.run(['gcloud', 'auth', 'print-identity-token'], stdout=subprocess.PIPE)
```
vkryachko: Pls add:
```suggestion
  result = subprocess.run(['gcloud', 'auth', 'print-identity-token'], stdout=subprocess.PIPE, check=True)
```

Otherwise you won't know if gcloud fails

ci/fireci/fireciplugins/coverage.py
```diff
@@ -0,0 +1,122 @@
...
+
+  return f'https://{domain}/view/gcs/{bucket}/{directory}/{path}'
+
+
+def post_request(data):
+  metrics_service_url = os.getenv('METRICS_SERVICE_URL')
+  endpoint = construct_request_endpoint()
+  headers = construct_request_header()
+
+  print('Posting to the metrics service ...')
```
vkryachko: pls use `logging.getLogger('fireci.coverage')` throughout.

ci/fireci/fireciplugins/coverage.py
```diff
@@ -0,0 +1,122 @@
...
+def calculate_coverage(element):
+  counter = element.find('counter[@type="LINE"]')
+  if counter is not None:
+    covered = int(counter.attrib['covered'])
+    missed = int(counter.attrib['missed'])
+    return covered / (covered + missed)
+  return 0
+
+
+def find_prow_job_link():
```
vkryachko: Consider making this a flag and have its `default` point to this function.

ci/fireci/fireciplugins/coverage.py
```diff
@@ -0,0 +1,122 @@
...
+  dir_pre_submit = f'pr-logs/pull/{repo_owner}_{repo_name}/{pull_number}'
+  dir_post_submit = "logs"
+  directory = dir_pre_submit if job_type == 'presubmit' else dir_post_submit
+  path = f'{job_name}/{build_id}'
+
+  return f'https://{domain}/view/gcs/{bucket}/{directory}/{path}'
+
+
+def post_request(data):
+  metrics_service_url = os.getenv('METRICS_SERVICE_URL')
```
vkryachko: consider making this an option with `envvar="METRICS_SERVICE_URL"`

ci/fireci/fireciplugins/coverage.py
vkryachko: additionally, this seems a good candidate for a cli flag as well with a default that makes the subprocess call, that way if one has a token it can be passed in without an additonal call to gcloud

ci/fireci/fireciplugins/coverage.py
yifanyang: By "option", do you mean a `@click.option`?

yifanyang: Oh, I get it.

https://click.palletsprojects.com/en/7.x/arguments/#environment-variables

ci/fireci/fireciplugins/coverage.py
yifanyang: Done.

ci/fireci/fireciplugins/coverage.py
yifanyang: Done.

ci/fireci/fireciplugins/coverage.py
yifanyang: Done.

ci/fireci/fireciplugins/coverage.py
yifanyang: Done.

PR #1387, opened 2020-03-24
Clean up error handling for report persistence

mrwillis21:
Ensure data written to disk is cleaned up properly when
read/write/serialization errors occur.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsUncaughtExceptionHandler.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsUncaughtExceptionHandler.java
@@ -46,7 +46,13 @@ class CrashlyticsUncaughtExceptionHandler implements Thread.UncaughtExceptionHan
   public void uncaughtException(Thread thread, Throwable ex) {
     isHandlingException.set(true);
     try {
-      crashListener.onUncaughtException(settingsDataProvider, thread, ex);
+      if (thread == null) {
+        Logger.getLogger().d("Could not handle uncaught exception; null thread");
+      } else if (ex == null) {
+        Logger.getLogger().d("Could not handle uncaught exception; null throwable");
+      } else {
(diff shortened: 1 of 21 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsUncaughtExceptionHandler.java
```diff
@@ -46,7 +46,13 @@ public CrashlyticsUncaughtExceptionHandler(
   public void uncaughtException(Thread thread, Throwable ex) {
     isHandlingException.set(true);
     try {
-      crashListener.onUncaughtException(settingsDataProvider, thread, ex);
+      if (thread == null) {
+        Logger.getLogger().d("Could not handle uncaught exception; null thread");
```
mrichards: What do you think about making some of this output logged as errors, since they are basically logging failed asserts?

mrwillis21: I think in this particular case that makes sense.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -169,66 +184,52 @@ public void deleteFinalizedReport(String sessionId) {
...
-    writeTextFile(new File(outputDirectory, sessionId), TRANSFORM.reportToJson(report));
-  }
-
-  // TODO: Deal with potential runtime exceptions
-  public void finalizeReports(String currentSessionId, long sessionEndTime) {
-    // TODO: Need to implement procedure to skip finalizing the current session when this is
-    //  called on app start, but keep the current session when called at crash time. Currently
-    //  this only works when called at app start.
-    List<File> sessionDirectories = capAndGetOpenSessions(currentSessionId);
+  /**
```
mrichards: This is a lot cleaner! 😍

PR #1392, opened 2020-03-27
Include branch information in metrics uploading.

yifanyang:
The branch information is used by the binary size dashboard.

This was written before coverage upload in `fireci` is implemented. I'll move this part of the logic to `fireci` as well when I get a chance. Right now, I want to make the data flow for dashboard work first.


buildSrc/src/main/groovy/com/google/firebase/gradle/plugins/measurement/MetricsReportUploader.java
```diff
@@ -16,6 +16,7 @@
 
 import org.gradle.api.Project;
 
+// TODO(yifany): Move to fireci and reuse the metrics uploading code there.
```
vkryachko: thanks

PR #1401, opened 2020-03-30
Update spec tests from changes in the firebase-js-sdk respository

dconeybe:
This change picks up changes to the spec tests, especially https://github.com/firebase/firebase-js-sdk/pull/2790, which renamed the "limboDocs" key in the JSON objects to "activeLimboDocs".

### DIFF
--- a/firebase-firestore/src/test/resources/json/limbo_spec_test.json
+++ b/firebase-firestore/src/test/resources/json/limbo_spec_test.json
@@ -1,20 +1,23 @@
 {
-  "Limbo documents are deleted without an existence filter": {
+  "Document remove message will cause docs to go in limbo": {
     "describeName": "Limbo Documents:",
-    "itName": "Limbo documents are deleted without an existence filter",
-    "tags": [],
+    "itName": "Document remove message will cause docs to go in limbo",
+    "tags": [
+    ],
     "config": {
-      "useGarbageCollection": true,
-      "numClients": 1
+      "numClients": 1,
+      "useGarbageCollection": true
     },
     "steps": [
       {
         "userListen": [
           2,
           {
-            "path": "collection",
-            "filters": [],
-            "orderBys": []
+            "filters": [
+            ],
+            "orderBys": [
+            ],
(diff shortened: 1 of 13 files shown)


firebase-firestore/src/test/resources/json/limbo_spec_test.json
```diff
@@ -3638,15 +3627,21 @@
                   "orderBys": []
                 }
               ],
-              "resumeToken": ""
+              "resumeToken": "resume-token-1000000"
             }
           }
         },
-        "clientIndex": 0
+        "clientIndex": 1
```
wilhuff: It looks like the generator isn't generating a stable order among client indexes?

dconeybe: I'll look into this and reply back.  There must have been a change between now and November 2019 (4 months ago) that affected this as the spec tests have not been regenerated since then.

dconeybe: Update: The culprit is https://github.com/firebase/firebase-js-sdk/pull/2693, discovered by performing a bisection.  I'm continuing to investigate.

dconeybe: It looks like the change in the spec test is expected; however, in the culprit PR the spec tests for Android (and probably iOS) were not updated accordingly.  I will create a separate PR to bring the spec tests up-to-date from the PR so that this PR can encompass only a single update to the spec tests.

dconeybe: I've created a PR, https://github.com/firebase/firebase-js-sdk/pull/2843, to produce stable JSON from `generate_spec_json.js`.  I will hold off merging this PR until that PR is resolved so that a commit with a noisy diff in the spec tests can be limited to one commit.

dconeybe: Update: commit 19ab883 was added to this PR with the stable JSON.

PR #1404, opened 2020-03-31
Pre-gzip the crashlytics payloads before sending them up

jakeouellette:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/BytesBackedNativeSessionFile.java
```diff
@@ -53,12 +56,24 @@ public InputStream getStream() {
     return isEmpty()
         ? null
         : CrashlyticsReport.FilesPayload.File.builder()
-            .setContents(bytes)
+            .setContents(asGzippedBytes())
```
mrwillis21, line 59: This now becomes possibly nullable, where it was not before. Was that intended?

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/TestUtils.java
```diff
@@ -33,4 +35,15 @@ public static void writeStringToFile(String s, File f) throws IOException {
       }
     }
   }
+
+  public static byte[] gzipToBytes(byte[] compressed, int numBytes) {
```
mrwillis21, line 39: Can we call this `gzipToRawBytes`, or `inflateGzip` or somesuch?

mrwillis21: Also, using a `ByteArrayOutputStream` would allow you to not have to pass in the number of bytes of the uncompressed output.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/BytesBackedNativeSessionFile.java
```diff
@@ -53,12 +56,24 @@ public InputStream getStream() {
...
             .build();
   }
 
   private boolean isEmpty() {
     return bytes == null || bytes.length == 0;
   }
+
+  @Nullable
+  private byte[] asGzippedBytes() {
+    try (ByteArrayOutputStream bos = new ByteArrayOutputStream();
```
mrwillis21, line 70: Is there value in using the same `gzip` function in both classes, so as to be assured that they, in fact, work the same way?

mrwillis21: Talked about this offline. Decided it's fine.

PR #1415, opened 2020-04-01
DeferredValue changes for ServerValue.increment()

schmidt-sebastian:
Ports https://github.com/firebase/firebase-ios-sdk/pull/5149/commits/aa67e19051a9b92d6fca709e709d3ac58beb00ed


firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -99,32 +100,29 @@ static Object resolveComplexDeferredValue(
...
+        resolveDeferredLeafValue(
+            rawPriority,
+            existing.getImmediateChild(ChildKey.fromString(".priority")),
+            serverValues);
     if (data.isLeafNode()) {
-      Object value = resolveDeferredValue(data.getValue(), existing, serverValues);
-      if (!value.equals(data.getValue()) || !priority.equals(data.getPriority())) {
-        return NodeUtilities.NodeFromJSON(value, priority);
+      Object value = resolveDeferredLeafValue(data.getValue(), existing, serverValues);
+      if (!value.equals(data.getValue()) || !Utilities.equals(priority, rawPriority)) {
```
inlined, line 124: Oof... I hate files and classes called `utilities`

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
```diff
@@ -145,22 +143,22 @@ public void visitChild(ChildKey name, Node child) {
...
       } else {
         return holder.getRootNode();
       }
     }
   }
 
   public static CompoundWrite resolveDeferredValueMerge(
-      CompoundWrite merge, Node existing, final Map<String, Object> serverValues) {
+      CompoundWrite merge, SyncTree syncTree, Path path, final Map<String, Object> serverValues) {
     CompoundWrite write = CompoundWrite.emptyWrite();
```
inlined, line 155: Interesting. In iOS this starts as merge and is modified with addWrite only where the resolved value changes. Is this a pessimisation?

firebase-database/src/main/java/com/google/firebase/database/core/SyncTree.java
```diff
@@ -249,12 +249,12 @@ public boolean isEmpty() {
                 if (write.isOverwrite()) {
                   Node resolvedNode =
                       ServerValues.resolveDeferredValueSnapshot(
-                          write.getOverwrite(), existing, serverValues);
```
inlined, line 252: I think you can delete "existing" above.

firebase-database/src/androidTest/java/com/google/firebase/database/DataTest.java
```diff
@@ -2747,7 +2747,18 @@ public void onComplete(DatabaseError error, boolean committed, DataSnapshot curr
   }
 
   @Test
-  public void testServerIncrementOverwritesExistingData()
+  public void testServerIncrementOverwritesExistingDataOnline()
+      throws DatabaseException, TimeoutException, InterruptedException {
+    serverIncrementOverwritesExistingData(true);
```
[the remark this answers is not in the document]
schmidt-sebastian: Done. Android Studio does this already implicitly, but it still makes sense to do until GitHub catches up.

firebase-database/src/main/java/com/google/firebase/database/core/DeferredValueProvider.java
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+package com.google.firebase.database.core;
+
+import com.google.firebase.database.snapshot.ChildKey;
+import com.google.firebase.database.snapshot.Node;
+import java.util.ArrayList;
+
+/** A DeferredValueProvider computes the value of a Node only when {@link #node()} is invoked. */
+public class DeferredValueProvider implements ValueProvider {
```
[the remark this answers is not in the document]
schmidt-sebastian: It does, but I think it is still a good idea - especially since these classes are so small.

firebase-database/src/main/java/com/google/firebase/database/core/SyncTree.java
schmidt-sebastian: Good catch. This also removed:

```
ArrayList<Long> excludeThis = new ArrayList<>();
excludeThis.add(write.getWriteId());
calcCompleteEventCache(write.getPath(), excludeThis);
```

That's a bit suspicious, but this never existing on iOS.

firebase-database/src/main/java/com/google/firebase/database/core/ServerValues.java
schmidt-sebastian: I tried to make it match iOS just now but it broke a bunch of tests.

PR #1421, opened 2020-04-02
Revert "Remove organization ID from DataTransport payloads"

mrwillis21:
Reverts firebase/firebase-android-sdk#1411

Found an issue with the backend, rolling this back until that's sorted.


mrwillis21:
Backend issue was resolved before it became necessary to merge this.

PR #1424, opened 2020-04-03
Firebase Data Collection:- Add tri states to the data collection api

VinayGuthal:
Add tri state for the dataCollectionsetter in firebase-common

### DIFF
--- a/firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
+++ b/firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
@@ -16,8 +16,6 @@ package com.google.firebase;
 
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.os.Build;
-import androidx.core.content.ContextCompat;
 import androidx.test.core.app.ApplicationProvider;
 import com.google.firebase.internal.DataCollectionConfigStorage;
 import java.util.function.Consumer;
@@ -47,15 +45,8 @@ final class DataCollectionTestUtil {
     }
   }
 
-  private static Context directBootSafe(Context applicationContext) {
-    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
-      return applicationContext;
-    }
-    return ContextCompat.createDeviceProtectedStorageContext(applicationContext);
-  }
-
   static SharedPreferences getSharedPreferences() {
-    return directBootSafe(ApplicationProvider.getApplicationContext())
+    return ApplicationProvider.getApplicationContext()
         .getSharedPreferences(
             FIREBASE_APP_PREFS + FirebaseApp.getPersistenceKey(APP_NAME, OPTIONS),
             Context.MODE_PRIVATE);
--- a/firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
+++ b/firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
@@ -52,10 +52,17 @@ final class DataCollectionTestUtil {
             Context.MODE_PRIVATE);
   }
 
-  static void setSharedPreferencesTo(boolean enabled) {
-    getSharedPreferences()
-        .edit()
-        .putBoolean(DataCollectionConfigStorage.DATA_COLLECTION_DEFAULT_ENABLED, enabled)
-        .commit();
+  static void setSharedPreferencesTo(Boolean enabled) {
+    if (enabled != null) {
+      getSharedPreferences()
+          .edit()
+          .putBoolean(DataCollectionConfigStorage.DATA_COLLECTION_DEFAULT_ENABLED, enabled)
+          .commit();
+    } else {
+      getSharedPreferences()
+          .edit()
+          .remove(DataCollectionConfigStorage.DATA_COLLECTION_DEFAULT_ENABLED)
+          .apply();
+    }
   }
 }
--- a/firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
+++ b/firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
@@ -45,11 +45,31 @@ final class DataCollectionTestUtil {
     }
   }
 
+  static SharedPreferences getSharedPreferences(Context context) {
+    return context.getSharedPreferences(
+        FIREBASE_APP_PREFS + FirebaseApp.getPersistenceKey(APP_NAME, OPTIONS),
+        Context.MODE_PRIVATE);
+  }
+
   static SharedPreferences getSharedPreferences() {
     return ApplicationProvider.getApplicationContext()
         .getSharedPreferences(
             FIREBASE_APP_PREFS + FirebaseApp.getPersistenceKey(APP_NAME, OPTIONS),
             Context.MODE_PRIVATE);
+  }
+
+  static void setSharedPreferencesTo(Context context, Boolean enabled) {
+    if (enabled != null) {
+      getSharedPreferences(context)
+          .edit()
+          .putBoolean(DataCollectionConfigStorage.DATA_COLLECTION_DEFAULT_ENABLED, enabled)
+          .commit();
+    } else {
+      getSharedPreferences(context)
+          .edit()
+          .remove(DataCollectionConfigStorage.DATA_COLLECTION_DEFAULT_ENABLED)
+          .apply();
+    }
   }
 
   static void setSharedPreferencesTo(Boolean enabled) {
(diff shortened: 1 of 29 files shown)


firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
```diff
@@ -37,7 +36,7 @@
   private final Context applicationContext;
   private final SharedPreferences sharedPreferences;
   private final Publisher publisher;
-  private final AtomicBoolean dataCollectionDefaultEnabled;
+  private Boolean dataCollectionDefaultEnabled;
```
vkryachko, line 39: Why is it no longer atomic? do we not care about concurrency anymore?

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
```diff
@@ -61,19 +61,28 @@ public boolean isEnabled() {
...
+      if (dataCollectionDefaultEnabled.compareAndSet(!manifestSetting, manifestSetting)) {
+        publisher.publish(
+            new Event<>(
+                DataCollectionDefaultChange.class,
+                new DataCollectionDefaultChange(manifestSetting)));
+      }
+    } else {
+      boolean apiSetting = Boolean.TRUE.equals(enabled);
+      sharedPreferences.edit().putBoolean(DATA_COLLECTION_DEFAULT_ENABLED, apiSetting).apply();
+      if (dataCollectionDefaultEnabled.compareAndSet(!apiSetting, apiSetting)) {
```
ashwinraghav, line 77: Can be extracted into a function?

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
```diff
@@ -61,19 +61,28 @@ public boolean isEnabled() {
...
 
-  public void setEnabled(boolean enabled) {
-    if (dataCollectionDefaultEnabled.compareAndSet(!enabled, enabled)) {
-      sharedPreferences.edit().putBoolean(DATA_COLLECTION_DEFAULT_ENABLED, enabled).apply();
-
-      publisher.publish(
-          new Event<>(DataCollectionDefaultChange.class, new DataCollectionDefaultChange(enabled)));
+  public void setEnabled(Boolean enabled) {
+    if (enabled == null) {
+      sharedPreferences.edit().remove(DATA_COLLECTION_DEFAULT_ENABLED).apply();
```
ashwinraghav, line 66: Should this also live inside the synchronized block?

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
```diff
@@ -61,19 +61,28 @@ public boolean isEnabled() {
...
+      boolean manifestSetting = readManifestDataCollectionEnabled();
+      if (dataCollectionDefaultEnabled.compareAndSet(!manifestSetting, manifestSetting)) {
+        publisher.publish(
+            new Event<>(
+                DataCollectionDefaultChange.class,
+                new DataCollectionDefaultChange(manifestSetting)));
+      }
+    } else {
+      boolean apiSetting = Boolean.TRUE.equals(enabled);
```
ashwinraghav, line 75: Synchronized block?

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
VinayGuthal: I think shared preferences itself makes sure that it's synchornized

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
VinayGuthal: I think shared preferences itself makes sure that it's synchornized

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
VinayGuthal: The publisher part?

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
ashwinraghav: Sure, you may however run into a situation where one thread sets the shared prefs value and another one broadcasts the message. So they need to be synchronized around the same mutex

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
ashwinraghav: I mean Lines 69-73 and lines 77-82 seem extractable into a reusable function

VinayGuthal: yeah makes sense i will do that

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
VinayGuthal: I dont think a synchronized block is needed in this case. Since the method is doing only one write and no read from the shared prefs. And since readManifestDataCollectionEnabled always returns the same value based on the manifest file

vkryachko: You *do* need to synchronize here, before your change synchronization was happening with the atomic bool, but since you are checking for null now, a bunch of races become possible, i.e. concurrent threads can enter different branches of this if block and cause mess up the resulting state.

VinayGuthal: actually nvm the setEnabled method should be in a sync block..
setEnabled(true) -> Thread1
setEnabled(false) -> Thread2
This might result in the shared preferences having the value of false but the dataCollection having the value of True based on the order of execution

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
```diff
@@ -43,15 +43,14 @@ public DataCollectionConfigStorage(
...
     this.applicationContext = directBootSafe(applicationContext);
     this.sharedPreferences =
-        applicationContext.getSharedPreferences(
+        this.applicationContext.getSharedPreferences(
             FIREBASE_APP_PREFS + persistenceKey, Context.MODE_PRIVATE);
     this.publisher = publisher;
     this.dataCollectionDefaultEnabled = new AtomicBoolean(readAutoDataCollectionEnabled());
   }
 
   private static Context directBootSafe(Context applicationContext) {
```
vkryachko, line 52: You may want to rename this to `deviceProtectedContext` since you changed the logic inside

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
```diff
@@ -61,19 +60,26 @@ public boolean isEnabled() {
     return dataCollectionDefaultEnabled.get();
   }
 
-  public void setEnabled(boolean enabled) {
+  private synchronized void updateDataCollectionDefaultEnabled(boolean enabled) {
     if (dataCollectionDefaultEnabled.compareAndSet(!enabled, enabled)) {
```
vkryachko, line 64: Since the method is now synchronized there is no point for the bool to be atomic. Also very important to make isEnabled() synchronized as well.

firebase-common/data-collection-tests/src/test/java/com/google/firebase/DataCollectionTestUtil.java
```diff
@@ -52,10 +52,17 @@ static SharedPreferences getSharedPreferences() {
             Context.MODE_PRIVATE);
```
vkryachko, line 52: How does this test work if it's using user credential encrypted context?(i.e. not the direct boot one?)

VinayGuthal: Its because of the config sdk=19

firebase-common/src/main/java/com/google/firebase/internal/DataCollectionConfigStorage.java
vkryachko: Actually I think this method is even redundant and you can just use `ContextCompat.createDeviceProtectedStorageContext()` directly as it does the same version check.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/CrashlyticsControllerTest.java
```diff
@@ -842,9 +842,8 @@ public void testUploadDisabledThenEnabled() throws Exception {
     Task<Void> task = controller.submitAllReports(1.0f, testSettingsDataProvider.getAppSettings());
 
     arbiter.setCrashlyticsDataCollectionEnabled(true);
-
+    
     await(task);
-
```
vkryachko, line 847: pls revert this change, lgtm otherwise

PR #1431, opened 2020-04-07
Add large payload support to datatransport.

vkryachko:
Large payloads are stored in a separate table as 80KB chunks of data keyed
by (sequence_num,event_id).
The motivation for the change is to avoid the Android SQLite
implementation limitation, that uses `CursorWindow`s of a limited size
that is unknown at runtime and vendors can change arbitrarily, hence the
conservative choice of 80KB instead of 2048KB that is present in AOSP
source code.

The way it manifests is that it is possible to write rows in the the db
that is impossible to read back as an exception gets thrown.

An altertative approach that was considered was to store file paths in
the DB and store blobs in the file-system. But that would introduce more
complexity to the storage system:

* Loss of ACID semantics, i.e. if we fail to commit we may fail to
  delete the underlying file.
* Should we store it in cache dir or in file dir?
* What if cache is cleared but the event reference still exists in the db?
* etc.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
@@ -13,7 +13,6 @@
 // limitations under the License.
 package com.google.android.datatransport.runtime.scheduling.persistence;
 
-import android.content.ContentValues;
 import android.content.Context;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteOpenHelper;
@@ -106,10 +105,7 @@ final class SchemaManager extends SQLiteOpenHelper {
 
   private static final SchemaManager.Migration MIGRATE_TO_V4 =
       db -> {
-        db.execSQL("ALTER TABLE events ADD COLUMN inline BOOLEAN DEFAULT 1");
-        ContentValues values = new ContentValues();
-        values.put("inline", 1);
-        db.update("events", values, null, new String[0]);
+        db.execSQL("ALTER TABLE events ADD COLUMN inline BOOLEAN NOT NULL DEFAULT 1");
         db.execSQL(CREATE_PAYLOADS_TABLE_V4);
       };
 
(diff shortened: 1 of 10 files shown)


vkryachko:
@mrwillis21 fyi, still a work in progress but feedback is welcome

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -95,12 +104,24 @@
   private static final SchemaManager.Migration MIGRATE_TO_V3 =
       db -> db.execSQL("ALTER TABLE events ADD COLUMN payload_encoding TEXT");
 
+  private static final SchemaManager.Migration MIGRATE_TO_V4 =
+      db -> {
+        db.execSQL("ALTER TABLE events ADD COLUMN inline BOOLEAN DEFAULT 1");
```
ashwinraghav, line 109: What happens to existing inline events that may be too large. Should we cleanup since migration seems impossible ?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
```diff
@@ -95,12 +104,24 @@
...
+        ContentValues values = new ContentValues();
+        values.put("inline", 1);
+        db.update("events", values, null, new String[0]);
```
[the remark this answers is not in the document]
vkryachko: I thought I needed to backfill values, but it turns out I did not, updated the migration test to verify that it's backfilled.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManager.java
vkryachko: I think it's highly unlikely that we have such events out there and the cleanup would be pretty involved if we don't want to lose data, we would have to:

* query all events with len(payload) > X
* for each event select X bytes chunk by chunk and put into new table
* clear inline + payload fields

And all of this would have to be maintained forever.

So given how unlikely it is that we have large payloads out there, I'd prefer not to bother with this. wdyt?

vkryachko:
**Benchmark results with `androidx.benchmark`**

The test was storing 1MB of payload into differently configured stores:

* **LargeInlineBlobs**: maxBlobSize = 2MB
* **SmallInlineBlobs**: maxBlobSize = 80KB

**Results**
StoreConfig | Operation | Time
--- | --- | ---:
LargeInlineBlobs | persist | 73,020ns
SmallInlineBlobs | persist | 71,753ns
LargeInlineBlobs | load | 1,978,386ns
SmallInlineBlobs | load | 2,375,521ns

PR #1434, opened 2020-04-07
Update crashlytics global data collection

VinayGuthal:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
```diff
@@ -107,13 +116,29 @@ public boolean isAutomaticDataCollectionEnabled() {
   }
 
   @SuppressLint({"CommitPrefEdits", "ApplySharedPref"})
-  public void setCrashlyticsDataCollectionEnabled(boolean enabled) {
-    crashlyticsDataCollectionEnabled = enabled;
-    crashlyticsDataCollectionExplicitlySet = true;
-    sharedPreferences.edit().putBoolean(FIREBASE_CRASHLYTICS_COLLECTION_ENABLED, enabled).commit();
+  public synchronized void setCrashlyticsDataCollectionEnabled(Boolean enabled) {
```
ashwinraghav, line 119: We can help our future selves by adding some comments here :)

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/CrashlyticsControllerTest.java
```diff
@@ -842,6 +842,21 @@ public void testUploadDisabledThenEnabled() throws Exception {
...
 
     arbiter.setCrashlyticsDataCollectionEnabled(true);
+    assertTrue(arbiter.isAutomaticDataCollectionEnabled());
+
+    when(mockEditor.putBoolean(PREFS_KEY, false)).thenReturn(mockEditor);
+    when(mockPrefs.getBoolean(PREFS_KEY, true)).thenReturn(false);
+    arbiter.setCrashlyticsDataCollectionEnabled(false);
+    assertFalse(arbiter.isAutomaticDataCollectionEnabled());
+
+    when(mockPrefs.contains(PREFS_KEY)).thenReturn(false);
```
ashwinraghav, line 852: Can you split this up into separate tests that express the conditions they check?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
mrwillis21, line 119: What is this method synchronizing that was missed before?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
```diff
@@ -107,13 +116,29 @@ public boolean isAutomaticDataCollectionEnabled() {
...
+    if (enabled == null) {
```
mrwillis21, line 120: Question: this setting, conceptually, should be "clear the shared preferences setting, then reset the values based on the manifest". Is there any way we could consolidate that second part with the same conceptual code (i.e. setting the values based on the manifest) in the constructor, and make a method out of it? That would 1) make this method simpler and more compact, and 2) make sure that both times (on construction and also here) we set the values the same way.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
VinayGuthal: Yes looks like it was missed before.

mrwillis21: My question was more the *what* is being synchronized, rather than "was this missed". I'm not 100% sure this needs to be synchronized, but if some state in here does need to be, we should probably be more explicit about it and try to use a synchronized block with a named lock for clarity.

mrwillis21: Talked offline about this: going to leave as it is for now and potentially polish it in a subsequent commit.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -463,4 +463,23 @@ public boolean didCrashOnPreviousExecution() {
   public void setCrashlyticsCollectionEnabled(boolean enabled) {
     core.setCrashlyticsCollectionEnabled(enabled);
   }
+
+  /**
+   * Enables/disables/clears automatic data collection config by Crashlytics.
```
RelaxedPear, line 468: Suggest replacing line 468 with:

`* Allows you to enable or disable the automatic data collection configuration in Crashlytics.`

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -463,4 +463,23 @@ public boolean didCrashOnPreviousExecution() {
...
+   *
+   * <p>If this is set, it overrides the data collection settings provided by the Android Manifest,
```
RelaxedPear, line 470: Suggest replacing lines 470 and 471 with:

"* If not set, defaults to `True`. When set to `False`, disables automatic collection by overriding the data collections settings in the AndroidManifest.xml, as well as any Firebase-wide automatic data collection settings. When set to `null`, clears the Crashlytics data collection configuration (in this case, enabling data collection becomes dependent on other settings)."

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -463,4 +463,23 @@ public boolean didCrashOnPreviousExecution() {
...
+   * as well as any Firebase-wide automatic data collection settings.
+   *
+   * <p>If automatic data collection is disabled for Crashlytics, crash reports are stored on the
+   * device. To check for reports, use the {@link #checkForUnsentReports()} method. Use {@link
+   * #sendUnsentReports()} to upload existing reports even when automatic data collection is
+   * disabled. Use {@link #deleteUnsentReports()} to delete any reports stored on the device without
+   * sending them to Crashlytics.
+   *
+   * @param enabled whether to enable automatic data collection. The value of null would clear the
```
RelaxedPear, line 479: Suggest replacing line 479 with:

"`@param` Determines whether to enable or disable automatic data collection. If set to `null`, clears the Crashlytics data collection configuration."

PR #1438, opened 2020-04-08
Remove a hard assert that no longer holds due to mirror queries.

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -220,9 +220,6 @@ private ViewSnapshot initializeViewAndComputeSnapshot(Query query, int targetId)
     View view = new View(query, queryResult.getRemoteKeys());
     View.DocumentChanges viewDocChanges = view.computeDocChanges(queryResult.getDocuments());
     ViewChange viewChange = view.applyChanges(viewDocChanges, synthesizedCurrentChange);
-    hardAssert(
```
schmidt-sebastian, line 223: I believe the reason that this assert was here is that this method does not call `updateTrackedLimboDocuments`. 

So, we either need to convince ourselves that any limbo document encountered here will already be in limbo thanks to the originating query (and document this assumption) or call `updateTrackedLimboDocuments`.

wu-hui: Added a call to updateTrackedLimboDocuments

schmidt-sebastian:
LGTM. Please remember to port to iOS and Web.

PR #1443, opened 2020-04-09
Logging error response from FIS servers

ankitaj224:
Looks like this: https://screenshot.googleplex.com/dEO4gqR3fVx


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -161,6 +164,8 @@ public InstallationResponse createFirebaseInstallation(
           return readCreateResponse(httpURLConnection);
         }
 
+        Log.w(FIS_TAG, readErrorResponse(httpURLConnection));
```
andirayo, line 167: How about creating a method for this to prevent code duplication:

```
private void logFisCommunicationError(HttpURLConnection conn) {
  String logString = [...];
  Log.w(FIS_TAG, logString);
}
```

Please ignore/ack if you think this makes no sense.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -161,6 +164,8 @@ public InstallationResponse createFirebaseInstallation(
...
+
         if (httpResponseCode == 429 || (httpResponseCode >= 500 && httpResponseCode < 600)) {
           retryCount++;
           continue;
```
andirayo, line 171: I would ask you to please add another log message to the case that the request fails as `ResponseCode.BAD_CONFIG`, when we know that the application is misconfigured.
Since this is the main concern we are addressing, I recommend to add another log message, preferably as "Error" or as "Critical" - I'm not sure which one is more accurate:

```
Log.e/c(FIS_TAG, "Firebase Installations can not communicate with Firebase server APIs due to invalid configuration. Please update your Firebase initialization process and set valid Firebase options (API key, Project ID, Application ID) when initializing Firebase."
```

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -486,4 +495,19 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
         ? 0L
         : Long.parseLong(expiresIn.substring(0, expiresIn.length() - 1));
   }
+
+  // Read the error message from the response.
+  private String readErrorResponse(HttpURLConnection conn) throws IOException {
```
andirayo, line 500: If we are only using this for logging, should we swallow that IOException? Wdyt?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -486,4 +495,19 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
...
+    try (BufferedReader reader =
+        new BufferedReader(new InputStreamReader(conn.getErrorStream(), UTF_8))) {
+      StringBuilder response = new StringBuilder();
+      for (String input = reader.readLine(); input != null; input = reader.readLine()) {
+        response.append(input).append('\n');
+      }
+
+      return String.format(
+          "The FIS server responded with an error. HTTP response: [%d %s %s]",
```
andirayo, line 509: **FIS**
"Firebase Installations API" or "Firebase Installations server API" instead of "FIS server".
We don't use FIS in public documentation.
Also it uses the term "Firebase" so developers are more likely to recognize that this is used by their Firebase service.

**Colon**
Please add a colon between ResponseMessage and ResponseBody. I think that is the standard when printing these errors (not 100% sure).

**Suggestion**
```
"Error when communicating with the Firebase Installations server API. HTTP response: [%d %s: %s]"
```

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Done

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Done. PTAL

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Makes sense. Done

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -486,4 +511,27 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
...
+
+  private static void logFisCommunicationError(HttpURLConnection conn) {
+    String logString = readErrorResponse(conn);
+    if (!TextUtils.isEmpty(logString)) {
+      Log.w(FIS_TAG, logString);
+    }
+  }
+
+  // Read the error message from the response.
+  private static String readErrorResponse(HttpURLConnection conn) {
```
andirayo, line 523: @Nullable

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -286,19 +294,23 @@ public void deleteFirebaseInstallation(
...
 
       int httpResponseCode = httpURLConnection.getResponseCode();
-      httpURLConnection.disconnect();
 
       if (httpResponseCode == 200 || httpResponseCode == 401 || httpResponseCode == 404) {
         return;
       }
 
+      logFisCommunicationError(httpURLConnection);
+      httpURLConnection.disconnect();
```
andirayo, line 303: If you return before this httpURLConnection seems to stay connected, because you only call #disconnect over here.

ankitaj224: Good catch. Fixed it.

PR #1457, opened 2020-04-13
Testing CI

ashwinraghav:
Failures that we've faced to get to these fixes
1. [ Robolectric freaks out](https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/1457/check-changed/1249842709921796101): While I am yet to find any documentation that states a lack of support for java 11 on older versions, the error seems pretty [explicit](https://github.com/robolectric/robolectric/blob/72af504d3bae70947ae921225cf32549e76035f1/utils/reflector/src/main/java/org/robolectric/util/reflector/UnsafeAccess.java#L15), and the newer versions of robolectric **require** [java 9 or newer](https://github.com/robolectric/robolectric/releases/tag/robolectric-4.3.1) when targetting newer devices. Seems to have done the trick

2. Mismatch between robolectric versions in java and ktx seemed to be a problem]. See test titled "testUsesTargetMappingToExecuteQueries" [here](https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/1457/check-changed/1249842709921796101).

### DIFF
--- a/firebase-firestore/ktx/ktx.gradle
+++ b/firebase-firestore/ktx/ktx.gradle
@@ -56,7 +56,7 @@ dependencies {
     testImplementation 'com.fasterxml.jackson.core:jackson-databind:2.9.8'
     testImplementation 'com.google.android.gms:play-services-tasks:16.0.1'
     testImplementation 'com.google.protobuf:protobuf-lite:3.0.1'
-    testImplementation "org.robolectric:robolectric:$robolectricVersion"
+    testImplementation "org.robolectric:robolectric:4.3.1"
     testImplementation 'junit:junit:4.12'
     testImplementation "com.google.truth:truth:$googleTruthVersion"
 }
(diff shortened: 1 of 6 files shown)


firebase-firestore/ktx/ktx.gradle
```diff
@@ -56,7 +56,7 @@ dependencies {
     testImplementation 'com.fasterxml.jackson.core:jackson-databind:2.9.8'
     testImplementation 'com.google.android.gms:play-services-tasks:16.0.1'
     testImplementation 'com.google.protobuf:protobuf-lite:3.0.1'
-    testImplementation "org.robolectric:robolectric:$robolectricVersion"
```
vkryachko, line 59: Would it make sense to upgrade robolectricVersion in the root build.gradle file for it to apply to all projects?

PR #1458, opened 2020-04-13
Removing try-with-resources

ankitaj224:
try-with-resources desugaring in google3 is incorrect. This causes copybara import to google3 fail. Raised a bug with Gms core. Meanwhile, removing try-with-resources from FIS SDK.


wilhuff:
Could you explain the reason for this change so that others can learn from it?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -205,13 +205,16 @@ private void writeFIDCreateRequestBodyToOutputStream(
...
+    OutputStream outputStream = urlConnection.getOutputStream();
+    if (outputStream == null) {
+      throw new IOException("Cannot send request to FIS servers. No OutputStream available.");
+    }
+    GZIPOutputStream gzipOutputStream = new GZIPOutputStream(outputStream);
+    try {
+      gzipOutputStream.write(jsonBytes);
+    } finally {
+      gzipOutputStream.close();
+      outputStream.close();
```
andirayo, line 217: 2 Questions:
a) Do we want to catch the IOExceptions here?
Thought process is this:
If an IOException is thrown during closing, do we really want to interrupt the normal process of FIS?
b) If your answer to (a) is "no", could you please add a comment explaining your decision?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -525,8 +528,8 @@ private static void logFisCommunicationError(HttpURLConnection conn) {
   // Read the error message from the response.
   @Nullable
   private static String readErrorResponse(HttpURLConnection conn) {
-    try (BufferedReader reader =
-        new BufferedReader(new InputStreamReader(conn.getErrorStream(), UTF_8))) {
+    BufferedReader reader = new BufferedReader(new InputStreamReader(conn.getErrorStream(), UTF_8));
```
andirayo, line 531: This can throw an IOException, yeah?
Not sure, but I think you should do the following:
```
Buffered reader = null;
try {
 reader = new BufferedReader(new InputStreamReader(conn.getErrorStream(), UTF_8));
```

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -536,6 +539,14 @@ private static String readErrorResponse(HttpURLConnection conn) {
           conn.getResponseCode(), conn.getResponseMessage(), response);
     } catch (IOException ignored) {
       return null;
+    } finally {
+      try {
+        if (reader != null) {
```
andirayo, line 544: I recommend pulling the if-clause around the try-clause.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Caught the exception. But, I hate these checks and code. It was so much cleaner and easy to maintain before.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Buffer reader creation can't. It could throw null pointer exception which is handled. PTAL.

ankitaj224:
> Could you explain the reason for this change so that others can learn from it?

Done.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -205,12 +205,19 @@ private void writeFIDCreateRequestBodyToOutputStream(
...
-      if (outputStream == null) {
-        throw new IOException("Cannot send request to FIS servers. No OutputStream available.");
-      }
-      try (GZIPOutputStream gzipOutputStream = new GZIPOutputStream(outputStream)) {
-        gzipOutputStream.write(jsonBytes);
+    OutputStream outputStream = urlConnection.getOutputStream();
+    if (outputStream == null) {
+      throw new IOException("Cannot send request to FIS servers. No OutputStream available.");
+    }
+    GZIPOutputStream gzipOutputStream = new GZIPOutputStream(outputStream);
```
andirayo, line 212: Probably an unrelated question, feel free to just acknowledge:
Can this ever be NULL, similar to outputStream above?
Probably not, huh?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -525,8 +532,12 @@ private static void logFisCommunicationError(HttpURLConnection conn) {
   // Read the error message from the response.
   @Nullable
   private static String readErrorResponse(HttpURLConnection conn) {
-    try (BufferedReader reader =
-        new BufferedReader(new InputStreamReader(conn.getErrorStream(), UTF_8))) {
+    InputStream errorStream = conn.getErrorStream();
+    if (errorStream == null) {
+      return null;
+    }
```
andirayo, line 538: nice!

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: It cant be null. I validated that. It throw NPE or returns a valid GzipOutputStream.

PR #1463, opened 2020-04-14
add a comment on regex import

ChaoqunCHEN:


firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -15,10 +15,12 @@
 package com.google.firebase.installations;
 
 import android.text.TextUtils;
+
```
andirayo, line 18: please remove empty line

PR #1470, opened 2020-04-15
Added precondition format checks for App-Id and Api-Key

ChaoqunCHEN:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -146,14 +146,22 @@ private void preConditionChecks() {
...
+                    + "Please refer to https://firebase.google.com/support/privacy/init-options.");
   }
 
   /** Returns the Project Id or Project Number for the Firebase Project. */
   @Nullable
   String getProjectIdentifier() {
-    return TextUtils.isEmpty(firebaseApp.getOptions().getProjectId())
-        ? firebaseApp.getOptions().getGcmSenderId()
-        : firebaseApp.getOptions().getProjectId();
+    return firebaseApp.getOptions().getProjectId();
```
andirayo, line 164: Is there no test that you have to modify?
(I mean a test that created FirebaseApp without projectId but with gcmSenderId.)

ChaoqunCHEN: No, we don't have

PR #1474, opened 2020-04-16
Refactoring to help with ANR

ashwinraghav:


firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +244,38 @@ private boolean shouldIgnoreCache(String event) {
...
-                          .doOnError(e -> Logging.logw("Service fetch error: " + e.getMessage()))
-                          .onErrorResumeNext(Maybe.empty()); // Absorb service failures
+              Maybe<InstanceIdResult> getIID = Maybe.fromCallable(() -> {
+                Task<InstanceIdResult> t = firebaseInstanceId.getInstanceId();
+                // This should be safe since these callbacks never happen on the UI thread.
+                Tasks.await(t);
+                if(t.isSuccessful() && t.getResult() != null) {
+                  return t.getResult();
+                } else {
+                  return null;
```
ashwinraghav, line 254: Previously you were returning an expired response here which would have cached an expired campaign. Now nothing would be cached. I assume that is acceptable too?
Worth checking what happens on IID fetch failure (say when there is no network) and whether we are able to eventually recover.

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +244,38 @@ private boolean shouldIgnoreCache(String event) {
...
+                }
+              });
+
+              Function<CampaignImpressionList, Maybe<FetchEligibleCampaignsResponse>> serviceFetch = campaignImpressionList -> {
+                if (!dataCollectionHelper.isAutomaticDataCollectionEnabled()) {
+                  return Maybe.just(FetchEligibleCampaignsResponse.newBuilder().setExpirationEpochTimestampMillis(1).build());
```
ashwinraghav, line 260: This is also worth modelling the same way as the IID failure above

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +244,38 @@ private boolean shouldIgnoreCache(String event) {
...
+                }
+
+                return getIID.flatMap(iid -> Maybe.fromCallable(() -> apiClient.getFiams(iid, campaignImpressionList)))
```
vkryachko, line 263: Instead of flatMap + wrap into Maybe, can you just use a map() instead?

ashwinraghav: Good point. cc @JasonAHeron worth changing on your PR

PR #1480, opened 2020-04-17
Rewrite code for binary size collection.

yifanyang:
- Move apk generation code into a separate gradle project.
- Move metrics uploading code into fireci.
- A failure from a SDK does not affect metrics collection for other SDKs.
- Misc. cleanup.


ci/fireci/fireci/prow_utils.py
```diff
@@ -0,0 +1,45 @@
...
+  dir_post_submit = "logs"
+  directory = dir_pre_submit if job_type == 'presubmit' else dir_post_submit
+  path = f'{job_name}/{build_id}'
+
+  return f'https://{domain}/view/gcs/{bucket}/{directory}/{path}'
+
+
+def gcloud_identity_token():
+  """Returns an identity token with the current gcloud service account."""
+  result = subprocess.run(['gcloud', 'auth', 'print-identity-token'], stdout=subprocess.PIPE, check=True)
```
ashwinraghav, line 44: I also see this being passed in as an arg from the prow config. Do we need both?

ci/fireci/fireci/uploader.py
```diff
@@ -0,0 +1,59 @@
...
+
+  _logger.info(f'Response: {result.text}')
+
+
+def _construct_request_endpoint():
+  repo_owner = os.getenv('REPO_OWNER')
+  repo_name = os.getenv('REPO_NAME')
+  branch = os.getenv('PULL_BASE_REF')
+  base_commit = os.getenv('PULL_BASE_SHA')
+  head_commit = os.getenv('PULL_PULL_SHA')
```
ashwinraghav, line 50: PULL_PULL_SHA  : Typo?

ci/fireci/fireci/uploader.py
```diff
@@ -0,0 +1,59 @@
...
+_logger = logging.getLogger('fireci.uploader')
+
+
+def post_report(
+    test_report,
+    metrics_service_url=os.getenv('METRICS_SERVICE_URL'),
+    access_token=prow_utils.gcloud_identity_token()):
+  """Post a report to the metrics service backend."""
+
+  endpoint = _construct_request_endpoint()
```
ashwinraghav, line 31: I realise you have just moved this, but Consider using https://uplink.readthedocs.io/en/stable/ if you want to take a declarative approach to building the http client

ci/fireci/fireciplugins/binary_size.py
```diff
@@ -0,0 +1,101 @@
...
+  help='When running for pull requests, only size of affected SDKs are collected.'
+)
+@click.option(
+  '--log',
+  default=prow_utils.prow_job_log_link,
+  help='The link to the log of the prow job, which runs this coverage check.'
+)
+@click.option(
+  '--metrics-service-url',
+  envvar='METRICS_SERVICE_URL',
```
ashwinraghav, line 42: Im trying to map where these values are populated, since there are no defaults and it does not seem to be wired in the prow configs either.

yifanyang:
/test fireci

(with python:3.7 image)

ci/fireci/fireci/gradle.py
```diff
@@ -30,7 +30,7 @@ def P(name, value):
 
 
 @stats.measure_call('gradle')
-def run(*args, gradle_opts='', workdir=None):
+def run(*args, gradle_opts='', workdir=None, gradle_bin_dir='.', check=True):
```
vkryachko, line 33: Instead of changing workdir it might be simpler to run `./gradlew -p api-size ...`, wdyt?

ci/fireci/fireci/uploader.py
vkryachko: Just a consideration: Let's make sure we are confident in the future of uplink, i.e. requests is [`#4`](https://pypistats.org/top) most popular package on pypi while uplink is much less popular, so let's make sure it's not a personal project that is likely to get abandoned.

ashwinraghav: Fair. Are there other packages that you are aware of that provide a retrofit like service?

ashwinraghav: On second thought, this looks pretty good the way it is. No action required

ci/fireci/fireci/prow_utils.py
yifanyang: With the way the job is invoked in prow, this is not needed. The reason why I want to keep it is that it makes it easier for me to test the code (e.g. uploading data from my laptop to a staging metric service in GCP). Does it make sense?

ci/fireci/fireci/uploader.py
yifanyang: This env var is actually indeed called `PULL_PULL_SHA`. See prow doc [here](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables).

ci/fireci/fireciplugins/binary_size.py
yifanyang: It is actually defined as a preset in [prow configs](https://team.git.corp.google.com/firebase-android-core/test-infra/+/refs/heads/master/prow/config.yaml#118).

ci/fireci/fireci/gradle.py
yifanyang: Done. I didn't even know that such a gradle flag exists. Thanks for pointing it out!

Issue #1436, opened 2020-04-08
Crash: android.database.sqlite.SQLiteCantOpenDatabaseException: 


ReCursia:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.6.1
  * Firebase Component: i have problem in /transport directory
  * Components versions:
    firebaseCloudMessaging  20.1.0
    firebaseStorage 16.0.1
    firebaseConfig 16.1.0
    firebasePerfomance       :19.0.5

#### Steps to reproduce:

I have crashes SQLite for Android version >= 7
I do have `<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />`
in AndroidManifest.xml.
Does it write to external storage or it uses internal storage? 
#### Relevant Code:

```
android.database.sqlite.SQLiteCantOpenDatabaseException: 
  at android.database.sqlite.SQLiteConnection.nativeOpen (SQLiteConnection.java)
  at android.database.sqlite.SQLiteConnection.open (SQLiteConnection.java:284)
  at android.database.sqlite.SQLiteConnection.open (SQLiteConnection.java:215)
  at android.database.sqlite.SQLiteConnectionPool.openConnectionLocked (SQLiteConnectionPool.java:705)
  at android.database.sqlite.SQLiteConnectionPool.open (SQLiteConnectionPool.java:272)
  at android.database.sqlite.SQLiteConnectionPool.open (SQLiteConnectionPool.java:239)
  at android.database.sqlite.SQLiteDatabase.openInner (SQLiteDatabase.java:1292)
  at android.database.sqlite.SQLiteDatabase.open (SQLiteDatabase.java:1247)
  at android.database.sqlite.SQLiteDatabase.openDatabase (SQLiteDatabase.java:903)
  at android.database.sqlite.SQLiteDatabase.openDatabase (SQLiteDatabase.java:893)
[...]

PR #1486, opened 2020-04-20
Datatransport failures should not crash the app.

vkryachko:
This is achieved by protecting the executor thread by a catch all. This
makes sure that failures are logged to logcat but do not cause the
underlying thread to crash itself.

Mitigates #1436 until a more detailed repro is provided.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/SafeLoggingExecutor.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/SafeLoggingExecutor.java
@@ -1,0 +1,53 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/SafeLoggingExecutor.java
```diff
@@ -0,0 +1,53 @@
...
+    SafeLoggingRunnable(Runnable delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override
+    public void run() {
+      try {
+        delegate.run();
+      } catch (Exception e) {
+        Logging.e("Executor", "Background execution failure.", e);
```
ashwinraghav, line 49: We should consider logging non fatals to crashlytics in the future. In this case, that could off course loop around and come right back to the problem

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ExecutionModule.java
```diff
@@ -25,6 +25,6 @@
   @Singleton
   @Provides
   static Executor executor() {
-    return Executors.newSingleThreadExecutor();
+    return new SafeLoggingExecutor(Executors.newSingleThreadExecutor());
```
ashwinraghav, line 28: Wow!
DI to the rescue!

PR #1488, opened 2020-04-21
Fix FIAM ANR Issue 1430

JasonAHeron:


firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +243,37 @@ private boolean shouldIgnoreCache(String event) {
...
+                      Logging.logi(
+                          "Automatic data collection is disabled, not attempting campaign fetch from service.");
+                      return Maybe.just(cacheExpiringResponse());
+                    }
+
+                    // blocking get occurs on the IO thread because that's what we observeOn above
+                    return Maybe.fromCallable(getIID::blockingGet)
+                        .map(
+                            iid ->
+                                validIID(iid)
```
ashwinraghav, line 260: consider using `filter`

```
Maybe.fromCallable(getIID::blockingGet)
.filter(validIID())
.map(apiClient.getFiams(iid, campaignImpressionList))
.switchIfEmpty(cacheExpiringResponse())
```

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +243,37 @@ private boolean shouldIgnoreCache(String event) {
...
-                          .onErrorResumeNext(Maybe.empty()); // Absorb service failures
+                  campaignImpressionList -> {
+                    if (!dataCollectionHelper.isAutomaticDataCollectionEnabled()) {
+                      Logging.logi(
+                          "Automatic data collection is disabled, not attempting campaign fetch from service.");
+                      return Maybe.just(cacheExpiringResponse());
+                    }
+
+                    // blocking get occurs on the IO thread because that's what we observeOn above
+                    return Maybe.fromCallable(getIID::blockingGet)
```
ashwinraghav, line 257: You should be able to just return getIID here, which is already a Maybe

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +243,37 @@ private boolean shouldIgnoreCache(String event) {
                       .defaultIfEmpty(CampaignImpressionList.getDefaultInstance())
                       .onErrorResumeNext(Maybe.just(CampaignImpressionList.getDefaultInstance()));
 
+              Maybe<InstanceIdResult> getIID = taskToMaybe(firebaseInstanceId.getInstanceId());
```
ashwinraghav, line 246: Previously: When IID was failing, you were previously returning an expired response from the API Client.
I presume this would be cached, and the next time you lookup the cache, you would find that it has expired and make a service call.

Now: You this maybe would resolve to an error, that on line 274 would print the service failure.

Can you confirm that this logic is accurate by actually testing IID Fetch failure ?

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/ApiClient.java
```diff
@@ -143,17 +111,11 @@ private ClientSignals getClientSignals() {
   }
 
   private ClientAppInfo getClientAppInfo(InstanceIdResult instanceIdResult) {
-    ClientAppInfo.Builder builder =
```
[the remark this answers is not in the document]
JasonAHeron: no just removing the checks because now we do them outside of this class

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
JasonAHeron: done

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
JasonAHeron: ah but I need to do blocking get which turns it into a not maybe, if I don't do blocking get here the waiting ends up hitting the main thread below

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
JasonAHeron: So right now the behavior is essentially FIAM will fetch infinitely if it encounters an error

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InAppMessageStreamManager.java
```diff
@@ -232,22 +243,35 @@ private boolean shouldIgnoreCache(String event) {
...
-                              resp -> impressionStorageClient.clearImpressions(resp).subscribe())
-                          .doOnSuccess(analyticsEventsManager::updateContextualTriggers)
-                          .doOnSuccess(testDeviceHelper::processCampaignFetch)
-                          .doOnError(e -> Logging.logw("Service fetch error: " + e.getMessage()))
-                          .onErrorResumeNext(Maybe.empty()); // Absorb service failures
+                  campaignImpressionList -> {
+                    if (!dataCollectionHelper.isAutomaticDataCollectionEnabled()) {
+                      Logging.logi(
+                          "Automatic data collection is disabled, not attempting campaign fetch from service.");
+                      return Maybe.just(cacheExpiringResponse());
```
ashwinraghav, line 253: Thinking out loud: Can this be modelled as an empty as well, similar to IID?

JasonAHeron: hmm maybe but I think this is pretty clear

JasonAHeron: like it could map to empty or something like that but I think it's not worth it

PR #1495, opened 2020-04-23
Add clarifying javadoc to FirebaseCrashlytics#setCrashlyticsCollectionEnabled

mrichards:
This is not a behavior change; it just clarifies the existing behavior in the javadoc.


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -338,7 +338,10 @@ public boolean didCrashOnPreviousExecution() {
    * disabled. Use {@link #deleteUnsentReports()} to delete any reports stored on the device without
    * sending them to Crashlytics.
    *
-   * @param enabled whether to enable automatic data collection.
+   * @param enabled whether to enable automatic data collection. When set to `false`, the new value
```
RelaxedPear, line 341: Suggest: 

`* @param specifies whether to enable automatic data collection...`

PR #1498, opened 2020-04-24
Port ComponentProvider

schmidt-sebastian:
Ports the component provider from Web: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/component_provider.ts

It also updates `FirestoreClient` and `SpecTestRunner` to use the new component provider. Note that the initialization in `FirestoreClient` is slightly different as there is no memory-only build, so we can decide much later which component provider to use.

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/spec/MemorySpecTest.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/spec/MemorySpecTest.java
@@ -14,8 +14,14 @@
 
 package com.google.firebase.firestore.spec;
 
+import com.google.firebase.firestore.core.ComponentProvider;
+import com.google.firebase.firestore.core.MemoryComponentProvider;
+import com.google.firebase.firestore.local.LocalSerializer;
+import com.google.firebase.firestore.local.LruGarbageCollector;
+import com.google.firebase.firestore.local.MemoryPersistence;
 import com.google.firebase.firestore.local.Persistence;
-import com.google.firebase.firestore.local.PersistenceTestHelpers;
+import com.google.firebase.firestore.model.DatabaseId;
+import com.google.firebase.firestore.remote.RemoteSerializer;
 import java.util.Set;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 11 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -241,7 +241,8 @@ public static void log(String line) {
   // Methods for tracking state of writes.
   //
 
-  abstract Persistence getPersistence(boolean garbageCollectionEnabled);
+  protected abstract ComponentProvider initializeComponentProvider(
+      ComponentProvider.Configuration configuration, final boolean garbageCollectionEnabled);
```
var-const, line 245: Nit: I think Java recommendations discourage `final` on parameters and local variables.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/MemorySpecTest.java
```diff
@@ -33,11 +39,23 @@ protected boolean isExcluded(Set<String> tags) {
...
 
   @Override
-  Persistence getPersistence(boolean garbageCollectionEnabled) {
-    if (garbageCollectionEnabled) {
-      return PersistenceTestHelpers.createEagerGCMemoryPersistence();
-    } else {
-      return PersistenceTestHelpers.createLRUMemoryPersistence();
-    }
+  protected MemoryComponentProvider initializeComponentProvider(
+      ComponentProvider.Configuration configuration, final boolean garbageCollectionEnabled) {
```
var-const, line 43: Same nit re. `final`.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
schmidt-sebastian: Removed. I added it as it is used in an inner class, but it is "effectively final" without it.

PR #1209, opened 2020-02-06
Fix  ClassNotFoundException for com.google.firebase.remoteconfig.ktx.…

b95505017:
…FirebaseConfigKtxRegistrar

Should be `FirebaseRemoteConfigKtxRegistrar`

PR #1501, opened 2020-04-24
Add missing test to RemoteConfig KTX

rlazo:


rlazo:
This is a follow up to #1209

PR #1503, opened 2020-04-27
feat: Implement firebase-crashlytics-ktx

rosariopfernandes:
Adding Kotlin Extensions for the new Crashlytics SDK :)

Not only this allows to use `Firebase.crashlytics`, but also setting custom keys:

```kotlin
Firebase.crashlytics.setCustomKeys(
    "str_key" to "hello",
    "bool_key" to true,
    "int_key" to 1,
    "long_key" to 1L,
    "float_key" to 1.0f,
    "double_key" to 1.0
)
```

### DIFF
--- a/firebase-crashlytics/ktx/src/main/kotlin/com/google/firebase/crashlytics/ktx/FirebaseCrashlytics.kt
+++ b/firebase-crashlytics/ktx/src/main/kotlin/com/google/firebase/crashlytics/ktx/FirebaseCrashlytics.kt
@@ -26,21 +26,12 @@ import com.google.firebase.platforminfo.LibraryVersionComponent
 val Firebase.crashlytics: FirebaseCrashlytics
     get() = FirebaseCrashlytics.getInstance()
 
-fun FirebaseCrashlytics.setCustomKeys(vararg pairs: Pair<String, Any>) {
-    for ((key, value) in pairs) {
-        when (value) {
-            is Boolean -> setCustomKey(key, value)
-            is Double -> setCustomKey(key, value)
-            is Float -> setCustomKey(key, value)
-            is Int -> setCustomKey(key, value)
-            is Long -> setCustomKey(key, value)
-            is String -> setCustomKey(key, value)
-            else -> {
-                val valueType = value::class.java.componentType!!.canonicalName
-                throw IllegalArgumentException("Illegal value type $valueType for key \"$key\"")
-            }
-        }
-    }
+/**
+ * Associates all key-value parameters with the reports
+ */
+fun FirebaseCrashlytics.setCustomKeys(init: KeyValueBuilder.() -> Unit) {
+    val builder = KeyValueBuilder(this)
+    builder.init()
 }
 
 internal const val LIBRARY_NAME: String = "fire-cls-ktx"
--- a/firebase-crashlytics/ktx/src/main/kotlin/com/google/firebase/crashlytics/ktx/FirebaseCrashlytics.kt
+++ b/firebase-crashlytics/ktx/src/main/kotlin/com/google/firebase/crashlytics/ktx/FirebaseCrashlytics.kt
@@ -38,7 +38,7 @@ internal const val LIBRARY_NAME: String = "fire-cls-ktx"
 
 /** @suppress */
 @Keep
-class FirebaseCrashlyticsRegistrar : ComponentRegistrar {
+class FirebaseCrashlyticsKtxRegistrar : ComponentRegistrar {
     override fun getComponents(): List<Component<*>> =
             listOf(LibraryVersionComponent.create(LIBRARY_NAME, BuildConfig.VERSION_NAME))
 }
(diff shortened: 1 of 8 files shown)


ashwinraghav:
Thanks for helping us with your contribution. One of us will be sure to take a closer look soon. 
Just a heads up that there's bound to be a bit of back and forth since we'd have to run a formal API review within the company before we can actually ship this. So it's possible that there is some back and forth. Just a heads up :)

rosariopfernandes:
@ashwinraghav I totally understand! No problems at all :)

firebase-crashlytics/ktx/src/test/java/com/google/firebase/crashlytics/ktx/CrashlyticsTests.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.google.firebase.platforminfo.UserAgentPublisher
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+
+const val APP_ID = "APP_ID"
+const val API_KEY = "API_KEY"
```
andirayo, line 33: const val APP_ID = "1:1234567890:android:321abc456def7890"
const val API_KEY = "AIzaSyDOCAbC123dEf456GhI789jKl012-MnO"

I recommend using values for fake Firebase options that could pass simple checks for validity in order to prevent tests failing for future initialization checks.

rosariopfernandes: @andirayo  Thank you! I've updated these values

firebase-crashlytics/ktx/src/main/kotlin/com/google/firebase/crashlytics/ktx/FirebaseCrashlytics.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.google.firebase.components.ComponentRegistrar
+import com.google.firebase.crashlytics.FirebaseCrashlytics
+import com.google.firebase.ktx.Firebase
+import com.google.firebase.platforminfo.LibraryVersionComponent
+
+/** Returns the [FirebaseCrashlytics] instance of the default [FirebaseApp]. */
+val Firebase.crashlytics: FirebaseCrashlytics
+    get() = FirebaseCrashlytics.getInstance()
+
+fun FirebaseCrashlytics.setCustomKeys(vararg pairs: Pair<String, Any>) {
```
vkryachko, line 29: This is definitely an improvement compared to using plain java api, a few considerations though:
* Some other ktx sdks use dsl-like notation for similar use-cases, e.g. [Firebase.analytics.logEvent](https://firebase.google.com/docs/reference/kotlin/com/google/firebase/analytics/ktx/package-summary?hl=en#logevent), so it might make sense to something similar here for the sake of consistency. The dsl could even support `infix` notation like `Pair` does(not sure if it would be a good idea though).
* The dsl would also be more type-safe, i.e. there won't be a need for `IllegalArgumentException` below, as the block can be made to only have the methods with argument types crashlytics supports.

So I would suggest something along the lines of:
```kotlin
crashlytics.setCustomKeys {
  key("foo", "bar")
  key("int", 42)
  // or
  "foo" to "bar" // this might not be a good idea as it shadows the Pair constructor function.
}
```

wdyt?

rosariopfernandes: @vkryachko since `analytics.logEvent` uses DSL, I think we should maintain consistency. Ive updated the code to match your suggestion. PTAL

firebase-crashlytics/ktx/src/main/kotlin/com/google/firebase/crashlytics/ktx/FirebaseCrashlytics.kt
```diff
@@ -0,0 +1,44 @@
...
+fun FirebaseCrashlytics.setCustomKeys(init: KeyValueBuilder.() -> Unit) {
+    val builder = KeyValueBuilder(this)
+    builder.init()
+}
+
+internal const val LIBRARY_NAME: String = "fire-cls-ktx"
+
+/** @suppress */
+@Keep
+class FirebaseCrashlyticsRegistrar : ComponentRegistrar {
```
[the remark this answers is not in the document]
rosariopfernandes: Oops! Done :)

rlazo:
Thanks for the patience @rosariopfernandes the PR has been approved. Feel free to merge it. And thank you for contributions!

rosariopfernandes:
@rlazo Thanks for confirming the approval! :)

I don't have write access to the Firebase repos, do you mind merging it? :sweat_smile:

rlazo:
Done 😄

PR #1507, opened 2020-04-28
Move FIS precondition format check to installation constructor

ChaoqunCHEN:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -142,17 +142,17 @@ public Thread newThread(Runnable r) {
    * FirebaseOptions#getApplicationId()} , and ({@link FirebaseOptions#getProjectId()} are not null
    * or empty.
    */
-  private void preConditionChecks() {
-    Preconditions.checkNotEmpty(getApplicationId());
-    Preconditions.checkNotEmpty(getProjectIdentifier());
-    Preconditions.checkNotEmpty(getApiKey());
```
ankitaj224, line 148: Did you remove these methods?

getApplicationId, getProjectIdentifier, getApiKey

Also, for getProjectIdentifier we wanted to either use project id or sender id.

ChaoqunCHEN: - No, I didn't remove those methods. I just made preConditionChecks to be static and move it to constructor from getId() and getToken().

- getProjectIdentifier was changed to take only projectId on last PR as requested by Rayo.

ankitaj224: nit: If those methods arent used remove them.

ChaoqunCHEN: Thanks, those methods are used somewhere else. So I keep them.

PR #1508, opened 2020-04-28
revert "Move precondition to constructor"

ChaoqunCHEN:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -229,6 +229,7 @@ String getName() {
   @NonNull
   @Override
   public Task<InstallationTokenResult> getToken(boolean forceRefresh) {
+    preConditionChecks();
```
andirayo, line 232: what about deleteToken?

ChaoqunCHEN: We never had it on deleteToken. Shall we add it?

Issue #1519, opened 2020-05-01
OverlappingFileLockException crashes with up to date libraries versions


JeremyR34:
* Android Studio version: 3.6.3
  * Firebase Components & versions:
   com.google.firebase:firebase-core			17.4.0
   com.google.firebase:firebase-analytics		17.4.0
   com.google.firebase:firebase-messaging		20.1.6
   com.crashlytics.sdk.android:crashlytics		2.10.1
   com.google.firebase:firebase-dynamic-links	19.1.0
   com.google.firebase:firebase-database		19.3.0
   com.google.firebase:firebase-config			19.1.4

### Describe the problem

After releasing a new version ni production with up to date Firebase libraries we have many new crashes (Google Play developer console & Firebase Crashlytics).

#### Steps to reproduce:

Up-to-date libraries versions.
Notes
- it seems to happen only on Android 7 & 8 devices.
- it may be related to this: https://github.com/firebase/firebase-android-sdk/pull/1234

#### Relevant Code:

```
Fatal Exception: java.nio.channels.OverlappingFileLockException
       at sun.nio.ch.SharedFileLockTable.checkList(SharedFileLockTable.java:255)
       at sun.nio.ch.SharedFileLockTable.add(SharedFileLockTable.java:152)
       at sun.nio.ch.FileChannelImpl.lock(FileChannelImpl.java:1080)
       at java.nio.channels.FileChannel.lock(FileChannel.java:1053)
       at com.google.firebase.installations.CrossProcessLock.acquire(CrossProcessLock.java:52)
       at com.google.firebase.installations.FirebaseInstallations.getPrefsWithGeneratedIdMultiProcessSafe(FirebaseInstallations.java:376)
       at com.google.firebase.installations.FirebaseInstallations.doNetworkCall(FirebaseInstallations.java:327)
       at com.google.firebase.installations.FirebaseInstallations.lambda$doRegistrationInternal$0(FirebaseInstallations.java:323)
       at com.google.firebase.installations.FirebaseInstallations$$Lambda$5.run(FirebaseInstallations.java:4)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
       at java.lang.Thread.run(Thread.java:764)
```

Another kind of new crashes we have that seems also related (FirebaseInstallations):
```
[...]

PR #1521, opened 2020-05-01
Fixing Bad File descriptor IOException to avoid app crashes.

ankitaj224:
Addresses #1519


firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -51,9 +51,10 @@ static CrossProcessLock acquire(Context appContext, String lockName) {
       // This method blocks until it can retrieve the lock.
       lock = channel.lock();
       return new CrossProcessLock(channel, lock);
-    } catch (IOException e) {
+    } catch (IOException | Error e) {
       // Certain conditions can cause file locking to fail, such as out of disk or bad permissions.
       // In any case, the acquire will fail and return null instead of a held lock.
+      // NOTE: In Java 7 & 8, FileKey creation failure wraps IOException into Error.
```
andirayo, line 57: It does not always wrap I think - we only got 1 developer mentioning this to us.
I think we should say "`might wrap`" instead of "`wraps`".
How about adding a link to https://bugs.openjdk.java.net/browse/JDK-8025619?

ankitaj224: Done.

PR #1530, opened 2020-05-05
FIAM FIS Integration 

JasonAHeron:


firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -76,7 +76,7 @@
  * </ul>
  *
  * <p>To delete the Instance ID and the data associated with it, see {@link
- * com.google.firebase.iid.FirebaseInstanceId#deleteInstanceId}.
+ * FirebaseInstallationsApi#delete.
```
rlazo, line 79: missing } after delete

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InstallationIdResult.java
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+package com.google.firebase.inappmessaging.internal;
+
+import android.text.TextUtils;
+import com.google.firebase.installations.InstallationTokenResult;
+
+public class InstallationIdResult {
+  private String id;
+  private InstallationTokenResult token;
```
prakhar1989, line 22: private final

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/InstallationIdResult.java
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.inappmessaging.internal;
+
+import android.text.TextUtils;
+import com.google.firebase.installations.InstallationTokenResult;
+
+public class InstallationIdResult {
```
prakhar1989, line 20: Do we need this to be `public`.. I think package private would work

prakhar1989, line 20: final

prakhar1989, line 20: docs on this class?

prakhar1989, line 20: Ideally we should be using https://github.com/google/auto/blob/master/value/userguide/index.md for these value objects. If its not too much work, consider adding using this lib

JasonAHeron: turns out we access this from two different packages for tests

PR #1532, opened 2020-05-07
Migrate all tests to kotlin, remove groovy plugin from buildSrc.

vkryachko:


rosariopfernandes:
@vkryachko If I may ask, what's the reason for this change? (besides the fact that Kotlin is more pleasant to read and write :wink:)

vkryachko:
hey @rosariopfernandes , this work is an intermediate step towards Kotlin in buildSrc. The motivation is
1. Maintaining groovy has been a pain for us, due to it's dynamic typing, i.e. you're never sure if what you write will actually work
2. So we started migrating logic to java, but it is really verbose to write and has so many nested lambdas, i.e. https://github.com/firebase/firebase-android-sdk/blob/631484a1407da3dde3cfce23324a466ffd3f01b3/buildSrc/src/main/java/com/google/firebase/gradle/plugins/Dokka.java#L99-L126

So our goal is to end up with code in Kotlin, for that we had to remove the groovy plugin as it causes issues when the kotlin plugin is enabled(at least in the version of gradle we are currently stuck with).

So now that groovy is gone, we can start writing concise(like we did in groovy) and type-safe(like we did in Java) code :)

rosariopfernandes:
@vkryachko Thanks for the detailed explanation :)

PR #1534, opened 2020-05-07
Migrating to new protolite runtime

ashwinraghav:
### Required reading : [Protos in dependencies ](https://github.com/google/protobuf-gradle-plugin#protos-in-dependencies)
|   configuration | `protobuf`| `implementation`|
| :---| :---: | :---: |
|  `.proto` files available to proto compiler | ✅ | ✅|
| Generated classes compiled into artifact   |✅| ❌|

### Problem 1 - Choice of runtime

The `protobuf-lite` runtime has been renamed to `protobuf-javalite`. These libraries are incompatible since generated code has **not** been repackaged as part of the migration. Libraries that use `protobuf-lite` have one of three choices
1. Never migrate to `javalite` and sacrifice any fixes made upstream.
2. Migrate to `javalite` and potentially break downstream consumers using older versions of the runtime.
3. Repackage the new runtime and ship a vendored version.

We are going with 2, until we move away from protos all together. cc @vkryachko 
Currently some customers resort to [fatjaring](https://github.com/gitprelimtek/protobuf-javalite-firebase-wellknowntypes). Even with our solution, this may not go away for customers who are stuck on older runtimes.

### Problem 2  - `com.google.firebase:protolite-well-known-types`
 
Our well known types library comprises of the set of `.proto` and files and the corresponding `.class` files that firebase libraries depend on. These were scoped to 2 conceptual sets
1. Proto's [well known types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf), brought in as ` protobuf 'com.google.protobuf:protobuf-java:3.6.1'`
2 . Cloud's [backend](https://github.com/googleapis/api-common-protos) protos, brought in as `protobuf 'com.google.api.grpc:proto-google-common-protos:1.12.0'`

Our well knon types library helps firestore and inappmessaging `import` protos, as well as take run time dependencies on them.

In addition to renaming the runtime artifact, the javalite runtime **now** includes well known types class files and `.proto` files i.e 1. above.

|  Library  | protobuf-lite | protobuf-javalite |
| :---| :---: | :---: |
|  Contain the `.class` files for [well known types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf)   | ❌ | ✅|
|  Contain the `.proto` files for [well known types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf)   | ❌ | ✅**|

**_javalite has the classfiles for [Duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration), but not the `.proto` file. This seems like a [bug](https://github.com/protocolbuffers/protobuf/issues/7094)_

As a result, we are going to have to figure out a way to have the "Duration.proto" be available to firebase projects and the protoc compiler, but **never** shipping its class files, since that would conflict with the javalite runtime.

### Problem 3 Descriptor.proto
In addition to the well known types, firebase libraries depend on `descriptor.proto`, which is **not a well known type** and we need to have this be available at compile (and runtime) to firebase libraries.

### Solution

We change the well known types libraries in the following ways.

1. We remove `protobuf 'com.google.protobuf:protobuf-java:3.6.1'` (previously used to vendor the well known types), since we now use `implementation 'com.google.protobuf:protobuf-javalite:3.11.0'` that contains these proto and class files.

2. To solve Problem 3, we manually delete the Duration files in a post compile phase.

3. We continue vendoring in the  cloud protos `protobuf("com.google.api.grpc:proto-google-common-protos:1.17.0")`

# Action items
1. Write some firehorn tests for fiam and remote config. These tests for firestore were the only tests that caught the incompatibilities. See commit history.
2. Investigate why we need to continue vendoring cloud protos. Ideally, we should be able to just to take a dep on it. 
```
implementation("com.google.api.grpc:proto-google-common-protos:1.17.0"){
        exclude group: "com.google.protobuf", module: "protobuf-java"
}
```
3. Is there a cleaner approach than having to delete these class files ? These are the options that failed
`protobuf files('lib/')`
`compile files('lib/')`
It seems like the compile dep on "files" will bring in the class files into the aar.


protolite-well-known-types/protolite-well-known-types.gradle
```diff
@@ -50,13 +50,9 @@ android {
...
 }
 
-configurations.all {
-    exclude group: "com.google.protobuf", module: "protobuf-java"
-}
-
 dependencies {
     implementation 'com.google.protobuf:protobuf-javalite:3.11.0'
-    protobuf 'com.google.protobuf:protobuf-javalite:3.11.0'
+    protobuf 'com.google.protobuf:protobuf-java:3.11.4'
```
ashwinraghav, line 55: The `protobuf` declaration is different from a `implementation`.

IIRC, this will simply make the proto files in the artifact available to all consumers of this dep. It should not vendor the protos itself. It should simply make it available to the proto compiler

ashwinraghav:
> # Binary Size Report
> ### Affected SDKs
> No changes between base commit ([b8c0062](https://github.com/firebase/firebase-android-sdk/commit/b8c00623a8dcdf19a0bb0f871e97e0001725531c)) and head commit (7506c69d).
> 
> ### Test Logs
> * `Base (b8c00623)`: https://android-ci.firebaseopensource.com/view/gcs/android-ci/logs/postsubmit-binary-size/1258888792878616577
> * `Head (7506c69d)`: https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/1534/binary-size/1259575448220733440
> 
> ### Notes
> Head commit (7506c69d) is created by Prow via merging commits: [b8c0062](https://github.com/firebase/firebase-android-sdk/commit/b8c00623a8dcdf19a0bb0f871e97e0001725531c) [d976df0](https://github.com/firebase/firebase-android-sdk/commit/d976df0a390886aa5719877344853c86425414af).

@yifanyang Can you help investigate why no changes are detected. I would expect this change to have an impact on binary size.

rlazo:
Last smoke-test failure was caused by

```
Process crashed while executing setValueShouldTriggerListenerWithNewlySetData(com.google.firebase.testing.DatabaseTest):
f.c.c.c: Field messages_ for b.b.g.a.a.a.a.n not found. Known fields are [public b.b.h.M$i b.b.g.a.a.a.a.n.g, public long b.b.g.a.a.a.a.n.h, public static final b.b.g.a.a.a.a.n b.b.g.a.a.a.a.n.e, public static volatile b.b.h.pa b.b.g.a.a.a.a.n.f]
	at f.c.f.b.a$k.accept(Functions.java:2)
	at f.c.f.h.c.a(LambdaSubscriber.java:8)
	at f.c.f.e.b.r$a.a(FlowableObserveOn.java:21)
	at f.c.f.e.b.r$c.b(FlowableObserveOn.java:7)
	at f.c.f.e.b.r$a.run(FlowableObserveOn.java:5)
	at f.c.a.a.c$b.run(HandlerScheduler.java:1)
	at android.os.Handler.handleCallback(Handler.java:789)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6600)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:772)
Caused by: java.lang.RuntimeException: Field messages_ for b.b.g.a.a.a.a.n not found. Known fields are [public b.b.h.M$i b.b.g.a.a.a.a.n.g, public long b.b.g.a.a.a.a.n.h, public static final b.b.g.a.a.a.a.n b.b.g.a.a.a.a.n.e, public static volatile b.b.h.pa b.b.g.a.a.a.a.n.f]
	at b.b.h.ja.a(MessageSchema.java:112)
	at b.b.h.ja.a(MessageSchema.java:54)
	at b.b.h.ja.a(MessageSchema.java:80)
	at b.b.h.Z.a(ManifestSchemaFactory.java:26)
	at b.b.h.sa.a(Protobuf.java:3)
	at b.b.h.sa.a(Protobuf.java:7)
```

ashwinraghav:
> Last smoke-test failure was caused by
> 
> ```
> Process crashed while executing setValueShouldTriggerListenerWithNewlySetData(com.google.firebase.testing.DatabaseTest):
> f.c.c.c: Field messages_ for b.b.g.a.a.a.a.n not found. Known fields are [public b.b.h.M$i b.b.g.a.a.a.a.n.g, public long b.b.g.a.a.a.a.n.h, public static final b.b.g.a.a.a.a.n b.b.g.a.a.a.a.n.e, public static volatile b.b.h.pa b.b.g.a.a.a.a.n.f]
> 	at f.c.f.b.a$k.accept(Functions.java:2)
> 	at f.c.f.h.c.a(LambdaSubscriber.java:8)
> 	at f.c.f.e.b.r$a.a(FlowableObserveOn.java:21)
> 	at f.c.f.e.b.r$c.b(FlowableObserveOn.java:7)
> 	at f.c.f.e.b.r$a.run(FlowableObserveOn.java:5)
> 	at f.c.a.a.c$b.run(HandlerScheduler.java:1)
> 	at android.os.Handler.handleCallback(Handler.java:789)
> 	at android.os.Handler.dispatchMessage(Handler.java:98)
> 	at android.os.Looper.loop(Looper.java:164)
> 	at android.app.ActivityThread.main(ActivityThread.java:6600)
> 	at java.lang.reflect.Method.invoke(Native Method)
> 	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
> 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:772)
> Caused by: java.lang.RuntimeException: Field messages_ for b.b.g.a.a.a.a.n not found. Known fields are [public b.b.h.M$i b.b.g.a.a.a.a.n.g, public long b.b.g.a.a.a.a.n.h, public static final b.b.g.a.a.a.a.n b.b.g.a.a.a.a.n.e, public static volatile b.b.h.pa b.b.g.a.a.a.a.n.f]
> 	at b.b.h.ja.a(MessageSchema.java:112)
> 	at b.b.h.ja.a(MessageSchema.java:54)
> 	at b.b.h.ja.a(MessageSchema.java:80)
> 	at b.b.h.Z.a(ManifestSchemaFactory.java:26)
> 	at b.b.h.sa.a(Protobuf.java:3)
> 	at b.b.h.sa.a(Protobuf.java:7)
> ```

These seem like flakes, unrelated to the current PR ?

rlazo:
Probably a more important context of the exception

```
Caused by: java.lang.RuntimeException: Field messages_ for b.b.g.a.a.a.a.n not found. Known fields are [public b.b.h.M$i b.b.g.a.a.a.a.n.g, public long b.b.g.a.a.a.a.n.h, public static final b.b.g.a.a.a.a.n b.b.g.a.a.a.a.n.e, public static volatile b.b.h.pa b.b.g.a.a.a.a.n.f]
	at b.b.h.ja.a(MessageSchema.java:112)
	at b.b.h.ja.a(MessageSchema.java:54)
	at b.b.h.ja.a(MessageSchema.java:80)
	at b.b.h.Z.a(ManifestSchemaFactory.java:26)
	at b.b.h.sa.a(Protobuf.java:3)
	at b.b.h.sa.a(Protobuf.java:7)
	at b.b.h.H.l(GeneratedMessageLite.java:2)
	at b.b.h.H$a.e(GeneratedMessageLite.java:3)
	at b.b.h.H$a.d(GeneratedMessageLite.java:1)
	at b.b.d.o.a.ra.a(InAppMessageStreamManager.java:21)
	at b.b.d.o.a.W.apply(InAppMessageStreamManager.java:8)
	at f.c.f.e.c.g$a.a(MaybeFlatten.java:4)
```

so it seems like it's caused by a field in the `InAppMessageStreamManager` class. I haven't been able to put a finger on it yet.

rlazo:
This failure is happening with **release builds only**, debug builds tests pass
```
SUCCESS (at: 2020-05-11T17:47:11.956198Z)
    /usr/bin/gcloud
    firebase
    test
    android
    run
    --type=instrumentation
    --no-record-video
    --app=/home/prow/go/src/github.com/firebase/firebase-android-sdk/smoke-tests/build-debug-headPiper-smoke-test/outputs/apk/debug/smoke-tests-debug.apk
    --test=/home/prow/go/src/github.com/firebase/firebase-android-sdk/smoke-tests/build-debug-headPiper-smoke-test/outputs/apk/androidTest/debug/smoke-tests-debug-androidTest.apk
    --project=fireescape-c4819
    --results-bucket=android-ci
    --results-dir=pr-logs/pull/firebase_firebase-android-sdk/1534/smoke-tests/1259888141813157890/artifacts/debug-headPiper-smoke-test
FAILURE (at: 2020-05-11T17:47:11.956455Z, status: 10)    /usr/bin/gcloud
    firebase
    test
    android
    run
    --type=instrumentation
    --no-record-video
    --app=/home/prow/go/src/github.com/firebase/firebase-android-sdk/smoke-tests/build-release-headPiper-smoke-test/outputs/apk/release/smoke-tests-release.apk
    --test=/home/prow/go/src/github.com/firebase/firebase-android-sdk/smoke-tests/build-release-headPiper-smoke-test/outputs/apk/androidTest/release/smoke-tests-release-androidTest.apk
    --project=fireescape-c4819
    --results-bucket=android-ci
    --results-dir=pr-logs/pull/firebase_firebase-android-sdk/1534/smoke-tests/1259888141813157890/artifacts/release-headPiper-smoke-test
```

build.gradle
```diff
@@ -49,6 +49,7 @@ ext {
     playServicesVersion = '16.0.1'
     supportAnnotationsVersion = '28.0.0'
     googleTruthVersion = '0.45'
+    grpcVersion = '1.29.0'
```
rlazo, line 52: 1.29.0 depends on guava 28.2 

https://github.com/grpc/grpc-java/blob/master/build.gradle#L53 

but we have a hard dependency on 27

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-inappmessaging/firebase-inappmessaging.gradle#L107

I'm curious on how does the tests worked with this issue.

rlazo:
Javalite uses some reflexion, which isn't compatible with proguard. An additional rule is needed to prevent fields to be removed as noted in

https://github.com/protocolbuffers/protobuf/issues/6463#issuecomment-553183215

rlazo:
Here's a description of the changes required to migrate our codebase
from `protobuf-lite` runtime to `protobuf-javalite` ([github](https://github.com/protocolbuffers/protobuf/blob/master/java/lite.md))

1.  **Migrated to newer versions** GRPC 1.28, Protoc 3.11.0, Javalite
    3.11.0. To simplify updating versions, versions come from a var in
    the base gradle file ([build.gradle](https://github.com/firebase/firebase-android-sdk/pull/1534/files#diff-c197962302397baf3a4cc36463dce5ea))
2.  **Remove artifact and javalite plugin declarations** The newer
    runtime comes included in the regular protoc, no longer do we need
    to use plugins for it ([gradle plugin ref](https://github.com/google/protobuf-gradle-plugin))
3.  **Vendor duration.proto** Javalite runtime includes the `Duration`
    message class, but not the proto definition. Since we have proto
    files depending on `Duration` we copied over the Duration proto
    definition. We only need the file during codegen, including it in
    the generated artifact would cause duplicated classes. We use the
    includelib workaround for this ([protobuf issue](https://github.com/protocolbuffers/protobuf/issues/7094) | [workaround](https://github.com/protocolbuffers/protobuf/issues/7094#issuecomment-637044968))
4.  **Vendor Descriptor proto** It's missing from the runtime ([source](https://github.com/firebase/firebase-android-sdk/pull/1534/files#diff-49e1881f5a290e3da933805304128352))
5.  **Add proguard rule to preserve internal fields** Javalite uses
    reflection, which requires us to include proguard rules that prevent
    some internal fields from being stripped
    ([protocolbuffers/protobuf#6463](https://github.com/protocolbuffers/protobuf/issues/6463#issuecomment-553183215) | [code](https://github.com/firebase/firebase-android-sdk/pull/1534/files#diff-3ab6fe7be8b518bab0591a21606740cd)).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
```diff
@@ -125,7 +125,7 @@ private ManagedChannel initChannel(Context context, DatabaseInfo databaseInfo) {
     // Wrap the ManagedChannelBuilder in an AndroidChannelBuilder. This allows the channel to
     // respond more gracefully to network change events (such as switching from cell to wifi).
     AndroidChannelBuilder androidChannelBuilder =
-        AndroidChannelBuilder.fromBuilder(channelBuilder).context(context);
```
ashwinraghav, line 128: Intentional?

protolite-well-known-types/README.md
```diff
@@ -29,26 +29,22 @@ A typical grpc and protobuf-lite configuration should look like
 ```groovy
```
ashwinraghav, line 29: I think the readme files in well known types need to change in a follow up PR

protolite-well-known-types/proguard-rules.pro
```diff
@@ -19,3 +19,7 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+-keepclassmembers class * extends com.google.protobuf.GeneratedMessageLite {
```
ashwinraghav, line 23: Can you add a comment on why this is needed? It seems like we did not previously had to preserve these classes

protolite-well-known-types/protolite-well-known-types.gradle
```diff
@@ -45,6 +45,7 @@ android {
         minSdkVersion project.minSdkVersion
         versionCode 1
         versionName '1.0'
+	consumerProguardFiles 'proguard-rules.pro'
```
ashwinraghav, line 48: Fix indentation.

ashwinraghav: I think all other SDKs have the file named proguard.txt. I think that file name may automagically be picked up. Worth confirming and removing this config

ashwinraghav:
It's also worth putting a finger on what exacly causes the substantial size size increase in config and the smaller increases in fiam and firestore.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/GrpcCallProvider.java
rlazo: Yes, `fromBuilder` is deprecated

https://github.com/grpc/grpc-java/blob/master/android/src/main/java/io/grpc/android/AndroidChannelBuilder.java#L90

protolite-well-known-types/proguard-rules.pro
rlazo: Done.

rlazo:
**A note about size implications of this change:**

The new `protobuf-javalite` artifact comes with a size and method count penalty. It includes more classes, and more method per class. See the attached image for a comparison of the same class using the new artifact (left), and using the old artifact (right).
Screen Shot 2020-06-23 at 12 43 47 PM

firebase-firestore/firebase-firestore.gradle
```diff
@@ -124,10 +118,10 @@ dependencies {
     compileOnly 'javax.annotation:jsr250-api:1.0'
 
     implementation 'androidx.annotation:annotation:1.1.0'
-    implementation 'io.grpc:grpc-stub:1.21.0'
-    implementation 'io.grpc:grpc-protobuf-lite:1.21.0'
-    implementation 'io.grpc:grpc-okhttp:1.21.0'
-    implementation 'io.grpc:grpc-android:1.21.0'
+    implementation "io.grpc:grpc-stub:$grpcVersion"
+    implementation "io.grpc:grpc-protobuf-lite:$grpcVersion"
```
rosariopfernandes, line 122: still using `protobuf-lite`

firebase-inappmessaging/firebase-inappmessaging.gradle
```diff
@@ -123,10 +116,10 @@ dependencies {
     compileOnly 'javax.annotation:jsr250-api:1.0'
 
     implementation 'com.google.android.gms:play-services-tasks:17.0.2'
-    implementation 'io.grpc:grpc-stub:1.21.0'
-    implementation 'io.grpc:grpc-protobuf-lite:1.21.0'
-    implementation 'io.grpc:grpc-okhttp:1.21.0'
-    implementation 'com.google.dagger:dagger:2.27'
+    implementation "io.grpc:grpc-stub:$grpcVersion"
+    implementation "io.grpc:grpc-protobuf-lite:$grpcVersion"
```
rosariopfernandes, line 120: still using `protobuf-lite`

rlazo:
@rosariopfernandes `grpc-protobuf-lite` actually depends on `protobuf-javalite` since 1.30

https://github.com/grpc/grpc-java/commit/48929c4d502f3613a52bb13551c7d6b2a34f80a2

rosariopfernandes:
@rlazo I see. I was wondering if this is why the smoke-tests ([e.g. #1690](https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/1690/smoke-tests/1275537297248358401/)) are failing with

```
java.lang.RuntimeException: Duplicate class com.google.protobuf.AbstractMessageLite found in modules protobuf-javalite-3.11.0.jar (com.google.protobuf:protobuf-javalite:3.11.0) and protobuf-lite-3.0.1.jar (com.google.protobuf:protobuf-lite:3.0.1)
```

rlazo:
Actually, since `1.28.0` which is the version we are using

https://repo1.maven.org/maven2/io/grpc/grpc-protobuf-lite/1.28.0/grpc-protobuf-lite-1.28.0.pom

Thanks for taking a look at this~

PR #1537, opened 2020-05-08
Log the available Firebase options along with the error stream.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -522,13 +525,21 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
...
       Log.w(FIS_TAG, logString);
+      Log.w(FIS_TAG, firebaseOptions);
     }
   }
 
+  private String availableFirebaseOptions(
+      @Nullable String appId, @NonNull String apiKey, @NonNull String projectId) {
+    return String.format(
+        "Firebase options used while communicating with Firebase server APIs: %s, %s, %s",
+        TextUtils.isEmpty(appId) ? "" : appId, apiKey, projectId);
```
andirayo, line 540: I don't think that this `TextUtils.isEmpty(appId) ? "" : appId` is necessary.
Just add appId and `String.format` will replace it by an empty String.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -522,13 +525,21 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
...
     if (!TextUtils.isEmpty(logString)) {
       Log.w(FIS_TAG, logString);
+      Log.w(FIS_TAG, firebaseOptions);
     }
   }
 
+  private String availableFirebaseOptions(
+      @Nullable String appId, @NonNull String apiKey, @NonNull String projectId) {
+    return String.format(
+        "Firebase options used while communicating with Firebase server APIs: %s, %s, %s",
```
andirayo, line 539: ... %s, %s (%s)", apiKey, projectId, appId

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -165,7 +165,8 @@ public InstallationResponse createFirebaseInstallation(
           return readCreateResponse(httpURLConnection);
         }
 
-        logFisCommunicationError(httpURLConnection);
+        logFisCommunicationError(
+            httpURLConnection, availableFirebaseOptions(appId, apiKey, projectID));
```
andirayo, line 169: rather than repeating `availableFirebaseOptions(` 3 times, I recommend calling `#availableFirebaseOptions` from within the `#logFisCommunicationError` method.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -522,13 +525,21 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
         : Long.parseLong(expiresIn.substring(0, expiresIn.length() - 1));
   }
 
-  private static void logFisCommunicationError(HttpURLConnection conn) {
+  private static void logFisCommunicationError(HttpURLConnection conn, String firebaseOptions) {
     String logString = readErrorResponse(conn);
     if (!TextUtils.isEmpty(logString)) {
       Log.w(FIS_TAG, logString);
+      Log.w(FIS_TAG, firebaseOptions);
```
andirayo, line 532: I recommend calling `#availableFirebaseOptions` here.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
andirayo: I would prefer this order, given that API key is the most important value and App ID is not always available.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Unless otherwise specified, passing a null argument to any method or constructor in this class will cause a NullPointerException to be thrown.

https://developer.android.com/reference/java/util/Formatter

Its safe to format on our end correctly vs depending on underlying Formatter impl.

andirayo: https://screenshot.googleplex.com/JVHfUAGMKGt

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -522,13 +522,25 @@ static long parseTokenExpirationTimestamp(String expiresIn) {
...
       Log.w(FIS_TAG, logString);
+      Log.w(FIS_TAG, availableFirebaseOptions(appId, apiKey, projectId));
     }
   }
 
+  private static String availableFirebaseOptions(
+      @Nullable String appId, @NonNull String apiKey, @NonNull String projectId) {
+    return String.format(
+        "Firebase options used while communicating with Firebase server APIs: %s, %s, %s",
+        apiKey, projectId, TextUtils.isEmpty(appId) ? "" : appId);
```
andirayo, line 541: It's not that I feel strongly about it, but this is not really necessary.

How about this compromise:
```
return String.format(
        "Firebase options used while communicating with Firebase server APIs: %s, %s%s",
        apiKey, projectId, TextUtils.isEmpty(appId) ? "" : ", " + appId);
```

PR #1546, opened 2020-05-11
Simplifying cross process lock impl.

ankitaj224:
Also, use BufferedOutputStream to persist firebase installations. BufferedOutputStream is faster than FileOutputStream when writing small chunks of data.


firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -48,8 +48,8 @@ static CrossProcessLock acquire(Context appContext, String lockName) {
       File file = new File(appContext.getFilesDir(), lockName);
       channel = new RandomAccessFile(file, "rw").getChannel();
       // Use the file channel to create a lock on the file.
-      // This method blocks until it can retrieve the lock.
-      lock = channel.lock();
+      // This method does not block and only attempts to acquire the lock.
+      lock = channel.tryLock();
```
vkryachko, line 52: hm iiuc using `tryLock()` here is equivalent to not using locks at all, i.e. it does not given any multiprocess synchronization at all, what is the reasoning behind it?

ankitaj224: I understand. Ideally I want to use lock() with timeout. File Channel doesnt have any methods with explicit timeouts. WDYT about putting the thread to sleep here explicitly?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -64,13 +64,15 @@ static CrossProcessLock acquire(Context appContext, String lockName) {
           lock.release();
         } catch (IOException e2) {
           // nothing to do here
+          Log.e(TAG, "encountered error while releasing the lock, ignoring", e);
```
ankitaj224, line 67: @vkryachko WDYT about these exceptions we ignore? Is it worth propagating them? Even if we throw them and that results in app crashes - Is there anything we can do about it?

vkryachko: I think igrnoring the exceptions should be fine, **but** one thing I found is that there seems to be a resource leak above, i.e. you never `close()` the `RandomAccessFile` which leads to its `fd` to not be release in a timely manner. I wonder if that could cause the file lock not to release properly?

In general I suggest you rewrite all `RandomAccessFile`, `FileChannel` and `FileLock` as one try-with-resources block(it's fully supported at this point).

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
vkryachko: I guess it could work, although (given my comment below) I would suggest you first try to repro by writing multiprocess tests without a fix below and see if you can reproduce the issue, then if it turns out to be the root cause, see if promptly closing the fd fixes the issue.

ankitaj224: I see. Makes sesnse. I will try that. For now reverted back the lock and closed the randomaccessfile fd.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
ankitaj224: Thanks for the suggestion. I changed the code to close RandomAccessFile in finally block.

Reason, we want the FileChannel and FileLock to be closed later by the caller method acquiring the lock. PTAL.

vkryachko: > Reason, we want the FileChannel and FileLock to be closed later by the caller method acquiring the lock. PTAL.

That makes sense, although I would suggest(optional) that you refactor it to make it possible to use `try-with-resources`, e.g. something like this should work:

```
return CrossProcessLock.whileLocked(context, "name", () -> {
  return persistedInstallation.insertOrUpdatePersistedInstallationEntry(prefs);
});
```

this makes sure locking/unlocking is handled inside `whileLocked` as opposed to being up to the caller of `lock()` which could lead to inconsistent behavior and lead to subtle bugs.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
vkryachko: also, have you seen the thread dumps from the IllegalStateException issue? it looks like it is not related to filelocks

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -81,8 +96,15 @@ static CrossProcessLock acquire(Context appContext, String lockName) {
...
-      lock.release();
-      channel.close();
+      if (lock != null) {
+        lock.release();
+      }
+      if (channel != null) {
+        channel.close();
+      }
+      if (randomAccessFile != null) {
+        randomAccessFile.close();
```
vkryachko, line 106: afaict, if any of the above calls throw, the file won't get closed. This is a one more reason to use try-with-resources as doing it manually is prone to bugs

ankitaj224: Updated as per your suggestion. Another look.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
ankitaj224: PTAL. I tried incorporating the changes you recommened.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
-    this.lock = lock;
+  interface CrossProcessPersistedInstallation {
+    PersistedInstallationEntry getPersistedInstallations();
   }
 
   /**
-   * Create and return a lock that is exclusive across processes. If another process has the lock
-   * then this call will block until it is released.
+   * Create a lock that is exclusive across processes. If another process has the lock then this
+   * call will block until it is released.
```
andirayo, line 36: Will there be a time out?
Let's say we only wait on this lock for 1 or 5s?
If we are waiting on a file lock, then 1s should be enough right? ... but even 2, 3, or 5 should be fine.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
+      Context applicationContext,
+      String lockName,
+      CrossProcessPersistedInstallation persistedInstallation) {
+
+    File file = new File(applicationContext.getFilesDir(), lockName);
+    try (RandomAccessFile randomAccessFile = new RandomAccessFile(file, "rw");
+        FileChannel channel = randomAccessFile.getChannel();
+        // Use the file channel to create a lock on the file.
+        // This method blocks until it can retrieve the lock.
+        FileLock lock = channel.lock()) {
```
andirayo, line 52: what happens to this variable `lock`?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
+  static PersistedInstallationEntry whileLocked(
+      Context applicationContext,
+      String lockName,
+      CrossProcessPersistedInstallation persistedInstallation) {
+
+    File file = new File(applicationContext.getFilesDir(), lockName);
+    try (RandomAccessFile randomAccessFile = new RandomAccessFile(file, "rw");
+        FileChannel channel = randomAccessFile.getChannel();
+        // Use the file channel to create a lock on the file.
+        // This method blocks until it can retrieve the lock.
```
andirayo, line 51: can we timeout here?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
 /** Use file locking to acquire a lock that will also block other processes. */
 class CrossProcessLock {
   private static final String TAG = "CrossProcessLock";
-  private final FileChannel channel;
-  private final FileLock lock;
 
-  private CrossProcessLock(FileChannel channel, FileLock lock) {
-    this.channel = channel;
-    this.lock = lock;
+  interface CrossProcessPersistedInstallation {
```
andirayo, line 30: Do we need this interface or can we just use a `Supplier`?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
-      File file = new File(appContext.getFilesDir(), lockName);
-      channel = new RandomAccessFile(file, "rw").getChannel();
-      // Use the file channel to create a lock on the file.
-      // This method blocks until it can retrieve the lock.
-      lock = channel.lock();
-      return new CrossProcessLock(channel, lock);
+  static PersistedInstallationEntry whileLocked(
+      Context applicationContext,
+      String lockName,
+      CrossProcessPersistedInstallation persistedInstallation) {
```
andirayo, line 45: Can we use `Supplier<PersistedInstallationEntry>` here?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
+    PersistedInstallationEntry getPersistedInstallations();
   }
 
   /**
-   * Create and return a lock that is exclusive across processes. If another process has the lock
-   * then this call will block until it is released.
+   * Create a lock that is exclusive across processes. If another process has the lock then this
+   * call will block until it is released.
    *
    * @param lockName the lockname is global across all processes of an app
```
andirayo, line 38: Please add @param for the new parameters.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
-    try {
-      File file = new File(appContext.getFilesDir(), lockName);
-      channel = new RandomAccessFile(file, "rw").getChannel();
-      // Use the file channel to create a lock on the file.
-      // This method blocks until it can retrieve the lock.
-      lock = channel.lock();
-      return new CrossProcessLock(channel, lock);
+  static PersistedInstallationEntry whileLocked(
+      Context applicationContext,
+      String lockName,
```
andirayo, line 44: Why is this a parameter?
Aren't we always using `LOCKFILE_NAME_GENERATE_FID`?
If yes, I recommend
* move the constant `LOCKFILE_NAME_GENERATE_FID` into this class
* remove this parameter and always use this lock file

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
-    FileChannel channel = null;
-    FileLock lock = null;
-    try {
-      File file = new File(appContext.getFilesDir(), lockName);
-      channel = new RandomAccessFile(file, "rw").getChannel();
-      // Use the file channel to create a lock on the file.
-      // This method blocks until it can retrieve the lock.
-      lock = channel.lock();
-      return new CrossProcessLock(channel, lock);
+  static PersistedInstallationEntry whileLocked(
```
andirayo, line 42: 1. why is this method package visible instead of private?
1. `whileLocked` is not a good method name. I recommend to use
    a. `#getLock` - it's pretty common that if a lambda is given it is executed with whatever the method name described. Example `file.open {f -> f.write(...)}`
    b.`supplySynchronized` - the purpose of this method is to run the given Supplier in a synchronized form.
    c. `supplyWithProcessSync` - ...

It would also be great to get rid of all the other parameters, so that it is clear that this method executes the given `Supplier` in a synchronized context and does nothing else.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
```diff
@@ -25,67 +26,40 @@
...
+      Context applicationContext,
+      String lockName,
+      CrossProcessPersistedInstallation persistedInstallation) {
+
+    File file = new File(applicationContext.getFilesDir(), lockName);
```
andirayo, line 47: #### Assumption
We always only lock 1 file and it's always the same file per Firebase Installation.

#### Recommendation
If that assumption is correct, I recommend:
* let's initialize the process lock file once when FIS gets initialized (in the constructor)
* let's get rid of the `applicationContext` parameter.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -403,34 +394,28 @@ private void insertOrUpdatePrefs(PersistedInstallationEntry prefs) {
...
-      }
+      return CrossProcessLock.whileLocked(
+          firebaseApp.getApplicationContext(),
+          LOCKFILE_NAME_GENERATE_FID,
+          this::readLatestPersistedInstallationsMultiProcessSafe);
+    }
+  }
+
+  private PersistedInstallationEntry readLatestPersistedInstallationsMultiProcessSafe() {
+    PersistedInstallationEntry prefs = persistedInstallation.readPersistedInstallationEntryValue();
```
andirayo, line 405: It's fine for this PR, but this is a dependency that violates Single Responsibility Principle.
A class called `CrossProcessLock` should not know anything about persisted installations.
We should segregate these responsibilities, have one class that does nothing but get a CrossProcssLock and/or executes methods synchronized.
Another class should know methods like reading and writing to disk.

The business logic should then just call
{pseudo-code}
`CrossProcessLock.supplySynchronized(PersistedInstallation::readFromDisc)`

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallation.java
```diff
@@ -124,6 +127,7 @@ private JSONObject readJSONFromFile() {
       }
       return new JSONObject(baos.toString());
     } catch (IOException | JSONException e) {
+      Log.e(TAG, "error encountered while reading persisted firebase installation." + e);
```
andirayo, line 130: String.format(Error while reading from disc (%s): %s", filename, exception)

If there is always only 1 file, you can omit the filename.

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallation.java
```diff
@@ -164,6 +169,7 @@ public PersistedInstallationEntry insertOrUpdatePersistedInstallationEntry(
       // There isn't a lot we can do when this happens, other than crash the process. It is a
       // bit nicer to eat the error and hope that the user clears some storage space on their
       // device.
+      Log.e(TAG, "error encountered while persisting firebase installation." + e);
```
andirayo, line 172: String.format(Error while writing to disc (%s): %s", filename, exception)

If there is always only 1 file, you can omit the filename.
Instead of "disc" we could also use "shared pref".

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
andirayo, line 52: Remind me again please:
Why can we use `try(<resource>)` here and couldn't use it before?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
ankitaj224: Supplier Added in API level 24

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
ankitaj224: Supplier Added in API Level 24

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
andirayo: After thinking more about this, I think a good method name is
`#synchonize`

Then when used it should be self-explanatory:
`crossProcessLock.synchronize( () -> someActionThatNeedsToBeSynchronized() );`

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
vkryachko: @andirayo What is the intent of timing out? is if acceptable to proceed if acquiring the lock fails? if so then what is the value of the lock in the first place?

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
vkryachko: good point, as an alternative I'd suggest considering making this method non-static and have the class take these parameters, that way the CrossProcessLock instance can be initialized by fis once and used as `lock.whileLocked(()-> doStuff());`

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
vkryachko: I suggest you make it generic at least on the return parameter(and optionally on the exception thrown to support checked exceptions seamlessly), i.e.

```java
interface Producer<T, E extends Throwable> {
  T produce() throws E;
}

static <T, E extends Throwable> T whileLocked(Producer<T, E> producer) {}
```

firebase-installations/src/main/java/com/google/firebase/installations/local/PersistedInstallation.java
vkryachko: I recommend against logging the exception as a string and instead passing it in separately:
`Log.e(msg, exc)`

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
ankitaj224: Like it. Updated.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
ankitaj224: We couldn't use try-with-resources before because there was an issue when migrating it to google3. That seems to be fixed now.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
fredquintana: I agree, there is no need to timeout here.

firebase-installations/src/main/java/com/google/firebase/installations/CrossProcessLock.java
fredquintana: Same question. Where do lock and channel get released and closed? Does closing the randomAcessFile release and close the others?

Issue #1540, opened 2020-05-11
Crash after calling task.getResult().getToken() - IllegalArgumentException: method POST must have a request body.


grebulon:
### [READ] Step 1: Are you in the right place?
yes

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.6.1
  * Firebase Component: firebase-messaging
  * Component version: 20.1.7

### [REQUIRED] Step 3: Describe the problem

Sometimes, after calling `task.getResult().getToken()` my app crashes with the following stack.
This happens occasionally, usually once a week. After that the app keeps crashing on startup and the only thing that stops it from crashing is to **delete the cache** from Android's app settings.
I suspect it has to do with caching of the token because after cleaning the cache will make the problem goes away (until the next cache refresh)

```
java.lang.IllegalArgumentException: method POST must have a request body.
com.android.okhttp.Request$Builder.method(Request.java:267)
com.android.okhttp.internal.huc.JavaApiConverter.createOkResponseForCacheGet(JavaApiConverter.java:216)
com.android.okhttp.internal.huc.CacheAdapter.get(CacheAdapter.java:53)
com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:316)
com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:483)
com.android.okhttp.internal.huc.HttpURLConnectionImpl.connect(HttpURLConnectionImpl.java:135)
com.android.okhttp.internal.huc.HttpURLConnectionImpl.getOutputStream(HttpURLConnectionImpl.java:266)
com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.getOutputStream(DelegatingHttpsURLConnection.java:219)
com.android.okhttp.internal.huc.HttpsURLConnectionImpl.getOutputStream(HttpsURLConnectionImpl.java:30)
com.google.firebase.installations.remote.FirebaseInstallationServiceClient.writeRequestBodyToOutputStream(com.google.firebase:firebase-installations@@16.3.0:208)
com.google.firebase.installations.remote.FirebaseInstallationServiceClient.writeGenerateAuthTokenRequestBodyToOutputStream(com.google.firebase:firebase-installations@@16.3.0:248)
com.google.firebase.installations.remote.FirebaseInstallationServiceClient.generateAuthToken(com.google.firebase:firebase-installations@@16.3.0:369)
[...]

PR #1555, opened 2020-05-12
Disable FIS network calls to use cached response.

ankitaj224:
Possible fix for  #1540


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -407,6 +407,7 @@ private static void logBadConfigError() {
   private HttpURLConnection openHttpURLConnection(URL url, String apiKey) throws IOException {
     HttpURLConnection httpURLConnection = (HttpURLConnection) url.openConnection();
     httpURLConnection.setConnectTimeout(NETWORK_TIMEOUT_MILLIS);
+    httpURLConnection.setUseCaches(false);
```
andirayo, line 410: Aren't you worried that this might trigger unexpected errors?
Have you tested this already or should we test it maybe?

vkryachko:
@ankitaj224 have we confirmed that this indeed fixes the issue? it's not obvious to me how caching would result in empty request bodies being sent to the server

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
vkryachko: Additionally, are we sure that `POST`s actually respect this value? I would imagine that posts are just never cached

ankitaj224: Since Android 4.4, it uses okhttp for all Http calls. 

setUseCaches is set to true by default. 

The [ResponseCache](https://developer.android.com/reference/java/net/ResponseCache) implementation decides which resources should be cached, and for how long they should be cached. If a request resource cannot be retrieved from the cache, then the protocol handlers will fetch the resource from its original location. The settings for URLConnection#useCaches controls whether the protocol is allowed to use a cached response.

From the user reported stack trace:
`com.android.okhttp.internal.huc.JavaApiConverter.createOkResponseForCacheGet(JavaApiConverter.java:216)
`

And the [javadocs](https://github.com/square/okhttp/blob/ee5cf502889ed896a67dac57662d19bc469bb876/okhttp-android-support/src/main/java/okhttp3/internal/huc/JavaApiConverter.java#L204-L210) of JavaApiConverter.createOkResponseForCacheGet: **Creates an OkHttp Response using the supplied Request and CacheResponse to supply the data** 

Looks like there is response caching internally. 
 
Regarding POST not being cached: 

From the [rfc](https://www.ietf.org/rfc/rfc2616.txt) : Responses to POST method are not cacheable, unless the response includes appropriate Cache-Control or Expires header fields. 

Cache-Control is set to private for getToken response.

@vkryachko Do you see any issue in disabling the cache?

ankitaj224: Also, I think we can benefit from setting Cache-Control request header as  **no-cache**

ankitaj224:
> @ankitaj224 have we confirmed that this indeed fixes the issue? it's not obvious to me how caching would result in empty request bodies being sent to the server

Hi @vkryachko Sadly, no. We couldn't reproduce this issue. However, from my understanding of the APIs we always want a network call to be enforced. So, I assume its safe to disable any sort of caching on FIS APIs.

Also, the developer mentioned that he could test it out on his beta group.

PR #1557, opened 2020-05-13
Fix a multiple-definition linker error in the native API.

kmandrika:
Remove unnecessary API implementation functions.
Fix warnings.


kmandrika:
@googlebot I fixed it.

kmandrika:
Yes, it is no longer necessary

PR #1570, opened 2020-05-18
Completing getId call with the disk reads on the caller thread.  

ankitaj224:
2. Fixing delete API to access persisted FID using cross process safe locks.
3. Minor code cleanup


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -520,7 +509,7 @@ private PersistedInstallationEntry fetchAuthTokenFromServer(
    * storage.
    */
   private Void deleteFirebaseInstallationId() throws FirebaseInstallationsException, IOException {
-    PersistedInstallationEntry entry = persistedInstallation.readPersistedInstallationEntryValue();
+    PersistedInstallationEntry entry = getPrefsWithGeneratedIdMultiProcessSafe();
```
fredquintana, line 512: This will create a new fid and delete it. I think it is better the way it was.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -213,9 +217,11 @@ String getName() {
   @Override
   public Task<String> getId() {
     preConditionChecks();
-    Task<String> task = addGetIdListener();
-    backgroundExecutor.execute(this::doGetId);
-    return task;
+    synchronized (lockGetFid) {
```
fredquintana, line 220: I see that you want to emulate the old behavior of this being serialized on a sinle thread, but I don't know of adding an additional synchronization block is the right way to go.

For one thing, it isn't multi-process safe.

fredquintana: Actually, everything doGetId() does it already thread and process safe, so this doesn't need its own protection, does it?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Reason for this: If there is one thread that is triggering getId or getToken and another triggering delete. If the delete read is not multi-process safe, and tries to read the same data locked by other thread acquiring CrossProcessLock - leads to OverLappingFileLockException

ankitaj224: As discussed offline: created and used cross-process safe read prefs method.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: I missed that local variables are thread safe. PTAL

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -73,6 +73,10 @@
 
   /* used for thread-level synchronization of generating and persisting fids */
   private static final Object lockGenerateFid = new Object();
+
+  /* used for thread-level synchronization of getting a fid. */
```
andirayo: How about this comment:
/** Thread level synchronization for reading FID from disk. */

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -213,9 +217,11 @@ String getName() {
...
   public Task<String> getId() {
     preConditionChecks();
-    Task<String> task = addGetIdListener();
-    backgroundExecutor.execute(this::doGetId);
-    return task;
+    synchronized (lockGetFid) {
+      TaskCompletionSource<String> taskCompletionSource = new TaskCompletionSource<>();
+      taskCompletionSource.trySetResult(doGetId());
+      return taskCompletionSource.getTask();
+    }
```
andirayo: If we do the `cachedFid` solution, then it has to happen outside of the `synchronized` block of course.
The setting of `cachedFid` could happen anywhere where you think it makes sense.
```
if (null != cachedFid) {
  return cachedFid
}
synchronized (...)
```

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -295,16 +288,12 @@ private void triggerOnException(PersistedInstallationEntry prefs, Exception exce
...
-    doRegistrationInternal(false);
-  }
-
-  private final void doGetAuthTokenForceRefresh() {
-    doRegistrationInternal(true);
+  private String doGetId() {
+    PersistedInstallationEntry prefs = getPrefsWithGeneratedIdMultiProcessSafe();
+    // Execute network calls (CreateInstallations) to the FIS Servers on a separate executor
+    // i.e networkExecutor
+    networkExecutor.execute(() -> doNetworkCall(false));
```
andirayo, line 290: This method name is misleading.
How about `#doNetworkCallIfNecessary`?

PR #1571, opened 2020-05-18
Cache FID to avoid multiple disk reads.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -68,6 +68,8 @@
   private final ExecutorService backgroundExecutor;
   private final ExecutorService networkExecutor;
 
+  private String cachedFid = null;
```
andirayo, line 71: please add a comment here, example:
/** FID of this FirebaseInstallations instance. Cached after the first successful read from local storage. */

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -215,7 +217,11 @@ String getName() {
   public Task<String> getId() {
     preConditionChecks();
     TaskCompletionSource<String> taskCompletionSource = new TaskCompletionSource<>();
-    taskCompletionSource.trySetResult(doGetId());
+    if (cachedFid != null) {
+      taskCompletionSource.trySetResult(cachedFid);
+    } else {
```
andirayo, line 222: Do this as first line of `#doGetId()`.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -454,6 +460,7 @@ private PersistedInstallationEntry registerFidWithServer(PersistedInstallationEn
 
     switch (response.getResponseCode()) {
       case OK:
+        cachedFid = response.getFid();
```
andirayo, line 463: I recommend to remove this and only set cachedFid when you successfully read from SharedPrefs.
You solution is probably better because 1 less read of SharedPrefs, but potentially more error cases.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -519,7 +527,7 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
             "Failed to delete a Firebase Installation.", Status.BAD_CONFIG);
       }
     }
-
+    cachedFid = null;
```
andirayo, line 530: Maybe first line of this method?
Worst case we have 1 unnecessary disk read.

ankitaj224:
> I would recommend to do all the logic inside of `#doGetId`:
> 
> ```
>   private String doGetId() {
>     // only read FID once from disk per session. Note: cachedFid will be reset if FID is deleted.
>     if (cachedFid != null) {
>         return cachedFid;
>     }
>     PersistedInstallationEntry prefs = getPrefsWithGeneratedIdMultiProcessSafe();
>     networkExecutor.execute(() -> doNetworkCall(false));
>     
>     if (prefs.status == REGISTERED) {  // pseudo-code
>         cachedFid = prefs.getFirebaseInstallationId();
>     }
>     return prefs.getFirebaseInstallationId();
>   }
> ```

LGTM. PTAL - updated the PR.

PR #1577, opened 2020-05-20
Refactor Crashlytics/Analytics integration

mrwillis21:
- Register with Firebase Analytics at initialization time
- Create separate pipelines for breadcrumbs and Crashlytics origin
events
- Create a facade for a synchronous blocking call to log the
app_exception event
- Create stub implementations for breadcrumbs and app_exception
event logger for when Firebase Analytics is not available


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
```diff
@@ -0,0 +1,86 @@
...
+        Logger.getLogger().d("Interrupted while awaiting app exception callback from FA listener.");
+      }
+
+      this.eventLatch = null;
+    }
+  }
+
+  @Override
+  public void onEvent(@NonNull String name, @NonNull Bundle params) {
+    final CountDownLatch eventLatch = this.eventLatch;
```
mrichards, line 76: I wasn't really sure about this reference...it is a sneaky clever way to guard against a weird race condition that theoretically shouldn't happen. 🙃

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
```diff
@@ -0,0 +1,86 @@
...
+      int timeout,
+      TimeUnit timeUnit) {
+    this.baseAnalyticsEventLogger = baseAnalyticsEventLogger;
+    this.timeout = timeout;
+    this.timeUnit = timeUnit;
+  }
+
+  @Override
+  public void logEvent(@NonNull String name, @Nullable Bundle params) {
+    synchronized (latchLock) {
```
mrichards, line 52: Is it worth making a comment somewhere that we only ever expect at most ONE of these events per app execution? That fact influences the design, though I think the way you have it implemented is resilient to multiple events (though you could end up with a bunch calls waiting on the syncrhonized(latchLock) block, which is fine).

Maybe not necessary since the design works either way?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -61,13 +74,52 @@
       nativeComponent = new MissingNativeComponent();
     }
 
+    final Onboarding onboarding = new Onboarding(app, context, idManager, arbiter);
+
+    final BreadcrumbSource breadcrumbSource;
+    final AnalyticsEventLogger analyticsEventLogger;
+
+    if (analyticsConnector != null) {
```
mrichards, line 82: This might be a good place for a comment describing the overall approach to the analytics integration, or at least point to one of the class javadocs that describes it.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
mrwillis21: Yep! Get a local reference for the function scope, then if the global reference is changed by a separate thread, it won't interfere! :)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
mrwillis21: I think there's definitely the possibility we could end up with multiple queued-up calls to this function, so it may be worth changing this later to allow for sending multiple events at once.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
mrwillis21: Happy to.

PR #1585, opened 2020-05-22
Add per-instance log level API to Remote Config.

danasilver:


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,126 @@
...
+  private final String tag;
+  private int logLevel;
+
+  private ConfigLogger(String tag) {
+    this.tag = tag;
+    this.logLevel = Log.INFO;
+  }
+
+  /** Get the logger for {@param namespace} or create one if it does not already exist. */
+  public static ConfigLogger getLogger(String namespace) {
```
erikeldridge, line 39: Given [RC instances are per-namespace](https://github.com/firebase/firebase-android-sdk/blob/d8f4dc14a58bfe607a455c8c984dd7327f9b6523/firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java#L48), and the logger is per-instance, I don't think we need to maintain a map of namespace-to-logger.

danasilver: I'm trying to think through whether that's necessary too... My thinking is that we need to use the same logger in the various classes injected into the instance of RC (ConfigFetchHttpClient, ConfigCacheClient, etc.) so the same instance of RC always uses the same logger, however those are created before and injected into the RC instance in [RemoteConfigComponent](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java) so they can't get the logger from the RC instance.

What do you think? I'd definitely prefer not needing to maintain the map and get the logger by namespace - would be much cleaner than passing the namespace around RemoteConfigComponent.

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -644,6 +642,21 @@ public void setDefaults(@XmlRes int resourceId) {
         });
   }
 
+  /**
+   * Set the log level for this {@link FirebaseRemoteConfig}'s namespace. Log messages at or above
```
erikeldridge, line 646: I think "instance" is a more accurate differentiation than "namespace". Instances happen to be differentiated by namespace, among other things, but this change adds a per-instance log level. Also, these docs are external-facing, but namespace is an internal detail I wouldn't expect 3p customers to be aware of.

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -644,6 +642,21 @@ public void setDefaults(@XmlRes int resourceId) {
...
+   * {@param logLevel} will be written.
+   *
+   * @param logLevel Log level as specified by {@link android.util.Log}.
+   */
+  public void setLogLevel(int logLevel) {
+    logger.setLogLevel(logLevel);
+  }
+
+  /** Get the log level for this {@link FirebaseRemoteConfig}'s namespace. */
```
erikeldridge, line 655: As above: "Get the log level for this {@link FirebaseRemoteConfig} instance"

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -149,10 +153,12 @@ FirebaseRemoteConfig getDefault() {
   @VisibleForTesting
   @KeepForSdk
   public synchronized FirebaseRemoteConfig get(String namespace) {
-    ConfigCacheClient fetchedCacheClient = getCacheClient(namespace, FETCH_FILE_NAME);
-    ConfigCacheClient activatedCacheClient = getCacheClient(namespace, ACTIVATE_FILE_NAME);
-    ConfigCacheClient defaultsCacheClient = getCacheClient(namespace, DEFAULTS_FILE_NAME);
+    ConfigLogger logger = ConfigLogger.getLogger();
+    ConfigCacheClient fetchedCacheClient = getCacheClient(namespace, FETCH_FILE_NAME, logger);
```
erikeldridge, line 157: Nice!

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,117 @@
...
+  private int logLevel;
+
+  private ConfigLogger(String tag) {
+    this.tag = tag;
+    this.logLevel = Log.INFO;
+  }
+
+  /** Get the logger for {@param namespace} or create one if it does not already exist. */
+  public static ConfigLogger getLogger() {
+    return new ConfigLogger(TAG);
```
erikeldridge, line 37: Given `TAG` is effectively constant, I don't see a reason to parameterize it, and if we don't have any params, I think it's ok to just call `new ConfigLogger()` in a few places.

If we need to know what the tag is in tests, I think it's ok to make it public (esp since we're in the internal package) and visible for testing, or leave it parameterized and just have a no-arg constructor overload.

If a creation method is handy, I think we should name it something that implies creation, eg `ConfigLogger.create()`. `getLogger` seems like `getInstance`, but it isn't a singleton.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,117 @@
...
+
+  private final String tag;
+  private int logLevel;
+
+  private ConfigLogger(String tag) {
+    this.tag = tag;
+    this.logLevel = Log.INFO;
+  }
+
+  /** Get the logger for {@param namespace} or create one if it does not already exist. */
```
erikeldridge, line 35: Stale comment

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,117 @@
...
+  /** Send a {@link Log#WARN} log message and log the exception. */
+  public void w(String text, Throwable throwable) {
+    if (canLog(Log.WARN)) {
+      Log.w(tag, text, throwable);
+    }
+  }
+
+  /** Send a {@link Log#ERROR} log message and log the exception. */
+  public void e(String text, Throwable throwable) {
+    if (canLog(Log.WARN)) {
```
erikeldridge, line 84: Typo: WARN --> ERROR

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigLoggerTest.java
```diff
@@ -0,0 +1,77 @@
...
+  public void setLogLevel_suppressesLogsBelowLevel() {
+    logger.setLogLevel(Log.ERROR);
+    logger.w("Warning");
+
+    List<LogItem> logs = ShadowLog.getLogsForTag(TAG);
+    assertThat(logs).isEmpty();
+  }
+
+  @Test
+  public void getLogLevel_getsSetLogLevel() {
```
erikeldridge, line 72: Awkward phrasing given the level isn't called "SetLogLevel". Something like `getLogLevel_getsCurrentLevel` would be more accurate.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
danasilver: Good call, that simplifies things.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @author danasilver
+ */
+public class ConfigLogger {
+  @VisibleForTesting static final String TAG = "FirebaseRemoteConfig";
+
+  private int logLevel;
+
+  public ConfigLogger() {
+    this.logLevel = Log.INFO;
```
erikeldridge, line 31: Would it makes sense to default to the highest level (`VERBOSE`) since the system level will override? (I see the [system level](https://developer.android.com/studio/debug/am-logcat#level) also defaults to the highest level.)

danasilver: Oh interesting, that's a good find. My intention was to mirror the system default, which I thought was `INFO` based on the [`isLoggable()`](https://developer.android.com/reference/android/util/Log#isLoggable(java.lang.String,%20int)) docs. I tested using `adb logcat` on the command line (rather than Android Studio, which has a dropdown to control the log level, so I'm not sure how its default works) and a verbose message appeared without setting any level.

Makes sense to use `VERBOSE` as the default for the RC logger, too.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigLogger.java
```diff
@@ -0,0 +1,111 @@
...
+  public void w(String text) {
+    w(text, null);
+  }
+
+  /** Send a {@link Log#ERROR} log message. */
+  public void e(String text) {
+    e(text, null);
+  }
+
+  private boolean canLog(int level) {
```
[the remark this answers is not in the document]
danasilver: I was approaching this with more with the goal of reducing log spam, primarily `WARNING` level logs [when no value exists for a key](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigGetParameterHandler.java#L412), especially when RC is used indirectly via the Performance SDK.

Since we currently use the Android [`Log`](https://developer.android.com/reference/android/util/Log) API rather than `java.util.logging` (same as the other Firebase SDKs) I'd bias towards keeping with Android Log and its levels rather than using`java.util.logging` or mixing the Android logger with Java log levels that don't match in name. I think it's a good point that the level is just an `int`, so `Integer.MAX_VALUE` can be used to silence all logging.

PR #1591, opened 2020-05-25
Initial implementation of FieldDescriptor.

vkryachko:


encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
```diff
@@ -0,0 +1,132 @@
...
+    if (!(o instanceof FieldDescriptor)) {
+      return false;
+    }
+
+    FieldDescriptor that = (FieldDescriptor) o;
+
+    if (!name.equals(that.name)) {
+      return false;
+    }
+    return properties.equals(that.properties);
```
rlazo, line 89:
```suggestion
    return name.equals(that.name) &&
        properties.equals(that.properties);
```

PR #1594, opened 2020-05-26
Initial implementation of ExtraProperty annotation processor.

vkryachko:


encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/annotations/AnnotationBuilder.java
```diff
@@ -0,0 +1,237 @@
...
+ * Generates an annotation implementation.
+ *
+ * <p>For example, given the following annotation
+ * <pre>{@code
+ * @interface MyAnnotation {
+ *   int value();
+ *   String str() default "hello";
+ * }
+ * }</pre>
+ * The produced code will look like:
```
rlazo, line 48: Would it be useful to document that the `*Impl` is also generated, not only the `At*`?

vkryachko: done

PR #1602, opened 2020-05-28
Add ability to publish JAR SDKs

vkryachko:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/Tasks.kt
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.gradle.plugins
+
+const val JAVADOC_TASK_NAME = "firebaseJavadoc"
```
rlazo, line 17: where is this used?

vkryachko: Looks like it's only needed in the release repo, will move it there.

PR #1606, opened 2020-05-31
Changed TypeTokenContainer from interface to class

James201311:
- Changed `TypeTokenContainer` from an interface to a class.
- Changed field `typeArguments` of type `TypeTokenContainer` in non-generic `ClassToken` from null to an empty instance. 
- Changed `TypeToken` to throw an exception when a non-supported type `<? super Foo>` is passed to create a `TypeToken`, instead of throwing an exception only when trying to get the type arguments from it.

### DIFF
--- a/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
+++ b/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
@@ -53,7 +53,8 @@ public abstract class TypeToken<T> {
     if (superclass instanceof Class) {
       throw new IllegalArgumentException("Missing type parameters");
     }
-    @SuppressWarnings("ConstantConditions") // Safe because superclass is not instanceof Class
+    // Safe because superclass is not instanceof Class
+    @SuppressWarnings("ConstantConditions")
     Type type = ((ParameterizedType) superclass).getActualTypeArguments()[0];
     return of(type);
   }
@@ -84,8 +85,8 @@ public abstract class TypeToken<T> {
       return new ArrayToken<T>(TypeToken.of(componentType));
     } else if (type instanceof ParameterizedType) {
       ParameterizedType parameterizedType = (ParameterizedType) type;
-      @SuppressWarnings(
-          "unchecked") // Safe because rawType of parameterizedType is always instance of Class<T>
+      // Safe because rawType of parameterizedType is always instance of Class<T>
+      @SuppressWarnings("unchecked")
       Class<T> rawType = (Class<T>) parameterizedType.getRawType();
       Type[] types = parameterizedType.getActualTypeArguments();
       TypeToken[] typeTokens = new TypeToken[types.length];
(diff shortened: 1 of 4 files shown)


encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -118,7 +110,7 @@ private TypeToken() {}
 
     private ClassToken(Class<T> token) {
       this.rawType = token;
-      this.typeArguments = null;
+      this.typeArguments = new TypeTokenContainer(null);
```
vkryachko, line 113: suggestion: Since `TypeTokenContainer` is immutable, let's have a singleton `TypeTokenContainer .EMPTY` instance that can be shared by all non-generic classes as opposed to allocating one each time.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -15,13 +15,33 @@
...
 /**
  * {@link TypeTokenContainer} is used to get actual type parameter in a generic class at given
  * index.
  */
-// TODO: change from interface to class
-public interface TypeTokenContainer {
+public class TypeTokenContainer {
+  private final TypeToken<?>[] typeTokens;
+
+  TypeTokenContainer(@Nullable Type[] types) {
```
vkryachko, line 28: Consider moving the conversion logic `Type->TypeToken` into the TypeToken class itself, this would increase cohesion of this class and will not require making `TypeToken#of(Type)` non-private, effectively exposing its implementation details.

vkryachko: Also might make sense to make the constructor public and class `final`.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -15,13 +15,33 @@
...
+      }
+    }
+    this.typeTokens = typeTokens;
+  }
+
   @NonNull
-  <T> TypeToken<T> at(int index);
+  public <T> TypeToken<T> at(int index) {
+    if (index >= typeTokens.length)
+      throw new NullPointerException("No type token at index: " + index);
```
vkryachko, line 44: I think `IllegalArgumentException` is more appropriate here.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -15,13 +15,33 @@
...
+        typeTokens[i] = TypeToken.of(types[i]);
+      }
+    }
+    this.typeTokens = typeTokens;
+  }
+
   @NonNull
-  <T> TypeToken<T> at(int index);
+  public <T> TypeToken<T> at(int index) {
+    if (index >= typeTokens.length)
```
vkryachko, line 43: nit: what if the index is less than zero?

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -15,13 +15,33 @@
...
+      throw new NullPointerException("No type token at index: " + index);
+    return (TypeToken<T>) typeTokens[index];
```
vkryachko, line 45: Doesn't this produce an unchecked cast warning? if so worth adding a `@SuppressWarnings`?

PR #1609, opened 2020-06-01
Avoid throwing explcit IOExceptions.

ankitaj224:
Also, handle slow network connection issues by retrying and back off.


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -444,7 +444,7 @@ private String readExistingIidOrCreateFid(PersistedInstallationEntry prefs) {
 
   /** Registers the created Fid with FIS servers and update the persisted state. */
   private PersistedInstallationEntry registerFidWithServer(PersistedInstallationEntry prefs)
-      throws IOException {
+      throws FirebaseInstallationsException, IOException {
```
andirayo, line 447: Can we just throw 1 exception type here?
Or where happens the unification (of exception types) for users?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -527,7 +532,7 @@ private Void deleteFirebaseInstallationId() throws FirebaseInstallationsExceptio
             /*projectID= */ getProjectIdentifier(),
             /*refreshToken= */ entry.getRefreshToken());
 
-      } catch (FirebaseException exception) {
+      } catch (FirebaseException | IOException exception) {
```
andirayo, line 535: Why do we catch `FirebaseException` here?

a) Do we ever use any Exception type other than `FirebaseInstallationsException`?
If yes: why? If no: I recommend to check for `FirebaseInstallationsException` here.
b) Why do we catch `FirebaseInstallationsException`?
catching and rethrowing the same exception is an antipattern.
Can we not just leave the FirebaseInstallationsException as is?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallationsException.java
```diff
@@ -29,6 +29,11 @@
      * Project.
      */
     BAD_CONFIG,
+    /**
+     * The service is currently unavailable. This is a most likely a transient condition and may be
+     * corrected by retrying with a backoff.
```
andirayo, line 34: What "backoff" are you referring to?
Exponential backoff?

If this is a public comment, I think I would either omit this addition or frame it like this:
"This is a most likely due to a transient condition and may be corrected by retrying. We recommend exponential backoff when retrying requests."

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -138,7 +138,7 @@ public InstallationResponse createFirebaseInstallation(
       @NonNull String projectID,
       @NonNull String appId,
       @Nullable String iidToken)
-      throws IOException {
+      throws FirebaseInstallationsException, IOException {
```
andirayo, line 141: Why are we still throwing `IOException`?
Can we get around this and only throw 1 exception type?
See comment below.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -179,12 +179,16 @@ public InstallationResponse createFirebaseInstallation(
 
         // Return empty installation response with BAD_CONFIG response code after max retries
         return InstallationResponse.builder().setResponseCode(ResponseCode.BAD_CONFIG).build();
+      } catch (IOException ignored) {
+        retryCount++;
       } finally {
         httpURLConnection.disconnect();
```
andirayo, line 185: If we catch and ignore `IOException` here, can we remove `IOException` from the method interface?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -352,7 +360,7 @@ public TokenResult generateAuthToken(
       @NonNull String fid,
       @NonNull String projectID,
       @NonNull String refreshToken)
-      throws IOException {
+      throws FirebaseInstallationsException, IOException {
```
andirayo, line 363: Same question as before.
Btw, just as a comment, no action item for this PR: quite a bit of code duplication going on here.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -407,27 +419,37 @@ private static void logBadConfigError() {
...
-    if (heartbeatInfo != null && userAgentPublisher != null) {
-      HeartBeat heartbeat = heartbeatInfo.getHeartBeatCode(FIREBASE_INSTALLATIONS_ID_HEARTBEAT_TAG);
-      if (heartbeat != HeartBeat.NONE) {
-        httpURLConnection.addRequestProperty(USER_AGENT_HEADER, userAgentPublisher.getUserAgent());
-        httpURLConnection.addRequestProperty(
-            HEART_BEAT_HEADER, Integer.toString(heartbeat.getCode()));
+  private HttpURLConnection openHttpURLConnection(URL url, String apiKey)
+      throws FirebaseInstallationsException {
+    try {
+      HttpURLConnection httpURLConnection = (HttpURLConnection) url.openConnection();
```
andirayo, line 425: is it maybe enough to wrap this line with the try-catch block?
Maybe it gives more clarity about what can go wrong here.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: I will address the code duplication in another PR. Updated the exception handling.  PTAL. Thanks

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Update to check only `FirebaseInstallationsException`.

Agreed on the anti-pattern. Updated the code to remove this code.

I don't know the reason why we had this anti-pattern behavior before.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -407,8 +412,16 @@ private static void logBadConfigError() {
             + " initializing Firebase.");
   }
 
-  private HttpURLConnection openHttpURLConnection(URL url, String apiKey) throws IOException {
-    HttpURLConnection httpURLConnection = (HttpURLConnection) url.openConnection();
+  private HttpURLConnection openHttpURLConnection(URL url, String apiKey)
+      throws FirebaseInstallationsException {
+    HttpURLConnection httpURLConnection;
+    try {
+      httpURLConnection = (HttpURLConnection) url.openConnection();
```
andirayo, line 419: Why are you casting this URLConnection?
Are you using methods that are specific to HttpURLConnection?
If not, I strongly recommend programming against the interface than the implementation.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -321,12 +313,27 @@ public void deleteFirebaseInstallation(
...
       }
     }
 
-    throw new IOException();
+    throw new FirebaseInstallationsException(
+        "Firebase Installations Service is unavailable. Please try again later.",
+        Status.UNAVAILABLE);
+  }
+
+  private URL getURL(String resourceName) throws FirebaseInstallationsException {
```
andirayo, line 328: +1 for adding this method!
An idea for a different name (feel free to ignore):
#getFullyQualifiedRequestUri
#getFullyQualifiedApiEndpoint

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -321,12 +313,27 @@ public void deleteFirebaseInstallation(
...
+    try {
+      return new URL(
+          String.format(
+              "https://%s/%s/%s",
+              FIREBASE_INSTALLATIONS_API_DOMAIN, FIREBASE_INSTALLATIONS_API_VERSION, resourceName));
+    } catch (MalformedURLException e) {
+      throw new FirebaseInstallationsException(e.getMessage(), Status.UNAVAILABLE);
```
andirayo, line 335: This seems to be a pretty dramatic error.
How about we add some message here, e.g.
"Unexpected internal error while creating request URI."
I don't know.
Feel free to ignore.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Yes, I am using methods specific to HttpURLConnection

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallationsException.java
```diff
@@ -29,6 +29,12 @@
      * Project.
      */
     BAD_CONFIG,
+
+    /**
+     * The service is currently unavailable. This is a most likely due to a transient condition and
+     * may be corrected by retrying. We recommend exponential backoff when retrying requests."
```
ciarand, line 35: extra quote?

PR #1612, opened 2020-06-02
implemented hashcode, equals, toString methods for TypeToken in decoders

James201311:


encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -41,4 +42,48 @@ public TypeTokenContainer(@NonNull TypeToken[] typeTokens) {
     TypeToken<T> typeToken = (TypeToken<T>) typeTokens[index];
     return typeToken;
   }
+
+  @Override
+  public int hashCode() {
```
vkryachko, line 47: I think this whole method could be just one line:
```java
return Arrays.hashCode(typeTokens);
```
wdyt?

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -41,4 +42,48 @@ public TypeTokenContainer(@NonNull TypeToken[] typeTokens) {
...
+      return false;
+    }
+    TypeTokenContainer that = (TypeTokenContainer) o;
+    if (this.typeTokens.length != that.typeTokens.length) {
+      return false;
+    } else {
+      for (int i = 0; i < typeTokens.length; i++) {
+        if (!this.typeTokens[i].equals(that.typeTokens[i])) return false;
+      }
+      return true;
```
vkryachko, line 70:
```suggestion
    return Arrays.equals(typeTokens, that.typeTokens);
```

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
```diff
@@ -41,4 +42,48 @@ public TypeTokenContainer(@NonNull TypeToken[] typeTokens) {
...
+    }
+  }
+
+  @NonNull
+  @Override
+  public String toString() {
```
vkryachko, line 76: I don't think this one needs a `toString()`

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -157,5 +186,34 @@ private ArrayToken(TypeToken<?> componentType) {
...
+
+    @NonNull
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder();
+      sb.append("ArrayToken{ ");
+      sb.append("ComponentType: ").append(componentType);
+      sb.append(" }");
+      return sb.toString();
+    }
```
vkryachko, line 217: Then this one could be `${componentType}[]`, wdyt?

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -140,6 +141,34 @@ private ClassToken(Class<T> rawType, TypeTokenContainer typeArguments) {
...
+
+    @NonNull
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder();
+      sb.append("ClassToken{ ");
+      sb.append("RawType: ").append(rawType).append(", ");
+      sb.append("TypeArguments: ").append(typeArguments).append("");
+      sb.append(" }");
+      return sb.toString();
```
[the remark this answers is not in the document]
vkryachko: Or to disambiguate, could it be something like `TypeToken{Optionsl<String>[]`?

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
James201311: Good call!!

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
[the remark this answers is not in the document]
James201311: Great! I'll do that.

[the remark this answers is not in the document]
James201311: It turns out that i can do `TypeToken{TypeToken{Optional<TypeToken{String}>}[]}` or `Optional<String>[]` without any efforts. However, If we want `TypeToken{Optional<String>[]}`, it would involve more code since TypeTokens are nested. wdyt? @vkryachko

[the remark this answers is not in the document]
vkryachko: yeah, that's expected. But it seems that it's pretty straightforward to achieve, i.e. by adding an abstract method like `toStringRaw()` to the type token and have all subclasses implement it, while `toString()` itself is only needed in `TypeToken`, wdyt?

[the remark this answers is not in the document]
James201311: Niceeeeee! I am impressed.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -113,6 +112,15 @@
 
   private TypeToken() {}
 
+  @NonNull
+  abstract String getTypeTokenLiteral();
+
+  @NonNull
+  @Override
+  public String toString() {
+    return "TypeToken:{" + getTypeTokenLiteral() + "}";
```
vkryachko, line 121: nit 
```suggestion
    return "TypeToken{" + getTypeTokenLiteral() + "}";
```

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeTokenContainer.java
vkryachko: I think this one still applies

PR #1613, opened 2020-06-03
Synchronize cached fid

ankitaj224:
Also, contains minor fixes suggested by Ciaran.


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -68,7 +68,7 @@
   private final ExecutorService networkExecutor;
   /* FID of this Firebase Installations instance. Cached after successfully registering and
   persisting the FID locally. NOTE: cachedFid resets if FID is deleted.*/
-  private String cachedFid = null;
+  private String cachedFid;
```
ciarand, line 71: might as well mark this @GuardedBy too

ankitaj224: Done.

PR #1617, opened 2020-06-04
Add a deprecated setter for the data collection

VinayGuthal:


firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -389,6 +389,23 @@ public void setDataCollectionDefaultEnabled(Boolean enabled) {
...
+   * specific mechanisms.
+   *
+   * <p>This api has been deprecated. Please use the api which takes in a Java boolean
+   *
+   * @hide
+   */
+  @KeepForSdk
+  @Deprecated
+  public void setDataCollectionDefaultEnabled(boolean enabled) {
+    checkNotDeleted();
```
vkryachko, line 405: consider remoging this line, since the other method already calls `checkNotDeleted()`

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -389,6 +389,23 @@ public void setDataCollectionDefaultEnabled(Boolean enabled) {
...
   }
 
+  /**
+   * Enable or disable automatic data collection across all SDKs.
+   *
+   * <p>Note: this value is respected by all SDKs unless overridden by the developer via SDK
+   * specific mechanisms.
+   *
+   * <p>This api has been deprecated. Please use the api which takes in a Java boolean
+   *
```
vkryachko, line 399:
```suggestion
   *
   * @deprecated Use {@link #setDataCollectionDefaultEnabled(Boolean)} instead.
```

PR #1623, opened 2020-06-05
Increase FIS Network Timeout.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -86,7 +86,7 @@
 
   private static final String API_KEY_HEADER = "x-goog-api-key";
 
-  private static final int NETWORK_TIMEOUT_MILLIS = 10000;
+  private static final int NETWORK_TIMEOUT_MILLIS = 30000;
```
ciarand, line 89: should this have a comment explaining why 30s is the right number and 10s is too low?

ankitaj224: Done. PTAL.

Also, what's the recommended Timeout settings in Android? I couldn't find any guidelines around that. I see numbers varying from 30s to 60s.

ciarand: I don't know if there's a central piece of advice for this particular problem. It also might be better to just retry the connection after 10s instead of waiting a full 30s for it to timeout. Or by checking the current network type and adjusting the timeout based on that (i.e. an "edge" network would have a longer timeout than a 4g one). Is there a specific bug you're trying to solve with this that has more info?

PR #1624, opened 2020-06-05
init implementation for decoder

James201311:


encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -0,0 +1,119 @@
...
+       * reader.endArray();
+       * return l.toArray();
+       */
+    }
+    return null;
+  }
+
+  private <T> T decodeObject(TypeToken.ClassToken<T> classToken) throws IOException {
+    if(classToken.getRawType().isPrimitive() || classToken.getRawType().equals(String.class))
+      return decodePrimitives(classToken.getRawType());
```
vkryachko, line 81: nit: at some later point we should think of how to avoid auto-boxing for primitives, i.e. decodeObject returns `T` which will automatically auto-box any primitive type like `int` in a `Integer` only to unpack it back when it's injected into the class that is being decoded. So this is an optimization opportunity that can be saved for later.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -0,0 +1,119 @@
...
+//TODO: remove Exception from method header
+//TODO: surpass warnings
+//TODO: Java Docs
+public class JsonDataDecoderBuilderContext implements DataDecoder {
+  private Map<TypeToken.ClassToken<?>, ObjectDecoderContextImpl<?>> objectDecodersCtx = new HashMap<>();
+  private Map<TypeToken.ClassToken<?>, TypeCreator<?>> typeCreators = new HashMap<>();
+  private JsonReader reader;
+
+  public JsonDataDecoderBuilderContext(@NonNull Map<TypeToken.ClassToken<?>, ObjectDecoder<?>> objectDecoders) {
+    for (Map.Entry<TypeToken.ClassToken<?>, ObjectDecoder<?>> entry : objectDecoders.entrySet()) {
```
vkryachko, line 41: Ideally we should make this processing only on demand, as doing this upfront may be wasteful if most registered decoders are never needed.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+    reader.endObject();
+
+    TypeCreator<T> creator = (TypeCreator<T>) typeCreators.get(classToken);
+    return (T) creator.create(creationCtx);
+  }
+
+  private <T> T decodePrimitives(Class<T> clazz) throws IOException {
+    //TODO: try catch
+    if (clazz.equals(Boolean.class)) {
```
vkryachko, line 106: nit: `Boolean.class` is not primitive, `boolean.class` is. same for `Integer` vs `int`, etc.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/DataDecoder.java
```diff
@@ -0,0 +1,27 @@
...
+import androidx.annotation.NonNull;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
+
+public interface DataDecoder {
+  //TODO: remove Exception from method header
+  @NonNull
+  <T> T decode(@NonNull InputStream input, @NonNull TypeToken<T> typeToken) throws IOException;
```
vkryachko, line 26: nit: we may want to swap the argument order to be consistent with `DataEncoder`

encoders/firebase-decoders-json/src/test/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContextTest.java
```diff
@@ -0,0 +1,137 @@
...
+      FieldRef.Primitive<Integer> aField = ctx.decodeInteger(FieldDescriptor.of("a"));
+      FieldRef.Primitive<Boolean> bField = ctx.decodeBoolean(FieldDescriptor.of("b"));
+      FieldRef.Boxed<T> tField = ctx.decode(FieldDescriptor.of("t"), (TypeToken.ClassToken) ctx.getTypeArguments().at(0));
+      FieldRef.Boxed<SubFoo<String>> subFooField = ctx.decode(FieldDescriptor.of("subFoo"), TypeToken.of(new Safe<SubFoo<String>>(){}));
+
+      return (creationCtx ->
+              new Foo<T>(
+                      creationCtx.getInteger(aField),
+                      creationCtx.getBoolean(bField),
+                      (T) creationCtx.get(tField),
```
vkryachko, line 89: curious why you had to cast to `(T)` here?

encoders/firebase-decoders-json/src/test/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContextTest.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  static class SubFooObjectDecoder<T> implements ObjectDecoder<SubFoo<T>> {
+
+    @NonNull
+    @Override
+    public TypeCreator<SubFoo<T>> decode(@NonNull ObjectDecoderContext<SubFoo<T>> ctx) {
+      FieldRef.Primitive<Integer> aField = ctx.decodeInteger(FieldDescriptor.of("a"));
+      FieldRef.Primitive<Boolean> bField = ctx.decodeBoolean(FieldDescriptor.of("b"));
+      FieldRef.Boxed<T> tField = ctx.decode(FieldDescriptor.of("t"), (TypeToken.ClassToken) ctx.getTypeArguments().at(0));
```
vkryachko, line 103: Did you have to cast to `(TypeToken.ClassToken)` here? I think the types should align correctly without the cast.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldRef.java
```diff
@@ -0,0 +1,72 @@
...
+  @NonNull
+  public String getName() {
+    return name;
+  }
+
+  protected FieldRef(@NonNull String name) {
+    this.name = name;
+  }
+
+  public static class Primitive<T> extends FieldRef<T> {
```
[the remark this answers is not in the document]
James201311: Each `ObjectDecoder` keeps its decoding **logics** into a `ObjectDecoderContextIml` where each field name corresponds to a `FieldRef`. The decoding **logic** inside `ObjectDecoderContextIml` is simply an container `Map<String, FieldRef<?>> refs` which refers each field name and its corresponding `FieldRef`.

Each `ObjectDecoder` keeps its decoding **value** from given `inputStream` into a `CreationContext`. The decoding **values** inside `CreationContext` is also an container `Map<FieldRef<?>, Object> ctx` which contains all the values the FieldRef are refering to.

Therefore, whenever we are given the new `InputStream`, We wouldn't create `ObjectDecoderContextIml` again, but only update the values each `FieldRef` pointing to.

The reason we need a `rawType` inside `Primitive<T>` or `ClassToken` inside `Boxed<T>` is that:
each `ObjectDecoderContextIml` should not only knows where each field is referring to but also its anticipated type. Therefore, I nested their type information into `FiedRef`, getting these types information make it possible to determine which method to use when looping the json data, such as `nextInt(), nextBoolean, etc.` 

The reason I make `Primitive` and `Boxed` extennd `FieldRef` is that
no matter they are primitive or object, they are field references. Besides, I want to make the containers inside  `ObjectDecoderContextIml` and `CreationContext` be simple `Map<String, FieldRef<?>>`

@vkryachko 
wdyt? are there any better ways to do it?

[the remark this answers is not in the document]
vkryachko: > The reason we need a rawType inside Primitive<T> or ClassToken inside Boxed<T>

I see that makes sense, I would suggest though that both store a typeToken exposed through the `FieldRef`'s abstract `getTypeToken()` so that you don't have to cast in your implementation code.

> The reason I make Primitive and Boxed extennd FieldRef is that no matter they are primitive or object, they are field references

sgtm

> CreationContext is also an container Map<FieldRef<?>, Object> ctx which contains all the values
This would be a performance challenge later as we don't want to "auto-box" primitives

> We wouldn't create ObjectDecoderContextIml again, but only update the values each FieldRef pointing to

Let's make sure this approach works with recursively defined types and does not overwrite things in unexpected ways, i.e. the type like the following might have a problem:

```java
class Foo {
  String field;
  @Nullable Foo foo;
```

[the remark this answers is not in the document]
James201311: That is a very good call!
Actually it will take care of these recursively defined types well after I enable it to read `null` since each object will have its own `CreationContext` and it won't overwrite things.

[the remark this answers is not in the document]
James201311: Also, `TypeToken` didn't support `TypeVariable` right now, but in order to support decode nested Generic type it should. I will have a PR for that first, then come back to support the recursively defined type.

encoders/firebase-decoders-json/src/test/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContextTest.java
James201311: since generic type `T` in `Box<T>` is downgrade to `Object`, then `creationCtx.get(tField)` give us an `Object` types, therefore i have to explicitly cast it to `T`. 

I will have more investigations on it to see how can I avoid explicitly casting it.

vkryachko: At runtime sure, but it should not matter at compile time where it is `T` which is exactly what `Foo`'s constructor expects

vkryachko: Oh, I think I see the problem, you need to remove `<T>` on `CreationContext<T>` and make it just `CreationContext`, without this the `TypeCreator` interface is defined with a raw type, hence all generics inference gets disabled and `.get()` return `Object` instead of `T`.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/CreationContextImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    ctx.put(ref, val);
+  }
+
+  void put(@NonNull FieldRef<?> ref, double val) {
+    ctx.put(ref, val);
+  }
+
+  void put(@NonNull FieldRef<?> ref, boolean val) {
+    ctx.put(ref, val);
+  }
```
vkryachko, line 55: Since these auto-box anyway, consider removing these methods altogether for now.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldRef.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public abstract class FieldRef<T> {
+  private TypeToken<T> typeToken;
+
+  @NonNull
+  public abstract TypeToken<T> getTypeToken();
+
+  /** Used to represent primitive data type. */
+  public static final class Primitive<T> extends FieldRef<T> {
+    @NonNull public static Primitive<Integer> INT = new Primitive<>(TypeToken.of(int.class));
```
vkryachko, line 34: nit, pls make them all final, otherwise anyone can reassign them.

vkryachko, line 34: Additionally, I'd suggest you move these constants out into `FieldRef` itself, and also move `Boxed#of()` into FieldRef as sell. It will make the public interface to `FieldRefs` less hierarchical which might make it simpler, i.e.:

```java
class FieldRef<T> {
  static Primitive<Boolean> BOOLEAN = //...;
  static Primitive<Short> SHORT = //...;
  // etc,
  static Boxed<T> of(TypeToken<T> token) {}
}
```

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldRef.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /** Use it to represent Boxed Data type */
+  public static final class Boxed<T> extends FieldRef<T> {
+    @NonNull
+    public static <T> Boxed<T> of(@NonNull TypeToken<T> typeToken) {
+      return new Boxed<T>(typeToken);
+    }
+
+    private Boxed(@NonNull TypeToken<T> typeToken) {
+      super.typeToken = typeToken;
```
vkryachko, line 68: let's make a precondition check that typeToken is not in fact primitive and throw `IllegalArgumentException` if it is, wdyt?

James201311: I agree! this way we can ensure `Boxed` only contains non-primitive types and force to create `Primitive` by a pre-defined constant value.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldRef.java
```diff
@@ -29,22 +29,29 @@
...
 
-    @NonNull public static Primitive<Character> CHAR = new Primitive<>(TypeToken.of(char.class));
+  @NonNull
+  public static <T> Boxed<T> of(@NonNull TypeToken<T> typeToken) {
+    return new Boxed<T>(typeToken);
+  }
+
+  /** Used to represent primitive data type. */
+  public static final class Primitive<T> extends FieldRef<T> {
+    @NonNull public static Primitive<Integer> INT = new Primitive<>(TypeToken.of(int.class));
```
vkryachko, line 54: looks like you forgot to move `INT`

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldRef.java
vkryachko: 👍 can you also add a test for it?

PR #1628, opened 2020-06-08
Refetch token if attempt was invalidated

schmidt-sebastian:
Port of https://github.com/firebase/firebase-js-sdk/pull/3184

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
@@ -17,13 +17,14 @@ package com.google.firebase.firestore.auth;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.android.gms.tasks.Task;
+import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.auth.GetTokenResult;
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -85,19 +88,21 @@ public FirebaseAuthCredentialsProvider(InternalAuthProvider authProvider) {
...
+        Executors.DIRECT_EXECUTOR,
         task -> {
           synchronized (this) {
             // Cancel the request since the token changed while the request was outstanding so the
             // response is potentially for a previous user (which user, we can't be sure).
             if (savedCounter != tokenCounter) {
-              throw new FirebaseFirestoreException(
-                  "getToken aborted due to token change", Code.ABORTED);
+              Logger.debug(LOG_TAG, "getToken aborted due to token change");
+              return getToken();
```
var-const, line 99: General observation: it seems a little unfortunate that `getToken` and `getResult().getToken()` look so similar while having a very different effect. In general, I'd expect a `get`-prefixed function to be a simple getter, so it might be better to rename `getToken` to something like `fetchToken` to better convey its semantics. Of course, this is absolutely out of scope for this PR.

schmidt-sebastian: Ack :)

PR #1629, opened 2020-06-08
Have fiamd-ktx depend on fiam-ktx.

vkryachko:
This makes fiam-ktx symbols available for developers without having them
to directly depend on both fiam-ktx and fiamd-ktx.


firebase-inappmessaging-display/ktx/ktx.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     implementation project(':firebase-components')
     implementation project(':firebase-common:ktx')
     implementation project(':firebase-inappmessaging')
```
rosariopfernandes, line 51: Maybe we could remove fiam, considering that fiam-ktx already has a dependency on it?

PR #1644, opened 2020-06-11
feat(storage-ktx): destructure ListResult and TaskSnapshots

rosariopfernandes:
This is actually a **Feature Request**. It should add [Kotlin destructuring declarations](https://kotlinlang.org/docs/reference/multi-declarations.html) to the classes: `FileDownloadTask.TaskSnapshot`, `StreamDownloadTask.TaskSnapshot`, `UploadTask.TaskSnapshot` and `ListResult`.

With the destructuring declarations, instead of:
```kotlin
fileDownloadTask.addOnSuccessListener { taskSnapshot ->
    val bytesTransferred = taskSnapshot.bytesTransferred
    val totalCount = taskSnapshot.totalByteCount
    val progress = bytesTransferred * 100 / totalCount
}
```
Kotlin developers will be able to use:
```kotlin
fileDownloadTask.addOnSuccessListener { (bytesTransferred, totalCount) ->
    val progress = bytesTransferred * 100 / totalCount
}
```

Here's a full list of all the declarations added in this PR:
<table>
<tr>
  <th>Class</th>
  <th>Destructured</th>
</tr>
<tr>
  <td>FileDownloadTask.TaskSnapshot</td>
  <td>(bytesTransferred, totalByteCount)</td>
</tr>
<tr>
  <td>StreamDownloadTask.TaskSnapshot</td>
  <td>(bytesTransferred, totalByteCount, stream)</td>
</tr>
<tr>
  <td>UploadTask.TaskSnapshot</td>
  <td>(bytesTransferred, totalByteCount, metadata, uploadSessionUri)</td>
</tr>

<tr>
  <td>ListResult</td>
  <td>(items, prefixes, pageToken)</td>
</tr>
</table>

I should also point out that these arguments are optional, so for example, if a developer wishes to only get `ListResult.items` and not `ListResult.prefixes` or `ListResult.pageToken`, he may omit them:
```kotlin
storageRef.listAll().addOnSuccessListener { (items) ->
    // handle results
}
```


rlazo:
Hi @rosariopfernandes thanks for the proposal. I like the idea of adding destructure here! We'll evaluate this through our internal API review process. I'll keep you posted.

firebase-storage/ktx/src/main/kotlin/com/google/firebase/storage/ktx/Storage.kt
```diff
@@ -45,6 +50,90 @@ fun storageMetadata(init: StorageMetadata.Builder.() -> Unit): StorageMetadata {
...
+ * @return the totalByteCount of the [FileDownloadTask.TaskSnapshot]
+ */
+operator fun FileDownloadTask.TaskSnapshot.component2() = totalByteCount
+
+/**
+ * Destructuring declaration for [ListResult] to provide its items
+ *
+ * @return the items of the [ListResult]
+ */
+operator fun ListResult.component1(): MutableList<StorageReference> = items
```
vkryachko, line 121: Should the developers be able to mutate the list or can we return just `List<StorageReference>` here?
cc @schmidt-sebastian

firebase-storage/ktx/src/main/kotlin/com/google/firebase/storage/ktx/Storage.kt
```diff
@@ -45,6 +50,90 @@ fun storageMetadata(init: StorageMetadata.Builder.() -> Unit): StorageMetadata {
...
+
+/**
+ * Destructuring declaration for [ListResult] to provide its prefixes
+ *
+ * @return the prefixes of the [ListResult]
+ */
+operator fun ListResult.component2(): MutableList<StorageReference> = prefixes
```
vkryachko, line 128: same question here?

firebase-storage/ktx/src/main/kotlin/com/google/firebase/storage/ktx/Storage.kt
rosariopfernandes: @vkryachko Seems like you're right. The java code does return an immutable list:

https://github.com/firebase/firebase-android-sdk/blob/334aa44e0f7a3418dbe5cb0da3dcc0abd442c967/firebase-storage/src/main/java/com/google/firebase/storage/ListResult.java#L86-L89

The `MutableList` return type was  mistakenly added by the IDE when auto-fixing this lint warning:

Screenshot from 2020-07-13 17-22-57

schmidt-sebastian: `List<StorageReference>` is the correct return type.

rosariopfernandes: Thanks for the confirmation @schmidt-sebastian 

Pushed a new commit with the update.

schmidt-sebastian:
@rosariopfernandes Just wanted to chime in since I am cc'ed on this: This PR will likely sit here for a while (2-3 weeks) as we obtain internal API approval. There is likely nothing more we need.

rosariopfernandes:
@schmidt-sebastian Yes, I understand. I was just trying to fix the coverage report. Sorry for spamming your notifications inbox.

firebase-storage/ktx/src/main/kotlin/com/google/firebase/storage/ktx/Storage.kt
```diff
@@ -44,6 +49,90 @@ fun storageMetadata(init: StorageMetadata.Builder.() -> Unit): StorageMetadata {
     return builder.build()
 }
 
+/**
+ * Destructuring declaration for [UploadTask.TaskSnapshot] to provide bytesTransferred
```
rlazo, line 53: Add a period at the end of the sentence, here and below.

firebase-storage/ktx/src/test/kotlin/com/google/firebase/storage/ktx/StorageTest.kt
```diff
@@ -105,6 +113,64 @@ class StorageTests : BaseTestCase() {
         assertThat(metadata.getContentType()).isEqualTo("text/html")
         assertThat(metadata.getCacheControl()).isEqualTo("no-cache")
     }
+
+    @Test
+    fun `ListResult destructuring declarations work`() {
+        val mockListResult = Mockito.mock(ListResult::class.java)
```
rlazo, line 119: You're mocking the final class `ListResult`, which ends up requiring the use of `mock-maker-inline` which according to [the docs](https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2) is still *incubating*. 

For this particular case, you can create a helper class in the package `com.google.firebase.storage` that can access  the package private constructor directly. We've already done something similar in 

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/ktx/src/test/java/com/google/firebase/firestore/TestUtil.java

rosariopfernandes: @rlazo the package `com.google.firebase.storage` already had a `TestUtil` class, so I added a `listResult()` method in that same class. PTAL and let me know if that's ok

rlazo: 👍

rlazo:
@rosariopfernandes `check-changed` is failing with

```
> Task :firebase-storage:ktx:compileReleaseUnitTestKotlin FAILED
w: Argument -module-name is passed multiple times. Only the last value will be used: com.google.firebase-firebase-storage-ktx
e: /home/prow/go/src/github.com/firebase/firebase-android-sdk/firebase-storage/ktx/src/test/kotlin/com/google/firebase/storage/ktx/StorageTest.kt: (29, 36): Unresolved reference: TestUtil
e: /home/prow/go/src/github.com/firebase/firebase-android-sdk/firebase-storage/ktx/src/test/kotlin/com/google/firebase/storage/ktx/StorageTest.kt: (119, 30): Unresolved reference: TestUtil
e: /home/prow/go/src/github.com/firebase/firebase-android-sdk/firebase-storage/ktx/src/test/kotlin/com/google/firebase/storage/ktx/StorageTest.kt: (119, 50): Type inference failed: Not enough information to infer parameter T in inline fun <T> listOf(): List<T>
Please specify it explicitly.

e: /home/prow/go/src/github.com/firebase/firebase-android-sdk/firebase-storage/ktx/src/test/kotlin/com/google/firebase/storage/ktx/StorageTest.kt: (119, 60): Type inference failed: Not enough information to infer parameter T in inline fun <T> listOf(): List<T>
Please specify it explicitly.
```

rosariopfernandes:
@rlazo Because the `TestUtil` class was on a different module's test directory, it was inaccessible from the -ktx module. I tried adding it to the -ktx sourceSet like this:

```groovy
sourceSets {
        // ...
        test.java {
            // ...
            srcDir '../src/test/java'
        }
        // ...
}
```

But it would try to import all the other test classes as well, and some of them were showing errors due to the difference in package names.

As a workaround, I created a new directory (/test/kotlin/) and a new TestUtil class (still under the `com.firebase.storage` package) exclusively for the ktx module. Seems like it does the trick.

rlazo:
Would it be possible to add it to `firebase-storage/src/testUtil/java/com/google/firebase/storage` ? That could eliminate the need for a new directory.

rosariopfernandes:
@rlazo I did that, but needed to make 2 additional changes:
1. Renamed the class to `KtxTestUtil`, because it was conflicting with the existing `TestUtil` class.
2. Added Java 8 to the `ktx` module, because some classes from the `testUtil` package use Java 8 features (e.g. Lambda functions).

PR #1646, opened 2020-06-12
Send the FIS auth token in the RC fetch headers.

danasilver:


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -78,6 +78,8 @@
   private static final String X_ANDROID_PACKAGE_HEADER = "X-Android-Package";
   private static final String X_ANDROID_CERT_HEADER = "X-Android-Cert";
   private static final String X_GOOGLE_GFE_CAN_RETRY = "X-Google-GFE-Can-Retry";
+  private static final String INSTALLATIONS_AUTH_TOKEN_HEADER =
+      "X-Goog-Firebase-Installations-Auth";
```
erikeldridge, line 82: Thinking (no action required): this is comparable to usage of the same header in RC's iOS SDK: https://github.com/firebase/firebase-ios-sdk/blob/c07c89ee7582e84da1c5aafaa94b38a53bcfbdc2/FirebaseRemoteConfig/Sources/RCNFetch.m#L49

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java
```diff
@@ -240,22 +242,22 @@ public void fetch_localeUsesToStringBelowLollipop() throws Exception {
...
   @Test
-  public void fetch_instanceIdTokenIsNull_doesNotThrowException() throws Exception {
+  public void fetch_installationAuthTokenIsNull_doesNotThrowException() throws Exception {
     setServerResponseTo(noChangeResponseBody, SECOND_ETAG);
 
-    FetchResponse fetchResponse = fetchWithoutIidToken();
+    FetchResponse fetchResponse = fetchWithoutInstallationAuthToken();
 
-    assertWithMessage("Fetch() failed with null instance id token!")
+    assertWithMessage("Fetch() failed with null installation auth token!")
```
erikeldridge, line 260: Thinking (no action required): this is public error message, so the phrasing should match the public docs. The [FirebaseInstallations.getToken docs](https://firebase.google.com/docs/reference/android/com/google/firebase/installations/FirebaseInstallations#getToken(boolean)) describe "authentication token for the Firebase installation", which matches the phrasing here.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
erikeldridge: ... also FIAM and FIS SDKs:
https://github.com/firebase/firebase-ios-sdk/search?q=x-goog-firebase-installations-auth&unscoped_q=x-goog-firebase-installations-auth

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigConstants.java
```diff
@@ -31,6 +31,10 @@
   /**
    * Keys of fields in the Fetch request body that the client sends to the Firebase Remote Config
    * server.
+   *
+   * <p>{@code INSTANCE_ID} and {@code INSTANCE_ID_TOKEN} are legacy names for the fields that used
```
ankitaj224, line 35: We might need some clarity here. The javadoc gives the impression that INSTANCE_ID_TOKEN can be fetched from FIS SDK, which is incorrect. We can explicitly call out INSTANCE_ID_TOKEN is replaced by  installation auth token fetched from FIS SDK.

danasilver: Good call. Updated.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java
andirayo: Great callout, thank you!
We had a meeting recently with DevRel and Tech Writers and decided to name the authentication tokens for Firebase installations: "installation auth tokens".
So, the action item is probably on us in the FIS team.

PR #1667, opened 2020-06-15
Support Required/Optional Field

James201311:


encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -92,6 +91,20 @@
     return creationCtx;
   }
 
+  private <T> void handleNullValue(FieldRef<T> fieldRef, CreationContextImpl creationCtx)
+      throws IOException {
+    reader.nextNull();
```
rlazo, line 96: Since both checks below fail if the value is `null` I think it's safe for you to *only* read it if you are going to use it (line 105)

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldRef.java
```diff
@@ -25,35 +27,50 @@
...
   public static final class Primitive<T> extends FieldRef<T> {
-    private Primitive(@NonNull TypeToken<T> typeToken) {
+    private Primitive(FieldDescriptor fieldDescriptor, @NonNull TypeToken<T> typeToken) {
+      if (typeToken instanceof TypeToken.ClassToken) {
+        Class<?> clazz = ((TypeToken.ClassToken<T>) typeToken).getRawType();
+        if (!clazz.isPrimitive())
+          throw new IllegalArgumentException(
+              "FieldRef.Primitive<T> can only be used to hold primitive type.\n"
+                  + clazz
+                  + "was found.");
```
rlazo, line 71:
```suggestion
                  + " was found.");
```

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
James201311: `handleNullValue` will only be called when the next token is null, and the only purpose of `reader.nextNull()` is to consume the null token, so that the reader can continue reading other tokens, therefore, we need to call it. 
And it's always safe to do it,  since the `fiedRef` is always non-null, even if its pointing value is null.

rlazo: Well, technically, if the `fieldRef` fails any of the checks below we throw a runtime exception, which cannot be caught, so the reader is not going to be consumed any further :-). So, if you move it to line 104, it will still work. But, I think you can leave it as is for clarity.

James201311: got it, good call!

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
```diff
@@ -43,11 +43,23 @@
 public final class FieldDescriptor {
 
   private final String name;
+  private final boolean required;
+  private final Object defaultValue;
```
vkryachko, line 47: A few consideration regarding `defaultValue`:
* major: It's use will be limited for "composite" types in the case of `@Encodable`-driven code generation. The issue is that annotations cannot have non-constant values in them, so I would not be able to express something like `@Field(defaultValue = new MyPojo(new SubPojo(), "foo"))` and furthermore annotations cannot contain `null` values, so I cannot even say `@Field(defaultValue =null)`.
* minor: It does not appear to be type-safe, i.e. i can put an `int` inside a fieldDescriptor for `float`
* And last - while may not be too big a deal - it auto-boxes for primitives.

All that said, we have to think more about how(if at all) we allow specifying defaults, and if we don't - using proto3-like defaults might be a good start. wdyt?

James201311:
align with Proto3 to support default value instead.
https://github.com/firebase/firebase-android-sdk/pull/1722

PR #1668, opened 2020-06-15
ArrayToken Type Argument

James201311:
Changed type argument in `ArrayToken` to align with the type argument of its component type.

### DIFF
--- a/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
+++ b/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
@@ -81,7 +81,14 @@ public abstract class TypeToken<T> {
       throw new IllegalArgumentException("<? super T> is not supported");
     } else if (type instanceof GenericArrayType) {
       Type componentType = ((GenericArrayType) type).getGenericComponentType();
-      return new ArrayToken<T>(TypeToken.of(componentType));
+      @SuppressWarnings("unchecked")
+      /*
+       Safe, since class ArrayToken<T> extends TypeToken<T[]>, therefore,
+       type parameter in ArrayToken is always the "de-array" type parameter of TypeToken.
+       In this case, type parameter in componentType is the "de-array" type parameter of T.
+      */
+      TypeToken<T> typeToken = (TypeToken<T>) new ArrayToken(TypeToken.of(componentType));
+      return typeToken;
     } else if (type instanceof ParameterizedType) {
       ParameterizedType parameterizedType = (ParameterizedType) type;
       // Safe because rawType of parameterizedType is always instance of Class<T>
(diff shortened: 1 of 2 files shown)


James201311:
> While the types kind of align nicely here, imagine the code like you have in the decoder context:
> 
> ```java
> <T> T decode(TypeToken<T> token) {
>   if (token instanceof ClassToken) {
>     // can easily cast to ClassToken as T is the same
>     ClassToken<T> cls = (ClassToken<T>) token;
>   }
>   if (token instanceof ArrayToken) {
>     // this means T is an array of some Unknown[], what to cast it to?
>     ArrayToken<Unknown> cls = (ClassToken<Unknown>) token;
>   }
> }
> ```
> 
> Since we don't know `Unknown` how can we write the code above? And if we can't what are the possible workarounds?

If we cast it back to `ArrayToken<?>` and past it to another generic method `decodeArrayToken` will it capture the `?` 

```
else if (typeToken instanceof TypeToken.ArrayToken) {
      return (T) decodeArrayToken((TypeToken.ArrayToken<?>) typeToken);
}

private <T> T[] decodeArrayToken(TypeToken.ArrayToken<T> arrayToken);
```

vkryachko:
> > While the types kind of align nicely here, imagine the code like you have in the decoder context:
> > ```java
> > <T> T decode(TypeToken<T> token) {
> >   if (token instanceof ClassToken) {
> >     // can easily cast to ClassToken as T is the same
> >     ClassToken<T> cls = (ClassToken<T>) token;
> >   }
> >   if (token instanceof ArrayToken) {
> >     // this means T is an array of some Unknown[], what to cast it to?
> >     ArrayToken<Unknown> cls = (ClassToken<Unknown>) token;
> >   }
> > }
> > ```
> > 
> > 
> > Since we don't know `Unknown` how can we write the code above? And if we can't what are the possible workarounds?
> 
> If we cast it back to `ArrayToken<?>` and pass it to another generic method `decodeArrayToken` will it capture the `?`
> 
> ```
> else if (typeToken instanceof TypeToken.ArrayToken) {
>       return (T) decodeArrayToken((TypeToken.ArrayToken<?>) typeToken);
> }
> 
> private <T> T[] decodeArrayToken(TypeToken.ArrayToken<T> arrayToken);
> ```
> 
> Just have some experiments, seems like it works. wdyt? @vkryachko

iiuc this would work for boxed arrays but if you want to return say a primitive array like `int[]` it will not compile. Although it might work out with some casts, not sure.

Additionally there is a weird edge case for primitives:
Say you have a `TypeToken<int[]>`, its `ArrayToken` will automatically be promoted to `ArrayToken<Integer>` as generics cannot be parameterized by primitives. At this point it's not possible to distinguish primitive `int` arrays from boxed `Integer` arrays at the type level.

All that said, I like the idea in general and not strongly opposed to it, it may just lead to a dead end due to issues/inconsistencies I mentioned above. We can try it out and see where it leads us and if it solves more problems than it may cause. wdyt?

James201311:
> > > While the types kind of align nicely here, imagine the code like you have in the decoder context:
> > > ```java
> > > <T> T decode(TypeToken<T> token) {
> > >   if (token instanceof ClassToken) {
> > >     // can easily cast to ClassToken as T is the same
> > >     ClassToken<T> cls = (ClassToken<T>) token;
> > >   }
> > >   if (token instanceof ArrayToken) {
> > >     // this means T is an array of some Unknown[], what to cast it to?
> > >     ArrayToken<Unknown> cls = (ClassToken<Unknown>) token;
> > >   }
> > > }
> > > ```
> > > 
> > > 
> > > Since we don't know `Unknown` how can we write the code above? And if we can't what are the possible workarounds?
> > 
> > 
> > If we cast it back to `ArrayToken<?>` and pass it to another generic method `decodeArrayToken` will it capture the `?`
> > ```
> > else if (typeToken instanceof TypeToken.ArrayToken) {
> >       return (T) decodeArrayToken((TypeToken.ArrayToken<?>) typeToken);
> > }
> > 
> > private <T> T[] decodeArrayToken(TypeToken.ArrayToken<T> arrayToken);
> > ```
> > 
> > 
> > Just have some experiments, seems like it works. wdyt? @vkryachko
> 
> iiuc this would work for boxed arrays but if you want to return say a primitive array like `int[]` it will not compile. Although it might work out with some casts, not sure.
> 
> Additionally there is a weird edge case for primitives:
> Say you have a `TypeToken<int[]>`, its `ArrayToken` will automatically be promoted to `ArrayToken<Integer>` as generics cannot be parameterized by primitives. At this point it's not possible to distinguish primitive `int` arrays from boxed `Integer` arrays at the type level.
> 
> All that said, I like the idea in general and not strongly opposed to it, it may just lead to a dead end due to issues/inconsistencies I mentioned above. We can try it out and see where it leads us and if it solves more problems than it may cause. wdyt?

Good Point!!! I will think more about it, then we can discuss it more in our 1:1,

James201311:
@vkryachko @rlazo 
This PR is no longer needed since it will make arrays of primitive harder to decode. 

if type argument `T` in `ArrayToken<T>` aligns with the type argument of its component type. Then `T` will auto-boxed to its wrapper class, then end up with an array of its Wrapper Class, also casting them back to primitives array is also hard. For example,  we have an `int[]` represented by `ArrayToken<int>` where `int` will be au-boxed into to `Integer`, after decoding we will end up with `Integer[]`, casting them back to `int[]` increase the complexity and requires more work. 
However, if we keep type argument `T` to align with its own type we won't have the same issue, since `int[]` is a reference type, it won't be wrapped.

PR #1672, opened 2020-06-17
Unified emulator settings for Database

samtstern:
This is the first PR for http://go/firebase-emulator-connection-api

This implements the global changes needed in `FirebaseApp` as well as the changes in `FirebaseDatabase` to support them.

TODO:

- [x] Properly hide the `FirebaseApp` changes so that this doesn't leak before Functions / Firestore are implemented.

### DIFF
--- a/firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
+++ b/firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
@@ -14,6 +14,19 @@
 
 package com.google.firebase;
 
+import static com.google.android.gms.common.util.Base64Utils.decodeUrlSafeNoPadding;
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.firebase.common.testutil.Assert.assertThrows;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
 import android.content.BroadcastReceiver;
 import android.content.ComponentCallbacks2;
 import android.content.Context;
@@ -21,11 +34,9 @@ import android.content.ContextWrapper;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.os.UserManager;
-
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 import androidx.test.InstrumentationRegistry;
 import androidx.test.runner.AndroidJUnit4;
-
 import com.google.android.gms.common.api.internal.BackgroundDetector;
 import com.google.common.base.Defaults;
 import com.google.firebase.auth.FirebaseAuth;
@@ -39,13 +50,6 @@ import com.google.firebase.emulators.EmulatorSettings;
 import com.google.firebase.emulators.FirebaseEmulator;
 import com.google.firebase.platforminfo.UserAgentPublisher;
 import com.google.firebase.testing.FirebaseAppRule;
-
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.stubbing.Answer;
-
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -55,19 +59,11 @@ import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-
-import static com.google.android.gms.common.util.Base64Utils.decodeUrlSafeNoPadding;
-import static com.google.common.truth.Truth.assertThat;
-import static com.google.firebase.common.testutil.Assert.assertThrows;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.reset;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.stubbing.Answer;
 
 /** Unit tests for {@link com.google.firebase.FirebaseApp}. */
 // TODO(arondeak): uncomment lines when Firebase API targets are in integ.
@@ -436,9 +432,7 @@ public class FirebaseAppTest {
     EmulatedServiceSettings databaseSettings =
         new EmulatedServiceSettings.Builder("10.0.2.2", 9000).build();
     EmulatorSettings emulatorSettings =
-        new EmulatorSettings.Builder()
-            .addEmulatedService(emulator, databaseSettings)
-            .build();
+        new EmulatorSettings.Builder().addEmulatedService(emulator, databaseSettings).build();
 
     // Set twice
     firebaseApp.enableEmulators(emulatorSettings);
@@ -455,9 +449,7 @@ public class FirebaseAppTest {
     EmulatedServiceSettings databaseSettings =
         new EmulatedServiceSettings.Builder("10.0.2.2", 9000).build();
     EmulatorSettings emulatorSettings =
-        new EmulatorSettings.Builder()
-            .addEmulatedService(emulator, databaseSettings)
-            .build();
+        new EmulatorSettings.Builder().addEmulatedService(emulator, databaseSettings).build();
     firebaseApp.enableEmulators(emulatorSettings);
 
     // Access (as if from the Database SDK)
--- a/firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
+++ b/firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
@@ -14,6 +14,48 @@
 
 package com.google.firebase;
 
+import android.content.BroadcastReceiver;
+import android.content.ComponentCallbacks2;
+import android.content.Context;
+import android.content.ContextWrapper;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.os.UserManager;
+
+import androidx.localbroadcastmanager.content.LocalBroadcastManager;
+import androidx.test.InstrumentationRegistry;
+import androidx.test.runner.AndroidJUnit4;
+
+import com.google.android.gms.common.api.internal.BackgroundDetector;
+import com.google.common.base.Defaults;
+import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.components.EagerSdkVerifier;
+import com.google.firebase.components.InitTracker;
+import com.google.firebase.components.TestComponentOne;
+import com.google.firebase.components.TestComponentTwo;
+import com.google.firebase.components.TestUserAgentDependentComponent;
+import com.google.firebase.emulators.EmulatedServiceSettings;
+import com.google.firebase.emulators.EmulatorSettings;
+import com.google.firebase.emulators.FirebaseEmulators;
+import com.google.firebase.platforminfo.UserAgentPublisher;
+import com.google.firebase.testing.FirebaseAppRule;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.stubbing.Answer;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import static com.google.android.gms.common.util.Base64Utils.decodeUrlSafeNoPadding;
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.firebase.common.testutil.Assert.assertThrows;
@@ -26,41 +68,6 @@ import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
-
-import android.content.BroadcastReceiver;
-import android.content.ComponentCallbacks2;
-import android.content.Context;
-import android.content.ContextWrapper;
-import android.content.Intent;
-import android.content.IntentFilter;
-import android.os.UserManager;
-import androidx.localbroadcastmanager.content.LocalBroadcastManager;
-import androidx.test.InstrumentationRegistry;
-import androidx.test.runner.AndroidJUnit4;
-import com.google.android.gms.common.api.internal.BackgroundDetector;
-import com.google.common.base.Defaults;
-import com.google.firebase.auth.FirebaseAuth;
-import com.google.firebase.components.EagerSdkVerifier;
-import com.google.firebase.components.InitTracker;
-import com.google.firebase.components.TestComponentOne;
-import com.google.firebase.components.TestComponentTwo;
-import com.google.firebase.components.TestUserAgentDependentComponent;
-import com.google.firebase.platforminfo.UserAgentPublisher;
-import com.google.firebase.testing.FirebaseAppRule;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.stubbing.Answer;
 
 /** Unit tests for {@link com.google.firebase.FirebaseApp}. */
 // TODO(arondeak): uncomment lines when Firebase API targets are in integ.
@@ -415,6 +422,43 @@ public class FirebaseAppTest {
     // Because default is true.
     firebaseApp.setDataCollectionDefaultEnabled(null);
     assertTrue(firebaseApp.isDataCollectionDefaultEnabled());
+  }
+
+  @Test
+  public void testEnableEmulators_shouldAllowDoubleSetBeforeAccess() {
+    Context mockContext = createForwardingMockContext();
+    FirebaseApp firebaseApp = FirebaseApp.initializeApp(mockContext);
+
+    EmulatedServiceSettings databaseSettings =
+        new EmulatedServiceSettings.Builder("10.0.2.2", 9000).build();
+    EmulatorSettings emulatorSettings =
+        new EmulatorSettings.Builder()
+            .addEmulatedService(FirebaseEmulators.DATABASE, databaseSettings)
+            .build();
+
+    // Set twice
+    firebaseApp.enableEmulators(emulatorSettings);
+    firebaseApp.enableEmulators(emulatorSettings);
+  }
+
+  @Test
+  public void testEnableEmulators_shouldThrowIfSetAfterAccess() {
+    Context mockContext = createForwardingMockContext();
+    FirebaseApp firebaseApp = FirebaseApp.initializeApp(mockContext);
+
+    EmulatedServiceSettings databaseSettings =
+        new EmulatedServiceSettings.Builder("10.0.2.2", 9000).build();
+    EmulatorSettings emulatorSettings =
+        new EmulatorSettings.Builder()
+            .addEmulatedService(FirebaseEmulators.DATABASE, databaseSettings)
+            .build();
+    firebaseApp.enableEmulators(emulatorSettings);
+
+    // Access (as if from the Database SDK)
+    firebaseApp.getEmulatorSettings().getServiceSettings(FirebaseEmulators.DATABASE);
+
+    // Try to set again
+    assertThrows(IllegalStateException.class, () -> firebaseApp.enableEmulators(emulatorSettings));
   }
 
   /** Returns mock context that forwards calls to targetContext and localBroadcastManager. */
--- a/firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
+++ b/firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
@@ -429,8 +429,7 @@ public class FirebaseAppTest {
     // dependency for this test.
     FirebaseEmulator emulator = FirebaseEmulator.forName("database");
 
-    EmulatedServiceSettings databaseSettings =
-        new EmulatedServiceSettings.Builder("10.0.2.2", 9000).build();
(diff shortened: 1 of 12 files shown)


firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
```diff
@@ -56,11 +55,19 @@
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.stubbing.Answer;
+
+import static com.google.android.gms.common.util.Base64Utils.decodeUrlSafeNoPadding;
```
samtstern, line 59: I'm gonna guess I have a bad IDE setting?  I used google-java-format

samtstern:
@vkryachko well `common` will always have to depend on these emulator classes and they are useless outside of `common` so I don't see a reason to break them out.  They're just tiny data classes they won't gain any actual functionality.

firebase-common/src/main/java/com/google/firebase/emulators/FirebaseEmulators.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+ *
+ * @see com.google.firebase.FirebaseApp#enableEmulators(EmulatorSettings)
+ * @see EmulatorSettings
+ * @see EmulatedServiceSettings
+ * @hide
+ */
+@KeepForSdk
+public enum FirebaseEmulators {
```
samtstern, line 30: @vkryachko if these were string constants rather than an enum would that help?  Then I don't think all BaaS would have to be released after each emulator is released, only `common` and the SDK which corresponds to the new emulator.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -142,6 +148,20 @@ public FirebaseOptions getOptions() {
...
 
+  /**
+   * Returns the specified {@link EmulatorSettings}.
+   *
+   * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+   *
+   * @hide
+   */
+  @KeepForSdk
+  @Nullable
```
vkryachko, line 159: Would it make sense to make it `@NotNull` and return a default "empty" settings object. It should help with the following:
* Make it more null-safe and simplify integrating SDKs as they won't have to deal with `null` twice as I can see in the implementation below.
*iiuc it will help make the "set-after-use" logic below more safe, i.e. with the null case it seems possible to first initialize firestore and then call `enableEmulators()`

wdyt?

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
```diff
@@ -0,0 +1,103 @@
...
+      Preconditions.checkState(
+          !settingsMap.containsKey(emulator),
+          "Cannot call addEmulatedService twice for " + emulator.toString());
+      this.settingsMap.put(emulator, settings);
+      return this;
+    }
+
+    @NonNull
+    public EmulatorSettings build() {
+      return new EmulatorSettings(settingsMap);
```
vkryachko, line 66: nit: consider making a copy of the map, otherwise if a reference to the builder is kept around, one can modify the underlying map unintentionally.

firebase-common/src/main/java/com/google/firebase/emulators/FirebaseEmulators.java
vkryachko, line 30: While making this an enum makes the API simple and type safe, it couples firebase-common to all baas sdks.

To avoid the coupling I suggest to either resort to using `String`s with each string defined as a constant on the target SDK, i.e. `FirebaseFirestore.EMULATOR = "firestore"` or something.
Or wrapping that string in a class here in common like `data class Emulator(serviceName: String)` and initializing the SDK constant with it instead of the string.

A potential additional advantage of this is that one can only configure an emulator for say `firestore` if it's actually present in their app.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -305,6 +325,28 @@ public static FirebaseApp initializeApp(
...
+   *
+   * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+   *
+   * @param emulatorSettings the emulator settings for all services.
+   * @hide
+   */
+  @KeepForSdk
+  public void enableEmulators(@NonNull EmulatorSettings emulatorSettings) {
+    checkNotDeleted();
+    Preconditions.checkState(
```
vkryachko, line 344: How will this interact with firestore's `terminate()`? afaicr one can terminate one instance of firestore at which point the call to `FirebaseFirestore#getInstance()` will create a new firestore instance, should we allow the following sequence of calls?

```java
firestore.terminate();
app.updateFirestoresEmulator();
Firebase.firestore.use();
```

firebase-common/src/main/java/com/google/firebase/emulators/FirebaseEmulators.java
vkryachko: @samtstern imo it would, although additionally - as I mentioned below - ideally those constants should not live inside common either but rather in their respective sdks.

samtstern: Ah ok your suggestion below is a good one!

firebase-common/src/androidTest/java/com/google/firebase/FirebaseAppTest.java
samtstern: Thanks for the pointer to that task.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: That is done in the `EmulatorSettings()` constructor with `unmodifiableMap`

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
samtstern: Thanks for calling this out, I resolved all the other comments since they were straightforward but I will have to look into this one more deeply.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
vkryachko: The problem with `unmodifiableMap` is that it only a read-only **view** into the underlying map, so modifications to the underlying map will be seen by the view.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
samtstern: Ok actually I do *not* want to allow that sequence of calls.  One of the main reasons for this is to make sure that app developers can reason about their emulation settings at a single point before *any* Firebase service has done any communication.

If you call `terminate()` the settings should still apply.  If you want to nuke them you'll need to call `FirebaseApp#delete()` and then make a new app.

WDYT?

vkryachko: That's a good point and sgtm. You may want to confirm this behavior with firestore folks though

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: Ah thanks for the clarification, resolved.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
samtstern: Will do during the Firestore PR.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Fetch the emulation settings for a single Firebase service. Once this method has been called
+   * {@link #isAccessed()} will return true.
+   *
+   * @hide
+   */
+  @Nullable
+  public EmulatedServiceSettings getServiceSettings(@NonNull FirebaseEmulator emulator) {
+    accessed.set(true);
```
schmidt-sebastian, line 96: The class as implemented appears immutable. It doesn't seem possible to call   `addEmulatedService` and modify an existing instance, so it is not immediately obvious why we need the "accessed" check. 

Would it make more sense to perform this check in `FirebaseApp` and enforce that `enableEmulators` can only be called once? Along the same line, could `enableEmulators` take `FirebaseEmulator emulator,  EmulatedServiceSettings settings` directly, which would remove the need for `EmulatorSettings` API.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -73,6 +81,13 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
       originalPathString = originalPathString.replace("+", " ");
       Validation.validateRootPathString(originalPathString);
 
+      // TODO: Should log this out
+      if (serviceSettings != null) {
+        repoInfo.secure = false;
+        repoInfo.host = serviceSettings.host + ":" + serviceSettings.port;
+        repoInfo.internalHost = repoInfo.host;
```
schmidt-sebastian, line 88: Should we add an explicit way to decide which credentials to use? I am asking because of Yuchen's feedback here: https://github.com/firebase/firebase-js-sdk/pull/3228

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
```diff
@@ -0,0 +1,59 @@
...
+    private final int port;
+
+    /**
+     * Create a new EmulatedServiceSettings builder.
+     *
+     * @param host the host where the local emulator is running. If you want to access 'localhost'
+     *     from an Android Emulator use '10.0.2.2' instead.
+     * @param port the port where the local emulator is running.
+     */
+    public Builder(@NonNull String host, int port) {
```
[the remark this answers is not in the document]
samtstern: I've always been told that required arguments should be in the `Builder` constructor.  The awkward thing here is that the `Builder` has no other methods at all but in the API discussion it was clear that we may want to extend this object one day (`setInstance()` maybe or `setCredentials` as you suggest below) so having a `Builder` now rather than a plain constructor is more future proof.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
```diff
@@ -0,0 +1,104 @@
...
+    this.settingsMap = Collections.unmodifiableMap(settingsMap);
+  }
+
+  /**
+   * Determine if any Firebase SDK has already accessed the emulator settings. When true, attempting
+   * to change the settings should throw an error.
+   *
+   * @hide
+   */
+  public boolean isAccessed() {
```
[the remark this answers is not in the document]
samtstern: I like that better, will change.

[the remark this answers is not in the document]
samtstern: It's actually not the same, see below.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: It's not about preventing someone calling `enableEmulators` twice, it's about someone calling it *after* the Database/Firestore/etc SDK has already consulted `FirebaseApp.getEmulatorSettings()`.

It's sort of like `firestore.settings({ })` where you can call it all you want but once the SDK "boots up" we throw if you call it.

As for changing the API to be repeated calls to `enableEmulators`:
  1. That would be a pretty significant departure from the API approved by the council.
  2. The idea here is that developers should have a *single* place where they control all the emulator settings in their app.  So I am personally in favor of a single call to `enableEmulators()` with all of the service settings in one object.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
samtstern: I do think that would be a nice addition one day but this API was approved cross-platform for Android, iOS, and JS without any mention of a credentials setting so I think I'd need to do a new API review for that.

Do you anticipate bugs like that JS one in this implementation?

yuchenshi: This should be low risk since Android SDK is not reused in Admin SDKs. The expected behavior is the same as production -- get credentials from Firebase Auth if used, or no credentials if not.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
schmidt-sebastian: Okay, understood. I think we should still try to simplify the code flow, which would make it more obvious to code reviewers like me. Can we make this explicit in Firebase App (by adding an internal `freezeSettings()` or something)? ideally, trying to understand this should not overheat my brain, but it is also very hot today.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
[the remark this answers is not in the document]
vkryachko: +1 on required parameters as builder constructor parameters

Although to the point of being future-proof: it seems that it's possible to be future-proof by not having a builder at all at this point and expose the constructor directly, then add the builder when(and most importantly **if**) needed and tell people to use the builder for additional functionality. wdyt?

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: Yes that's a good idea.

samtstern: Added a hidden `freeze()` method on `EmulatorSettings` and made it the responsibility of `FirebaseApp` to call it.

Added a warning if somehow someone is getting service settings from a non-frozen `EmulatorSettings`, which shouldn't be possible.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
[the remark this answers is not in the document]
samtstern: I am open to whatever you think.  I guess let's do the shorter thing now and do the builder later.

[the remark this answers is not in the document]
samtstern: Builder removed.

samtstern:
@schmidt-sebastian all of your comments were about the FirebaseApp stuff (and they were useful!) but I added you as a database reviewer.

1. Is there anyone else I should add?
2. Do you have any concerns about the database stuff?

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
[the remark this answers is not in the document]
vkryachko: Whatever we decide has to go through api review, has it happened yet btw?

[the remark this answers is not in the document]
samtstern: Yes the API (go/firebase-emulator-connection-api) was approved before I started this PR.

IMO we don't have to go back through API review to change a Builder to a non-Builder, especially since you are the Android platform approver.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
vkryachko: Hm, that's interesting. What's become more apparent to me with your latest changes is the observation that `EmulatorSettings` are not really concerned about being frozen, it's `FirebaseApp` that both `freezes` and checks for `frozen` on it.

This kind of results in the situation where the `frozen` logic is spread out among both `FirebaseApp` and `EmulatorSettings`. Would it make sense to make `EmulatorSettings` just a pure immutable pojo and have `FirebaseApp` store the `frozen` bit since it's the only one who uses it and has full control of the value?

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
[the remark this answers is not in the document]
vkryachko: Just to clarify, I don't have a vote in the API council, I just offer my recommendations :)

That said, it would not hurt to drop an email to your api review thread, but it's definitely up to you.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: That's cool with me!  I agree that's where it should live, I just thought maybe people didn't want that kind of state on `FirebaseApp`.  I'll move it.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -142,6 +145,20 @@ public FirebaseOptions getOptions() {
...
+   * Returns the specified {@link EmulatorSettings} or a default.
+   *
+   * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+   *
+   * @hide
+   */
+  @NonNull
+  public EmulatorSettings getEmulatorSettings() {
+    checkNotDeleted();
+    emulatorSettings.freeze();
```
schmidt-sebastian: If this remains hidden, then the frozen state could just be a simple atomic boolean here. If this is ever public, then this won't work since we wouldn't want to freeze the settings just because this method was called. I would vote we go the simple route now and use a single (atomic) boolean that manages the state in Firebase app.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: Moved boolean state to `FirebaseApp`

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -32,7 +34,13 @@
 public class Utilities {
   private static final char[] HEX_CHARACTERS = "0123456789abcdef".toCharArray();
 
-  public static ParsedUrl parseUrl(String url) throws DatabaseException {
+  public static ParsedUrl parseUrl(@NonNull String url) {
+    return Utilities.parseUrl(url, null);
+  }
+
+  public static ParsedUrl parseUrl(
+      @NonNull String url, @Nullable EmulatedServiceSettings serviceSettings)
```
schmidt-sebastian, line 42: s/serviceSettings/emulatorSettings ?

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -73,6 +81,13 @@ public static ParsedUrl parseUrl(String url) throws DatabaseException {
       originalPathString = originalPathString.replace("+", " ");
       Validation.validateRootPathString(originalPathString);
 
+      // TODO: Should log this out
+      if (serviceSettings != null) {
+        repoInfo.secure = false;
+        repoInfo.host = serviceSettings.host + ":" + serviceSettings.port;
+        repoInfo.internalHost = repoInfo.host;
+      }
+
```
schmidt-sebastian, line 90: As for comments for here, I think this code is fine. A cleaner way might be https://gist.github.com/schmidt-sebastian/e1ddba17d468ae72c4104fa84aae2d9a, but it's optional.

samtstern: You're right that is better, I implemented that diff.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -110,6 +111,9 @@
   private final FirebaseOptions options;
   private final ComponentRuntime componentRuntime;
 
+  private final AtomicBoolean emulatorSettingsFrozen = new AtomicBoolean(false);
+  private EmulatorSettings emulatorSettings = EmulatorSettings.getDefault();
```
vkryachko, line 115: nit: now that settings are immutable, consider sharing the default instance across app instances.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Settings to connect a single Firebase service to a local emulator.
+ *
+ * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+ *
+ * @see EmulatorSettings
+ * @hide
+ */
+public class EmulatedServiceSettings {
```
vkryachko, line 27: nit:
```suggestion
public final class EmulatedServiceSettings {
```

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Settings that control which Firebase services should access a local emulator, rather than
+ * production.
+ *
+ * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+ *
+ * @see com.google.firebase.FirebaseApp#enableEmulators(EmulatorSettings)
+ * @hide
+ */
+public class EmulatorSettings {
```
vkryachko, line 33: nit:
```suggestion
public final class EmulatorSettings {
```

firebase-common/src/main/java/com/google/firebase/emulators/FirebaseEmulator.java
```diff
@@ -0,0 +1,46 @@
...
+ * Identifier Firebase services that can be emulated using the Firebase Emulator Suite.
+ *
+ * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+ *
+ * @see com.google.firebase.FirebaseApp#enableEmulators(EmulatorSettings)
+ * @see EmulatorSettings
+ * @see EmulatedServiceSettings
+ * @hide
+ */
+public class FirebaseEmulator {
```
vkryachko, line 29: nit: 
```suggestion
public final class FirebaseEmulator {
```

firebase-common/src/main/java/com/google/firebase/emulators/FirebaseEmulator.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public final String name;
```
vkryachko, line 31: Might be worth making a getter for it: go/ej3e-16

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
samtstern: Ah yeah good call!

firebase-common/src/main/java/com/google/firebase/emulators/EmulatedServiceSettings.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+ *
+ * @see EmulatorSettings
+ * @hide
+ */
+public final class EmulatedServiceSettings {
+
+  public final String host;
+  public final int port;
```
vkryachko, line 30: same here pls :)

samtstern:
```
12: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':firebase-firestore:firebase-test-labUpload'.
> Process 'command 'gcloud'' finished with non-zero exit value 10
```

That's unrelated and this PR was previously green, so gonna merge anyway.

PR #1683, opened 2020-06-18
support decode array token

James201311:
- [] will be decoded as empty array.

### DIFF
--- a/encoders/firebase-decoders-json/firebase-decoders-json.gradle
+++ b/encoders/firebase-decoders-json/firebase-decoders-json.gradle
@@ -54,7 +54,6 @@ dependencies {
     testImplementation "com.google.truth:truth:$googleTruthVersion"
     testImplementation 'org.mockito:mockito-core:2.25.0'
     implementation project(path: ':encoders:firebase-encoders-json')
-    implementation project(path: ':encoders:firebase-encoders-processor:test-support')
 }
 
 tasks.withType(JavaCompile) {
--- a/encoders/firebase-decoders-json/firebase-decoders-json.gradle
+++ b/encoders/firebase-decoders-json/firebase-decoders-json.gradle
@@ -54,6 +54,7 @@ dependencies {
     testImplementation "com.google.truth:truth:$googleTruthVersion"
     testImplementation 'org.mockito:mockito-core:2.25.0'
     implementation project(path: ':encoders:firebase-encoders-json')
+    implementation project(path: ':encoders:firebase-encoders-processor:test-support')
 }
 
 tasks.withType(JavaCompile) {
--- a/encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
+++ b/encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
@@ -85,18 +85,13 @@ public class JsonDataDecoderBuilderContext implements DataDecoder {
 
   private static <T, E> T convertGenericListToArray(
       List<E> list, TypeToken.ArrayToken<T> arrayToken) {
-    if (list.size() == 0) return null;
     @SuppressWarnings("unchecked")
     TypeToken<E> componentTypeToken = (TypeToken<E>) arrayToken.getComponentType();
-    if (componentTypeToken instanceof TypeToken.ClassToken) {
-      TypeToken.ClassToken<E> componentClassToken = (TypeToken.ClassToken<E>) componentTypeToken;
-      Class<E> componentClass = componentClassToken.getRawType();
-      if (componentClassToken.getRawType().isPrimitive()) {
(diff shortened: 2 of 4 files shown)


vkryachko:
I am about to look at the code, but have a question regarding:

> [] will be decoded as null

What's the motivation for this? Why not decode it as empty array instead?

James201311:
> I am about to look at the code, but have a question regarding:
> 
> > [] will be decoded as null
> 
> What's the motivation for this? Why not decode it as empty array instead?

Haven't figure out a good way to initialize a generic empty array, will work on it.

James201311:
> > I am about to look at the code, but have a question regarding:
> > > [] will be decoded as null
> > 
> > 
> > What's the motivation for this? Why not decode it as empty array instead?
> 
> Haven't figure out a good way to initialize a generic empty array, will work on it.

One possible solution is that we move the `rawType` to `TypeToken` level instead of only inside `ClassToken` so that `ArrayToken` will also be able to get its raw type, therefore, we can create generic array no matter if the given list is empty.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -50,26 +53,115 @@
...
+    TypeToken<E> componentTypeToken = (TypeToken<E>) arrayToken.getComponentType();
+    if (componentTypeToken instanceof TypeToken.ClassToken) {
+      TypeToken.ClassToken<E> componentClassToken = (TypeToken.ClassToken<E>) componentTypeToken;
+      Class<E> componentClass = componentClassToken.getRawType();
+      if (componentClassToken.getRawType().isPrimitive()) {
+        return convertGenericListToPrimitiveArray(list, componentClass, arrayToken);
+      }
+    }
+    @SuppressWarnings("unchecked") // Safe, list is not empty
+    E[] arr = (E[]) Array.newInstance(list.get(0).getClass(), list.size());
```
James201311, line 99: if the first element is `null`, it will fail.

vkryachko:
> > > I am about to look at the code, but have a question regarding:
> > > > [] will be decoded as null
> > > 
> > > 
> > > What's the motivation for this? Why not decode it as empty array instead?
> > 
> > 
> > Haven't figure out a good way to initialize a generic empty array, will work on it.
> 
> One possible solution is that we move the `rawType` to `TypeToken` level instead of only inside `ClassToken` so that `ArrayToken` will also be able to get its raw type, therefore, we can create generic array no matter if the given list is empty.

+1, the choice of the type for the array should definitely be driven by the `TypeToken`, not values in the list.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -50,26 +53,110 @@
...
+    if (classToken.getRawType().isPrimitive()) {
+      return decodePrimitive(classToken);
+    } else if (isSingleValue(classToken)) {
+      return decodeSingleValue(classToken);
+    } else {
+      return decodeObject(classToken);
+    }
+  }
+
+  private <T, E> T decodeArrayToken(TypeToken.ArrayToken<T> arrayToken) throws IOException {
```
vkryachko, line 74: What's the purpose of the `E` type parameter? can you use `Object` in `TypeToken` below instead?

vkryachko:
@James201311 pls also update the PR's description, as it looks like empty arrays are no longer decoded as `null` values.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
James201311: My initial thought was to align with `convertGenericListToArray` since `convertGenericListToArray` need `E` to create `E[]` generic array to assign value to its elements.

James201311: But it's a generic method, maybe I don't need to align?

encoders/firebase-decoders-json/firebase-decoders-json.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     testImplementation "com.google.truth:truth:$googleTruthVersion"
     testImplementation 'org.mockito:mockito-core:2.25.0'
     implementation project(path: ':encoders:firebase-encoders-json')
+    implementation project(path: ':encoders:firebase-encoders-processor:test-support')
```
vkryachko: you should not use this dependency, it's only needed to be able to test encoders-processor, depending on `:encoders:firebase-encoders-json` should be enough(which you already do).

PR #1685, opened 2020-06-18
Move `rawType` to `TypeToken`

James201311:
Move the `rawType` to `TypeToken` level instead of only inside `ClassToken` so that `ArrayToken` will also be able to get its raw type, which enables us to decode arrays of object.

Geting `Class<?>` type from `GenericArrayType` is not straightforward and hard, we will not support `GenericArrayType` for now.

### DIFF
--- a/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
+++ b/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
@@ -36,8 +36,7 @@ import java.lang.reflect.WildcardType;
  * </ol>
  */
 public abstract class TypeToken<T> {
-  @NonNull
-  protected Class<T> rawType;
+  @NonNull protected Class<T> rawType;
 
   @NonNull
   public Class<T> getRawType() {
@@ -92,7 +91,7 @@ public abstract class TypeToken<T> {
       throw new IllegalArgumentException("<? super T> is not supported");
     } else if (type instanceof GenericArrayType) {
       throw new IllegalArgumentException(
-          "Generic Arrays are not supported, " + "please use Lists instead.");
+          "Generic Arrays are not supported, please use Lists instead.");
     } else if (type instanceof ParameterizedType) {
       ParameterizedType parameterizedType = (ParameterizedType) type;
       // Safe because rawType of parameterizedType is always instance of Class<T>
@@ -162,8 +161,7 @@ public abstract class TypeToken<T> {
         return false;
       }
       ClassToken<?> that = (ClassToken<?>) o;
-      return rawType.equals(that.getRawType())
-          && this.typeArguments.equals(that.typeArguments);
+      return rawType.equals(that.getRawType()) && this.typeArguments.equals(that.typeArguments);
     }
 
     @NonNull
--- a/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
+++ b/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
@@ -86,6 +86,7 @@ public abstract class TypeToken<T> {
       }
       throw new IllegalArgumentException("<? super T> is not supported");
     } else if (type instanceof GenericArrayType) {
+      // TODO: Support GenericArrayType
       throw new IllegalArgumentException("GenericArrayType is not supported.");
     } else if (type instanceof ParameterizedType) {
       ParameterizedType parameterizedType = (ParameterizedType) type;
--- a/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
+++ b/encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
@@ -36,6 +36,12 @@ import java.lang.reflect.WildcardType;
  * </ol>
  */
 public abstract class TypeToken<T> {
+  private Class<T> rawType;
+
+  @NonNull
+  public Class<T> getRawType() {
+    return rawType;
+  }
 
   /**
    * Return an {@link TypeToken} to represent generic type {@code T}.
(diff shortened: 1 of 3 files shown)


encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -36,6 +36,12 @@
  * </ol>
  */
 public abstract class TypeToken<T> {
+  private Class<T> rawType;
```
vkryachko, line 39: pls make it final and set through the private constructor.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -126,32 +132,26 @@ public String toString() {
...
-
     @NonNull
     public TypeTokenContainer getTypeArguments() {
       return typeArguments;
     }
 
     @Override
     public int hashCode() {
-      return 11 * rawType.hashCode() + typeArguments.hashCode();
+      return 11 * super.rawType.hashCode() + typeArguments.hashCode();
```
vkryachko, line 154: I don't think you need `super` here or anywhere above and below.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -80,8 +86,8 @@
       }
       throw new IllegalArgumentException("<? super T> is not supported");
     } else if (type instanceof GenericArrayType) {
-      Type componentType = ((GenericArrayType) type).getGenericComponentType();
-      return new ArrayToken<T>(TypeToken.of(componentType));
+      // TODO: Support GenericArrayType
+      throw new IllegalArgumentException("GenericArrayType is not supported.");
```
vkryachko, line 90: I've read a bit more on generic arrays and it looks like they are even more weird than I'd thought, i.e. there is not even a "sane" way to create one - it just won't compile - even though the type `Optional<String>[]` itself is legal in java.

So I would suggest you remove the todo and just leave it as explicitly not supported.

Also consider using the error message rtdb uses https://github.com/firebase/firebase-android-sdk/blob/a7096ac2a65842f9528243b2451e5b172de4c789/firebase-database/src/main/java/com/google/firebase/database/core/utilities/encoding/CustomClassMapper.java#L200

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -180,7 +181,8 @@ String getTypeTokenLiteral() {
   public static class ArrayToken<T> extends TypeToken<T> {
     private final TypeToken<?> componentType;
 
-    private ArrayToken(TypeToken<?> componentType) {
+    private ArrayToken(Class<T> rawType, TypeToken<?> componentType) {
+      super.rawType = rawType;
       this.componentType = componentType;
     }
```
vkryachko, line 187: Consider removing the second argument and instead doing something like:

```suggestion
    private ArrayToken(Class<T> rawType) {
      if(!rawType.isArray()) {
        throw new IllegalArgumentException("blah-blah");
      }
      super(rawType);
      this.componentType = TypeToken.of(rawType.getComponentType());
    }
```

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
James201311: since `ArrayToken` is inner class and `rawType` is private, we need `super` to access the `rawType` or use `getRawType()`

vkryachko: how about making it protected instead?

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -180,8 +179,12 @@ String getTypeTokenLiteral() {
   public static class ArrayToken<T> extends TypeToken<T> {
```
vkryachko, line 179: I think you also need to update equals and hashCode now that it has a rawType.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -181,9 +179,12 @@ String getTypeTokenLiteral() {
...
     private final TypeToken<?> componentType;
 
-    private ArrayToken(Class<T> rawType, TypeToken<?> componentType) {
-      super.rawType = rawType;
-      this.componentType = componentType;
+    private ArrayToken(Class<T> rawType) {
+      super(rawType);
```
James201311, line 183: super() have to be the first statement in a constructor

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
James201311: Yes, I am thinking if we make it protected, anyone will be able to access it when they extent the `TypeToken`, maybe we loose the advantage of putting `ArrayToken` and `ClassToken` inside `TypeToken`? wdyt?

vkryachko: No one will be able to extend it as it has a private constructor

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/TypeToken.java
```diff
@@ -80,8 +91,8 @@
       }
       throw new IllegalArgumentException("<? super T> is not supported");
     } else if (type instanceof GenericArrayType) {
-      Type componentType = ((GenericArrayType) type).getGenericComponentType();
-      return new ArrayToken<T>(TypeToken.of(componentType));
+      throw new IllegalArgumentException(
+          "Generic Arrays are not supported, " + "please use Lists instead.");
```
vkryachko, line 95:
```suggestion
          "Generic Arrays are not supported, please use Lists instead.");
```

PR #1688, opened 2020-06-19
Functions unified emulator settings

samtstern:
Follow up to #1672 but for `FirebaseFunctions`

### DIFF
--- a/firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java
+++ b/firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java
@@ -25,6 +25,7 @@ import androidx.test.runner.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
+import com.google.firebase.emulators.EmulatedServiceSettings;
 import com.google.firebase.functions.FirebaseFunctionsException.Code;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -38,6 +39,8 @@ import org.junit.runner.RunWith;
 @RunWith(AndroidJUnit4.class)
 public class CallTest {
   private static FirebaseApp app;
+
+  private static final EmulatedServiceSettings NO_EMULATOR = null;
 
   @BeforeClass
   public static void setUp() {
@@ -91,7 +94,7 @@ public class CallTest {
               HttpsCallableContext context = new HttpsCallableContext("token", null);
               return Tasks.forResult(context);
             },
-            null);
+            NO_EMULATOR);
 
     HttpsCallableReference function = functions.getHttpsCallable("tokenTest");
     Task<HttpsCallableResult> result = function.call(new HashMap<>());
@@ -112,7 +115,7 @@ public class CallTest {
               HttpsCallableContext context = new HttpsCallableContext(null, "iid");
               return Tasks.forResult(context);
             },
-            null);
+            NO_EMULATOR);
 
     HttpsCallableReference function = functions.getHttpsCallable("instanceIdTest");
     Task<HttpsCallableResult> result = function.call(new HashMap<>());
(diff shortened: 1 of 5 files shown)


firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -47,6 +50,13 @@
 /** FirebaseFunctions lets you call Cloud Functions for Firebase. */
 public class FirebaseFunctions {
 
+  /**
+   * Emulator identifier, see {@link
+   * com.google.firebase.emulators.EmulatorSettings.Builder#addEmulatedService(FirebaseEmulator,
+   * EmulatedServiceSettings)}
+   */
+  public static final FirebaseEmulator EMULATOR = FirebaseEmulator.forName("functions");
```
samtstern, line 58: Should this require a new `api.txt`?  If so the bot did not warn me for the Database PR...

firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java
```diff
@@ -90,7 +90,8 @@ public void testToken() throws InterruptedException, ExecutionException {
             () -> {
               HttpsCallableContext context = new HttpsCallableContext("token", null);
               return Tasks.forResult(context);
-            });
+            },
+            null);
```
davidmotson, line 94: would it be possible to get a named variable for these nulls? Its unclear what exactly they mean for the test?

Something like "NO_CALLBACK" or something.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -75,13 +85,26 @@
   private String urlFormat = "https://%1$s-%2$s.cloudfunctions.net/%3$s";
 
   FirebaseFunctions(
-      Context context, String projectId, String region, ContextProvider contextProvider) {
+      Context context,
+      String projectId,
+      String region,
+      ContextProvider contextProvider,
+      @Nullable EmulatedServiceSettings emulatorSettings) {
```
davidmotson, line 92: Is there any value to holding this reference? Or is it only useful for this URL?

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
davidmotson: Unsure, Would it not fail tests/RC confirmation if that was the case?

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
samtstern: Yeah we don't need it after we use it.

firebase-functions/src/androidTest/java/com/google/firebase/functions/CallTest.java
samtstern: Good idea.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
rosariopfernandes: It should. Seems like the bot failed silently. See #1692

samtstern: @rosariopfernandes I'll add a fix for that here:
https://github.com/firebase/firebase-android-sdk/pull/1690

samtstern: I think I can add a fix for that here:
https://github.com/firebase/firebase-android-sdk/pull/1690

I don't think there will be a database release before that is resolved.

rosariopfernandes: @samtstern I'll change the base branch of my PR to the base branch of #1690

PR #1693, opened 2020-06-19
Cleaning up and addressing issues with FIAM docs

JasonAHeron:


firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/model/CardMessage.java
```diff
@@ -148,8 +148,8 @@ public Action getSecondaryAction() {
   }
 
   /**
-   * @deprecated Use getPrimaryAction() instead. Gets the primary {@link Action} associated with
-   *     this message
+   * @deprecated This method is deprecated. Use {@link #getPrimaryAction()} or {@link
```
rachelsaunders, line 151: I _think_ that by having the @deprecated that the "This method is deprecated." bit will automatically get generated when we generate the docs.

markarndt:
@JasonAHeron 
Please also check https://github.com/firebase/firebase-android-sdk/blob/master/firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/FirebaseInAppMessaging.java. In master and this branch, still have reference to Instance ID.

PR #1702, opened 2020-06-23
Update released version of database.

rlazo:


rlazo:
> Are there other versions that need to be updated too ?.

Not according to our tooling

PR #1705, opened 2020-06-23
Retry GoogleDataTransport Connection Errors

samedson:
- In cases where there is bad connection, the current behavior deletes the cached reports because they are caught as an `IOException`. They appear to be encoding errors, when in reality it's an error opening the `OutputStream` that is the connection.

 - This changes it so that when the `IOException` is due to connection problems, we retry the upload instead of deleting it. 

This is where we transform status numbers into success, fatal, and transient errors: [CctTransportBackend.java](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L329-L356)

 - 400-level errors (except 404) are treated as fatal errors, which are deleted
 - 500-level errors are treated as transient errors, which are retried

This is where we decide to delete fatal errors, and retry transient errors: [Uploader.java](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java#L128-L144)

### DIFF
--- a/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
+++ b/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
@@ -48,9 +48,11 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.net.ConnectException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.net.UnknownHostException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Calendar;
@@ -268,6 +270,9 @@ final class CctTransportBackend implements TransportBackend {
       // JsonWriter often writes one character at a time.
       dataEncoder.encode(
           request.requestBody, new BufferedWriter(new OutputStreamWriter(outputStream)));
+    } catch (ConnectException | UnknownHostException e) {
+      Logging.e(LOG_TAG, "Couldn't open connection, returning with 500", e);
+      return new HttpResponse(500, null, 0);
     } catch (EncodingException | IOException e) {
       Logging.e(LOG_TAG, "Couldn't encode request, returning with 400", e);
       return new HttpResponse(400, null, 0);


ashwinraghav:
Links were super helpful.
It seems like we model IOExceptions as transient errors [already](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L354)

This would result in retrying ? What am i missing

samedson:
@ashwinraghav  The issue is IOException from this error don't bubble up to that `try catch` because they're caught right here, and returned as a `HttpResponse(400, null, 0);`.

So when this error happens that we are testing, we get a valid [response](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L330) here. And then it becomes a [fatalError here](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L349) before, because it was a 400. Not since we return a `HttpResponse(500, null, 0);`, it becomes a [transientError](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L347)

ashwinraghav:
> @ashwinraghav The issue is IOException from this error don't bubble up to that `try catch` because they're caught right here, and returned as a `HttpResponse(400, null, 0);`.
> 
> So when this error happens that we are testing, we get a valid [response](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L330) here. And then it becomes a [fatalError here](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L349) before, because it was a 400. Not since we return a `HttpResponse(500, null, 0);`, it becomes a [transientError](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L347)

Makes sense.

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -268,6 +270,9 @@ private HttpResponse doSend(HttpRequest request) throws IOException {
       // JsonWriter often writes one character at a time.
       dataEncoder.encode(
           request.requestBody, new BufferedWriter(new OutputStreamWriter(outputStream)));
+    } catch (ConnectException | UnknownHostException e) {
```
ashwinraghav, line 273: Thinking out loud: This seems like a pretty conservative approach, as opposed to treating all IO Exceptions as retry-able.

ashwinraghav:
Can you add a test similar to https://github.com/firebase/firebase-android-sdk/blob/0ecb1148b1843d410bbf6db888115223cbf50ad0/transport/transport-backend-cct/src/test/java/com/google/android/datatransport/cct/CctTransportBackendTest.java#L299

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
samedson: Yeah exactly - the reason we did this is there are legit IO Exceptions that would not be retryable, eg. encoding errors of some sort

samedson:
@ashwinraghav we tried to write a test, but it seems really difficult with the current code setup, and with WireMock. We're gonna punt for now and sync up with Vlad on Friday to see if he knows a way.

PR #1690, opened 2020-06-19
Unified emulator settings for Firestore

samtstern:
Follow up to #1672 but for `FirebaseFirestore`

Note:

- This now throws an `IlegalStateException` if the developer tries to set the host through the `EmulatedServiceSettings` as well as `FirebaseFirestoreSettings`.  The settings method already throws unchecked exceptions so I thought this was the correct behavior, but I could convert it to a warning.

PR #1712, opened 2020-06-24
Un-hide EmulatorSettings API

samtstern:
Follow up to #1672 #1688 #1690

Questions:

- [x] Where is the CHANGELOG for `firebase-common` and `firebase-functions`?
- [x] Should I also bump the `-ktx` library versions?

### DIFF
--- a/firebase-common/api.txt
+++ b/firebase-common/api.txt
@@ -2,8 +2,10 @@
 package com.google.firebase {
 
   public class FirebaseApp {
+    method public void enableEmulators(@NonNull com.google.firebase.emulators.EmulatorSettings);
     method @NonNull public android.content.Context getApplicationContext();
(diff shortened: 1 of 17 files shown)


samtstern:
Assigning to @ashwinraghav for review since Vlad is out.

rlazo:
About bumping `-ktx`, please do. Thanks!

samtstern:
@rlazo where do I do that?  I don't see a `gradle.properties` for the KTX SDKs that contains a version number.  Do they even have their own versions or do they automatically inherit?

rlazo:
argh... my bad. Yes, they inherit the version from the parent's `version` var. No work needed!

rlazo:
Changelogs for `firebase-common` and `firebase-functions` are generated during the release process, you'll not be able to find them here.

PR #1717, opened 2020-06-25
Support Collections decoding

James201311:


encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -145,6 +148,83 @@
...
+    T map = null;
+    try {
+      map = classToken.getRawType().getDeclaredConstructor().newInstance();
+    } catch (Exception e) {
+      throw new IllegalArgumentException(
+          classToken.getRawType()
+              + " cannot be initialized.\n"
+              + "Do not pass abstract class and an interface.\n"
+              + e);
+    }
```
James201311, line 188: Are there better ways to deal with Exceptions?
or avoid using reflect?

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -145,6 +148,101 @@
...
+    T collection = null;
+    try {
+      collection = classToken.getRawType().getDeclaredConstructor().newInstance();
+    } catch (Exception e) {
+      throw new IllegalArgumentException(
+          classToken.getRawType()
+              + " cannot be initialized.\n"
+              + "Do not pass abstract class and an interface.\n"
+              + e);
+    }
```
James201311, line 214: same as above

rlazo: ?

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
vkryachko: We don't want to use reflection, I would say let's start with supporting `HashMap` and `LinkedHashMap` and add support for any others later if we have to.

In general users should be encouraged to have fields of interface type `Map` not concrete implementations, so I think supporting `Map`, `HashMap` and `LinkedHashMap` should be good enough.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -145,6 +160,109 @@
...
+    } else if (clazz.equals(PriorityQueue.class)) {
+      return (T) new PriorityQueue<>();
+    } else if (clazz.equals(HashSet.class) || clazz.equals(Set.class)) {
+      return (T) new HashSet<>();
+    } else if (clazz.equals(LinkedHashSet.class)) {
+      return (T) new LinkedHashSet<>();
+    } else if (clazz.equals(TreeSet.class) || clazz.equals(SortedSet.class)) {
+      return (T) new TreeSet<>();
+    }
+    throw new IllegalArgumentException(classToken + " not supported.");
```
vkryachko, line 240: Consider throwing `EncodingException` instead as this is the exception we advertise can be thrown.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -145,6 +160,109 @@
...
+      return (T) new HashMap<>();
+    } else if (clazz.equals(LinkedHashMap.class)) {
+      return (T) new LinkedHashMap<>();
+    } else if (clazz.equals(TreeMap.class) || clazz.equals(SortedMap.class)) {
+      return (T) new TreeMap<>();
+    } else if (clazz.equals(ArrayList.class) || clazz.equals(List.class)) {
+      return (T) new ArrayList<>();
+    } else if (clazz.equals(ArrayDeque.class)
+        || clazz.equals(Deque.class)
+        || clazz.equals(Queue.class)) {
```
vkryachko, line 229: Not sure how useful these are in the context of DTOs so we may want to remove them not to encourage folks to model queues with their json payloads, but I don't feel strongly.

Same for priority queue below.

James201311: Not sure if I understand correctly, the reason discouraging people from using queues is that the order of the decoded Json is not preserved?
But we should also support that if a client of Firestore doesn't want to have a dedicated network model(DTOs), but he wants to store the ordered results(get from orderby('x')) into a queue without mapping them.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -67,6 +80,8 @@
       return decodePrimitive(classToken);
     } else if (isSingleValue(classToken)) {
       return decodeSingleValue(classToken);
+    } else if (isPreDefinedObject(classToken)) {
+      return decodePreDefinedObject(classToken);
```
vkryachko, line 84: Looks like these 2 methods do more than one thing, consider splitting their functionality in to map and collection specific stuff, context: https://en.wikipedia.org/wiki/Single-responsibility_principle

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -145,6 +160,109 @@
...
+  }
+
+  private <K, V, T extends Map<K, V>> T decodeMap(TypeToken.ClassToken<T> classToken)
+      throws IOException {
+    TypeToken<K> keyTypeToken = classToken.getTypeArguments().at(0);
+    TypeToken<V> valTypeToken = classToken.getTypeArguments().at(1);
+    if (!isSingleValue(keyTypeToken))
+      throw new IllegalArgumentException(keyTypeToken + " cannot be used as Map key.");
+
+    T map = newInstance(classToken);
```
vkryachko, line 188: Same here, I would suggest you have 2 methods `newMap`, `newCollection`

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
vkryachko: makes sense

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -145,6 +160,109 @@
...
+    } else if (clazz.equals(Long.class)) {
+      return (K) (Long) Long.parseLong(keyLiteral);
+    } else if (clazz.equals(Character.class)) {
+      return (K) (Character) keyLiteral.charAt(0);
+    } else if (clazz.equals(Byte.class)) {
+      return (K) (Byte) Byte.parseByte(keyLiteral);
+    } else if (clazz.equals(String.class)) {
+      return (K) keyLiteral;
+    } else {
+      throw new IllegalArgumentException("Excepted Single Value Type. " + clazz + " was found.");
```
vkryachko, line 262: `EncodingException`

PR #1718, opened 2020-06-25
New Feature Suggestion for Cloud Firestore

allusai:
I would like to suggest a new feature for Firebase and Cloud Firestore. One thing I've seen a lot of Firebase developers say is that although Cloud Firestore's real-time write capabilities are helpful to them in most situations, often they have write requests that are unimportant and don't need to be done right away. One example of this is *logging*. Let's say a company wants to track what pages the user is visiting on their app in order to improve their app for the long-term. It's not too important to them to have this data be updated in real-time to the database, they just need it eventually. The same situation comes up with *monitoring users and data analytics*, there's a lot of frequent write requests that would be made to the database for data collection purposes, but there's no need for it to be updated in real-time with the full power that Firebase offers. Especially for big projects or popular apps, there's a lot of low-priority logging and collecting related tasks that doesn't directly affect users at all (and thus doesn't need to be done in real-time).

And these unimportant writes which will happen relatively frequently actually slow down the main app unnecessarily for the users. A lot of developers want a way where they can put off these low priority write requests for later and also conduct them in bulk (sending 10 logging related writes at once, rather than individually as soon as they are created). That would free up the main thread to work on other important tasks such as updating the UI or loading search results, rather than constantly being interrupted by the need to send low-priority logging updates to the backend/database.

That's why I developed this new feature that I call the **Lazy Database Uploader**. Basically whenever you want to write data to your Cloud Firestore backend, but that data isn't too important and you want to upload it as a batch with other requests, you can do that by adding it to the Lazy Database Uploader's queue and setting your threshold of how many requests you want to wait for before uploading all of them, in bulk. For example, for my app, I can say I want to wait for 5 logging-related writes to be requested and then I will execute all of them at once rather than interrupting my main thread every time a logging-related write comes up. When using this feature, I can also set a "maximum delay" which means I can say "after 30 seconds, if the threshold of 5 writes in the queue hasn't been met, go ahead and upload the 2 writes which are in the queue anyway." This prevents you from having to wait too long to get the data into your database.

I added the two source code files in my commit and I tried my best to adhere to Google's Java code style guidelines and use other files in the repo as a guide. Based on my research on Firebase users, I believe this feature can be really valuable to developers like me. If you have any further questions or suggestions for me, please let me know!

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/LDURequest.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/LDURequest.java
@@ -1,0 +1,27 @@
+import java.lang.String;
+import java.util.Map;
+
+/**
+* Represents a request in the LazyDatabaseUploader queue. It consists of the
+* name of collection the user wants to change, a document name, and the requestInfo
+* of their request. The requestInfo of their request is what usually goes inside
+* the parentheses of the db.collection(col).document(doc).set(requestInfo) method
+* It is a map with key type String and value type Object and it's the most important
+* part of the request (since it's what you want to write to the database).
+*/
+public class LDURequest {
+    private String collectionName;
+    private String documentName;
+    private Map<String, Object> requestInfo;
+   
+    public LDURequest(String colName, String docName, Map<String, Object> ct)
+    {
+        collectionName = colName;
+        documentName = docName;
+        requestInfo = ct;
+    }
+   
+    public String getColName() { return collectionName; }
+    public String getDocName() { return documentName; }
+    public Map<String,Object> getContent() { return requestInfo; }
+}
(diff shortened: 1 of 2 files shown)


wilhuff:
Hi @allusai, thanks for your interest!

While I appreciate where you're coming from on this, the Firestore API isn't the place to deal with this problem. There are many trade-offs to balance for this kind of solution, and while there may be users that have similar problems to the one you're describing, they may want different choices than the ones you've made here. For example, your change buffers pending writes in RAM, but these writes will be lost if the process is restarted before the buffer is flushed. Others may want a guarantee that writes are durably persisted. Similarly, while your change merely batches changes together, others may want to wait until the user is on Wi-Fi to minimize data charges.

Making a flexible API that solves all of these problems is a large problem space and falls well outside the charter of our project. We can't accept this PR because it starts us down the road of solving all these cases and this isn't something we have the resources to tackle.

Feel free to use `LazyDatabaseUploader` in your own projects or distribute it separately. Alternatively, consider using [`WorkManager`](https://developer.android.com/topic/libraries/architecture/workmanager) from Android Jetpack, which is specifically designed to address all these different cases. Also note that depending on how you're using the event data you're logging, you may benefit from logging these as custom events in Firebase Analytics, which is specifically designed to upload event data in an unobtrusive fashion.

allusai:
Hello @wilhuff, thank you for the consideration! That makes sense actually, I feel like the code I wrote might be just the initial seed of a much larger API which takes into account the tradeoffs and developer requirements that you mentioned. But you're right, to make a robust project that addresses those concerns isn't super easy. Thank you for your suggestions!

PR #1719, opened 2020-06-25
Repack apk files into zip format to generate stable size measurements.

yifanyang:


apk-size/apk-size.gradle
```diff
@@ -24,6 +24,10 @@ buildscript {
     }
 }
 
+plugins {
+    id "com.dorongold.task-tree" version "1.5"
```
vkryachko, line 28: What is this plugin for?

yifanyang: https://github.com/dorongold/gradle-task-tree

This is used to print out a tree of task dependencies. It is added to the `apk-size` project only. The main project will not be affected.

PR #1722, opened 2020-06-26
support default value decoding

James201311:


encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -274,6 +244,35 @@
     }
   }
 
+  @SuppressWarnings("unchecked")
+  private <T> T defaultValue(TypeToken<T> typeToken) {
+    if (typeToken instanceof TypeToken.ArrayToken) {
+      return convertGenericListToArray(null, (TypeToken.ArrayToken<T>) typeToken);
```
vkryachko, line 250: Does it have to be `null` here, could we use `Collections.emptyList()` instead?

James201311: Both of them works, but `Collections.emptyList()` is better, since it explicitly indicates the default value of array is `[]`. Great Call!

encoders/firebase-decoders-json/src/test/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContextDefaultValueTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+  @Test
+  public void defaultNumericValues_areDecodeCorrectly() throws IOException {
+    Map<Class<?>, ObjectDecoder<?>> objectDecoders = new HashMap<>();
+    objectDecoders.put(Foo.class, new FooObjectDecoder());
+    JsonDataDecoderBuilderContext jsonDataDecoderBuilderContext =
+        new JsonDataDecoderBuilderContext(objectDecoders);
+
+    String json =
+        "{\"i\":null, \"s\":null, \"l\":null, \"d\":null, \"f\":null, \"b\":null, \"c\":null, \"ii\":null, \"ss\":null, \"ll\":null, \"dd\":null, \"ff\":null, \"bb\":null, \"cc\":null, \"str\":null, \"obj\":null}";
```
vkryachko, line 164: Will defaults work for missing json values, i.e. if the above json did not contain "i" at all, would foo.i equal 0?

James201311: It won't, but we can support it by adding `getOrDefault` for each `get` in the `CreationContextImpl`.
My concern is that: if it missed the whole entry, will it still be considered as valid?

vkryachko: > if it missed the whole entry, will it still be considered as valid?

I think so, otherwise it will not be an optional field afaict. That said, I think it makes sense for `getInt` to return the default value unless the field is marked as required(which I don't think we have a notion for yet), it also makes `getInt` consistent with `get` which returns `null` today

James201311: Great call! 
Also, since `CreationContext` is the place to have value-related logic, and `DataDecoder` is the place to have decoding logic. 
I will move the default value logic into `CreationContext`, and `DataDecoder` will simply link the `fieldRef` with `null` value when the input missed the  entry or encountered `null` value .

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -87,8 +86,7 @@
     return convertGenericListToArray(list, arrayToken);
   }
 
-  private static <T, E> T convertGenericListToArray(
-      List<Object> list, TypeToken.ArrayToken<T> arrayToken) {
+  static <T, E> T convertGenericListToArray(List<Object> list, TypeToken.ArrayToken<T> arrayToken) {
```
James201311, line 89: this method works like a utility method, does it worth the efforts to move it into a helper file?

PR #1735, opened 2020-06-29
Initial Implementation of Decoder Builder

James201311:


James201311:
Builder should also have ability to set if the data decoder should ignore `null` value, will re-open this pr after https://github.com/firebase/firebase-android-sdk/pull/1722 is merged.

PR #1748, opened 2020-07-02
Fix kotlin lint issue.

rlazo:


rlazo:
the smoke-tests breakage is expected.

PR #1764, opened 2020-07-07
Patch caching fid in memory on successful registration.

ankitaj224:
Also, updating caching FID logic due to the reverted PR.


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -241,7 +248,11 @@ String getName() {
   public Task<InstallationTokenResult> getToken(boolean forceRefresh) {
     preConditionChecks();
     Task<InstallationTokenResult> task = addGetAuthTokenListener();
-    backgroundExecutor.execute(() -> doGetAuthToken(forceRefresh));
+    if (forceRefresh) {
+      backgroundExecutor.execute(this::doGetAuthTokenForceRefresh);
+    } else {
+      backgroundExecutor.execute(this::doGetAuthTokenWithoutForceRefresh);
+    }
```
andirayo, line 255: How about we just call `#doGetAuthToken` directly (or the new name respectively).
```
backgroundExecutor.execute(() -> this::doGetAuthToken(forceRefresh));
```

ankitaj224:
> As for the multi-threading with the tasks, executors, and listeners my understanding is limited, unfortunately. :/

No worries :) Fred reviewed those changes in another PR #1767

PR #1774, opened 2020-07-15
Adding a descriptive auth error message.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -109,6 +109,11 @@ public Thread newThread(Runnable r) {
           + "with Firebase server APIs: It identifies your application with Firebase."
           + "Please refer to https://firebase.google.com/support/privacy/init-options.";
 
+  private static final String AUTH_ERROR_MSG =
+      "Installation ID could not be validated with the Firebase servers (maybe deleted). "
```
andirayo, line 113: "The Installation ID [...]" or "Your Installation ID [...]"
I'm not sure, maybe without is best. :)

"[...] (maybe it was deleted)."

"[...] new Installation ID and auth token."

Issue #1680, opened 2020-06-18
160002227: Attempt to invoke virtual method 'com.google.firebase.inappmessaging.model.MessageType com.google.firebase.inappmessaging.model.InAppMessage.getMessageType()' on a null object reference


belrvn:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 3.6
  * Firebase Component: inappMessaging
  * Component version: 19.0.7
### [REQUIRED] Step 3: Describe the problem

Steps to reproduce:
I set up campaign which should show modal triggered by "on_foreground" +  "app_launch" + custom event from Tag Manager. When opening an application, on the main screen the crash happens. 

#### Steps to reproduce:

After publishing a test campaign the following error leads to app crash.

#### Relevant Code:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.google.firebase.inappmessaging.model.MessageType com.google.firebase.inappmessaging.model.InAppMessage.getMessageType()' on a null object reference
        at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.extractActions(FirebaseInAppMessagingDisplay.java:445)
        at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.inflateBinding(FirebaseInAppMessagingDisplay.java:313)
        at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.access$000(FirebaseInAppMessagingDisplay.java:81)
        at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay$1.run(FirebaseInAppMessagingDisplay.java:289)
        at android.os.Handler.handleCallback(Handler.java:907)
        at android.os.Handler.dispatchMessage(Handler.java:105)
        at android.os.Looper.loop(Looper.java:216)
        at android.app.ActivityThread.main(ActivityThread.java:7625)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:524)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:987)
```

Belka1000867:
https://github.com/firebase/firebase-android-sdk/issues/274

Belka1000867:
Not sure yet, but it looks like it happens for release version.. For debug version seems fine. Need to double check, but I will probably tell it more next week

CsabaMiomni:
[...]

PR #1779, opened 2020-07-17
Added null check before accessing message type

SadmanSarar:
Added null check before accessing message type in FirebaseInAppMessagingDisplay.java:259
This pull request should solve this [issue](https://github.com/firebase/firebase-android-sdk/issues/1680) (ref no. 160002227).

### DIFF
--- a/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
+++ b/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
@@ -256,6 +256,11 @@ public class FirebaseInAppMessagingDisplay extends FirebaseInAppMessagingDisplay
 
     final BindingWrapper bindingWrapper;
 
+    if (inAppMessage.getMessageType() == null) {
+      Logging.loge("The message type is null and not supported.");
+      return;
+    }
+
     switch (inAppMessage.getMessageType()) {
       case BANNER:
         bindingWrapper = bindingWrapperFactory.createBannerBindingWrapper(config, inAppMessage);


rosariopfernandes:
@SadmanSarar good work putting up a PR, but I feel like that's not the cause of #1680, because `null` message types will end up in the `default` branch of the switch case right below the `if` statement you added. And it does exactly the same thing you did:

https://github.com/firebase/firebase-android-sdk/blob/37dd2b18e3754b4dd3c2357b01e6b5f5a5f474cb/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java#L272-L275

---

The exception trace in that issue shows:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.google.firebase.inappmessaging.model.MessageType com.google.firebase.inappmessaging.model.InAppMessage.getMessageType()' on a null object reference
        at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.extractActions(FirebaseInAppMessagingDisplay.java:445)
        at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.inflateBinding(FirebaseInAppMessagingDisplay.java:313)
```
Which indicates that the `inAppMessage` object is null when passed as argument to the `extractActions` method:

https://github.com/firebase/firebase-android-sdk/blob/37dd2b18e3754b4dd3c2357b01e6b5f5a5f474cb/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java#L440-L442

I suggest we wait for more feedback from the author of #1680 before making any changes to the SDK.

Belka1000867:
> I suggest we wait for more feedback from the author of #1680 before making any changes to the SDK.

What kind of addition would you need?

shaifulroni:
@rosariopfernandes Why do you think the **getMessageType()** is null here? It clearly stated that the **message** object itself is null.

rosariopfernandes:
@shaifulroni that's what I'm saying. The `message` object is null, but this PR is checking if `getMessageType()` is null, which I don't think will fix the issue.

shaifulroni:
@rosariopfernandes Yes, as you were saying about the default branch in the switch/case and I missed the line in the middle. Sorry about that. Then again this PR should be close as it doesn't do anything actually.

PR #1802, opened 2020-07-22
Re-implement Unified Emulator Settings

samtstern:
History:
#1672 #1688 #1690 #1712 #1737 

This follows Proposal 2 at:
http://go/firebase-emulator-connection-api

TODO:
- [x] Regenerate `api.txt`
- [x] Version bumps


samtstern:
Not sure what's going on with the tests, every test I have run locally passes (connected and unit)

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -324,23 +321,14 @@ public static FirebaseApp initializeApp(
...
 
   /**
-   * Specify which services should access local emulators for this FirebaseApp instance.
-   *
-   * <p>For example, if the {@link EmulatorSettings} contain {@link
-   * com.google.firebase.emulators.EmulatedServiceSettings} for {@link FirebaseDatabase#EMULATOR},
-   * then calls to Cloud Firestore will communicate with the emulator rather than production.
-   *
-   * <p>TODO(samstern): Un-hide this once Firestore, Database, and Functions are implemented
+   * This should <b>only</b> ever be called from inside FirebaseFoo.useEmulator().
```
vkryachko, line 324: If it should now only be called by `useEmulator()` methods, why does FirebaseApp need to host this functionality as opposed to each sdk handling it individually?

vkryachko: If we want to have a central place for this config but it does not require a public API, then the recommended approach is to use Components as opposed to adding functionality to `FirebaseApp`

samtstern: I guess there's no need to centralize it anymore.  We may want it one day, but that's for then I guess.

samtstern: Actually it's pretty hard to make `FirebaseDatabase.getInstance(app, url)` work without this centralization ... so I'll do a component.

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
```diff
@@ -0,0 +1,55 @@
...
+  public EmulatedServiceSettings get(@NonNull String name) {
+    return settings.get(getKeyFor(name));
+  }
+
+  public void set(@NonNull String name, @Nullable EmulatedServiceSettings serviceSettings) {
+    settings.put(getKeyFor(name), serviceSettings);
+  }
+
+  private String getKeyFor(@NonNull String name) {
+    return this.app.getName() + '-' + name;
```
vkryachko, line 53: There will be one instance of `EmulatorSettings` per `FirebaseApp` so you don't need to prefix the map keys, and as a result it looks like you can drop the `FirebaseApp` dependency.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
vkryachko: hm actually, rtdb has [FirebaseDatabaseComponent](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabaseComponent.java), could it be used to host the emulator settings as opposed to having a component in firebase-common?

firebase-common/src/main/java/com/google/firebase/emulators/EmulatorSettings.java
samtstern: Sure.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
samtstern: Done.

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -202,7 +192,9 @@ public DatabaseReference getReferenceFromUrl(@NonNull String url) {
           "Can't pass null for argument 'url' in " + "FirebaseDatabase.getReferenceFromUrl()");
     }
 
-    ParsedUrl parsedUrl = Utilities.parseUrl(url, getEmulatorServiceSettings(this.app));
+    ParsedUrl parsedUrl =
+        Utilities.parseUrl(
+            url, getApp().get(FirebaseDatabaseComponent.class).getEmulatorSettings());
```
samtstern, line 197: @vkryachko what's the proper way to do this?  Lint says I shouldn't do this outside of `getInstance` but I can't store a reference to the component in `getInstance` since the instance comes from the component ...

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -302,9 +294,23 @@ public synchronized void setPersistenceCacheSizeBytes(long cacheSizeInBytes) {
...
+   */
+  public void useEmulator(@NonNull String host, int port) {
+    if (this.repo != null) {
+      throw new IllegalStateException(
+          "Cannot call useEmulator() after instance has already been initialized.");
+    }
+
+    getApp()
+        .get(FirebaseDatabaseComponent.class)
+        .setEmulatorSettings(new EmulatedServiceSettings(host, port));
```
schmidt-sebastian, line 313: Why do we need to round-trip through Firebase App? It seems like this could now stay within the SDK.

I am also not sure if I understand the merits of proposal 2. We now have two places next to each other which both configure settings that can end up conflicting. Isn't the thing that is actually missing from our settings object just an "isEmulator" property? Or even - is this something the Emulator could just tell us?

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
vkryachko: In this case `FirebaseDatabaseComponent` is the owner of `FirebaseDatabase` instances and responsible for creating them, so I would just recommend passing this component to the `FirebaseDatabase`'s constructor here:
https://github.com/firebase/firebase-android-sdk/blob/cc79cc638975677eadb5d742a5423312a82debf4/firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabaseComponent.java#L66

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
samtstern: RTDB has no settings object ... did you mean for Firestore?  

I am at the end of my wits here I truly don't care how we design this.  I only did Proposal #2 because I got told off after proposal #1 was already implemented and merged so I reverted it.  I can't justify this API because it's not mine.

The reason this has to be app-scoped is because if you call `FirebaseDatabase.getInstance(app, url)` we need to apply the same emulator settings.

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
samtstern: Ok so it's safe to pass the component down?  Thanks!

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
schmidt-sebastian: I think I meant the url parameter in `FirebaseDatabase.getInstance(app, url)`, which is RTDB's "Settings" object. I will review this PR for what it is later today (looks fine glancing over it without too much depth).

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -302,9 +294,23 @@ public synchronized void setPersistenceCacheSizeBytes(long cacheSizeInBytes) {
...
+  /**
+   * Modify this FirebaseDatabase instance to communicate with the Realtime Database emulator.
+   *
+   * <p>Note: this must be called before this instance has been used to do any database operations.
+   *
+   * @param host the emulator host (ex: 10.0.2.2)
+   * @param port the emulator port (ex: 9000)
+   */
+  public void useEmulator(@NonNull String host, int port) {
+    if (this.repo != null) {
```
vkryachko, line 306: What is the expected behavior in the following scenario?

```kotlin
val custom = Firebase.database("http://localhost:8000")
readAndWriteTo(custom)
val default = Firebase.database.useEmulator("example.com", 123)
```

afaict this will cause custom to keep talking to localhost, while default will talk to the emulator.

Do we want to throw in this case?

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
samtstern: No more going through app.

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
samtstern: No we do not.  That was something we could have done in the old "hold it all in FirebaseApp" proposal but it's not longer a feasible edge case to handle.

samtstern:
@schmidt-sebastian @vkryachko all review comments addressed, TODOs done, tests are now passing.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -217,6 +203,25 @@ public void setFirestoreSettings(@NonNull FirebaseFirestoreSettings settings) {
...
+   * @param port the emulator port (ex: 8080)
+   */
+  public void useEmulator(@NonNull String host, int port) {
+    if (this.client != null) {
+      throw new IllegalStateException(
+          "Cannot call useEmulator() after instance has already been initialized.");
+    }
+
+    this.emulatorSettings = new EmulatedServiceSettings(host, port);
+    ;
```
schmidt-sebastian, line 221: Stray semicolon

firebase-database/src/main/java/com/google/firebase/database/core/RepoInfo.java
```diff
@@ -61,6 +63,15 @@ public URI getConnectionURL(String optLastSessionId) {
     return URI.create(url);
   }
 
+  public void applyEmulatorSettings(@Nullable EmulatedServiceSettings settings) {
+    if (settings == null) {
+      return;
+    }
+
+    this.host = settings.getHost() + ":" + settings.getPort();
```
schmidt-sebastian, line 71: As discussed offline, `internalHost` needs to be updated as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -244,7 +249,7 @@ private FirebaseFirestoreSettings mergeEmulatorSettings(
 
     if (!FirebaseFirestoreSettings.DEFAULT_HOST.equals(settings.getHost())) {
       throw new IllegalStateException(
-          "Cannot specify the host in FirebaseFirestoreSettings when EmulatedServiceSettings is provided.");
+          "Cannot specify the host in FirebaseFirestoreSettings when emulator settings are provided.");
```
schmidt-sebastian, line 252: This is allowed in the RTDB proposal. No strong opinion here, but just wanted to point this out.

samtstern: So would you prefer the Firestore settings override?

schmidt-sebastian: I do, but I defer to your judgement.

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -202,7 +193,9 @@ public DatabaseReference getReferenceFromUrl(@NonNull String url) {
           "Can't pass null for argument 'url' in " + "FirebaseDatabase.getReferenceFromUrl()");
     }
 
-    ParsedUrl parsedUrl = Utilities.parseUrl(url, getEmulatorServiceSettings(this.app));
+    ParsedUrl parsedUrl = Utilities.parseUrl(url);
+    parsedUrl.repoInfo.applyEmulatorSettings(this.emulatorSettings);
```
schmidt-sebastian, line 197: I think this call needs to happen after the if statement below. You probably want to support this:

```
var db = FirebaseDatabase.getInstance('http://foo');
db.useEmulator('localhost');
db.getReferenceFromUrl('http://foo/bar');
````

Please add test as well.

File: firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java [hunk 10]
```diff
@@ -213,17 +209,24 @@ URL getURL(String function) {
...
-   * @param origin The origin of the local emulator, such as "http://10.0.2.2:5005".
-   */
+  /** @deprecated see {@link #useEmulator(String, int)} */
   public void useFunctionsEmulator(@NonNull String origin) {
     Preconditions.checkNotNull(origin, "origin cannot be null");
     urlFormat = origin + "/%2$s/%1$s/%3$s";
   }
 
+  /**
+   * Modify this FirebaseDatabase instance to communicate with the Cloud Functions emulator.
```
schmidt-sebastian, line 219: s/FirebaseDatabase/FirebaseFunctions/

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
samtstern: There is a test for this case actually.  `getReferenceFromUrl` is overridden by emulator settings.  We'll use the namespace from the URL, but the emulator host/port.

schmidt-sebastian: What about:

```
var db = FirebaseDatabase.getInstance('http://foo');
db.useEmulator('localhost');
db.getReferenceFromUrl('http://bar/baz');
```

That doesn't throw anymore, does it?

samtstern: Will add that to the test.

vkryachko:
looks like at least one of the failures is legit:
```
com.google.firebase.firestore.FirebaseFirestoreTest > getInstance_withEmulator_mergeSettingsFailure FAILED
    java.lang.AssertionError at FirebaseFirestoreTest.java:72
```

samtstern:
@vkryachko thanks you're right we no longer throw there.

PR #1803, opened 2020-07-22
Migrate reflective encoder to `FieldDescriptor`.

vkryachko:
The change also adds support for `@ExtraProperty`s and `@Ignore`.

### DIFF
--- a/encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoderProvider.java
+++ b/encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoderProvider.java
@@ -18,7 +18,6 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.firebase.encoders.EncodingException;
 import com.google.firebase.encoders.FieldDescriptor;
-import com.google.firebase.encoders.FieldDescriptor.Builder;
 import com.google.firebase.encoders.ObjectEncoder;
 import com.google.firebase.encoders.ObjectEncoderContext;
 import com.google.firebase.encoders.annotations.Encodable;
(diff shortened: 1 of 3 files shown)


encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoderProvider.java
```diff
@@ -63,26 +68,48 @@ public void encode(@Nullable Object obj, @NonNull ObjectEncoderContext ctx) thro
   @NonNull
   @Override
   public <T> ObjectEncoder<T> get(@NonNull Class<T> type) {
-    Map<String, EncodingDescriptor> fields = new HashMap<>();
+    Map<FieldDescriptor, EncodingDescriptor> fields = new LinkedHashMap<>();
```
James201311, line 71: wondering what would be the use case of `LinkedHashMap`?  are we going to assume the decoded json order to be associated with its getter methods?

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoderProvider.java
```diff
@@ -63,26 +68,48 @@ public void encode(@Nullable Object obj, @NonNull ObjectEncoderContext ctx) thro
...
+      return null;
+    }
+
+    String name = toGetterName(method, fieldAnnotation);
+    if (name == null) {
+      return null;
+    }
+    FieldDescriptor.Builder builder = FieldDescriptor.builder(name);
+    for (Annotation annotation : method.getAnnotations()) {
+      if (!annotation.annotationType().isAnnotationPresent(ExtraProperty.class)) {
```
James201311, line 105: if we have Annotation Processor implemented, should we just check the annotation type is registered? in that case, we may not need `ExtraProperty`?

vkryachko: we still need `ExtraProperty` for the code-gen decoders as at compile time we don't yet know what annotations will be registered at runtime.

Additionally, having this encoder implementation make assumptions about the environment it is running it would increase coupling to that environment, not something we want.

Also note that this encoder is a singleton and it caches information about types, so it can be run by differently configured encoder contexts.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectEncoderProvider.java
vkryachko: oops, that was not intended. removed

PR #1813, opened 2020-07-23
inital implementation of AnnotationFieldHandler

James201311:


encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
```diff
@@ -64,7 +64,12 @@ public String getName() {
   @Nullable
   @SuppressWarnings("unchecked")
   public <T extends Annotation> T getProperty(@NonNull Class<T> type) {
-    return (T) properties.get(type);
+    return type.cast(properties.get(type));
+  }
+
+  @NonNull
+  public Map<Class<? extends Annotation>, Annotation> getProperties() {
```
vkryachko, line 71: The map is an implementation detail, let's only expose the `Set<Annotation>`.

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
```diff
@@ -64,7 +64,12 @@ public String getName() {
   @Nullable
   @SuppressWarnings("unchecked")
   public <T extends Annotation> T getProperty(@NonNull Class<T> type) {
-    return (T) properties.get(type);
+    return type.cast(properties.get(type));
```
vkryachko, line 67: curious why this change was needed?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon May 21 15:18:27 PDT 2018
+#Wed Jul 22 15:50:08 ADT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
```
vkryachko, line 6: pls revert this change

build.gradle
```diff
@@ -29,7 +29,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.2'
+        classpath 'com.android.tools.build:gradle:3.6.1'
```
vkryachko, line 32: pls revert this change

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -38,11 +42,19 @@
       new HashMap<>();
   private Map<TypeToken.ClassToken<?>, TypeCreator<?>> typeCreators = new HashMap<>();
   private JsonReader reader;
+  private Map<Class<?>, FieldModifier<?>> fieldModifier = new HashMap<>();
```
vkryachko, line 44: let's make this final, as well as objectDecoders above.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -38,11 +42,19 @@
...
 
   JsonDataDecoderBuilderContext(@NonNull Map<Class<?>, ObjectDecoder<?>> objectDecoders) {
     this.objectDecoders = objectDecoders;
   }
 
+  JsonDataDecoderBuilderContext(
+      @NonNull Map<Class<?>, ObjectDecoder<?>> objectDecoders,
+      Map<Class<?>, FieldModifier<?>> fieldModifier) {
```
vkryachko, line 52: I assume this argument should be `@NonNull` as well?

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -166,15 +178,42 @@
...
+    }
+    return val;
+  }
+
+  private <TField, U extends Annotation> TField applyFieldModifier(Class<U> annotationType, FieldDescriptor fieldDescriptor, TField value, Class<TField> fieldType) {
+    @SuppressWarnings("unchecked") //Safe, because each entry in fieldModifier always have the same type parameter
+    FieldModifier<U> modifier = (FieldModifier<U>) fieldModifier.get(annotationType);
+    U annotation = fieldDescriptor.getProperty(annotationType);
+    if (modifier == null)
+      throw new IllegalArgumentException("FieldModifier: " + annotationType + " is not registered.");
```
vkryachko: I don't think we should throw in this case. I might not care about a particular annotation but still want to be able to decode.

A hypothetical example: decode pojo from a firebase function call and save into firestore. the pojo could contain `@DocumentId`, but functions don't care about it, so they will just ignore it as opposed to failing to decode.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -166,15 +178,42 @@
...
+      val = applyFieldModifier(annotationType, fieldDescriptor, val, fieldTypeToken.getRawType());
+    }
+    return val;
+  }
+
+  private <TField, U extends Annotation> TField applyFieldModifier(Class<U> annotationType, FieldDescriptor fieldDescriptor, TField value, Class<TField> fieldType) {
+    @SuppressWarnings("unchecked") //Safe, because each entry in fieldModifier always have the same type parameter
+    FieldModifier<U> modifier = (FieldModifier<U>) fieldModifier.get(annotationType);
+    U annotation = fieldDescriptor.getProperty(annotationType);
+    if (modifier == null)
```
vkryachko, line 217: pls use curly braces everywhere: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -166,15 +178,42 @@
...
+    FieldDescriptor fieldDescriptor = decoderCtx.getFieldDescriptors(fieldName);
+    Object val = decodeField(fieldRef.getTypeToken(), fieldDescriptor);
+    creationCtx.put(fieldRef, val);
+  }
+
+  private <TField, U extends Annotation> TField decodeField(TypeToken<TField> fieldTypeToken, FieldDescriptor fieldDescriptor) throws IOException {
+    TField val = decode(fieldTypeToken);
+
+    //apply annotation modifier
+    for (Class<? extends Annotation> annotationType: fieldDescriptor.getProperties().keySet()) {
```
vkryachko: instead of iterating the types, I suggest you iterate the values, then you don't need to pass the `fieldDescriptor` to `applyFieldModifier()`.

To get the type of the annotation you can just do `annotation.annotationType()`.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/ObjectDecoderContextImpl.java
```diff
@@ -47,6 +48,15 @@ private ObjectDecoderContextImpl(TypeToken.ClassToken<T> classToken) {
     }
   }
 
+  @NonNull
+  FieldDescriptor getFieldDescriptors(@NonNull String fieldName) {
+    if (fieldDescriptors.containsKey(fieldName)) {
+      return fieldDescriptors.get(fieldName);
+    } else {
+      throw new IllegalArgumentException(fieldName + " was not register in ObjectDecoder.");
```
vkryachko, line 56: I don't think `IllegalArgumentException` is appropriate here, it usually indicates that the caller of the library provided some incorrect input. In this case it seems that the only time this could happen is if there is a bug in the library implementation, pls double check if that is the case and if so, change to `AssertionError`.

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
James201311: sorry, that was not intended. will remove it.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/ObjectDecoderContextImpl.java
James201311: Great Point here! I totally agree. it should never be null as long as the `fieldName` have associated `FieldRef`, in that case, it would make sense to make `FieldRef` as the key of the `fieldDescriptors`.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/FieldModifier.java
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.decoders;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import java.lang.annotation.Annotation;
+
+public interface FieldModifier<U extends Annotation> {
```
vkryachko, line 21: nit: since it only handles annotated fields, would it make sense to name it like `AnnotatedFieldHandler`?

encoders/firebase-decoders-json/src/test/java/com/google/firebase/decoders/json/JsonDataDecoderExtraPropertyTest.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  @Test
+  public void customizedAnnotation_shouldProcessCorrectly() throws IOException {
+    Map<Class<?>, ObjectDecoder<?>> objectDecoders = new HashMap<>();
+    Map<Class<?>, FieldModifier<?>> processors = new HashMap<>();
+
+    objectDecoders.put(Foo.class, new FooObjectDecoder());
+    processors.put(Default.class, processor);
```
vkryachko, line 116: I think you can just replace the processor with an inline lambda expression here, wdyt?

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilderContext.java
```diff
@@ -46,14 +50,23 @@
...
+  private final Map<Class<?>, ObjectDecoder<?>> objectDecoders;
   private Map<TypeToken.ClassToken<?>, ObjectDecoderContextImpl<?>> objectDecoderContexts =
       new HashMap<>();
   private Map<TypeToken.ClassToken<?>, TypeCreator<?>> typeCreators = new HashMap<>();
   private JsonReader reader;
+  private final Map<Class<?>, FieldModifier<?>> fieldModifier;
 
   JsonDataDecoderBuilderContext(@NonNull Map<Class<?>, ObjectDecoder<?>> objectDecoders) {
     this.objectDecoders = objectDecoders;
+    this.fieldModifier = Collections.emptyMap();
```
vkryachko, line 62:
```suggestion
    this(objectDecoders, Collections.emptyMap());
```

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
vkryachko: looks like you forgot to remove it

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
```diff
@@ -62,9 +64,13 @@ public String getName() {
...
   @Nullable
-  @SuppressWarnings("unchecked")
   public <T extends Annotation> T getProperty(@NonNull Class<T> type) {
-    return (T) properties.get(type);
+    return type.cast(properties.get(type));
+  }
+
+  @NonNull
+  public Set<Annotation> getAllAnnotations() {
+    return new HashSet<>(properties.values());
```
vkryachko, line 73: Instead of copying, consider using `Collections.unmodifiableSet(properties.values())`, this would prevent modification and avoid a full copy.

vkryachko: furthermore, I think it would be even cleaner if you stored the properties above as `this.properties = Collections.unmodifiableMap(properties)`. then you don't need to allocate a new unmodifiableSet here at all, just `return properties.values()`

encoders/firebase-decoders-json/src/test/java/com/google/firebase/decoders/json/JsonDataDecoderExtraPropertyTest.java
James201311: just tried to place them in an inline lambda expression, but seems like the compiler won’t be able to determine those generic types where we have to cast them multiple times, will use an anonymous class instead.

encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/AnnotatedFieldHandler.java
```diff
@@ -18,7 +18,21 @@
 import androidx.annotation.Nullable;
 import java.lang.annotation.Annotation;
 
-public interface FieldModifier<U extends Annotation> {
+/**
+ * {@link AnnotatedFieldHandler} provide a way to handle field values based on its annotation.
+ *
+ * <p>Each annotation type should be registered along with associated {@link AnnotatedFieldHandler},
+ * non-registered annotation will be skipped in the process of decoding.
```
vkryachko, line 25: Worth also noting that annotations themselves need to be annotated with `@ExtraProperty`

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/FieldDescriptor.java
```diff
@@ -64,13 +63,14 @@ public String getName() {
...
   public <T extends Annotation> T getProperty(@NonNull Class<T> type) {
-    return type.cast(properties.get(type));
+    return (T) properties.get(type);
   }
 
   @NonNull
-  public Set<Annotation> getAllAnnotations() {
-    return new HashSet<>(properties.values());
+  public Collection<Annotation> getAllAnnotations() {
+    return properties.values();
```
vkryachko, line 73: please also change the constructor to `this.properties = Collections.unmodifiableMap(properties)`

James201311: the only place called the private constructor is the `build` method, and `Collections.unmodifiableMap(properties)` was applied to the `properties` before passing to the constructor.

PR #1815, opened 2020-07-23
initial implementation of decoder builder

James201311:


encoders/firebase-decoders-json/src/main/java/com/google/firebase/decoders/DecoderConfig.java
```diff
@@ -0,0 +1,27 @@
...
+import androidx.annotation.NonNull;
+
+/**
+ * Implemented by concrete {@link DataDecoder} builders.
+ *
+ * <p>Used by clients to configure decoders without coupling to a particular decoder format.
+ */
+public interface DecoderConfig<T extends DecoderConfig<T>> {
+  @NonNull
+  <U> T register(@NonNull Class<U> clazz, @NonNull ObjectDecoder<U> objectDecoder);
```
vkryachko, line 26: Would it make sense to allow the following?
```suggestion
  <U> T register(@NonNull Class<U> clazz, @NonNull ObjectDecoder<? extends U> objectDecoder);
```

I can't see how it would cause problems in principle. i.e. when a class `T` is needed, it is ok to return a subclass of that at the type level. Not sure if it would be useful in practice, but seems pretty cheap to change.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+  @Override
+  public <T> JsonDataDecoderBuilder register(
+      @NonNull Class<T> clazz, @NonNull ObjectDecoder<T> objectDecoder) {
+    objectDecoders.put(clazz, objectDecoder);
+    return this;
+  }
+
+  @NonNull
+  public DataDecoder build() {
+    return new JsonDataDecoderBuilderContext(objectDecoders);
```
vkryachko, line 38: nit:
```suggestion
    return new JsonDataDecoderContext(objectDecoders);
```

Let's rename this class as it's not a builder.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.google.firebase.decoders.json;
+
+import androidx.annotation.NonNull;
+import com.google.firebase.decoders.DataDecoder;
+import com.google.firebase.decoders.DecoderConfig;
+import com.google.firebase.decoders.ObjectDecoder;
+import java.util.HashMap;
+
+public final class JsonDataDecoderBuilder implements DecoderConfig<JsonDataDecoderBuilder> {
```
vkryachko, line 23: is my understanding correct that you will add ability to register annotation processors later?

James201311: yes, it will be able to register `AnnotationFieldHandler`, and `fallbackObjectDecoder`.

PR #1819, opened 2020-07-24
Reflective Decoder with Delegation(Fully customized)

James201311:


James201311:
Go with Annotation Field Handler approach, keep a record here.

PR #1827, opened 2020-07-28
initial implementation of reflective decoder

James201311:


encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
```diff
@@ -0,0 +1,214 @@
...
+    private final Map<String, ReflectiveFieldSetter> fieldSetters = new HashMap<>();
+
+    private ReflectiveObjectDecoderImpl() {}
+
+    @NonNull
+    @Override
+    public TypeCreator<T> decode(@NonNull ObjectDecoderContext<T> ctx) {
+      Class<T> clazz = ctx.getTypeToken().getRawType();
+      readMethods(clazz);
+      readFields(clazz);
```
vkryachko, line 60: Will fields overwrite the methods that were just read above?

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
```diff
@@ -0,0 +1,214 @@
...
+        for (Method method : methods) {
+          if (!shouldIncludeSetter(method)) {
+            continue;
+          }
+
+          String fieldName = fieldName(method);
+          if (descriptors.get(fieldName) == null) {
+            descriptors.put(
+                fieldName,
+                buildFieldDescriptor(decodingKey(method), method.getDeclaredAnnotations()));
```
vkryachko, line 78: We don't want to include all annotations in the descriptor, only ones that are themselves annotated with `@ExtraProperty`.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
```diff
@@ -0,0 +1,214 @@
...
+      while (currentClass != Object.class && currentClass != null) {
+        for (Field field : currentClass.getDeclaredFields()) {
+          if (!shouldIncludeField(field)) {
+            continue;
+          }
+          String fieldName = field.getName();
+          if (descriptors.get(fieldName) == null) {
+            descriptors.put(
+                fieldName,
+                buildFieldDescriptor(decodingKey(field), field.getDeclaredAnnotations()));
```
vkryachko, line 99: same here

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
James201311: no, it won't. Fields will only be written when their associated setter methods were not found.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
James201311: Great, then `@Encodable.Field` should be annoated with `@ExtraProperty` as well, since it is the extra property of the field even if it's an internal annotation.

encoders/firebase-encoders-json/src/main/java/com/google/firebase/encoders/annotations/Encodable.java
```diff
@@ -48,6 +48,7 @@
   /** Field configuration. */
   @Target(ElementType.METHOD)
   @Retention(RetentionPolicy.RUNTIME)
+  @ExtraProperty
```
vkryachko, line 51:
```suggestion
```

I don't think we want Field itseld to have this annotation.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveDecoderHelper.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+    return builder.build();
+  }
+
+  static boolean isInLine(FieldDescriptor fieldDescriptor) {
+    Encodable.Field annotation = fieldDescriptor.getProperty(Encodable.Field.class);
+    return annotation != null && annotation.inline();
+  }
+
+  static String fieldName(Method method) {
```
vkryachko, line 45: We want to make sure the field name is consistent with whatever the reflective *encoder* does, can you make sure they are indeed consistent and if not, reconsile them?

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveFieldSetter.java
```diff
@@ -0,0 +1,101 @@
...
+package com.google.firebase.encoders.reflective;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+
+abstract class ReflectiveFieldSetter {
```
vkryachko, line 24: please make this an interface

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveFieldSetter.java
```diff
@@ -0,0 +1,101 @@
...
+  static ReflectiveFieldSetter of(@NonNull Field field) {
+    return new FieldSetter(field);
+  }
+
+  @NonNull
+  static ReflectiveFieldSetter of(@NonNull Method setter) {
+    return new MethodSetter(setter);
+  }
+
+  static class FieldSetter extends ReflectiveFieldSetter {
```
vkryachko, line 44: extract Field and MethodSetters into top level classes pls.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveDecoderHelper.java
James201311: yes, they are consistent, but i will change the implementation to align with encoder as well.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
```diff
@@ -0,0 +1,210 @@
...
+  @Override
+  public <T> ObjectDecoder<T> get(@NonNull Class<T> clazz) {
+    return new ReflectiveObjectDecoderImpl<>();
+  }
+
+  private static class ReflectiveObjectDecoderImpl<T> implements ObjectDecoder<T> {
+
+    private final Map<String, FieldRef<?>> refs = new HashMap<>();
+    private final Map<String, FieldDescriptor> descriptors = new HashMap<>();
+    private final Map<String, ReflectiveSetter> reflectiveSetters = new HashMap<>();
```
vkryachko, line 51: This looks like a lot of bookkeeping spread across 3 maps, is it ever the case that only a subset of these maps contains a certain key? if not consider merging into one map

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
```diff
@@ -0,0 +1,210 @@
...
+        refs.put(fieldName, ref);
+      }
+    }
+
+    private TypeToken<?> getFieldTypeToken(Type type, ObjectDecoderContext<?> ctx) {
+      if (type instanceof TypeVariable) {
+        TypeVariable[] typeVariables = ctx.getTypeToken().getRawType().getTypeParameters();
+        for (int i = 0; i < typeVariables.length; i++) {
+          if (typeVariables[i].equals(type)) {
+            return ctx.getTypeArgument(i);
```
vkryachko, line 152: don't we want to throw a meaningful exception in the case we could not find the TypeVariable in the class definition? By "meaningful" I mean, not the generic on that `TypeToken.of` will throw, but rather something along the lines of "class foo<T,U,V> does not have type variable W".

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveInitializer.java
```diff
@@ -0,0 +1,42 @@
...
+final class ReflectiveInitializer {
+
+  @NonNull
+  static <T> T newInstance(TypeToken.ClassToken<T> classToken) {
+    try {
+      Constructor<T> constructor = classToken.getRawType().getDeclaredConstructor();
+      return constructor.newInstance();
+    } catch (NoSuchMethodException e) {
+      // TODO: try JVM sun.misc.Unsafe to allocate an instance
+      throw new RuntimeException(e);
```
vkryachko, line 31: This is not a very good exception, let's tell the developer what is wrong in each case and how they could fix it.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveSetter.java
```diff
@@ -0,0 +1,31 @@
...
+interface ReflectiveSetter {
+  void set(@NonNull Object obj, @Nullable Object val);
+
+  @NonNull
+  Class<?> getFieldRawType();
+
+  @Nullable
+  Type getFieldGenericType();
+
+  boolean isDecodedInline();
```
vkryachko, line 30: food fo thought(no immediate action required): it seems that it could be useful to just store the `inline` flag inside the field descriptor. This will help by:
* not needing a method here
* and furthermore would allow us to remove the `decodeInline()` method altogether making the api surface smaller and the implementation below simpler.

wdyt?

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveObjectDecoderProvider.java
James201311: Great call, I will wrap them into a `ReflectiveDecoderFieldContext`, which will also make the `ReflectiveSetter` be a single purpose setter.

encoders/firebase-encoders-reflective/firebase-encoders-reflective.gradle
```diff
@@ -52,5 +52,6 @@ dependencies {
     testImplementation "com.google.truth:truth:$googleTruthVersion"
     testImplementation 'org.mockito:mockito-core:2.25.0'
     testImplementation 'com.fasterxml.jackson.core:jackson-databind:2.11.1'
+    implementation project(path: ':encoders:firebase-decoders-json')
```
vkryachko, line 55:
```suggestion
    implementation project(':encoders:firebase-decoders-json')
```

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveInitializer.java
```diff
@@ -0,0 +1,47 @@
...
+
+final class ReflectiveInitializer {
+
+  @NonNull
+  static <T> T newInstance(TypeToken.ClassToken<T> classToken) {
+    try {
+      Constructor<T> constructor = classToken.getRawType().getDeclaredConstructor();
+      return constructor.newInstance();
+    } catch (NoSuchMethodException e) {
+      // TODO: try JVM sun.misc.Unsafe to allocate an instance
```
vkryachko, line 31: is it something you intend to actually do, if not consider removing this todo

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveInitializer.java
```diff
@@ -0,0 +1,47 @@
...
+              + classToken
+              + " does not define a no-argument constructor. If you are using ProGuard, make "
+              + "sure these constructors are not stripped.");
+    } catch (IllegalAccessException e) {
+      throw new RuntimeException(e);
+    } catch (InstantiationException e) {
+      throw new RuntimeException(e);
+    } catch (InvocationTargetException e) {
+      throw new RuntimeException(e);
+    }
```
vkryachko, line 43: Looks like you missed all these cases, like constructor is not public, etc.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveDecoderFieldContextImp.java
```diff
@@ -0,0 +1,181 @@
...
+import java.lang.reflect.AccessibleObject;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+class ReflectiveDecoderFieldContextImp<T> implements ReflectiveDecoderFieldContext<T> {
```
vkryachko, line 32: nit: it's common to name these things with `Impl`, not `Imp`
```suggestion
class ReflectiveDecoderFieldContextImpl<T> implements ReflectiveDecoderFieldContext<T> {
```

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveInitializer.java
```diff
@@ -35,11 +36,11 @@
               + " does not define a no-argument constructor. If you are using ProGuard, make "
               + "sure these constructors are not stripped.");
     } catch (IllegalAccessException e) {
-      throw new RuntimeException(e);
+      throw new EncodingException(e.toString());
```
vkryachko, line 39: We have to preserve the cause chain to make the error more useful, I would suggest something like this in all cases below:
```java
} catch (IllegalAccessException | InstantiationException | InvocationTargetException e) {
  throw new EncodingException("Could not instantiate type " + classToken, e);
}
```

not the second parameter to the exception constructor.

vkryachko: pls also add the `e` argument to the catch block above.

James201311: Multi-catch requires API level 19.

vkryachko: No really, if you set language level to java 8, the desugarer will kick in and desugar to java 6.

i.e.  see https://github.com/firebase/firebase-android-sdk/blob/master/firebase-common/firebase-common.gradle#L49

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ObjectDecoderProvider.java
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.encoders.reflective;
+
+import androidx.annotation.NonNull;
+import com.google.firebase.decoders.ObjectDecoder;
+
+interface ObjectDecoderProvider {
```
vkryachko, line 20: Looks like you're not using it for anything, consider removing altogether

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ReflectiveDecoderFieldContext.java
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.encoders.reflective;
+
+import com.google.firebase.decoders.FieldRef;
+import com.google.firebase.encoders.FieldDescriptor;
+import java.lang.reflect.Type;
+
+interface ReflectiveDecoderFieldContext<T> {
```
vkryachko, line 21: Same here, consider removing the interface and renaming the `Impl` class to this name.

encoders/firebase-encoders-reflective/src/main/java/com/google/firebase/encoders/reflective/ObjectDecoderProvider.java
James201311: I was trying to align with the behaviour of `ReflectiveEncoder`, but it turns out it's redundant. I will remove them.

encoders/firebase-decoders-json/src/json/java/com/google/firebase/decoders/json/JsonDataDecoderBuilder.java
```diff
@@ -20,15 +20,21 @@
...
 import com.google.firebase.decoders.ValueDecoder;
+import com.google.firebase.encoders.EncodingException;
 import java.lang.annotation.Annotation;
 import java.util.HashMap;
 import java.util.Map;
 
 public final class JsonDataDecoderBuilder implements DecoderConfig<JsonDataDecoderBuilder> {
+  private static final ObjectDecoder<Object> DEFAULT_FALLBACK_DECODER =
+      (ctx) -> {
+        throw new EncodingException(ctx.getTypeToken().getRawType() + " is not register.");
```
vkryachko, line 31:
```suggestion
        throw new EncodingException(ctx.getTypeToken().getRawType() + " is not registered.");
```

PR #1831, opened 2020-07-29
Add platform logging

VinayGuthal:


VinayGuthal:
/test device-check-changed
Details

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatLogSource.java
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package com.google.firebase.heartbeatinfo;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nonnull;
+
+/** The class is not public to ensure other components cannot depend on it. */
+@AutoValue
+abstract class HeartBeatLogSource {
```
vkryachko, line 22: To me it sounds like a "consumer" rather than a logSource, I would suggest you rename it to `HeartBeatConsumer`.

vkryachko: Additionally it could be just an empty interface rather than a class(you would have to make it public in this case).

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatLogSource.java
```diff
@@ -0,0 +1,29 @@
...
+  static HeartBeatLogSource create(String name) {
+    return new AutoValue_HeartBeatLogSource(name);
+  }
+
+  @Nonnull
+  public abstract String getLogSourceName();
```
vkryachko, line 28: this value is not used anywhere, consider removing altogether.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatLogSourceComponent.java
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.heartbeatinfo;
+
+import com.google.firebase.components.Component;
+
+/** Factory to create a component that publishes the version of an SDK */
```
vkryachko, line 19: pls change the javadocs in this class

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -20,20 +20,27 @@
...
 public class DefaultHeartBeatInfo implements HeartBeatInfo {
 
   private HeartBeatInfoStorage storage;
 
-  private DefaultHeartBeatInfo(Context context) {
+  private boolean doesSourceExist;
+
+  private DefaultHeartBeatInfo(Context context, Set<HeartBeatLogSource> logSources) {
     storage = HeartBeatInfoStorage.getInstance(context);
+    doesSourceExist = (logSources.size() > 0);
```
vkryachko, line 36: you may want to hold on to the set and check if it's empty every time, this will make it future proof for when we add support for dynamic loading.

vkryachko, line 36: nit: delegate to the other constructor to avoid logic duplication.
```suggestion
    this(HeartBeatInfoStorage.getInstance(context), doesSourceExist = (logSources.size() > 0));
```

VinayGuthal: ahh this is not possible since the other class can only be accessed via tests.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -51,10 +60,59 @@ private DefaultHeartBeatInfo(Context context) {
     return HeartBeat.NONE;
   }
 
+  @Override
+  public Task<List<HeartBeatResult>> getAndClearStoredHeartBeatInfo() {
```
vkryachko, line 64: iiuc you just wrapped the result in a `Task` but did not actually make it asynchronous, so it will block anyway.

VinayGuthal: Oops sorry about that. Updated

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -18,22 +18,57 @@
...
   private HeartBeatInfoStorage storage;
 
-  private DefaultHeartBeatInfo(Context context) {
+  private Set<HeartBeatConsumer> consumers;
+
+  private ExecutorService backgroundExecutor;
+
+  private static final ThreadFactory THREAD_FACTORY =
+      new ThreadFactory() {
+        private final AtomicInteger mCount = new AtomicInteger(1);
```
vkryachko, line 46: why do use a counter if the thread pool has at most one thread?

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -18,22 +18,57 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /** Provides information as whether to send heart beat or not. */
 public class DefaultHeartBeatInfo implements HeartBeatInfo {
 
   private HeartBeatInfoStorage storage;
```
vkryachko, line 38:
```suggestion
  private final HeartBeatInfoStorage storage;
```

here and throughout

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -18,22 +18,57 @@
...
 
-  private DefaultHeartBeatInfo(Context context) {
+  private Set<HeartBeatConsumer> consumers;
+
+  private ExecutorService backgroundExecutor;
```
vkryachko, line 42:
```suggestion
  private Executor backgroundExecutor;
```

since you only use the `execute()` method

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -18,22 +18,57 @@
...
+        private final AtomicInteger mCount = new AtomicInteger(1);
+
+        @Override
+        public Thread newThread(Runnable r) {
+          return new Thread(
+              r, String.format("heartbeat-information-executor-%d", mCount.getAndIncrement()));
+        }
+      };
+
+  private DefaultHeartBeatInfo(Context context, Set<HeartBeatConsumer> consumers) {
```
vkryachko, line 55: pls have this constructor delegate to the one below

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -51,10 +86,69 @@ private DefaultHeartBeatInfo(Context context) {
...
+  @Override
+  public Task<List<HeartBeatResult>> getAndClearStoredHeartBeatInfo() {
+    TaskCompletionSource<List<HeartBeatResult>> taskCompletionSource = new TaskCompletionSource<>();
+    this.backgroundExecutor.execute(
+        () -> {
+          ArrayList<HeartBeatResult> heartBeatResults = new ArrayList<>();
+          boolean shouldSendGlobalHeartBeat = false;
+          List<SdkHeartBeatResult> sdkHeartBeatResults = storage.getStoredHeartBeats(true);
+          long lastGlobalHeartBeat = storage.getLastGlobalHeartBeat();
+          for (int i = 0; i < sdkHeartBeatResults.size(); i++) {
```
vkryachko, line 98: if you don't need the indexes, pls use `for(Result result: sdkHeartBeatResults)`

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -51,10 +86,69 @@ private DefaultHeartBeatInfo(Context context) {
     return HeartBeat.NONE;
   }
 
+  @Override
+  public Task<List<HeartBeatResult>> getAndClearStoredHeartBeatInfo() {
+    TaskCompletionSource<List<HeartBeatResult>> taskCompletionSource = new TaskCompletionSource<>();
+    this.backgroundExecutor.execute(
```
vkryachko, line 92: Even though `Tasks.call()` is deprecated I suggest you use it as it makes sure that the task actually completes in the face of exceptions and errors, i.e. see b/157690112 for how tasks can hang indefinitely if an exception happens.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -51,10 +86,69 @@ private DefaultHeartBeatInfo(Context context) {
...
+        });
+    return taskCompletionSource.getTask();
+  }
+
+  @Override
+  public Task storeHeartBeatInfo(@NonNull String heartBeatTag) {
+    TaskCompletionSource<Boolean> taskCompletionSource = new TaskCompletionSource<>();
+    if (consumers.size() <= 0) {
+      taskCompletionSource.setResult(true);
+      return taskCompletionSource.getTask();
```
vkryachko, line 128: use `Tasks.forResult(true)` ?

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfo.java
```diff
@@ -44,4 +49,10 @@ public int getCode() {
 
   @NonNull
   HeartBeat getHeartBeatCode(@NonNull String heartBeatTag);
+
+  @NonNull
+  Task storeHeartBeatInfo(@NonNull String heartBeatTag);
```
vkryachko, line 54: why is it a raw `Task`? should it be `Task<Boolean>`? also consider renaming the method to something like `shouldStoreHeardBeatInfo()`

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
```diff
@@ -50,15 +74,63 @@ static synchronized HeartBeatInfoStorage getInstance(Context applicationContext)
...
+    if (shouldClear) clearStoredHeartBeats();
+    return sdkHeartBeatResults;
+  }
+
+  void clearStoredHeartBeats() {
+    heartBeatSharedPreferences.edit().clear().apply();
+  }
+
+  boolean isValidHeartBeat(long base, long target) {
+    return !((new Date(base).getDate()) == (new Date(target).getDate()));
```
vkryachko, line 123: I wonder what will happen if the device changes timezone, can we use utc here? Additionally getDate() seems deprecated.

Also is it possible that the date is the same but the month is different?

VinayGuthal: no the date would include a combination of month day and year. Yeah can use utc

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfo.java
VinayGuthal: hmm the method itself would storeHeartBeatInformation. It returns a task so that we can wait on the task to complete if needed. The return type doesn't matter in this case but TaskCompletionSource requires a returnType.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
vkryachko: what I mean is that `date.getDate()` returns the day of the month, so there is no info about which month it is or which year

VinayGuthal: Ah okay I thought it handled all the 3. I will change it.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
VinayGuthal: hmm so the other constructor is test only and can't be accessed via this

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
VinayGuthal: Regarding the first question, it shouldn't matter if the device changes timezone it should still be utc bcz we are getting the date based on the milliseconds which are the number of milliseconds since epoch (System.getMillis)

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfo.java
vkryachko: in this case use `Task<Void>`, but `raw types` break type safety

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
vkryachko: in this case pls change the annotation to `@VisibleForTesting` so that linter does not complain

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
```diff
@@ -50,15 +75,66 @@ static synchronized HeartBeatInfoStorage getInstance(Context applicationContext)
...
+  }
+
+  void clearStoredHeartBeats() {
+    heartBeatSharedPreferences.edit().clear().apply();
+  }
+
+  boolean isValidHeartBeat(long base, long target) {
+    Date baseDate = new Date(base);
+    Date targetDate = new Date(target);
+    SimpleDateFormat formatter = new SimpleDateFormat("dd/MM/yyyy z");
```
vkryachko, line 126: consider extracting this formatter into a constant since it never changes.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
```diff
@@ -50,15 +75,66 @@ static synchronized HeartBeatInfoStorage getInstance(Context applicationContext)
...
+    Collections.sort(sdkHeartBeatResults);
+    if (shouldClear) clearStoredHeartBeats();
+    return sdkHeartBeatResults;
+  }
+
+  void clearStoredHeartBeats() {
+    heartBeatSharedPreferences.edit().clear().apply();
+  }
+
+  boolean isValidHeartBeat(long base, long target) {
```
vkryachko, line 123: looking at the signature, iiuc this method does not have anything to do with heartbeat, but rather checks if 2 timestamps belong to the same day in utc, consider renaming the method to something like

```suggestion
  boolean static isSameDateUtc(long oneMillis, long otherMillis) {
```

not `static`, since you don't use any member state of this class.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
```diff
@@ -30,17 +36,38 @@
...
+  HeartBeatInfoStorage(
+      SharedPreferences preferences, SharedPreferences heartBeatSharedPreferences) {
     this.sharedPreferences = preferences;
+    this.heartBeatSharedPreferences = heartBeatSharedPreferences;
+  }
+
+  @VisibleForTesting
+  @RestrictTo(RestrictTo.Scope.TESTS)
+  int getHeartBeatCount() {
+    return this.heartBeatCount;
```
vkryachko, line 70: looks like you're updating it inside synchronized, but reading it unprotected, this is a race. consider making this method synchronized as well

vkryachko: hm, actually: are you using this variable to control when to clean up stale heartbeats? if so, is it ok that this variable will not survive app restarts? wouldn't it cause storage to grow in an unbounded manner?

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfoTest.java
```diff
@@ -15,20 +15,50 @@
...
+        {
+          add(new HeartBeatConsumer() {});
+        }
+      };
+  private DefaultHeartBeatInfo heartBeatInfo;
+
+  @Before
+  public void setUp() {
+    executor = new ThreadPoolExecutor(0, 1, 30L, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
+    storeOnCompleteListener = new TestOnCompleteListener<>();
```
vkryachko, line 58: what is the purpose of these listeners? why not just use `Tasks.await(info.storeHeartBeatInfo(testSdk))` for example?

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
VinayGuthal: Hmm the method is just restricted for Testing so hadn't made it synchronized.  Hmmm but good catch, maybe it makes sense to put this as well under shared preferences?

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfoTest.java
VinayGuthal: hmm I used Tasks.await before but it would cause the looper wait issue as there could be more than one threads. Hence the need of the listener similar to what has been done in firebase-installations.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -51,10 +80,64 @@ private DefaultHeartBeatInfo(Context context) {
...
+          }
+          if (lastGlobalHeartBeat > 0) {
+            storage.updateGlobalHeartBeat(lastGlobalHeartBeat);
+          }
+          return heartBeatResults;
+        });
+  }
+
+  @Override
+  public Task storeHeartBeatInfo(@NonNull String heartBeatTag) {
```
vkryachko, line 118:
```suggestion
  public Task<Void> storeHeartBeatInfo(@NonNull String heartBeatTag) {
```

vkryachko (approved):

VinayGuthal:
The public api surface has changed for the subproject firebase-common:
error: Added method com.google.firebase.FirebaseApp.getNameHello() [AddedMethod]

Please update the api.txt files for the subprojects being affected by this change by running ./gradlew ${subproject}:generateApiTxtFile. Also perform a major/minor bump accordingly.

Issue #1766, opened 2020-07-08
IllegalStateException FirebaseCrashlytics.getInstance


ShivamPokhriyal:
We recently migrated from Fabric to FirebaseCrashlytics and started seeing IllegalStateExceptions. 

Copying over the stacktrace from play console.  
```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:5929)
  at android.app.ActivityThread.access$1200 (ActivityThread.java:209)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1681)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:193)
  at android.app.ActivityThread.main (ActivityThread.java:6724)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:493)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:911)
Caused by: java.lang.IllegalStateException: 
  at com.google.firebase.FirebaseApp.getInstance (FirebaseApp.java:184)
  at com.google.firebase.crashlytics.FirebaseCrashlytics.getInstance (FirebaseCrashlytics.java:112)
  at <package_name>.CustomApplication.onCreate (CommCareApplication.java:195)
  at android.app.Instrumentation.callApplicationOnCreate (Instrumentation.java:1154)
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:5924)
  at android.app.ActivityThread.access$1200 (ActivityThread.java:209)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1681)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:193)
  at android.app.ActivityThread.main (ActivityThread.java:6724)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:493)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:911)
```

This crash appears to be only happening for a particular device. 
```
Blackview BV5500Pro (Android 9)
```

I don't have any more details, neither do I have that device to test it out. So posting it here, if someone else also notices the issue on the same device or some other devices as well.

janbolat:
[...]

PR #1840, opened 2020-07-31
Initialize Firebase in direct boot mode

vkryachko:
Note that even though `FirebaseApp itself initializes, none of the other SDKs get initialized until the user unlocks the devices.
See #1766


vkryachko:
@ashwinraghav you see any concerns with doing this?

PR #1842, opened 2020-07-31
Increase number of retries to 18

mrwillis21:
This will extend the timeout period for high-priority events to 6 days, up from 1.5 days.


mrwillis21:
Discussed with Vlad, closing this in favor of a more precise approach.

PR #1855, opened 2020-08-05
Log-adjusted backoff logic.

vkryachko:
The effect of the change is that initial attempts with small delay
are spread out in time more, which increases the overall time that
events are stored.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/SchedulerConfig.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/SchedulerConfig.java
@@ -65,6 +65,8 @@ public abstract class SchedulerConfig {
 
   private static final long ONE_SECOND = 1000;
 
+  private static final long BACKOFF_LOG_BASE = 10000;
+
   public static SchedulerConfig getDefault(Clock clock) {
     return SchedulerConfig.builder()
         .addConfig(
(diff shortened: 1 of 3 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/SchedulerConfig.java
```diff
@@ -135,10 +137,20 @@ public long getScheduleDelay(Priority priority, long minTimestamp, int attemptNu
     long timeDiff = minTimestamp - getClock().getTime();
     ConfigValue config = getValues().get(priority);
 
-    long delay = Math.max(((long) Math.pow(2, attemptNumber - 1)) * config.getDelta(), timeDiff);
+    long delay = Math.max(adjustedExponentialBackoff(attemptNumber, config.getDelta()), timeDiff);
     return Math.min(delay, config.getMaxAllowedDelay());
   }
 
+  private long adjustedExponentialBackoff(int attemptNumber, long delta) {
```
mrwillis21, line 144: I know this algorithm is being used just because it works out the way we want - but it would help to have some commentary about why we're using this rather complex logic, for anyone down the line wondering why this was chosen.

vkryachko: Oops, I saw your comment after merging. I can create a follow up PR, but fwiw the explanation is present in the commit message.

mrwillis21: Right, but when just looking at the code in the future, it would be nice not to have to do a `git blame` to figure out the intent. :)

PR #1872, opened 2020-08-10
Add != and NOT_IN queries (not public)

thebrianchen:
Porting from [web](https://github.com/firebase/firebase-js-sdk/pull/3557).

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -20,7 +20,6 @@ import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCol
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollectionWithDocs;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
-import static com.google.firebase.firestore.testutil.TestUtil.assertSetEquals;
 import static com.google.firebase.firestore.testutil.TestUtil.expectError;
 import static com.google.firebase.firestore.testutil.TestUtil.map;
 import static java.util.Arrays.asList;
@@ -32,6 +31,7 @@ import static org.junit.Assert.assertTrue;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.firebase.firestore.Query.Direction;
 import com.google.firebase.firestore.testutil.EventAccumulator;
@@ -514,7 +514,7 @@ public class QueryTest {
     expectedDocsMap.remove("i");
 
     QuerySnapshot snapshot = waitFor(collection.whereNotEqualTo("zip", 98101L).get());
-    assertSetEquals(expectedDocsMap.values(), querySnapshotToValues(snapshot));
+    assertEquals(Lists.newArrayList(expectedDocsMap.values()), querySnapshotToValues(snapshot));
 
     // With objects.
     expectedDocsMap = Maps.newHashMap(allDocs);
@@ -522,14 +522,14 @@ public class QueryTest {
     expectedDocsMap.remove("h");
     expectedDocsMap.remove("i");
     snapshot = waitFor(collection.whereNotEqualTo("zip", map("code", 500)).get());
-    assertSetEquals(expectedDocsMap.values(), querySnapshotToValues(snapshot));
+    assertEquals(Lists.newArrayList(expectedDocsMap.values()), querySnapshotToValues(snapshot));
 
     // With Null.
     expectedDocsMap = Maps.newHashMap(allDocs);
     expectedDocsMap.remove("h");
     expectedDocsMap.remove("i");
     snapshot = waitFor(collection.whereNotEqualTo("zip", null).get());
-    assertSetEquals(expectedDocsMap.values(), querySnapshotToValues(snapshot));
+    assertEquals(Lists.newArrayList(expectedDocsMap.values()), querySnapshotToValues(snapshot));
 
     // With NaN.
     expectedDocsMap = Maps.newHashMap(allDocs);
@@ -537,7 +537,7 @@ public class QueryTest {
     expectedDocsMap.remove("i");
     expectedDocsMap.remove("j");
     snapshot = waitFor(collection.whereNotEqualTo("zip", Double.NaN).get());
-    assertSetEquals(expectedDocsMap.values(), querySnapshotToValues(snapshot));
+    assertEquals(Lists.newArrayList(expectedDocsMap.values()), querySnapshotToValues(snapshot));
   }
 
   // TODO(ne-queries): Re-enable once emulator support is added to CI.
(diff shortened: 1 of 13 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -445,15 +520,44 @@ private void validateOrderByFieldMatchesInequality(
     }
   }
 
+  /**
+   * Given an operator, returns the set of operators that cannot be used with it.
+   *
+   * <p>Operators in a query must adhere to the following set of rules: 1. Only one array operator
```
wilhuff, line 526: To make a list use

```html
<ol>
<li>Only one array operator ...
<li>One one disjunctive ...
</ol>
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FieldFilter.java
```diff
@@ -64,20 +64,24 @@ public static FieldFilter create(FieldPath path, Operator operator, Value value)
...
       } else {
         hardAssert(
             operator != Operator.ARRAY_CONTAINS && operator != Operator.ARRAY_CONTAINS_ANY,
             operator.toString() + "queries don't make sense on document keys");
         return new KeyFieldFilter(path, operator, value);
       }
     } else if (Values.isNullValue(value)) {
-      if (operator != Filter.Operator.EQUAL) {
+      // TODO(ne-queries): Update error message to include != operator.
+      if (operator != Filter.Operator.EQUAL && operator != Filter.Operator.NOT_EQUAL) {
```
wilhuff, line 77: `Operator` must already be imported here (see references to `Operator.ARRAY_CONTAINS`) so you can drop the `Filter` in `Filter.Operator`.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/KeyFieldNotInFilter.java
```diff
@@ -0,0 +1,39 @@
+// Copyright 2019 Google LLC
```
wilhuff, line 1: 2020

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Util.java
```diff
@@ -234,4 +240,22 @@ public static int compareByteStrings(ByteString left, ByteString right) {
     }
     return Util.compareIntegers(left.size(), right.size());
   }
+
+  public static List<DocumentKey> extractDocumentKeysFromArrayValue(
```
wilhuff, line 244: This isn't likely to be a widely used or general utility. Better to make this a package private static member of e.g. `KeyFieldInFilter` and just call it from the other one.

(Another clue that this doesn't belong here is that you're referring to `KeyFieldInFilter` in the error message. This is therefore pretty specific to those classes.)

firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -607,6 +611,18 @@ public static ByteString streamToken(String contents) {
...
 
+  /**
+   * Asserts that the contents of actual list is equal to that of the expected one.
+   *
+   * @param expected A list of the expected contents of the set, in any order.
+   * @param actual The list to compare against.
+   * @param <T> The type of the values of in common between the expected list and actual set.
+   */
+  // PORTING NOTE: JUnit and XCTest use reversed conventions on expected and actual values :-(.
+  public static <T> void assertArrayEquals(List<T> expected, List<T> actual) {
```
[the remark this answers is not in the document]
thebrianchen: Done. Thanks for the explanation!

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -483,6 +487,83 @@ public void testQueriesFireFromCacheWhenOffline() {
...
+
+    // Search for zips not matching 98101.
+    Map<String, Map<String, Object>> expectedDocsMap = Maps.newHashMap(allDocs);
+    expectedDocsMap.remove("a");
+    expectedDocsMap.remove("h");
+    expectedDocsMap.remove("i");
+
+    QuerySnapshot snapshot = waitFor(collection.whereNotEqualTo("zip", 98101L).get());
+    assertArrayEquals(
+        Lists.newArrayList(expectedDocsMap.values()), querySnapshotToValues(snapshot));
```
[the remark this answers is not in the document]
thebrianchen: Removed here and throughout.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -483,6 +486,79 @@ public void testQueriesFireFromCacheWhenOffline() {
...
+    Map<String, String> docD = map("key", "bb");
+    Map<String, Map<String, Object>> testDocs =
+        map(
+            "aa", docA,
+            "ab", docB,
+            "ba", docC,
+            "bb", docD);
+    CollectionReference collection = testCollectionWithDocs(testDocs);
+    QuerySnapshot docs = waitFor(collection.whereNotEqualTo(FieldPath.documentId(), "aa").get());
+    assertEquals(asList(docB, docC, docD), querySnapshotToValues(docs));
```
wilhuff, line 559: Apologies for what's going to be an irritating question (since I didn't pick up on this earlier), but why do you `assertSetEquals(x, expected)` in the test above but `assertEquals(asList(x), expected)` here when we're testing the same underlying operator?

To me it seems like either the ordering matters or it doesn't. If ordering of query results is predictable, `assertSetEquals` seems wrong because it allows the result to be unordered when our customers really care that it's ordered. The reverse is also true: if ordering *doesn't* matter, then this test is too strict, and could create a false positive.

To my understanding, query results always have a well defined ordering (breaking ties on key ordering). This seems to point to `assertSetEquals` being wrong above.

What's your take on this? Why the discrepancy?

(The same concern applies to the `whereNotIn`, below).

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
wilhuff: The list wrapper element is required (`<ol>` indicates an ordered list (aka numbered); use `<ul>` for unordered (aka bulleted)).

firebase-firestore/src/main/java/com/google/firebase/firestore/core/NotInFilter.java
```diff
@@ -0,0 +1,36 @@
+// Copyright 2019 Google LLC
```
wilhuff, line 1: 2020 here too

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
thebrianchen: Thanks for catching this actually! I originally ran into an issue where I was converting the snapshots into a HashSet in order to easily remove the unneeded documents, but converting them back into a List resulted in an ordering that didn't match the query snapshots from the query. Instead of trying to create the `expectedDocsMap` in the correct order, I took the "shortcut" of comparing them as sets to avoid the ordering errors without realizing that order matters in query results.

I took a look at the code, and realized that Using `Maps.newHashMap()` keeps the documents in the correct order. Switched back to using `assertEquals` with `Lists.newArrayList()`.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/NotInFilter.java
thebrianchen: done.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
thebrianchen: Done for reals.

PR #1877, opened 2020-08-12
add platform logging to firebase data transport

VinayGuthal:


firebase-datatransport/src/main/java/com/google/firebase/datatransport/FirebaseTransportFactory.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public <T> Transport<T> getTransport(
+      String name,
+      Class<T> payloadType,
+      Encoding payloadEncoding,
+      Transformer<T, byte[]> payloadTransformer) {
+    transportFactory.getTransport(
+        PLATFORM_LOGGING_SOURCE_NAME, FirebasePlatformInfo.class, FirebasePlatformInfo::getPlatformInfo());
```
VinayGuthal, line 55: Need a correct byte transformer here

firebase-datatransport/firebase-datatransport.gradle
```diff
@@ -14,8 +14,21 @@
 
 plugins {
     id 'firebase-library'
+    id 'com.google.protobuf'
```
vkryachko, line 17: we can't afford a proto dependency here due to size and compatibility issues, pls use json instead, the easiest way to do it is to use `@Encodable`.

PR #1883, opened 2020-08-17
Update koltin version in buildSrc

rlazo:


buildSrc/build.gradle
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 plugins {
-    id "org.gradle.kotlin.kotlin-dsl" version "1.2.6"
+    id "org.gradle.kotlin.kotlin-dsl" version "1.3.6"
```
vkryachko, line 16: This will most likely not work, the kotlin dsl version depends on the version of Gradle used, so bumping it here is not something we want as in this case behavior is undefined.

The same might apply to the kotlin plugin below, but I am not sure.

rlazo:
Yeap, this needs further research. Marking the PR as draft for the time being.

PR #1900, opened 2020-08-20
docs(storage-ktx): document kotlin destructuring declarations

thatfiredev:
~**DO NOT MERGE** before the next release of storage-ktx.~

This should document the new features added in #1644


thatfiredev:
KTX docs were removed from firebaseopensource.com in #2363, so I think this PR is no longer needed.

PR #1939, opened 2020-09-04
Update the README instructions to clarify the importance of pulling submodules

var-const:


var-const:
Tested by doing a fresh clone running the `git clone` command provided and verifying that `./gradlew :firebase-firestore:check` succeeds (whereas it fails for the previous iteration of the command). Let me know if there are any other targets I should check.

README.md
```diff
@@ -53,9 +53,12 @@ https://firebase.google.com.
 ## Getting Started
 
 * Install the latest Android Studio (should be 3.0.1 or later)
-* Clone the repo (`git clone git@github.com:firebase/firebase-android-sdk.git`)
+* Clone the repo (`git clone --recurse-submodules --remote-submodules git@github.com:firebase/firebase-android-sdk.git`)
```
rlazo, line 56: We don't use `--remote-submodules` when syncing because it pulls the submodules from "head" in their respective repos, not from the exact commit we want to. 

Each project is responsible to update their remotes if deemed necessary.

README.md
```diff
@@ -53,9 +53,12 @@ https://firebase.google.com.
...
+    * When cloning the repo, it is important to get the submodules as well. If
+    you have already cloned the repo without the submodules, you can update the
+    submodules by running `git submodule update --init --recursive`.
 * Import the firebase-android-sdk gradle project into Android Studio using the
-  **Import project(Gradle, Eclipse ADT, etc.** option.
+  **Import project(Gradle, Eclipse ADT, etc.)** option.
```
rlazo, line 61: Nice catch, thanks!

README.md
var-const: Done, thanks!

rlazo (approved):

PR #1947, opened 2020-09-08
Fix NotInFilter matching

thebrianchen:
`whereNotIn(path, Array(null))` always returns an empty list, which means that `NotInFilter.matches()` needs to also not match any documents if the filter's value contains `null`. 

Also added a null serialization test for notIn.


wilhuff (approved):
LGTM

PR #1333, opened 2020-03-09
 Adding a TimeSource in Utils to easily inject and test the auth token expiration.

ankitaj224:
Refer to b/149959987

PR #1954, opened 2020-09-09
Adding ResponseHandler to implement backoff retry strageies based on response code.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,76 @@
...
+    nextAllowedRequestTime = utils.currentTimeInMillis() + backOffTime;
+    if (backOffTime == BACKOFF_TIME_24H_IN_MILLIS) {
+      retryCount = 0;
+    }
+  }
+
+  private synchronized long getBackoffTime(int responseCode) {
+    // Quickly increasing dynamically configured back-off strategy for Retryable errors. Read more:
+    // https://cloud.google.com/storage/docs/exponential-backoff.
+    if (!isNonRetryableError(responseCode)) {
```
maksymmalyhin, line 53: nit: double negation looks a bit hard to read to me.

maksymmalyhin, line 53: What will happen in the case of the request success? Will we still backoff? It seems that the backoff reset is not implemented.

maksymmalyhin (commented):
Mostly LGTM, on the logic part except the question on the success handling.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
ankitaj224: Thanks for catching this case. Addressed it :)

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  public boolean isRequestAllowed() {
+    long currentTime = utils.currentTimeInMillis();
+    boolean isRequestAllowed =
+        attemptCount == 0 || currentTime > nextAllowedRequestTime;
+
+    if (currentTime > nextAllowedRequestTime) {
+      synchronized (this) {
+        nextAllowedRequestTime = Long.MAX_VALUE;
```
maksymmalyhin, line 81: I wonder what's the intended used of  `isRequestAllowed` method? Will this line lead to the method returning `false` until `setNextAllowedRequestTime()` method is called?

firebase-installations/src/test/java/com/google/firebase/installations/remote/ResponseHandlerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    responseHandler.setNextAllowedRequestTime(RETRYABLE_RESPONSE_CODE);
+    assertFalse(responseHandler.isRequestAllowed());
+    responseHandler.setNextAllowedRequestTime(RETRYABLE_RESPONSE_CODE);
+    assertTrue(responseHandler.isRequestAllowed());
+  }
+
+  // Adds specified minutes to the CURRENT_TIME_IN_MILLIS.
+  private long addMinutesToCurrentTime(int minutes) {
+    return CURRENT_TIME_IN_MILLIS + TimeUnit.MINUTES.toMillis(minutes);
+  }
```
maksymmalyhin, line 102: I think tests for the reset case will be beneficial. Tests that cover ResponseHandler.java lines 80-83 as well.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @GuardedBy("this")
+  private int attemptCount = 0;
+
+  ResponseHandler(Utils utils) {
+    this.utils = utils;
+  }
+
+  public synchronized void setNextAllowedRequestTime(int responseCode) {
+    if (responseCode == 200) {
```
maksymmalyhin, line 43: If I'm not mistaken, there may be other success codes we receive, e.g. 201 when FIS created. Would it be better to be more generic and treat all codes between 200 and 300 as success?

maksymmalyhin (approved):
LGTM, a couple minor comments/questions.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -15,13 +15,18 @@
...
 import com.google.firebase.installations.local.PersistedInstallationEntry;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
-/** Util methods used for {@link FirebaseInstallations} */
-class Utils {
+/**
+ * Util methods used for {@link FirebaseInstallations}.
+ *
+ * @hide
```
andirayo, line 27: out of curiosity:
what does this mean/do?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -43,7 +48,12 @@ public boolean isAuthTokenExpired(PersistedInstallationEntry entry) {
 
   /** Returns current time in seconds. */
   public long currentTimeInSecs() {
-    return TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis());
+    return TimeUnit.MILLISECONDS.toSeconds(currentTimeInMillis());
+  }
+
+  /** Returns current time in milliseconds. */
+  public long currentTimeInMillis() {
+    return System.currentTimeMillis();
```
andirayo, line 56: I probably asked this before:
Any chance we can inject a timesource here or any instance that can be overwritten in different environments?
Not important at all.
But if we plan to do something about this, then maybe add a comment or TODO.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -53,4 +63,8 @@ static boolean isValidAppIdFormat(@Nullable String appId) {
   static boolean isValidApiKeyFormat(@Nullable String apiKey) {
     return API_KEY_FORMAT.matcher(apiKey).matches();
   }
+
+  public long getRandomMillis() {
```
andirayo, line 67: What's the point of this method?
Maybe a JavaDoc?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -15,13 +15,18 @@
...
 import java.util.regex.Pattern;
 
-/** Util methods used for {@link FirebaseInstallations} */
-class Utils {
+/**
+ * Util methods used for {@link FirebaseInstallations}.
+ *
+ * @hide
+ */
+public class Utils {
```
andirayo, line 29: Generally not a fan of classes that are called "Utils".
Seems like a dump for all methods that have no home, haha

Maybe we can have a time class that deals with all time related methods and has all time related constants, and call it "TimeSource" or "SystemTime" or "Clock" or something?
And another class that deals with validation and input checking, called "Validator" or something.

Even better would be if all these methods could find a home in the respective model classes (assuming that we have model classes).
Example: `#isAuthTokenExpired` goes into the `AuthToken` model class.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.firebase.installations.Utils;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The {@link ResponseHandler} class calculates the next allowed request time. Also, decides whether
+ * a network request to FIS servers is allowed to execute.
+ *
+ * @hide
+ */
+class ResponseHandler {
```
andirayo, line 27: It seems to me that this class is not handling responses, but rather response stati.
How about one of the following names:
* RequestLimiter  (my favorite so far)
* BackoffStrategy
* ResponseStatusHandler

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -30,7 +35,7 @@
    * Checks if the FIS Auth token is expired or going to expire in next 1 hour {@link
    * #AUTH_TOKEN_EXPIRATION_BUFFER_IN_SECS}.
    */
-  public boolean isAuthTokenExpired(PersistedInstallationEntry entry) {
+  public boolean isAuthTokenExpired(@NonNull PersistedInstallationEntry entry) {
```
andirayo, line 38: static?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -43,7 +48,12 @@ public boolean isAuthTokenExpired(PersistedInstallationEntry entry) {
 
   /** Returns current time in seconds. */
   public long currentTimeInSecs() {
```
andirayo, line 50: static?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -43,7 +48,12 @@ public boolean isAuthTokenExpired(PersistedInstallationEntry entry) {
...
-    return TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis());
+    return TimeUnit.MILLISECONDS.toSeconds(currentTimeInMillis());
+  }
+
+  /** Returns current time in milliseconds. */
+  public long currentTimeInMillis() {
```
andirayo, line 55: static?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
andirayo, line 67: static?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
andirayo, line 29: final

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+  private static final long BACKOFF_TIME_30_MINS_IN_MILLIS = TimeUnit.MINUTES.toMillis(30);
+  private final Utils utils;
+
+  @GuardedBy("this")
+  private long nextAllowedRequestTime = Long.MAX_VALUE;
+
+  @GuardedBy("this")
+  private int attemptCount = 0;
+
+  ResponseHandler(Utils utils) {
```
andirayo, line 38: Why do we need a Utils instance?
Aren't all methods in Utils class methods (a.k.a. "static")?

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
andirayo: +1

andirayo: I would add a method similar `#isNonRetryableError`, e.g. `#isSuccessful`.
Usually `HttpResponse` instances already have such a method.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @hide
+ */
+class ResponseHandler {
+  private static final long BACKOFF_TIME_24H_IN_MILLIS = TimeUnit.HOURS.toMillis(24);
+  private static final long BACKOFF_TIME_30_MINS_IN_MILLIS = TimeUnit.MINUTES.toMillis(30);
+  private final Utils utils;
+
+  @GuardedBy("this")
+  private long nextAllowedRequestTime = Long.MAX_VALUE;
```
andirayo, line 33: How about "`nextRequestTime`".
The word "allowed" seems to be noise.
Not feeling strongly about it.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @GuardedBy("this")
+  private int attemptCount = 0;
+
+  ResponseHandler(Utils utils) {
+    this.utils = utils;
+  }
+
+  public synchronized void setNextAllowedRequestTime(int responseCode) {
```
andirayo, line 42: `#setNextRequestTime`?

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    // Quickly increasing dynamically configured back-off strategy for Retryable errors. Read more:
+    // https://cloud.google.com/storage/docs/exponential-backoff.
+    return (long)
+        Math.min(
+            Math.pow(2, attemptCount) + utils.getRandomMillis(), BACKOFF_TIME_30_MINS_IN_MILLIS);
+  }
+
+  // Response codes classified as non-retryable for FIS API, all other response codes will be
+  // treated as retryable-errors. Read more: go/fis-api-error-code-classification.
+  private static boolean isNonRetryableError(int responseCode) {
```
andirayo, line 71: Usually it's better to not use negations in method (or variable) names.
Options:
* `isConfigurationProblem` (my favorite so far)
* `isRetryableError` and use negations above

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
andirayo: How about instead we set `attemptCount = 0`.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+      resetAllowedRequestTime();
+      return;
+    }
+    attemptCount++;
+    long backOffTime = getBackoffTime(responseCode);
+    nextAllowedRequestTime = utils.currentTimeInMillis() + backOffTime;
+  }
+
+  private synchronized void resetAllowedRequestTime() {
+    nextAllowedRequestTime = Long.MAX_VALUE;
```
andirayo, line 53: Why do we set this at all?
Can we just ignore this value or set it to `-1`?

It seems like you are abusing some existing constant (i.e. `Long.MAX_VALUE`) for semantic purposes.
If that is the case, I recommend to rather introduce your own constant that has a speaking name.
Example: `NEXT_REQUEST_TIME_IMMEDIATELY` or something similar.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    attemptCount = 0;
+  }
+
+  private synchronized long getBackoffTime(int responseCode) {
```
andirayo, line 57: #getBackoffDuration

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    // https://cloud.google.com/storage/docs/exponential-backoff.
+    return (long)
+        Math.min(
+            Math.pow(2, attemptCount) + utils.getRandomMillis(), BACKOFF_TIME_30_MINS_IN_MILLIS);
+  }
+
+  // Response codes classified as non-retryable for FIS API, all other response codes will be
+  // treated as retryable-errors. Read more: go/fis-api-error-code-classification.
+  private static boolean isNonRetryableError(int responseCode) {
+    return responseCode == 400 || responseCode == 403;
```
andirayo, line 72: constants?
don't we have those defined somewhere already?

firebase-installations/src/test/java/com/google/firebase/installations/remote/ResponseHandlerTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.robolectric.RobolectricTestRunner;
+
+/** Tests for {@link ResponseHandler}. */
+@RunWith(RobolectricTestRunner.class)
+public class ResponseHandlerTest {
+  private static final int NON_RETRYABLE_RESPONSE_CODE = 403;
+  private static final int RETRYABLE_RESPONSE_CODE = 500;
```
andirayo, line 34: `SOME_RETRYABLE_RESPONSE_CODE`
or
`RETRYABLE_RESPONSE_CODE_EXAMPLE`

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public boolean isRequestAllowed() {
+    long currentTime = utils.currentTimeInMillis();
+    boolean isRequestAllowed = attemptCount == 0 || currentTime > nextAllowedRequestTime;
```
andirayo, line 77: I don't think this is a problem, but I want to call it out:
If the end-user changes the system-time to the past, the app will stop working.
This problem will be fixed by restarting the app and usually end-users changing system-clock know this - at least I knew whenever I have done similar things in the past, haha
Maybe add a short comment here to call this out?

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: Mockito doesn't allow to mock static methods. As a result methods in this class are not static.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: https://github.com/firebase/firebase-android-sdk/pull/1954/files#r487182669

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: https://github.com/firebase/firebase-android-sdk/pull/1954/files#r487182669

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: https://github.com/firebase/firebase-android-sdk/pull/1954/files#r487182669

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: Correct, I have a open bug & PR addressing this request. https://github.com/firebase/firebase-android-sdk/pull/1333
I`ll address that PR shortly after submitting this.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
ankitaj224: @maksymmalyhin 
isRequestAllowed will be called before making a network call to FIS Servers. This method will return true if its a first call i.e attemptCount = 0 or if currentTime is greater than the set nextRequestTime.

The way I imagined plugging these methods was `setNextAllowedRequestTime()` will be called on every responseCode resulting in correctly setting `attemptNumber` & `nextRequestTime`.

But I guess, we can do without explicitly resetting `nextRequestTime`. WDYT?

ankitaj224: @andirayo I don't think setting it attemptCount = 0 will be correct.

Let's say FIS servers respond 429/ 5xx for consecutive requests.

`Response1 code 500 : At Duration T1 with attemptCount = 1, nextRequestTime = T2`

Next Request to communicate with FIS servers is at T3 > T2. Since this is true, we allow the request to proceed. But resetting attemptCount = 0 will beat the purpose of exponential backoff. 

If this request also fails, attemptCount should be 2:
`Response2 code 500: At  Duration T3 with attemptCount = 2, nextRequestTime = T4`

Please correct me if I have misunderstood something here. 

Thanks.

maksymmalyhin: IMO this logic places unnecessary restrictions on the client and makes some potential use cases complicated e.g. if the client invokes `isRequestAllowed` and then doesn't send a request because of another condition, there is no a straightforward way to call `isRequestAllowed` and get expected result. I would prefer keep returning `true` until a signal from the client (via `setNextAllowedRequestTime()`) is received. So, basically, I would just remove lines 80-83. WDYT?

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
ankitaj224: Agreed on ignoring the value instead of resetting it. I initially had a thought without resetting it will not allow some requests to FIS servers. But on rethinking the logic, it will work correctly without resetting the nextRequestTime.
Thanks for pointing it out. PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
ankitaj224: Added a comment. Thanks.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
ankitaj224: For testing purpose. Easier to mock & test the behavior.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: I dont have a model class per se but I can clean up this class in another PR. Probably the one where I inject the Timesource .

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
andirayo: Please add a respective comment (not Javadoc) to the code.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
andirayo: a) Please add a respective comment to the code.
b) Just to be sure we are on the same page:
If this class were to use a proper `TimeSource`, this issue would not exist and tests could mock the TimeSource instance instead.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
andirayo: Please add a respective comment (not JavaDoc) to the method.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
andirayo: Just seeing this. Some comments above may allude to this ...

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+// limitations under the License.
+
+package com.google.firebase.installations.remote;
+
+import androidx.annotation.GuardedBy;
+import com.google.firebase.installations.Utils;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The {@link RequestLimiter} class calculates the next allowed request time. Also, decides whether
```
andirayo, line 22: Let's remove the word "allowed".
The next sentence is explaining that it also checks if a request is allowed.

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+ * a network request to FIS servers is allowed to execute.
+ *
+ * @hide
+ */
+class RequestLimiter {
+  private static final long BACKOFF_TIME_24H_IN_MILLIS = TimeUnit.HOURS.toMillis(24);
+  private static final long BACKOFF_TIME_30_MINS_IN_MILLIS = TimeUnit.MINUTES.toMillis(30);
```
andirayo, line 29: These constant names currently only describe the actual value of the constant, but as reader I don't understand their business purpose.
It would be better, if the constant name describes the semantic meaning or what the constant is used for.

How about these names:
MAXIMUM_BACKOFF_INTERVAL_FOR_CONFIGURATION_ERRORS
MAXIMUM_BACKOFF_INTERVAL_FOR_ SERVER_ERRORS

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+  private long nextRequestTime;
+
+  @GuardedBy("this")
+  private int attemptCount;
+
+  RequestLimiter(Utils utils) {
+    this.utils = utils;
+  }
+
+  public synchronized void setNextRequestTime(int responseCode) {
```
andirayo, line 42: JavaDoc for public method?

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+  private final Utils utils;
+
+  @GuardedBy("this")
+  private long nextRequestTime;
+
+  @GuardedBy("this")
+  private int attemptCount;
+
+  RequestLimiter(Utils utils) {
+    this.utils = utils;
```
andirayo, line 39: Maybe add a comment that this is injected to facilitate testability of TimeSource?

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
andirayo: Maybe replace "...INTERVAL..." with  "...DURATION..."
I have just seen that you use the term "duration" below, instead of interval.
For some reason, I prefer "interval" today, haha, but
I think both are fine as long as we stay consistent.

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+    if (isSuccessful(responseCode)) {
+      resetAttemptCount();
+      return;
+    }
+    attemptCount++;
+    long backOffTime = getBackoffDuration(responseCode);
+    nextRequestTime = utils.currentTimeInMillis() + backOffTime;
+  }
+
+  private synchronized void resetAttemptCount() {
```
andirayo, line 52: This method seems to describe its implementation, rather than what it does.

I'm not sure if this what is happening here, but what about this name:
`#resetBackoffStrategy`
or
`#resetBackoffInterval`

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+    attemptCount = 0;
+  }
+
+  private synchronized long getBackoffDuration(int responseCode) {
+    // Fixed 24 hours silence period for non-retryable errors. Read more: b/160751425.
```
andirayo, line 57: "[...] non-retryable server errors. [...]"

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+    // Quickly increasing dynamically configured back-off strategy for Retryable errors. Read more:
+    // https://cloud.google.com/storage/docs/exponential-backoff.
+    return (long)
+        Math.min(
+            Math.pow(2, attemptCount) + utils.getRandomMillis(), BACKOFF_TIME_30_MINS_IN_MILLIS);
+  }
+
+  // Response codes classified as retryable for FIS API. Read more on FIS response codes:
+  // go/fis-api-error-code-classification.
```
andirayo, line 69: Please add that these are server errors and response code 429 (i.e. TOO_MANY_REQUESTS)

andirayo (approved):
Looks good!

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: a) Done. 
b) Correct.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: When applied to a package, class, method or field, @hide removes that node and all of its children from the public documentation.

firebase-installations/src/main/java/com/google/firebase/installations/remote/ResponseHandler.java
ankitaj224: Agreed. I realized that there is no need to reset the nextRequestTime here. Changed it.

Thanks for the reveiw.

PR #1972, opened 2020-09-10
Make the language reported via headers configurable

var-const:
See https://github.com/firebase/firebase-ios-sdk/pull/6264 for the equivalent change in the iOS repo.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -39,6 +39,7 @@ import com.google.firebase.firestore.core.FirestoreClient;
 import com.google.firebase.firestore.local.SQLitePersistence;
 import com.google.firebase.firestore.model.DatabaseId;
 import com.google.firebase.firestore.model.ResourcePath;
+import com.google.firebase.firestore.remote.FirestoreChannel;
 import com.google.firebase.firestore.remote.GrpcMetadataProvider;
 import com.google.firebase.firestore.util.AsyncQueue;
 import com.google.firebase.firestore.util.Executors;
@@ -634,5 +635,9 @@ public class FirebaseFirestore {
       throw new IllegalArgumentException(
           "Provided document reference is from a different Cloud Firestore instance.");
     }
+  }
+
+  void setClientLanguage(String language) {
+    FirestoreChannel.setClientLanguage(language);
   }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -19,6 +19,7 @@ import static com.google.firebase.firestore.util.Preconditions.checkNotNull;
 
 import android.app.Activity;
 import android.content.Context;
+import androidx.annotation.Keep;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -281,10 +278,24 @@ public void invalidateToken() {
...
+  }
+
+  private static String getDefaultClientLanguage() {
+    // Note: there is no good way to get the Java language version on Android
+    // (System.getProperty("java.version") returns "0", for example).
+    return "gl-java/";
+  }
+
+  private String getGoogApiClientValue() {
+    return String.format("%s fire/%s grpc/", clientLanguage, BuildConfig.VERSION_NAME);
```
var-const, line 292: Unfortunately, the method to retrieve the gRPC version is not public in the Java SDK.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -635,4 +636,15 @@ void validateReference(DocumentReference docRef) {
...
   }
+
+  /**
+   * Sets the language of the public API in the format of "gl-<language>/<version>" where version
+   * might be blank, e.g. `gl-cpp/`. The provided string is used as is.
+   *
+   * Note: this method is package-private because it is expected to only be called via JNI (which
+   * ignores access modifiers).
+   */
+  static void setClientLanguage(String languageToken) {
```
wilhuff, line 647: Proguard will rewrite this method name in release builds of users' applications. We need to annotate this, likely with [`@KeepForSdk`](https://github.com/firebase/firebase-android-sdk#keepforsdk) or possibly `@Keep`, though I don't know which applies to package-private methods.

Note that this method does not have to be `public` to be accessible from JNI (JNI ignores access control modifiers), but this can only work when the name isn't shortened.

The argument here should have a nullability annotation, probably `@NonNull`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -39,19 +39,16 @@
  * Wrapper class around io.grpc.Channel that adds headers, exception handling and simplifies
  * invoking RPCs.
  */
-class FirestoreChannel {
+public class FirestoreChannel {
```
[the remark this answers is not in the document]
var-const: Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -39,19 +39,16 @@
...
   private static final Metadata.Key<String> RESOURCE_PREFIX_HEADER =
       Metadata.Key.of("google-cloud-resource-prefix", Metadata.ASCII_STRING_MARSHALLER);
 
-  // TODO: The gRPC version is determined using a package manifest, which is not available
-  // to us at build time or runtime (it's empty when building in google3). So for now we omit the
-  // version of grpc.
-  private static final String X_GOOG_API_CLIENT_VALUE =
-      "gl-java/ fire/" + BuildConfig.VERSION_NAME + " grpc/";
+  /** The client language reported via the X_GOOG_API_CLIENT_HEADER. */
+  private static String clientLanguage = getDefaultClientLanguage();
```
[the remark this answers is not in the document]
var-const: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
var-const: Hmm, looking at the way `@KeepForSdk` is described, it appears to have a more limited effect than `@Keep`, so it seems using `@Keep` is safer (apparently `@KeepForSdk` still allows for the method to be stripped in end-developer apps). Feel free to correct me, though.

wilhuff (approved):
LGTM

PR #1980, opened 2020-09-14
Remove session ID tracking from session reporting coordinator.

mrwillis21:
Multi-process apps can have multiple instances, which can result
in one instance having stale state. Relying on the passed-in
session ID as the source of truth ensures the SRC is always
operating on the most recent open session.


kmandrika (approved):

mrichards (approved):
👍

PR #1981, opened 2020-09-14
Fix ConcurrentModificationException in FirestoreMultiDbComponent.onDeleted()

dconeybe:


firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreMultiDbComponent.java
```diff
@@ -79,10 +80,15 @@ public synchronized void remove(@NonNull String databaseId) {
...
-    // Shuts down all database instances and remove them from registry map when App is deleted.
-    for (Map.Entry<String, FirebaseFirestore> entry : instances.entrySet()) {
-      entry.getValue().terminateInternal();
-      instances.remove(entry.getKey());
+    // Shut down all database instances and remove them from the registry map. To avoid
+    // ConcurrentModificationException, make a copy of the entries instead of using an iterator from
+    // the `instances` map directly.
+    for (Map.Entry<String, FirebaseFirestore> entry : new ArrayList<>(instances.entrySet())) {
+      entry.getValue().terminate();
+      if (instances.containsKey(entry.getKey())) {
```
dconeybe, line 88: This is just a sanity check to make sure that the `instances` map is in the expected state after the `for` loop completes. If `terminate()` is ever changed to not remove the instance then that would go unnoticed, breaking the contract of this method.

wilhuff: `hardAssert` is what you want for this.

dconeybe: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/FirestoreMultiDbComponentTest.java
```diff
@@ -0,0 +1,72 @@
...
+    FirestoreMultiDbComponent component = createComponent(firebaseApp);
+    FirebaseFirestore firebaseFirestore1 = component.get("db1");
+    FirebaseFirestore firebaseFirestore2 = component.get("db2");
+
+    component.onDeleted(firebaseApp.getName(), firebaseApp.getOptions());
+
+    assertNotSame(firebaseFirestore1, component.get("db1"));
+    assertNotSame(firebaseFirestore2, component.get("db2"));
+    // Clean up the instances registered by the calls above.
+    component.onDeleted(firebaseApp.getName(), firebaseApp.getOptions());
```
wilhuff, line 51: Is this cleanup actually required? There's no shared state, so it seems like it would be fine to just drop the component and delete this line and the comment.

If this clean-up *is* required, then this should probably be within a finally block, because `assertNotSame` will throw on a failure and then the cleanup won't happen.

dconeybe: No, it's probably not required. I've removed it.

wilhuff (approved):
LGTM

What a blast from the past!

PR #1988, opened 2020-09-16
Clarifying why FIS SDK reads SHA1 hash.

ankitaj224:


firebase-installations/src/main/java/com/google/firebase/installations/local/IidStore.java
```diff
@@ -167,6 +167,13 @@ private String readPublicKeyFromLocalStorageAndCalculateInstanceId() {
     }
   }
 
+  /**
+   * FirebaseInstallations SDK uses the SHA1 hash for backwards compatibility with the legacy
+   * InstanceID SDK. The SHA1 hash is used to access Instance IDs stored on the device and not for
+   * any security relevant process. This is a one-time step that allows migration of old client
+   * identifiers. Cryptographic security is not needed here so potential hash collisions are not a
+   * problem.
+   */
```
andirayo, line 176: Please don't put this in JavaDoc, but rather as a code comment above the actual usage of SHA1.

andirayo (approved):

PR #2002, opened 2020-09-23
Minor fixes addressing previous comments during cl review

ankitaj224:
Minor fix.


firebase-installations/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,8 @@
     package="com.google.firebase.installations">
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <!--Although the *SdkVersion is captured in gradle build files, this is required for non gradle builds-->
```
andirayo, line 6: If possible and known, can you please add a reason why Android SDK v14 is required, e.g. because we use some HTTP client that is not available in v13...

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -66,10 +66,12 @@ static boolean isValidApiKeyFormat(@Nullable String apiKey) {
...
-  request.*/
+  /**
+   * Returns a random number of milliseconds less than or equal to 1000. This helps to avoid cases
+   * where many clients get synchronized by some situation and all retry at once, sending requests
+   * in synchronized waves. The value of random_number_milliseconds is recalculated after each retry
+   * request.
+   */
   public long getRandomMillis() {
     // Mockito doesn't allow to mock static methods. As a result this util method is not static.
     return (long) (Math.random() * 1000);
```
andirayo, line 77: Just for completeness:
A random number generator should, similar to a timesource, be injected into code to allow deterministic testing of code.

No action item! I just wanted to shared this with you.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -66,10 +66,12 @@ static boolean isValidApiKeyFormat(@Nullable String apiKey) {
...
-   where many clients get synchronized by some situation and all retry at once, sending requests
-   in synchronized waves. The value of random_number_milliseconds is recalculated after each retry
-  request.*/
+  /**
+   * Returns a random number of milliseconds less than or equal to 1000. This helps to avoid cases
+   * where many clients get synchronized by some situation and all retry at once, sending requests
+   * in synchronized waves. The value of random_number_milliseconds is recalculated after each retry
+   * request.
+   */
   public long getRandomMillis() {
```
andirayo, line 75: How about renaming this to one of
* `#getRandomTimeInterval`
* `#getRandomDelay`
* `#getRandomDelayForSyncPrevention`

The goal is to have a name that semantically describes what a method is used for, rather than a name that describes its implementation.

andirayo (approved):

firebase-installations/src/main/AndroidManifest.xml
ankitaj224: I have no clue. I will try & findout and update the file. But all Firebase sdks minSDK support is API 14.

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
ankitaj224: I like it. Thanks for the suggestion. I added a TODO and will follow up the change in a separate PR.

PR #2003, opened 2020-09-23
Enable request limiter for CreateInstallation, GenerateAuthToken request calls.

ankitaj224:
Preventing erroneous requests.


firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -143,7 +147,7 @@ public InstallationResponse createFirebaseInstallation(
     String resourceName = String.format(CREATE_REQUEST_RESOURCE_NAME_FORMAT, projectID);
     int retryCount = 0;
     URL url = getFullyQualifiedRequestUri(resourceName);
-    while (retryCount <= MAX_RETRIES) {
+    while (requestLimiter.isRequestAllowed() || shouldServerErrorRetry) {
```
andirayo, line 150: Question:
When we should not send requests to the server, we throw a `FirebaseInstallationsException` with message "Firebase Installations Service is unavailable."? Is that correct and desired behavior?
If yes, why not throw this exception before the while loop?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -158,15 +162,17 @@ public InstallationResponse createFirebaseInstallation(
         writeFIDCreateRequestBodyToOutputStream(httpURLConnection, fid, appId);
 
         int httpResponseCode = httpURLConnection.getResponseCode();
+        requestLimiter.setNextRequestTime(httpResponseCode);
```
andirayo, line 165: What happens in case of exceptions?
Do we need to set a `nextRequestTime`?
I assume not, just checking if you considered that case.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -158,15 +162,17 @@ public InstallationResponse createFirebaseInstallation(
...
         if (httpResponseCode == 200) {
           return readCreateResponse(httpURLConnection);
         }
 
         logFisCommunicationError(httpURLConnection, appId, apiKey, projectID);
 
-        if (httpResponseCode == 429 || (httpResponseCode >= 500 && httpResponseCode < 600)) {
+        if (httpResponseCode >= 500 && httpResponseCode < 600) {
           retryCount++;
+          shouldServerErrorRetry = retryCount <= MAX_RETRIES;
```
andirayo, line 175: I'm not sure if I like this code!
There is several things wrong with this I think:
a) When I read the While loop condition above, it is unclear to me what `shouldServerErrorRetry` is and I have to search the code to find where the variable is set.
b) If avoidable (which is almost always the case) local variables should **not** change in value, because it makes code hard to read - this is actually a great example why. If you have to do it at all costs, you should mark the local variable as `@Var`.
c) I just realize it's not a local variable, but an instance variable. Why is that? That is even more confusing to me. Are you saying that other parallel processes can change the value of the variable and affect this while loop? (c.1) I doubt that this is the case. (c.2) And if it's the case, I doubt that it's a good idea!
d) I also liked using a variable named `retryCount` in the while loop condition better than a variable called `shouldServerErrorRetry`... With the `retryCount` it is immediately obvious, that we will retry sending the same request for some times... `shouldServerErrorRetry` does not really give that much information.

I recommend:
Leave the `retryCount <= MAX_RETRIES` in the while loop.
Even better, get rid of the while loop and replace it with a `for (int retryCount = 0; retryCount <= MAX_RETRIES; retryCount++)`.
That explains so much better what is happening here than a local variable that is changed somewhere else.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -158,15 +162,17 @@ public InstallationResponse createFirebaseInstallation(
...
+        requestLimiter.setNextRequestTime(httpResponseCode);
 
         if (httpResponseCode == 200) {
           return readCreateResponse(httpURLConnection);
         }
 
         logFisCommunicationError(httpURLConnection, appId, apiKey, projectID);
 
-        if (httpResponseCode == 429 || (httpResponseCode >= 500 && httpResponseCode < 600)) {
+        if (httpResponseCode >= 500 && httpResponseCode < 600) {
```
andirayo, line 173: What happened to `429`?
We `logBadConfigError` for that status now? 
If that is the case, I think it's not correct.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -158,15 +162,17 @@ public InstallationResponse createFirebaseInstallation(
         writeFIDCreateRequestBodyToOutputStream(httpURLConnection, fid, appId);
 
         int httpResponseCode = httpURLConnection.getResponseCode();
+        requestLimiter.setNextRequestTime(httpResponseCode);
 
         if (httpResponseCode == 200) {
```
andirayo, line 167: This obviously seemed to have worked in the past, but maybe it's better we check for a successful code here?
I'm not sure, but I think the FIS server responds with `201` in case of a successful registatration.

firebase-installations/src/main/java/com/google/firebase/installations/remote/RequestLimiter.java
```diff
@@ -45,7 +45,7 @@
   // Based on the response code, calculates the next request time to communicate with the FIS
   // servers.
   public synchronized void setNextRequestTime(int responseCode) {
-    if (isSuccessful(responseCode)) {
+    if (isSuccessfulOrRequiresNewFIDCreation(responseCode)) {
```
andirayo, line 48: camel case:
`isSuccessfulOrRequiresNewFidCreation`

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Oh right. Changed it to throw an exception on 429.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Correct, we should & do not set nextRequestTime on exceptions. Instead retry once before throwing unavailable  exception.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Makes sense. Changed it to throw the exception before the loop. 

Now about the desired behavior: We should throw an exception. But I am unclear if it should state "Firebase Installations Service is unavailable". WDYT?

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
ankitaj224: Changed it . PTAL.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
```diff
@@ -112,6 +115,7 @@ public FirebaseInstallationServiceClient(
     this.context = context;
     this.userAgentPublisher = publisher;
     this.heartbeatInfo = heartbeatInfo;
+    this.requestLimiter = new RequestLimiter(Utils.getInstance());
```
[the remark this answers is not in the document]
ankitaj224: Optional not supported due to minSDK version.
This is for testing this class easily.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -132,7 +132,7 @@ public Thread newThread(Runnable r) {
         new FirebaseInstallationServiceClient(
             firebaseApp.getApplicationContext(), publisher, heartbeatInfo),
         new PersistedInstallation(firebaseApp),
-        new Utils(),
+        Utils.getInstance(),
         new IidStore(firebaseApp),
         new RandomFidGenerator());
```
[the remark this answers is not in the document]
andirayo: (fyi: this was not addressed or acknowledged)

firebase-installations/src/main/java/com/google/firebase/installations/Utils.java
```diff
@@ -31,11 +32,28 @@
...
+  private Utils(Clock clock) {
     this.clock = clock;
   }
+
+  // Factory method that always returns the same Utils instance.
+  public static Utils getInstance() {
+    if (singleton == null) {
+      singleton = new Utils(SystemClock.getInstance());
+    }
+    return singleton;
```
andirayo, line 47: Please avoid code duplication:

```
public static Utils getInstance() {
  return getInstance(SystemClock.getInstance());
}

File: firebase-installations/src/main/java/com/google/firebase/installations/Utils.java [hunk 10]
```diff
@@ -31,11 +32,28 @@
...
+  }
+
+  public static Utils getTestInstance(Clock clock) {
```
andirayo, line 50: Please add JavaDoc, that the first call will define the Clock used.
If a Utils instance has already been initialized, the parameter will be ignored and the old instance will be returned.

firebase-installations/src/main/java/com/google/firebase/installations/remote/FirebaseInstallationServiceClient.java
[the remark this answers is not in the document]
andirayo: By "optional" I mean that you have 2 constructors, one with `Utils` parameter, one without.
The constructor without `Utils` parameter just calls `Utils.getInstance()`, so we don't need this implementation detail here.

andirayo (approved):

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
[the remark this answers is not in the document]
ankitaj224: Acknowledged. I will address this in a following PR.

PR #2009, opened 2020-09-25
Send the GMP app ID along with user agent

var-const:


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
```diff
@@ -29,32 +30,51 @@
...
       Metadata.Key.of("x-firebase-client", Metadata.ASCII_STRING_MARSHALLER);
 
+  private static final Metadata.Key<String> GMP_APP_ID_HEADER =
+      Metadata.Key.of("x-firebase-gmpid", Metadata.ASCII_STRING_MARSHALLER);
+
   public FirebaseClientGrpcMetadataProvider(
       @NonNull Provider<UserAgentPublisher> userAgentPublisherProvider,
-      @NonNull Provider<HeartBeatInfo> heartBeatInfoProvider) {
+      @NonNull Provider<HeartBeatInfo> heartBeatInfoProvider,
+      FirebaseOptions firebaseOptions) {
```
wilhuff, line 49: Declare this `@NonNull` or `@Nullable`.

wilhuff (approved):
LGTM

var-const:
@VinayGuthal Vinay, could you please take a look at this PR as well? @vkryachko @maksymmalyhin Vlad, Maksym, tagging you in case you have any feedback as well. Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
var-const: Done (went with `@Nullable` mostly because I'm not really sure how this thing with dependency injection works and accounting for null seemed safer).

var-const:
Will merge now to make sure this PR makes it into the release.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirebaseClientGrpcMetadataProvider.java
vkryachko: I suggest you use `@NonNull` here as `FirebaseOptions` are always present, I would also (optionally) recommend changing your dependency on `FirebaseOptions` to `required` here: https://github.com/firebase/firebase-android-sdk/blob/22d1615474da1bfb71652447f65c0797818ecdb7/firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java#L51

VinayGuthal:
This looks good to me.. Sorry for the late reply

PR #2012, opened 2020-09-29
Always send the user agent regardless of the heartbeat value

var-const:
Port of https://github.com/firebase/firebase-ios-sdk/pull/6592


wilhuff (approved):
LGTM

vkryachko:
@var-const Looks like this PR broke your integ tests, see the `device-check-changed` failure.

Issue #1996, opened 2020-09-21
Firebase Crashlytics assumes use of proguard rules shipped with the android SDK


jzbrooks:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 4.0.1
  * Firebase Component: Crashlytics (Database, Firestore, Storage, Functions, etc)
  * Component version: 17.2.1

### [REQUIRED] Step 3: Describe the problem

Our team doesn't use the default proguard ruleset. Being more specific with what we keep and don't keep helps us keep our app lean. Since that ruleset isn't included, we don't keep `values()` methods on every enum in the system.

This causes a crash when uploading crashes to firebase. [`Priority`](https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-api/src/main/java/com/google/android/datatransport/Priority.java) is used reflectively in the `EnumMap` within `PriorityMapping`.

```
         AndroidRuntime  E  FATAL EXCEPTION: main
                         E  Process: com.faithlife.mobile, PID: 7912
                         E  java.lang.RuntimeException: Testing crashlytics sdk upgrade
                         E      at com.faithlife.mobile.mainmenu.MainMenuFragment.O0(MainMenuFragment.kt:10)
                         E      at com.faithlife.mobile.mainmenu.MainMenuFragment$onViewCreated$3.d(MainMenuFragment.kt:3)
                         E      at at.onClick(Unknown Source:2)
                         E      at android.view.View.performClick(View.java:6597)
                         E      at com.google.android.material.button.MaterialButton.performClick(MaterialButton.java:2)
                         E      at android.view.View.performClickInternal(View.java:6574)
                         E      at android.view.View.access$3100(View.java:778)
                         E      at android.view.View$PerformClick.run(View.java:25885)
                         E      at android.os.Handler.handleCallback(Handler.java:873)
[...]

PR #2016, opened 2020-09-29
Fix reflection error in com.google.android.datatransport.Priority when using aggressive Proguard rules.

mrichards:
The enum's values are accessed via reflection, so aggressive Proguard
rules can cause it to be stripped out.
e.g., https://github.com/firebase/firebase-android-sdk/issues/1996

### DIFF
--- a/transport/transport-api/proguard.txt
+++ b/transport/transport-api/proguard.txt
@@ -1,2 +1,5 @@
 -dontwarn com.google.auto.value.AutoValue
 -dontwarn com.google.auto.value.AutoValue$Builder
+-keepclassmembers enum com.google.android.datatransport.Priority {
+     *;
+ }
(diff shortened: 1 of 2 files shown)


mrichards:
@vkryachko Are you aware of other places where reflection is used in datatransport? This rule will fix a Crashlytics customer crash. Could there be similar issues lurking for other SDKs?

vkryachko:
@mrichards I think the underlying issue is the fact that datatransport uses an `EnumMap` internally, which for reasons beyond my understanding uses reflection, I got burnt by an `EnumSet` previously and changed it to a plain `HashSet`. I wonder if a cleaner fix in this case would be to change `EnumMap` to `HashMap` in https://github.com/firebase/firebase-android-sdk/blob/12ae453dc92424ce3163089c0d3ec33060b4fbcc/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/util/PriorityMapping.java#L24

mrichards:
> @mrichards ... I wonder if a cleaner fix in this case would be to change `EnumMap` to `HashMap` in...

Yes, that worked. See most recent commit.

vkryachko (approved):
LGTM, thanks for the fix!

PR #2026, opened 2020-09-30
Remove timestampInSnapshots Setting

schmidt-sebastian:
Snapshot data will now always use Timestamp values. Use `Timestamp.toDate()` to convert to `java.util.Date` if required.

`timestampsInSnapshots` was deprecated in January 2019 and has defaulted to true since then.

Port of https://github.com/firebase/firebase-ios-sdk/pull/6622


thebrianchen (approved):
lgtm

PR #2034, opened 2020-10-02
Remove protobuf dependency from RC SDK.

danasilver:


firebase-config/firebase-config.gradle
```diff
@@ -24,6 +24,8 @@ firebaseLibrary {
     publishSources = true
 }
 
+// Compile Protobuf definition files. Remote Config does not depend on Protobuf but includes
+// definitions used by the ABT and FIAM SDKs.
```
vkryachko, line 28: afaict I don't think that is accurate, instead ABT has protos used by fiam and RC, can you please double check that?

rlazo: ABT only exposes `experiment_payload` proto, which is only used in RC in the class removed in this PR. There are other mentions in the configs proto, but it's as bytes

https://osscs.corp.google.com/firebase-sdk/firebase-android-sdk/+/master:firebase-config/src/proto/com/google/android/gms/config/proto/config_persistence.proto;l=42?q=experiment%20payload&ss=firebase-sdk%2Ffirebase-android-sdk:firebase-config%2F

FIAM does depend on the class
https://osscs.corp.google.com/firebase-sdk/firebase-android-sdk/+/master:firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/AbtIntegrationHelper.java;l=57?q=experimentpayload&ss=firebase-sdk%2Ffirebase-android-sdk:firebase-inappmessaging%2F from ABT

firebase-abt/firebase-abt.gradle
```diff
@@ -87,4 +86,5 @@ dependencies {
     testImplementation 'io.grpc:grpc-testing:1.12.0'
     testImplementation 'androidx.test.espresso:espresso-core:3.2.0'
 
+    compileOnly "com.google.protobuf:protobuf-javalite:$javaliteVersion"
```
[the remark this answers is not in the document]
rlazo: AFAIK firebase-abt cannot be used stand alone. Can it?

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -209,11 +205,7 @@ private ConfigCacheClient getCacheClient(String namespace, String configStoreTyp
     return getCacheClient(context, appId, namespace, configStoreType);
   }
 
-  /**
-   * The {@link LegacyConfigsHandler} needs access to multiple cache clients, and the simplest way
-   * to provide it access is to keep this method public and static.
-   */
-  public static ConfigCacheClient getCacheClient(
+  private ConfigCacheClient getCacheClient(
```
rlazo, line 208: from the javadoc seems like the only usefulness of the method was for use with the `LegacyConfigsHandler`, is it still needed?

danasilver: Ah yeah - don't need this method and the one above.

firebase-abt/firebase-abt.gradle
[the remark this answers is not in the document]
danasilver: ABT can't be used alone (since you have to run the ABT test on something else) but we should update the comment if we don't move the proto definition into FIAM.

rlazo (approved):

vkryachko (approved):

firebase-inappmessaging/src/proto/firebase/inappmessaging/proto/experiment_payload.proto
```diff
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 syntax = "proto3";
-package developers.mobile.abt;
 
-option java_outer_classname = "FirebaseAbt";
+package firebase.inappmessaging;
```
eldhosembabu, line 17: Will there be any issues with serialization if we change the package name as I assume there could be a similar proto used at server side?

vkryachko: nope, there should not be any issues

eldhosembabu: Thanks for clarifying...I saw another proto from targeting inside FIAM maintaining the targeting package structure(.../firebase-inappmessaging/src/proto/developers/mobile/targeting/proto/client_signals.proto)...so was a little confused and added this question.

eldhosembabu (approved):

firebase-inappmessaging/src/proto/firebase/inappmessaging/proto/experiment_payload.proto
danasilver: Ah yeah I saw that but decided to keep it as-is since it was already in FIAM.

PR #2053, opened 2020-10-08
Moving proto dependency from abt to owned proto in fiam.

eldhosembabu:
Draft PR

- Moved experiment_payload.proto from abt to fiam project under folder firebase-inappmessaging/src/proto/developers/mobile/abt/experiment_payload.proto.
-Changed proto dependency in firebase-inappmessaging/src/proto/google/internal/firebase/inappmessaging/v1/campaign.proto to use the newly moved proto.


eldhosembabu:
Closing this PR since the change is already addressed in https://github.com/firebase/firebase-android-sdk/pull/2034

PR #2067, opened 2020-10-15
Add support for fireperf e2e test in fireci.

yifanyang:


vkryachko (approved):
👍

ci/fireci/fireciplugins/fireperf.py
```diff
@@ -0,0 +1,85 @@
...
+def _find_fireperf_plugin_version():
+  local_maven_repo_dir = pathlib.Path.home().joinpath('.m2', 'repository')
+  artifacts_path = local_maven_repo_dir.joinpath('com', 'google', 'firebase', 'perf-plugin')
+  versions = [pom.parent.name for pom in artifacts_path.rglob('*.pom')]
+  snapshots = list(filter(lambda v: v.endswith('-SNAPSHOT'), versions))
+
+  if snapshots:
+    return snapshots[0]
+  else:
+    raise click.ClickException('Cannot find a snapshot version in local maven repo.')
```
ramanpreetSinghKhinda, line 85: What if there are multiple versions available? May be cleaning up the local maven repo before kicking the build might help to reduce the noise.

vkryachko: maven local is guarantted to be empty when the job starts, so checking should not be required

ramanpreetSinghKhinda (approved):

PR #2073, opened 2020-10-16
Add CustomModel class framework.

annzimmer:


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
```diff
@@ -0,0 +1,63 @@
...
+ * on a device.
+ */
+public class CustomModel {
+  private String name;
+  private long downloadId;
+  private long fileSize;
+  private String modelHash;
+  private String localFilePath = "";
+
+  public @NonNull String getName() {
```
vkryachko, line 32: nit
```suggestion
  @NonNull public String getName() {
```
here and throughout

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Used to store information about custom models that are being downloaded or are already downloaded
+ * on a device.
+ */
+public class CustomModel {
+  private String name;
+  private long downloadId;
+  private long fileSize;
+  private String modelHash;
+  private String localFilePath = "";
```
vkryachko, line 30: Is the class intended to be mutated? Why not make all fields `final`?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public @NonNull String getName() {
+    return name;
+  }
+
+  public @Nullable File getFile() {
```
vkryachko, line 36: What can a model without a file be use for?

annzimmer: It can be used to get the downloadId, which can be used to get the progress of the file's download status.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
annzimmer: Added a comment to the class. At this point I'm just adding the public parts - enough to get the API headers written in the next PR. The file associated with the model can be updated but will only be supported as a protected method call via the downloader. So no public way to modify the class. The name however should be final. There will be a few other update calls and private fields in the next few PRs.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
annzimmer: done.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
vkryachko: Not sure I understand. How can I get a hold of an instance of `CustomModel` before the download has completed?
It looks like the only ways to get a `CustomModel` is to call download and wait for its task to complete, or list all "downloaded" models

annzimmer: Good point, I forgot to add the addOnProgressListener to the getModel in the API design - it was discussed and planned (hence the download id). So, I added example. In the addOnProgressListener, the CustomModel is returned with the download id and can then be used to updated the progress bar, etc. Does this clarify?

vkryachko: Thanks, it does. But it also reveals a potential design issue, namely, most of the getters have the following javadoc: "When models is not downloaded returns X, otherwise Y".

imo, the fact that the model has 2 modes of operation, indicates that type has low cohesion, and really represents 2 separate things:

* `ModelDownload` - contains info about the download that is in progress. Contains (modelName, downloadId), passed to onProgressListeners
* `CustomModel` - describes a fully downloaded model. Contains (name, file, size, hash), available only upon download completion.

This could reduce developer confusion, as they won't have to guess which state they are in(downloading vs downloaded). wdyt?

annzimmer: We will actually have everything except the file during the progress listener. We have the model size and hash in advanced. So the difference is during "in progress" you'd use the download id to get download status information (which could be used to see if the download had started and % complete) and during "on success" you'd get the actual model file. We added the in progress functionality with download id so that a progress bar could be created and populated as desired.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/CustomModel.java
vkryachko: Making this class mutable and updating it as model state changes has many consequences:

1. It must be made thread-safe as updates to it may happen on a thread different from where the developer is using it.
2. You would also have to keep track of all instances of `CustomModel` that are being used and update them all.

So my question is, is it worth it? Can the class be made immutable and for any state updates a new `CustomModel` instance can be created and handed out to progress listeners?

vkryachko (approved):

PR #2077, opened 2020-10-19
Add tests for Crashlytics DataCollectionArbiter

mrichards:
Basic test for the setter and getter. The real benefit of this change
is catching any changes to the internal API that will break the Unity
plugin.


mrwillis21 (approved):
Nice!

PR #2078, opened 2020-10-19
Make some changes that were causing build warnings/errors on copybara import

tejasd:
Errors/Warnings fixed:

1. https://errorprone.info/bugpattern/FutureReturnValueIgnored
2. Requirement of strings to be translatable
3. Compress certain images which can be compressed lossless
4. Add minSdkVersion to 16 (for non gradle builds)


firebase-crashlytics/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,7 @@
           package="com.google.firebase.crashlytics" >
 
     <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-sdk android:minSdkVersion="16"/>
```
mrwillis21, line 6: Can we add a comment that this is for non-gradle builds, a la https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/AndroidManifest.xml#L19 ?

tejasd: Done! Thanks for catching that! (Also I mistakenly looked at post copybara change :) (copybara uncomments <uses-sdk android:minSdkVersion="16"/>)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsBackgroundWorker.java
```diff
@@ -44,6 +44,7 @@
   // A thread local to keep track of which thread belongs to this executor.
   private ThreadLocal<Boolean> isExecutorThread = new ThreadLocal<>();
 
+  @SuppressWarnings("FutureReturnValueIgnored")
```
mrwillis21, line 47: Any chance we can scope this to line 51 instead of having it on the whole constructor?

firebase-crashlytics/src/main/AndroidManifest.xml
mrwillis21: Thanks for fixing that up! :)

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsBackgroundWorker.java
tejasd: Is this better?

mrwillis21: Yes, thank you! :)

mrwillis21 (approved):

Issue #2069, opened 2020-10-16
Uncaught exception in Firebase Database runloop (19.5.0)


imKarthikMuru:
After updating Android Studio 4.1 facing java.lang.RuntimeException: Uncaught exception in Firebase Database runloop (19.5.0). Tried many ways but no use.

#Firebase #1936

mandeepsinghn:
I am getting the same error with the following debug Log:
E/Persistence: Caught Throwable.
    java.lang.AssertionError: hardAssert failed: 
        at com.google.firebase.database.core.utilities.Utilities.hardAssert(Utilities.java:234)
        at com.google.firebase.database.core.utilities.Utilities.hardAssert(Utilities.java:229)
        at com.google.firebase.database.snapshot.ChildrenNode.updateChild(ChildrenNode.java:293)
        at com.google.firebase.database.snapshot.ChildrenNode.updateChild(ChildrenNode.java:296)
        at com.google.firebase.database.snapshot.ChildrenNode.updateChild(ChildrenNode.java:296)
        at com.google.firebase.database.snapshot.ChildrenNode.updateChild(ChildrenNode.java:296)
        at com.google.firebase.database.android.SqlPersistenceStorageEngine.loadNested(SqlPersistenceStorageEngine.java:979)
        at com.google.firebase.database.android.SqlPersistenceStorageEngine.serverCache(SqlPersistenceStorageEngine.java:350)
        at com.google.firebase.database.core.persistence.DefaultPersistenceManager.serverCache(DefaultPersistenceManager.java:160)
        at com.google.firebase.database.core.SyncTree$11.call(SyncTree.java:513)
        at com.google.firebase.database.core.SyncTree$11.call(SyncTree.java:461)
        at com.google.firebase.database.core.persistence.DefaultPersistenceManager.runInTransaction(DefaultPersistenceManager.java:239)
        at com.google.firebase.database.core.SyncTree.addEventRegistration(SyncTree.java:460)
        at com.google.firebase.database.core.SyncTree.keepSynced(SyncTree.java:736)
        at com.google.firebase.database.core.Repo.keepSynced(Repo.java:647)
        at com.google.firebase.database.Query$4.run(Query.java:255)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:428)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
[...]

PR #2079, opened 2020-10-19
Enforce asserts only in debug builds

schmidt-sebastian:
In 19.4, we replaced Java `assert` statements with custom asserts to address lint warnings, as `assert` is disabled on Android. This however caused the clients to crash, which indicates that at least some of our asserts are problematic. We can safely revert to the previous state by only enabling these asserts in debug builds.

Fixes https://github.com/firebase/firebase-android-sdk/issues/2069

### DIFF
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -1,6 +1,12 @@
-# Unreleased
+# 19.5.1 (Unreleased)
+- [fixed] Fixes a regression in v19.4 that may cause assertion failures,
+  especially when persistence is enabled.
+
+# 19.5.0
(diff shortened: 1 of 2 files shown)


VinayGuthal (approved):

vkryachko (approved):
Generally LGTM, since it used to be the behavior prior to migrating off of `assert`, but could this issue indicate that there are subtle bugs that cause internal state to become in violation of its invariants? Is it worth digging deeper into the root cause of those failures?

schmidt-sebastian:
@vkryachko Yes to all. Note that most of these asserts also exist on other platforms and do not fire.

PR #2085, opened 2020-10-20
Implement Deferred<T> lint check.

vkryachko:
The check ensures that methods marked with `@DeferredApi` are only used
in the context of `Deferred#whenAvailable(DeferredHandler)` callback.


tools/lint/src/main/kotlin/DeferredApiDetector.kt
```diff
@@ -0,0 +1,153 @@
...
+                UMethod::class.java, true,
+                UAnonymousClass::class.java, ULambdaExpression::class.java
+        ) as? PsiMethod
+        return hasAnnotation(context, method)
+    }
+
+    private fun lambdaMethod(element: UElement): PsiMethod? {
+        val lambda = element.getParentOfType<ULambdaExpression>(
+                ULambdaExpression::class.java, true, UMethod::class.java, UAnonymousClass::class.java)
+        if (lambda != null) {
```
rlazo, line 84: To reduce nesting 

```
if (lambda == null) {
 return null
}
```

tools/lint/src/main/kotlin/DeferredApiDetector.kt
```diff
@@ -0,0 +1,153 @@
...
+                    return resolved.allMethods.firstOrNull { it.hasModifier(JvmModifier.ABSTRACT) }
+                }
+            }
+        }
+        return null
+    }
+
+    private fun hasAnnotation(context: JavaContext, calledMethod: PsiMethod?): Boolean {
+        var method = calledMethod
+        if (method != null) {
```
rlazo, line 98: same

rlazo (approved):

PR #2087, opened 2020-10-21
Query get API for RTDB

IanWyszynski:
Port of https://github.com/firebase/firebase-js-sdk/pull/3812

### DIFF
--- a/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
+++ b/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
@@ -3431,7 +3431,6 @@ public class QueryTest {
     FirebaseApp app =
         appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
-    db.setLogLevel(Logger.Level.DEBUG);
     DatabaseReference node = db.getReference();
     db.goOffline();
     try {
--- a/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
+++ b/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
@@ -36,7 +36,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
@@ -3422,50 +3421,48 @@ public class QueryTest {
   @Test
   public void emptyQueryGet() throws DatabaseException, InterruptedException {
     FirebaseApp app =
-        appForDatabaseUrl(IntegrationTestValues.getAltNamespace(), UUID.randomUUID().toString());
+        appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
-    db.useEmulator("10.0.2.2", 9000);
-    DatabaseReference node = db.getReference();
     try {
-      Tasks.await(node.get());
+      Tasks.await(db.getReference("dummy/").setValue(42L));
+      assertNull(Tasks.await(db.getReference("null/").get()).getValue());
     } catch (ExecutionException e) {
-      assertEquals(e.getCause().getMessage(), "Client is offline");
-      return;
+      fail(e.getMessage());
     }
-    fail();
   }
 
   @Test
   public void offlineQueryGet() throws DatabaseException, InterruptedException {
     FirebaseApp app =
-        appForDatabaseUrl(IntegrationTestValues.getAltNamespace(), UUID.randomUUID().toString());
+        appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
-    db.useEmulator("10.0.2.2", 9000);
     DatabaseReference node = db.getReference();
-    DatabaseConfig cfg = IntegrationTestHelpers.newTestConfig();
-    IntegrationTestHelpers.goOffline(cfg);
+    try {
+      Tasks.await(node.setValue(42L));
+    } catch (ExecutionException e) {
+      fail();
+    }
+    db.goOffline();
     try {
       Tasks.await(node.get());
     } catch (ExecutionException e) {
       assertEquals(e.getCause().getMessage(), "Client is offline");
       return;
     }
-    fail();
+    fail("Client get succeeded even though offline.");
   }
 
   @Test
   public void getQueryBasic() throws DatabaseException, InterruptedException {
     FirebaseApp app =
-        appForDatabaseUrl(IntegrationTestValues.getAltNamespace(), UUID.randomUUID().toString());
+        appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
-    db.useEmulator("10.0.2.2", 9000);
     DatabaseReference node = db.getReference();
-    final Semaphore semaphore = new Semaphore(0);
     try {
       Tasks.await(node.setValue(42));
       assertEquals(42L, Tasks.await(node.get()).getValue());
     } catch (ExecutionException e) {
-      fail();
+      fail(e.getMessage());
     }
   }
 
@@ -3475,57 +3472,31 @@ public class QueryTest {
     FirebaseApp app =
         appForDatabaseUrl(IntegrationTestValues.getAltNamespace(), UUID.randomUUID().toString());
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
-    db.useEmulator("10.0.2.2", 9000);
     DatabaseReference ref = db.getReference();
-    DatabaseConfig cfg = IntegrationTestHelpers.newTestConfig();
-    ReadFuture future = ReadFuture.untilNonNull(ref);
-    ref.setValue(42);
-    assertEquals(42L, future.waitForLastValue());
-    IntegrationTestHelpers.goOffline(cfg);
-    try {
-      assertEquals(42L, Tasks.await(ref.get()).getValue());
-    } catch (ExecutionException e) {
-      fail();
-    }
-  }
-
-  @Test
-  public void getQueryHitsCacheWhenOffline()
-      throws InterruptedException, ExecutionException, TimeoutException, TestFailure {
-    FirebaseApp readerApp =
-        appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
-    FirebaseApp writerApp =
-        appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
-    FirebaseDatabase readerDb = FirebaseDatabase.getInstance(readerApp);
-    FirebaseDatabase writerDb = FirebaseDatabase.getInstance(writerApp);
-    readerDb.useEmulator("10.0.2.2", 9000);
-    writerDb.useEmulator("10.0.2.2", 9000);
-    DatabaseReference reader = readerDb.getReference("/foo");
-    DatabaseReference writer = writerDb.getReference("/foo");
-
-    WriteFuture write = new WriteFuture(writer, 42L);
-    assertNull(write.timedGet());
-
     final Semaphore semaphore = new Semaphore(0);
-    reader.addValueEventListener(
+    ValueEventListener listener =
         new ValueEventListener() {
           @Override
           public void onDataChange(@NonNull DataSnapshot snapshot) {
-            if (Objects.requireNonNull(snapshot.getValue()).equals(42L)) {
+            if (snapshot.getValue() != null && snapshot.getValue().equals(42L)) {
               semaphore.release();
             }
           }
 
           @Override
           public void onCancelled(@NonNull DatabaseError error) {}
-        });
+        };
+    ref.addValueEventListener(listener);
+    ref.setValue(42L);
     IntegrationTestHelpers.waitFor(semaphore);
-    reader.getDatabase().goOffline();
+    db.goOffline();
     try {
-      DataSnapshot snapshot = Tasks.await(reader.get());
-      assertEquals(42L, snapshot.getValue());
+      // Since we still have a listener on `ref`, the 42L should be cached here.
+      assertEquals(42L, Tasks.await(ref.get()).getValue());
     } catch (ExecutionException e) {
-      fail();
+      fail(e.getMessage());
+    } finally {
+      ref.removeEventListener(listener);
     }
   }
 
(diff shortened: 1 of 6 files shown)


IanWyszynski:
@schmidt-sebastian Can you take a quick pass over this to check implementation strategy?

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3406,101 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+    DatabaseConfig cfg = IntegrationTestHelpers.newTestConfig();
+    final Semaphore semaphore = new Semaphore(0);
+    node.get()
+        .addOnCompleteListener(
+            new OnCompleteListener<DataSnapshot>() {
+              @Override
+              public void onComplete(@NonNull Task<DataSnapshot> task) {
+                assertTrue(task.isSuccessful());
+                assertNotNull(task.getException());
+                assertEquals(task.getException().getMessage(), "Client offline with empty cache!");
```
schmidt-sebastian, line 3421: Nit: Drop the exclamation mark.

But my larger question is why we are offline here. Is this because we kick off going online only after this API call finishes (like we discussed before)?

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3406,101 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
     IntegrationTestHelpers.waitFor(semaphore);
   }
 
+  @Test
+  public void emptyQueryGet() throws DatabaseException, InterruptedException {
+    DatabaseReference node = IntegrationTestHelpers.getRandomNode();
+    DatabaseConfig cfg = IntegrationTestHelpers.newTestConfig();
+    final Semaphore semaphore = new Semaphore(0);
+    node.get()
```
schmidt-sebastian, line 3414: FWIW, you can just do:

```
try {
DataSnapshot snap = node.get().get();
fail(...);
} catch (ExcecutionException e) {
...
}
```

Optional... especially since I just remembered that this might not work since it may block the UI thread that is running the test.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3406,101 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+                assertEquals(task.getException().getMessage(), "Client offline with empty cache!");
+              }
+            });
+  }
+
+  @Test
+  public void getQueryBasic() throws DatabaseException, InterruptedException {
+    DatabaseReference ref = IntegrationTestHelpers.getRandomNode();
+    final Semaphore semaphore = new Semaphore(0);
+    ref.setValue(42)
```
schmidt-sebastian, line 3450: If calling `.get()` works for these tests, this could be as simple as

`ref.setValue(42).get()` followed by `ref.get().get()`.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -344,6 +346,21 @@ public void listen(
...
 
+  @Override
+  public Task<Object> get(List<String> path, Map<String, Object> queryParams) {
+    QuerySpec query = new QuerySpec(path, queryParams);
+    TaskCompletionSource<Object> source = new TaskCompletionSource<>();
+    Task<Object> task;
+    if (connected()) {
+      task = sendGet(query, source);
+    } else {
+      source.setException(new Exception("Client offline!"));
```
schmidt-sebastian, line 357: Please drop exclamation mark as this is user facing. Maybe make this "Client is offline".

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -462,6 +467,37 @@ public void onRequestResult(String optErrorCode, String optErrorMessage) {
...
+        .continueWithTask(
+            new Continuation<Object, Task<DataSnapshot>>() {
+              @Override
+              public Task<DataSnapshot> then(@NonNull Task<Object> task) throws Exception {
+                TaskCompletionSource<DataSnapshot> source = new TaskCompletionSource<>();
+                if (!task.isSuccessful()) {
+                  Node cached =
+                      serverSyncTree.calcCompleteEventCache(query.getPath(), new ArrayList<>());
+                  if (cached.isEmpty()) {
+                    source.setException(new Exception("Client offline with empty cache!"));
```
schmidt-sebastian, line 482: Please also port the logging changes from JS.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -1049,6 +1066,32 @@ public void onResponse(Map<String, Object> response) {
...
+    request.put(REQUEST_PATH, ConnectionUtils.pathToString(query.path));
+    request.put(REQUEST_QUERIES, query.queryParams);
+    sendAction(
+        REQUEST_ACTION_QUERY,
+        request,
+        new ConnectionRequestCallback() {
+          @Override
+          public void onResponse(Map<String, Object> response) {
+            String status = (String) response.get(REQUEST_STATUS);
+            if (status.equals("ok")) {
```
schmidt-sebastian, line 1080: Do we need to special case `404` as well and treat it as a successful response?

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -1049,6 +1066,32 @@ public void onResponse(Map<String, Object> response) {
...
+              String pathString = (String) response.get(SERVER_DATA_UPDATE_PATH);
+              List<String> path = ConnectionUtils.stringToPath(pathString);
+              Object body = response.get(SERVER_DATA_UPDATE_BODY);
+              Long tagNumber = ConnectionUtils.longFromObject(response.get(SERVER_DATA_TAG));
```
schmidt-sebastian, line 1084: Does this need to be added to the Web PR?

IanWyszynski: No I think this was a mistake in this PR. Gets won't send tags.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
IanWyszynski: We would return null if no data matches the query.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
IanWyszynski: My intent for this test was to check that the proper exception is returned by get() when the client is offline.

IanWyszynski: That is a mistake, this test should never be offline.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3407,140 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+    IntegrationTestHelpers.waitFor(semaphore);
+    db.goOffline();
+    try {
+      // Since we still have a listener on `ref`, the 42L should be cached here.
+      assertEquals(42L, Tasks.await(ref.get()).getValue());
+    } catch (ExecutionException e) {
+      fail(e.getMessage());
+    } finally {
+      ref.removeEventListener(listener);
+    }
```
schmidt-sebastian, line 3500: Please also add a test that shows the the value is still cached even when there are no remaining listeners. The SDK should keep up to 10 MB of least recently used data. For this to work, you have to track the query via the persistence manager. See here: https://github.com/firebase/firebase-android-sdk/blob/d92096257b3e750c72757fc7c4bc9d7393db2046/firebase-database/src/main/java/com/google/firebase/database/core/SyncTree.java#L505

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3407,140 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+          @Override
+          public void onDataChange(@NonNull DataSnapshot snapshot) {
+            if (snapshot.getValue() != null && snapshot.getValue().equals(42L)) {
+              semaphore.release();
+            }
+          }
+
+          @Override
+          public void onCancelled(@NonNull DatabaseError error) {}
+        };
```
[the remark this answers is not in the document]
schmidt-sebastian: After talking offline, this is not possible since we need to run line 3489 and 3490.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3407,165 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+          public void onDataChange(@NonNull DataSnapshot snapshot) {
+            if (snapshot.getValue() != null && snapshot.getValue().equals(42L)) {
+              semaphore.release();
+            }
+          }
+
+          @Override
+          public void onCancelled(@NonNull DatabaseError error) {}
+        });
+    IntegrationTestHelpers.waitFor(semaphore);
```
schmidt-sebastian, line 3566: I think my comment earlier wasn't applicable... but it does seem that this is now a ReadFuture.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -109,6 +113,41 @@ public String toString() {
     }
   }
 
+  private static class OutstandingGet {
+    private Map<String, Object> request;
+    private ConnectionRequestCallback onComplete;
+    private String action;
+    private AtomicBoolean sent;
```
schmidt-sebastian, line 120:
```suggestion
    private final Map<String, Object> request;
    private final ConnectionRequestCallback onComplete;
    private final String action;
    private final AtomicBoolean sent;
```

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -109,6 +113,41 @@ public String toString() {
...
+      return request;
+    }
+
+    private boolean markSent() {
+      return sent.compareAndSet(false, true);
+    }
+
+    private boolean wasSent() {
+      return sent.get();
+    }
```
schmidt-sebastian, line 148: This method seems unused.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -344,6 +388,64 @@ public void listen(
...
+              }
+            }));
+
+    if (!connected()) {
+      executorService.schedule(
+          new Runnable() {
+            @Override
+            public void run() {
+              OutstandingGet get = outstandingGets.get(readId);
+              if (get == null || !get.markSent()) {
```
schmidt-sebastian, line 428: It took me a while to wade through all the callsites, but it seems like `executorService` is single-threaded and the is responsible for all access to `outstandingGets`. This means that `markSent` doesn't need to use an AtomicBoolean.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -1049,6 +1162,34 @@ public void onResponse(Map<String, Object> response) {
         });
   }
 
+  private void sendGet(final Long readId) {
+    hardAssert(canSendReads(), "sendGet called when we can't send gets");
+    OutstandingGet get = outstandingGets.get(readId);
+    if (!get.markSent()) {
+      if (logger.logsDebug()) {
+        logger.debug("get" + readId + " already sent or cancelled, ignoring.");
```
schmidt-sebastian, line 1170: This should only ever be hit when cancelled, or not?

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -1049,6 +1162,34 @@ public void onResponse(Map<String, Object> response) {
...
+          public void onResponse(Map<String, Object> response) {
+            OutstandingGet currentGet = outstandingGets.get(readId);
+            if (currentGet == get) {
+              outstandingGets.remove(readId);
+              get.getOnComplete().onResponse(response);
+            } else {
+              if (logger.logsDebug())
+                logger.debug(
+                    "Ignoring on complete for get " + readId + " because it was removed already.");
+            }
```
schmidt-sebastian, line 1188:
```suggestion
            } else if (logger.logsDebug()) {
                logger.debug(
                    "Ignoring on complete for get " + readId + " because it was removed already.");
            }
```

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -462,6 +468,61 @@ public void onRequestResult(String optErrorCode, String optErrorMessage) {
...
+                    });
+          }
+        });
+    return source
+        .getTask()
+        .continueWithTask(
+            new Continuation<DataSnapshot, Task<DataSnapshot>>() {
+              @Override
+              public Task<DataSnapshot> then(@NonNull Task<DataSnapshot> task) throws Exception {
+                serverSyncTree.setQueryInactive(query.getSpec());
```
schmidt-sebastian, line 520: Can this be done in the `OnCompletedListener`?

firebase-database/src/main/java/com/google/firebase/database/core/SyncTree.java
```diff
@@ -454,6 +454,28 @@ public boolean isEmpty() {
...
+  public void setQueryInactive(QuerySpec query) {
+    persistenceManager.runInTransaction(
+        new Callable<Void>() {
+          @Override
+          public Void call() {
+            persistenceManager.setQueryInactive(query);
+            return null;
+          }
+        });
+  }
```
schmidt-sebastian, line 477: This is much easier than I thought it would be. Thanks for figuring this out.

schmidt-sebastian (commented):
Thanks for the updates, looks very nice!

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -140,11 +139,12 @@ private ConnectionRequestCallback getOnComplete() {
...
-    }
-
-    private boolean wasSent() {
-      return sent.get();
+      boolean prev = sent;
+      if (prev) {
+        return false;
+      }
+      this.sent = true;
+      return true;
```
schmidt-sebastian, line 141: Looks like we can inline `prev` and drop the `this.`.

This function should probably also have a short comment on what the return type is - or a more descriptive name.

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -417,15 +416,16 @@ public void onResponse(Map<String, Object> response) {
...
+    outstandingGets.put(readId, outstandingGet);
 
     if (!connected()) {
       executorService.schedule(
           new Runnable() {
             @Override
             public void run() {
               OutstandingGet get = outstandingGets.get(readId);
-              if (get == null || !get.markSent()) {
+              if (get == null || get != outstandingGet || !get.markSent()) {
```
schmidt-sebastian: Could we just use `outstandingGet` here rather than retrieve it from the list?

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3401,6 +3407,145 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+    FirebaseDatabase db = FirebaseDatabase.getInstance(app);
+    assertNull(Tasks.await(db.getReference(UUID.randomUUID().toString()).get()).getValue());
+  }
+
+  @Test
+  public void offlineQueryGet() throws DatabaseException, InterruptedException {
+    FirebaseApp app =
+        appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
+    FirebaseDatabase db = FirebaseDatabase.getInstance(app);
+    db.setLogLevel(Logger.Level.DEBUG);
```
schmidt-sebastian, line 3434: Please remove

schmidt-sebastian (approved):

PR #2095, opened 2020-10-22
Implement Provider.get() assignment lint check.

vkryachko:


tools/lint/src/main/kotlin/ProviderAssignmentDetector.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    private fun isProviderGet(method: PsiMethod): Boolean {
+        if (!method.parameterList.isEmpty) {
+            return false
+        }
+        return (method.parent as? PsiClass)?.let {
+            return it.qualifiedName == PROVIDER
+        } ?: false
```
rlazo, line 61: I think you can go with 

```
(method.parent as? PsiClass)?.let {
            return it.qualifiedName == PROVIDER
        }
return false
```

and simplify it a bit since the `return` is closer to the `false` that get evaluated conditionally by the `?:`

rlazo (approved):

PR #2098, opened 2020-10-23
Add FirebaseModelDownloader public API Headers

annzimmer:
As approved during API review.

### DIFF
--- a/firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
+++ b/firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
@@ -14,10 +14,13 @@
 package com.google.firebase.ml.modeldownloader;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
```diff
@@ -50,6 +53,38 @@ public static FirebaseModelDownloader getInstance(@NonNull FirebaseApp app) {
     return app.get(FirebaseModelDownloader.class);
   }
 
+  // Get the downloaded model file based on download type and conditions.
```
rlazo, line 56: Usually for multiline comments we use `/*` instead of `//`. 

Also, if you want to get the comment below rendered as a list you would need to use `<ul>` and `<li>`, eg https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/Transaction.java#L38

annzimmer: Done - thanks.

rlazo (approved):

PR #2100, opened 2020-10-24
Add custom domain support to callable functions

kroikie:
### Discussion
[Googlers: see API review here](https://docs.google.com/document/d/1cSIcGCW2QR1OLHX4RDCVNhpBUolnG6tfNpJBR2pn1aM/edit#)

### Testing
- Added new unit tests

### API Changes
- API reviewed

This change is in line with other similar changes for JS and iOS sdk:
[JS implemented](https://github.com/firebase/firebase-js-sdk/pull/3825)
[iOS implemented](https://github.com/firebase/firebase-ios-sdk/pull/6787)

### DIFF
--- a/firebase-functions/firebase-functions.gradle
+++ b/firebase-functions/firebase-functions.gradle
@@ -48,12 +48,6 @@ android {
     testOptions.unitTests.includeAndroidResources = true
 }
 
-configurations {
-    all {
-        exclude module: 'commons-logging'
-    }
-}
-
 dependencies {
     implementation project(':firebase-common')
     implementation project(':firebase-components')
@@ -69,7 +63,6 @@ dependencies {
     implementation 'com.google.firebase:firebase-iid-interop:17.0.0'
 
     implementation 'com.squareup.okhttp3:okhttp:3.12.1'
-    implementation 'commons-validator:commons-validator:1.7'
 
     annotationProcessor 'com.google.auto.value:auto-value:1.6.2'
 
--- a/firebase-functions/firebase-functions.gradle
+++ b/firebase-functions/firebase-functions.gradle
@@ -48,6 +48,12 @@ android {
     testOptions.unitTests.includeAndroidResources = true
 }
 
+configurations {
+    all {
+        exclude module: 'commons-logging'
+    }
+}
+
 dependencies {
     implementation project(':firebase-common')
     implementation project(':firebase-components')
@@ -63,6 +69,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-iid-interop:17.0.0'
 
     implementation 'com.squareup.okhttp3:okhttp:3.12.1'
+    implementation 'commons-validator:commons-validator:1.7'
 
     annotationProcessor 'com.google.auto.value:auto-value:1.6.2'
 
--- a/firebase-functions/src/androidTest/java/com/google/firebase/functions/FirebaseFunctionsTest.java
+++ b/firebase-functions/src/androidTest/java/com/google/firebase/functions/FirebaseFunctionsTest.java
@@ -65,6 +65,14 @@ public class FirebaseFunctionsTest {
 
     URL withRegion = functionsWithRegion.getURL("my-endpoint");
     assertEquals("http://10.0.2.2:5001/my-project/my-region/my-endpoint", withRegion.toString());
+
+    FirebaseFunctions functionsWithCustomDomain =
+        FirebaseFunctions.getInstance(app, "https://mydomain.com");
+    functionsWithCustomDomain.useEmulator("10.0.2.2", 5001);
+
+    URL withCustomDOmain = functionsWithCustomDomain.getURL("my-endpoint");
+    assertEquals(
+        "http://10.0.2.2:5001/my-project/us-central1/my-endpoint", withCustomDOmain.toString());
   }
 
   @Test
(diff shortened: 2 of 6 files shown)


kroikie:
This PR combines the region and custom domain in a single `regionOrCustomDomain` FirebaseFunctions constructor parameter. This parameter is checked for being a valid URL, this check is done using the Apache commons UrlValidator. This option provided a clean way of checking if `regionOrCustomDomain` was a valid URL or not.

If adding this dependency is an issue there are other options that I would be happy to switch to:

1. Try to create a URL and catch the resulting exception (like done in the JS implementation)
1. Use a RegEx to see if the parameter is a URL or not
1. Check if the parameter starts with `http` or `https` since no region will have this prefix

firebase-functions/firebase-functions.gradle
```diff
@@ -63,6 +69,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-iid-interop:17.0.0'
 
     implementation 'com.squareup.okhttp3:okhttp:3.12.1'
+    implementation 'commons-validator:commons-validator:1.7'
```
samtstern, line 72: I think we can get rid of this new dependency (see comments below)

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -73,26 +74,40 @@
...
+      String regionOrCustomDomain,
       ContextProvider contextProvider) {
     this.app = app;
     this.client = new OkHttpClient();
     this.serializer = new Serializer();
     this.contextProvider = Preconditions.checkNotNull(contextProvider);
     this.projectId = Preconditions.checkNotNull(projectId);
-    this.region = Preconditions.checkNotNull(region);
+
+    UrlValidator validator = new UrlValidator(customDomainSchemes);
```
samtstern, line 103: I think we can just use the standard library `URL` class and do `new URL(regionOrCustomDomain)` and then catch `MalformedUrlException`:
https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#URL(java.lang.String)

Not quite as good as what you have but worth saving a whole dependency.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -73,26 +74,40 @@
...
+    if (validator.isValid(regionOrCustomDomain)) {
+      this.region = null;
```
samtstern, line 105: If you look at the iOS and JS PRs we set `region` to the default value here.  This is needed in case the person calls `useEmulator`

firebase-functions/src/androidTest/java/com/google/firebase/functions/FirebaseFunctionsTest.java
```diff
@@ -38,6 +38,14 @@ public void testGetUrl() {
...
     url = functions.getURL("my-endpoint");
     assertEquals("https://us-central1-my-project.cloudfunctions.net/my-endpoint", url.toString());
+
+    functions = FirebaseFunctions.getInstance(app, "https://mydomain.com");
+    url = functions.getURL("my-endpoint");
+    assertEquals("https://mydomain.com/my-endpoint", url.toString());
+
+    functions = FirebaseFunctions.getInstance(app, "https://mydomain.com/foo");
+    url = functions.getURL("my-endpoint");
+    assertEquals("https://mydomain.com/foo/my-endpoint", url.toString());
```
samtstern, line 48: See my comment below, but add a test about using `useEmulator` with a custom domain.  I think it would fail right now.

samtstern (changes requested):

firebase-functions/firebase-functions.gradle
kroikie: Done

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
kroikie: Done

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
kroikie: Done, created a DEFAULT_REGION constant seems like `us-central1` literal is used quite a lot.

firebase-functions/src/androidTest/java/com/google/firebase/functions/FirebaseFunctionsTest.java
kroikie: Done.

kroikie:
@samtstern looks like the smoke test has an issue with ML:

`Could not find com.google.firebase:firebase-ml-natural-language`

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
```diff
@@ -85,14 +88,29 @@
...
+
+    boolean isRegion;
+    try {
+      new URL(regionOrCustomDomain);
+      isRegion = false;
+    } catch (MalformedURLException malformedURLException) {
+      isRegion = true;
+    }
+
+    if (isRegion) {
```
samtstern, line 107: Nit: you could just move this logic inside of the `try/catch` and not need `isRegion` at all.  But totally fine either way.

samtstern (approved):

samtstern:
@kroikie LGTM I don't think the smoke test thing is your fault but hopefully someone else can chime in.

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
kroikie: Since `region` and `customDomain` are final they can't be set both in the try and the catch.

I'm gussing that Java considers the whole try and not just the lines that could cause the exception.

samtstern: Ah interesting, that's fair then!

davidmotson (approved):

PR #2108, opened 2020-10-27
Add development_platform for Unity crash reports

mrichards:
Crashlytics reports sent via DataTransport erroneously did not set the
development_platform field for Unity apps. This results in incorrect
internal metrics for Unity apps. b/171214452

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
@@ -137,12 +137,20 @@ public class CrashlyticsReportDataCapture {
   }
 
   private CrashlyticsReport.Session.Application populateSessionApplicationData() {
-    return CrashlyticsReport.Session.Application.builder()
-        .setIdentifier(idManager.getAppIdentifier())
-        .setVersion(appData.versionCode)
-        .setDisplayVersion(appData.versionName)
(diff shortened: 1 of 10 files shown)


mrichards:
@mrwillis21 I'm looking for feedback on the overall approach, and if you think it is worth adding in the install_source field in a similar way.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
```diff
@@ -137,12 +137,20 @@ public Event captureEventData(
   }
 
   private CrashlyticsReport.Session.Application populateSessionApplicationData() {
-    return CrashlyticsReport.Session.Application.builder()
-        .setIdentifier(idManager.getAppIdentifier())
-        .setVersion(appData.versionCode)
-        .setDisplayVersion(appData.versionName)
-        .setInstallationUuid(idManager.getCrashlyticsInstallId())
-        .build();
+    CrashlyticsReport.Session.Application.Builder builder =
```
mrwillis21, line 140: For consistency, can we make this `final`?

mrwillis21 (approved):
One minor nit, but otherwise this seems great to me.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java
mrichards: Done

PR #2115, opened 2020-10-28
Allow null and NaN values in filters

thebrianchen:
Porting from [web](https://github.com/firebase/firebase-js-sdk/pull/3960).


wilhuff (approved):
LGTM

PR #2137, opened 2020-11-04
Fix a race condition in LazySet.

vkryachko:
The race could result in us not updating the actualSet with the newly
aded provider.


rlazo (approved):
Great catch!

PR #2162, opened 2020-11-09
Fix broken code links in comments of SetOptions.java

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
```diff
@@ -26,10 +26,10 @@
 
 /**
  * An options object that configures the behavior of {@code set()} calls. By providing one of the
- * SetOptions objects returned by {@link #merge}, {@link #mergeFields} and {@link #mergeFieldPaths},
- * the {@code set()} calls in {@link DocumentReference}, {@link WriteBatch} and {@link Transaction}
- * can be configured to perform granular merges instead of overwriting the target documents in their
- * entirety.
+ * SetOptions objects returned by {@link #merge()}, {@link #mergeFields(List<String>)} and {@link
```
vkryachko, line 29: I don't think it's valid syntax. it should be something like:
```suggestion
 * SetOptions objects returned by {@link #merge()}, {@link #mergeFields(List)} and {@link
```

i.e. without generic parameters.

vkryachko (commented):
SetOptions.java is not the only file mentioned in the bug, please make sure you fix the rest of the issues.

firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
wu-hui: Done.

wu-hui (commented):
The original bug does not mention, but I think the other place is DocumentId.java?

vkryachko (approved):
Thanks

PR #2195, opened 2020-11-23
Adding a FidListener that propagates fid changes to the clients.

ankitaj224:


firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -51,4 +52,12 @@
...
   @NonNull
   Task<Void> delete();
+
+  /**
+   * Register a listener to receive fid changes.
+   *
+   * @param listener implementation of the {@code FidListener} to handle fid changes.
+   * @hide
+   */
+  void registerFidListener(FidListener listener);
```
vkryachko, line 62: Please add `@DeferredApi` annotation to this method. It's needed for proper dynamic module loading that is upcoming across Firebase SDKs.

vkryachko: Also, it might make sense to add a `NonNull` annotation onto the listener argument.

firebase-installations-interop/src/main/java/com/google/firebase/installations/internal/FidListener.java
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.installations.internal;
+
+import androidx.annotation.NonNull;
+
+/**
+ * <aside class="warning"><strong>Provides an inter-operational interface only</strong>; instead,
+ * use {@link FidListener} to call a listener when a Fid changes.</aside>
```
vkryachko, line 21: It looks like this javadoc discourages use of this interface and then immediately suggests to use the same interface instead. Can you clarify the intent here?

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -271,6 +275,14 @@ String getName() {
     return Tasks.call(backgroundExecutor, this::deleteFirebaseInstallationId);
   }
 
+  /** Register a callback {@link FidListener} to receive fid changes. */
+  @Override
+  public void registerFidListener(@Nullable FidListener listener) {
```
vkryachko, line 280: Should it instead be?
```suggestion
  public void registerFidListener(@NonNull FidListener listener) {
```

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -271,6 +275,14 @@ String getName() {
...
+    synchronized (this) {
```
vkryachko, line 281: it looks equivalent to just making the whole method `saynchronized`, please consider doing that.

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -271,6 +275,14 @@ String getName() {
...
+      this.fidListener = listener;
```
vkryachko, line 282: Looks like you're only supporting one listener at a time, I wonder if we want to allow multiple listeners, otherwise it may be surprising to clients that their listener is removed just because someone else registered a listener as well.

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -51,4 +52,12 @@
    */
   @NonNull
   Task<Void> delete();
+
+  /**
+   * Register a listener to receive fid changes.
```
vkryachko, line 57: Should there be a way to "unregister" the listener? A common way to do it is to have the `register` method return a "registration" object, that has a `remove()` method.

ankitaj224: Done. PTAL if the changes look.

firebase-installations-interop/firebase-installations-interop.gradle
```diff
@@ -40,7 +40,8 @@ android {
 
 dependencies {
 	implementation 'com.google.android.gms:play-services-tasks:17.0.0'
+    implementation project(path: ':firebase-annotations')
```
vkryachko, line 43: nit: please fix indentation in this file.

firebase-installations-interop/src/main/java/com/google/firebase/installations/internal/HandleFidListener.java
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+package com.google.firebase.installations.internal;
+
+/**
+ * Interface for un-registering a previously registered listener.
+ *
+ * @hide
+ */
+public interface HandleFidListener {
```
vkryachko, line 22: nit:
```suggestion
public interface FidListenerHandle {
```

firebase-installations-interop/src/main/java/com/google/firebase/installations/internal/HandleFidListener.java
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * Unregisters a previously registered {@link FidListener}.
+   *
+   * @hide
+   */
+  void unregister(FidListener fidListener);
```
vkryachko, line 29: The handle is supposed to represent a registration so it shouldn't take a listener as a parameter. For an example see
https://github.com/firebase/firebase-android-sdk/blob/c7d3073d6fe2c168d93acfac88c45ec49a583601/firebase-firestore/src/main/java/com/google/firebase/firestore/ListenerRegistration.java#L17-L25

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -271,6 +278,21 @@ String getName() {
...
+  @Override
+  public synchronized HandleFidListener registerFidListener(@NonNull FidListener listener) {
+    this.fidListeners.add(listener);
+    return new HandleFidListener() {
+      @Override
+      public void unregister(FidListener fidListener) {
+        if (fidListeners.contains(fidListener)) {
+          fidListeners.remove(fidListener);
+        }
+      }
```
vkryachko, line 292: While the registration is synchronized, the unregister is not, please consider something like
```suggestion
      public synchronized void unregister() {
        fidListeners.remove(listener);
      }
```

firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
```diff
@@ -453,6 +454,40 @@ public void testReadToken_withJsonformatting() {
...
+        .thenReturn(
+            TEST_INSTALLATION_RESPONSE
+                .toBuilder()
+                .setUri("/projects/" + TEST_PROJECT_ID + "/installations/" + TEST_FID_2)
+                .setFid(TEST_FID_2)
+                .build());
+
+    FakeFidListener fidListener = new FakeFidListener();
+    // Register the FidListener
+    firebaseInstallations.registerFidListener(fidListener);
```
vkryachko, line 472: Please consider adding a test for removing a registration as well.

firebase-installations-interop/src/main/java/com/google/firebase/installations/internal/HandleFidListener.java
ankitaj224: PTAL. Thanks

firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
ankitaj224: Updated this test to check the removing registration code as well. PTAL. Thanks

vkryachko:
CI failures look legit, see: https://gist.github.com/vkryachko/d1cd35daaf62a63ff7aff9a94e1a28e4 regarding GuardedBy

ankitaj224:
> CI failures look legit, see: https://gist.github.com/vkryachko/d1cd35daaf62a63ff7aff9a94e1a28e4 regarding GuardedBy

Thanks for pointing this out, Vlad. Fixed it. 

Just curious: how did you generate this?

vkryachko:
@ankitaj224 No worries, I did not generate it, it was in the build log of the `check-changed` presubmit:
https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/2195/check-changed/1333493750147584001/

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
```diff
@@ -278,16 +278,22 @@ String getName() {
...
+  public synchronized FidListenerHandle registerFidListener(@NonNull FidListener listener) {
+    fidListeners.add(listener);
+    return new FidListenerHandle() {
       @Override
-      public void unregister(FidListener fidListener) {
-        if (fidListeners.contains(fidListener)) {
-          fidListeners.remove(fidListener);
+      public void unregister() {
+        synchronized (FirebaseInstallations.this) {
+          if (fidListeners.contains(listener)) {
```
vkryachko, line 294: nit: consider removing the `contains` check as remove handles it internally.

vkryachko (approved):

firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java
ankitaj224: Thanks. Done

PR #2200, opened 2020-11-25
Make Api Information task comment the diff on the PR

VinayGuthal:
Tested at https://github.com/firebase/firebase-android-sdk/pull/1831 
apiInformation task would still show failure status since we are doing exit(1) before posting the comment

### DIFF
--- a/ci/fireci/fireci/commands.py
+++ b/ci/fireci/fireci/commands.py
@@ -72,12 +72,7 @@ def smoke_tests(app_build_variant, test_apps_dir):
 def api_information(auth_token, repo_name, issue_number):
   """Comments the api information on the pr"""
 
-  try:
-  	gradle.run('apiInformation')
-  except:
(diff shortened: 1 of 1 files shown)


ci/fireci/fireci/commands.py
```diff
@@ -72,7 +72,13 @@ def smoke_tests(app_build_variant, test_apps_dir):
 def api_information(auth_token, repo_name, issue_number):
   """Comments the api information on the pr"""
 
-  gradle.run('apiInformation')
```
vkryachko, line 75: You can instead call
```python
gradle.run('apiInformation', '--continue', check=False)
```

Also note the continue flag, it will make sure apiInformation runs for all sdks instead of stopping upon first failure.

vkryachko (approved):
Please update the PR title to better reflect its intent. LGTM otherwise

PR #2207, opened 2020-11-27
Update protobuf dep to 3.14

rlazo:


vkryachko:
I wonder if the smoke tests failure is legit?

rlazo:
> I wonder if the smoke tests failure is legit?

They are all failures in `firebase-perf`, and in every case it's because the perf artifact was compiled depending on 3.11. They are expected.

vkryachko (approved):

PR #2216, opened 2020-12-01
Don't crash SDK is socket.close() fails

schmidt-sebastian:
On some Pixel devices, `socket.close()` throws an NPE (might be related to https://github.com/google/conscrypt/issues/331).

Fixes b/164528104

### DIFF
--- a/firebase-database/src/main/java/com/google/firebase/database/tubesock/WebSocket.java
+++ b/firebase-database/src/main/java/com/google/firebase/database/tubesock/WebSocket.java
@@ -261,8 +261,8 @@ public class WebSocket {
     if (socket != null) {
       try {
         socket.close();
-      } catch (IOException e) {
-        throw new RuntimeException(e);
+      } catch (Throwable e) {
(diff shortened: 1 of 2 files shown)


firebase-database/src/main/java/com/google/firebase/database/tubesock/WebSocket.java
```diff
@@ -261,8 +261,8 @@ private synchronized void closeSocket() {
     if (socket != null) {
       try {
         socket.close();
-      } catch (IOException e) {
-        throw new RuntimeException(e);
+      } catch (Throwable e) {
```
vkryachko, line 264: catching `Throwable` seems a little bit aggressive as it will also catch `java.lang.Error`s which, according to its javadoc, "a reasonable application should not try to catch". Would it be sufficient to only catch `Exception` for example?

schmidt-sebastian: Fixed. I only mean to catch IOException and NPEs here.

vkryachko (approved):

Issue #2186, opened 2020-11-19
Firebase remoteConfig return no value


KevynBct:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version:  4.1.1
  * Firebase Component:  firebase-config/firebase-config-ktx
  * Component version: 20.0.1

Hello everyone,

Since I've updated the component `firebase-config` from 20.0.0 to 20.0.1, I can't get values from `Firebase.remoteConfig`.

Nothing has changed in my code since previous version :

First I fetch and active in my `MainActivity`

    val configSettings = remoteConfigSettings {
        minimumFetchIntervalInSeconds = 3600
    }

    Firebase.remoteConfig.setConfigSettingsAsync(configSettings)

    Firebase.remoteConfig.fetchAndActivate().addOnCompleteListener(this) { task ->

        if (task.isComplete) {
            // Some code here
        }

    }

And then in a fragment I get values :

    with (Firebase.remoteConfig) {

        val a = getValue("key1").asString()
        val b = getValue("key2").asString()

    }

T1mmyK0:
I have the same issue.
When updating the app to the new RemoteConfig version, no values are received anymore and the default values apply.
Reinstalling fixes the issue, but I cannot ask all my users to reinstall the app or clear data.

AgiMaulana:
I've encountered the same issue. The value has been synced in previous remote config SDK version when it updated with the version of 20.0.1 the data was lost. Although can be resolved simply by staling the remote config this is not the convenient way I think.

AgiMaulana:
This is reproducible, try to build with the version 19.2.0 everything is okay, then rebuild with the version 20.0.1 remote config return no value for the given key

T1mmyK0:
@AgiMaulana Same here. And when downgrading to 20.0.0 it works fine again. I have downgraded RemoteConfig in my production apk again.

danasilver:
Looking into this. I'm unable to reproduce it in a new Kotlin app with `firebase-config-ktx:20.0.1` and `firebase-analytics-ktx:18.0.0` as dependencies. @vic-flair do you mind taking a look, too?

bocops:
@danasilver I'm seeing the same as others have reported, with a `firebase-config:20.0.x` dependency and a `firebase-analytics:18.0.0` dependency (transitively via `firebase-core`):

1. Fresh install (via Android Studio, "Debug app") of app with a `20.0.0` dependency.
[...]

PR #2217, opened 2020-12-01
Check for missing metadata when constructing ConfigContainer.

danasilver:


erikeldridge (approved):

rlazo (approved):

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
```diff
@@ -86,11 +86,18 @@ private ConfigContainer(
    * <p>The {@code containerJson} must not be modified.
    */
   static ConfigContainer copyOf(JSONObject containerJson) throws JSONException {
+    // Personalization metadata may not have been written yet.
+    JSONObject personalizationMetadataJSON =
+        containerJson.optJSONObject(PERSONALIZATION_METADATA_KEY);
```
erikeldridge, line 91: Thinking out loud: `getJSONObject` throws if the given key isn't found, which would result in the new `ConfigContainer` not being created, as noted in https://github.com/firebase/firebase-android-sdk/issues/2186#issuecomment-736355842 Metadata depends on integrations with RC and may not be defined, ie it's not necessarily an error case, and may be a common case (eg for a project that never uses metadata). Using `optJSONObject` instead facilitates checking for null and assigning a default value rather than throwing.

PR #2222, opened 2020-12-02
Standardize support for Firebase products that integrate with Remote Config.

vic-flair:


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -16,22 +16,41 @@
...
+  public static final String ARM_VALUE = "arm_value";
+  public static final String PERSONALIZATION_ID = "personalization_id";
+  public static final String ARM_INDEX = "arm_index";
+  public static final String GROUP = "group";
+
+  public static final String ANALYTICS_PULL_EVENT_INTERNAL = "_fpc";
+  public static final String CHOICE_ID = "choiceId";
+  public static final String CHOICE_ID_KEY = "_fpid";
+
+  private static final Converter<String, String> CONVERTER =
```
danasilver, line 41: Do we need to use a `Converter` for this? Seems like we're just converting constants and it adds the guava dependency.

vic-flair: Removed the `Converter`. Is Guava a heavy dependency?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,21 +17,37 @@
...
   /** The app's Firebase Analytics client. */
   private final AnalyticsConnector analyticsConnector;
 
+  /** A map of Remote Config parameter key to Personalization ID. */
+  private final Map<String, String> armsCache;
+
   /** Creates an instance of {@code Personalization}. */
   public Personalization(@NonNull AnalyticsConnector analyticsConnector) {
     this.analyticsConnector = analyticsConnector;
+    armsCache = Collections.synchronizedMap(new HashMap<String, String>());
```
danasilver, line 50: I think this can be initialized with the variable declaration rather than in the constructor.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,21 +17,37 @@
...
+
+  public static final String ANALYTICS_PULL_EVENT_INTERNAL = "_fpc";
+  public static final String CHOICE_ID = "choiceId";
+  public static final String CHOICE_ID_KEY = "_fpid";
 
   /** The app's Firebase Analytics client. */
   private final AnalyticsConnector analyticsConnector;
 
+  /** A map of Remote Config parameter key to Personalization ID. */
+  private final Map<String, String> armsCache;
```
danasilver, line 45: Looks like this is used to prevent logging the same key and personalization id combination twice? Do we expect the personalization id to change for a particular key? Maybe we can rename the variable to reflect the usage.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,21 +17,37 @@
...
 
 public class Personalization {
   public static final String ANALYTICS_ORIGIN_PERSONALIZATION = "fp";
-  public static final String ANALYTICS_PULL_EVENT = "_fpc";
-  public static final String ARM_KEY = "_fpid";
-  public static final String ARM_VALUE = "_fpct";
-  static final String PERSONALIZATION_ID = "personalizationId";
+
+  public static final String ANALYTICS_PULL_EVENT = "personalization_assignment";
+  public static final String ARM_KEY = "arm_key";
```
danasilver, line 29: Would consider adding a comment above the block saying what the "key" and non-key named variables are used for.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -74,8 +82,7 @@ public void setUp() {
 
     doAnswer(invocation -> FAKE_LOGS.add(invocation.getArgument(2)))
         .when(mockAnalyticsConnector)
-        .logEvent(
-            eq(ANALYTICS_ORIGIN_PERSONALIZATION), eq(ANALYTICS_PULL_EVENT), any(Bundle.class));
+        .logEvent(eq(ANALYTICS_ORIGIN_PERSONALIZATION), anyString(), any(Bundle.class));
```
danasilver, line 85: Could us the `or` operator from [`AdditionalMatchers`](https://www.javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/AdditionalMatchers.html) to get either of `ANALYTICS_PULL_EVENT` or `ANALYTICS_PULL_EVENT_INTERNAL`.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -99,12 +105,16 @@ public void logArmActive_singlePersonalizationKey_loggedOnce() {
     verify(mockAnalyticsConnector, times(1))
         .logEvent(
             eq(ANALYTICS_ORIGIN_PERSONALIZATION), eq(ANALYTICS_PULL_EVENT), any(Bundle.class));
-    assertThat(FAKE_LOGS).hasSize(1);
+    verify(mockAnalyticsConnector, times(1))
```
danasilver, line 108: Do we want to verify the other (internal) log event here, too?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -58,9 +74,29 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
+    }
+
     Bundle params = new Bundle();
-    params.putString(ARM_KEY, metadata.optString(PERSONALIZATION_ID));
+    params.putString(ARM_KEY, key);
     params.putString(ARM_VALUE, values.optString(key));
+    params.putString(PERSONALIZATION_ID_KEY, personalizationId);
+    params.putInt(ARM_INDEX_KEY, metadata.optInt(ARM_INDEX, -1));
+    params.putString(GROUP, metadata.optString(GROUP));
     analyticsConnector.logEvent(ANALYTICS_ORIGIN_PERSONALIZATION, ANALYTICS_PULL_EVENT, params);
```
danasilver, line 95: Can these log statements fail? Are we okay with one succeeding if the other doesn't?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -58,9 +74,29 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
+    }
+
+    synchronized (armsCache) {
+      if (armsCache.get(key) == personalizationId) {
+        return;
+      }
+      armsCache.put(key, personalizationId);
+    }
+
     Bundle params = new Bundle();
```
danasilver, line 89: Might call these `logParams` and `internalLogParams` since "parameters" otherwise refers to RC parameters.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -115,16 +125,20 @@ public void logArmActive_multiplePersonalizationKeys_loggedMultiple() {
...
-    params1.putString(ARM_VALUE, "value1");
-    assertThat(FAKE_LOGS.get(0).toString()).isEqualTo(params1.toString());
-
+    params1.putString(CHOICE_ID_KEY, "id1");
     Bundle params2 = new Bundle();
-    params2.putString(ARM_KEY, "id2");
-    params2.putString(ARM_VALUE, "value2");
-    assertThat(FAKE_LOGS.get(1).toString()).isEqualTo(params2.toString());
+    params2.putString(CHOICE_ID_KEY, "id2");
+    assertThat(FAKE_LOGS)
```
danasilver, line 139: Same question here - should we test the internal event?

danasilver (changes requested):
Some questions and comments about usage. We should make sure we test the upgrade path between SDK versions, too.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: Oh, that's actually supposed to be choice ID. I think I forgot to change it when we were discussing what fields should be called. We do expect choice IDs to change, and we need to log each at least once. There will probably some de-duping on the back-end too, but this is just a quick way to reduce some of that work.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: I think it should be OK. As long as it's not happening too frequently, the stats should be close enough.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,18 +17,36 @@
...
   public static final String ANALYTICS_ORIGIN_PERSONALIZATION = "fp";
-  public static final String ANALYTICS_PULL_EVENT = "_fpc";
-  public static final String ARM_KEY = "_fpid";
-  public static final String ARM_VALUE = "_fpct";
-  static final String PERSONALIZATION_ID = "personalizationId";
+
+  // Constants with LOG_KEY suffix are how the corresponding ones without it are identified on
+  // Google Analytics.
+  public static final String ANALYTICS_PULL_EVENT = "personalization_assignment";
+  public static final String ARM_KEY = "arm_key";
```
danasilver, line 31: `ARM_LOG_KEY`?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,18 +17,36 @@
...
+
+  public static final String ANALYTICS_PULL_EVENT_INTERNAL = "_fpc";
+  public static final String CHOICE_ID = "choiceId";
+  public static final String CHOICE_ID_LOG_KEY = "_fpid";
 
   /** The app's Firebase Analytics client. */
   private final AnalyticsConnector analyticsConnector;
 
+  /** A map of Remote Config parameter key to choice ID. */
+  private final Map<String, String> armsCache =
```
danasilver, line 47: I'd still like to rename this to something more descriptive. Maybe `loggedArms` or `loggedChoiceIds`?

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -88,43 +103,79 @@ public void logArmActive_nonPersonalizationKey_notLogged() {
...
-    params.putString(ARM_KEY, "id1");
-    params.putString(ARM_VALUE, "value1");
-    assertThat(FAKE_LOGS.get(0).toString()).isEqualTo(params.toString());
+    Bundle logParams = new Bundle();
+    logParams.putString(ARM_KEY, "key1");
+    logParams.putString(ARM_VALUE, "value1");
+    logParams.putString(PERSONALIZATION_ID_LOG_KEY, "p13n1");
+    logParams.putInt(ARM_INDEX_LOG_KEY, 0);
+    logParams.putString(GROUP, "BASELINE");
+    Bundle internalLogParams = new Bundle();
```
danasilver, line 132: Would separate these from the `logParams` with an empty line for readability.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -88,43 +103,79 @@ public void logArmActive_nonPersonalizationKey_notLogged() {
...
+            any(Bundle.class));
+    assertThat(FAKE_LOGS).hasSize(4);
+
+    Bundle logParams1 = new Bundle();
+    logParams1.putString(ARM_KEY, "key1");
+    logParams1.putString(ARM_VALUE, "value1");
+    logParams1.putString(PERSONALIZATION_ID_LOG_KEY, "p13n1");
+    logParams1.putInt(ARM_INDEX_LOG_KEY, 0);
+    logParams1.putString(GROUP, "BASELINE");
+    Bundle logParams2 = new Bundle();
```
danasilver, line 161: Same here - newline would help readability.

danasilver (changes requested):
A couple naming and formatting followups.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -58,9 +76,29 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
-    Bundle params = new Bundle();
-    params.putString(ARM_KEY, metadata.optString(PERSONALIZATION_ID));
-    params.putString(ARM_VALUE, values.optString(key));
-    analyticsConnector.logEvent(ANALYTICS_ORIGIN_PERSONALIZATION, ANALYTICS_PULL_EVENT, params);
+    String choiceId = metadata.optString(CHOICE_ID);
+    if (choiceId.isEmpty()) {
+      return;
+    }
+
+    synchronized (armsCache) {
```
erikeldridge, line 84: Mind adding a comment explaining why we need this? 

Thinking: this ensures we only log one event per param-choice pair. This map is bounded by the number of params. Choices are made on each fetch, so this isn't a long-lived cache.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -58,9 +76,29 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
+      if (choiceId.equals(armsCache.get(key))) {
+        return;
+      }
+      armsCache.put(key, choiceId);
+    }
+
+    Bundle logParams = new Bundle();
+    logParams.putString(ARM_KEY, key);
```
erikeldridge, line 92: Minor: there are a lot of things called "key" in this file. Since the `key` arg passed to the `logArmActive` fn is actually the RC param, renaming it to something like `rcParameter` would be more self-descriptive.

erikeldridge, line 92: Thinking: the [design](go/sage-p13n-bq) lists four things we want to log:
1. Developer friendly identifier of the personalization (either the rc param, or an id they can map back to rc param), satisfied by the `arm_key` field
2. The choice that was made (a string value), satisfied by the `arm_value` field
3. The group the user was in (“p13n” or “baseline”), satisfied by the `group` field
4. Personalization ID, satisfied by the `personalization_id` field

Minor: renaming `arm_key` to something like `rc_parameter` would be more self-descriptive given it communicates an RC parameter.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,18 +17,36 @@
...
 public class Personalization {
   public static final String ANALYTICS_ORIGIN_PERSONALIZATION = "fp";
-  public static final String ANALYTICS_PULL_EVENT = "_fpc";
-  public static final String ARM_KEY = "_fpid";
-  public static final String ARM_VALUE = "_fpct";
-  static final String PERSONALIZATION_ID = "personalizationId";
+
+  // Constants with LOG_KEY suffix are how the corresponding ones without it are identified on
+  // Google Analytics.
+  public static final String ANALYTICS_PULL_EVENT = "personalization_assignment";
```
erikeldridge, line 30: Minor: it's unclear to me what "pull" means here and below. Also, the `ANALYTICS_..._INTERNAL` naming is helpful below, but we only use it in one place. Given we have external and internal events, and Analytics differentiates "event" from "param" field types (per the [events proposal](go/p13n-events)), something like  `ANALYTICS_EXTERNAL_EVENT` would be more self-descriptive here. In general, I'd vote to use  `ANALYTICS_EXTERNAL_` and `ANALYTICS_INTERNAL_` prefixes to consistently differentiate field names for the external and internal events, and `..._EVENT` and `..._PARAM` to differentiate event names from param names using Analytics' nomenclature.

Thinking: the external event name is arbitrary, we just need to document what it is. `personalization_assignment` seems self-descriptive to me 👍

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -58,9 +76,29 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
+    Bundle logParams = new Bundle();
+    logParams.putString(ARM_KEY, key);
+    logParams.putString(ARM_VALUE, values.optString(key));
+    logParams.putString(PERSONALIZATION_ID_LOG_KEY, metadata.optString(PERSONALIZATION_ID));
+    logParams.putInt(ARM_INDEX_LOG_KEY, metadata.optInt(ARM_INDEX, -1));
+    logParams.putString(GROUP, metadata.optString(GROUP));
+    analyticsConnector.logEvent(ANALYTICS_ORIGIN_PERSONALIZATION, ANALYTICS_PULL_EVENT, logParams);
+
+    Bundle internalLogParams = new Bundle();
+    internalLogParams.putString(CHOICE_ID_LOG_KEY, choiceId);
```
erikeldridge, line 100: Thinking: the [design](go/p13n-choice-metadata) lists one thing we want to log:
1. Choice ID, satisfied by the `_fpid` param

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
erikeldridge: 👍  I'd vote to use the `_loggedInternallyAndExternally` suffix for the test below too, ie `logArmActive_multiplePersonalizationKeys_loggedMultiple` --> `logArmActive_multiplePersonalizationKeys_loggedInternallyAndExternally`

erikeldridge (commented):
Seems technically correct to me 👍  I dropped a couple comments re event field naming to improve readability, but since they concern naming, which can be contentious, I flagged them as minor. To avoid miscommunication, I'll defer formal approval to @danasilver, since he's the Android lead.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
erikeldridge: 👍  I agree that'd be helpful

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -58,9 +76,29 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
+      }
+      armsCache.put(key, choiceId);
+    }
+
+    Bundle logParams = new Bundle();
+    logParams.putString(ARM_KEY, key);
+    logParams.putString(ARM_VALUE, values.optString(key));
+    logParams.putString(PERSONALIZATION_ID_LOG_KEY, metadata.optString(PERSONALIZATION_ID));
+    logParams.putInt(ARM_INDEX_LOG_KEY, metadata.optInt(ARM_INDEX, -1));
+    logParams.putString(GROUP, metadata.optString(GROUP));
```
andyliangdong, line 96: I'm not sure where the metadata come from? is this logging server side's decision?

andyliangdong (approved):
I'm not quite familiar with SDK dev process, thus I'm approving from language point of view.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: Changed it to `EXTERNAL_RC_PARAMETER_PARAM` per Erik's comment.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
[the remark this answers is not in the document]
vic-flair: I don't know of any reason why they wouldn't be OK.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
vic-flair: Yeah, from the server.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -53,14 +72,36 @@ public void logArmActive(@NonNull String key, @NonNull ConfigContainer configCon
...
+    // logging.
+    synchronized (loggedChoiceIds) {
+      if (choiceId.equals(loggedChoiceIds.get(rcParameter))) {
+        return;
+      }
+      loggedChoiceIds.put(rcParameter, choiceId);
+    }
+
+    Bundle logParams = new Bundle();
+    logParams.putString(EXTERNAL_RC_PARAMETER_PARAM, rcParameter);
```
erikeldridge, line 95: This looks great. Much easier to read 👍

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
```diff
@@ -88,43 +123,46 @@ public void logArmActive_nonPersonalizationKey_notLogged() {
...
+        .logEvent(eq(ANALYTICS_ORIGIN_PERSONALIZATION), eq(INTERNAL_EVENT), any(Bundle.class));
+    assertThat(FAKE_LOGS).hasSize(4);
+
+    assertThat(FAKE_LOGS)
+        .comparingElementsUsing(TO_STRING)
+        .containsAtLeast(LOG_PARAMS_1.toString(), LOG_PARAMS_2.toString())
+        .inOrder();
+    assertThat(FAKE_LOGS)
+        .comparingElementsUsing(TO_STRING)
+        .containsAtLeast(INTERNAL_LOG_PARAMS_1.toString(), INTERNAL_LOG_PARAMS_2.toString())
```
danasilver, line 165: Seems like this could be combined with the above into one statement with all four logs?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/Personalization.java
```diff
@@ -17,18 +17,37 @@
...
+  public static final String EXTERNAL_GROUP_PARAM = "group";
+
+  public static final String INTERNAL_EVENT = "_fpc";
+  public static final String CHOICE_ID = "choiceId";
+  public static final String INTERNAL_CHOICE_ID_PARAM = "_fpid";
 
   /** The app's Firebase Analytics client. */
   private final AnalyticsConnector analyticsConnector;
 
+  /** A map of Remote Config parameter key to choice ID. */
```
danasilver, line 47: Related to removing the comment below about reducing logging - I'd just note that here as something like "Remote Config parameter key and Choice ID pairs that have already been logged to Analytics."

danasilver (approved):

danasilver:
@vic-flair Would also be good to add a test that logging a duplicate choice ID does not log to GA.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/PersonalizationTest.java
vic-flair: We can't combine it if we want to check the ordering too, since it's not guaranteed that the external ones come before the internal ones.

PR #2223, opened 2020-12-02
Enabling Test Orchestrator 

ramanpreetSinghKhinda:
b/174660554

Enabling Test Orchestrator to run each of app's instrumentation tests independently without accumulating shared states.

References:
  - https://firebase.google.com/docs/test-lab/android/firebase-console#new_instrumentation_tests_with_orchestrator
 - http://cloud/sdk/gcloud/reference/firebase/test/android/run#--use-orchestrator

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.java
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/device/FirebaseTestServer.java
@@ -62,6 +62,7 @@ public class FirebaseTestServer extends TestServer {
         "--type=instrumentation",
         "--app=" + testedApkPath,
         "--test=" + testApk,
+        "--use-orchestrator",
         "--no-auto-google-login",
         "--no-record-video",
         "--no-performance-metrics",


vkryachko:
@ramanpreetSinghKhinda while you're at it, do you mind also increasing the timeout to 45m? This should help with your e2e test failures we are seeing.
for reference: https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--timeout

ramanpreetSinghKhinda:
> @ramanpreetSinghKhinda while you're at it, do you mind also increasing the timeout to 45m? This should help with your e2e test failures we are seeing.
> for reference: https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--timeout

I think the current timeout is `15 min`. Our tests usually finish within `9 min` and the timeout in b/174660554 is not actually  because our test is taking longer. It's because the android test runner is just waiting and not starting the test. If we increase the timeout it will just hold and waste resources for that timeout period.

vkryachko (approved):

rlazo (approved):

ramanpreetSinghKhinda:
I see past many runs of smoke tests are failing. Is this a known issue @yifanyang?

/test smoke-tests

ramanpreetSinghKhinda:
Per discussion with @yifanyang 

It's safe to merge the PR. The smoke test has been failing since last weekend and Yifan is working on it.

PR #2230, opened 2020-12-04
Ml logging

annzimmer:


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseMlException.java
```diff
@@ -0,0 +1,159 @@
...
+import com.google.firebase.FirebaseException;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * Represents an Exception resulting from an operation on a {@link FirebaseModelDownloader}. Error
+ * mappings should remain consistent with the original firebase_ml_sdk whenever possible.
+ */
+public class FirebaseMlException extends FirebaseException {
+  /** The operation completed successfully. FirebaseMLException will never have a status of OK. */
```
vkryachko, line 30: If it will never have this status, why include it as a possibility in the type?
i.e. it seems that by doing it you're making this a runtime error as opposed to a compile-time error.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/DataTransportMlStatsSender.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class DataTransportMlStatsSender {
+  private static final String FIREBASE_ML_STATS_NAME = "FIREBASE_ML_LOG_SDK";
+  private final Transport<FirebaseMlStat> transport;
+
+  @NonNull
+  public static DataTransportMlStatsSender create(@NonNull Context context) {
+    TransportRuntime.initialize(context);
+    final Transport<FirebaseMlStat> transport =
+        TransportRuntime.getInstance()
```
vkryachko, line 40: You should not be calling into the runtime, instead you should inject the `TransportFactory` instance through components(which I think you already did above), why does it need to be done here as well?

firebase-ml-modeldownloader/firebase-ml-modeldownloader.gradle
```diff
@@ -43,9 +43,15 @@ android {
...
 dependencies {
+    implementation project(':encoders:firebase-encoders')
+    implementation project(':encoders:firebase-encoders-json')
     implementation project(':firebase-common')
     implementation project(':firebase-components')
+    implementation project(':firebase-datatransport')
     implementation project(':firebase-installations-interop')
+    implementation project(':transport:transport-api')
+    implementation project(':transport:transport-backend-cct')
+    implementation project(':transport:transport-runtime')
```
vkryachko, line 54: There is no need to declare these deps, please only depend on `transport-api` and `firebase-datatransport`

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseMlException.java
annzimmer: I removed for now - I think it was meant as a default in the old code but I can just not allow that here.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/DataTransportMlStatsSender.java
annzimmer: Updated - hopefully correctly. Still crashes in my test app, but I might not be adding it right. Let me know if you have time to help.

firebase-ml-modeldownloader/firebase-ml-modeldownloader.gradle
annzimmer: Done

firebase-ml-modeldownloader/firebase-ml-modeldownloader.gradle
```diff
@@ -43,25 +60,43 @@ android {
...
+    implementation project(':firebase-datatransport')
     implementation project(':firebase-installations-interop')
+    implementation project(':transport:transport-api')
     runtimeOnly project(':firebase-installations')
 
+    implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'com.google.android.gms:play-services-tasks:17.2.0'
     implementation 'com.google.auto.service:auto-service-annotations:1.0-rc6'
     implementation 'javax.inject:javax.inject:1'
+    compile 'com.google.protobuf:protobuf-lite:3.0.0'
```
vkryachko, line 76: this is the wrong protolite runtime, you need to use protobuf-javalite(see. Also note that you should add it under `testImplementation` not `compile` or `implementation`, as doing so will make this a dependency of your SDK, not just your tests.

See example here https://github.com/firebase/firebase-android-sdk/blob/f4d1987fdad3932f7fa5a9fb7c61a24b725e6f62/firebase-crashlytics-ndk/firebase-crashlytics-ndk.gradle#L128

PR #2236, opened 2020-12-07
Delete model

annzimmer:


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
```diff
@@ -187,7 +192,16 @@ public static FirebaseModelDownloader getInstance(@NonNull FirebaseApp app) {
    */
   @NonNull
   public Task<Void> deleteDownloadedModel(@NonNull String modelName) {
-    throw new UnsupportedOperationException("Not yet implemented.");
+
+    TaskCompletionSource<Void> taskCompletionSource = new TaskCompletionSource<>();
+    executor.execute(
+        () -> {
+          // remove all files associated with this model and then clean up model references.
+          fileManager.deleteAllModels(modelName);
```
rlazo, line 200: In case the app crashes, or the phone dies, after models have been deleted but before their details have been removed from sharedPreferences, what would happen?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
```diff
@@ -118,6 +126,21 @@ private int getLatestCachedModelVersion(@NonNull File modelDir) {
   @Nullable
   File getModelFileDestination(@NonNull CustomModel model) throws Exception {
     File destFolder = getDirImpl(model.getName());
+    if (!destFolder.exists()) {
+      Log.d(TAG, "model folder does not exist, creating one: " + destFolder.getAbsolutePath());
+      if (!destFolder.mkdirs()) {
+        System.out.println("Failed to create model folder" + destFolder);
+        //        throw new FirebaseMLException(
```
rlazo, line 133: Intentional?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
annzimmer: The model would be listed but when they tried to get the file it would return null... I think this is acceptable. I did update the getFile to handle this case a little more cleanly. If you can think of a better solution. I'm open to suggestions. I've also made a bug b/175144342 to add a start-up routine that will check that models with filePaths actually point to files and will clean them up if they don't.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
annzimmer: yes - the logging which will be the commented out lines is currently being tested. Almost there.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
rlazo: LGTM

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
rlazo: Sure. Ping me when they're done so I can do a last review before granting approval. Thanks!

rlazo (approved):

PR #2277, opened 2020-12-21
Implement `startAfter` and `endBefore` for RTDB queries

IanWyszynski:
[go/rtdb-exclusive-query-filters](go/rtdb-exclusive-query-filters)

`startAfter(value, key)` means:

If `key == undefined or null`, `startAt(value, MAX_NAME)`, else, `startAt(value, key + 1)`, where `+ 1` means either the next integer (if the key is parseable as an int), or the lexicographically next key.


firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -1427,6 +1784,25 @@ public void nullPrioritiesIncludedInStartAt2() throws DatabaseException, Interru
     DeepEquals.assertEquals(expected, result);
   }
 
+  @Test
+  public void nullPrioritiesIncludedInStartAfter2() throws DatabaseException, InterruptedException {
```
schmidt-sebastian, line 1788: Do you know what the `2` is meant to indicate? If not, I suggest we remove it from here and above.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -2079,6 +2455,43 @@ public void startAtWithTwoArgumentsWorksCase1169()
     assertEquals(2, i);
   }
 
+  @Test
+  public void startAfterWithTwoArgumentsWorksCase1169()
```
schmidt-sebastian, line 2459: The case number looks like it refers to an old bug about startAt, which is not applicable here. I would drop it (at least from here)

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -2920,6 +3333,94 @@ public void onCancelled(DatabaseError error) {}
     assertEquals("a", removedSecond.get(0));
   }
 
+  // TODO(wyszynski): endBefore
+  @Test
+  public void case2003CorrectlyGetEventsForStartAfterEndAtQueriesWhenPriorityChanges()
```
schmidt-sebastian, line 3338: Same comment about case number.

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
       });
   }*/
 
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default.
+   *
+   * @param value The value to start at, inclusive
```
schmidt-sebastian, line 286: s/inclusive/exclusive (and maybe s/at/after, but then I am no longer sure whether this is inclusive or exclusive)

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+  @NonNull
+  public Query startAfter(@Nullable String value) {
+    return startAt(value, ChildKey.getMaxName().asString());
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default.
+   *
+   * @param value The value to start at, inclusive
```
schmidt-sebastian, line 298: Same as above

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+  @NonNull
+  public Query startAfter(double value) {
+    return startAt(value, ChildKey.getMaxName().asString());
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default.
+   *
+   * @param value The value to start at, inclusive
```
schmidt-sebastian, line 310: Same as above

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+   * @return A Query with the new constraint
+   * @since 2.0
```
schmidt-sebastian, line 312: We are well past version 2 now :)

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+  public Query startAfter(boolean value) {
+    return startAt(value, ChildKey.getMaxName().asString());
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default, and additionally only child
+   * nodes with a key greater than or equal to the given key.
+   *
+   * @param value The priority to start at, inclusive
```
schmidt-sebastian, line 324: Same as above

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+        value != null ? new StringNode(value, PriorityUtilities.NullPriority()) : EmptyNode.Empty();
+    return startAfter(node, key);
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default, and additionally only child
+   * nodes with a key greater than or equal to the given key.
+   *
+   * @param value The priority to start at, inclusive
```
schmidt-sebastian, line 340: Same as above

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+  public Query startAfter(double value, @Nullable String key) {
+    return startAfter(new DoubleNode(value, PriorityUtilities.NullPriority()), key);
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default, and additionally only child
+   * nodes with a key greater than or equal to the given key.
+   *
+   * @param value The priority to start at, inclusive
```
schmidt-sebastian, line 354: Same as above

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,92 @@ public void run() {
...
+   * @param key The key to start at, inclusive
+   * @return A Query with the new constraint
+   * @since 2.0
```
schmidt-sebastian, line 357: Same as above

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -1112,6 +1112,8 @@ private void putInternal(
     long writeId = this.writeCounter++;
 
     outstandingPuts.put(writeId, new OutstandingPut(action, request, onComplete));
+    logger.info("Can send writes = " + canSendWrites());
+    logger.info("Trying to send puts = " + outstandingPuts.get(writeId).getRequest());
```
schmidt-sebastian, line 1116: This seems like leftover debug information.

IanWyszynski:
@schmidt-sebastian This is ready for another round of review.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3054,6 +3661,45 @@ public void onCancelled(DatabaseError error) {}
     IntegrationTestHelpers.waitFor(done);
   }
 
+  @Test
+  public void integerKeysBehaveNumericallyStartAfter1()
```
schmidt-sebastian, line 3665: Another strange suffix. If these are used to distinguish test cases, then we should come up with more specific names.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3142,6 +3788,46 @@ public void onCancelled(DatabaseError error) {}
     IntegrationTestHelpers.waitFor(done);
   }
 
+  @Test
+  public void integerKeysBehaveNumerically3StartAfter()
```
schmidt-sebastian, line 3792: Please come up with a more specific test name.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -3522,28 +4208,28 @@ public void onCancelled(@NonNull DatabaseError error) {}
...
-
-    assertNull(new WriteFuture(writer, 42L).timedGet());
-    assertEquals(42L, Tasks.await(reader.get()).getValue());
-
-    readerDb.goOffline();
-
-    Semaphore semaphore = new Semaphore(0);
-    assertNotNull(ReadFuture.untilEquals(reader, 42L).timedGet());
-  }
+  //  @Test
```
schmidt-sebastian, line 4211: Please fix.

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,96 @@ public void run() {
...
+  public Query startAfter(boolean value) {
+    return startAt(value, ChildKey.getMaxName().asString());
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default, and additionally only child
+   * nodes with a key greater than or equal to the given key.
+   *
+   * @param value The priority to start at
```
schmidt-sebastian, line 326: s/priority/value

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,96 @@ public void run() {
...
+   */
+  @NonNull
+  public Query startAfter(@Nullable String value, @Nullable String key) {
+    Node node =
+        value != null ? new StringNode(value, PriorityUtilities.NullPriority()) : EmptyNode.Empty();
+    return startAfter(node, key);
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
```
schmidt-sebastian, line 339: The first greater should be "equal", no?

firebase-database/src/main/java/com/google/firebase/database/Query.java
```diff
@@ -278,6 +279,96 @@ public void run() {
...
+  public Query startAfter(double value, @Nullable String key) {
+    return startAfter(new DoubleNode(value, PriorityUtilities.NullPriority()), key);
+  }
+
+  /**
+   * Create a query constrained to only return child nodes with a value greater than the given
+   * value, using the given orderBy directive or priority as default, and additionally only child
+   * nodes with a key greater than or equal to the given key.
+   *
+   * @param value The priority to start at
```
schmidt-sebastian, line 358: s/priority/value

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
```diff
@@ -57,6 +65,38 @@ public static synchronized String generatePushChildName(long now) {
     return result.toString();
   }
 
+  public static final String nextAfter(String key) {
```
schmidt-sebastian, line 68: Nit: This is usually called a successor.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
```diff
@@ -57,6 +65,38 @@ public static synchronized String generatePushChildName(long now) {
...
+    Validation.validateNullableKey(key);
+    Integer num = tryParseInt(key);
+    if (num != null) {
+      return String.valueOf(num + 1);
```
schmidt-sebastian, line 72: This looks like it could trigger an integer overflow. You could probably just use a long here, unless there are some specifications for valid keys that I am not aware of.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
```diff
@@ -15,14 +15,22 @@
...
 import java.util.Random;
 
 public class PushIdGenerator {
 
   private static final String PUSH_CHARS =
       "-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz";
 
+  private static final char MAX_PUSH_CHAR = 'z';
+
+  private static final char MIN_PUSH_CHAR = '-';
```
schmidt-sebastian, line 30: Nit: Flip assignments for clarity. 

You could also refer to `PUSH_CHARS` and get the values from there (which will be slower and more code, so I am not sure it is actually worth it).

firebase-database/src/main/java/com/google/firebase/database/core/utilities/Utilities.java
```diff
@@ -273,4 +273,8 @@ public static boolean equals(@Nullable Object left, @Nullable Object right) {
     }
     return left.equals(right);
   }
+
+  public static String next(@Nullable String value) {
```
schmidt-sebastian, line 277: This is unused.

firebase-database/src/main/java/com/google/firebase/database/snapshot/ChildKey.java
```diff
@@ -56,11 +56,11 @@ public boolean isPriorityChildName() {
     return this.equals(PRIORITY_CHILD_KEY);
   }
 
-  protected boolean isInt() {
+  public boolean isInt() {
```
schmidt-sebastian, line 59: These changes don't seem to be needed.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
IanWyszynski: I sort of prefer having separate sentinel values for this since they trigger edge-cases in the `successor`/`predecessor` code.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
IanWyszynski: Using a long should be fine, the limit on keys is 768 bytes:

https://firebase.google.com/docs/database/usage/limits#data_tree

IanWyszynski: The only other place we're using tryParseInt is in:

https://github.com/firebase/firebase-android-sdk/blob/e65f4876ff61b5791986c249b21cb094c37bd29e/firebase-database/src/main/java/com/google/firebase/database/snapshot/ChildrenNode.java#L134

Where it seems to be used to decide the number of results in an array. I think we should be able to change the method to `tryParseLong` without risk.

IanWyszynski: Upon further consideration, I think we should handle the overflow here. In our public docs, we assume at most 32-bit integer keys for ordering semantics:

https://firebase.google.com/docs/database/web/lists-of-data

I wonder if it's reasonable to return `MIN_PUSH_CHAR` here since that's the next string in our documented key ordering.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -1150,6 +1567,29 @@ public void startAtEndAtWithPriorityAndNameWorks2()
     helper.waitForEvents();
   }
 
+  @Test
+  public void startAfterEndAtWithPriorityAndNameWorks2()
```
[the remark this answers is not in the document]
IanWyszynski: Some of these didn't really add any value and were just blindly copy/pastes. I'm removed them.

IanWyszynski:
@schmidt-sebastian Ready for review.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -658,6 +770,57 @@ public void onCancelled(DatabaseError error) {}
     assertEquals("b", removed.get(0));
   }
 
+  @Test
+  public void setLimitEnsureChildRemovedAndChildAddedHitWhenLimitIsHitFromFrontWithStartAfter()
```
schmidt-sebastian, line 774: Totally optional:
s/WhenLimitIsHitFromFront/WhenDataEntersLimit

Here and below if you do update.

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
```diff
@@ -57,6 +65,49 @@ public static synchronized String generatePushChildName(long now) {
     return result.toString();
   }
 
+  public static final String successor(String key) {
+    Validation.validateNullableKey(key);
+    Integer num = tryParseInt(key);
+    if (num != null) {
+      if (Long.valueOf(num) + 1L > Integer.MAX_VALUE) {
```
schmidt-sebastian, line 72: You don't need to convert `num` to long. In Java, long+int=long (and the int is treated as a long)

firebase-database/src/main/java/com/google/firebase/database/core/utilities/PushIdGenerator.java
```diff
@@ -57,6 +65,49 @@ public static synchronized String generatePushChildName(long now) {
...
+      // If this key doesn't have all possible character slots filled,
+      // the lexicographical successor is the same string with the smallest
+      // possible character appended to the end.
+      next.append(MIN_PUSH_CHAR);
+      return next.toString();
+    }
+
+    int i = next.length() - 1;
+
+    while (i >= 0 && next.charAt(i) == MAX_PUSH_CHAR) {
```
schmidt-sebastian, line 90: Optional nit: Remove empty line above.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -504,6 +585,37 @@ public void setVariousLimitsWithStartAtNameWithServerData()
...
+    // lot of partial data back from the server. This all correct, and we end up in the correct
+    // state, but it's still kinda weird. Consider having ValueExpectationHelper deal with initial
+    // state.
+
+    new WriteFuture(ref, new MapBuilder().put("a", 1).put("b", 2).put("c", 3).put("d", 4).build())
+        .timedGet();
+
+    ValueExpectationHelper expectations = new ValueExpectationHelper();
+
+    expectations.add(ref.startAfter(null).limitToFirst(1), null);
```
schmidt-sebastian, line 604: Let's talk about this offline. My gut feeling is that this should match all results, since they order past `null`.

schmidt-sebastian (commented):
LGTM(ish), depending on how we handle the "startAfter(null)" case.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
IanWyszynski: I added an explanation here: https://github.com/firebase/firebase-ios-sdk/pull/7209/files#r556060181. 

I know this reads funny, but I think it is the intended behavior because we are indexing on priority.

IanWyszynski:
@schmidt-sebastian The copyright check is failing for an unrelated file. The device-check-changed check runs

```
./gradlew :firebase-database:firebase-test-labUpload
```

Which I ran locally on master and it failed in the same way it does here. I'm going to assume these CI checks are not needed.

firebase-database/src/test/java/com/google/firebase/database/PushIdGeneratorTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private static final char MAX_PUSH_CHAR = 'z';
+
+  private static final int MAX_KEY_LEN = 786;
+
+  @Test
+  public void testSuccessorSpecialValue() {
+    assertEquals(
+        PushIdGenerator.successor(String.valueOf(Integer.MAX_VALUE)),
+        Character.toString(MIN_PUSH_CHAR));
```
schmidt-sebastian, line 40: The expected state should come before the actual state. I think you need to flip the arguments everywhere.

schmidt-sebastian (approved):
LGTM. Please add changelog entry to https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/CHANGELOG.md

PR #2296, opened 2021-01-06
Implement LoadBundleTask

schmidt-sebastian:
This implements the logic for LoadBundleTask. It's basically a super advanced copy/paste job:

- It takes some API comments from the Web implementation (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L8374)
- It relies on the Android Task API whenever possible. For progress updates, the implementation is inspired by cs/piper///depot/google3/java/com/google/android/gmscore/integ/client/tasks/src/com/google/android/gms/tasks/TaskImpl.java
- Most Task API comments are copied from StorageTask (https://github.com/firebase/firebase-android-sdk/blob/master/firebase-storage/src/main/java/com/google/firebase/storage/StorageTask.java)

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java
@@ -15,6 +15,7 @@
 package com.google.firebase.firestore;
 
 import android.app.Activity;
+import androidx.annotation.GuardedBy;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.android.gms.tasks.Continuation;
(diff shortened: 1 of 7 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
```diff
@@ -27,39 +38,79 @@
...
+    if (totalDocuments != that.totalDocuments) return false;
+    if (bytesLoaded != that.bytesLoaded) return false;
+    if (totalBytes != that.totalBytes) return false;
+    return taskState == that.taskState;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = documentsLoaded;
+    result = 31 * result + totalDocuments;
```
dconeybe, line 110: Consider using `Objects.hash()` from the standard library instead of implementing your own hash code.

https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
```diff
@@ -27,39 +38,79 @@
   private int totalDocuments;
```
dconeybe, line 38: Should `totalDocuments`, `bytesLoaded`, and `totalBytes` be declared as `final` as well?

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
```diff
@@ -38,4 +58,208 @@ public void testImplementsAllTaskInterface() {
       }
     }
   }
+
+  @Test
+  public void testSuccessListener() throws InterruptedException {
```
dconeybe, line 63: Please add tests to verify that the correct thread is used to perform the callbacks.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java
```diff
@@ -15,183 +15,556 @@
...
+    synchronized (lock) {
+      progressListenerQueue.remove(new ManagedListener(/* executor= */ null, listener));
+    }
+  }
+
+  void setResult(@Nullable LoadBundleTaskProgress result) {
+    synchronized (lock) {
+      snapshot = result;
+      for (ManagedListener progressListener : progressListenerQueue) {
+        progressListener.maybeRun(result);
```
dconeybe, line 495: I think this code will throw ConcurrentModificationException if the callback adds or removes a progress listener. Please add a test for this and verify that this scenario is handled correctly. This comment also applies to `setException()` and `updateProgress()`.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java
```diff
@@ -15,183 +15,556 @@
...
+  private LoadBundleTaskProgress snapshot;
+
+  /**
+   * A TaskCompletionSource that is used to deliver all standard Task API events (e.g.
+   * `onComplete`).
+   */
+  private final TaskCompletionSource<LoadBundleTaskProgress> completionSource;
+
+  /**
+   * A delegate task derived from `completionSource`. All APIs events that don't involve progress
```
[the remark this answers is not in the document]
wilhuff: Also, Javadoc does not do markdown processing, so `` `completionSource` `` should be `{@code completionSource}`.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
schmidt-sebastian: Yeah, good catch.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
```diff
@@ -27,39 +38,79 @@
...
 
+  /** Returns the current state of the task. */
   @NonNull
   public TaskState getTaskState() {
     return taskState;
   }
+
+  /** If the task failed, returns the exception. Otherwise, returns null. */
+  @Nullable
+  public Exception getError() {
```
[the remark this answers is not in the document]
schmidt-sebastian: I was trying to use names in the Task API, but failed miserably :) It's also `Task#getException`

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
```diff
@@ -27,39 +38,79 @@
...
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+
+    LoadBundleTaskProgress that = (LoadBundleTaskProgress) o;
+
+    if (documentsLoaded != that.documentsLoaded) return false;
```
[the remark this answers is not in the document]
schmidt-sebastian: This is auto-generated so I will leave as is.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
```diff
@@ -27,39 +38,79 @@
...
+    if (totalDocuments != that.totalDocuments) return false;
+    if (bytesLoaded != that.bytesLoaded) return false;
+    if (totalBytes != that.totalBytes) return false;
+    return taskState == that.taskState;
```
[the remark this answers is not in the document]
schmidt-sebastian: Oops. Forgot to regenerate after adding the exception (which we also have to add to the API document)

cc @wu-hui

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTaskProgress.java
schmidt-sebastian: It's not available with Android 14 (requires API level 19+).

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
schmidt-sebastian: Done

schmidt-sebastian (commented):
Thanks for the thorough review. Feedback addressed.

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
```diff
@@ -115,6 +123,64 @@ public void testProgressListener() throws InterruptedException {
...
+  }
+
+  @Test
+  public void testProgressListenerFireOnSpecifiedExecutor() throws InterruptedException {
+    CountDownLatch latch = new CountDownLatch(2);
+
+    LoadBundleTask task = new LoadBundleTask();
+    task.addOnProgressListener(
+        p -> {
+          assertNotEquals(TEST_THREAD_NAME, Thread.currentThread().getName());
```
dconeybe, line 147: Can you confirm that this assertion failure will actually cause a test failure? Since the exception will be thrown on some other thread I just don't know if it will cause the test to fail.

dconeybe (approved):
LGTM, with a few optional/minor suggestions.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java
```diff
@@ -484,47 +493,54 @@ public LoadBundleTask addOnProgressListener(
...
+      currentListeners = new ArrayDeque<>(progressListeners);
+      progressListeners.clear();
     }
+    invokeListeners(currentListeners, snapshot);
     completionSource.setException(exception);
   }
 
-  void updateProgress(LoadBundleTaskProgress snapshot) {
+  void updateProgress(LoadBundleTaskProgress progressUpdate) {
+    Queue<ManagedListener> currentListeners;
```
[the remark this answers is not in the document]
schmidt-sebastian: I thought about this more: Since I am always invoking the callback on an executor, there should only be two scenarios:

- `updateProgress` is executed on the same thread as the callbacks. This means that all `execute` calls will only run after the loop has finished processing.
- `updateProgress` is executed on a different thread from the callbacks. This means that the other thread won't be able to grab the lock to modify the progress listener.

With that, I am reasonably confident that we don't need the extra copy and can simplify this quite a bit.

(FYI - I could implement your suggestion above by changing the data structure to our internal ImmutableSet/Map implementation, but I do still prefer to keep in sync with Android's Task API.

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
```diff
@@ -115,6 +123,64 @@ public void testProgressListener() throws InterruptedException {
...
+          latch.countDown();
```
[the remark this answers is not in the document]
schmidt-sebastian: Swapping these will allow the test to pass before the assertions trigger (as the scope of the test is only bound by the latches). I think what we need to do is add try/finally blocks.

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
schmidt-sebastian: The test shows the stacktrace, but doesn't fail. I added a super ugly hack to ensure the test fails when this happens. Let me know if you can think of something better.

firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java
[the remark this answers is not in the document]
dconeybe: Ahh yes that makes sense. I missed the fact that the callbacks were being enqueued on an executor. In that case, I rescind my comment about the ConcurrentModificationException.

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
dconeybe: Are you able to use `org.junit.rules.ErrorCollector`? If so, I think it would be a perfect fit for reporting assertion failures from other threads.

https://junit.org/junit4/javadoc/latest/org/junit/rules/ErrorCollector.html

schmidt-sebastian: I saw this when I googled around, but it didn't quite strike my mind that this is exactly what I need. Thanks!

PR #2301, opened 2021-01-08
Add BundleCache

schmidt-sebastian:
This adds the Bundle Cache + Schema changes + Tests (port from Web).

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleCache.java
@@ -1,0 +1,42 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+import androidx.annotation.Nullable;
+
+/** Provides methods to save and read Firestore bundles. */
+public interface BundleCache {
+  /**
+   * Gets the saved BundleMetadata for a given bundle id. Returns {@code null} if no bundles are
+   * found under the given id.
+   */
+  @Nullable
+  BundleMetadata getBundleMetadata(String bundleId);
+
+  /**
+   * Saves the BundleMetadata from a bundle into local storage, using its id as the persistent key.
+   */
+  void saveBundleMetadata(BundleMetadata metadata);
+
+  /**
+   * Gets a saved NamedQuery for the given query name. Returns {@code null} if no queries are found
+   * under the given name.
+   */
+  @Nullable
+  NamedQuery getNamedQuery(String queryName);
+
+  /** Saves a NamedQuery from a bundle, using its name as the persistent key. */
+  void saveNamedQuery(NamedQuery query);
+}
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleCache.java
@@ -19,16 +19,14 @@ import androidx.annotation.Nullable;
 /** Provides methods to save and read Firestore bundles. */
 public interface BundleCache {
   /**
-   * Gets the saved BundleMetadata for a given bundle id. Returns {@code null} if no bundles are
-   * found under the given id.
+   * Gets the saved metadata for a given bundle id. Returns {@code null} if no bundles are found
+   * under the given id.
    */
   @Nullable
-  BundleMetadata getBundleMetadata(String bundleId);
+  Bundle getBundleMetadata(String bundleId);
 
-  /**
-   * Saves the BundleMetadata from a bundle into local storage, using its id as the persistent key.
-   */
-  void saveBundleMetadata(BundleMetadata metadata);
+  /** Saves the metadata from a bundle into local storage, using its id as the persistent key. */
+  void saveBundleMetadata(Bundle metadata);
 
   /**
    * Gets a saved NamedQuery for the given query name. Returns {@code null} if no queries are found
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleMetadata.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleMetadata.java
@@ -19,34 +19,53 @@ import com.google.firebase.firestore.model.SnapshotVersion;
 /** Represents a Firestore bundle saved by the SDK in its local storage. */
 /* package */ class BundleMetadata {
   private final String bundleId;
-  private final int schemaVersion;
+  private final int version;
   private final SnapshotVersion createTime;
+  private final int totalDocuments;
+  private final long totalBytes;
 
-  public BundleMetadata(String bundleId, int schemaVersion, SnapshotVersion createTime) {
+  public BundleMetadata(
+      String bundleId,
+      int version,
+      SnapshotVersion createTime,
+      int totalDocuments,
+      long totalBytes) {
     this.bundleId = bundleId;
-    this.schemaVersion = schemaVersion;
+    this.version = version;
     this.createTime = createTime;
+    this.totalDocuments = totalDocuments;
+    this.totalBytes = totalBytes;
   }
 
   /**
-   * @return Id of the bundle. It is used together with `createTime` to determine if a bundle has
-   *     been loaded by the SDK.
+   * Returns the ID of the bundle. It is used together with `createTime` to determine if a bundle
+   * has been loaded by the SDK.
    */
   public String getBundleId() {
     return bundleId;
   }
 
-  /** @return Schema version of the bundle. */
-  public int getSchemaVersion() {
-    return schemaVersion;
+  /** Returns the schema version of the bundle. */
+  public int getVersion() {
+    return version;
   }
 
   /**
-   * @return Snapshot version of the bundle if created by the Server SDKs, or else
-   *     SnapshotVersion.MIN.
+   * Returns the snapshot version of the bundle if created by the Server SDKs, or else
+   * SnapshotVersion.MIN.
    */
   public SnapshotVersion getCreateTime() {
     return createTime;
+  }
+
+  /** Returns the number of documents in the bundle. */
+  public int getTotalDocuments() {
+    return totalDocuments;
+  }
+
+  /** Returns the size of the bundle in bytes, excluding this BundleMetadata. */
+  public long getTotalBytes() {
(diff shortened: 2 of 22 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLBundleCacheTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.google.firebase.firestore.local;
+
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest = Config.NONE)
+public class SQLBundleCacheTest extends BundleCacheTestCase {
```
wu-hui, line 23: SQLiteBundleCacheTest ?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/BundleCacheTestCase.java
```diff
@@ -0,0 +1,130 @@
...
+        new NamedQuery(
+            "query-1", query("collection").limitToFirst(1), new SnapshotVersion(Timestamp.now()));
+    bundleCache.saveNamedQuery(expectedQuery);
+
+    NamedQuery actualQuery = bundleCache.getNamedQuery("query-1");
+    assertEquals(expectedQuery, actualQuery);
+  }
+
+  @Test
+  public void testReturnsSavedLimitToLastQueries() {
```
wu-hui, line 110: Note the point of having a limit-to-last test is queries saved in bundles are encoded as target with an extra boolean saying it is limit to last. 

So I think constructing a limit to last query here defeat the purpose, checking the JS implementation, it is constructing a limit to first query with the extra boolean (by creating the proto class manually), to ensure limit to last queries in bundles are translated correctly.

Also check `local_serializer.ts/fromBundleQuery` to see the actual translation for limit queries.

wu-hui (changes requested):
Thank you so much for helping out!

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleCache.java
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+import androidx.annotation.Nullable;
+
+/** Provides methods to save and read Firestore bundles. */
+public interface BundleCache {
+  /**
+   * Gets the saved BundleMetadata for a given bundle id. Returns {@code null} if no bundles are
```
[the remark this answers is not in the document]
schmidt-sebastian: In general, I tend to use literal references for types that are used in the signature that I am commenting on. I think you can reason about it if you replace `code` with `link` (which I would do if I implemented your suggestion). This would look as such:

```
/**
   * Gets the saved {@link BundleMetadata} for a given bundle id.
   */
BundleMetadata getBundleMetadata(String bundleId);
```

The generated JavaDoc HTML would not how two links to BundleMetadata, as the return type already links to the source location.

Let me know if this is convincing. If not, I'll change it :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleCache.java
```diff
@@ -0,0 +1,42 @@
...
+   * found under the given id.
+   */
+  @Nullable
+  BundleMetadata getBundleMetadata(String bundleId);
+
+  /**
+   * Saves the BundleMetadata from a bundle into local storage, using its id as the persistent key.
+   */
+  void saveBundleMetadata(BundleMetadata metadata);
```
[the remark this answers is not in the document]
schmidt-sebastian: I actually filed a bug about this: https://b.corp.google.com/issues/176907699

The other data access classes write the same data that they read and convert to Proto when the data is accessed. We can maybe use the only the proto classes and remove our model classes, but I think we should settle on one data type and use it throughout this interface.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleMetadata.java
```diff
@@ -0,0 +1,94 @@
...
+package com.google.firebase.firestore.local;
+
+import com.google.firebase.firestore.model.SnapshotVersion;
+
+/** Represents a Firestore bundle saved by the SDK in its local storage. */
+/* package */ class BundleMetadata {
+  private final String bundleId;
+  private final int version;
+  private final SnapshotVersion createTime;
+  private final int totalDocuments;
```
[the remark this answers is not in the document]
schmidt-sebastian: Okay, I got confused there. I renamed this class to Bundle and removed the extra fields.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundledDocumentMetadata.java
```diff
@@ -0,0 +1,77 @@
...
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+import com.google.firebase.firestore.model.DocumentKey;
+import com.google.firebase.firestore.model.SnapshotVersion;
+import java.util.List;
+
+/** Metadata describing a Firestore document saved in the bundle. */
+public class BundledDocumentMetadata {
```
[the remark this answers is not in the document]
schmidt-sebastian: I think I might need this class since I cannot directly access a JSON object, but need to convert JSON in Java before I can use it in the BundleLoader. I added a TODO to see if I can remove it.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/BundleCacheTestCase.java
schmidt-sebastian: Since I updated the data layer to use only NamedQuery, I don't have a way to inject Protos in this test. I do have a test thought that likely serves the same purpose (`testEncodesLimitToLastQuery` in the file below). This test actually uncovered an issue I had during the development of this PR, since I originally called toTarget() before serialization which flipped the order by constraints.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLBundleCacheTest.java
schmidt-sebastian: Yes!

schmidt-sebastian (commented):
Thanks for the review!

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
```diff
@@ -266,4 +269,32 @@ TargetData decodeTargetData(com.google.firebase.firestore.proto.Target targetPro
         lastLimboFreeSnapshotVersion,
         resumeToken);
   }
+
+  public com.google.firestore.proto.BundledQuery encodeQuery(Query query) {
+    // We store LimitToLast as metadata information below. We remove it from the query itself as
+    // `toTarget()` flips the order by constraints otherwise.
+    Query queryWithoutLimitToLast =
```
wu-hui, line 276: I think we could avoid this by parsing json into proto class directly. This is required here because `Query` the model object has its own handling logic for limit to last. 

In order to make the interface of `BundleSerializer` consistent, we can introduce a new `local.BundledQuery` which holds a proto `StructuredQuery`. `local.NamedQuery` holds a `BundledQuery` instead of `Query`.

In `BundleSerializer` the interface stays unchanged, but the implementation will work with `StructuredQuery` directly, without having to deal with `Query`.

This way, the `encodeQuery` here can be eliminated. We can also have the test for limit to last queries the same way as in JS, constructing a `NamedQuery.bundledQuery` directly and save it, read it back from persistence and make sure it is a proper limit to last query.

wu-hui (changes requested):
Sorry, it took me a while to figure out. I originally imagined `BundleSerializer` in the other PR will be parsing into proto classes. 

I still think we should consider parsing into proto class of `BundledQuery` at least, to avoid the awkward `limitToFirst`/`limitToLast` dance. See below a detailed proposal. 

Please let me know if this makes sense at all, we can work this out over GVC if it makes it quicker.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
schmidt-sebastian: There are two problems I encountered:

- It it is not possible to instantiate a Query object from a Target. If that was possible, `BundledQuery` would contain a Target with an additional bit that describes the limit. 
- It is also not possible to get the original order constraints from the query, since `toTarget()` flips the ordering constraint. The hack here aims to undo this.

Furthermore, the Android SDK cannot deal with StructuredQuery directly. Instead, it expects the data types that we use in Target. If we don't deserialize here, we have to at another point (e.g. to allow a user to modify a named query).

To clean this up, I think we need to somehow figure out how to change Query to only contain a Target and a Limit flag. The target object would then contain the constraints and filters. This would allow us to construct a Query from a Target and to retrieve the Target from a Query without modifying it.

Anyways, I will schedule some time to discuss.

schmidt-sebastian: I updated this based on our offline discussion and replaces the `queryWithoutLimitToLast` without something a little less hacky. Hope this works :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
```diff
@@ -271,12 +271,25 @@ TargetData decodeTargetData(com.google.firebase.firestore.proto.Target targetPro
...
 
   public com.google.firestore.proto.BundledQuery encodeQuery(Query query) {
-    // We store LimitToLast as metadata information below. We remove it from the query itself as
-    // `toTarget()` flips the order by constraints otherwise.
-    Query queryWithoutLimitToLast =
-        query.hasLimitToLast() ? query.limitToFirst(query.getLimitToLast()) : query;
+    // We don't use query.toTarget() here as we want to save the original query, and not the flipped
+    // query that we may send to the backend (if limitToLast() is used). This allows us to
+    // deserialize named queries again without inverting the order by constraints on read.
+    long limit =
```
wu-hui, line 277: This is better!

What do you think about eliminate `Query` in the picture completely and use `Target` only?

I am thinking:

- Make `NamedQuery` hold `Target` and a limit type enum.
- `BundleSerializer.decodeQuery` becomes `decodeTarget`.
-  This method here becomes `encodeNamedQuery`, the `Target` and limitType should both be available. The `decodeQuery` below becomes `decodeBundledQuery`.

wu-hui: Above should also make porting the JS test of saving and reading a limit to last query possible?

schmidt-sebastian: NamedQuery now contains a BundledQuery model. This matches the Protobuf and gets rid of all this indirection. I didn't go with Target/LimitType in the NameQuery as I needed a return type for the methods in LocalSerializer, which decodes both the limit type and the name query.

There are two tests that test the round-trip encoding. I will go through your "bundle to master" merge PR when I am close to finishing it up to ensure that I am not missing tests.

wu-hui (approved):
Thanks for working this out, LGTM!

PR #2308, opened 2021-01-09
JSON Decoder

schmidt-sebastian:
This adds a JSON-to-Protobuf serializer that is used for Bundles. It used org.json for JSON decoding, which is part of the Android SDK (API Level 1).

The Value-decoding tests are inspired by the existing tests in RemoteSerializerTest.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
@@ -23,6 +23,7 @@ import com.google.firebase.firestore.core.FieldFilter;
 import com.google.firebase.firestore.core.Filter;
 import com.google.firebase.firestore.core.OrderBy;
 import com.google.firebase.firestore.core.Query;
+import com.google.firebase.firestore.core.Target;
 import com.google.firebase.firestore.model.Document;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.FieldPath;
@@ -59,9 +60,9 @@ class BundleSerializer {
 
   public NamedQuery decodeNamedQuery(JSONObject namedQuery) throws JSONException {
     String name = namedQuery.getString("name");
-    Query query = decodeQuery(namedQuery.getJSONObject("bundledQuery"));
+    BundledQuery bundledQuery = decodeBundledQuery(namedQuery.getJSONObject("bundledQuery"));
     SnapshotVersion readTime = decodeSnapshotVersion(namedQuery.getJSONObject("readTime"));
-    return new NamedQuery(name, query, readTime);
+    return new NamedQuery(name, bundledQuery, readTime);
   }
 
   public BundleMetadata decodeBundleMetadata(String json) throws JSONException {
@@ -72,9 +73,7 @@ class BundleSerializer {
     String bundleId = bundleMetadata.getString("id");
     int version = bundleMetadata.getInt("version");
     SnapshotVersion createTime = decodeSnapshotVersion(bundleMetadata.getJSONObject("createTime"));
-    int totalDocuments = bundleMetadata.getInt("totalDocuments");
-    long totalBytes = bundleMetadata.getLong("totalBytes");
-    return new BundleMetadata(bundleId, version, createTime, totalDocuments, totalBytes);
+    return new BundleMetadata(bundleId, version, createTime);
   }
 
   public BundledDocumentMetadata decodeBundledDocumentMetadata(String json) throws JSONException {
@@ -134,7 +133,7 @@ class BundleSerializer {
     return new SnapshotVersion(new Timestamp(seconds, nanos));
   }
 
-  private Query decodeQuery(JSONObject bundledQuery) throws JSONException {
+  private BundledQuery decodeBundledQuery(JSONObject bundledQuery) throws JSONException {
     JSONObject structuredQuery = bundledQuery.getJSONObject("structuredQuery");
     verifyNoSelect(structuredQuery);
 
@@ -159,7 +158,8 @@ class BundleSerializer {
     int limit = decodeLimit(structuredQuery);
     Query.LimitType limitType = decodeLimitType(bundledQuery);
 
-    return new Query(parent, collectionGroup, filters, orderBys, limit, limitType, startAt, endAt);
+    return new BundledQuery(
+        new Target(parent, collectionGroup, filters, orderBys, limit, startAt, endAt), limitType);
   }
 
   private int decodeLimit(JSONObject structuredQuery) {
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
@@ -17,7 +17,6 @@ package com.google.firebase.firestore.local;
 import android.util.Base64;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
-
 import com.google.firebase.Timestamp;
 import com.google.firebase.firestore.core.Bound;
 import com.google.firebase.firestore.core.FieldFilter;
@@ -46,7 +45,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 /** A JSON serializer to deserialize Firestore Bundles. */
-/* package= */ class BundleSerializer {
+class BundleSerializer {
 
   private final RemoteSerializer remoteSerializer;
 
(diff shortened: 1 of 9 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,350 @@
...
+    long totalBytes = jsonObject.getLong("totalBytes");
+    return new BundleMetadata(bundleId, version, createTime, totalDocuments, totalBytes);
+  }
+
+  public BundledDocumentMetadata fromBundledDocumentMetadata(String json) throws JSONException {
+    JSONObject jsonObject = new JSONObject(json);
+    DocumentKey key = DocumentKey.fromName(jsonObject.getString("name"));
+    SnapshotVersion readTime = decodeSnapshotVersion(jsonObject.getJSONObject("readTime"));
+    boolean exists = jsonObject.optBoolean("exists", false);
+    JSONArray queriesJson = jsonObject.optJSONArray("queries");
```
wilhuff, line 80: An empty array is different from a missing array. Are you sure you want to conflate these two notions here? If so, it may be worth a comment on why that's OK.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,350 @@
...
+/** A JSON serializer to deserialize Firestore Bundles. */
+/* package= */ class BundleSerializer {
+
+  private final RemoteSerializer remoteSerializer;
+
+  public BundleSerializer(RemoteSerializer remoteSerializer) {
+    this.remoteSerializer = remoteSerializer;
+  }
+
+  public NamedQuery fromNamedQuery(String json) throws JSONException {
```
wilhuff, line 57: Suggestion: for consistency in the API here, consider making this take `JSONObject`. This way the source of the JSON doesn't affect the interface. If production code really does need to read a named query from a string, maybe leave that as an overload that parses and then invokes the other one?

wilhuff, line 57: This class is mixing "from" and "decode" prefixes for decoding methods. The other android serializers use "decode" so maybe stay consistent with that?

(We do have static factory methods with "from" names, but these work well because the class name gets prefixed: `DocumentKey.fromName` reads clearly, but just `jsonSerializer.fromName` would not.)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,350 @@
...
+    } else if (structuredQuery.has("fieldFilter")) {
+      decodeFieldFilter(result, structuredQuery.getJSONObject("fieldFilter"));
+    } else if (structuredQuery.has("unaryFilter")) {
+      decodeUnaryFilter(result, structuredQuery.getJSONObject("unaryFilter"));
+    }
+  }
+
+  private void decodeCompositeFilter(List<Filter> result, JSONObject compositeFilter)
+      throws JSONException {
+    JSONArray filters = compositeFilter.optJSONArray("filters");
```
wilhuff, line 203: Should this verify that the composite filter is AND?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+
+  private FieldPath decodeFieldReference(JSONObject fieldReference) throws JSONException {
+    return FieldPath.fromServerFormat(fieldReference.getString("fieldPath"));
+  }
+
+  private Query.LimitType decodeLimitType(JSONObject bundledQueryJson) {
+    String limitType = bundledQueryJson.optString("limitType", "FIRST");
+    return limitType.equals("FIRST")
```
wilhuff, line 321: Optional: consider rejecting invalid limit types.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/BundleSerializerTest.java
```diff
@@ -0,0 +1,701 @@
...
+      Value proto = Value.newBuilder().setBooleanValue(test).build();
+      assertValueRoundTrip(json, proto);
+    }
+  }
+
+  @Test
+  public void testDecodesIntegerValues() throws JSONException {
+    List<Long> tests = asList(Long.MIN_VALUE, -100L, -1L, 0L, 1L, 100L, Long.MAX_VALUE);
+    for (Long test : tests) {
+      String json = "{ integerValue: '" + test + "' }";
```
wilhuff, line 82: Are integer values always encoded as strings? If not, maybe we should have cases that verify that the integer encoded as an integer also works?

Separately, single quoted strings are valid JavaScript but invalid JSON. If we are testing for a string, shouldn't this be double-quoted?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/BundleSerializerTest.java
```diff
@@ -0,0 +1,701 @@
...
+        new BundledDocumentMetadata(
+            key("coll/doc"),
+            new SnapshotVersion(new com.google.firebase.Timestamp(1, 2)),
+            true,
+            Arrays.asList("query-1", "query-2"));
+    BundledDocumentMetadata actualMetadata = serializer.fromBundledDocumentMetadata(json);
+    assertEquals(expectedMetadata, actualMetadata);
+  }
+
+  private void assertValueRoundTrip(String json, Value proto) throws JSONException {
```
wilhuff, line 644: Since this isn't actually round-tripping the values, should this be named `assertDecodes`, `assertDecodesValue`, or something like that?

Also applies to other methods with a `RoundTrip` suffix.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
schmidt-sebastian: Renamed all methods to use `decode` prefix.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,350 @@
...
+/* package= */ class BundleSerializer {
+
+  private final RemoteSerializer remoteSerializer;
+
+  public BundleSerializer(RemoteSerializer remoteSerializer) {
+    this.remoteSerializer = remoteSerializer;
+  }
+
+  public NamedQuery fromNamedQuery(String json) throws JSONException {
+    JSONObject JSONObject = new JSONObject(json);
```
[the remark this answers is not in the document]
schmidt-sebastian: I figured this would GZIP better if I re-used the class names :) Thanks for catching. I changed all "JSONObject" variables to use the containing data type as the name.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,350 @@
...
+    int version = jsonObject.getInt("version");
+    SnapshotVersion createTime = decodeSnapshotVersion(jsonObject.getJSONObject("createTime"));
+    int totalDocuments = jsonObject.getInt("totalDocuments");
+    long totalBytes = jsonObject.getLong("totalBytes");
+    return new BundleMetadata(bundleId, version, createTime, totalDocuments, totalBytes);
+  }
+
+  public BundledDocumentMetadata fromBundledDocumentMetadata(String json) throws JSONException {
+    JSONObject jsonObject = new JSONObject(json);
+    DocumentKey key = DocumentKey.fromName(jsonObject.getString("name"));
```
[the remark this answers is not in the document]
schmidt-sebastian: I updated this a bit. I am using the RemoteSerializer to validate the key along and whether t it belongs to the current project, but throwing and InvalidArgumentException instead on an assert when it this is not true.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
schmidt-sebastian: It seems safer to treat a missing array and an empty array the same, despite the fact that a missing array might point to a problem in Server SDKs. Added comment.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
schmidt-sebastian: We can, but there is no other valid value for now. I added a validation.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
schmidt-sebastian: Done

firebase-firestore/src/test/java/com/google/firebase/firestore/local/BundleSerializerTest.java
schmidt-sebastian: It probably depends on the implementation of the SDK. Since we have to be able to read data produced by 7 different JSON implementations, we should probably err on the side of caution. I added another test.

I originally used double quotes throughout this file but it made the sources pretty hard to read. Added a comment to the top of the file.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/BundleSerializerTest.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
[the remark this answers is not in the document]
wilhuff: This one is still not fixed. Feel free to inline this variable to avoid the naming question :-).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/BundleSerializer.java
```diff
@@ -0,0 +1,383 @@
...
+    long seconds = timestamp.optLong("seconds", 0);
+    int nanos = timestamp.optInt("nanos", 0);
+    return new SnapshotVersion(new Timestamp(seconds, nanos));
+  }
+
+  private Query decodeQuery(JSONObject bundledQuery) throws JSONException {
+    JSONObject structuredQuery = bundledQuery.getJSONObject("structuredQuery");
+
+    ResourcePath parent = decodeName(bundledQuery.getString("parent"));
+    verifyNoSelect(structuredQuery);
```
wilhuff, line 145: This seems like it could be paired with the read of the `structuredQuery` (similar to the way `from` is validated immediately after reading it).

wilhuff (approved):
LGTM

PR #2310, opened 2021-01-10
Use custom tab only for web actions

dlackty:
Currently, FIAM uses Custom Tabs to open every action URI and results in errors for deep links.

Refering to [iOS SDK's implementation](https://github.com/firebase/firebase-ios-sdk/blob/b41a1273368e8af2af26d1dfcb135fc2307b94b3/FirebaseInAppMessaging/Sources/Runtime/FIRIAMActionURLFollower.m#L107-L129), which checks whether URI schems is http/https, I implemented a similar logic in this PR.

Not pretty sure if my test change is sufficient as this is my first time to contribute to the project. 
So please let me know if there's anything missing.

### DIFF
--- a/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
+++ b/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
@@ -534,7 +534,7 @@ public class FirebaseInAppMessagingDisplay extends FirebaseInAppMessagingDisplay
   }
 
   private void launchUriIntent(Activity activity, Uri uri) {
-    if (supportsCustomTabs(activity)) {
+    if (shouldOpenWeb(uri) && supportsCustomTabs(activity)) {
       // If we can launch a chrome view, try that.
       CustomTabsIntent customTabsIntent = new CustomTabsIntent.Builder().build();
       Intent intent = customTabsIntent.intent;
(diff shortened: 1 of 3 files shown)


dlackty:
According to the commit history, I guess @JasonAHeron might want to have a look. Sorry if I ping the wrong person.

firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -563,4 +563,9 @@ private boolean supportsCustomTabs(Activity activity) {
         activity.getPackageManager().queryIntentServices(customTabIntent, 0);
     return resolveInfos != null && !resolveInfos.isEmpty();
   }
+
+  private boolean shouldOpenWeb(Uri uri) {
```
eldhosembabu, line 567: wdyt about having a different name for this method? like ishttpOrHttpsUri or something else since that is what this method is checking.

eldhosembabu (commented):
This PR looks good to me.

dlackty:
@eldhosembabu Pushed some changes according the suggestions, and also rebased the branch. Thank you.

firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -563,4 +563,12 @@ private boolean supportsCustomTabs(Activity activity) {
         activity.getPackageManager().queryIntentServices(customTabIntent, 0);
     return resolveInfos != null && !resolveInfos.isEmpty();
   }
+
+  private boolean ishttpOrHttpsUri(Uri uri) {
+    if (uri == null) {
+      return false
```
samtstern, line 569:
```suggestion
      return false;
```

dlackty: Thanks. Sorry for the careless mistake.

eldhosembabu (approved):

samtstern:
@eldhosembabu this is green and approved, do you want to merge it?

PR #2316, opened 2021-01-12
Fixing issues related with package visibility changes for Android 11.

eldhosembabu:
Fixing issues related with package visibility changes for Android 11.

While targeting android 11, querying for activities or services which can handle intent returns 0 results. To support this querying in Android 11 and above we need to have the 'queries' tag mentioned in the manifest. Please see: https://developer.android.com/about/versions/11/privacy/package-visibility for details.


prakhar1989 (approved):
Awesome!

PR #2319, opened 2021-01-13
Remove support for legacy reports endpoint

mrwillis21:
The legacy reporting endpoint has been retired, this removes support for it from the SDK.


mrichards (approved):
@mrwillis21 & I have manually tested all P0 & P1 test cases across a variety of API levels and found no regressions, so my review pass was largely for style / clarity. LGTM, nice work Matt!

PR #2327, opened 2021-01-14
Make RTDB query get hit in-memory listener view cache first

IanWyszynski:
Motivated by https://b.corp.google.com/issues/177263312. When a query has an active listener, it's in-memory cached value is almost certainly the latest server-value, so we don't need to perform a network round-trip to get it.


firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -4474,77 +4479,99 @@ public void getQueryBasic() throws DatabaseException, InterruptedException, Exec
...
-    ref.addValueEventListener(listener);
-    ref.setValue(42L);
-    IntegrationTestHelpers.waitFor(semaphore);
-    db.goOffline();
-    try {
-      // Since we still have a listener on `ref`, the 42L should be cached here.
-      assertEquals(42L, Tasks.await(ref.get()).getValue());
+            @Override
+            public void onCancelled(@NonNull DatabaseError error) {
+              fail(error.getMessage());
```
schmidt-sebastian, line 4515: Note that this won't actually fail the test since it is running in a different thread. It will however print the error to the console, and the test should fail because the semaphore times out.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -4474,77 +4479,99 @@ public void getQueryBasic() throws DatabaseException, InterruptedException, Exec
...
+            }
+          });
+      IntegrationTestHelpers.waitFor(semaphore);
+
+      readerDb.goOffline();
+
+      assertEquals(Tasks.await(reader.get()).getValue(), 42L);
     } finally {
-      ref.removeEventListener(listener);
+      readerDb.goOnline();
```
schmidt-sebastian, line 4524: Should we remove the listener?

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -4474,77 +4479,99 @@ public void getQueryBasic() throws DatabaseException, InterruptedException, Exec
...
+    reader.addValueEventListener(listener);
+    IntegrationTestHelpers.waitFor(semaphore);
 
-    assertEquals(43L, Tasks.await(reader.get()).getValue());
+    readerDb.goOffline();
+    try {
+      reader.removeEventListener(listener);
+      assertEquals(Tasks.await(reader.get()).getValue(), 42L);
+    } finally {
+      readerDb.goOffline();
```
schmidt-sebastian, line 4569: goOnline?

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -496,6 +494,14 @@ public void onRequestResult(String optErrorCode, String optErrorMessage) {
         new Runnable() {
           @Override
           public void run() {
+            Node cached =
```
schmidt-sebastian, line 497: Maybe add explicit comment that this excludes persistence.

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -529,19 +530,18 @@ public void onComplete(@NonNull Task<Object> task) {
...
+                              @Override
+                              public void run() {
+                                serverSyncTree.setQueryInactive(query.getSpec());
+                              }
+                            });
                       }
                     });
           }
         });
-    return source
```
schmidt-sebastian, line 536: I think you could keep this as is and just pass the executor that `scheduleNow` uses as the first argument to `addOnCompleteListener`. That would be a bit easier to follow in my opinion.

firebase-database/src/main/java/com/google/firebase/database/core/SyncTree.java
```diff
@@ -857,6 +865,21 @@ private Tag tagForQuery(QuerySpec query) {
     return this.queryToTagMap.get(query);
   }
 
+  /** Similar to calcCompleteEventCache, but doesn't skip the root view cache. */
+  public Node calcCompleteEventCacheFromRoot(Path path, List<Long> writeIdsToExclude) {
+    ImmutableTree<SyncPoint> tree = this.syncPointTree;
```
schmidt-sebastian, line 870: You can drop "this." here and below (and IMHO you can just inline "tree").

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -4554,16 +4581,22 @@ public void getUpdatesPersistenceCacheWhenEnabled()
...
 
     readerDb.goOffline();
-
-    Semaphore semaphore = new Semaphore(0);
-    assertNotNull(ReadFuture.untilEquals(reader, 42L).timedGet());
+    try {
+      assertNotNull(ReadFuture.untilEquals(reader, 42L).timedGet());
+    } finally {
+      readerDb.goOnline();
+    }
```
schmidt-sebastian, line 4599: Can you add a test that verifies that we don't go back to the `SingleValueEvent` behavior. That is:

- Client 1: Read value:old from cache
- Client 1: Disconnect all listeners
- Client 2: Write value:new
- Client 1:  Get should now return value:new

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
IanWyszynski: I don't think the executor service in the DefaultRunLoop is available here:

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/core/utilities/DefaultRunLoop.java#L99

Since it's hidden behind the RunLoop interface:

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/core/RunLoop.java#L24

This feels like more trouble than it's worth for marginal readability gains. Let me know if I missed something.

schmidt-sebastian (commented):
Thanks for adding the test. As discussed offline, please move all state access to the run loop.

schmidt-sebastian (approved):

PR #2328, opened 2021-01-15
Key comparison for RTDB queries should correctly handle boundary sentinels

IanWyszynski:
In order for `startAfter` and `endBefore` to have the correct ordering semantics, we need to return the objects `MAX_KEY` and `MIN_KEY`. Otherwise, the comparison function won't work properly in all cases:

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/snapshot/ChildKey.java#L74-L76


schmidt-sebastian (commented):
Hm. Should we update the comparison function instead? This bug can show up in a number of places.

schmidt-sebastian (approved):

PR #2329, opened 2021-01-15
Add ByteBufferInputStream

schmidt-sebastian:
This will allow all three overloads for loadBundle() to use the same decoding logic.

See https://github.com/firebase/firebase-android-sdk/pull/2322 for how this all comes together.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/util/ByteBufferInputStream.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/util/ByteBufferInputStream.java
@@ -1,0 +1,46 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/util/ByteBufferInputStream.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public ByteBufferInputStream(ByteBuffer buf) {
+    this.buffer = buf;
+  }
+
+  public int read() {
+    if (!buffer.hasRemaining()) {
+      return -1;
+    }
+    return buffer.get() & 0xFF;
```
wu-hui, line 34: Add a comment here to explain why `& 0xFF`?

wu-hui (approved):
LGTM.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/ByteBufferInputStream.java
schmidt-sebastian: Done

PR #2333, opened 2021-01-16
Add Bundle Reader

schmidt-sebastian:
Port of the JS Bundle Reader + its tests.


schmidt-sebastian:
@wilhuff Added you as well to review the management of the CharBuffer.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -0,0 +1,198 @@
...
+   * <p>Returns a string decoded from the read bytes.
+   */
+  private String readJsonString(int length) throws IOException {
+    char[] c = new char[length];
+
+    int read = Math.min(length, buffer.remaining());
+    buffer.get(c, 0, read);
+
+    while (read < length) {
+      if (!pullMoreData()) {
```
wilhuff, line 164: This is a wasted copy. There's no reason to copy into the buffer only to use that to copy into the character array. You might as well just directly read into `c`. However, creating a `String` from a character array is yet another copy...

Probably the right resolution here is to make `c` a `StringBuilder` and then `append` the buffer contents into it. `CharBuffer` implements `CharSequence` so this should be straightforward.

Note that filling the buffer the way you're doing isn't the most efficient way to do this. For as many bytes as you've read, you might as well append them directly into the target StringBuilder.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -0,0 +1,198 @@
...
+   * @return whether more data is available
+   */
+  private boolean pullMoreData() throws IOException {
+    if (buffer.remaining() == 0) {
+      buffer.compact();
+    }
+
+    int read;
+    do {
+      read = dataReader.read(buffer);
```
wilhuff, line 186: This loop calls `read` at least twice if not at EOF. I think you could fix it by checking `while (read > 0 && buffer.remaining() > 0)`. That is, if it manages to fill the buffer in a single go, it will fall out of the loop without attempting to read into the full buffer.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -0,0 +1,198 @@
...
+import java.nio.CharBuffer;
+import org.json.JSONException;
+
+/**
+ * Reads the length-prefixed JSON stream for Bundles.
+ *
+ * <p>The class takes a bundle stream and presents abstractions to read bundled elements out of the
+ * underlying content.
+ */
+public class BundleReader extends BundleElement {
```
wu-hui, line 30: Why would this extend `BundleElement`?

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleElement.java
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.bundle;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public abstract class BundleElement {
```
wu-hui, line 20: Is this corresponded to `SizedBundleElement` in JS, and the "size" part is not implemented yet. But it seems like `BundlerReader` here directly provides `bytesRead`.

If it is intended as I understand it, I'd prefer to making it an mark interface, and move `fromJson` to `BundleReader`.

wu-hui (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleElement.java
schmidt-sebastian: It is meant to be like the sized element, but I think it is cleaner if it doesn't carry the size information in the Java port. This allows me to do instanceof checks. Instead, the BundleReader has separate state to expose the number of bytes read.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
schmidt-sebastian: Hmmm... removed.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -0,0 +1,198 @@
...
+  private final BundleSerializer serializer;
+  private final InputStreamReader dataReader;
+
+  @Nullable BundleMetadata metadata;
+  private CharBuffer buffer;
+  long bytesRead;
+
+  public BundleReader(BundleSerializer serializer, InputStream data) {
+    this.serializer = serializer;
+    dataReader = new InputStreamReader(data);
```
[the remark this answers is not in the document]
schmidt-sebastian: I used the `Charset.forName` lookup since it doesn't throw checked exceptions (the String overload for InputStreamReader does).

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -0,0 +1,198 @@
...
+   */
+  @Nullable
+  private BundleElement readNextElement() throws IOException, JSONException {
+    int length = readLength();
+    if (length == -1) {
+      return null;
+    }
+
+    String json = readJsonString(length);
+    bytesRead += (int) (Math.log10(length) + 1) + length;
```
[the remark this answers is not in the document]
schmidt-sebastian: I rewrote this to make this much more obvious.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
schmidt-sebastian: I tried to figure out how to use a StringBuilder here without making an additional copy. I didn't realize that I could just use `StringBuilder.append()`. Updated the code.

Thanks for the help!

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
schmidt-sebastian: I updated this to only read once. There is an outer loop that already polls so we don't need this here.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wilhuff: This is still computing characters read and not bytes.

wilhuff (commented):
Pretty much LGTM, though one remaining slightly larger than nit.

wu-hui (approved):
LGTM.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
schmidt-sebastian: Addressed and also added a test that fails without this fix.

wilhuff (approved):
LGTM

PR #2334, opened 2021-01-18
Mark Dependency.optional as deprecated.

vkryachko:


firebase-components/src/main/java/com/google/firebase/components/Dependency.java
```diff
@@ -47,30 +47,82 @@ private Dependency(Class<?> anInterface, @Type int type, @Injection int injectio
...
+  @Deprecated
   public static Dependency optional(Class<?> anInterface) {
     return new Dependency(anInterface, Type.OPTIONAL, Injection.DIRECT);
   }
 
+  /**
+   * Declares a deferred dependency.
+   *
+   * <p>Such dependencies are optional and may not be present be default. But they can become
+   * available if a dynamic module is installed that they are contained in.
```
rlazo, line 68:
```suggestion
   * available if a dynamic module that contains them is installed.
```

firebase-components/src/main/java/com/google/firebase/components/Dependency.java
```diff
@@ -47,30 +47,82 @@ private Dependency(Class<?> anInterface, @Type int type, @Injection int injectio
...
+   */
+  @Deprecated
   public static Dependency optional(Class<?> anInterface) {
     return new Dependency(anInterface, Type.OPTIONAL, Injection.DIRECT);
   }
 
+  /**
+   * Declares a deferred dependency.
+   *
+   * <p>Such dependencies are optional and may not be present be default. But they can become
```
rlazo, line 67:
```suggestion
   * <p>Such dependencies are optional and may not be present by default. But they can become
```

firebase-components/src/main/java/com/google/firebase/components/Dependency.java
```diff
@@ -47,30 +47,82 @@ private Dependency(Class<?> anInterface, @Type int type, @Injection int injectio
...
   public static Dependency setOf(Class<?> anInterface) {
     return new Dependency(anInterface, Type.SET, Injection.DIRECT);
   }
 
+  /**
+   * Declares an optional dependency.
+   *
+   * <p>Optional dependencies can be missing at runtime(being {@code null}) and dependents must be
+   * ready to handle that.
+   */
```
rlazo, line 102: This isn't deprecated?

vkryachko: Nope, this is now the recommended optional dependency, see the deprecation message in `Dependency.optional()`.

rlazo (approved):

PR #2340, opened 2021-01-19
Remove unused baseDoc argument

schmidt-sebastian:
Port from iOS https://github.com/firebase/firebase-ios-sdk/pull/7276

This likely needs to be ported to Web too (but I didn't check - just worked on this while verifying the iOS PR).


thebrianchen (approved):
lgtm

PR #2346, opened 2021-01-20
Implement BundleLoader

schmidt-sebastian:
BundleLoader is a port from Web, the tests are new since BundleLoader on Web only has integration tests. To make testing possible, I also added BundleListener


schmidt-sebastian:
@wu-hui This has now been merged with the latest changes and only contains the unreviewed diff.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleListener.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.firebase.firestore.bundle;
+
+import com.google.firebase.database.collection.ImmutableSortedMap;
+import com.google.firebase.database.collection.ImmutableSortedSet;
+import com.google.firebase.firestore.model.DocumentKey;
+import com.google.firebase.firestore.model.MaybeDocument;
+
+/** Interface implemented by components that can apply changes from a bundle. */
+public interface BundleListener {
```
wu-hui, line 23: Maybe call it `BundleStorage`? And update comment to `...that can apply data from bundles to local storage`?

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleLoader.java
```diff
@@ -0,0 +1,154 @@
...
+   */
+  public @Nullable LoadBundleTaskProgress addElement(BundleElement bundleElement, long byteSize) {
+    Preconditions.checkArgument(
+        !(bundleElement instanceof BundleMetadata), "Unexpected bundle metadata element.");
+
+    boolean updateProgress = false;
+
+    if (bundleElement instanceof NamedQuery) {
+      queries.add((NamedQuery) bundleElement);
+    } else if (bundleElement instanceof BundledDocumentMetadata) {
```
wu-hui, line 76: I am wondering if we could bring the implementation more consistent to the JS implementation, or maybe there is something I don't see that is blocking it?

This is mostly a concern of inconsistency, but JS also does not have to use a sorted map.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleLoader.java
```diff
@@ -0,0 +1,154 @@
...
+            totalDocuments,
+            bytesLoaded,
+            totalBytes,
+            null,
+            LoadBundleTaskProgress.TaskState.RUNNING)
+        : null;
+  }
+
+  /** Applies the loaded documents and queries to local store. Returns the document view changes. */
+  public ImmutableSortedMap<DocumentKey, MaybeDocument> applyChanges() {
```
wu-hui, line 116: This is a better name than `complete`!

wu-hui (commented):
Almost LGTM, I do wonder if we can make the `addElement` method more consistent to JS.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleListener.java
schmidt-sebastian: We seem to call these types of interfaces "Callback": OnlineStateCallback, SyncEngineCallback, StreamCallback, RemoteStoreCallback. I have updated this to use "callback" as well, but also used your comment enhancement as it still applies.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleLoader.java
schmidt-sebastian: There are two differences here that stand out to me:

- The metadata lookup in JS looks at the gathered documents. I decided to not follow this pattern as I would like to build up an `ImmutableSortedMap<DocumentKey, MaybeDocument>` directly as this is a common data structure I can use this map directly in LocalStore. As such, I need some place to denote the last stored document key. 
I can remove this altogether though since it is only used for asserts.

- The `updateProgress` check is done differently here. I prefer the `updateProgress` a little, but not enough to convince you that it is worth backporting. I updated it to be more aligned with the Web.

schmidt-sebastian (commented):
Thanks for the review!

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleLoader.java
wu-hui: Looks good.

wu-hui (approved):
Thanks, LGTM!

PR #2350, opened 2021-01-21
Use JSON timestamps for Bundles

schmidt-sebastian:
Sad panda :(


schmidt-sebastian:
cc @wu-hui

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
```diff
@@ -307,11 +323,92 @@ private void decodeGeoPoint(Value.Builder builder, JSONObject geoPoint) {
...
-  private void decodeTimestamp(Value.Builder builder, JSONObject timestamp) {
+  private Timestamp decodeTimestamp(String timestamp) {
+    // This method is copied from com.google.protobuf.util.Timestamps. See:
+    // https://chromium.googlesource.com/external/github.com/google/protobuf/+/HEAD/java/util/src/main/java/com/google/protobuf/util/Timestamps.java?autodive=0%2F#232
+
+    try {
+      int dayOffset = timestamp.indexOf('T');
+      if (dayOffset == -1) {
+        throw new IllegalArgumentException(
+            "Failed to parse timestamp: invalid timestamp \"" + timestamp + "\"");
```
wilhuff, line 334: Optional: you could simplify this message to just "Invalid timestamp ...".

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
```diff
@@ -307,11 +323,92 @@ private void decodeGeoPoint(Value.Builder builder, JSONObject geoPoint) {
...
+      int timezoneOffsetPosition = timestamp.indexOf('Z', dayOffset);
+      if (timezoneOffsetPosition == -1) {
+        timezoneOffsetPosition = timestamp.indexOf('+', dayOffset);
+      }
+      if (timezoneOffsetPosition == -1) {
+        timezoneOffsetPosition = timestamp.indexOf('-', dayOffset);
+      }
+      if (timezoneOffsetPosition == -1) {
+        throw new IllegalArgumentException(
+            "Failed to parse timestamp: missing valid timezone offset.");
```
wilhuff, line 345: Please include the text of the invalid timestamp

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
```diff
@@ -61,7 +80,7 @@ public NamedQuery decodeNamedQuery(String json) throws JSONException {
   public NamedQuery decodeNamedQuery(JSONObject namedQuery) throws JSONException {
     String name = namedQuery.getString("name");
     BundledQuery bundledQuery = decodeBundledQuery(namedQuery.getJSONObject("bundledQuery"));
-    SnapshotVersion readTime = decodeSnapshotVersion(namedQuery.getJSONObject("readTime"));
+    SnapshotVersion readTime = decodeSnapshotVersion(namedQuery.getString("readTime"));
```
wilhuff, line 83: Do we need to support both? That is, is it ever possible that users might end up with timestamp values encoded as a seconds/nanoseconds pair instead?

I'm specifically concerned because `Timestamp.toJSON` uses the object format: https://github.com/firebase/firebase-js-sdk/blob/f68bd3f29c87e98f15635e38c1f9847b7731fa30/packages/firestore/src/lite/timestamp.ts#L162. It seems that if users construct bundles for themselves they could get this wrong.

wilhuff: That is to say, it seems like logic like we have in the JS SDK is required for dealing with JSON timestamps, no?

https://github.com/firebase/firebase-js-sdk/blob/f68bd3f29c87e98f15635e38c1f9847b7731fa30/packages/firestore/src/model/normalize.ts#L31

schmidt-sebastian: I updated this PR to support both formats. I am not sure we need this going forward, but this ensures backwards compatibility with existing bundles.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
```diff
@@ -38,16 +38,35 @@
...
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
 /** A JSON serializer to deserialize Firestore Bundles. */
 class BundleSerializer {
 
+  private static final long MILLIS_PER_SECOND = 1000;
+
+  private static final SimpleDateFormat timestampFormat;
```
wilhuff, line 59: DateFormat objects are not safe to use from multiple threads. While any one Firestore instance may only use this from a single thread, multiple Firestore instances can run at the same time. You could make this safe by making this a member rather than a static.

schmidt-sebastian: I thought this problem didn't apply to us since we are single-threaded, but you are right that it shows up when multiple instances are used. Thanks for pointing this out.

wilhuff (approved):
LGTM

PR #2353, opened 2021-01-22
Add model file clean-up on initial load for all models found in share…

annzimmer:
…d preferences.


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
```diff
@@ -61,6 +63,24 @@ public static ModelFileManager getInstance() {
     return FirebaseApp.getInstance().get(ModelFileManager.class);
   }
 
+  void deleteNonLatestCustomModels() throws FirebaseMlException {
+    File root = getDirImpl("");
+    // for each custom model sub directory
```
rlazo, line 68: is the comment in the right position?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
```diff
@@ -61,6 +63,24 @@ public static ModelFileManager getInstance() {
...
+
+    boolean ret = true;
+    if (root.isDirectory()) {
+      for (File f : Preconditions.checkNotNull(root.listFiles())) {
```
rlazo, line 72: This will cause the app to crash if listFiles ever returns null. Is that intended?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileDownloadService.java
```diff
@@ -404,7 +409,12 @@ public File loadNewlyDownloadedModelFile(CustomModel model) {
           new CustomModel(
               model.getName(), model.getModelHash(), model.getSize(), 0, newModelFile.getPath()));
 
-      // todo(annzimmer) Cleans up the old files if it is the initial creation.
+      try {
```
rlazo, line 412: Since this is a clean-up task, it could probably be ran as a separate background task, WDYT?

annzimmer: I wasn't sure if I should account for potential (and unlikely) weird threading issues if an app downloads two models almost back to back... Not even sure this would fully address that issue. Happy to change if you think it's the appropriate solution.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
annzimmer: Good catch - adding fix in second location as well.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileManager.java
annzimmer: fixed.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileDownloadService.java
rlazo: Could you give an example of the situation? My mental model of the situation is as follows:
- ml downloads the latest model into a temp location
- once it's downloaded, it gets copied over to the definitive location.
- this code only checks for models in that definitive location

At which point the two downloads would collide?

Also, reading through it more carefully I noticed this: by calling `listFiles()` you get a snapshot of the files available at time X, then you iterate over it one by one, deleting the old models. Since the list of files is static, but the actual files are being deleted during iteration, could it be the case that you try to delete a file that has already being deleted earlier in the loop. Is this an issue?

annzimmer: The listFiles snapshot is actually good. At the top level, only sub-directories are listed (which is the model names) and in the deeper level, all but the current model file are deleted for that model name.

I was concerned about...
Model A is getting downloaded and as the first one to hit clean up starts the deletion process.
Meanwhile, Model B (already downloaded) has a update and starts downloading.

During cleanup triggered by model A, we start the deletion of files associated with model B. We get the current file name (local file) and start deleting, at exactly the same time the update for model B finishes and copies the new file for B into the directory and updated the local file path... so we delete the new file and are now in a bad state.

This is extremely unlikely but I can probably fix if I do based on file # instead of and not passed in local file name.  Let me see how hard that will be.

annzimmer: Okay - check this - it should fix the updated while deleting issue. 
I'm made the clean up into a task, is this the correct way to run it in the background?

rlazo: Thanks for the clarification, for a moment I though that all models were in the same directory (no subdirectories).

Sure thing, that's an approach. Let me know how it goes. Another possibility is to get the list of files and then check the name of the latest one, and then go deleting them by name. That way, if the file is added afterwards, it's not deleted.

annzimmer: Actually this was updated and should work - the file name is the index... that way if a new one is added it's always a higher index. so new files won't be deleted if the sharedPreferences are now slightly behind the temporary file being moved to the correct spot.

annzimmer: PTAL

rlazo (approved):

PR #2355, opened 2021-01-25
Ml logging

annzimmer:


rlazo:
Looks like you're also removing transport from the deps. Is it no longer needed?

annzimmer:
I moved the transport dep to the FirebaseMlLogger - that way I won't have to pass it into every class - just have each class that needed logging get an instance of the FirebaseMlLogger. Hoping that was appropriate.

rlazo:
> I moved the transport dep to the FirebaseMlLogger - that way I won't have to pass it into every class - just have each class that needed logging get an instance of the FirebaseMlLogger. Hoping that was appropriate.

Yes, definitely. Just checking 👍

rlazo (approved):

PR #2359, opened 2021-01-26
Updating `firebase-perf` external documentation as part of Open Sourcing

ramanpreetSinghKhinda:
Preparing `firebase-perf` to be OSS ready. 
  - Updating main README to include `firebase-perf`
  - Updating [firebaseopensource](https://firebaseopensource.com/projects/firebase/firebase-android-sdk/) to include `firebase-perf`

**Note:** Merge only after `firebase-perf` have been Open Sourced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ dependencies:
   * `firebase-inappmessaging-display`
   * `firebase-inappmessaging-display-ktx`
   * `firebase-perf`
+  * `firebase-perf-ktx`
   * `firebase-remote-config`
   * `firebase-remote-config-ktx`
   * `firebase-storage`
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ dependencies:
   * `firebase-inappmessaging-ktx`
   * `firebase-inappmessaging-display`
   * `firebase-inappmessaging-display-ktx`
+  * `firebase-perf`
   * `firebase-remote-config`
   * `firebase-remote-config-ktx`
   * `firebase-storage`
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,7 @@ This repository includes the following Firebase SDKs for Android:
   * `firebase-firestore`
   * `firebase-functions`
   * `firebase-inappmessaging-display`
+  * `firebase-perf`
   * `firebase-storage`
 
 For more information on building the SDKs from source or contributing,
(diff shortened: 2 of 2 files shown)


rlazo (approved):

ramanpreetSinghKhinda:
@rlazo Are there any other public documentation updates required for `fireperf`?

jeremyjiang-dev (commented):
Do we need to update https://github.com/firebase/firebase-android-sdk/blob/master/.opensource/project.json and https://github.com/firebase/firebase-android-sdk/tree/master/docs

rlazo:
> @rlazo Are there any other public documentation updates required for `fireperf`?

The only "public" doc that refers to the actual contents of the repo is the README.md So if that's updated, you're good to go.

ramanpreetSinghKhinda:
> Do we need to update https://github.com/firebase/firebase-android-sdk/blob/master/.opensource/project.json and https://github.com/firebase/firebase-android-sdk/tree/master/docs

Good point Jeremy. Let me update this in this PR.

rlazo:
> Do we need to update https://github.com/firebase/firebase-android-sdk/blob/master/.opensource/project.json and https://github.com/firebase/firebase-android-sdk/tree/master/docs

That's a great question. If anything needs to be changed there is **not** a blocker for OSS. It would be more an oversight from our (Core) part. I'll double check about those files.

rlazo (approved):

README.md
```diff
@@ -22,6 +22,7 @@ dependencies:
   * `firebase-inappmessaging-ktx`
   * `firebase-inappmessaging-display`
   * `firebase-inappmessaging-display-ktx`
+  * `firebase-perf`
```
rachelsaunders, line 25:
```suggestion
  * `firebase-perf`
  * `firebase-perf-ktx`
```

rachelsaunders (dismissed):

docs/README.md
```diff
@@ -12,6 +12,7 @@ This repository includes the following Firebase SDKs for Android:
   * `firebase-firestore`
   * `firebase-functions`
   * `firebase-inappmessaging-display`
+  * `firebase-perf`
```
jeremyjiang-dev, line 15: Do you wanna add another `firebase-perf` to the ktx section below?

jeremyjiang-dev: Never mind. Just saw the comments that we don't need it.

jeremyjiang-dev (approved):

visumickey (approved):

ramanpreetSinghKhinda:
@rlazo Looks like there are some tests which has been failing. I wonder why README changes are triggering tests for those SDKs in the first place.

Also how can we avoid getting blocked on these unrelated transient errors?

jeremyjiang-dev:
I wonder what is the best approach here for some unrelated flaky tests that are failing:
1. Should we ask people to fix them?
2. Should we merge without that test passing if we know that test is flaky by running it locally.

I think rerunning the tests may not help to solve the problem fundamentally.

ramanpreetSinghKhinda:
> I wonder what is the best approach here for some unrelated flaky tests that are failing:
> 
> 1. Should we ask people to fix them?
> 2. Should we merge without that test passing if we know that test is flaky by running it locally.
> 
> I think rerunning the tests may not help to solve the problem fundamentally.

cc @vkryachko @rlazo

PR #2362, opened 2021-01-26
Clean up and clarify Crashlytics logging.

mrwillis21:
Made some log messages clearer and easier to understand, also
reduced noise by changing the log level on existing logs to:

E: Unexpected/error state, not recoverable. Crashlytics will not function.

W: Unexpected/error state, but recoverable.

D: SDK lifecycle logs ("what" is happening).
   - Unix-style "no news is good news" logging, if something goes
     wrong, it will appear in W or E logs.

V: Finer-grained logging ("how" things are happening).
   - "Success" logs, as well as logging of internally-stored values
     (e.g. current mapping file ID) are logged at this level.

Also modified a few logs to *only* print in an error case.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
@@ -92,7 +92,6 @@ public class FirebaseCrashlytics {
 
     if (analyticsConnector != null) {
       // If FA is available, create a logger to log events from the Crashlytics origin.
-      Logger.getLogger().d("Firebase Analytics is available.");
       final CrashlyticsOriginAnalyticsEventLogger directAnalyticsEventLogger =
           new CrashlyticsOriginAnalyticsEventLogger(analyticsConnector);
 
@@ -107,7 +106,7 @@ public class FirebaseCrashlytics {
           subscribeToAnalyticsEvents(analyticsConnector, crashlyticsAnalyticsListener);
 
       if (analyticsConnectorHandle != null) {
-        Logger.getLogger().d("Firebase Analytics listener registered successfully.");
+        Logger.getLogger().d("Registered Firebase Analytics listener.");
         // Create the event receiver which will supply breadcrumb events to Crashlytics
         final BreadcrumbAnalyticsEventReceiver breadcrumbReceiver =
             new BreadcrumbAnalyticsEventReceiver();
@@ -132,7 +131,8 @@ public class FirebaseCrashlytics {
         // Set the blocking analytics event logger for Crashlytics.
         analyticsEventLogger = blockingAnalyticsEventLogger;
       } else {
-        Logger.getLogger().d("Firebase Analytics listener registration failed.");
+        Logger.getLogger()
+            .w("Could not register Firebase Analytics listener; a listener is already registered.");
         // FA is enabled, but the listener was not registered successfully.
         // We cannot listen for breadcrumbs.
         breadcrumbSource = new DisabledBreadcrumbSource();
@@ -142,7 +142,7 @@ public class FirebaseCrashlytics {
       }
     } else {
       // FA is entirely unavailable. We cannot listen for breadcrumbs or send events.
-      Logger.getLogger().d("Firebase Analytics is unavailable.");
+      Logger.getLogger().d("Firebase Analytics is not available.");
       breadcrumbSource = new DisabledBreadcrumbSource();
       analyticsEventLogger = new UnavailableAnalyticsEventLogger();
     }
@@ -161,7 +161,7 @@ public class FirebaseCrashlytics {
 
     final String googleAppId = app.getOptions().getApplicationId();
     final String mappingFileId = CommonUtils.getMappingFileId(context);
-    Logger.getLogger().d("Mapping file ID is: " + mappingFileId);
+    Logger.getLogger().v("Mapping file ID is: " + mappingFileId);
 
     final UnityVersionProvider unityVersionProvider = new ResourceUnityVersionProvider(context);
 
@@ -170,11 +170,11 @@ public class FirebaseCrashlytics {
       appData =
           AppData.create(context, idManager, googleAppId, mappingFileId, unityVersionProvider);
     } catch (PackageManager.NameNotFoundException e) {
-      Logger.getLogger().e("Could not retrieve app info, initialization failed.", e);
+      Logger.getLogger().e("Error retrieving app package info.", e);
       return null;
     }
 
-    Logger.getLogger().d("Installer package name is: " + appData.installerPackageName);
+    Logger.getLogger().v("Installer package name is: " + appData.installerPackageName);
 
     final ExecutorService threadPoolExecutor =
         ExecutorUtils.buildSingleThreadExecutorService("com.google.firebase.crashlytics.startup");
@@ -282,7 +282,7 @@ public class FirebaseCrashlytics {
    */
   public void recordException(@NonNull Throwable throwable) {
     if (throwable == null) { // Users could call this with null despite the annotation.
-      Logger.getLogger().w("Crashlytics is ignoring a request to log a null exception.");
+      Logger.getLogger().v("A null value was passed to recordException. Ignoring.");
       return;
     }
     core.logException(throwable);
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
@@ -161,7 +161,7 @@ public class FirebaseCrashlytics {
 
     final String googleAppId = app.getOptions().getApplicationId();
     final String mappingFileId = CommonUtils.getMappingFileId(context);
-    Logger.getLogger().v("Mapping file ID is: " + mappingFileId);
+    Logger.getLogger().d("Mapping file ID is: " + mappingFileId);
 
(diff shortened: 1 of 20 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -161,7 +161,7 @@
 
     final String googleAppId = app.getOptions().getApplicationId();
     final String mappingFileId = CommonUtils.getMappingFileId(context);
-    Logger.getLogger().d("Mapping file ID is: " + mappingFileId);
```
mrichards, line 164: Strongly suggest this line remain DEBUG since it is critical to many support cases.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -282,7 +282,7 @@ public static FirebaseCrashlytics getInstance() {
    */
   public void recordException(@NonNull Throwable throwable) {
     if (throwable == null) { // Users could call this with null despite the annotation.
-      Logger.getLogger().w("Crashlytics is ignoring a request to log a null exception.");
+      Logger.getLogger().v("A null value was passed to recordException. Ignoring.");
```
mrichards, line 285: This feels more like DEBUG (or WARN) than VERBOSE, since this indicates something unexpected.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
```diff
@@ -51,23 +51,25 @@ public BlockingAnalyticsEventLogger(
   @Override
   public void logEvent(@NonNull String name, @Nullable Bundle params) {
     synchronized (latchLock) {
-      Logger.getLogger().d("Logging Crashlytics event to Firebase");
+      Logger.getLogger()
```
mrichards, line 54: I think this line should remain DEBUG but keep the rest of the changes you made below about the analytics integration. If the message doesn't log successfully, we'll now see a warning or error. If there's nothing else printed to DEBUG, we can assume it succeeded.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
```diff
@@ -526,18 +521,18 @@ boolean finalizeSessions() {
     backgroundWorker.checkRunningOnThread();
 
     if (isHandlingException()) {
-      Logger.getLogger().d("Skipping session finalization because a crash has already occurred.");
+      Logger.getLogger().w("Skipping session finalization because a crash has already occurred.");
       return Boolean.FALSE;
     }
 
-    Logger.getLogger().d("Finalizing previously open sessions.");
+    Logger.getLogger().v("Finalizing previously open sessions.");
```
mrichards, line 528: I think the rest of this file are all great examples of where VERBOSE or WARN makes more sense than DEBUG. Nice!

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
```diff
@@ -782,9 +777,10 @@ File getNativeSessionFilesDir() {
 
   private Task<Void> logAnalyticsAppExceptionEvent(long timestamp) {
     if (firebaseCrashExists()) {
-      Logger.getLogger().d("Skipping logging Crashlytics event to Firebase, FirebaseCrash exists");
+      Logger.getLogger().v("Skipping logging Crashlytics event to Firebase, FirebaseCrash exists");
```
mrichards, line 780: I might keep this as DEBUG just so it is more likely we'll find it in the rare case it causes a problem for a customer. I don't feel super strongly about that though.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsUncaughtExceptionHandler.java
```diff
@@ -57,9 +57,7 @@ public void uncaughtException(Thread thread, Throwable ex) {
       Logger.getLogger().e("An error occurred in the uncaught exception handler", e);
     } finally {
       Logger.getLogger()
-          .d(
-              "Crashlytics completed exception processing."
-                  + " Invoking default exception handler.");
+          .d("Completed exception processing." + " Invoking default exception handler.");
```
mrichards, line 60: You can drop the "+" here, no?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
```diff
@@ -142,7 +142,7 @@ private void logDataCollectionState(boolean dataCollectionEnabled) {
             ? "global Firebase setting"
             : setInManifest ? FIREBASE_CRASHLYTICS_COLLECTION_ENABLED + " manifest flag" : "API";
     Logger.getLogger()
-        .d(
+        .v(
```
mrichards, line 145: This is another one that this is pretty useful for troubleshooting so I'd prefer to keep it as DEBUG.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/settings/network/DefaultSettingsSpiCall.java
```diff
@@ -122,7 +122,7 @@ public JSONObject invoke(SettingsRequest requestData, boolean dataCollectionToke
   /** package private for testing */
   JSONObject handleResponse(HttpResponse httpResponse) {
     final int statusCode = httpResponse.code();
-    logger.d("Settings result was: " + statusCode);
+    logger.v("Settings response code was: " + statusCode);
```
mrichards, line 125: If we're removing this from DEBUG, it is probably useful to add the response code the output below in 131. Something like: `Logger.e(Failed to retrieve settings (response=" + statusCode + ") from " + url);`

mrichards (approved):
I have several comments where I think keeping the output as DEBUG will help with customer troubleshooting. Admittedly, these are based more on my subjective experiences with debugging customer issues than an application of a consistent standard for picking logging levels.

I feel most strongly about the mapping file ID one. I'm approving in advance assuming you'll at least change that one. If you have a major difference of opinion for the other ones, I won't put up an argument.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsUncaughtExceptionHandler.java
mrwillis21: Yeah, this was a formatting thing. I'll fix it.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/settings/network/DefaultSettingsSpiCall.java
mrwillis21: Good call.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
mrwillis21: 👍  after thinking about this a bit more I think you're right.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
mrwillis21: 👍

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
mrwillis21: I'd intended this DEBUG log to cover that use case: https://github.com/firebase/firebase-android-sdk/pull/2362/files#diff-7db92a2657e4a466a2091ebc5233c28b7727d5a5c5121dd95d7e752518586256R783

The verbose log here just gives more details about any events logged through this class.

Thoughts?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
mrwillis21: Might give this one a bit more thought, there might be a way we can make this less noisy and more useful in a DEBUG log.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
mrwillis21: 👍

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/analytics/BlockingAnalyticsEventLogger.java
mrichards: Ah yeah, looks good. 👍

PR #2365, opened 2021-01-27
Clean up unused files and update README in Fireperf.

jeremyjiang-dev:


ramanpreetSinghKhinda:
I think we need proguard file. Isn't it?

https://github.com/FirebasePrivate/firebase-android-release/blob/master/firebase-perf/proguard-rules.pro

ramanpreetSinghKhinda (approved):

visumickey:
@jeremyjiang-dev Where did these files from at the first place? Atleast the crash dump.

visumickey (approved):

jeremyjiang-dev:
> @jeremyjiang-dev Where did these files from at the first place? Atleast the crash dump.

Discussed offline. This file is generated accidentally on my machine and it is fine to remove it.

jeremyjiang-dev:
> I think we need proguard file. Isn't it?
> 
> https://github.com/FirebasePrivate/firebase-android-release/blob/master/firebase-perf/proguard-rules.pro

The content in this file is all commented out. It is fine to remove it.

PR #2366, opened 2021-01-27
Run fireperf e2e test from the public reepo.

yifanyang:
Moving fireci plugin for fireperf end to end test from the release repo to the public repo.

### DIFF
--- a/ci/fireci/fireciplugins/fireperf.py
+++ b/ci/fireci/fireciplugins/fireperf.py
@@ -1,0 +1,85 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import click
+import contextlib
+import logging
+import os
+import pathlib
+
+from fireci import ci_command
+from fireci import gradle
+
+_logger = logging.getLogger('fireci.fireperf')
+
+
+@click.option(
+  '--plugin_repo_dir',
+  help='The location of the fireperf plugin repository.',
+  required=True,
+)
+@click.option(
+  '--target_environment',
+  type=click.Choice(['prod', 'autopush'], case_sensitive=False),
+  help='The target environment fireperf is built for.',
+  required=True,
+)
+@ci_command()
+def fireperf_e2e_test(target_environment, plugin_repo_dir):
+  """Run Firebase Performance end to end test."""
+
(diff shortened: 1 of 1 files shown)


ci/fireci/fireciplugins/fireperf.py
```diff
@@ -0,0 +1,85 @@
...
+def _find_fireperf_plugin_version():
+  local_maven_repo_dir = pathlib.Path.home().joinpath('.m2', 'repository')
+  artifacts_path = local_maven_repo_dir.joinpath('com', 'google', 'firebase', 'perf-plugin')
+  versions = [pom.parent.name for pom in artifacts_path.rglob('*.pom')]
+  snapshots = list(filter(lambda v: v.endswith('-SNAPSHOT'), versions))
+
+  if snapshots:
+    return snapshots[0]
+  else:
+    raise click.ClickException('Cannot find a snapshot version in local maven repo.')
```
ramanpreetSinghKhinda, line 85: I assume every run will be a fresh and clean run so we'll only have one version to pick which is the latest HEAD, right?

ramanpreetSinghKhinda (approved):

ci/fireci/fireciplugins/fireperf.py
yifanyang: Yes, exactly. There will be only one or none, depending on whether `:firebase-performance:performance-gradle:publishToMavenLocal` succeeds.

jeremyjiang-dev (approved):
The next step will be removing the same file in the `FirebasePrivate
/firebase-android-release` repo?

visumickey:
@yifanyang Thanks for making those changes. There are a couple of things I would like to learn a bit about:

1. How can we learn about errors when running the test? Eg. if there are any changes to the SDK that is not compatible with the plugin, it would good to learn them.
2. Where would we configure this script to run the E2E test app? Are there any other config changes required to enable this app?

yifanyang:
> The next step will be removing the same file in the `FirebasePrivate /firebase-android-release` repo?

Yes. It will be done in this [PR](https://github.com/FirebasePrivate/firebase-android-release/pull/834). I'll submit them together.

yifanyang:
> @yifanyang Thanks for making those changes. There are a couple of things I would like to learn a bit about:
> 
> 1. How can we learn about errors when running the test? Eg. if there are any changes to the SDK that is not compatible with the plugin, it would good to learn them.
> 2. Where would we configure this script to run the E2E test app? Are there any other config changes required to enable this e2e test app run?

The status of recent runs for e2e test can be found here:
* (autopush) https://prow-gob.gcpnode.com/job-history/gs/fireescape/logs/periodic-fireperf-e2e-test-autopush/
* (prod) https://prow-gob.gcpnode.com/job-history/gs/fireescape/logs/periodic-fireperf-e2e-test-prod/

We have also set up an alert for these two jobs, which sends out emails in case of failures. @vkryachko Can you add fireperf team as a recipient of the alert?

This script is actually invoked/configured [here](https://team.git.corp.google.com/firebase-android-core/test-infra/+/refs/heads/master/prow/config.yaml) (search `fireperf-e2e-test` in the file). The other config changes required for this migration is https://team-review.git.corp.google.com/c/firebase-android-core/test-infra/+/966612. I saw Raman has already added you as a reviewer though.

PR #2369, opened 2021-01-27
Expose Firestore Bundles

schmidt-sebastian:
WE HAVE REACHED THE END!

This PR adds the API endpoints, the SyncEngine logic and the integration tests for Bundles.

Some notable changes:

I rewrote EventAccumulator to support assertNoAdditionalEvents().
I added totalBytes/totalDocuments back to BundleMetadata as it made plumbing these values through much easier.


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -569,7 +570,9 @@ private void createBundleCache() {
                   + "bundle_id TEXT PRIMARY KEY, "
                   + "create_time_seconds INTEGER, "
                   + "create_time_nanos INTEGER, "
-                  + "schema_version INTEGER)");
+                  + "schema_version INTEGER, "
+                  + "total_documents INTEGER, "
+                  + "total_bytes INTEGER)");
```
wu-hui, line 575: total_documents and total_bytes are only used during loading, they are not of much use after it is loaded, and we should not persist them.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -501,6 +506,20 @@ private void doUnlisten(JSONArray unlistenSpec) throws Exception {
     queue.runSync(() -> eventManager.removeQueryListener(listener));
   }
 
+  private void doLoadBundle(String json) throws Exception {
```
wu-hui, line 509: This confuses me, how did spec tests run before this PR?

wu-hui (changes requested):
Very close to finish!

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleSerializer.java
```diff
@@ -51,7 +51,7 @@
 import org.json.JSONObject;
 
 /** A JSON serializer to deserialize Firestore Bundles. */
-class BundleSerializer {
+public class BundleSerializer {
```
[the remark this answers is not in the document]
schmidt-sebastian: We don't use `createTime` in the Mobile SDKs. The version we use is derived from `updateTime`.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: I have added and removed these fields quite a few times now. Unfortunately, it seems easier to just include them in BundleMetadata, as the combined state can then be passed around in one object. Note that unlike JS, the BundleLoader cannot just grab these values from the JSON stream since they need to be deserialized first. If we want to omit these values from persistence, I think they are only two strategies:

- We create a BundleMetadata and a BundleMetadataDuringLoading and only use BundleMetadata in Storage
- We don't write these values to persistence and use -1 or similar values to populate BundleMetadata when we deserialize

Both of these options seems much less clean to me.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
schmidt-sebastian: They have "no-android" tags and didn't run before they were removed in this PR.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
wu-hui: SG.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
wu-hui: I see, thanks.

wu-hui (approved):
LGTM

PR #2372, opened 2021-01-28
Fix error caused by merging multiple PRs.

annzimmer:


annzimmer:
The last 2 PRs caused an error after merge... it's a simple fix, but everything is failling here with google.auth module missing...

PR #2375, opened 2021-01-29
Iterate forward on local serializer mutation batch decoding

thebrianchen:
Porting from [iOS](https://github.com/firebase/firebase-ios-sdk/pull/7276/files#diff-f6031b0454b2782196a25405bedaafd9083538172c176c17961c709f5118fb2fR338).

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalSerializerTest.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalSerializerTest.java
@@ -204,18 +204,6 @@ public final class LocalSerializerTest {
 
   // TODO(b/174608374): Remove these tests once we perform a schema migration.
   @Test
-  public void testOnlyTransformThrowsError() {
-    WriteBatch batchProto =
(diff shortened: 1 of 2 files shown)


schmidt-sebastian (approved):

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalSerializerTest.java
```diff
@@ -202,18 +202,6 @@ public void testTransformAndTransformThrowError() {
     assertThrows(AssertionError.class, () -> serializer.decodeMutationBatch(batchProto));
   }
 
-  // TODO(b/174608374): Remove these tests once we perform a schema migration.
-  @Test
-  public void testOnlyTransformThrowsError() {
```
schmidt-sebastian, line 207: FWIW You could still assert that the first mutation is not a transform. Optional.

thebrianchen: Eh, I'll pass on this one since I don't want to back-back-backport it to iOS again.

PR #2380, opened 2021-01-29
Add build number to Remote Config

andreaowu:
Allow users to make conditions using build number like they can for iOS apps

Build version is an internal tracking version of an app, not shown to users (in gradle files, this is versionCode)
App version is the external released version of an app shown to users (in gradle files, this is versionName)

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigConstants.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigConstants.java
@@ -44,6 +44,7 @@ public final class RemoteConfigConstants {
     RequestFieldKey.LANGUAGE_CODE,
     RequestFieldKey.PLATFORM_VERSION,
     RequestFieldKey.TIME_ZONE,
+    RequestFieldKey.APP_BUILD,
     RequestFieldKey.APP_VERSION,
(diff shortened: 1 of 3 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigConstants.java
```diff
@@ -59,6 +60,7 @@
     String PLATFORM_VERSION = "platformVersion";
     String TIME_ZONE = "timeZone";
     String APP_VERSION = "appVersion";
+    String APP_BUILD = "appBuild";
```
danasilver, line 63: Looks like a consistent pattern that the [iOS constants use underscores](https://github.com/firebase/firebase-ios-sdk/blob/f3bb73fc8950f4cc6041d3e66ab6b40ee05cc698/FirebaseRemoteConfig/Sources/RCNDevice.m#L29) and Android uses camel case

danasilver (approved):
Looks good - linking the [`PackageInfoCompat#getLongVersionCode`](https://developer.android.com/reference/androidx/core/content/pm/PackageInfoCompat#getLongVersionCode(android.content.pm.PackageInfo)) docs since they explain how that number represents `versionCode` and `versionCodeMajor` should the latter be used as well.

PR #2381, opened 2021-01-29
Add another layer of url expiry checking.

annzimmer:


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/ModelFileDownloadService.java
```diff
@@ -159,7 +159,14 @@ public static ModelFileDownloadService getInstance() {
 
     // schedule new download of model file
     Log.d(TAG, "Need to download a new model.");
-    Long newDownloadId = scheduleModelDownload(customModel);
+    Long newDownloadId = null;
+    try {
+      newDownloadId = scheduleModelDownload(customModel);
+    } catch (FirebaseMlException fex) {
```
rlazo, line 165: If the exception is triggered by a different reason (code) would it be useful to log it?

annzimmer: Good point, also found more appropriate error codes to use and updated testing proto to match.

rlazo (approved):

PR #2383, opened 2021-01-29
Mutable Documents

schmidt-sebastian:
This is yet another big rewrite of the document/mutation logic that aims to provide stronger ownership guarantees for the C++ client.

With this PR, Documents and ObjectValues are now mutable. All mutations directly modify the Document and documents can transition from MaybeDocument to Document, NoDocument and UnknownDocument (and back). I think this is safe because every document lookup starts in the RemoteDocumentCache: In the SQLRemoteDocumentCache we always create new document anyways, and in the MemoryDocumentCache I now clone the documents.

I also merged all existing classes into Document. A Document that is not found in the RemoteDocumentCache is now  `INVALID`, which allows mutations to transition the document to a valid state at a later point.


wilhuff (commented):
In many ways, this is a replay of firebase/firebase-js-sdk#1886. mikelehen argued there that removing the type distinction between `Document` and `MaybeDocument` was harmful and I came to agree. Please review discussion in that PR and let's discuss.

The idea I was proposing w/r/t making mutations operate on a mutable document builder was narrow: within the mutation types, operate on the builder, but then once done applying mutations, convert to the appropriate existing immutable document type afterward. This would significantly limit the blast radius of the change.

schmidt-sebastian:
I have tried to re-introduce the MaybeDocument chain by changing the current document to be a sort of builder for the old types. I am however having a bit of trouble finding a good place to draw the line between where the MutableDocument and the MaybeDocument should be used. The obvious place the LocalDocumentsView. It's input sources would be MutableDocuments (e.g. the RemoteDocumentStore and all mutation logic) and it would return MaybeDocuments. This however falls apart in two places:

- When the LocalStore first writes a new mutation, it uses the LocalDocumentsView to generate the documents that contain the base state. It then passes these MaybeDocuments to the MutationBatch to quickly apply just the new mutation - but the MutationBatch APIs now takes MutableDocuments. This can be solved by going back to the LocalDocumentsView to compute the new state, but that means that we perform some work twice.

- When the LocalStore applies RemoteEvents, it should also ideally take the input as MutableDocuments. If the documents inside of a RemoteEvent are not MutableDocuments, then we would need to convert them before applying mutations. If, however, these documents are MutableDocuments, then all logic in LocalStore `applyRemoteEvent` would have to deal with all this new state and mostly stay as is in this PR. Since that logic really is the one part of our code base that does any of "tricky" stuff with the internal document state, I am not sure if that defeats the purpose of re-introducing MaybeDocuments.

The last argument also applies to drawing the boundary at LocalStore - if RemoteStore, LocalStore, LocalDocumentsView and RemoteDocumentCache all use MutableDocuments, then the remaining part of the code that uses Documents would be fairly minimal - and plus, we would have to run a conversion between MutableDocumentMap and MaybeDocumentMap for every read from LocalStore.

I am probably missing something here, but these are my thoughts from today.

schmidt-sebastian:
Ping :)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/SetMutation.java
```diff
@@ -72,40 +71,32 @@ public String toString() {
...
     // accepted the mutation so the precondition must have held.
-
-    SnapshotVersion version = mutationResult.getVersion();
-
-    ObjectValue newData = value;
     if (mutationResult.getTransformResults() != null) {
-      List<Value> transformResults =
-          serverTransformResults(maybeDoc, mutationResult.getTransformResults());
-      newData = transformObject(newData, transformResults);
+      Map<FieldPath, Value> transformResults =
```
wilhuff, line 80: Elsewhere in this PR you changed `getTransformResults` to be non-nullable. That actually seems unrelated to the core of this change, and introduces the risk that it wouldn't be caught in review. Consider decoupling that into a separate change.

Otherwise, the line above needs to be a !`isEmpty()` test.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/SetMutation.java
```diff
@@ -72,40 +71,32 @@ public String toString() {
...
+          serverTransformResults(document, mutationResult.getTransformResults());
+      ObjectValue localValue = value.clone();
+      localValue.set(transformResults);
+      document.asFoundDocument(mutationResult.getVersion(), localValue).withCommittedMutations();
+    } else {
+      document.asFoundDocument(mutationResult.getVersion(), value.clone()).withCommittedMutations();
```
wilhuff, line 86: You could avoid duplicating most of the arguments to `asFoundDocument` by following the old form of this method:

```java
ObjectValue newData = value.clone();
if (mutationResult.getTransformResults() != null) {
  transformResults = ...;
  newData.set(transformResults);
}

return document.asFoundDcument(mutationResult.getVersion(), newData).withCommittedMutations();
```

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/SetMutation.java
schmidt-sebastian: Separated this out as a stand alone PR: https://github.com/firebase/firebase-android-sdk/pull/2428

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/SetMutation.java
schmidt-sebastian: Done (but some of this is no longer applicable)

schmidt-sebastian (commented):
Feedback addressed (hopefully)

schmidt-sebastian:
I pushed a couple of renamings. This PR now uses setX for all methods that modify the document state. I also renamed Document to MutableDocument. I did experiment with adding a DocumentBuilder or keeping Document separate from MutableDocunent, but it leads to some strange APIs in the RemoteDocumentCache  as it would store Documents but return Builders. Let me know what you think.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -93,33 +93,25 @@ public Precondition getPrecondition() {
   }
 
   /**
-   * Applies this mutation to the given MaybeDocument for the purposes of computing a new remote
-   * document. If the input document doesn't match the expected state (e.g. it is null or outdated),
-   * an `UnknownDocument` can be returned.
+   * Applies this mutation to the given Document for the purposes of computing a new remote document
+   * If the input document doesn't match the expected state (e.g. it is null or outdated), an
+   * `UnknownDocument` may be created
```
dconeybe, line 98: "may be created" is somewhat misleading, since the `MutableDocument` is modified in place. Please update.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -93,33 +93,25 @@ public Precondition getPrecondition() {
...
-  public abstract MaybeDocument applyToRemoteDocument(
-      @Nullable MaybeDocument maybeDoc, MutationResult mutationResult);
+  public abstract void applyToRemoteDocument(
+      MutableDocument document, MutationResult mutationResult);
 
   /**
-   * Applies this mutation to the given MaybeDocument for the purposes of computing the new local
-   * view of a document. Both the input and returned documents can be null.
+   * Applies this mutation to the given Document for the purposes of computing the new local view of
+   * a document. Both the input and returned documents can be null.
```
dconeybe, line 108: "Both the input and returned documents can be null" is now somewhat incorrect since the method now returns null. Please update.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -166,9 +156,9 @@ static SnapshotVersion getPostMutationVersion(@Nullable MaybeDocument maybeDoc)
    * @param serverTransformResults The transform results received by the server.
    * @return The transform results list.
    */
-  protected List<Value> serverTransformResults(
-      @Nullable MaybeDocument maybeDoc, List<Value> serverTransformResults) {
-    ArrayList<Value> transformResults = new ArrayList<>(fieldTransforms.size());
+  protected Map<FieldPath, Value> serverTransformResults(
+      MutableDocument maybeDoc, List<Value> serverTransformResults) {
```
dconeybe, line 160: Please rename `maybeDoc`. Also on line 193.

dconeybe (commented):
My review is very much incomplete, but I thought I'd post the comments I've left so far on 7 of the 92 files in this review :) Most of my comments concern naming and vestigial documentation that needs updating.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -218,18 +203,21 @@ private MaybeDocument getDocument(DocumentKey key, List<MutationBatch> inBatches
...
         DocumentKey key = mutation.getKey();
-        MaybeDocument baseDoc = results.get(key);
-        MaybeDocument mutatedDoc = mutation.applyToLocalView(baseDoc, batch.getLocalWriteTime());
-        if (mutatedDoc instanceof Document) {
-          results = results.insert(key, (Document) mutatedDoc);
-        } else {
+        MutableDocument document = results.get(key);
+        if (document == null) {
+          document =
+              new MutableDocument(key); // Create invalid document to apply mutations on top of
```
dconeybe, line 209: ultranit: Consider moving the comment "// Create invalid document..." onto its own line so that the assignment to `document` can also happen on its own line instead of wrapping around.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
```diff
@@ -48,30 +45,27 @@ public LocalSerializer(RemoteSerializer rpcSerializer) {
...
-      builder.setHasCommittedMutations(true);
+    if (document.isNoDocument()) {
+      builder.setNoDocument(encodeNoDocument(document));
+    } else if (document.isFoundDocument()) {
+      builder.setDocument(encodeDocument(document));
+    } else if (document.isUnknownDocument()) {
+      builder.setUnknownDocument(encodeUnknownDocument(document));
     } else {
-      throw fail("Unknown document type %s", document.getClass().getCanonicalName());
+      throw fail("Unknown Document %s", document);
```
dconeybe, line 59: Consider rewording this error message since the concept of an "unknown document" is already taken, and would have caused the "else if" branch above to be entered had the document been "unknown".

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -59,75 +57,67 @@ public void remove(DocumentKey key) {
...
-      MaybeDocument maybeDoc = entry.getValue().first;
-      if (!(maybeDoc instanceof Document)) {
-        continue;
-      }
-
-      SnapshotVersion readTime = entry.getValue().second;
-      if (readTime.compareTo(sinceReadTime) <= 0) {
-        continue;
-      }
+      MutableDocument doc = entry.getValue().first;
```
dconeybe, line 98: Personally, I prefer the "guard" style from the original code. The new code mixes an "if" statement and a guard, which is arguably more confusing. For readability, could you switch to exclusively use the "guard" style?

i.e.

```java
while (iterator.hasNext()) {
  Map.Entry<DocumentKey, Pair<MutableDocument, SnapshotVersion>> entry = iterator.next();

  DocumentKey key = entry.getKey();
  if (!queryPath.isPrefixOf(key.getPath())) {
    break;
  }

  MutableDocument doc = entry.getValue().first;
  if (! doc.isFoundDocument()) {
    continue;
  }
  
  SnapshotVersion readTime = entry.getValue().second;
  if (readTime.compareTo(sinceReadTime) <= 0) {
    continue;
  }

  if (!query.matches(doc)) {
    continue;
  }
  
  result = result.insert(doc.getKey(), doc.clone());
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/RemoteDocumentCache.java
```diff
@@ -37,10 +35,10 @@
    * <p>The cache key is extracted from {@code maybeDocument.key}. If there is already a cache entry
    * for the key, it will be replaced.
    *
-   * @param maybeDocument A Document or NoDocument to put in the cache.
+   * @param document A Document or NoDocument to put in the cache.
```
dconeybe, line 38: nit: The mention of `NoDocument` here is obsolete. Also, `Document` should not begin with an uppercase letter because it is no longer the name of a class. Also on lines 50 and 58.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/RemoteDocumentCache.java
```diff
@@ -49,19 +47,18 @@
...
+  MutableDocument get(DocumentKey documentKey);
 
   /**
    * Looks up a set of entries in the cache.
    *
    * @param documentKeys The keys of the entries to look up.
-   * @return The cached Document or NoDocument entries indexed by key. If an entry is not cached,
-   *     the corresponding key will be mapped to a null value.
+   * @return The cached Document or NoDocument entries indexed by key. If an entry is not cached, an
+   *     invalid document is returned.
```
dconeybe, line 59: This statement that an "invalid document is returned" is incorrect because the method returns a *Map*, not a *document*. The original comment can probably be tweaked, something like "If an entry is not cached, the corresponding key will be mapped to an invalid document."

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationBatch.java
```diff
@@ -76,19 +76,16 @@ public MutationBatch(
...
-          maybeDoc.getKey().equals(documentKey),
-          "applyToRemoteDocument: key %s doesn't match maybeDoc key %s",
-          documentKey,
-          maybeDoc.getKey());
-    }
+  public void applyToRemoteDocument(
+      DocumentKey documentKey, MutableDocument document, MutationBatchResult batchResult) {
+    hardAssert(
+        document.getKey().equals(documentKey),
+        "applyToRemoteDocument: key %s doesn't match maybeDoc key %s",
```
dconeybe, line 86: nit: fix this outdated mention of "maybeDoc" in this assertion message.

dconeybe (commented):
Here is another round of feedback. My brain has turned to mush so I'll resume, and hopefully finish, my first pass of the review tomorrow.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
```diff
@@ -0,0 +1,206 @@
...
+    if (hasCommittedMutations != document.hasCommittedMutations) return false;
+    if (!key.equals(document.key)) return false;
+    if (!version.equals(document.version)) return false;
+    if (type != document.type) return false;
+    return value.equals(document.value);
+  }
+
+  @Override
+  public int hashCode() {
+    return key.hashCode();
```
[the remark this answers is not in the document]
schmidt-sebastian: Uff. The risk is absolutely there, especially as we further modify this code in the coming years. I spend the last couple of days trying to see how we could mitigate this risk without creating additional document copies (that would defeat the purpose of the PR a bit). In the end, I added a `Document` interface that provides getters but no setters. These Documents are now used in Views. While this does not 100% remove the risk as the documents can still be mutated in other parts of the code, the vast majority of the code base is now shielded from regressions. Let me know what you think.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationBatch.java
schmidt-sebastian: Fixed

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -162,13 +154,13 @@ static SnapshotVersion getPostMutationVersion(@Nullable MaybeDocument maybeDoc)
    * result of applying a transform) for use after a mutation containing transforms has been
    * acknowledged by the server.
    *
-   * @param maybeDoc The current state of the document after applying all previous mutations.
+   * @param mutableDocument The current state of the document after applying all previous mutations.
    * @param serverTransformResults The transform results received by the server.
    * @return The transform results list.
```
dconeybe, line 159: ultranit: "list" is no longer accurate, since the return type has changed to a map. Please update. Also on line 192.

dconeybe (approved):
Overall this LGTM. My comments about naming are "optional" since coming up with the perfect names is proving to be challenging and probably not worth a bunch of back and forth. Moreover, we can always change then names later.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
```diff
@@ -0,0 +1,225 @@
...
+
+import androidx.annotation.NonNull;
+import com.google.firestore.v1.Value;
+
+public final class MutableDocument implements Document, Cloneable {
+
+  private enum DocumentType {
+    /**
+     * Represents the initial state of a MutableDocument when only the document key is known.
+     * Invalid documents transition to other states as mutations are applied. If a document remain
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

firebase-firestore/ktx/src/test/java/com/google/firebase/firestore/testutil/TestUtil.java
```diff
@@ -62,28 +62,22 @@ public static SnapshotVersion version(long versionMicros) {
     return new SnapshotVersion(new Timestamp(seconds, nanos));
   }
 
-  public static Document doc(String key, long version, Map<String, Object> data) {
-    return new Document(
-        key(key), version(version), wrapObject(data), Document.DocumentState.SYNCED);
+  public static MutableDocument doc(String key, long version, Map<String, Object> data) {
+    return new MutableDocument(key(key)).setFoundDocument(version(version), wrapObject(data));
```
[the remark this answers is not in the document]
schmidt-sebastian: I did - but in our other TestUtil (https://github.com/firebase/firebase-android-sdk/blob/c61f53c66ffb72197ad14fec9ad0590aad2960aa/firebase-firestore/src/testUtil/java/com/google/firebase/firestore/testutil/TestUtil.java). I will add it here as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -74,38 +71,31 @@ IndexManager getIndexManager() {
   /**
    * Returns the the local view of the document identified by {@code key}.
    *
-   * @return Local view of the document or null if we don't have any cached state for it.
+   * @return Local view of the document or a null if we don't have any cached state for it.
```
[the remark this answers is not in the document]
schmidt-sebastian: It does not, it returns an invalid document instead of `null`. Fixed.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -74,38 +71,31 @@ IndexManager getIndexManager() {
...
+    MutableDocument document = remoteDocumentCache.get(key);
     for (MutationBatch batch : inBatches) {
-      document = batch.applyToLocalView(key, document);
+      batch.applyToLocalView(document);
     }
-
     return document;
   }
 
   // Returns the view of the given {@code docs} as they would appear after applying all mutations in
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed. Thanks for catching this as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
schmidt-sebastian: Fixed (here and below)

schmidt-sebastian (commented):
Thanks for the review! Next one will be at least 10x as long :) (okay, maybe not)

PR #2384, opened 2021-01-30
Update getModelDownloadId based on API review.

annzimmer:


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
```diff
@@ -488,12 +488,26 @@ public void setStatsCollectionEnabled(boolean enabled) {
...
+    if (getModelTask == null) {
+      CustomModel localModel = sharedPreferencesUtil.getDownloadingCustomModelDetails(modelName);
+      if (localModel != null) {
+        return Tasks.forResult(localModel.getDownloadId());
+      }
+      return Tasks.forResult(0L);
+    }
+
+    long downloadId = 0;
+    while (downloadId == 0 && !getModelTask.isComplete()) {
```
rlazo, line 503: Since getModelTask is a task, would it be better to model this as `.addOnCompleteListener` or `.onComplete` ? the while would keep checking constantly.

annzimmer: I need it to return when either downloadId is not 0 or the task completes... can I do that with a listener?
Basically in most cases where a download is happening we'll get the downloadId quickly (we don't wait for the task to complete here), but in a few cases no download will be triggered and we will return 0 when the task is completed (usually very quickly).

rlazo: oh I see, I wasn't considering that scenario. Thanks for the clarification!

rlazo (approved):

PR #2394, opened 2021-02-01
Implement message parsing.

vkryachko:
The parser processes input `FileDescriptorProto`s from protoc into a
type graph that is later fed into `CodeGenerator`.

### DIFF
--- a/encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
+++ b/encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
@@ -44,24 +44,24 @@ interface DescriptorParser {
  */
 class DefaultParser @Inject constructor(private val config: CodeGenConfig) : DescriptorParser {
     override fun parse(files: List<FileDescriptorProto>): List<UserDefined> {
-        val parsedMessages = files.asSequence().flatMap { file ->
+        val parsedTypes = files.asSequence().flatMap { file ->
             val javaPackage = "${config.vendorPackage}.${file.`package`}"
 
             val parent = Owner.Package(file.`package`, javaPackage, file.name)
             parseEnums(parent, file.enumTypeList).plus(parseMessages(parent, file.messageTypeList))
-        }.toList()
+        }
 
         val extensions = discoverExtensions(files)
 
-        val messagesWithExtensions = parsedMessages.map { msg ->
-            val msgExtensions = extensions[msg.protobufFullName]
-            if (msg !is Message || msgExtensions == null) {
-                return@map msg
+        val typesWithExtensions = parsedTypes.map { type ->
+            val msgExtensions = extensions[type.protobufFullName]
+            if (type !is Message || msgExtensions == null) {
+                return@map type
             }
-            msg.withMoreFields(msgExtensions)
-        }
-        resolveReferences(messagesWithExtensions)
-        return messagesWithExtensions.filter {
+            type.withMoreFields(msgExtensions)
+        }.toList()
+        resolveReferences(typesWithExtensions)
+        return typesWithExtensions.filter {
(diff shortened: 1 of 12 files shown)


encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
```diff
@@ -27,9 +33,128 @@ interface DescriptorParser {
...
 class DefaultParser @Inject constructor(private val config: CodeGenConfig) : DescriptorParser {
-    override fun parse(protoFiles: List<FileDescriptorProto>): Collection<UserDefined> {
-        return listOf()
+    override fun parse(files: List<FileDescriptorProto>): List<UserDefined> {
+        val parsedMessages = files.asSequence().flatMap { file ->
+            val javaPackage = "${config.vendorPackage}.${file.`package`}"
+
+            val parent = Owner.Package(file.`package`, javaPackage, file.name)
+            parseEnums(parent, file.enumTypeList).plus(parseMessages(parent, file.messageTypeList))
+        }.toList()
```
rlazo, line 52: do you need to convert it to list first?

encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
```diff
@@ -27,9 +33,128 @@ interface DescriptorParser {
...
+
+        val extensions = discoverExtensions(files)
+
+        val messagesWithExtensions = parsedMessages.map { msg ->
+            val msgExtensions = extensions[msg.protobufFullName]
+            if (msg !is Message || msgExtensions == null) {
```
rlazo, line 58: if msg isn't Message, wouldn't the protobufFullName above fail before reaching this point? Also

encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
```diff
@@ -27,9 +33,128 @@ interface DescriptorParser {
...
+                        field.determineType(),
+                        field.number,
+                        field.label == FieldDescriptorProto.Label.LABEL_REPEATED))
+            }
+        }
+        return extensions.build()
+    }
+
+    private fun parseEnums(parent: Owner, enums: List<DescriptorProtos.EnumDescriptorProto>): Sequence<ProtoEnum> {
+        return enums.asSequence().map { enum ->
```
rlazo, line 107: curiosity, why don't you need to escape enum as \`enum\`?

vkryachko: apparently Kotlin calls it a [Modifier keyword](https://kotlinlang.org/docs/reference/keyword-reference.html#modifier-keywords) which can be used as an identifier as well.

encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
vkryachko: good point, removed

encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Parsing.kt
vkryachko: It wouldn't fail, because if it's not a `Message` then it's a `ProtoEnum` which also has the `protobufFullName` property.

I renamed the variables to make it more clear

rlazo (approved):

PR #2401, opened 2021-02-02
code deduplication

jmwski:


firebase-database/src/main/java/com/google/firebase/database/core/SyncTree.java
```diff
@@ -551,60 +591,16 @@ public Node getServerValue(QuerySpec query) {
...
+            CacheNode serverCache = result.node();
+            if (serverCache == null) {
               // Hit persistence
               CacheNode persistentServerCache = persistenceManager.serverCache(query);
               if (persistentServerCache.isFullyInitialized()) {
                 serverCache = persistentServerCache;
               } else {
-                serverCacheNode = EmptyNode.Empty();
-                ImmutableTree<SyncPoint> subtree = syncPointTree.subtree(path);
+                Node serverCacheNode = EmptyNode.Empty();
```
aimiebell, line 602: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

Issue #2107, opened 2020-10-27
Firebase In-App Messaging errors


ArcherEmiya05:
We are receiving tons of error from Crashlytics and Cloud Console with In-App Messaging dependency and as I checked other report issue with In-App Messaging it seems no fix action has been taken. This module definitely need some revamp.

![image](https://user-images.githubusercontent.com/38008900/97276690-82849f00-1872-11eb-868a-14b60bb44d08.png)
![image](https://user-images.githubusercontent.com/38008900/97276759-96300580-1872-11eb-895f-5fa4f2188033.png)
![image](https://user-images.githubusercontent.com/38008900/97276922-cd061b80-1872-11eb-9d0b-d1633848f3cc.png)
![image](https://user-images.githubusercontent.com/38008900/97276967-d8f1dd80-1872-11eb-986c-0f0fedf4ce6a.png)
![image](https://user-images.githubusercontent.com/38008900/97277028-e909bd00-1872-11eb-93e9-9807c2ba4b18.png)
![image](https://user-images.githubusercontent.com/38008900/97277044-f030cb00-1872-11eb-9b47-2e54d985faab.png)
![image](https://user-images.githubusercontent.com/38008900/97277081-0179d780-1873-11eb-8121-6e5dadef7aca.png)
![image](https://user-images.githubusercontent.com/38008900/97277119-0dfe3000-1873-11eb-8234-cf73c50976f4.png)

rlazo:
Hi @ArcherEmiya05 what Firebase SDKs are you using? Which versions?  Thanks!

prakhar1989:
Hey @ArcherEmiya05 thanks for bringing this to our attention. Can you please share what SDK version are you seeing the crash on? Also from your crashlytics report, can you confirm what if there's a particular set of devices and/or Android OS version that this is prevalent on? It'll help us narrow down the issue faster.

Thanks!

eldhosembabu:
Hi @ArcherEmiya05 ,

Thanks for reporting this issue.

The issue is already fixed in the latest in-app-messaging version 19.1.4.

Please use that version and let us know whether its reproducible.

PR #2402, opened 2021-02-02
Fixing BadTokenException while showing FIAM

eldhosembabu:
Fixing BadTokenException while showing FIAM.

Related Bug : https://github.com/firebase/firebase-android-sdk/issues/2107

This is a possible fix as the BadTokenException usually happens when we try to add/show new views with a finishing activity context.


firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/internal/FiamWindowManager.java
```diff
@@ -54,6 +54,11 @@ public void show(@NonNull final BindingWrapper bindingWrapper, @NonNull Activity
       return;
     }
 
+    if(activity.isFinishing()){
+      Logging.loge("Activity is finishing or not having valid window token. Cannot show Fiam");
```
JasonAHeron, line 58: Activity is finishing or does not have valid window token. Cannot show FIAM

JasonAHeron (approved):

firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/internal/FiamWindowManager.java
eldhosembabu: Done

Issue #2311, opened 2021-01-11
Firestore .get .(addOnCompleteListener) not receiveing response in firestore sdk 21.4.3 or newer


markic22:
- Android Studio version: **4.1.1**
- Firebase Component: **Firestore**
- Component version: **21.4.3 or higher**

#### Steps to reproduce:

I'm having issue with firestore Android sdk. We were using firestore sdk 22.0.0 but we found an issue with our application, that in some cases, data wouldn't load successfully. After investigation, we figured out that the issue was happening when user had closed application and when there was big number of changes happening, first try to load the data was not successful. If you closed and opened application again (called .addOnCompleteListener again) the data was loaded ok, so the issue is only when you call listener for the first time since there were a lot of changes made to the documents. The strange thing is that in this .get() none of the listeners got a response (neither addOnCompleteListener, addOnFailureListener or addOnCanceledListener). I tried to wait for hours (8 documents in this collection collection, but in other collections there were hundreds of changes if not thousands).

`   private fun getAssignmentConsumptions() {
    Log.d("firestoreService", "firestoreService: getAssignmentConsumptions, started")
    getFacilityDocumentRef(facilityId).collection(AssignmentConsumption.COLLECTION).get().addOnCompleteListener { snapshots ->
            assignmentConsumptionsList = snapshots.result?.documents?.toModels<AssignmentConsumption>()?.toMutableList() ?: mutableListOf()
            Log.d("firestoreService", "firestoreService: getAssignmentConsumptions, done")
            assignmentConsumptionsListBehaviourRelay.accept(assignmentConsumptionsList.toList())
    }.addOnFailureListener {
        Log.d("firestoreService", "firestoreService: getAssignmentConsumptions, onFailure")
    }.addOnCanceledListener {
        Log.d("firestoreService", "firestoreService: getAssignmentConsumptions, onCanceled")
    }
}`

Then I tried to use firestore 19.0.0 version. And it worked as it should. Then I decided that I would figure out when was this "issue" introduced into firestore project. The last version that works OK for us is the 21.4.2 version, anything above this version, this .get() call gets 0 response.

In documentation for version 21.4.3 it says this:

> Cloud Firestore version 21.4.3 Cloud Firestore now limits the number of concurrent document lookups it will perform when resolving inconsistencies in the local cache.

As we can see there were some changes done for local cache, but it seems that this broke loading/resolving local data with new changes, where we don't get response about what is happening.

[...]

PR #2404, opened 2021-02-03
Dequeue limbo resolutions when their respective queries are stopped

dconeybe:
Fix a bug where enqueued limbo resolutions are left in the queue even after all targets that care about their resolutions are stopped. This erroneous behavior was reported in https://github.com/firebase/firebase-android-sdk/issues/2311 against the Android SDK, but the bug is also present in the Web and iOS SDKs. This PR is a port of the equivalent fix in the Web SDK: https://github.com/firebase/firebase-js-sdk/pull/4395. This bug was introduced when limbo resolution throttling was implemented almost a year ago (https://github.com/firebase/firebase-js-sdk/pull/2790).

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -4,9 +4,9 @@
   Bundles contain pre-packaged data produced with the NodeJS Server SDK and
   can be used to populate Firestore's cache without reading documents from
   the backend.
-- [fixed] Fix a Firestore bug where local cache inconsistencies were
-  unnecessarily being resolved, resulting in the incompletion of `Task` objects
-  returned from `get()` invocations (#2404).
+- [fixed] Fixed a Firestore bug where local cache inconsistencies were
+  unnecessarily being resolved, causing the `Task` objects returned from `get()`
+  invocations to never complete (#2404).
 
 # (22.0.2)
 - [changed] A write to a document that contains FieldValue transforms is no
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -6,7 +6,7 @@
   the backend.
 - [fixed] Fix a Firestore bug where local cache inconsistencies were
   unnecessarily being resolved, resulting in the incompletion of `Task` objects
-  returned from `get()` invocations.
+  returned from `get()` invocations (#2404).
 
 # (22.0.2)
 - [changed] A write to a document that contains FieldValue transforms is no
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -4,6 +4,9 @@
   Bundles contain pre-packaged data produced with the NodeJS Server SDK and
   can be used to populate Firestore's cache without reading documents from
   the backend.
+- [fixed] Fix a Firestore bug where local cache inconsistencies were
+  unnecessarily being resolved, resulting in the incompletion of `Task` objects
+  returned from `get()` invocations.
 
 # (22.0.2)
 - [changed] A write to a document that contains FieldValue transforms is no
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
@@ -696,8 +696,9 @@ public class SyncEngine implements RemoteStore.RemoteStoreCallback {
   private void pumpEnqueuedLimboResolutions() {
     while (!enqueuedLimboResolutions.isEmpty()
         && activeLimboTargetsByKey.size() < maxConcurrentLimboResolutions) {
-      DocumentKey key = enqueuedLimboResolutions.iterator().next();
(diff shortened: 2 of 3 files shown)


firebase-firestore/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
   Bundles contain pre-packaged data produced with the NodeJS Server SDK and
   can be used to populate Firestore's cache without reading documents from
   the backend.
+- [fixed] Fix a Firestore bug where local cache inconsistencies were
+  unnecessarily being resolved, resulting in the incompletion of `Task` objects
```
schmidt-sebastian, line 8: s/incompletion/something that is easier to parse

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -694,7 +696,8 @@ private void trackLimboChange(LimboDocumentChange change) {
   private void pumpEnqueuedLimboResolutions() {
     while (!enqueuedLimboResolutions.isEmpty()
         && activeLimboTargetsByKey.size() < maxConcurrentLimboResolutions) {
-      DocumentKey key = enqueuedLimboResolutions.remove();
+      DocumentKey key = enqueuedLimboResolutions.iterator().next();
+      enqueuedLimboResolutions.remove(key);
```
schmidt-sebastian, line 700:
```suggestion
      Iterator<DocumentKey> it = enqueuedLimboResolutions.iterator();
      DocumentKey key = it.next();
      it.remove();
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -21,6 +21,8 @@
 import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
```
schmidt-sebastian, line 25: One of our goals is to eventually kick out Guava from our dependencies. Can you find another solution?

firebase-firestore/CHANGELOG.md
dconeybe: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
dconeybe: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -708,15 +710,15 @@ private void pumpEnqueuedLimboResolutions() {
...
     // Make a defensive copy as the Map continues to be modified.
-    return new HashMap<>(activeLimboTargetsByKey);
+    return new HashMap(activeLimboTargetsByKey);
   }
 
   @VisibleForTesting
-  public Queue<DocumentKey> getEnqueuedLimboDocumentResolutions() {
-    // Make a defensive copy as the Queue continues to be modified.
-    return new ArrayDeque<>(enqueuedLimboResolutions);
+  public LinkedHashSet<DocumentKey> getEnqueuedLimboDocumentResolutions() {
```
schmidt-sebastian, line 719: Super nit (here and above): We usually try to use more generic types in our return types, as this allows us to change the implementation without changing the callsites. If the callsites don't require the `HashMap` or `LinkedHashSet` functionality, I would just use `Map<>` and `Set<>`. 

Please note that this might not apply if `getEnqueuedLimboDocumentResolutions()` only uses the Set API but its consumers rely on the iteration order of a LinkedHashSet. Then I would still return `LinkedHashSet` as we will not be able to replace it later. You should still generify the signature in  line 713 though.

schmidt-sebastian (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
```diff
@@ -694,7 +694,9 @@ private void trackLimboChange(LimboDocumentChange change) {
   private void pumpEnqueuedLimboResolutions() {
     while (!enqueuedLimboResolutions.isEmpty()
         && activeLimboTargetsByKey.size() < maxConcurrentLimboResolutions) {
-      DocumentKey key = enqueuedLimboResolutions.remove();
+      Iterator<DocumentKey> it = enqueuedLimboResolutions.iterator();
```
schmidt-sebastian, line 697: Looks like you need to import `Iterator`.

dconeybe: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/SyncEngine.java
dconeybe: Done.

PR #2406, opened 2021-02-03
Add DI to FirebasePerfRegistrar.java

visumickey:


firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -132,13 +133,17 @@ public static FirebasePerformance getInstance() {
...
-      FirebaseInstallationsApi firebaseInstallationsApi,
-      Provider<TransportFactory> transportFactoryProvider) {
-
+  FirebaseApp firebaseApp;
+  Provider<RemoteConfigComponent> firebaseRemoteConfigProvider;
+  FirebaseInstallationsApi firebaseInstallationsApi;
+  Provider<TransportFactory> transportFactoryProvider;
+
+  @Inject
+  public FirebasePerformance(
```
visumickey, line 142: This need not be Public.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -62,4 +57,18 @@
             .build(),
         LibraryVersionComponent.create("fire-perf", BuildConfig.VERSION_NAME));
   }
+
+  public FirebasePerformance providesFirebasePerformance(ComponentContainer container) {
```
jposuna, line 61: Should this be static? (If yes, please update line 51 to not use "this")

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -132,13 +133,17 @@ public static FirebasePerformance getInstance() {
...
-  /** This constructor is invoked from {@link com.google.firebase.perf.FirebasePerfRegistrar}. */
-  FirebasePerformance(
-      FirebaseApp firebaseApp,
-      Provider<RemoteConfigComponent> firebaseRemoteConfigProvider,
-      FirebaseInstallationsApi firebaseInstallationsApi,
-      Provider<TransportFactory> transportFactoryProvider) {
+  FirebaseApp firebaseApp;
+  Provider<RemoteConfigComponent> firebaseRemoteConfigProvider;
+  FirebaseInstallationsApi firebaseInstallationsApi;
+  Provider<TransportFactory> transportFactoryProvider;
```
jposuna, line 139: Can these be final?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -147,6 +152,10 @@ public static FirebasePerformance getInstance() {
         RemoteConfigManager.getInstance(),
         ConfigResolver.getInstance(),
         GaugeManager.getInstance());
```
jposuna, line 154: Do we still need the this(...)?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformanceModule.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.firebase.remoteconfig.RemoteConfigComponent;
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+public class FirebasePerformanceModule {
+  private final FirebaseApp firebaseApp;
+  private final FirebaseInstallationsApi firebaseInstallations;
+  private Provider<RemoteConfigComponent> remoteConfigComponentProvider;
+  private Provider<TransportFactory> transportFactoryProvider;
```
jposuna, line 31: Can these be final as well?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jposuna: Potentially also private

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
visumickey: Yes, it should be and private as well.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: For now yes, I'm working on figuring out a way to get this removed as well.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformanceModule.java
visumickey: Done!

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: We are trying to unify to have a common constructor. Currently there are 2 separate constructors, one for testing and one for the registrar which should ideally be unified.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: Made it private. Will move it to final as well soon!

firebase-perf/firebase-perf.gradle
```diff
@@ -109,6 +109,8 @@ dependencies {
     // 3rd Party Deps
     implementation 'com.squareup.okhttp3:okhttp:3.12.1'
     compileOnly group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.6'
+    implementation 'com.google.dagger:dagger:2.27'
```
blundell, line 112: doesn't this mean anyone depending on this library has to then pull in dagger?

What if they don't use dagger for their DI?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: Done!

firebase-perf/firebase-perf.gradle
visumickey: Apparently the underlying dependency of Firebase Performance (Google Transport) already uses Dagger. So, this should not come as a surprise to the developers using Firebase Performance. Hopefully the cost is well justified for a better quality predictable SDK.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jposuna: Just checking for the "final" bit - is it something coming in a different PR?

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: This is done now.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
```diff
@@ -0,0 +1,94 @@
...
+ * @hide
+ */
+@Module
+public class FirebasePerformanceModule {
+  private final FirebaseApp firebaseApp;
+  private final FirebaseInstallationsApi firebaseInstallations;
+  private final Provider<RemoteConfigComponent> remoteConfigComponentProvider;
+  private final Provider<TransportFactory> transportFactoryProvider;
+
+  //  private final RemoteConfigManager remoteConfigManager;
```
yingdai3, line 41: Remove these lines?

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
```diff
@@ -0,0 +1,94 @@
...
+      @NonNull FirebaseApp firebaseApp,
+      @NonNull FirebaseInstallationsApi firebaseInstallations,
+      @NonNull Provider<RemoteConfigComponent> remoteConfigComponentProvider,
+      @NonNull Provider<TransportFactory> transportFactoryProvider) {
+    this.firebaseApp = firebaseApp;
+    this.firebaseInstallations = firebaseInstallations;
+    this.remoteConfigComponentProvider = remoteConfigComponentProvider;
+    this.transportFactoryProvider = transportFactoryProvider;
+
+    //    this.remoteConfigManager = RemoteConfigManager.getInstance();
```
yingdai3, line 55: Same here.

yingdai3 (changes requested):

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.firebase.installations.FirebaseInstallationsApi;
+import com.google.firebase.perf.config.ConfigResolver;
+import com.google.firebase.perf.internal.GaugeManager;
+import com.google.firebase.perf.internal.RemoteConfigManager;
+import com.google.firebase.remoteconfig.RemoteConfigComponent;
+import dagger.Module;
+import dagger.Provides;
+
+/**
+ * Provider for FirebasePerformance
```
ramanpreetSinghKhinda, line 30: nit: `{@link FirebasePerformance}.`

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
```diff
@@ -0,0 +1,31 @@
...
+
+import androidx.annotation.NonNull;
+import com.google.firebase.perf.FirebasePerformance;
+import com.google.firebase.perf.injection.modules.FirebasePerformanceModule;
+import dagger.Component;
+
+/**
+ * Dagger component to create FirebasePerformanceComponent Objects.
+ *
+ * @hide
```
ramanpreetSinghKhinda, line 25: I know we have another bug to filter out the use-case for `@hide` annotations but do we need to add this for new APIs/Classes that we are writing?

FYI - https://github.com/firebase/firebase-android-sdk/tree/master#annotations

---

_If we want to do it later across project that is fine too. In that case, please ignore._

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Provides
+  ConfigResolver providesConfigResolver() {
+    return ConfigResolver.getInstance();
+  }
+
+  @Provides
+  GaugeManager providesGaugeManager() {
+    return GaugeManager.getInstance();
+  }
```
ramanpreetSinghKhinda, line 93: Should these APIs (for initializing `RemoteConfigManager`, `ConfigResolver`, `GaugeManager`) be part of `FirebasePerformanceModule`?

Also those objects are not initialized in Daggery way but I assume that's what the goal for this week.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
```diff
@@ -0,0 +1,94 @@
...
+
+    //    this.remoteConfigManager = RemoteConfigManager.getInstance();
+    //    this.configResolver = ConfigResolver.getInstance();
+    //    this.gaugeManager = GaugeManager.getInstance();
+  }
+
+  @Provides
+  FirebaseApp providesFirebaseApp() {
+    return firebaseApp;
+  }
```
ramanpreetSinghKhinda, line 63: _Just for Info_ How are we using these APIs? Currently I don't see it being called from anywhere?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -164,6 +153,7 @@ public static FirebasePerformance getInstance() {
    * @param gaugeManager The GaugeManager instance.
    */
   @VisibleForTesting
+  @Inject
```
ramanpreetSinghKhinda, line 156: Since now we've changed `FirebasePerformance` object to be dependency injected don't we want to update the [corresponding tests](https://github.com/firebase/firebase-android-sdk/blob/6270c03055591f55b4206c06373a7184012952f7/firebase-perf/src/test/java/com/google/firebase/perf/FirebasePerformanceTest.java#L565-L573)?

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+@Component(modules = {FirebasePerformanceModule.class})
+public interface FirebasePerformanceComponent {
+  @NonNull
+  FirebasePerformance getFirebasePerformance();
```
ramanpreetSinghKhinda, line 30: Although currently `FirebasePerfRegistrar` is the only place where `FirebasePerformance` object is created so we should be fine but since Dagger allows us to define [Singleton](https://dagger.dev/dev-guide/) scope why not use that to guarantee singleton behavior?

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
jeremyjiang-dev: Yes, it is not explicitly called anywhere. The method means that whenever we need a `FirebaseApp` type object, we will get `firebaseApp` as indicated in the methods body. The method name does not really matter. 

See https://dagger.dev/api/2.12/dagger/Provides.html. "The method's return type is bound to its returned value. The component implementation will pass dependencies to the method as parameters."

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
jeremyjiang-dev: You are right. We will do this later this week in another PR.

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
jeremyjiang-dev: We use `@hide` to hide the public methods from generating api.txt. Would you like to link the bug id? 
Yes, we will do it later across project for the annotation fix.

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
jeremyjiang-dev: Good observation. We will introduce singleton with scopes in the next PR.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/FirebasePerformanceModule.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: Dependency injection does not change the way we create objects and we did not modify any existing constructors. Also, [the dagger dev site](https://dagger.dev/dev-guide/testing.html) says "You don’t have to use Dagger for single-class unit tests". Therefore, we don't have to update the tests. 
@visumickey What do you think?

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
ramanpreetSinghKhinda: Sg! b/177448055 is the bug.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
ramanpreetSinghKhinda: Maybe fine for now (since the change scope is very limited currently) but I assume as we expand on this we would want to change the way how we call constructors from the tests.

ramanpreetSinghKhinda (approved):

yingdai3 (approved):
LGTM!

PR #2410, opened 2021-02-04
Initial implementation of protoc codegen.

vkryachko:


encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/CodeGenerator.kt
```diff
@@ -14,10 +14,477 @@
...
+     * Example generated enum:
+     *
+     * ```java
+     * import com.google.firebase.encoders.proto.ProtoEnum;
+     *
+     * public enum Foo implements ProtoEnum {
+     *   DEFAULT(0),
+     *   EXAMPLE(1);
+     *
+     *   private final int number;
```
rlazo, line 158:
```suggestion
     *   private final int number_;
```

encoders/protoc-gen-firebase-encoders/src/main/kotlin/com/google/firebase/encoders/proto/codegen/Types.kt
```diff
@@ -149,6 +154,8 @@ sealed class Owner(val scopeSeparator: Char) {
     }
 }
 
+private val SNAKE_REGEX = "_[a-zA-Z]".toRegex()
```
rlazo, line 157: SNAKE_CASE_REGEX

rlazo (approved):

PR #2413, opened 2021-02-05
Adjust returned information from url errorStream

annzimmer:


annzimmer:
Oops, added the functional test to the write branch. It's pretty small. I'll leave it here unless you object.

rlazo (approved):

PR #2418, opened 2021-02-08
Make Auth an deferred dependency

schmidt-sebastian:
Support Auth as an optional dependencies.

Tested manually via test app (which does not yet use actual dynamic module support).


firebase-database/src/main/java/com/google/firebase/database/android/AndroidAuthTokenProvider.java
```diff
@@ -21,65 +21,66 @@
...
-      public void addTokenChangeListener(
-          ExecutorService executorService, TokenChangeListener listener) {
-        executorService.execute(() -> listener.onTokenChange(null));
-      }
+    IdTokenListener idTokenListener =
+        tokenResult ->
+            executorService.execute(
+                () -> tokenListener.onTokenChange(/* nullable */ tokenResult.getToken()));
+    deferredAuthProvider.whenAvailable(
+        // TODO: Could "whenAvailable" return the Auth instance rather than a provider for Auth?
```
vkryachko, line 77: The return of the provider is intentional since it's more general and makes things lazy by default - so that we don't have to initialize with inside components unless get() is called on its provider.

vkryachko (commented):
The general direction LGTM

firebase-database/src/main/java/com/google/firebase/database/android/AndroidAuthTokenProvider.java
dconeybe: Could `Deferred` be modified to extend `Provider`? As I'm working on changes for Firestore, this would be incredibly convenient because some usages need to react instantly when the provider becomes available (i.e. via `whenAvailable()`) and other instances need to get the instance and do something with it if it is non-null (which would be achieved by calling `get()`). 

Since [OptionalProvider](https://github.com/firebase/firebase-android-sdk/blob/4a366a823000b37790c1af4145366fee914489e3/firebase-components/src/main/java/com/google/firebase/components/OptionalProvider.java#L30) already implements both `Deferred` and `Provider`, this would be an incredibly easy change to implement. Also, it would satisfy @schmidt-sebastian's request for an `isAvailable()` method.

dconeybe: @vkryachko I've implemented this suggestion in https://github.com/firebase/firebase-android-sdk/pull/2432. I'll officially add you both as reviewers once the GitHub Actions pass.

vkryachko:
Hey, firendly ping. Is there anything else needed to productionize this?

FYI the test failure needs some attention

```
UnitTestHelpers.java:170: error: cannot find symbol
    config.setAuthTokenProvider(AndroidAuthTokenProvider.forUnauthenticatedAccess());
```

schmidt-sebastian:
I fixed the compile failure. I don't think there is much else to do here until we have an actual test case that takes advantage of this feature (but this shouldn't block this PR). I will try to get it past the flaky tests today.

vkryachko (approved):

PR #2428, opened 2021-02-09
Make TransformResult non-nullable

schmidt-sebastian:
Gil asked that this should not be part of #2383, but I think it still makes sense as an indepedent simplification

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
@@ -159,7 +159,7 @@ public abstract class LocalStoreTestCase {
             version,
             transformResult != null
                 ? Collections.singletonList(TestUtil.wrap(transformResult))
-                : null);
+                : Collections.emptyList());
     MutationBatchResult result =
         MutationBatchResult.create(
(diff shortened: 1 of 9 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -159,7 +159,7 @@ private void acknowledgeMutation(long documentVersion, @Nullable Object transfor
             version,
             transformResult != null
                 ? Collections.singletonList(TestUtil.wrap(transformResult))
-                : null);
+                : Collections.emptyList());
```
thebrianchen, line 162: optional nit: you can just remove the Nullable annotation in the signature and change this method's overload to pass in an empty list instead of `null`.

thebrianchen (approved):
lgtm

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
schmidt-sebastian: Hm, this would have been nice but unfortunately `wrap` throws when it is passed an empty Object.

PR #2429, opened 2021-02-09
Update functions instances used by device-tests and smoke-tests.

yifanyang:
- upgrade the version of node.js used to build functions from 6 to 12
- add source code for functions instance used by smoke-tests


firebase-functions/src/androidTest/backend/functions/package.json
```diff
@@ -2,14 +2,11 @@
   "name": "functions",
   "description": "Cloud Functions for Firebase",
   "dependencies": {
-    "firebase-admin": "~4.2.1",
-    "firebase-functions": "^0.5.7"
-  },
-  "devDependencies": {
-    "@google-cloud/functions-emulator": "1.0.0-beta.4"
```
rlazo, line 9: We don't need the emulator anymore?

smoke-tests/src/androidTest/backend/functions/functions/index.js
```diff
@@ -0,0 +1,29 @@
+const functions = require("firebase-functions");
```
rlazo, line 1: Since this is actual code, AFAIK it needs the copyright header.

firebase-functions/src/androidTest/backend/functions/package.json
yifanyang: This npm module has been deprecated: https://www.npmjs.com/package/@google-cloud/functions-emulator. One of the recommended alternative is to start emulator with `firebase-tools`, which is the same tool used to deploy functions. So I think it's no longer needed here.

smoke-tests/src/androidTest/backend/functions/functions/index.js
yifanyang: Done.

rlazo (approved):

PR #2432, opened 2021-02-10
Add dynamic module support to Firestore

dconeybe:
Context: https://developer.android.com/guide/app-bundle/play-feature-delivery

Note to Googlers: This work is tracked by b/177833559.


firebase-components/src/main/java/com/google/firebase/inject/Deferred.java
```diff
@@ -37,11 +37,11 @@
  * }
  * }</pre>
  */
-public interface Deferred<T> {
+public interface Deferred<T> extends Provider<T> {
   /** Used by dependers to register their callbacks. */
   interface DeferredHandler<T> {
     @DeferredApi
-    void handle(Provider<T> provider);
+    void handle(T instance);
```
vkryachko, line 44: I would like to avoid this, since for one it's a breaking change and also is less general and breaks laziness(if  eagerness is not required by other products).

What is this achieving other than removing the need to call `get()` once?

vkryachko (changes requested):

firebase-components/src/main/java/com/google/firebase/inject/Deferred.java
dconeybe: The main advantage is indeed the removal of the need to call `get()`: instead of giving you an indirect reference to the data it gives you the data directly. When I used this code I first thought "why isn't it just giving me the instance? why is it giving me a provider to the instance?" I see your points about lazy initialization though.

What if the argument to `handle()` is changed from `Provider<T>` to `Deferred<T>` and the `this` instance is passed as the argument? Since `Deferred` now extends `Provider` this provides one authoritative way to get the instance: call `get()` on the `Deferred` object.

Also, since this is an internal-only API (and I don't see any products using `Deferred` yet), could you explain the concerns about this being a breaking change?

dconeybe: Note: The discussion about modifying `Deferred` to extend `Provider` occurred here: https://github.com/firebase/firebase-android-sdk/pull/2418#pullrequestreview-586863949

vkryachko: I feel like `Deferred` and `Provider` are completely disjoint concepts - even though related in the context of dependency injection. The fact that `OptionalProvider` implements both is just an internal implementation detail that I would prefer not to leak into the interface.

> What if the argument to handle() is changed from Provider<T> to Deferred<T> and the this instance is passed as the argument? Since Deferred now extends Provider this provides one authoritative way to get the instance: call get() on the Deferred object.

I guess we could do that, but I don't see the benefits of this approach, afaict it's completely isomorphic to the existing interface however:

* has the downside I described above
* provides 2 ways of doing the same thing, i.e. calling get on the deferred itself or on the provider passed into `handle()`
* constrains the implementation of deferred as its contract becomes more complicated.

Looks like rtdb's change has no problem with deferred not extending provider and instead uses an atomic reference to determine if auth is present or not, can we do the same in firestore?

vkryachko (changes requested):

dconeybe:
@vkryachko I've reverted the changes to `Deferred` and `OptionalProvider`. I see what you mean about `Deferred` and `Provider` being disjoint concepts. I still feel that adding the ability to get the value from the `Deferred` is useful but it's not a deal breaker.

firebase-components/src/main/java/com/google/firebase/inject/Deferred.java
dconeybe: I've reverted the changes to `Deferred` and `OptionalProvider`. It is indeed possible to use `Deferred` without the changes I suggested and I've made those changes as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -40,7 +44,7 @@
 
   private static final String LOG_TAG = "FirebaseAuthCredentialsProvider";
 
-  private final InternalAuthProvider authProvider;
+  private AtomicReference<Provider<InternalAuthProvider>> internalAuthProviderProviderRef;
```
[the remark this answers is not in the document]
vkryachko: You could also consider storing the `InternalAuthProvider` directly since you seem to need the provider value right away(hence laziness is not an option for you) 
```suggestion
  private final AtomicReference<InternalAuthProvider> internalAuthProviderRef;
```

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -60,30 +64,47 @@
...
     tokenCounter = 0;
 
-    authProvider.addIdTokenListener(idTokenListener);
+    deferredAuthProvider.whenAvailable(
+        provider -> {
+          synchronized (this) {
+            internalAuthProviderProviderRef.set(provider);
+            currentUser = getUser(internalAuthProviderProviderRef.get());
+            if (changeListener != null) {
+              changeListener.onValue(currentUser);
```
vkryachko, line 90: `whenAvailable` will run on the UI thread, is it ok to execute the changeListener on it?

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -60,30 +64,47 @@
...
 
   @Override
   public synchronized Task<String> getToken() {
     boolean doForceRefresh = forceRefresh;
     forceRefresh = false;
-    Task<GetTokenResult> res = authProvider.getAccessToken(doForceRefresh);
+
+    InternalAuthProvider internalAuthProvider = internalAuthProviderProviderRef.get().get();
+    if (internalAuthProvider == null) {
+      return Tasks.forException(new FirebaseApiNotAvailableException("auth is not available"));
```
vkryachko, line 104: Are you sure you want to throw in this case? Shouldn't this case be treated the same as it is in the EmptyCredentialsProvider and return `Task.forResult(null)`?
https://github.com/firebase/firebase-android-sdk/blob/18d574fe59119c4422a3acf4a566c20bd187e0fb/firebase-firestore/src/main/java/com/google/firebase/firestore/auth/EmptyCredentialsProvider.java#L25-L29

Additionally, is `EmptyCredentialsProvider` still needed or can it be removed?

firebase-firestore/src/test/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProviderTest.java
```diff
@@ -0,0 +1,354 @@
...
+
+  /** An implementation of {@link Deferred} whose provider never becomes available. */
+  private static final class UnavailableDeferred implements Deferred<InternalAuthProvider> {
+
+    @Override
+    public void whenAvailable(@NonNull DeferredHandler<InternalAuthProvider> handler) {}
+  }
+
+  /** An implementation of {@link Deferred} whose provider is always available. */
+  private static final class ImmediateDeferred implements Deferred<InternalAuthProvider> {
```
schmidt-sebastian, line 300: Would it make sense to move these classes to separate package or utility class, so that they can be used from both test files? We could have something similar to the `Tasks` class with "Deferred.forX()" style factory methods.

I also wonder if `ImmediateDeferred` should be part of firebase-common.

firebase-firestore/src/test/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProviderTest.java
```diff
@@ -0,0 +1,354 @@
...
+
+    private final Object lock = new Object();
+    private final List<DeferredHandler<InternalAuthProvider>> handlers = new ArrayList<>();
+    private final AtomicReference<Provider<InternalAuthProvider>> internalAuthProviderProviderRef =
+        new AtomicReference<>();
+
+    @Override
+    public void whenAvailable(@NonNull DeferredHandler<InternalAuthProvider> handler) {
+      Provider<InternalAuthProvider> internalAuthProviderProvider;
+      synchronized (lock) {
```
schmidt-sebastian, line 328: Again, I think we should not be mixing `synchronized` and `AtomicReference`. 

Since this is a test class, you could also synchronize on the object itself and use `public synchronized void whenAvailable()` as the method signature.

I also wonder if these methods need to be synchronized at all. Does Mockito invoke the methods from a different thread?

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -124,12 +145,17 @@ public synchronized void setChangeListener(@NonNull Listener<User> changeListene
...
+    }
   }
 
-  /** Returns the current {@link User} as obtained from the given FirebaseApp instance. */
-  private User getUser() {
-    @Nullable String uid = authProvider.getUid();
+  /** Returns the current {@link User} as obtained from the given InternalAuthProvider. */
+  private static User getUser(Provider<InternalAuthProvider> provider) {
+    InternalAuthProvider internalAuthProvider = provider.get();
+    @Nullable String uid = (internalAuthProvider == null) ? null : internalAuthProvider.getUid();
```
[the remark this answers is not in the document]
schmidt-sebastian: Never mind, I got confused because this is a `ProviderProvider`. I think this will be more obvious when this is just a `Provider`.

schmidt-sebastian (commented):
I added some more comments (after guessing what your reply on my last round was :) )

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -60,30 +64,47 @@
...
+          provider.get().addIdTokenListener(idTokenListener);
+        });
   }
 
   @Override
   public synchronized Task<String> getToken() {
     boolean doForceRefresh = forceRefresh;
     forceRefresh = false;
-    Task<GetTokenResult> res = authProvider.getAccessToken(doForceRefresh);
+
```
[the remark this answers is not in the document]
dconeybe: What do you mean by "inline"?

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -40,50 +44,56 @@
...
+  private final IdTokenListener idTokenListener = result -> onIdTokenChanged();
 
-  /** The listener to be notified of credential changes (sign-in / sign-out, token changes). */
-  @Nullable private Listener<User> changeListener;
+  /**
+   * The {@link Provider} that gives access to the {@link InternalAuthProvider} instance; initially,
+   * its {@link Provider#get} method returns {@code null}, but will be changed to a new {@link
+   * Provider} once the "auth" module becomes available.
+   */
+  @Nullable
```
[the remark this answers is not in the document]
dconeybe: Fixed.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
```diff
@@ -40,50 +44,56 @@
...
+  @GuardedBy("this")
+  private Provider<InternalAuthProvider> internalAuthProviderProvider = () -> null;
```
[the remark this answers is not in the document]
dconeybe: You are correct; however, there is a custom lint check that complains when a "deferred" object is saved to an instance variable. Storing the `Provider` makes the lint check happy.

The lint error is "Provider.get() assignment to a field detected" (https://github.com/firebase/firebase-android-sdk/blob/f56a977990a6dc1aee1de9d50a94e69225246abd/tools/lint/src/main/kotlin/ProviderAssignmentDetector.kt#L47-L50)

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
dconeybe: What are your thoughts on this in light of the code in `FirestoreCallCredentials` that I linked to?

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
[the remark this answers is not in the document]
vkryachko: hm, it seems that this lint check is too strict in this case, meaning that it is not unreasonable to assign to a field from a `@DeferredApi` method - which the `whenAvailable()` callback lambda is an instance of. Please file a bug at go/fireescape-bug and feel free to `@SuppressLint` it or leave your code as is for now.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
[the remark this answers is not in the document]
schmidt-sebastian: You can ignore this since this is existing code, but I can't figure out why we need this extra variable. By inlining I mean replace "doForceRefresh" with "forceRefresh" (and setting "forceRefresh" to false a bit later).

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
schmidt-sebastian: Your answer wasn't quite what I predicted. You have changed my mind. I was expecting the consumer to be a bit simpler than it actually is. Please disregard.

firebase-firestore/src/test/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProviderTest.java
```diff
@@ -0,0 +1,354 @@
...
+  @Test
+  public void getTokenShouldReturnAFailedTaskIfProviderIsNotAvailable() {
+    FirebaseAuthCredentialsProvider firebaseAuthCredentialsProvider =
+        new FirebaseAuthCredentialsProvider(new UnavailableDeferred());
+
+    Task<String> task = firebaseAuthCredentialsProvider.getToken();
+
+    assertThat(task.isComplete()).isTrue();
+    assertThat(task.isSuccessful()).isFalse();
+    assertThat(task.getException()).isInstanceOf(FirebaseApiNotAvailableException.class);
```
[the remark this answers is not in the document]
schmidt-sebastian: "I won't update the test because my code is correct".

Came up with an answer for you.

schmidt-sebastian (approved):
This LGTM, but I do think we should use `SuppressLint` and get rid of the ProviderProvider.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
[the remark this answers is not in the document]
dconeybe: I can't figure it out either. The only difference it would make would be if `internalAuthProvider.getAccessToken()` throws an exception, which seems like a pretty extreme case against which I doubt we need to guard. I've implemented this "inlining" of `doForceRefresh`.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
dconeybe: Ok. I've left the `FirebaseApiNotAvailableException` in there.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java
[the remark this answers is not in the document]
dconeybe: Done. I've changed the instance variable to simply `@Nullable private InternalAuthProvider internalAuthProvider`, added a `SuppressLint` to the constructor, and filed b/181014061 about the the erroneous lint check.

PR #2442, opened 2021-02-16
Add useEmulator support for Firebase Storage

samtstern:
Add `useEmulator` method for the Storage SDK

Testing:

- Added some super-basic unit tests
- Modified the test app to support using the emulator, tested against an early build. This seemed to work for most operations, others failed on the emulator side but no issues with the plumbing.

TODO:

- [ ] Decide if `getRefFromUrl()` should accept localhost URLs and check them against emulator setings.

cc @abeisgoat

### DIFF
--- a/firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
+++ b/firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
@@ -166,7 +166,6 @@ public class FirebaseStorage {
    * @param port the emulator port (for example, 9000)
    */
   public void useEmulator(@NonNull String host, int port) {
-    // TODO(samstern): Throw if used already?
     this.emulatorSettings = new EmulatedServiceSettings(host, port);
   }
 
--- a/firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
+++ b/firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
@@ -24,6 +24,7 @@ import com.google.android.gms.common.internal.Preconditions;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
 import com.google.firebase.auth.internal.InternalAuthProvider;
+import com.google.firebase.emulators.EmulatedServiceSettings;
 import com.google.firebase.inject.Provider;
 import com.google.firebase.storage.internal.Util;
 import java.io.UnsupportedEncodingException;
@@ -50,6 +51,8 @@ public class FirebaseStorage {
   private long sMaxUploadRetry = 10 * DateUtils.MINUTE_IN_MILLIS; //  10 * 60 * 1000
   private long sMaxDownloadRetry = 10 * DateUtils.MINUTE_IN_MILLIS; //  10 * 60 * 1000
   private long sMaxQueryRetry = 2 * DateUtils.MINUTE_IN_MILLIS; //  2 * 60 * 1000
+
+  @Nullable private EmulatedServiceSettings emulatorSettings;
 
   FirebaseStorage(
       @Nullable String bucketName,
@@ -152,6 +155,19 @@ public class FirebaseStorage {
       Log.e(TAG, "Unable to parse url:" + url, e);
       throw new IllegalArgumentException(STORAGE_URI_PARSE_EXCEPTION);
     }
+  }
+
+  /**
+   * Modifies this FirebaseStorage instance to communicate with the Storage emulator.
+   *
+   * <p>Note: Call this method before using the instance to do any storage operations.
+   *
+   * @param host the emulator host (for example, 10.0.2.2)
+   * @param port the emulator port (for example, 9000)
+   */
+  public void useEmulator(@NonNull String host, int port) {
+    // TODO(samstern): Throw if used already?
+    this.emulatorSettings = new EmulatedServiceSettings(host, port);
   }
 
   /**
@@ -306,5 +322,10 @@ public class FirebaseStorage {
   @Nullable
   InternalAuthProvider getAuthProvider() {
     return mAuthProvider != null ? mAuthProvider.get() : null;
+  }
+
+  @Nullable
+  EmulatedServiceSettings getEmulatorSettings() {
+    return emulatorSettings;
   }
 }
--- a/firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
+++ b/firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
@@ -59,7 +59,9 @@ class GetDownloadUrlTask implements Runnable {
 
     if (!TextUtils.isEmpty(downloadTokens)) {
       String downloadToken = downloadTokens.split(",", -1)[0];
-      Uri.Builder uriBuilder = NetworkRequest.getDefaultURL(storageRef.getStorageUri(), storageRef.getEmulatorSettings()).buildUpon();
+      Uri.Builder uriBuilder =
+          NetworkRequest.getDefaultURL(storageRef.getStorageUri(), storageRef.getEmulatorSettings())
+              .buildUpon();
       uriBuilder.appendQueryParameter("alt", "media");
       uriBuilder.appendQueryParameter("token", downloadToken);
       return uriBuilder.build();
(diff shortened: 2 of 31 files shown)


firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
```diff
@@ -154,6 +157,19 @@ public static FirebaseStorage getInstance(@NonNull FirebaseApp app, @NonNull Str
...
+  /**
+   * Modifies this FirebaseStorage instance to communicate with the Storage emulator.
+   *
+   * <p>Note: Call this method before using the instance to do any storage operations.
+   *
+   * @param host the emulator host (for example, 10.0.2.2)
+   * @param port the emulator port (for example, 9000)
+   */
+  public void useEmulator(@NonNull String host, int port) {
+    // TODO(samstern): Throw if used already?
```
samtstern, line 169: Need to make sure to address this

firebase-storage/src/main/java/com/google/firebase/storage/internal/Util.java
```diff
@@ -104,13 +101,7 @@ public static Uri normalize(@NonNull FirebaseApp app, @Nullable String s)
       if (scheme != null
           && (equals(scheme.toLowerCase(), "http") || equals(scheme.toLowerCase(), "https"))) {
         String lowerAuthority = uri.getAuthority().toLowerCase();
-        int indexOfAuth;
-        try {
```
samtstern, line 108: Not sure what this block did before but `RemoteException` is no longer thrown by anything in the path here so the compiler told me to drop this.

schmidt-sebastian: `RemoteException` is dedicated to the long term memory of GMSCore. If you think we should not include references to these dark times, then we can remove it :)

firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
```diff
@@ -307,4 +322,9 @@ public FirebaseApp getApp() {
   InternalAuthProvider getAuthProvider() {
     return mAuthProvider != null ? mAuthProvider.get() : null;
   }
+
+  @Nullable
+  EmulatedServiceSettings getEmulatorSettings() {
```
schmidt-sebastian, line 327: Would it make sense to "generify" this support and just expose host, port and URL? This will make it less likely that I will abuse this feature for non-Emulator settings in a future PR.

firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
```diff
@@ -71,7 +73,8 @@ private Uri extractDownloadUrl(JSONObject response) {
   @Override
   public void run() {
     final NetworkRequest request =
-        new GetMetadataNetworkRequest(storageRef.getStorageUri(), storageRef.getApp());
+        new GetMetadataNetworkRequest(
+            storageRef.getStorageUri(), storageRef.getApp(), storageRef.getEmulatorSettings());
```
schmidt-sebastian, line 77: Can `getStorageUri` return an URL that includes the Emulator? You will not have to expose `StorageReference#getEmulatorSettings` and you will likely not have to change much in the NetworkRequest classes.

firebase-storage/test-app/src/main/res/layout/content_main.xml
```diff
@@ -286,13 +286,24 @@
...
+    <Button
+        android:textSize="7sp"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="Use Emulator"
+        android:id="@+id/useEmulator"
+        android:layout_below="@+id/listSinglePage"
+        android:layout_alignParentLeft="true" />
+
+    <!-- Eleventh Row -->
```
schmidt-sebastian, line 298: Oh, you uncovered the darkest secret for the Storage SDK - the button layout in the test app. I hope you enjoyed it? :)

schmidt-sebastian (commented):
This is fine, but I would have done it a bit differently. Let me know if my suggestions make sense.

Let's have the follow up discussion regarding `referenceFromUrl` over email.

firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
samtstern: Maybe? But that means I have to pass two things all the way down to the request instead of having one thing to encapsulate it.

firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
samtstern: That URI is always a `gs://` URI now so I'm not sure where I could stuff in the emulator details.

For example it would be something like `gs://foo.appspot.com/child` and the code makes the assumption that `getAuthority()` returns `foo.appspot.com` and `getPath()` returns `child`.

Do you think it's safe to turn this into a non-gs URI?  Seems like a risky change.

firebase-storage/test-app/src/main/res/layout/content_main.xml
samtstern: I had to zoom my screen a lot to see those tiny buttons.

firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
schmidt-sebastian: I tried to play with it. It seems much cleaner but it is also a fair bit of work. This is a draft state: https://gist.github.com/schmidt-sebastian/950943121b49b6542f9b4fceda54c540

I gave up after 45 minutes or so, but if you feel like this is worth pursuing you or I could pick it back up. Otherwise, we can probably leave it as is.

firebase-storage/src/main/java/com/google/firebase/storage/FirebaseStorage.java
schmidt-sebastian: You can pass a Uri. This could then return the host. You might even be able to replace all of `EmulatorSettings` with a `Uri host`.

firebase-storage/src/main/java/com/google/firebase/storage/GetDownloadUrlTask.java
samtstern: 🙏 thank you! You always end up sending me a patch file when I get your review :-)

samtstern:
@schmidt-sebastian I introduced a class called `StorageReferenceUri` which is just a utility so you can pass around a single object and get the `gs://` or the `http://` view on it whenever you need it.

This way nothing below `StorageReference` needs to know about `EmulatedServiceSettings` and we don't have to do more string parsing or deal with ambigious `Uri` arguments.

firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java
```diff
@@ -85,35 +89,24 @@
...
-  public static Uri getDefaultURL(@NonNull Uri gsUri) {
-    Preconditions.checkNotNull(gsUri);
-    String pathWithoutBucket = getPathWithoutBucket(gsUri);
-    Uri.Builder uriBuilder = sNetworkRequestUrl.buildUpon();
-    uriBuilder.appendPath("b");
-    uriBuilder.appendPath(gsUri.getAuthority());
-    uriBuilder.appendPath("o");
-    uriBuilder.appendPath(pathWithoutBucket);
-    return uriBuilder.build();
+  public static Uri getBaseUrl(@Nullable EmulatedServiceSettings emulatorSettings) {
```
schmidt-sebastian, line 103: I think this is not used anymore.

firebase-storage/src/main/java/com/google/firebase/storage/internal/StorageReferenceUri.java
```diff
@@ -0,0 +1,75 @@
...
+import android.net.Uri;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.firebase.emulators.EmulatedServiceSettings;
+import com.google.firebase.storage.network.NetworkRequest;
+
+/**
+ * Utility class to encapsulate the two "views" of a storage URI in a single object.
+ *
+ * @hide
```
schmidt-sebastian, line 26: I think you don't need `@hide` here since it is in an internal package. The rest of the classes are a bit inconsistent here.

schmidt-sebastian (approved):
This looks a lot cleaner than before and and much much cleaner than my patch. My goal was to reduce the number of arguments to the different request classes, but my patch added even more arguments. Thanks for discarding it :) 

I do think we should replace EmulatorServiceSettings with a host Uri, but I will leave this up to you.

PR #2443, opened 2021-02-16
New class and methods to implement bulk custom key/value logging.

elenadoty:
Bulk key/value logging reduces the computational overhead of logging multiple custom keys and values all at once by caching all pairs at once rather than individually.


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/CustomKeysAndValues.java
```diff
@@ -0,0 +1,86 @@
...
+      return new CustomKeysAndValues(this);
+    }
+  }
+
+  protected CustomKeysAndValues(@NonNull Builder builder) {
+    this.keysAndValues = builder.keysAndValues;
+  }
+
+  @NonNull
+  public Map<String, String> getCustomValues() {
```
samedson, line 83: Is there a way to keep this private from customers?

mrwillis21: Agreed, I don't think we should have a getter here.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/CustomKeysAndValues.java
```diff
@@ -0,0 +1,81 @@
...
+      return this;
+    }
+
+    @NonNull
+    public CustomKeysAndValues build() {
+      return new CustomKeysAndValues(this);
+    }
+  }
+
+  protected CustomKeysAndValues(@NonNull Builder builder) {
```
mrwillis21, line 78: We should make this `private` or package private (if you want access to it from tests), so it can't be extended, and is only able to be created by the `Builder`.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/CustomKeysAndValues.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Helper class which handles the storage and conversion to strings of key/value pairs with
+ * heterogenous value types.
+ */
+public class CustomKeysAndValues {
+
+  public final Map<String, String> keysAndValues;
```
mrwillis21, line 27: I'd recommend making this package private, that way the `setCustomKeys` method in `FirebaseCrashlytics` will have access but it won't be accessible from the outside.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/UserMetadata.java
```diff
@@ -63,6 +60,40 @@ public void setCustomKey(String key, String value) {
...
+
+      if (attributes.containsKey(key)) {
+        currentKeys.put(key, value);
+      } else {
+        newKeys.put(key, value);
+      }
+    }
+    attributes.putAll(currentKeys);
+
+    // Add new keys if there is space
```
mrwillis21, line 81: Should we at least try to add *some* of them? This is a tricky question since the map will be unordered, but disallowing *any* new values also seems suboptimal.

mrwillis21 (changes requested):

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/UserMetadata.java
elenadoty: Added an alphabetical sort to add however many we can in a predictable way

mrwillis21 (approved):

PR #2450, opened 2021-02-19
Additional javadoc fixes.

rlazo:


firebase-perf/src/main/java/com/google/firebase/perf/internal/SessionAwareObject.java
```diff
@@ -28,6 +28,7 @@
    * Updates the SessionAwareObject with the new sessionId.
    *
    * @param session The new PerfSession.
+   * @hide
```
jeremyjiang-dev, line 31: Why do you need to add `@hide` here? And do you need to add `@hide` to all the classes in the `internal` directory?

rlazo: @jeremyjiang-dev our javadoc pipeline was failing because an error here, related to the fact that the `PerfSession` (the return type of the function) is hidden. I'm not entirely sure why it failed, since `SessionAwareObject` is marked as `@hidden` too, which should mean that no javadoc is generated. In any case, adding this annotation fixed the issue. No additional annotation would be needed.

jeremyjiang-dev: Thanks for the explanation! LGTM.

jeremyjiang-dev (approved):

PR #2458, opened 2021-02-22
Removing the isPerformanceMonitoringEnabled check in FirebasePerfProvider and moving the getSharedPreferences to a separate thread.

raymondlam:
To address StrictMode violation in b/171411615.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/provider/FirebasePerfProvider.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/provider/FirebasePerfProvider.java
@@ -61,13 +61,11 @@ public class FirebasePerfProvider extends ContentProvider {
     ConfigResolver configResolver = ConfigResolver.getInstance();
     configResolver.setContentProviderContext(getContext());
 
-    if (configResolver.isPerformanceMonitoringEnabled()) {
-      AppStateMonitor.getInstance().registerActivityLifecycleCallbacks(getContext());
-      AppStartTrace appStartTrace = AppStartTrace.getInstance();
-      appStartTrace.registerActivityLifecycleCallbacks(getContext());
+    AppStateMonitor.getInstance().registerActivityLifecycleCallbacks(getContext());
+    AppStartTrace appStartTrace = AppStartTrace.getInstance();
+    appStartTrace.registerActivityLifecycleCallbacks(getContext());
 
-      mHandler.post(new AppStartTrace.StartFromBackgroundRunnable(appStartTrace));
-    }
+    mHandler.post(new AppStartTrace.StartFromBackgroundRunnable(appStartTrace));
 
     // In the case of cold start, we create a session and start collecting gauges as early as
     // possible.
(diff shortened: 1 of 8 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/provider/FirebasePerfProvider.java
```diff
@@ -61,13 +61,11 @@ public void attachInfo(Context context, ProviderInfo info) {
     ConfigResolver configResolver = ConfigResolver.getInstance();
     configResolver.setContentProviderContext(getContext());
```
ramanpreetSinghKhinda, line 62: We don't even need to initialize this class if we are not using the flag value atm.

ramanpreetSinghKhinda (approved):

firebase-perf/src/test/java/com/google/firebase/perf/internal/AppStateMonitorTest.java
```diff
@@ -387,6 +388,7 @@ public void screenTrace_perfMonDisabledAtBuildTime_traceNotCreated() {
     Activity activityWithNonHardwareAcceleratedView =
         createFakeActivity(/* isHardwareAccelerated =*/ true);
     ConfigResolver configResolver = ConfigResolver.getInstance();
+    configResolver.setDeviceCacheManager(new DeviceCacheManager(new FakeDirectExecutorService()));
```
ramanpreetSinghKhinda, line 391: We should not need to set Device Cache in any other test that doesn't directly talk to Device Cache. Is any test erroring out if we don't set this?

*Same for all references below.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/TraceTest.java
```diff
@@ -80,7 +81,9 @@ public void setUp() {
     ConfigResolver.clearInstance();
 
     context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE).edit().clear().commit();
-    ConfigResolver.getInstance().setApplicationContext(context);
+    ConfigResolver configResolver = ConfigResolver.getInstance();
+    configResolver.setDeviceCacheManager(new DeviceCacheManager(new FakeDirectExecutorService()));
+    configResolver.setApplicationContext(context);
```
ramanpreetSinghKhinda, line 86: I think we can remove this. Is any test erroring out?

raymondlam: Similar with the above comment on (setIsPerformanceCollectionEnabled).

firebase-perf/src/test/java/com/google/firebase/perf/internal/AppStateMonitorTest.java
raymondlam: Since the getSharedPreferences call is now kicked off in a background thread, we need to make sure that tests uses the FakeDirectExecutorService. Otherwise, there might be a race condition with when the getSharedPreferences are available if it is down in a background thread.

Calls like `setIsPerformanceCollectionEnabled` use the device manager under the hood and can have different results depending on when getSharedPreferences is enabled.

firebase-perf/src/main/java/com/google/firebase/perf/provider/FirebasePerfProvider.java
raymondlam: The goal is to kick off the background thread as soon as possible. If we lazy call it, it will not be available when we need to use a flag.

ramanpreetSinghKhinda (approved):

jeremyjiang-dev (approved):

PR #2467, opened 2021-02-25
Add @Singleton annotation to FirebasePerformance class and component.

visumickey:


jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
```diff
@@ -18,13 +18,15 @@
...
 /**
  * Dagger component to create FirebasePerformanceComponent Objects.
  *
  * @hide
  */
 @Component(modules = {FirebasePerformanceModule.class})
+@Singleton
 public interface FirebasePerformanceComponent {
   @NonNull
   FirebasePerformance getFirebasePerformance();
```
ramanpreetSinghKhinda, line 32: I am still new to this but shouldn't the Singleton be placed with `getFirebasePerformance()` object instead?

vkryachko (approved):

firebase-perf/src/main/java/com/google/firebase/perf/injection/components/FirebasePerformanceComponent.java
visumickey: Depends. In this case, both does not make a difference much. We are making the Firebase Performance Component to be a singleton as well instead of just making the instance singleton.

PR #2468, opened 2021-02-26
Post-Merge cleanup of MutableDocuments

schmidt-sebastian:
Some small fixes to MutableDocuments that I noticed while porting to JS (more likely coming). I also added the often asked for factory functions since it helps with readability (otherwise the reader does not know what `new MutableDocument` means).

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
@@ -21,11 +21,10 @@ import com.google.firestore.v1.Value;
  * Represents a document in Firestore with a key, version, data and whether it has local mutations
  * applied to it.
  *
- * <p>Documents can transition between states via {@link
- * #convertToFoundDocument(SnapshotVersion,ObjectValue)}, {@link
- * #convertToNoDocument(SnapshotVersion)} and {@link #convertToUnknownDocument(SnapshotVersion)}. If
- * a document does not transition to one of these states even after all mutations have been applied,
- * {@link #isValidDocument()} returns false and the document should be removed from all views.
+ * <p>Documents can transition between states via {@link #convertToFoundDocument}, {@link
+ * #convertToNoDocument} and {@link #convertToUnknownDocument}. If a document does not transition to
+ * one of these states even after all mutations have been applied, {@link #isValidDocument} returns
+ * false and the document should be removed from all views.
  */
 public final class MutableDocument implements Document, Cloneable {
 
@@ -66,6 +65,10 @@ public final class MutableDocument implements Document, Cloneable {
   private ObjectValue value;
   private DocumentState documentState;
 
+  private MutableDocument(DocumentKey key) {
+    this.key = key;
+  }
+
   private MutableDocument(
       DocumentKey key,
       DocumentType documentType,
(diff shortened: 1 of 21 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
```diff
@@ -17,6 +17,15 @@
 import androidx.annotation.NonNull;
 import com.google.firestore.v1.Value;
 
+/**
+ * Represents a document in Firestore with a key, version, data and whether it has local mutations
+ * applied to it. Documents can start out as {@link DocumentType#INVALID} and transition to a valid
```
dconeybe, line 22: "Documents can start out as..." is not as relevant now since factory methods create the object directly into the desired state. Also, `DocumentType` is a private implementation detail that should not be mentioned in the javadocs for the class. This entire paragraph about starting in an invalid state and transitioning to a different state can probably just be deleted.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
```diff
@@ -73,6 +78,46 @@ private MutableDocument(
     this.value = value;
   }
 
+  /**
+   * Creates a document with no known version or data, but which can serve as base document for
+   * mutations.
+   */
+  public static MutableDocument newInvalidDocument(DocumentKey documentKey) {
+    return new MutableDocument(
```
dconeybe, line 86: These factory methods appear to duplicate the logic defined in their corresponding `convertToXXX()` methods. To be a bit more "dry" (go/dry), consider reviving the constructor that takes the `DocumentKey`, changing its visibility to `private`, using that constructor, and then calling the corresponding `convertToXXX()` method, returning its result.

dconeybe (approved):
♥ the new factory methods!

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
schmidt-sebastian: Updated the comment.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
schmidt-sebastian: I opted to not call into the one argument constructor since keeping it initializes some of the values twice. The one argument constructor would call the default constructor, which then creates new objects, only to be immediately destroyed by the mutating function.

Furthermore, as least some of the code duplication would have to be moved to the one argument constructor.

dconeybe: How about this idea. The only constructor is `MutableDocument(DocumentKey)`, which initializes `this.key` and does not assign anything to the other instance variables (they will be initialized to null). If this constructor is called from the factory methods, followed by the corresponding `convertToXXX()` method, then all instance variables will be initialized and none will be multiply initialized. I realize that this leaves the object temporarily in a weird state after calling the constructor; however, since it is private you have complete control over how the newly-constructed object is used and can prevent ever exposing an object in this intermediate state.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
```diff
@@ -19,12 +19,13 @@
...
  * Represents a document in Firestore with a key, version, data and whether it has local mutations
- * applied to it. Documents can start out as {@link DocumentType#INVALID} and transition to a valid
- * states via {@link #convertToFoundDocument(SnapshotVersion,ObjectValue)}, {@link
- * #convertToNoDocument(SnapshotVersion)} and {@link #convertToUnknownDocument(SnapshotVersion)}.
+ * applied to it.
  *
- * <p>Invalid documents serve as base documents for mutations. If a document remains invalid even
- * after all mutations have been applied, it should be removed from all views.
+ * <p>Documents can transition between states via {@link
+ * #convertToFoundDocument(SnapshotVersion,ObjectValue)}, {@link
```
dconeybe, line 25: Optional: For maintainability and brevity of this documentation, you can omit the arguments in the `@link` tags. For example,  `{@link #convertToFoundDocument(SnapshotVersion,ObjectValue)}` can simply be `{@link #convertToFoundDocument}`. Both javadoc generation and IDE code navigation will infer the correct arguments.

dconeybe (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
schmidt-sebastian: Interesting. I thought that is how it worked, but I couldn't get the linking to work. It turns out it only works without `()`.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
schmidt-sebastian: I like that. Updated.

dconeybe (approved):

PR #2472, opened 2021-03-01
Make CpuGaugeCollector and MemoryGaugeCollector DI compatible.

jeremyjiang-dev:


firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Provider for {@link com.google.firebase.perf.gauges}.
+ *
+ * @hide
+ */
+@Module
+public class GaugeCollectorsModule {
+
+  public static final long INVALID_CPU_COLLECTION_FREQUENCY = -1;
```
ramanpreetSinghKhinda, line 34: nit: May be rename this to `INVALID_COLLECTION_FREQUENCY` and then you can use this across Cpu and Memory.

visumickey (approved):

ramanpreetSinghKhinda (dismissed):

ramanpreetSinghKhinda:
Are you planning to update tests in the separate PR?

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
jeremyjiang-dev: The `INVALID_CPU_COLLECTION_FREQUENCY` in module is only used for CpuGaugeCollector.

jeremyjiang-dev:
> Are you planning to update tests in the separate PR?

We may not need to update tests. The tests are already using the constructors that I marked as @Inejct

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
```diff
@@ -96,6 +97,7 @@ private CpuGaugeCollector() {
   }
 
   @VisibleForTesting
+  @Inject
```
ramanpreetSinghKhinda, line 100: It doesn't allow me to add a comment on the unchanged code so adding it here.

Why are we still keeping the private constructor and the `sharedInstance` reference?

*same for memory collector class.

jeremyjiang-dev: We are adding DI step by step. First, make all classes constructor DI compatible. Second, remove all shareInstance and getInstance(). Let me know if you have questions.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+/**
+ * Provider for {@link com.google.firebase.perf.gauges}.
+ *
+ * @hide
+ */
+@Module
+public class GaugeCollectorsModule {
```
ramanpreetSinghKhinda, line 32: Is it preferred to create module for specific classes like `Cpu/MemoryGaugeCollector` in this case or should it be functionality wise for e.g. `GaugeModule` for all gauge related things?

The reason I ask this is because I will have some stuff for `GaugeMetadataManager` and I think it would make more sense to have it unified functionality wise.

jeremyjiang-dev: Yes. I created the module because I think `Cpu/MemoryGaugeCollector` will be a self-contained module. We can also group them by directory (although some directories need to be restructured)

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
ramanpreetSinghKhinda: I think we chose bottom up approach to avoid coming back to these classes again. Do you think it would not be feasible to have those `sharedInstance` and `getInstance()` replaced now? And if yes, what's blocking us?

fyi: even I need to answer this question for the GaugeMetadaManager class changes (_which I am on the way_)

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
ramanpreetSinghKhinda: +1 on directory restructure but cases where functionality is well contained lets have a single module. Can you rename this module to `GaugeModule` in that case? 

fyi I have created `GaugeModule` class for `GaugeMetadaManager` but whosoever push the changes last can resolve any merge errors.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
ramanpreetSinghKhinda: Looks like Dagger created this class called `GaugeMetadataManager_Factory` for `GaugeMetadataManager` with public static methods for me to create new instance. 

I think if you mark your class as `@Singleton` you should also be able to get the singleton instance and hence won't need to have your own `getInstance()` method.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
jeremyjiang-dev: Sounds good!

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
```diff
@@ -96,6 +97,7 @@ private CpuGaugeCollector() {
...
   CpuGaugeCollector(
       ScheduledExecutorService cpuMetricCollectorExecutor,
       String fakeProcFileName,
```
jposuna, line 103: Can you update the name here? This is no longer a "fakeProcFileName".

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: 1. Do you think it would not be feasible to have those sharedInstance and getInstance() replaced now?
Not really. For example, GaugeManager calls CpuGaugeManager.getInstance() and we cannot remove `getInstance()` unless make GaugeManager DI compatible.

2.I think if you mark your class as `@Singleton` you should also be able to get the singleton instance and hence won't need to have your own `getInstance()` method.
You are right @Singleton will make the class singleton. However, for the same reason above, we cannot remove `getInstance()` method now. Once we have all classes DI compatible, we can easily remove `getInstance()` usages in its caller and inject the class into the caller's constructors.

jposuna, line 100: Also commenting here since I can't comment on the corresponding lines. As you do the exercise of DI, it would be good to do small changes / cleanups in the class structure. For example, here there is a mix of `private final`, `private`, `public`, etc variables, all mixed. 

It would be great to have "blocks" of variables so that we can instantly see the characteristics of a particular variable just through the block they belong to. It also allows us to easily see outliers that should not exist (like a public non-static variable: `cpuMetricReadings`)

Please do this throughout all these classes (no behaviour change, just cleaning up the variables section of the class)

jposuna, line 100: Since cpuMetricReadings is always initialized to the same thing, why not initialize it where it's declared? It doesn't actually depend on the constructor's parameters.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Provides
+  ScheduledExecutorService providesScheduledExecutorService() {
+    return Executors.newSingleThreadScheduledExecutor();
+  }
+
+  @Provides
+  String providesProcFileName() {
+    int pid = android.os.Process.myPid();
+    String procFileName = "/proc/" + Integer.toString(pid) + "/stat";
```
jposuna, line 44: Usually, it's preferable to use String.format over string concatenation. It makes the strings easier to read, and provides some compile-time checking with typing (since you could do %d for the `pid` instead of converting it to a string).

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
```diff
@@ -0,0 +1,62 @@
...
+    return procFileName;
+  }
+
+  @Provides
+  long providesClockTicksPerSecond() {
```
jposuna, line 49: It feels like we should annotate more than just `long` here. Similar to the comment about `string` above.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public static final long INVALID_CPU_COLLECTION_FREQUENCY = -1;
+
+  @Provides
+  ScheduledExecutorService providesScheduledExecutorService() {
+    return Executors.newSingleThreadScheduledExecutor();
+  }
+
+  @Provides
+  String providesProcFileName() {
```
jposuna, line 42: We should add more annotations that just `string` here. There can be a wide variety of strings to be injected, and we need to provide some way to differentiate between them. Otherwise when a String is requested, Dagger won't know which of the String providers it should use (it does not match by variable name or method name; it matches by type, IIRC)

jeremyjiang-dev: Initially I thought it's okay to not make it not @Named if there is only one provider with String type, but I like your suggestion for better clarity. Done.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeCollectorsModule.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: Good catch! Done.

jeremyjiang-dev: Great suggestion! I think it makes sense to do some small clean up on variables. Updated.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
```diff
@@ -74,39 +76,41 @@
...
   }
 
   @VisibleForTesting
+  @Inject
   CpuGaugeCollector(
       ScheduledExecutorService cpuMetricCollectorExecutor,
-      String fakeProcFileName,
-      long clockTicksPerSecond) {
-    cpuMetricReadings = new ConcurrentLinkedQueue<>();
+      @Named("proc file name") String procFileName,
```
vkryachko, line 105: Prefer custom annotations to `@Named`, http://go/di-practices/prefer-custom-qualifiers

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
vkryachko: Do I understand correctly that this class will be used by both DI injection(from other classes) as well as with getInstance() methods? if so, be careful as you may end up with multiple instances of this class, one created and maintained by dagger as a singleton, and one that you maintain youself in getInstance()

jposuna: Thanks for updating it! There's still the following that could use some sorting:

```
  @Nullable private static CpuGaugeCollector sharedInstance = null;
  @Nullable private ScheduledFuture cpuMetricCollectorJob = null;
  private final ScheduledExecutorService cpuMetricCollectorExecutor;
  private long cpuMetricCollectionRateMs = UNSET_CPU_METRIC_COLLECTION_RATE;
  private final long clockTicksPerSecond;
  private final String procFileName;
```

There's 3 separate blocks there:

- private static
- private final
- private

jposuna: The idea is to make the class DI-compatible, while maintaining the getInstance() method to avoid having a huge PR. While getInstance() exists, all users of the class should be using that, and then, in a single PR, we should remove the getInstance() method and use the injected instance instead.

**It is important to do the step in one go once we migrate because of what @vkryachko is saying - we won't actually have a singleton if we update usages separately.**

ramanpreetSinghKhinda: Isn't there a way for us to get the singleton instance created by Dagger?

I know with Constructor injection we don't even need `getInstance()` but I was wondering to remove the `getInstance()` from the `CpuGaugeCollector` and replace the usages of `CpuGaugeCollector.getInstance()` with `DaggerGeneratedCpuGaugeCollector.getInstance()` (_or something along the lines_)?

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
```diff
@@ -46,10 +48,10 @@
 /** @hide */
 public class CpuGaugeCollector {
```
ramanpreetSinghKhinda, line 49: You will also need to annotate this and the Memory class with `@Singleton` and make any other necessary change (if required) in the specific component/module.

ramanpreetSinghKhinda (approved):
As long as all other comments are addressed.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: Thanks for the reference! Done.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: Updated.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
```diff
@@ -0,0 +1,66 @@
+// Copyright 2021 Google LLC
```
jposuna, line 1: Usually, it's a good idea to have the Modules close to their dependencies (i.e. in their same package).

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
jeremyjiang-dev: Do you have any reference for this? I am seeing inappmessaging put all modules together https://github.com/firebase/firebase-android-sdk/tree/master/firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/injection/modules. The reason could be that it is easier to visualize all modules at once.

jposuna: The main reason is because the module is part of the package's functionality. Putting all modules in a shared package means that there is no relation between the module and the classes that need it (other than maybe by name) - it breaks cohesion.

Internally when working with Guice, we always place the module in the same package as the classes it serves. That's the point of the "package" - that it's one "package" with everything necessary to use it.

jeremyjiang-dev: Sounds good to me! Updated.

firebase-perf/src/main/java/com/google/firebase/perf/gauges/CpuGaugeCollector.java
jeremyjiang-dev: There is no way to remove `getInstance` now unless making the caller DI compatible. As Julio mentioned, this PR is our first step as making the current classes DI compatible without remove `getInstance`, which has the singleton problem as mentioned above and not in a releasable state. Therefore, we are merging to the feature branch called `perf-dagger`.

PR #2474, opened 2021-03-02
Making GaugeMetadataManager class DI compatible.

ramanpreetSinghKhinda:
Making `GaugeMetadataManager` class DI compatible.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
@@ -12,22 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.firebase.perf.injection.components;
+package com.google.firebase.perf.injection.modules;
 
-import androidx.annotation.NonNull;
-import com.google.firebase.perf.FirebasePerformance;
-import com.google.firebase.perf.injection.modules.FirebasePerformanceModule;
-import com.google.firebase.perf.injection.modules.GaugeModule;
-import dagger.Component;
+import android.app.ActivityManager;
+import android.content.Context;
+import dagger.Module;
+import dagger.Provides;
 
 /**
- * Dagger component to create FirebasePerformanceComponent Objects.
+ * Provider for {@link com.google.firebase.perf.internal.GaugeManager} and related classes.
  *
  * @hide
  */
-@Component(modules = {FirebasePerformanceModule.class, GaugeModule.class})
-public interface FirebasePerformanceComponent {
+@Module
+public class GaugeModule {
 
-  @NonNull
-  FirebasePerformance getFirebasePerformance();
+  @Provides
+  Runtime providesRuntime() {
+    return Runtime.getRuntime();
+  }
+
+  @Provides
+  ActivityManager providesActivityManager(Context applicationContext) {
+    return (ActivityManager) applicationContext.getSystemService(Context.ACTIVITY_SERVICE);
+  }
 }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
@@ -15,6 +15,7 @@
 package com.google.firebase.perf.injection.modules;
 
 import android.app.ActivityManager;
+import android.app.ActivityManager.MemoryInfo;
 import android.content.Context;
 import dagger.Module;
 import dagger.Provides;
@@ -35,5 +36,13 @@ public class GaugeModule {
   @Provides
   ActivityManager providesActivityManager(Context applicationContext) {
     return (ActivityManager) applicationContext.getSystemService(Context.ACTIVITY_SERVICE);
+  }
+
+  @Provides
+  MemoryInfo providesMemoryInfo(ActivityManager activityManager) {
+    MemoryInfo memoryInfo = new ActivityManager.MemoryInfo();
+    activityManager.getMemoryInfo(memoryInfo);
+
+    return memoryInfo;
   }
 }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
@@ -50,13 +50,14 @@ class GaugeMetadataManager {
 
   @Inject
   GaugeMetadataManager(
-      Runtime runtime, Context applicationContext, ActivityManager activityManager) {
+      Runtime runtime,
+      Context applicationContext,
+      ActivityManager activityManager,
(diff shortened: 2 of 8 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
```diff
@@ -0,0 +1,39 @@
...
+
+import android.app.ActivityManager;
+import android.content.Context;
+import dagger.Module;
+import dagger.Provides;
+
+/**
+ * Provider for {@link com.google.firebase.perf.internal.GaugeManager} and related classes.
+ *
+ * @hide
```
jposuna, line 25: Why do we have this `@hide` annotation?

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
```diff
@@ -40,22 +41,21 @@
...
-    this(Runtime.getRuntime(), applicationContext);
-  }
 
   @VisibleForTesting
-  GaugeMetadataManager(Runtime runtime, Context applicationContext) {
-    this.runtime = runtime;
+  @Inject
+  GaugeMetadataManager(
+      Runtime runtime, Context applicationContext, ActivityManager activityManager) {
     this.applicationContext = applicationContext;
```
jeremyjiang-dev, line 55: Why do you need to add `ActivityManager activityManager` in the constructor? Can you use `applicationContext.getSystemService(Context.ACTIVITY_SERVICE)` to get the activityManager?

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
ramanpreetSinghKhinda: It's historical. All of our javadocs are annotated with `@hide` and I don't why.

I have a b/177448055 to revisit our use of annotations.

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
```diff
@@ -40,22 +41,21 @@
...
-    this.activityManager =
-        (ActivityManager) applicationContext.getSystemService(Context.ACTIVITY_SERVICE);
+    this.activityManager = activityManager;
+    this.runtime = runtime;
+
     memoryInfo = new ActivityManager.MemoryInfo();
```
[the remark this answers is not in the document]
ramanpreetSinghKhinda: `MemoryInfo` is actually a static inner class of `ActivityManager`. So doing `new ActivityManager.MemoryInfo()` will create an empty `MemoryInfo` object for us which is filled with details once we do `activityManager.getMemoryInfo(memoryInfo)`.

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
ramanpreetSinghKhinda: As part of the DI changes I am also looking into what dependencies we were previously instantiating within the class itself and should actually be injected from outside. This is one of those.

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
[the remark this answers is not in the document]
jposuna: Would it make sense to receive it injected then?

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
jeremyjiang-dev: I see. In the test, you may want to mock `activityManager` first and then pass it to `GaugeMetadataManager`'s constructor. The current test does not seem to benefit from your constructor change.

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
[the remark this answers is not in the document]
ramanpreetSinghKhinda: Done.

Discussed offline and TLDR is we should inject (_in most cases_) everything that is class scoped.

firebase-perf/src/main/java/com/google/firebase/perf/internal/GaugeMetadataManager.java
ramanpreetSinghKhinda: It is being [shadowed](http://robolectric.org/extending/) instead of mocked as we can't mock some Android specific API's directly:

https://github.com/firebase/firebase-android-sdk/blob/ad95b179508f9246dba768786f2a85154ec9b9d1/firebase-perf/src/test/java/com/google/firebase/perf/internal/GaugeMetadataManagerTest.java#L77-L78

https://github.com/firebase/firebase-android-sdk/blob/ad95b179508f9246dba768786f2a85154ec9b9d1/firebase-perf/src/test/java/com/google/firebase/perf/internal/GaugeMetadataManagerTest.java#L91

---

FYI: I am now passing [spy](https://javadoc.io/static/org.mockito/mockito-core/3.8.0/org/mockito/Mockito.html#13) instead of actual objects to make our mocks/fakes clear.

jeremyjiang-dev: Thanks for the clarification on spy! My point was that you passed the real object instead of the spy one 
https://github.com/firebase/firebase-android-sdk/blob/ad95b179508f9246dba768786f2a85154ec9b9d1/firebase-perf/src/test/java/com/google/firebase/perf/internal/GaugeMetadataManagerTest.java#L66
so that you are not actually benefiting from making `activityManager` as injected.
Now I am seeing that you are passing the spy object instead of the real obejct here https://github.com/firebase/firebase-android-sdk/blob/a1b4308439ca1b40a09e4ec527ea6a0e7d8d114e/firebase-perf/src/test/java/com/google/firebase/perf/internal/GaugeMetadataManagerTest.java#L72-L75.
This is exactly what I was suggesting.

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
```diff
@@ -0,0 +1,48 @@
+// Copyright 2021 Google LLC
```
jposuna, line 1: Usually it's a good idea to have the modules together with the class it applies to.

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/GaugeModule.java
ramanpreetSinghKhinda: Make sense! Done

PR #2480, opened 2021-03-04
RTDB Transactions

jmwski:
[go/rtdb-transactions](go/rtdb-transactions)

### DIFF
--- a/firebase-database/src/main/java/com/google/firebase/database/DatabaseTransaction.java
+++ b/firebase-database/src/main/java/com/google/firebase/database/DatabaseTransaction.java
@@ -1,0 +1,73 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.database;
+
+import androidx.annotation.NonNull;
+
+import com.google.android.gms.tasks.Tasks;
+import com.google.firebase.database.annotations.Nullable;
+
+import java.util.concurrent.ExecutionException;
+
+public class DatabaseTransaction {
+
+    public static class DatabaseTransactionContext {
+
+        private final com.google.firebase.database.core.Transaction transaction;
+        private final FirebaseDatabase database;
+
+        DatabaseTransactionContext(com.google.firebase.database.core.Transaction transaction, FirebaseDatabase database) {
+            this.transaction = transaction;
+            this.database = database;
+        }
+
+        @Nullable
+        public DataSnapshot get(@NonNull Query query) throws FirebaseDatabaseException {
+            try {
+                return Tasks.await(transaction.get(query));
+            } catch (ExecutionException e) {
+                if (e.getCause() instanceof FirebaseDatabaseException) {
+                    throw ((FirebaseDatabaseException) e.getCause());
+                }
+                throw new RuntimeException((e.getCause()));
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        @Nullable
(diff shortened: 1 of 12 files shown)


firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -972,6 +1008,10 @@ public void run() {
     }
   }
 
+  public <TResult> Task<TResult> runTransaction(Function<com.google.firebase.database.core.Transaction, Task<TResult>> updateFunction) {
```
IanWyszynski, line 1011: How should we queue transactions?

Currently, I'm thinking about having something like Firestore's `TransactionRunner` hold a reference to some queue. I'm not sure how that queue should look though.

Existing RTDB transactions have a tree of queues, each path has a queue that is a `List<TransactionData>`.

Transactions ready to be sent to the backend (in state `RUN`) are aggregated and sent in batches in `[sendReadyTransactions](https://github.com/firebase/firebase-android-sdk/blob/0de237d26e0336e83c8a494dec9aea9c8a69eece/firebase-database/src/main/java/com/google/firebase/database/core/Repo.java#L997)`.

The actual `compareAndPut` network request is done [here](https://github.com/firebase/firebase-android-sdk/blob/0de237d26e0336e83c8a494dec9aea9c8a69eece/firebase-database/src/main/java/com/google/firebase/database/core/Repo.java#L1055).  

All of the transactions at a location are batched into a single update, where the client figures out what hash of that location was before any updates, and then constructs a datasnapshot node using `[updateChild](https://github.com/firebase/firebase-android-sdk/blob/0de237d26e0336e83c8a494dec9aea9c8a69eece/firebase-database/src/main/java/com/google/firebase/database/core/Repo.java#L1047)`.

With these new transactions, it doesn't make sense to have a path-based queue. We still want a queue because transactions can be retried, but a `List<Transaction>` might be sufficient.

firebase-database/src/main/java/com/google/firebase/database/DatabaseTransaction.java
```diff
@@ -0,0 +1,70 @@
...
+    DatabaseTransactionContext(
+        com.google.firebase.database.core.Transaction transaction, FirebaseDatabase database) {
+      this.transaction = transaction;
+      this.database = database;
+    }
+
+    @Nullable
+    public DataSnapshot get(@NonNull Query query) throws FirebaseDatabaseException {
+      try {
+        return Tasks.await(transaction.get(query));
```
schmidt-sebastian, line 38: For the future: We need to define whether these reads should reflect pending writes or whether we should reject reads for locations that were written to.

firebase-database/src/main/java/com/google/firebase/database/DatabaseTransaction.java
```diff
@@ -0,0 +1,70 @@
...
+      } catch (ExecutionException e) {
+        if (e.getCause() instanceof FirebaseDatabaseException) {
+          throw ((FirebaseDatabaseException) e.getCause());
+        }
+        throw new RuntimeException((e.getCause()));
```
schmidt-sebastian, line 43: This should throw a FirebaseDatabaseException, shouldn't it?

firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -268,6 +272,26 @@ public synchronized void setLogLevel(@NonNull Logger.Level logLevel) {
...
 
+  /**
+   * Run a transaction capable of reading, and then mutating multiple locations.
+   *
+   * @param updateFunction
+   * @param <TResult>
+   * @return
+   */
+  @Nullable
+  public <TResult> Task<TResult> runTransaction(
```
schmidt-sebastian, line 283: This needs an overload to allow users to specify their own executor.

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -490,6 +494,42 @@ public void onRequestResult(String optErrorCode, String optErrorMessage) {
...
+                    defaultExecutor,
+                    new Continuation<Node, DataSnapshot>() {
+                      @Override
+                      public DataSnapshot then(@NonNull Task<Node> task) throws Exception {
+                        return InternalHelpers.createDataSnapshot(
+                            query.getRef(),
+                            IndexedNode.from(task.getResult(), query.getSpec().getIndex()));
+                      }
+                    })
+                .addOnCompleteListener(
```
schmidt-sebastian, line 514: You should be able to merge these two callbacks.

PR #2484, opened 2021-03-05
Make Config Resolver and its dependencies DI compatible

visumickey:


firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/RemoteConfigManagerModule.java
```diff
@@ -0,0 +1,46 @@
...
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+public class RemoteConfigManagerModule {
+
+    @Provides
+    FirebaseRemoteConfig providesFirebaseRemoteConfig() {
+        return null;
+    }
```
jeremyjiang-dev, line 35: Why do you need to create `RemoteConfigManagerModule.java`? We should provide RemoteConfigManager, ImmutableBundle, and DeviceCacheManager for configResolver class.

jeremyjiang-dev: NVM. I realized that this PR is trying to make configResolver's dependencies DI compatible as well.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/RemoteConfigManagerModule.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+public class RemoteConfigManagerModule {
+
+    @Provides
```
jeremyjiang-dev, line 32: You need to mark @Nullable in provider method https://dagger.dev/api/2.11/dagger/Provides.html

firebase-perf/src/main/java/com/google/firebase/perf/internal/RemoteConfigManager.java
```diff
@@ -49,7 +53,7 @@
 
   private static final String FIREPERF_FRC_NAMESPACE_NAME = "fireperf";
 
-  private static final RemoteConfigManager sharedInstance = new RemoteConfigManager();
+  private static final RemoteConfigManager sharedInstance = DaggerRemoteConfigManagerComponent.create().getRemoteConfigManager();
```
jeremyjiang-dev, line 56: You cannot use `DaggerRemoteConfigManagerComponent.create().getRemoteConfigManager()` to get `RemoteConfigManager` when you are inside `RemoteConfigManager`. Instead, you have to let Dagger create and inject `RemoteConfigManager` to the class that calls `RemoteConfigManager.getInstance`. For the same reason, `RemoteConfigManagerComponent.java` may not be needed.

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
```diff
@@ -70,7 +71,7 @@
    * @param remoteConfigManager the Remote Config values set by Firebase Performance
    * @param metadataBundle a bundle of metadata values set by app developers in the AndroidManifest
    */
-  @VisibleForTesting
+  @Inject
```
jeremyjiang-dev, line 74: Mark this class as `@Singleton`.

firebase-perf/src/main/java/com/google/firebase/perf/internal/RemoteConfigManager.java
```diff
@@ -63,19 +67,7 @@
...
-            /* maximumPoolSize= */ 1,
-            /* keepAliveTime= */ 0L,
-            TimeUnit.SECONDS,
-            new LinkedBlockingQueue<Runnable>()),
-        /* firebaseRemoteConfig= */ null // set once FirebaseRemoteConfig is initialized
-        );
-  }
-
-  @VisibleForTesting
+  @Inject
```
jeremyjiang-dev, line 70: Mark this class as `@Singleton`.

firebase-perf/src/main/java/com/google/firebase/perf/injection/modules/RemoteConfigManagerModule.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.Executor;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import dagger.Module;
+import dagger.Provides;
+
+@Module
+public class RemoteConfigManagerModule {
```
jeremyjiang-dev, line 30: The best practice is to put the module into its corresponding package. It seems like that `RemoteConfigManagerModule` can be placed in config package.

PR #2494, opened 2021-03-10
Remove test duplication

schmidt-sebastian:
This merges the ObjectValue and MutableObjectValue tests. They used to be different since we had ObjectValue and ObjectValueBuilder. The PR also makes an effort to remove test duplication, but makes no guarantees here.


firebase-firestore/src/test/java/com/google/firebase/firestore/model/ObjectValueTest.java
```diff
@@ -32,7 +34,19 @@
...
+public class ObjectValueTest {
+
+  private String fooString = "foo";
+  private Value fooValue = wrap(fooString);
+  private String barString = "bar";
+  private Value barValue = wrap(barString);
+  private Value emptyObject = Value.newBuilder().setMapValue(MapValue.getDefaultInstance()).build();
+
+  @Test
+  public void supportsEmptyObjectValues() {
```
dconeybe, line 46: nit: All of the other test methods in this class have their names prefixed with "test". Can you rename `supportsEmptyObjectValues` to also follow this convention?

dconeybe (approved):

firebase-firestore/src/test/java/com/google/firebase/firestore/model/ObjectValueTest.java
schmidt-sebastian: Done

PR #2507, opened 2021-03-15
Fix test failures from internal presubmit

wu-hui:
1. Use proper executor service for spec tests.
2. Skip `@Hide` annotated Task interface.


firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
```diff
@@ -61,7 +61,20 @@
 
   @Test
   public void testImplementsAllTaskInterface() {
+    // Check if the internal gms Hide annotation is accessible.
+    Class hideClazz = null;
+    try {
+      hideClazz = Class.forName("com.google.android.gms.common.internal.Hide");
```
schmidt-sebastian, line 67: If we found this class, should we just exit the test?

schmidt-sebastian (approved):

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
wu-hui: Why? The presence of this class will only indicate we are running on internal platform, we still need to implement all methods, isn't it?

wu-hui (commented):
Merging it for now. We can apply your suggestion later if that is what is needed here.

firebase-firestore/src/test/java/com/google/firebase/firestore/LoadBundleTaskTest.java
schmidt-sebastian: Please disregard. I though the statements in line 74 were or-ed together. LGTM.

PR #2518, opened 2021-03-17
Make FirebasePerformance initialization lazy.

jeremyjiang-dev:
- Changes:
  - Made FirebasePerformance initialization lazy and moved initialization in the `onActivityResumed()` callback. Based on our definition of [app startup time](https://firebase.google.com/docs/perf-mon/app-start-foreground-background-traces?platform=android#app-start), initializing FirebasePerformance after `onActivityResume()` will not be included in the app startup time. Traces generated before FirebasePerformance initialization will be added to the pending [`pendingEventsQueue`](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java#L123) and will be dispatched when `FirebasePerformance` is available.
 - Verification:
   - Verified app start trace, network traces, and trace with annotation
 - Impact
   - Saved 6.6ms(~10%) app startup time


firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -56,7 +56,6 @@
             // components that are not required during initialization so as not to force initialize
             // them at app startup (refer
             // https://github.com/google/guice/wiki/InjectingProviders#providers-for-lazy-loading).
-            .eagerInDefaultApp()
```
vkryachko, line 59: Can you update the comment above to reflect that the SDK is no longer eager?

jeremyjiang-dev: Removed.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -85,8 +85,9 @@ public static AppStateMonitor getInstance() {
 
   private ApplicationProcessState mCurrentState = ApplicationProcessState.BACKGROUND;
 
-  private Set<WeakReference<AppStateCallback>> mClients =
+  private Set<WeakReference<AppStateCallback>> appStateClients =
```
visumickey, line 88: Can we use the term subscribers? may be activityStateSubscribers, appStateSubscribers? Client is too generic a term.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -254,16 +256,43 @@ public void registerForAppState(WeakReference<AppStateCallback> client) {
...
+  /**
+   * Register a client to receive app cold start update.
+   *
+   * @param client an AppColdStartCallback instance.
+   * @hide
+   */
+  /** @hide */
+  public void registerForAppColdStart(AppColdStartCallback client) {
+    synchronized (appStateClients) {
+      appColdStartClients.add(client);
```
visumickey, line 273: Should we check if the subscriber is already registered before adding to the list?

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -254,16 +256,43 @@ public void registerForAppState(WeakReference<AppStateCallback> client) {
...
+  /**
+   * Unregister the client to stop receiving app cold start update.
+   *
+   * @param client an AppColdStartCallback instance.
+   * @hide
+   */
+  /** @hide */
+  public void unregisterForColdStart(AppColdStartCallback client) {
+    synchronized (appStateClients) {
+      appColdStartClients.remove(client);
```
visumickey, line 286: Should we check if the client is already in the list before attempting to remove? What happens if the client is not there - Would it lead it a crash?

firebase-perf/src/test/java/com/google/firebase/perf/application/AppStateMonitorTest.java
```diff
@@ -722,6 +723,35 @@ public void activityStateChanges_multipleClients_callbackCalledOnEachClient() {
     assertThat(clientState.get(client3)).isEqualTo(ApplicationProcessState.BACKGROUND);
   }
 
+  @Test
```
visumickey, line 726: Can we add test cases for the PubSub setup? Eg. adding subscriber multiple times, removing multiple times to make sure we do not crash?

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -207,6 +208,7 @@ public synchronized void onActivityResumed(Activity activity) {
       updateAppState(ApplicationProcessState.FOREGROUND);
       if (mIsColdStart) {
         // case 1: app startup.
+        sendAppColdStartUpdate();
```
ramanpreetSinghKhinda, line 211: Although I have concerns on how we measure our cold startup time and I don't think initializing FirebasePerformance here will actually solve the startup time issue. However, I do agree with the current issue with dropbox and the traces provided by them (_which does not include `onResume()` as part of the measurements which is strange to me_) and this might help solve their use-case but not for all.

As discussed for now it's ok to rely on cold startup callback here but this need to change.

ramanpreetSinghKhinda (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
ramanpreetSinghKhinda: Since we are adding to a `Set` it won't be problem.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
ramanpreetSinghKhinda: It won't crash https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html#remove(java.lang.Object)

However, I don't think we are unregistering it anywhere? May I am missing it.

jeremyjiang-dev: We are not unregistering it anywhere. I added this method just for completion. Should I remove this method?

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
jeremyjiang-dev: Agree. The main reason that I include initialization code here is because our Fireperf's definition of [app startup time](https://firebase.google.com/docs/perf-mon/app-start-foreground-background-traces?platform=android#app-start), which does not count onResume() for app startup time. We might need to update how we measure app startup time based on Android vitals's definition.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
jeremyjiang-dev: To answer my own question, dead code is bad "Engineers spend time reading it for no value. They spend additional time trying to puzzle out why it's there, until they convince themselves there is no reason." (go/deadcode). So I will delete this method.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -254,16 +256,43 @@ public void registerForAppState(WeakReference<AppStateCallback> client) {
...
-      mClients.remove(client);
+    synchronized (appStateClients) {
+      appStateClients.remove(client);
+    }
+  }
+
+  /**
+   * Register a client to receive app cold start update.
+   *
+   * @param client an AppColdStartCallback instance.
```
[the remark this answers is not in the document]
jeremyjiang-dev: Done.

firebase-perf/src/test/java/com/google/firebase/perf/application/AppStateMonitorTest.java
jeremyjiang-dev: Added two more test case:
 - appColdStart_multipleSubscriber_callbackIsCalled
 - appColdStart_singleSubscriberRegistersForMultipleTimes_oneCallbackIsCalled

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
ramanpreetSinghKhinda: Filed b/183657588 for this.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -52,11 +52,6 @@
             .add(Dependency.required(FirebaseInstallationsApi.class))
```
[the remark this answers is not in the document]
visumickey: +1 But think if this is worth the optimization. And quite possible that we can iterate on this in a future iteration.

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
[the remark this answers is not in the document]
ramanpreetSinghKhinda: Filed http://b/183665997 to be taken care in future.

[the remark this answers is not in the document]
jeremyjiang-dev: Thank you, Raman!

PR #2524, opened 2021-03-18
Make ABT Dynamic-Module compliant.

vkryachko:
Changes the dependency on analytics to an optional Provider.

### DIFF
--- a/firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
+++ b/firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
@@ -23,6 +23,7 @@ import androidx.annotation.VisibleForTesting;
 import androidx.annotation.WorkerThread;
 import com.google.firebase.analytics.connector.AnalyticsConnector;
 import com.google.firebase.analytics.connector.AnalyticsConnector.ConditionalUserProperty;
+import com.google.firebase.inject.Provider;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.util.ArrayDeque;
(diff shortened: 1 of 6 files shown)


davidmotson (approved):

firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
```diff
@@ -350,14 +354,17 @@ private void throwAbtExceptionIfAnalyticsIsNull() throws AbtException {
...
+      connector.clearConditionalUserProperty(
+          experimentId, /*clearEventName=*/ null, /*clearEventParams=*/ null);
+    }
   }
 
   @WorkerThread
   private int getMaxUserPropertiesInAnalytics() {
     if (maxUserProperties == null) {
-      maxUserProperties = analyticsConnector.getMaxUserProperties(originService);
+      maxUserProperties = analyticsConnector.get().getMaxUserProperties(originService);
```
danasilver, line 367: Thinking: `@WorkerThread` methods don't need to check for null?

firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
```diff
@@ -333,11 +334,14 @@ private void removeExperiments(Collection<ConditionalUserProperty> experiments)
   }
 
   private void addExperimentToAnalytics(ConditionalUserProperty experiment) {
-    analyticsConnector.setConditionalUserProperty(experiment);
+    AnalyticsConnector connector = this.analyticsConnector.get();
+    if (connector != null) {
```
danasilver, line 338: Thinking: this is just a safety check, but `connector` will never be null here (and at the other checks) since `throwAbtExceptionIfAnalyticsIsNull()` is called first in the public methods

danasilver (approved):

firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
vkryachko: good point, removed to be consistent with previous behavior

firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
vkryachko: The check is not required for the same reason you mentioned below, the check is done prior to calling this method

PR #2526, opened 2021-03-18
Move the NetworkRequestBuilder package inside metrics to unify the builders

visumickey:


firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
```diff
@@ -1,9 +1,9 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
 //
```
vkryachko, line 2: Worth removing the double `@hide` javadoc comment below?

vkryachko (approved):

PR #2527, opened 2021-03-19
Create a separate package for Application Handlers

visumickey:


jeremyjiang-dev (approved):
LGTM. I wonder if we should rename `resource` folder to `resourcetype`. `resouce` folder makes me think about Android resource, such as layout.

visumickey:
> LGTM. I wonder if we should rename `resource` folder to `resourcetype`. `resouce` folder makes me think about Android resource, such as layout.

Interesting comment. Good point. May be as a part of the code convention change, we can call it as MetricType that would make it easier to understand. I agree resource is quite confusing.

PR #2532, opened 2021-03-22
Create native proxy object to check provider on each call and mediate…

davidmotson:
… between desired return value and default behavior)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/ProviderProxyNativeComponent.java
```diff
@@ -15,37 +15,60 @@
...
+
+  public ProviderProxyNativeComponent(Provider<CrashlyticsNativeComponent> provider) {
+    this.provider = provider;
+  }
+
   @Override
   public boolean hasCrashDataForSession(@NonNull String sessionId) {
+    if (provider.get() != null) {
+      return provider.get().hasCrashDataForSession(sessionId);
+    }
```
vkryachko, line 36: I recommend caching the result of `get()` instead of calling into it twice. This will make code more readable and more conventional as per [Injecting Providers](https://github.com/google/guice/wiki/InjectingProviders#injecting-providers) where in more traditional DI setups each call to `get()` could return a newly created instance if the component is not a singleton.

While all Firebase components are singletons, it could be beneficial to be consistent with common conventions for future readers of this code(including ourselves).

```suggestion
    Foo native = provider.get();
    if (native != null) {
      return native.hasCrashDataForSession(sessionId);
    }
```

Here and throughout

vkryachko (commented):
Could we add a test for the proxy where when the provider is filled with a value we would see the expected call into the native component?

vkryachko (approved):

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/ProviderProxyNativeComponentTest.java
```diff
@@ -0,0 +1,98 @@
...
+  private static final long TEST_LONG = 123;
+  private static final boolean TEST_BOOLEAN = true;
+  private static final int TEST_INT = 1234;
+
+  @Mock private CrashlyticsNativeComponent component;
+
+  @Before
+  public void setUp() throws Exception {
+    MockitoAnnotations.initMocks(this);
+  }
```
vkryachko, line 36: nit: for this simple case you can just use the more concise:
```suggestion
  private final CrashlyticsNativeComponent component = mock(CrashlyticsNativeComponent.class);
```

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/ProviderProxyNativeComponentTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+import static org.mockito.ArgumentMatchers.eq;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+public class ProviderProxyNative {
```
vkryachko: Please add a `@RunWith(AndroidJUnit4.class)`

PR #2533, opened 2021-03-23
Organize the access specifiers for the config package classes according to the package requirements.

visumickey:


jeremyjiang-dev (commented):
Two questions:
1. Can we make `ConfigResolver`, `DeviceCacheManager`, and `RemoteConfigManager` as final class. According to [Effective Java](https://effectivejava.googleplex.com/third-edition#page=114), we should design and document for inheritance or else prohibit it.
2. If we do not design for inheritance for the above classes, we need to change `protected` to default package access.

Issue #2539, opened 2021-03-23
FR: Scope Firestore SnapshotListeners to a LifecycleOwner


rosariopfernandes:
## What feature would you like to see?

I would like to be able to scope a Firestore SnapshotListener to a [LifecycleOwner](https://developer.android.com/reference/androidx/lifecycle/LifecycleOwner) the same way we're able to scope a Listener to an [Activity](https://developer.android.com/reference/android/app/Activity):

```kotlin
val query: Query = Firebase.firestore.collection("restaurants")
query.addSnapshotListener(viewLifecycleOwner) { snapshot, error ->
    // do something
}
```

### 1. As Fragment scoped listeners

One of the recommended best-practices on Modern Android Development is to let a [single Activity host multiple fragments](https://developer.android.com/guide/navigation/navigation-getting-started). 
It would be nice if we could, for example bind a SnapshotListener to Fragment A, and when we navigate to Fragment B, the listener from A gets removed so we can add a different listener in fragment B.

Note that Fragments provide a [getViewLifecycleOwner()](https://developer.android.com/reference/androidx/fragment/app/Fragment#getViewLifecycleOwner()) method that already allows developers to scope other kinds of listeners (eg. [LiveData#observe()](https://developer.android.com/reference/androidx/lifecycle/LiveData#observe(androidx.lifecycle.LifecycleOwner,%20androidx.lifecycle.Observer%3C?%20super%20T%3E))) to the [Fragment's View](https://developer.android.com/reference/androidx/fragment/app/Fragment#getView()) lifecycle, which means the listener is automatically removed once the fragment is no longer visible (user navigated away, for example).

### 2. As custom components scoped listeners

The [androidx.lifecycle](https://developer.android.com/reference/androidx/lifecycle/package-summary) package allows developers to [create custom lifecycle-aware components](https://developer.android.com/topic/libraries/architecture/lifecycle).
With the feature proposed here, developers would be able to easily scope snapshot listenerers to their custom components.

### 3. To (slightly) improve [FirebaseUI-Android](https://github.com/firebase/FirebaseUI-Android)?

FirebaseUI has its own implementation of the feature proposed here. It prompts the user to provide a `LifecycleOwner`, and then removes the snapshot listener when the `LifecycleOwner`'s `ON_STOP` Event is called:

[...]

PR #2541, opened 2021-03-24
feat(firestore): scope snapshot listeners to a LifecycleOwner

thatfiredev:
This PR should implement the Feature Requested in #2539. It introduces the following public APIs:
- `Query.addSnapshotListener(LifecycleOwner, EventListener)`
- `DocumentReference.addSnapshotListener(LifecycleOwner, EventListener)`
- `FirebaseFirestore.addSnapshotsInSyncListener(LifecycleOwner, EventListener)`

I was trying to follow the same approach of [ActivityScope](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/core/ActivityScope.java), so I created a new `LifecycleOwnerScope` class under the `firestore.core` package.


wu-hui:
First of all, thank you so much for your contribution! I'd like to see this gets merged.

This is a public API change however, which means this needs to go through internal review process. I will try to make this happen, but it might take 1-2 week.

Also, I hope it is not too much to ask, but can you add a test to `src/androidTest/java/com/google/firebase/firestore/ListenerRegistrationTest.java` while we are waiting for the internal review.

rosariopfernandes:
@wu-hui Thanks for taking this through the internal API review :)

Sure, I'll add a test for it

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java
```diff
@@ -455,7 +459,39 @@ public ListenerRegistration addSnapshotListener(
...
+   * <p>The listener will be automatically removed during {@link LifecycleOwner's ON_STOP Event'}.
+   *
+   * @param lifecycleOwner The LifecycleOwner to scope the listener to.
+   * @param metadataChanges Indicates whether metadata-only changes (i.e. only {@code
+   *     DocumentSnapshot.getMetadata()} changed) should trigger snapshot events.
+   * @param listener The event listener that will be called with the snapshots.
+   * @return A registration object that can be used to remove the listener.
+   */
+  @NonNull
+  public ListenerRegistration addSnapshotListener(
```
wu-hui, line 482: I think we also need a 

```
public ListenerRegistration addSnapshotListener(
      @NonNull LifecycleOwner lifecycleOwner, @NonNull EventListener<DocumentSnapshot> listener)
```

because there is an equivalent for `Activity`.

rosariopfernandes: Nice catch! Just added.

wu-hui: Actually, one of my colleague pointed out that it might be a compilation error to try to pass `AppCompatActivity` here, because it is both `LifeCycleOwner` and `Activity`, and I just confirmed it.

I don't see an easy way out unfortunately..

rosariopfernandes: @wu-hui I forgot `AppCompatActivity` is both... I think the compilation error might occur in all of the APIs added here. Developers will need to explicitly cast it (eg. `addSnapshotListener((LifecyleOwner) appCompatActivity, listener)`)

A possible way out would be exposing one more method that takes `AppCompatActivity` as an argument and internally we would cast it to LifecycleOwner. But the problem with this is that it'll require the Firestore SDK to depend on `androidx.appcompat`.

wu-hui: Right. I think neither will work perfectly.

To accommodate `LifeCycleOwner`, we might have to consider revamping the API around life cycle management for listener. I will try to make a proposal internally first. 

This does mean I cannot merge this PR as-is. Will check back with you when we have a clear idea.

PR #2543, opened 2021-03-24
Improve Fireperf SDK startup time.

jeremyjiang-dev:
- Removed console logging in `PerfSession.create()`, which reduces the startup time by 15 ms (~25% of total startup time)
   - Side effect: There will no logging for logging verbose/non-verbose session. However, the logging provides little value to developers and developers can easily add logging by themselves.
 - Delay the initialization of `applicationInfoBuilder`, which reduces the startup time by 8 ms (~20% of total startup time)
   - Side effect: None

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/session/PerfSession.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/session/PerfSession.java
@@ -50,11 +50,6 @@ public class PerfSession implements Parcelable {
     PerfSession session = new PerfSession(sessionId, new Clock());
     session.setGaugeAndEventCollectionEnabled(shouldCollectGaugesAndEvents());
 
-    AndroidLogger.getInstance()
-        .debug(
-            "Creating a new %s Session: %s",
-            session.isVerbose() ? "Verbose" : "Non Verbose", sessionId);
-
     return session;
   }
 
--- a/firebase-perf/src/main/java/com/google/firebase/perf/session/PerfSession.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/session/PerfSession.java
@@ -20,7 +20,6 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import com.google.firebase.perf.config.ConfigResolver;
-import com.google.firebase.perf.logging.AndroidLogger;
 import com.google.firebase.perf.util.Clock;
 import com.google.firebase.perf.util.Timer;
 import com.google.firebase.perf.v1.SessionVerbosity;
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/session/PerfSession.java
```diff
@@ -50,11 +49,6 @@ public static PerfSession create() {
     PerfSession session = new PerfSession(sessionId, new Clock());
     session.setGaugeAndEventCollectionEnabled(shouldCollectGaugesAndEvents());
 
-    AndroidLogger.getInstance()
-        .debug(
-            "Creating a new %s Session: %s",
-            session.isVerbose() ? "Verbose" : "Non Verbose", sessionId);
-
```
ramanpreetSinghKhinda, line 57: I assume this is not the only place where we are logging during app startup. Even if it is simply removing the log might solve our current use-case but not a clean solution IMO.

My thought is to update the way we log so instead of doing this:

```
logger.debug("This is a formatted log with String (%s), Integer (%d), Boolean (%b) and Float (%f)", "MyString", 10, true, 0.4);
````

You can do something like this:

**AndroidLogger.java**

```
    public interface LogFunction {
        String get();
    }


    private void debug(LogFunction logFunction, Object... args) {
        if(isLogcatEnabled) {
            Log.d(LOG_TAG, String.format(Locale.ENGLISH, function.get(), args));
        }
    } 
```

**Use Case**

```
logger.debug(() -> "This is a formatted log with String (%s), Integer (%d), Boolean (%b) and Float (%f)", "MyString", 10, true, 0.4);

```

The benefit it will have is:
 - You don't have to remove any log
 - This will make the entire log line (including string object creation) to happen only if `isLogcatEnabled` is `true`

jeremyjiang-dev: Thanks for the great suggestion! Discussed offline. This is a great suggestion of how we can save logging time. The benefit is that we will create string objects only when we enable debug logging. 
For this particular use case for logging verbose/non-verbose session, since it provides little value and action items to developers, we will drop this log.

visumickey: Great suggestion Raman! I love this idea and approach. Curious as to how much is this log line impacting our startup time. May be this is a better refactor we can do across our SDK to make our logging efficient and cheaper? (Instead of doing this short change just for now?)

jeremyjiang-dev: According to the analysis Raman did, this logging takes 15 ms (25% of total startup time). Raman is creating a bug to refactor logging using this new approach.

ramanpreetSinghKhinda: I filed b/183635842 to update logging specifically.

ramanpreetSinghKhinda (approved):

PR #2544, opened 2021-03-24
Cleaning up unnecessary use of "@hide annotations" (b/177448055)

ramanpreetSinghKhinda:
## Few Points:

* We should be using `@hide` javadoc annotation to hide certain packages/classes/methods (_which are declared public_) to be considered a part of public API.

* Double use of `@hide` javadoc annotation is not required. This was a requirement with GmsCore (go/gmscore-hide-annotation-lsc). Also that second annotation was `@Hide` Java annotation (_which is a GmsCore annotation_) and not `@hide` javadoc annotation (_which is a general javadoc annotation_) which seems like happened during code migration to GitHub.

* This annotation should only be used in public classes/methods.

* Adding this annotation to the **method** will hide that method.

* Adding this annotation to the **class level** will hide the entire class and its public methods.

* Adding this annotation to the **package** declaration in the`package-info.java` class makes the entire classes within that package to be hidden. Note that this does not work transitively for child packages and so each child package would also need there own `package-info.java` class.

* There's no way currently to not hide some of the classes/methods when the entire package is marked hidden. So in that case where a package contains some classes/methods which has to be public than we have to apply this annotation class/method wise. 

## What does `hiding` means?
 - No javadoc will be generated for that package/class/method
 - api.txt won't get updated for that package/class/method

## FAQs

 - Does this PR trigger any public api change?
   No, I have already validated everything by running the below command and no change was detected (_**caveat:** sometimes you have to touch the `api.txt` file to trigger the run_).

  ```
  firebase-android-sdk$./gradlew firebase-perf:generateApiTxtFile
  ```

WANT_LGTM = @vkryachko

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -423,7 +423,6 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
   }
 
   /** An interface to be implemented by subscribers which needs to receive app state update. */
-  /** An interface to be implemented by clients which needs to receive app state update. */
   public static interface AppStateCallback {
     public void onUpdateAppState(ApplicationProcessState newState);
   }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -45,13 +45,7 @@ import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
-/**
- * Trace timer implementation to send foreground and background session log.
- *
- * @hide
- */
-
-/** @hide */
+/** Trace timer implementation to send foreground and background session log. */
 public class AppStateMonitor implements ActivityLifecycleCallbacks {
 
   private static final AndroidLogger logger = AndroidLogger.getInstance();
@@ -125,8 +119,6 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
     }
   }
 
-  /** @hide */
-  /** @hide */
   public void incrementCount(@NonNull String name, long value) {
     // This method is called by RateLimiter.java when a log exceeds rate limit and to be dropped
     // It can be on any thread. sendSessionLog() method is called in callback methods from main UI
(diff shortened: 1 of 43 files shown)


jeremyjiang-dev (approved):
Thanks for the detailed PR description, which makes my review very easy! I recently merged `FirebasePerformanceInitializer.java` and `AppStateMonitor.java`, could you do a sanity check there?

ramanpreetSinghKhinda:
Merge this PR after https://github.com/firebase/firebase-android-sdk/pull/2545.

Reason: https://github.com/firebase/firebase-android-sdk/pull/2546#issuecomment-807487098

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -452,23 +422,13 @@ private boolean hasFrameMetricsAggregatorClass() {
...
   }
 
-  /**
-   * An interface to be implemented by subscribers which needs to receive app state update.
-   *
-   * @hide
-   */
-  /** @hide */
+  /** An interface to be implemented by subscribers which needs to receive app state update. */
+  /** An interface to be implemented by clients which needs to receive app state update. */
```
visumickey, line 426: Remove this line since this is redundant.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
```diff
@@ -507,37 +495,32 @@ protected void finalize() throws Throwable {
     }
   }
 
-  /** @hide */
   /** @hide */
```
visumickey, line 498: Do we need this? Why do we have two @hide annotations at times?

vkryachko (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
ramanpreetSinghKhinda: Ahh. Merge issues.

But Good catch!

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
ramanpreetSinghKhinda: Looks like you are looking at the older code.

visumickey (approved):

PR #2545, opened 2021-03-25
Updating variable names from the source classes to follow Java's code convention (b/177441811)

ramanpreetSinghKhinda:
## What this PR entails?
 - Refer b/177441811 for more details.
 - This PR is only for updating source classes and not the test classes.

## How to review this PR?
 - Look at the class level variable names and if that looks good to you then just move on to other class. Reason being the rest of the changes in the class are most likely the name refactoring change which are automatically done by the AndroidRefactoring tool and thus I don't expect any issues there.

## Is this an exhaustive change?
 - Not necessarily but this is a good first step to build upon. There could be many different ways of declaring variable names but I chose the one which is easy to follow IMHO.
 - Also there's so much more to improve on but I decided to just update class level variable names (_exception being `context` -> `appContext`_)

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
@@ -61,7 +61,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
   private final GaugeManager gaugeManager;
   private final TransportManager transportManager;
 
-  private final NetworkRequestMetric.Builder networkRequestBuilder = NetworkRequestMetric.newBuilder();
+  private final NetworkRequestMetric.Builder networkMetricBuilder = NetworkRequestMetric.newBuilder();
 
   @Nullable private String userAgent;
 
@@ -133,7 +133,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
       url = Utils.stripSensitiveInfo(url);
 
       // Limits the URL length to 2000 chars.
-      networkRequestBuilder.setUrl(Utils.truncateURL(url, Constants.MAX_URL_LENGTH));
+      networkMetricBuilder.setUrl(Utils.truncateURL(url, Constants.MAX_URL_LENGTH));
     }
 
     return this;
@@ -141,7 +141,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
 
   /** Gets the url for the current {@link NetworkRequestMetric}. */
   public String getUrl() {
-    return networkRequestBuilder.getUrl();
+    return networkMetricBuilder.getUrl();
   }
 
   /** Sets the user-agent for the current network request. */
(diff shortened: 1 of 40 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/network/InstrURLConnectionBase.java
```diff
@@ -38,11 +38,11 @@
...
-  private final HttpURLConnection mHttpUrlConnection;
-  private final NetworkRequestMetricBuilder mBuilder;
-  private long mTimeRequested = -1;
-  private long mTimeToResponseInitiated = -1;
-  private final Timer mTimer;
+  private final HttpURLConnection httpUrlConnection;
+  private final NetworkRequestMetricBuilder networkMetricBuilder;
+  private long timeRequested = -1;
+  private long timeToResponseInitiated = -1;
+  private final Timer timer;
```
jeremyjiang-dev, line 45: nit: We may want to group variables by modifiers and static/non-static variable for better readability. Here, we can group `private final` varibles together. No actions needed to check variables in other files because it is out of the scope of this PR, but this is something we can do when we examine the variables :)

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
```diff
@@ -61,7 +61,7 @@
   private final GaugeManager gaugeManager;
   private final TransportManager transportManager;
 
-  private final NetworkRequestMetric.Builder networkRequestMetricBuilder =
+  private final NetworkRequestMetric.Builder networkMetricBuilder =
```
jeremyjiang-dev, line 64: Should we call it builder? According to suggestions in the ticket

firebase-perf/src/main/java/com/google/firebase/perf/util/Timer.java
```diff
@@ -23,20 +23,20 @@
 /** A Timer class provides both wall-clock (epoch) time and high resolution time (nano time). */
 public class Timer implements Parcelable {
   /** Wall-clock time or epoch time in microseconds, */
-  private long mTime;
+  private long time;
```
jeremyjiang-dev, line 26: `timeInMicros`? According to your suggestions in the ticket, all variables should have contextual names.

firebase-perf/src/main/java/com/google/firebase/perf/network/InstrURLConnectionBase.java
ramanpreetSinghKhinda: Done!

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
ramanpreetSinghKhinda: Done!

firebase-perf/src/main/java/com/google/firebase/perf/util/Timer.java
ramanpreetSinghKhinda: Done!

ramanpreetSinghKhinda:
Submitting this PR as this contains with fix for Master (see https://github.com/firebase/firebase-android-sdk/pull/2546#issuecomment-807487098)

PR #2546, opened 2021-03-25
Updating variable names from the test classes to follow Java's code convention (b/177441811)

ramanpreetSinghKhinda:
## What this PR entails?
 - Refer b/177441811 for more details.
 - This PR is only for updating test classes and not the source classes.

## How to review this PR?
 - Look at the class level variable names and if that looks good to you then just move on to other class. Reason being the rest of the changes in the class are most likely the name refactoring change which are automatically done by the AndroidRefactoring tool and thus I don't expect any issues there.

## Is this an exhaustive change?
 - Not necessarily but this is a good first step to build upon. There could be many different ways of declaring variable names but I chose the one which is easy to follow IMHO.
 - Also there's so much more to improve on but I decided to just update class level variable names

### DIFF
--- a/firebase-perf/src/test/java/com/google/testing/timing/FakeDirectExecutorService.java
+++ b/firebase-perf/src/test/java/com/google/testing/timing/FakeDirectExecutorService.java
@@ -21,7 +21,7 @@ import java.util.concurrent.TimeUnit;
 /** Fake executor service that simply executes the runnable in the same thread. */
 public class FakeDirectExecutorService extends AbstractExecutorService {
 
-  private boolean IsShutdown = false;
+  private boolean isShutdown = false;
 
   public void execute(Runnable command) {
     command.run();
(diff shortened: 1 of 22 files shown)


firebase-perf/src/test/java/com/google/testing/timing/FakeDirectExecutorService.java
```diff
@@ -21,7 +21,7 @@
 /** Fake executor service that simply executes the runnable in the same thread. */
 public class FakeDirectExecutorService extends AbstractExecutorService {
 
-  private boolean shutdown = false;
+  private boolean IsShutdown = false;
```
jeremyjiang-dev, line 24: nit: isShutdown

jeremyjiang-dev (approved):

firebase-perf/src/test/java/com/google/testing/timing/FakeDirectExecutorService.java
ramanpreetSinghKhinda: Done

ramanpreetSinghKhinda:
This PR created a few conflicts when merging to Master. This has been resolved in https://github.com/firebase/firebase-android-sdk/pull/2545.

PR #2562, opened 2021-04-01
Fix personalization test and add check for optionalProvider.

rlazo:
The personalization test was using the default namespace, and because another instance of config already existed in that namespace, the test instance was not being created. This caused the personalization test to try to use real objects instead of the mocks, including the direct executor. This PR fixes this and uses a provider to fake the interaction when Analytics is present and when it's not.

### DIFF
--- a/firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
+++ b/firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.remoteconfig;
 
+import static androidx.test.ext.truth.os.BundleSubject.assertThat;
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.Truth.assertWithMessage;
 import static com.google.firebase.remoteconfig.AbtExperimentHelper.createAbtExperiment;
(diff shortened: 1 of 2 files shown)


firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -990,15 +999,45 @@ public void personalization_hasMetadata_successful() throws Exception {
                       any(Bundle.class));
               assertThat(fakeLogs).hasSize(2);
 
-              Bundle params1 = new Bundle();
-              params1.putString(Personalization.EXTERNAL_RC_PARAMETER_PARAM, "id1");
-              params1.putString(Personalization.EXTERNAL_ARM_VALUE_PARAM, "value1");
-              assertThat(fakeLogs.get(0).toString()).isEqualTo(params1.toString());
+              assertThat(fakeLogs.get(0))
+                  .string(EXTERNAL_PERSONALIZATION_ID_PARAM)
```
vkryachko, line 1003: TIL, cool library!

vkryachko (commented):
I wonder if the perf test failure a flake or is caused by the change in RC

rlazo:
> I wonder if the perf test failure a flake or is caused by the change in RC

Flake

vkryachko (approved):
LGTM, but let's wait for @danasilver to review before merging.

rlazo:
SGTM

danasilver (approved):

PR #2408, opened 2021-02-04
Android Support Library / AndroidX problem. Fix: Upgrade Picasso to avoid android support library dep

blundell:
**PROBLEM**

Firebase In App Messaging still has a dependency on the support library, when this can clash with AndroidX.

```
  /home/blundell/.../build.gradle: 
  Error: 
  
  Dependencies using groupId com.android.support and androidx.* can not be combined but found

  IdeMavenCoordinates
  {
	  myGroupId='com.android.support', 
	  myArtifactId='exifinterface', 
	  myVersion='27.1.0', 
	  myPacking='aar', 
	  myClassifier='null'
  } 

  and 
  
  IdeMavenCoordinates
  {
	  myGroupId='androidx.localbroadcastmanager', 
	  myArtifactId='localbroadcastmanager', 
	  myVersion='1.0.0', 
	  myPacking='aar', 
	  myClassifier='null'
  } 

  incompatible dependencies [GradleCompatible]

```
**INVESTIGATION**

Getting the dependency graph for the module that I had imported Firebase In App Messaging into (`./gradlew libraries:remoteMessaging:dependencies`) shows how FIAM depends on FIAM-display, which depends on Picasso, which depends on a android.support library. Seen here:

```
+--- com.google.firebase:firebase-inappmessaging-display-ktx:19.1.4
|    +--- androidx.annotation:annotation:1.1.0 -> 1.2.0-alpha01
|    +--- com.google.firebase:firebase-common:19.5.0 (*)
|    +--- com.google.firebase:firebase-common-ktx:19.5.0 (*)
|    +--- com.google.firebase:firebase-components:16.1.0 (*)
|    +--- com.google.firebase:firebase-inappmessaging:19.1.4
...
|    +--- com.google.firebase:firebase-inappmessaging-display:19.1.4
|    |    +--- androidx.appcompat:appcompat:1.1.0 -> 1.2.0
|    |    |    +--- androidx.annotation:annotation:1.1.0 -> 1.2.0-alpha01
|    |    |    +--- androidx.core:core:1.3.0 -> 1.5.0-alpha04 (*)
|    |    |    +--- androidx.cursoradapter:cursoradapter:1.0.0
|    |    |    |    \--- androidx.annotation:annotation:1.0.0 -> 1.2.0-alpha01
|    |    |    +--- androidx.fragment:fragment:1.1.0 (*)
|    |    |    +--- androidx.appcompat:appcompat-resources:1.2.0
...
[...]

Issue #1736, opened 2020-06-29
161731384: Move to Picasso that has no support deps after it is released


ubuntudroid:
## What feature would you like to see?
Right now you need to enable jetifier if you pull in the firebase-inappmessaging-display dependency as reported in #535 . This is due to a dependency on picasso which pulls in com.android.support:exifinterface:

```
com.android.support:exifinterface:27.1.0
\--- com.squareup.picasso:picasso:2.71828
     +--- debugCompileClasspath (requested com.squareup.picasso:picasso:{strictly 2.71828})
     \--- com.google.firebase:firebase-inappmessaging-display:19.0.7
          +--- debugCompileClasspath (requested com.google.firebase:firebase-inappmessaging-display)
          \--- com.google.firebase:firebase-bom:25.5.0
               \--- debugCompileClasspath
```

However, enabling jetifier, while undoubtedly handy in cases like these, unnecessarily slows down builds.

## How would you use it?

I'd disable jetifier and enjoy faster builds. 🚀

ashwinraghav:
Thanks for raising @ubuntudroid 
This seems like a valid concern given. We've [made the effort](https://android-developers.googleblog.com/2019/06/google-play-services-and-firebase.html#:~:text=Later%20this%20year%2C%20the%20Google,Jetpack%20features%20in%20your%20app.) to move away our direct dependencies on support over to jetpack libraries and this makes that effort futile.

I bet there are image processing libraries that don't have similar problems. The scope of the fix is really limited to migrating the dependency on Picasso out to some other library. The migration is tightly scoped to this one file if you are inclined to send a fix our way 🥇 

https://github.com/firebase/firebase-android-sdk/blob/d9ed980c39c58ddf55ab945baa120c2e0a9fa8cb/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/internal/FiamImageLoader.java#L31

ubuntudroid:
@ashwinraghav From what I see there are two options:

1. Stay with Picasso and wait for 3.0.0 to be released (no ETA from what I know though). Their SNAPSHOT version has already been migrated to androidx - however I'm pretty sure, using snapshot versions is out of the question for a widely used library as the Firebase SDK.
2. Migrate to some other library (Glide anyone?) as you suggested.

ashwinraghav:
[...]

PR #2578, opened 2021-04-09
androidx support: Update Picasso

jeremyrempel:
## What
* Update picasso to latest version

## Why
* Current version of picasso depends on `com.android.support:support-annotations:27.1.0` and `com.android.support:exifinterface:27.1.0`. This is blocking removing jetifier


rosariopfernandes:
This seems like a duplicate of #2408

Also see the [comment on #1736](https://github.com/firebase/firebase-android-sdk/issues/1736#issuecomment-719938843) for a workaround.

PR #2602, opened 2021-04-21
Add ApplicationExitInfo to CrashlyticsReport

tejasd:


mrichards (approved):
LGTM, though I'm a little dicey on AutoValue in general so if you aren't sure about that part, we should fine someone who knows it better.

PR #2610, opened 2021-04-22
 Disable fetching of Firebase Installation Id when data collection is disabled

mrichards:
Querying for the Firebase Installation Id makes a network call and potentially registers
new information to Firebase servers, so we should not access it from Crashlytics when
the data collection flag is disabled. Instead, we cache a synthetic FID to use as long
as collection is disabled. This can result in over-counting unique users who have experienced
a crash, but it better aligns with the privacy goals of the data collection flag.
    
See internal doc go/crashlytics-fid for more details.


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/ProviderProxyNativeComponentTest.java
```diff
@@ -19,6 +19,7 @@
 import androidx.test.runner.AndroidJUnit4;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
```
mrichards, line 22: Unrelated fix for compiler error when building the android tests. Full fix in https://github.com/firebase/firebase-android-sdk/pull/2609

mrober (approved):

PR #2611, opened 2021-04-22
Add test apps template and fireci commands to measure sdk startup times.

yifanyang:


vkryachko (commented):
My understanding was that mactobenchmark would contain:
* config.yaml
* template/ dir

However it contains many files that look like the could be generated, like the base gradle build as well as gradle wrapper itself(can we use gradle wrapper from the root directory)?

macrobenchmark/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+        google()
+        mavenCentral()
+    }
+    dependencies {
+        classpath "com.android.tools.build:gradle:4.2.0-rc01"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+
+        // NOTE: Do not place your application dependencies here; they belong
+        // in the individual module build.gradle files
+
```
rlazo, line 27: rm empty line

rlazo (approved):

ci/fireci/fireciplugins/startup_time.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@ci_command()
+def startup_time():
+  """Measures app startup times for Firebase SDKs."""
+
+  gradle.run('assembleAllForSmokeTests')
+  artifact_versions = _parse_artifacts()
+
+  os.chdir('macrobenchmark')
```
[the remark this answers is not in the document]
yifanyang: This file is now renamed to `macrobenchmark.py`. This method call is replaced with `with chdir('macrobenchmark'):`, at line 48.

macrobenchmark/build.gradle
yifanyang: This file is now moved to `macrobenchmark/template/build.gradle`. This exact blank line is now removed.

yifanyang:
@rlazo @vkryachko

I have updated this pull request with some changes:
- renamed `startup_time.py` to `macrobenchmark.py`
- removed gradle wrapper from src, but call gradle `wrapper` task to init at runtime
- executions of test app assembly and FTL uploading for different SDKs now run concurrently
- test apps modules are no longer subprojects of the same rootproject at `macrobenchmark` folder. Test apps modules of different SDKs are now subprojects to its own rootproject for each SDK

Can you help take another look at this pull request? Thanks!

ci/fireci/fireciplugins/macrobenchmark.py
```diff
@@ -0,0 +1,206 @@
...
+  with chdir('macrobenchmark'):
+    runners = [MacrobenchmarkTest(k, v, artifact_versions) for k, v in config.items()]
+    await asyncio.gather(*[x.run() for x in runners])
+
+  _logger.info('Macrobenchmark test finished.')
+
+
+async def _parse_artifact_versions():
+  proc = await asyncio.subprocess.create_subprocess_exec('./gradlew', 'assembleAllForSmokeTests')
+  await proc.wait()
```
rlazo, line 57: wouldn't line 56 await be enough? not sure

ci/fireci/fireciplugins/macrobenchmark.py
```diff
@@ -0,0 +1,206 @@
...
+
+  with open('build/m2repository/changed-artifacts.json', 'r') as json_file:
```
rlazo, line 59: 'r' is the default

ci/fireci/fireciplugins/macrobenchmark.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+async def _parse_config_yaml():
+  with open('macrobenchmark/config.yaml', 'r') as yaml_file:
+    return yaml.safe_load(yaml_file)
+
+
+async def _create_gradle_wrapper():
+  with open('macrobenchmark/settings.gradle', 'w'):
+    pass
```
rlazo, line 76: is this intended?

ci/fireci/fireciplugins/macrobenchmark.py
```diff
@@ -0,0 +1,206 @@
...
+_logger = logging.getLogger('fireci.macrobenchmark')
+
+
+@ci_command()
+def macrobenchmark():
+  """Measures app startup times for Firebase SDKs."""
+  asyncio.run(_launch_macrobenchmark_test())
+
+
+async def _launch_macrobenchmark_test():
```
vkryachko, line 39: What is the motivation for all the `async`s? not sure it's warranted for this simple use case

ci/fireci/fireciplugins/macrobenchmark.py
yifanyang: Sounds to me it is required. So if I don't `await` on it, it gives me this warning:
```
/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/unix_events.py:878: RuntimeWarning: A loop is being detached from a child watcher with pending handlers
  RuntimeWarning)
```
My understanding is that it means the underneath [`Future` object](https://docs.python.org/3/library/asyncio-future.html#future-object) may not be settled before my code exit. Also in the [python doc](https://docs.python.org/3/library/asyncio-subprocess.html) for asyncio subprocess, the first [example](https://docs.python.org/3/library/asyncio-subprocess.html#subprocesses) and the last [example](https://docs.python.org/3/library/asyncio-subprocess.html#examples) also `await` on `proc.communicate()` and `proc.wait()` respectively.

This also reminds me that I forgot to await at other places in the code. The problem was masked previously. I fixed in the new commit.

ci/fireci/fireciplugins/macrobenchmark.py
yifanyang: Done.

ci/fireci/fireciplugins/macrobenchmark.py
yifanyang: Yes. Without a `settings.gradle` file, `./gradlew wrapper` failed with this error:
```
FAILURE: Build failed with an exception.

* What went wrong:
Project directory '/Users/yifany/Documents/firebase/firebase-android-sdk/macrobenchmark' is not part of the build defined by settings file '/Users/yifany/Documents/firebase/firebase-android-sdk/settings.gradle'. If this is an unrelated build, it must have its own settings file.
```
And looks like an empty `settings.gradle` is good enough.

ci/fireci/fireciplugins/macrobenchmark.py
yifanyang: Here is how I come to all these async code:
1. gcloud calls takes long time, and I want to parallelize this part
2. now that test apps for different SDKs are isolated, to build them I need to run `./gradlew assemble assembleAndroidTest` one by one in each rootprojects for all SDKs (previously all test apps are subprojects of the same root project, so I only had to run one gradle command, and `org.gradle.parallel=true` takes care of the parallelization). Anyway I think I want to parallelize this part as well
3. with Popen and blocking wait, I'm able to make apk assembling and FTL uploading for different SDKs concurrent, respectively. But with async code, it's convenient for me to even make uploading for one SDK only depend on its corresponding assembling.
4. It also seems easier for me to handle stdout/stderr streaming (with color and sdk name decoration) to parent process from multiple subprocesses with async.
5. at this point, I feel like it's actually easier for me to write async code.

I do agree that it brings a lot of (unnecessarily) complexity. Maybe there are simpler ways to do it without async, and I just have not found them. What do you think? I'd like to have a quick GVC if you have time?

vkryachko: Actually, I take it back. It does seem to provide a lot of value, not the least of which is that it provides concurrency out of the box without having to deal with it explicitly and makes the code to read simple. Using asyncio sgtm

PR #2616, opened 2021-04-23
Update versions

rlazo:


vkryachko (approved):
Looks like `firebase-annotations` is not being updated

rlazo:
> Looks like `firebase-annotations` is not being updated

Fixed. This updated needed to be manual since it was the first release and the tooling expects the `latestReleasedVersion` string to exist.

vkryachko (approved):

PR #2617, opened 2021-04-23
Update default token refresh rate

rosalyntan:
Update the default token refresh rate:

`refresh time = IAT + (0.5*TTL) + 5 min`


malcolmdeck (approved):
Just checking, this presumes a minimum TTL of >10 minutes, yeah?

rosalyntan:
> Just checking, this presumes a minimum TTL of >10 minutes, yeah?

Yup, the min TTL should be at least 15 min, but I also have some handling for if we get a TTL that's unexpectedly small from the backend.

PR #2618, opened 2021-04-23
Update getTotalBytes() docs

schmidt-sebastian:
The method can return -1 for chunked/gzipped downloads. I didn't go into detail since the exact causes for this behavior are not clear to me.


schmidt-sebastian:
cc @cynthiajoan

schmidt-sebastian:
cc @rachelsaunders

PR #2619, opened 2021-04-23
Implement platform logging for App Check

rosalyntan:


appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
```diff
@@ -61,13 +63,15 @@
...
-    checkNotNull(projectId);
-    this.apiKey = apiKey;
-    this.appId = appId;
-    this.projectId = projectId;
+  public NetworkClient(@NonNull FirebaseApp firebaseApp) {
+    checkNotNull(firebaseApp);
+    this.firebaseAppCheck = (DefaultFirebaseAppCheck) FirebaseAppCheck.getInstance(firebaseApp);
+    this.apiKey = firebaseApp.getOptions().getApiKey();
+    this.appId = firebaseApp.getOptions().getApplicationId();
+    this.projectId = firebaseApp.getOptions().getProjectId();
```
vkryachko, line 71: Consider go/di-practices/inject-direct-dependencies

rosalyntan: Thanks for the link! I passed `FirebaseApp` directly into `NetworkClient` to avoid adding the `:appcheck:firebase-appcheck-interop` dependency to `firebase-appcheck-safetynet.gradle` and `firebase-appcheck-debug.gradle` (for accessing `FirebaseAppCheck#getInstance`).

I think it's fine to add the interop dependency, but it seems cleaner to only have that dependency in the core App Check SDK.

vkryachko: I see, one thing I would suggest is to consider injecting the NetworkClient via firebase components instead of having both `safetynet` and `debug` sdks creating instantiating it on their own. This would also allow you to register both of these sdks with the platform logging user-agent.

(Optional)

vkryachko (commented):
Overall LGTM,

Please get a review by @VinayGuthal before merging

VinayGuthal (approved):

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
rosalyntan: Ack, that makes sense! I've filed b/186549222 to follow up on this in a future PR.

Issue #2336, opened 2021-01-18
Documentation or nullability annotation of FirebaseDynamicLinks.getDynamicLink() intent parameter wrong


ubuntudroid:
The documentation of

https://github.com/firebase/firebase-android-sdk/blob/6ccc9de98aef8a559726c04a86ab3d47b35d6eed/firebase-dynamic-links/src/main/java/com/google/firebase/dynamiclinks/FirebaseDynamicLinks.java#L102

states, that the intent parameter

> can be null if the intent does not include the dynamic link. A non-null intent is necessary only when the app is launched directly using the dynamic link, such as when using App Links.

However the parameter is annotated with `@NonNull`. So either the documentation or the annotation is wrong and needs adaptation.

rlazo:
Hi @ubuntudroid, thanks for your report! The documentation is wrong, since the method implementing that interface does in fact expect the value to be non-null

https://github.com/firebase/firebase-android-sdk/blob/6ccc9de98aef8a559726c04a86ab3d47b35d6eed/firebase-dynamic-links/src/main/java/com/google/firebase/dynamiclinks/internal/FirebaseDynamicLinksImpl.java#L95

I'll follow up here when we have a fix ready.

rohandandavati:
b/178000308

davidmigloz:
What is the expected usage when you only care about the captured dynamic link after install? (e.g. because you have your own deep linking mechanism to handle app links that doesn't rely on Firebase Dynamic Links)

It'd good to have a `getDynamicLink()` without any parameters for that case (or allow to pass `null` as the documentation was suggesting before).

eldhosembabu:
Added a fix here: https://github.com/firebase/firebase-android-sdk/pull/2629

mikehardy:
This just bit us on react-native-firebase. We were coded up to wrap the API per docs, but a user took a null pointer exception because the implementation expects non-null

https://github.com/invertase/react-native-firebase/pull/5662

eldhosembabu:
#2629 is merged. Expect the fix to be rolled out in upcoming releases.  So closing this issue now.

PR #2629, opened 2021-04-28
FDL getDynamicLink(Intent) method to accept null intent.

eldhosembabu:
Changing the @NonNull to @Nullable for `Intent` argument in getDynamicLink(intent) method.

The intent parameter should be the intent that launched the application, or can be null if the intent does not include the dynamic link. A non-null intent is necessary only when the app is launched directly using the dynamic link, such as when using App Links.

Changes:
- Changed the annotation from  @NonNull to @Nullable.
- Modified getDynamicLink(Intent) method to handle null intents.
- Added test with null intent.

Related issue: https://github.com/firebase/firebase-android-sdk/issues/2336


vkryachko:
Is this PR still required or is it superseded by the recent FDL update?

warting:
> Is this PR still required or is it superseded by the recent FDL update?

@vkryachko It is still an issue in 20.1.0

vkryachko (approved):

PR #2632, opened 2021-04-29
Add console URL logging for unique trace metrics.

jeremyjiang-dev:


firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -191,6 +193,14 @@ public static FirebasePerformance getInstance() {
     gaugeManager.setApplicationContext(appContext);
 
     mPerformanceCollectionForceEnabledState = configResolver.getIsPerformanceCollectionEnabled();
+    if (isPerformanceCollectionEnabled()) {
+      Log.i(
+          LogWrapper.LOG_TAG,
+          String.format(
+              "Welcome to Firebase Performance Monitoring! Please visit %s in a minute to see details.",
```
visumickey, line 200: This definitely needs to be changed. As a default, below is my recommendation.

"Firebase performance successfully initialized. Visit %s in a minute to know the performance metrics of this application run".

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.perf.logging;
+
+public final class ConsoleUrlGenerator {
+  private static final String URL_BASE_PATH = "https://console.firebase.google.com";
+  private static final String UTM_MEDIUM = "perf-android-studio";
```
visumickey, line 19: Let not hardcoded it to android-studio. Developer do developer Android applications on other IDEs.

How about calling it "perf-android-ide"?

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Generate the console URL for Firebase Performance dashboard page.
+   *
+   * @param projectId the Firebase project ID
+   * @param packageName the name of this application's package
+   */
+  public static String generateDashboardUrl(String projectId, String packageName) {
+    return String.format(
+        "%s/project/%s/performance/app/android:%s/trends?utm_medium=%s",
```
visumickey, line 29: Just a thought, can we fetch this from remote config? And have this value as default value on client if RC does not have any value?

That way we are not worried about hardcoding the URL path on the client?

firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java
```diff
@@ -190,6 +194,7 @@ public void initialize(
       @NonNull Provider<TransportFactory> flgTransportFactoryProvider) {
 
     this.firebaseApp = firebaseApp;
+    projectId = firebaseApp.getOptions().getProjectId();
```
visumickey, line 197: this.projectId?

firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java
```diff
@@ -448,7 +454,15 @@ private boolean isAllowedToDispatch(PerfMetric perfMetric) {
 
   @WorkerThread
   private void dispatchLog(PerfMetric perfMetric) {
-    logger.info("Logging %s", getLogcatMsg(perfMetric));
+
+    // Logs the metrics to logcat plus console URL for every trace metric.
+    if (perfMetric.hasTraceMetric()) {
```
visumickey, line 459: Can you abstract a separate utility method isTraceMetric(perfMetric) and use that method here? That way the logic of determining if the provided metric is a trace metric is abstracted.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.perf.logging;
+
+public final class ConsoleUrlGenerator {
```
visumickey, line 17: Any unit tests for this? Lets make sure to create some to test if this is working as intended.

jeremyjiang-dev: Any suggestions on what tests to cover? I am not sure how to test logging functions.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: Sounds good! I have asked Rachel to take a look at the sentence.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
jeremyjiang-dev: My two cents:
1. I think it would be good if RC can have this value but RC might be an overkill. Values in RC should be expected to change over time so that we can remotely control the value. However, I do not expect the URL path to be changed very often. 
2. If the URL path gets changed, the FE should handle the redirection from the old URL to the new URL at least for a while.

firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java
jeremyjiang-dev: The purpose of `this` is to differentiate the input variable and instance variable if they have the same name. In my case, since there is no input param called `projectId`, it would be fine to use `projectId` without `this`. There is no universal standard to decide if `this` should be used in this case. I followed the convention in [other places in the codebase](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java#L203-L206) where `this` is not used for instance variables.

firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java
jeremyjiang-dev: Good suggestion! However, I would like to do this in a separate PR. `hasTraceMetric()`, `hasNetworkRequestMetric()`, `hasGaugeMetric()` have been the ways to determine the type of the `PerfMetric` and are used [here](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java#L582-L596) and many other places in the codebase. Fixing all of these could be out of the scope of this PR.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
jeremyjiang-dev: Good catch! Done. I have also added `utm_source` because I have tested that GA needs both `source` and `medium` for logging visits. PTAL.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
visumickey: Sounds good. Lets keep this aside for now and update SDKs when we change the endpoint in future.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
visumickey: We could just have static checks for the URLs that are generated from the URLGenerator. Just some straightforward checks to see the URL is as expected for the provided projectID and AppID would be sufficient.

firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java
visumickey: Sounds good. Thanks!

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
jeremyjiang-dev: Added some straightforward tests.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Generate the console URL for Firebase Performance dashboard page.
+   *
+   * @param projectId the Firebase project ID
+   * @param packageName the name of this application's package
+   */
+  public static String generateDashboardUrl(String projectId, String packageName) {
+    return String.format(
+        "%s/project/%s/performance/app/android:%s/trends?utm_source=%s&utm_medium=%s",
```
jposuna, line 30: `%s/project/%s/performance/app/android:` seems common to all the URLs; consider having it as a private function to avoid duplication.

firebase-perf/src/main/java/com/google/firebase/perf/transport/TransportManager.java
jposuna: My recommendation would be to be locally consistent. I.e. if a specific file does it, I'd recommend the whole file to do it. If we want to stop the practice, then we should update the whole file.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -191,6 +193,14 @@ public static FirebasePerformance getInstance() {
     gaugeManager.setApplicationContext(appContext);
 
     mPerformanceCollectionForceEnabledState = configResolver.getIsPerformanceCollectionEnabled();
+    if (isPerformanceCollectionEnabled()) {
+      Log.i(
```
jposuna, line 197: Why don't you use the AndroidLogger instead of exposing the LogWrapper? It has the added advantage that it already does the check as well.

firebase-perf/src/test/java/com/google/firebase/perf/logging/ConsoleUrlGeneratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+  private static final String TRACE_NAME = "test-trace";
+
+  @Test
+  public void testDashboardUrl() {
+    String url = ConsoleUrlGenerator.generateDashboardUrl(PROJECT_ID, PACKAGE_NAME);
+    assertThat(url)
+        .isEqualTo(
+            String.format(
+                "https://console.firebase.google.com/project/%s/performance/app/android:%s/trends?utm_source=perf-android-sdk&utm_medium=android-ide",
+                PROJECT_ID, PACKAGE_NAME));
```
jposuna, line 34: Avoid having logic in tests; I would just inline the constants (both in the function call, as well as the result).

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: Updated the sentence.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: The purpose is to log regardless of `isLogcatEnabled` defined in `AndroidLogger`. We want to log the console URL if FirebasePerformance is initialized per run.

firebase-perf/src/test/java/com/google/firebase/perf/logging/ConsoleUrlGeneratorTest.java
jeremyjiang-dev: Done.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jposuna: Thanks for the explanation. If `isLogcatEnabled = false`, wouldn't logging go against the user's expectation / desires?

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
jposuna: Did you upload the changes? I'm not seeing them here :)

jeremyjiang-dev: Whoops! Updated.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: You are right. This decision is made based on the fact that not many developers will turn on the `isLogcatEnabled` and the flag is off by default. As a result, we would like to log only the dashboard URL only once per run. Happy to hear your thoughts.

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: @jeremyjiang-dev How are we logging in the rest of the Firebase codebase? I would align with that. I would not recommend to supersede the choice of the developer to show/hide console logs.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
```diff
@@ -52,8 +52,19 @@ public static String generateCustomTraceUrl(String projectId, String packageName
    * @param name the trace name
    */
   public static String generateScreenTraceUrl(String projectId, String packageName, String name) {
+    String rootUrl = getRootUrl(projectId, packageName);
+    return String.format(
+        "%s/metrics/trace/SCREEN_TRACE/%s?utm_source=%s&utm_medium=%s",
+        rootUrl, name, UTM_SOURCE, UTM_MEDIUM);
+  }
+  /**
```
jposuna, line 60: Nit: Leave an empty line between the end of a function and the javadoc of the next function :)

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jposuna: > I would not recommend to supersede the choice of the developer to show/hide console logs

+1

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: Sg. I should have highlighted this in the design doc to catch attention.

firebase-perf/src/main/java/com/google/firebase/perf/logging/ConsoleUrlGenerator.java
jeremyjiang-dev: Done. Thanks for the speedy feedback!

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
jeremyjiang-dev: Done.

PR #2659, opened 2021-05-12
App Check Javadoc fixes

rosalyntan:


malcolmdeck:
Do we have a way to see what the generated Javadocs look like so we can see that it manifests correctly?

rachelsaunders:
Can we also fix the "short description" for getToken(): https://firebase.google.com/docs/reference/android/com/google/firebase/appcheck/AppCheckProvider

rosalyntan:
> Do we have a way to see what the generated Javadocs look like so we can see that it manifests correctly?
Not sure, but when hovering over a usage of `DebugAppCheckTestHelper` in Android Studio, the Javadoc that pops up looks correct.

> Can we also fix the "short description" for getToken(): https://firebase.google.com/docs/reference/android/com/google/firebase/appcheck/AppCheckProvider
Yup! That's included in this PR.

rachelsaunders (approved):

PR #2662, opened 2021-05-13
Fix fireperf RateLimiter replenishment behavior and unit alignment.

jfcong:
Fixed two bugs in the RateLimiter:
- Rate was represented by a long, which causes confusion around the units (per second vs per minute)
- the token replenishment behavior is wrong and keeps track of lastTimeTokenConsumed instead of lastTimeTokenReplenished, causing less tokens to be generated overtime

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
@@ -264,14 +264,14 @@ final class RateLimiter {
               0,
               (long)
                   (lastTimeTokenReplenished.getDurationMicros(now)
-                      * rate.getTokenPerSeconds()
+                      * rate.getTokensPerSeconds()
                       / MICROS_IN_A_SECOND));
       tokenCount = Math.min(tokenCount + newTokens, capacity);
       if (newTokens > 0) {
         lastTimeTokenReplenished =
             new Timer(
                 lastTimeTokenReplenished.getMicros()
-                    + (long) (newTokens * MICROS_IN_A_SECOND / rate.getTokenPerSeconds()));
+                    + (long) (newTokens * MICROS_IN_A_SECOND / rate.getTokensPerSeconds()));
       }
       if (tokenCount > 0) {
         tokenCount--;
--- a/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
@@ -16,6 +16,7 @@ package com.google.firebase.perf.transport;
 
 import static com.google.firebase.perf.metrics.resource.ResourceType.NETWORK;
 import static com.google.firebase.perf.metrics.resource.ResourceType.TRACE;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 import android.content.Context;
 import androidx.annotation.NonNull;
@@ -25,6 +26,7 @@ import com.google.firebase.perf.logging.AndroidLogger;
 import com.google.firebase.perf.metrics.resource.ResourceType;
 import com.google.firebase.perf.util.Clock;
 import com.google.firebase.perf.util.Constants;
+import com.google.firebase.perf.util.Rate;
 import com.google.firebase.perf.util.Timer;
 import com.google.firebase.perf.util.Utils;
 import com.google.firebase.perf.v1.NetworkRequestMetric;
@@ -34,7 +36,6 @@ import com.google.firebase.perf.v1.SessionVerbosity;
 import com.google.firebase.perf.v1.TraceMetric;
 import java.util.List;
 import java.util.Random;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Implement the Token Bucket rate limiting algorithm. The token bucket initially holds "capacity"
@@ -60,10 +61,10 @@ final class RateLimiter {
    * Construct a token bucket rate limiter.
    *
    * @param appContext the application's context.
-   * @param rate number of token generated per minute.
+   * @param rate the Rate object representing the number of tokens generated per specified time unit
    * @param capacity token bucket capacity
    */
-  public RateLimiter(@NonNull Context appContext, final double rate, final long capacity) {
+  public RateLimiter(@NonNull Context appContext, final Rate rate, final long capacity) {
     this(rate, capacity, new Clock(), getSamplingBucketId(), ConfigResolver.getInstance());
     this.isLogcatEnabled = Utils.isDebugLoggingEnabled(appContext);
   }
@@ -75,7 +76,7 @@ final class RateLimiter {
   }
 
   RateLimiter(
-      final double rate,
+      final Rate rate,
       final long capacity,
       final Clock clock,
       float samplingBucketId,
@@ -211,29 +212,29 @@ final class RateLimiter {
   static class RateLimiterImpl {
 
     private static final AndroidLogger logger = AndroidLogger.getInstance();
-    private static final long MICROS_IN_A_SECOND = TimeUnit.SECONDS.toMicros(1);
+    private static final long MICROS_IN_A_SECOND = SECONDS.toMicros(1);
 
     private final Clock clock;
     private final boolean isLogcatEnabled;
 
     // Last time a token is consumed.
-    private Timer lastTimeTokenConsumed;
+    private Timer lastTimeTokenReplenished;
 
-    // Number of new tokens generated per second.
-    private double rate;
+    // The Rate object representing the number of tokens generated per specified time unit
+    private Rate rate;
     // Token bucket capacity, also the initial number of tokens in the bucket.
     private long capacity;
     // Number of tokens in the bucket.
     private long tokenCount;
 
-    private double foregroundRate;
-    private double backgroundRate;
+    private Rate foregroundRate;
+    private Rate backgroundRate;
 
     private long foregroundCapacity;
     private long backgroundCapacity;
 
     RateLimiterImpl(
-        final double rate,
+        final Rate rate,
         final long capacity,
         final Clock clock,
         ConfigResolver configResolver,
@@ -243,7 +244,7 @@ final class RateLimiter {
       this.capacity = capacity;
       this.rate = rate;
       tokenCount = capacity;
-      lastTimeTokenConsumed = this.clock.getTime();
+      lastTimeTokenReplenished = this.clock.getTime();
       setRateByReadingRemoteConfigValues(configResolver, type, isLogcatEnabled);
       this.isLogcatEnabled = isLogcatEnabled;
     }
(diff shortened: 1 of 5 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
```diff
@@ -0,0 +1,64 @@
...
+// limitations under the License.
+
+package com.google.firebase.perf.util;
+
+import java.util.concurrent.TimeUnit;
+
+/** A Rate object representing the number of tokens per total specified time unit. */
+public class Rate {
+
+  private static final long NANO_IN_SECOND = 1000 * 1000 * 1000;
```
jposuna, line 22: Avoid doing maths on times. It's dangerous and standard libraries will do the conversions better than we can, so it's safer to use them.

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
```diff
@@ -0,0 +1,64 @@
...
+    this.numTimeUnits = numTimeUnits;
+    this.timeUnit = timeUnit;
+  }
+
+  /**
+   * Converts the rate to tokens per second.
+   *
+   * @return rate in tokens per second
+   */
+  public double getTokenPerSeconds() {
```
jposuna, line 52: Nit: s/getTokenPerSeconds/getTokensPerSecond

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
```diff
@@ -211,29 +212,29 @@ boolean getIsDeviceAllowedToSendNetworkEvents() {
...
-    private double backgroundRate;
+    private Rate foregroundRate;
+    private Rate backgroundRate;
 
     private long foregroundCapacity;
     private long backgroundCapacity;
 
     RateLimiterImpl(
-        final double rate,
+        final Rate rate,
```
jposuna, line 237: I'm noticing further down that there is a separate foreground and background rate, but they seem to be ignoring this initial one and just blindly overwriting it. Do you understand how it works?

firebase-perf/src/test/java/com/google/firebase/perf/util/RateTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+/** Unit tests for {@link Rate}. */
+@RunWith(RobolectricTestRunner.class)
+public class RateTest {
+
+  @Test
+  public void ratePerMinute() throws InterruptedException {
```
jposuna, line 29: Tests can throw Exception directly.

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
```diff
@@ -84,41 +91,215 @@ public void testRateLimitImpl() {
     // allow 2 logs every minute. token bucket capacity is 2.
     // clock is 0, token count is 2.
     RateLimiterImpl limiter =
-        new RateLimiterImpl(2.0d / 60, 2, mClock, mockConfigResolver, NETWORK, false);
+        new RateLimiterImpl(TWO_TOKENS_PER_MINUTE, 2, mClock, mockConfigResolver, NETWORK, false);
```
jposuna, line 94: What I'm seeing here is that with 2 tokens per minute, 4 requests in the first minute all succeed, which doesn't seem to make sense to me.

jfcong: The way I am seeing it is: if we have a full bucket of 2 tokens at time 0, and we replenish 2 tokens per minute, we will have four tokens available to be used in the first minute. Otherwise, are we saying we shouldn't replenish the bucket until it's empty?

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
jfcong: Fixed by using the standard Timeunit library.

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
jfcong: Fixed.

firebase-perf/src/test/java/com/google/firebase/perf/util/RateTest.java
jfcong: Fixed.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
jfcong: I think the foreground and background rate is set during initilization and read from remote config, but the passed in rate is still in use, until changeRate() is called. changeRate() is called by the TransportManager when an app updates its state (i.e., changes from foreground to background). At least that's my understanding.

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
```diff
@@ -14,14 +14,16 @@
...
+
 import java.util.concurrent.TimeUnit;
 
 /** A Rate object representing the number of tokens per total specified time unit. */
 public class Rate {
 
-  private static final long NANO_IN_SECOND = 1000 * 1000 * 1000;
-  private static final long MICRO_IN_SECOND = 1000 * 1000;
-  private static final long MILLI_IN_SECOND = 1000;
+  private static final long NANO_IN_SECOND = SECONDS.toNanos(1);
```
jposuna, line 24: If instead of keeping these as constants we just use them inline, we could have more straightforward conversions. E.g. NANOSECONDS.toSeconds(...), MILLISECONDS.toSeconds(...), etc.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
jposuna: But shouldn't the passed-in rate be either foreground or background? Why would there be a 3rd rate that is only used to initialize and then overwritten?

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
jposuna: I think this is slightly confusing since then we have double the configured capacity on the first minute. I think the idea to have an initial capacity of 2 is so that we don't start by throttling requests, not necessarily so we can have double the capacity during the first minute.

jfcong: Perhaps we can look at it differently. Even though conceptually to us, the rate is 2 tokens per minute, the implementation actually views the rate as token per second and calculates how many to refill per second. So in this example, the rate per second is 1/30. If viewed this way, we have multiplied the capacity during the first second by 60 times by allowing 2 events to go through within the first second but the rate per second is 1/30.

Reading more about the token bucket algorithm (wiki: https://en.wikipedia.org/wiki/Token_bucket), the capacity is there to allow a burst of events while the rate is there to regulate the average rate of events in the long run. I think that more events will be allowed to pass through when the capacity is more than the rate; capacity is the cap for the max number of events that can pass through, not the rate. For example, if the capacity is 500, and our rate is 2 per minute, we would allow all 500 events to go through in the first minute.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
jfcong: Perhaps because we don't know whether the app will be in foreground or background when it finishes initializing? I am not sure if there were other reasons that I am not aware of. We could remove the passed in rate and set it to either foreground or background on initialization if that makes more sense (but which one?)

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
jfcong: I made then inline, but I don't think I can use NANOSECONDS.toSeconds(...) etc. because that converts nanoseconds to second, but we want to convert 1 second to nanosecond, for example. I kept the inline variables to SECONDS.toNanos(1), etc.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
jposuna: @visumickey do you know why this is the way it is and whether it makes sense to keep it or change it?

jposuna: That sounds good - I'd recommend leaving it out of this PR though to not continue changing more things in a single PR.

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
```diff
@@ -0,0 +1,62 @@
...
+   * Constructs a Rate object.
+   *
+   * <p>For example, a rate of 3 tokens per minute can be represented by new Rate(3, 1, MINUTES) or
+   * new Rate(3, 60, SECONDS).
+   *
+   * @param numTokensPerTotalTimeUnit The number of tokens to be issued within the specified time
+   * @param numTimeUnits The number of specified time unit
+   * @param timeUnit The specified time unit
+   */
+  public Rate(long numTokensPerTotalTimeUnit, long numTimeUnits, TimeUnit timeUnit) {
```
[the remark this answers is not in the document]
jposuna: This is to prevent bugs like the existing one; if you only pass an int or long, it's very easy to get the units wrong (before we were mixing up minutes and seconds). By passing this, it's impossible to get it wrong.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
jfcong: Sounds good, I will leave that out of this PR. Although, is there a possibility that the app has not finished initializing yet so it's not in either foreground or background?

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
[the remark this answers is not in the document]
jfcong: I think the third parameter is there mainly for readability reasons. Before in this class, there are two conflicting comments about what the rate is: one says it's token per minute and the other one says it's token per second. By explicitly passing in the TimeUnit, we make it very clear that new Rate(3, 60, SECONDS) means 3 tokens per 60 seconds, and not 3 tokens per 60 minutes or vice versa.

Besides, by keeping the third parameter, we can also make testing more readable, by specifying rates of units other than SECONDS. For example, we can specify new Rate(2, 1, MINUTE) for 2 tokens per minute.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
visumickey: Yes, that is a possibility. But in those situations, it is fine to default it to the foreground rate.

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/util/Rate.java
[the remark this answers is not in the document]
visumickey: I definitely see the need for the readability part. My suggestion was to create the readability by calling the time unit to be always in seconds. Eg. (numberOfTokensForTimeLimit, timeLimitInSeconds)

The current is definitely super verbose and helpful. Was just feeling that verbosity can come by qualifying the parameter with better names.

Anyways, not too tied to my thoughts on this. Fine with this change as well.

visumickey (approved):

PR #2663, opened 2021-05-13
Change Uploader's `logAndUpdateState` to upload all pending events

andrewc94:
Instead of uploading one batch, the Uploader will now upload all pending events - this will get around a throughput issue that the uploader can experience.

### DIFF
--- a/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/UploaderIntegrationTest.java
+++ b/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/UploaderIntegrationTest.java
@@ -16,6 +16,7 @@ package com.google.android.datatransport.runtime;
 
 import static com.google.common.truth.Truth.assertThat;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
@@ -108,14 +109,12 @@ public class UploaderIntegrationTest {
     transport.send(stringEvent);
     verify(mockBackend, times(2))
         .send(eq(BackendRequest.create(Collections.singletonList(expectedEvent))));
-    verify(spyScheduler, times(1)).schedule(any(), eq(2));
+    verify(spyScheduler, times(1)).schedule(any(), eq(2), anyBoolean());
     Iterable<PersistedEvent> eventList = store.loadBatch(transportContext);
     assertThat(eventList).isNotEmpty();
     for (PersistedEvent persistedEvent : eventList) {
       assertThat(persistedEvent.getEvent()).isEqualTo(expectedEvent);
     }
-
-    assertThat(store.getNextCallTime(transportContext)).isEqualTo(0);
   }
 
   @Test
@@ -183,7 +182,6 @@ public class UploaderIntegrationTest {
         .send(eq(BackendRequest.create(Collections.singletonList(expectedEvent))));
     verify(spyScheduler, times(0)).schedule(any(), eq(2));
     assertThat(store.loadBatch(transportContext)).isEmpty();
-    assertThat(store.getNextCallTime(transportContext)).isEqualTo(0);
   }
 
   @Test
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
@@ -127,11 +127,12 @@ public class Uploader {
                     .build());
       }
       if (response.getStatus() == BackendResponse.Status.TRANSIENT_ERROR) {
+        long finalMaxNextRequestWaitMillis1 = maxNextRequestWaitMillis;
         guard.runCriticalSection(
             () -> {
               eventStore.recordFailure(persistedEvents);
               eventStore.recordNextCallTime(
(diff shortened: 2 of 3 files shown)


transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/UploaderIntegrationTest.java
```diff
@@ -108,14 +109,12 @@ public void uploader_transientError_shouldReschedule() {
     transport.send(stringEvent);
     verify(mockBackend, times(2))
         .send(eq(BackendRequest.create(Collections.singletonList(expectedEvent))));
-    verify(spyScheduler, times(1)).schedule(any(), eq(2));
+    verify(spyScheduler, times(1)).schedule(any(), eq(2), anyBoolean());
```
andrewc94, line 112: Here, we use a forced upload now. Even though this differs from the previous implementation, after reading https://github.com/firebase/firebase-android-sdk/pull/2393, I don't see why we wouldn't want to force an update - it's possible that if we hit a transient error that we still have pending events and would still want to schedule the update

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/UploaderIntegrationTest.java
```diff
@@ -108,14 +109,12 @@ public void uploader_transientError_shouldReschedule() {
...
     Iterable<PersistedEvent> eventList = store.loadBatch(transportContext);
     assertThat(eventList).isNotEmpty();
     for (PersistedEvent persistedEvent : eventList) {
       assertThat(persistedEvent.getEvent()).isEqualTo(expectedEvent);
     }
-
-    assertThat(store.getNextCallTime(transportContext)).isEqualTo(0);
```
andrewc94, line 118: This was removed because we now will store the next call time even during errors (to the current time + the max nextRequestCallTime seen by any previous successful uploads.) In the case where there are no previous uploads, although the value is different, I think the functionality should be the same since the difference is just 0 vs the current time.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -97,50 +97,56 @@ public void upload(TransportContext transportContext, int attemptNumber, Runnabl
...
-            workScheduler.schedule(transportContext, attemptNumber + 1);
-          } else {
-            eventStore.recordSuccess(persistedEvents);
-            if (response.getStatus() == BackendResponse.Status.OK) {
-              eventStore.recordNextCallTime(
-                  transportContext, clock.getTime() + response.getNextRequestWaitMillis());
-            }
-            if (eventStore.hasPendingEventsFor(transportContext)) {
-              workScheduler.schedule(transportContext, 1, true);
-            }
```
andrewc94, line 141: this bit wasn't carried over since we should be uploading all pending events

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -97,50 +97,56 @@ public void upload(TransportContext transportContext, int attemptNumber, Runnabl
...
+                    .setExtras(transportContext.getExtras())
+                    .build());
+      }
+      if (response.getStatus() == BackendResponse.Status.TRANSIENT_ERROR) {
+        guard.runCriticalSection(
+            () -> {
+              eventStore.recordFailure(persistedEvents);
+              return null;
+            });
+        eventStore.recordNextCallTime(transportContext, clock.getTime() + maxNextRequestWaitMillis);
```
vkryachko, line 135: Any reason these `eventStore` interactions are not part of the same critical section?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -97,50 +97,56 @@ public void upload(TransportContext transportContext, int attemptNumber, Runnabl
...
-              eventStore.recordNextCallTime(
-                  transportContext, clock.getTime() + response.getNextRequestWaitMillis());
-            }
-            if (eventStore.hasPendingEventsFor(transportContext)) {
-              workScheduler.schedule(transportContext, 1, true);
-            }
-          }
-          return null;
-        });
+    eventStore.recordNextCallTime(transportContext, clock.getTime() + maxNextRequestWaitMillis);
```
vkryachko, line 150: same here

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
andrewc94: Nope, that's a mistake

andrewc94: Ah sorry, I force pushed and wiped this out but the only change was wrapping this in the critical section

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
andrewc94: same

andrewc94:
Odd -  the stated failure passes locally for me and doesn't seem related to my changes? I'll try retesting again

vkryachko (approved):

PR #2676, opened 2021-05-19
Add firebase-common macrobenchmark.

vkryachko:


yifanyang (approved):
LGTM. I'll get the `application-id` part removed in a follow-up PR.

PR #2679, opened 2021-05-20
Add a comment about changing the CrashlyticsCore property

samedson:
Just a reminder


mrichards (changes requested):
In the past when we wanted to designated a stable internal API, we made changes such that the tests would fail if the API was changed, which would draw attention to the breaking change. Implementing something like [this](https://github.com/firebase/firebase-android-sdk/blob/1275cebd0eb732dd917ebc22279f31b72dc4da88/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiterTest.java#L64) is a better safeguard than the comment, IMO. WDYT?

samedson:
Doing https://github.com/firebase/firebase-android-sdk/pull/2884 instead

PR #2680, opened 2021-05-20
Clean up old sessions in the NDK

mrober:
Clean up old sessions in the NDK before initializing and after finalizing a session.

Tested manually with a test app.

### DIFF
--- a/firebase-crashlytics-ndk/src/main/java/com/google/firebase/crashlytics/ndk/NdkCrashFilesManager.java
+++ b/firebase-crashlytics-ndk/src/main/java/com/google/firebase/crashlytics/ndk/NdkCrashFilesManager.java
@@ -17,8 +17,13 @@ package com.google.firebase.crashlytics.ndk;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import java.io.File;
+import java.util.Arrays;
+import java.util.Comparator;
 
 class NdkCrashFilesManager implements CrashFilesManager {
+  private static final Comparator<? super File> LATEST_SESSION_FIRST =
+      (f1, f2) -> f2.getName().compareTo(f1.getName());
+  private static final int MAX_SESSIONS = 8;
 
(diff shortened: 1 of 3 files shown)


firebase-crashlytics-ndk/src/main/java/com/google/firebase/crashlytics/ndk/NdkCrashFilesManager.java
```diff
@@ -38,10 +43,21 @@ public File getSessionFileDirectory(String sessionId) {
...
 
   @Override
-  public void deleteSessionFilesDirectory(String sessionId) {
+  public void deleteSessionFileDirectory(String sessionId) {
     recursiveDelete(new File(rootPath, sessionId));
   }
 
+  @Override
+  public void cleanOldSessionFileDirectories() {
+    File[] sessionFileDirectories = rootPath.listFiles(File::isDirectory);
```
kmandrika, line 52: I'd double check with Marc about JDK version requirements - this syntax may not be supported.

mrichards: It should work now that we're requiring everyone to update to Java 8. I believe this syntax is 100% resolved in the compiler but it is probably worth testing on an API 19 emulator, which I believe is the oldest we support.

mrichards (approved):
LGTM, pending confirmation that JDK version isn't an issue.

firebase-crashlytics-ndk/src/main/java/com/google/firebase/crashlytics/ndk/NdkCrashFilesManager.java
mrober: I tested it on an emulator with API 16 and it works fine. We use this syntax in a few other places already.

mrober:
\test macrobenchmark

PR #2682, opened 2021-05-21
Upload macrobenchmark results to the metric service.

yifanyang:
- change to measure percentiles instead of mean/median/stdev

### DIFF
--- a/ci/fireci/fireciplugins/macrobenchmark.py
+++ b/ci/fireci/fireciplugins/macrobenchmark.py
@@ -106,7 +106,11 @@ async def _copy_google_services():
 
 async def _post_processing(results):
   # Upload successful measurements to the metric service
-  measurements = [i for x in results if not isinstance(x, Exception) for i in x]
+  measurements = []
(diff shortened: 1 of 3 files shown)


ci/fireci/fireciplugins/macrobenchmark.py
```diff
@@ -104,6 +104,19 @@ async def _copy_google_services():
     shutil.copyfile(src, dst)
 
 
+async def _post_processing(results):
+  # Upload successful measurements to the metric service
+  measurements = [i for x in results if not isinstance(x, Exception) for i in x]
```
rlazo, line 109: This line is getting a bit complex for a list comprehension, but feel free to leave it if you think is OK.

rlazo (approved):

ci/fireci/fireciplugins/macrobenchmark.py
yifanyang: Done. Use explicit for loop to assemble the list now.

PR #2683, opened 2021-05-22
Remove application-id from macrobenchmark config file.

yifanyang:
Also update version for some dependencies.

### DIFF
--- a/macrobenchmark/config.yaml
+++ b/macrobenchmark/config.yaml
@@ -1,22 +1,18 @@
 baseline:
   name: baseline
-  application-id: com.google.firebase.benchmark.baseline
 
 firebase-config:
   name: config
-  application-id: com.google.firebase.benchmark.config
   dependencies:
     - com.google.firebase:firebase-config-ktx
 
 firebase-common:
   name: common
-  application-id: com.google.firebase.benchmark.baseline
   dependencies:
     - com.google.firebase:firebase-common
 
 firebase-crashlytics:
   name: crash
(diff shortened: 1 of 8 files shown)


rlazo (approved):

macrobenchmark/config.yaml
```diff
@@ -1,76 +1,64 @@
...
     - com.google.firebase.crashlytics
 
 firebase-database:
   name: database
-  application-id: com.google.firebase.benchmark.database
   dependencies:
     - com.google.firebase:firebase-database-ktx
 
 firebase-dynamic-links:
   name: dynamiclinks
```
vkryachko, line 27: No immediate action required, but what is the purpose of the name given that we already have the section already named(i.e. firebase-dynamic-links)?

vkryachko (approved):

macrobenchmark/config.yaml
yifanyang: In the beginning, I thought there can be multiple apps built for one SDK, and this `name` can be used to distinguish those apps, but I suppose that is no longer the case. Right now it's used as part of the dir path for the generated test apps, but it is not necessary to have it in the path. Removing it requires some change, which I'll defer to a followup PR.

Regarding to application id, now that we only need one application id, I'm planning to add it to `fireescape-integ-tests` so that I can just use `preset-google-services` in prow jobs. What do you think?

vkryachko: sgtm

PR #2690, opened 2021-05-25
Functions to perform GET request when body data is empty

### LINKED ISSUE #2685: Functions always use POST request, even if there's no body data
### Describe your environment

  * Android Studio version: 4.2
  * Firebase Component: Functions (Database, Firestore, Storage, Functions, etc)
  * Component version: firebase-bom:26.1.1

#### Steps to reproduce:

Call a Functions GET endpoint without body data. The library always uses
```
Request.Builder().url(url).post(requestBody)
```
even if there's no body.

It should rather use
```
Request.Builder().url(url).get()
```
when there's nothing to send.

This is interfering with my API.

#### Relevant Code:

```
Firebase.functions
    getHttpsCallable(ENDPOINT)
    .call()
```

touficbatache:
Instead of always using a POST request, which is not always what devs want since express.js lets us build robust apis, functions will now perform a GET request when there is no body data.

Fixes and closes #2685

### DIFF
--- a/firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
+++ b/firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java
@@ -299,7 +299,12 @@ public class FirebaseFunctions {
     MediaType contentType = MediaType.parse("application/json");
     RequestBody requestBody = RequestBody.create(contentType, bodyJSON.toString());
 
-    Request.Builder request = new Request.Builder().url(url).post(requestBody);
+    Request.Builder request = new Request.Builder().url(url);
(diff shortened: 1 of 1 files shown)


touficbatache:
Hey @rlazo, would you mind checking out this PR? Thanks!

vkryachko:
@touficbatache thanks for taking interest in improving functions.

It does however look like this change would violate the protocol spec for callable functions: https://firebase.google.com/docs/functions/callable-reference#request_format_headers

Namely,

```
The HTTP request to a callable trigger endpoint must be a POST with the following headers:
```

So I don't think we can make this change without incurring a breaking change.

Issue #2637, opened 2021-05-03
Firestore takes up to 90 seconds to get new data after app returns to the foreground


jorgegil96:
* Android Studio version: 4.1.1
  * Firebase Component: Firestore
  * Component version: bom 27.1.0

If the app is backgrounded and the phone is idle for a while (say 1-2 hours), then, upon returning to the app, Firestore won't update with fresh data for up to 90 seconds. Which is a painful user experience because the data is stale and we can't do anything to force a refresh. My use case is a sports app where the user has to wait with the app open for over a minute to get fresh scores. Another person over in the flutter wrapper sdk was talking about a chat app.

I found a way to reliably reproduce this in a minimal sample project. You'll find the project and the logcat + steps in this drive folder:
https://drive.google.com/drive/folders/1bWPX4IKmQqs86-Khf0CqwwzKOLlqfOEe?usp=sharing

Copying repro steps here:
```
I wrote a sample app that shows live data from firestore. The data itself is just a timestamp
with the current time (US Central) that's written to firestore by a python script every minute. 
So if you look at the app and the time displayed doesn't match the current time, then we know 
there's an unreasonable delay.

Repro steps:
1. Disable LTE data, just keep Wifi connected
2. Kill app, open app (around 16:23:35)
3. Wait for app to receive update from network (~16:24:03)
4. Navigate home, open google chrome to put the app in background
5. Force idle (adb shell dumpsys deviceidle force-idle)
6. Keep looking at logcat to see updates every minute (~16:25:02, 16:26:02, 16:27:02, 16:28:02)
7. Wait until there's a connection error that kicks off exponential backoff (~16:28:15 we see a connection error)
8. Open app after 8 minutes (~16:36:10) 
- Note: I was just looking at the exp backoff delay retries (it was as high as 84000s, and as soon as I saw 59947 ms I opened the app)
9. Wait with the app open until we finally get a network update at (~16:37:48)
- Note: this is about 1 minute after opening up the app, which matches the last backoff timer :O
10. Revert force-idle (adb shell dumpsys deviceidle unforce)

Expected behavior:
As soon as you re-open app (step 8) it should check the network for new data.

Current behavior:
It is not until after 60ish seconds after re-opening app that we get network data.

Repro attemps: 5/5

See the txt file in drive for more info and some useful logs from Firestore and the sample app
```

[...]

PR #2693, opened 2021-05-25
Skip backoff when App enters Foreground

schmidt-sebastian:
We might lose network connectivity when a Firestore app is backgrounded. This PR changes our network monitor to skip the backoff when an app goes back in the foreground, which causes the SDK to immediately try to reconnect. 

Tested manually.

Fixes https://github.com/firebase/firebase-android-sdk/issues/2637


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
```diff
@@ -96,53 +99,60 @@ public void run() {
...
   }
 
+  private void configureBackgroundStateListener() {
+    BackgroundDetector.getInstance().addListener(this);
+  }
+
+  @Override
+  public void onBackgroundStateChanged(boolean background) {
+    if (!background && isConnected()) {
+      raiseCallbacks(/* connected= */ true);
```
var-const, line 109: Is invoking the callbacks enough to reset the backoff? (I presume it is, just checking)

var-const (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
schmidt-sebastian: Yeah, they already call `inhibitBackoff`.

PR #2671, opened 2021-05-17
Crashlytics Add Internal Keys feature for Unity Metadata

samedson:
- Allow us to pull internal Unity Metadata without impacting customers' Custom Keys limits.

PR #2697, opened 2021-05-26
Revert "Crashlytics Add Internal Keys feature for Unity Metadata"

samedson:
Reverts firebase/firebase-android-sdk#2671

We found this doesn't play nicely with the backend rollout.


mrichards (approved):
:(
We'll get it in for the next one.

PR #2698, opened 2021-05-27
update bom task for firebase-ads deprecation

davidmotson:

PR #2684, opened 2021-05-23
Bump up fireperf version and update Changelog.

jeremyjiang-dev:
1. Bump up Fireperf version from 20.0.0 to 20.0.1 for M97 release.
2. Remove `currentReleasedVersion` to avoid confusion.
3. Fix flaky tests in AppStateMonitorTest.java.

Note: this is not a blocker for M97 release.

PR #2700, opened 2021-05-28
Fix common tests to support patterns at end of string in useragent.

rlazo:
Old tests expected a whitespace between elements, but didn't account for the case when it's at the end of the string.

### DIFF
--- a/firebase-common/src/test/java/com/google/firebase/platforminfo/FirebasePlatformLoggingTest.java
+++ b/firebase-common/src/test/java/com/google/firebase/platforminfo/FirebasePlatformLoggingTest.java
@@ -23,6 +23,7 @@ import android.os.Build;
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.firebase.FirebaseOptions;
+import java.util.regex.Pattern;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.robolectric.annotation.Config;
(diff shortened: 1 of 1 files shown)


firebase-common/src/test/java/com/google/firebase/platforminfo/FirebasePlatformLoggingTest.java
```diff
@@ -79,7 +80,7 @@ public void test_watch_atNotHighEnoughApiLevel() {
         app -> {
           UserAgentPublisher ua = app.get(UserAgentPublisher.class);
 
-          assertThat(ua.getUserAgent()).contains("android-platform/ ");
+          assertThat(ua.getUserAgent()).containsMatch(Pattern.compile("android-platform/($|\\s)"));
```
vkryachko, line 83: Having played with this regex a bit, it does not look like what we want here. afaict it will *only* accept: `foo/` and `foo/ ` but not `foo/ bar/`.
What the test should ensure, is that if the value is not last, then there should be at least one space after it, so I suggest to use a pattern like `"android-platform/(\\s+.*)*"`. wdyt?

vkryachko: Also I am curious, is master broken because of this?

vkryachko (approved):

firebase-common/src/test/java/com/google/firebase/platforminfo/FirebasePlatformLoggingTest.java
rlazo: As discussed offline, the pattern does find the pattern :-) 

If not master, at least 2 other PRs: #2698 and #2684

PR #2702, opened 2021-05-31
Stop generating javadoc for DGC.

vkryachko:


buildSrc/src/main/java/com/google/firebase/gradle/MultiProjectReleasePlugin.java
```diff
@@ -64,13 +64,8 @@ public void apply(Project project) {
     boolean releaseJavadocs = toBoolean(System.getProperty("releaseJavadocs", "true"));
 
     File firebaseDevsiteJavadoc = new File(project.getBuildDir(), "firebase-javadocs/");
-    firebaseDevsiteJavadoc.mkdirs();
```
rlazo, line 67: why is this being removed?

rlazo (approved):

buildSrc/src/main/java/com/google/firebase/gradle/MultiProjectReleasePlugin.java
vkryachko: I removed it because it was redundant, from its javadoc:

```
The java.io.File.mkdirs() creates the directory named by this abstract pathname, including necessary and non-existent parent directories
```

So the call to mkdirs below, takes care of creating this "parent" directory

PR #2706, opened 2021-06-01
Update logging message with proper logging levels.

jeremyjiang-dev:
Summary
 - The main purpose for this PR is to make trace, network, and gauge logging messages more obvious in the logcat ([Example in Step 4.4](https://firebase.google.com/docs/perf-mon/get-started-android#view-log-messages)).
 - Update some logging messages with proper logging levels.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
@@ -293,7 +293,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
     if (isValidContentType(contentType)) {
       builder.setResponseContentType(contentType);
     } else {
-      logger.debug("The content type of the response is not a valid content-type:" + contentType);
+      logger.warn("The content type of the response is not a valid content-type:" + contentType);
     }
     return this;
   }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
@@ -70,7 +70,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
   public void updateSession(PerfSession session) {
     // Note(b/152218504): Being defensive to fix the NPE
     if (session == null) {
-      logger.debug("Unable to add new SessionId to the Network Trace. Continuing without it.");
+      logger.warn("Unable to add new SessionId to the Network Trace. Continuing without it.");
       return;
     }
 
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
@@ -70,7 +70,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
   public void updateSession(PerfSession session) {
     // Note(b/152218504): Being defensive to fix the NPE
     if (session == null) {
-      logger.info("Unable to add new SessionId to the Network Trace. Continuing without it.");
+      logger.debug("Unable to add new SessionId to the Network Trace. Continuing without it.");
       return;
     }
 
@@ -293,7 +293,7 @@ public final class NetworkRequestMetricBuilder extends AppStateUpdateHandler
     if (isValidContentType(contentType)) {
       builder.setResponseContentType(contentType);
     } else {
-      logger.info("The content type of the response is not a valid content-type:" + contentType);
+      logger.debug("The content type of the response is not a valid content-type:" + contentType);
     }
     return this;
   }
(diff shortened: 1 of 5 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
```diff
@@ -70,7 +70,7 @@
   public void updateSession(PerfSession session) {
     // Note(b/152218504): Being defensive to fix the NPE
     if (session == null) {
-      logger.info("Unable to add new SessionId to the Network Trace. Continuing without it.");
+      logger.debug("Unable to add new SessionId to the Network Trace. Continuing without it.");
```
jposuna, line 73: What criteria have you used to determine that this should be debug instead of warn?

jeremyjiang-dev: My criteria:
 - Debug: Useful message for debugging purposes. For example, the value of a remote config variable.
 - Warning: Unexpected behavior. The process might be continued, but take extra caution.

Therefore, I think this one should be `Warning`. Updated the PR :)

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
```diff
@@ -293,7 +293,7 @@ public NetworkRequestMetricBuilder setResponseContentType(@Nullable String conte
     if (isValidContentType(contentType)) {
       builder.setResponseContentType(contentType);
     } else {
-      logger.info("The content type of the response is not a valid content-type:" + contentType);
+      logger.debug("The content type of the response is not a valid content-type:" + contentType);
```
jposuna, line 296: Should this be warn?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
```diff
@@ -332,7 +332,7 @@ public NetworkRequestMetric build() {
     }
 
     if (isManualNetworkRequestMetric) {
-      logger.info(
+      logger.debug(
```
jposuna, line 335: Should this be warn?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
```diff
@@ -75,7 +75,7 @@
   public void updateSession(PerfSession session) {
     // Note(b/152218504): Being defensive to fix the NPE
     if (session == null) {
-      logger.info("Unable to add new SessionId to the Trace. Continuing without it.");
+      logger.debug("Unable to add new SessionId to the Trace. Continuing without it.");
```
jposuna, line 78: Should this be warn?

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
jposuna: Thanks! Left a few comments based on that criteria :)

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
jeremyjiang-dev: I think this could be `debug`. My interpretation of tis logging message is that manually created network metric cannot be built with `NetworkRequestMetricBuilder`. This is a check statement to inform that a manually created network metric is passed in the function. Therefore, a `debug` message could be sufficient.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
jeremyjiang-dev: My original thought is that this one can be either because `contentType` is extracted from response. We [check `contentType` length and disallow control characters](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java#L383-L399) and may cause false alarm. 
I updated to `warn` to match that this `contentType` is not expected from Fireperf's standard.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
jeremyjiang-dev: Yes. Updated.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
```diff
@@ -293,7 +293,7 @@ public NetworkRequestMetricBuilder setResponseContentType(@Nullable String conte
     if (isValidContentType(contentType)) {
       builder.setResponseContentType(contentType);
     } else {
-      logger.info("The content type of the response is not a valid content-type:" + contentType);
+      logger.warn("The content type of the response is not a valid content-type:" + contentType);
```
visumickey, line 296: Should this be a warning?

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/NetworkRequestMetricBuilder.java
jeremyjiang-dev: Refer to my previous comment:
"My original thought is that this one can be either Debug or Warn because contentType is extracted from response. We check contentType length and disallow control characters but this could be a false alarm.
I updated to warn to match that this contentType is not valid or expected from Fireperf's standard."

PR #2718, opened 2021-06-04
Fix logs, breadcrumbs and custom keys for ANRs

tejasd:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
```diff
@@ -895,7 +895,15 @@ private void writeApplicationExitInfoEventIfRelevant(String sessionId) {
...
       // happened during the session.
       if (applicationExitInfoList.size() != 0) {
-        reportingCoordinator.persistAppExitInfoEvent(sessionId, applicationExitInfoList.get(0));
+        final LogFileManager relevantSessionLogManager =
+            new LogFileManager(context, logFileDirectoryProvider, sessionId);
+        final UserMetadata relevantUserMetadata = new UserMetadata();
+        relevantUserMetadata.setCustomKeys(new MetaDataStore(getFilesDir()).readKeyData(sessionId));
+        reportingCoordinator.persistAppExitInfoEvent(
+            sessionId,
+            applicationExitInfoList.get(0),
```
pydichandra, line 904: I am assuming single app_exit_info entry gets created for app exit (and not multiple app_exit_info entries eg., ANR and OOM after an app exit). Is this assumption correct?

pydichandra (approved):

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java
tejasd: That assumption is most likely correct., but the documentation doesn't explicitly mention it, and I haven't looked at the implementation yet - but it does imply (https://developer.android.com/reference/kotlin/android/app/ApplicationExitInfo#gettraceinputstream)

However, we're currently only collecting ANRs, and so it's ok to only collect the last app_exit_info.

PR #2725, opened 2021-06-10
Update AnnotationBuilder to not rely on toString()

rlazo:
The toString() implementation of enum constants in annotations changed in JDK-8164819, this makes the previous behaviour more explicit.

### DIFF
--- a/encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/annotations/AnnotationBuilder.java
+++ b/encoders/firebase-encoders-processor/src/main/java/com/google/firebase/encoders/processor/annotations/AnnotationBuilder.java
@@ -136,7 +136,7 @@ public final class AnnotationBuilder {
 
           @Override
           protected Void defaultAction(Object o, Void unused) {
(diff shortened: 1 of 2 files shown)


vkryachko:
Looks like this is introducing a regression:

```
com/example/AtMyAnnotation.java:25: error: illegal start of expression
  private String strVal = default;
```

So I think for correctness we need to at least add `visitString`, and add tests that verify that other [possible types](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/AnnotationValueVisitor.html) work as expected

rlazo:
> Looks like this is introducing a regression:
> 
> ```
> com/example/AtMyAnnotation.java:25: error: illegal start of expression
>   private String strVal = default;
> ```
> 
> So I think for correctness we need to at least add `visitString`, and add tests that verify that other [possible types](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/AnnotationValueVisitor.html) work as expected

FIxed. PTAL

vkryachko (approved):

Issue #2672, opened 2021-05-18
[Performance][UIBlocking] AppStateMonitor Collect Activity Frame Data Prob


ethanhua:
![截屏2021-05-18 下午3 02 31](https://user-images.githubusercontent.com/22186185/118606503-148dce00-b7ea-11eb-815b-9586445bf304.png)

FirebasePerformance Log show that our slowFrame&frozenFrame data for each page is added on top of the data for all previous pages，maybe need to call `FrameMetricsAggregator.reset` method at activity stop lifecycle to solve this problems

![截屏2021-05-18 下午3 06 39](https://user-images.githubusercontent.com/22186185/118606997-a7c70380-b7ea-11eb-875e-73e043b26610.png)

Every time the `FrameMetricsAggregator.remove method` is called, it returns the same `mMetrics` instance that was not cleared at the stop time, resulting in all pages using the same data

![截屏2021-05-18 下午3 09 33](https://user-images.githubusercontent.com/22186185/118607361-10ae7b80-b7eb-11eb-832b-430a8777eccc.png)

aguatno:
Hi @ethanhua thanks for reporting. May we know what version of the SDK version you are using? Also, would you be able to create a project in GitHub that I could clone to reproduce? I'd like to make sure I'm using the same setup as your environment.

ethanhua:
@aguatno hello, my project sdk version: 
![截屏2021-05-21 上午10 57 27](https://user-images.githubusercontent.com/22186185/119075446-5786b600-ba23-11eb-864f-499336e7041a.png)
![截屏2021-05-21 上午10 57 52](https://user-images.githubusercontent.com/22186185/119075476-666d6880-ba23-11eb-99a3-142a20473b93.png)

i check the logic in source code of firebase perf module, this probs may exist in lasted version

yalunqin:
Thank you ethanhua, just FYI we are actively investigating this issue now. We will keep you posted.

jeremyjiang-dev:
PR #2736 is submitted to fix this issue. 
@ethanhua Thank you for reporting the issue and suggesting the fixes!

PR #2736, opened 2021-06-15
Fix inaccurate activity screen metrics.

jeremyjiang-dev:
This PR is to fix:
1. https://github.com/firebase/firebase-android-sdk/issues/2672 
2. Screen trace data not tracked for Android API levels 23 and below (b/191954905).

Summary of changes:
1. Replaced `frameMetricsAggregator.remove(activity)` with `frameMetricsAggregator.reset()` because `remove` does not clean the frame metrics, causes frame metrics for each activity is added on top of the data for all previous pages.
2. Removed [checks](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java#L405-L407) for hardware accelerated. The reason for adding this check is due to the crash by calling `frameMetricsAggregator.remove(activity)` (see b/133827763). However, this check will return false even if hardware acceleration is turned on by default. Now, if an activity is not hardware accelerated, the metrics will just be 0.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -289,7 +289,8 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
   public void onActivityPaused(Activity activity) {}
 
   /**
-   * Send screen trace.
+   * Send screen trace. If hardware acceleration is not enabled, all frame metrics will be zero and
+   * the trace will not be sent.
    *
    * @param activity activity object.
    */
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -149,6 +149,10 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
   public synchronized void onActivityStarted(Activity activity) {
     if (isScreenTraceSupported(activity) && configResolver.isPerformanceMonitoringEnabled()) {
       // Starts recording frame metrics for this activity.
+      /**
+       * TODO: Only add activities that are hardware acceleration enabled so that calling {@link
+       * FrameMetricsAggregator#remove(Activity)} will not throw exceptions.
+       */
       frameMetricsAggregator.add(activity);
       // Start the Trace
       Trace screenTrace = new Trace(getScreenTraceName(activity), transportManager, clock, this);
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -391,21 +390,13 @@ private void sendSessionLog(String name, Timer startTime, Timer endTime) {
   }
 
   /**
-   * Only send screen trace if FrameMetricsAggregator exists and the activity is hardware
-   * accelerated.
+   * Only send screen trace if FrameMetricsAggregator exists.
```
rlazo, line 393: In the PR description you mention that for non hw accelerated views this will be 0. Should the javadoc mention that anywhere? Not here, since this is a private method.

jeremyjiang-dev: Thank! Added a comment to `sendScreenTrace(Activity activity)`. There is no public method about screen traces because screen traces are measured automatically by Fireperf and users cannot measure screen traces themselves.

rlazo (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -308,7 +308,7 @@ private void sendScreenTrace(Activity activity) {
     int slowFrames = 0;
     int frozenFrames = 0;
     // Stops recording metrics for this Activity and returns the currently-collected metrics
-    SparseIntArray[] arr = frameMetricsAggregator.remove(activity);
+    SparseIntArray[] arr = frameMetricsAggregator.reset();
```
visumickey, line 311: This is the critical change for this PR.

Earlier we were stopping the aggregator for a particular activity. With reset, we don't do this per activity. So, concerned if there is any sideways impact for the app when doing reset.

jeremyjiang-dev: The sideway impact is that `frameMetricsAggregator` will be monitoring all the activities that are added even after `reset()` is called. However, from the perspective of measuring the correct screen performance, this is fine. The reason is that there will only be one activity alive at a moment. We start measuring the performance when the activity starts and stop measuring when the activity stops. When the activity stops, we will call `reset()` to clear the data in  `frameMetricsAggregator` so that the next activity will have clean data at the beginning.

jeremyjiang-dev: Discussed offline. Will add TODOs and comments.

PR #2737, opened 2021-06-16
Crashlytics smoke test stubs

mrichards:


smoke-tests/src/main/java/com/google/firebase/testing/CrashlyticsTest.java
```diff
@@ -0,0 +1,87 @@
...
+    // that events sent to Firebase Analytics are received by the Crashlytics breadcrumb handler.
+    try {
+      // need to use reflection to get the breadcrumb handler.
+      FirebaseCrashlytics crashlytics = FirebaseCrashlytics.getInstance();
+      Field coreField = crashlytics.getClass().getDeclaredField("core");
+      coreField.setAccessible(true);
+      Object core = coreField.get(crashlytics);
+      Field breadcrumbSourceField = core.getClass().getDeclaredField("breadcrumbSource");
+      breadcrumbSourceField.setAccessible(true);
+      BreadcrumbSource breadcrumbSource = (BreadcrumbSource)breadcrumbSourceField.get(core);
```
yifanyang, line 47:
```suggestion
      BreadcrumbSource breadcrumbSource = (BreadcrumbSource) breadcrumbSourceField.get(core);
```

yifanyang (approved):
Thanks for adding this test! LGTM.

yifanyang:
@vkryachko Adding you as a reviewer about the reflections and additional proguard rules.

smoke-tests/src/main/java/com/google/firebase/testing/CrashlyticsTest.java
```diff
@@ -0,0 +1,87 @@
...
+public final class CrashlyticsTest {
+
+  @Test
+  public void analyticsIntegration() {
+    // Validates that Firebase Analytics and Crashlytics interoperability is working, by confirming
+    // that events sent to Firebase Analytics are received by the Crashlytics breadcrumb handler.
+    try {
+      // need to use reflection to get the breadcrumb handler.
+      FirebaseCrashlytics crashlytics = FirebaseCrashlytics.getInstance();
+      Field coreField = crashlytics.getClass().getDeclaredField("core");
```
vkryachko, line 42: I feel like this use of reflection could cause maintenance problems down the road.

There seems to be an easy workaround, given the following observation: `core` is not a private field and is visible within the `com.google.firebase.crashlytics` package. @mrichards  Can we move this test into `src/main/java/com/google/firebase/crashlytics` and access the field directly?

As for breadcrumbSource, can we make it public and annotation with `@VisibleForTesting`? (could actually be used with `core` above as well)

mrichards: @vkryachko Please take a look at the updated implementation. I moved `CrashlyticsTest` to the crashlytics package, because making `core` public in FirebaseCrashlytics makes it visible in the IDE's autocomplete for customers, even when `@VisibleForTesting` was used.

LMK if this is consistent with your suggestion.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -156,7 +157,8 @@ public Void call() throws Exception {
     return new FirebaseCrashlytics(core);
   }
 
-  private final CrashlyticsCore core;
+  // package-private for smoke tests
+  final CrashlyticsCore core;
```
vkryachko, line 161: Could you still add `@VisibleForTesting`?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsCore.java
```diff
@@ -76,10 +77,12 @@
   private boolean didCrashOnPreviousExecution;
 
   private CrashlyticsController controller;
-
   private final IdManager idManager;
-  private final BreadcrumbSource breadcrumbSource;
+
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
vkryachko, line 82: I think otherwise = Private is the default, so no need to explicitly specify

vkryachko (approved):
LGTM with minor nits

PR #2744, opened 2021-06-17
Simplify proguard handling logic.

vkryachko:
This change removes custom logic that configures proguarding and
consumerProguardFiles for SDKs and instead makes all configuration
explicit per SDK.


buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -25,6 +25,7 @@
 import com.google.firebase.gradle.plugins.apiinfo.GetMetalavaJarTask;
 import com.google.firebase.gradle.plugins.ci.Coverage;
 import com.google.firebase.gradle.plugins.ci.device.FirebaseTestServer;
+import com.google.firebase.gradle.plugins.license.LicenseResolverPlugin;
```
rlazo, line 28: Is this change related to the proguard logic?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/license/LicenseResolverPlugin.java
```diff
@@ -40,7 +40,7 @@
  *       GenerateLicensesTask} to create the artifacts necessary for our release.
  * </ul>
  */
-class LicenseResolverPlugin implements Plugin<Project> {
+public class LicenseResolverPlugin implements Plugin<Project> {
```
rlazo, line 43: same

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
vkryachko: yes and no :)

It used to live in the root project configuration alongside the proguard logic. So it makes sense to move it here

rlazo (approved):

PR #2754, opened 2021-06-22
Initial code structure for App Distribution Android SDK 

andrewbibiloni:
Initial commit for App Distribution new build alerts Android SDK


firebase-app-distribution/firebase-app-distribution.gradle
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+}
+
+
+dependencies {
+    implementation 'org.jetbrains:annotations:15.0'
+    implementation project(path: ':firebase-components')
+    implementation 'com.google.android.gms:play-services-tasks:17.0.0'
+}
```
rachaprince, line 47: nit: could we add new lines to the end of these files?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionApi.java
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Updates app to the latest release. If the latest release is an APK, downloads the binary and
+   * starts an installation If the latest release is an AAB, directs the tester to the Play app to
+   * complete the download and installation.
+   *
+   * @throws an UPDATE_NOT_AVAIALBLE exception if
+   * @param updateProgressListener a callback function invoked as the update progresses
+   */
+  @NonNull
+  public abstract Task<Void> updateApp(@Nullable UpdateProgressListener updateProgressListener);
```
vkryachko, line 52: it's a bit strange to me that lack of update is an exception, do you have an api review doc I can take a look at?

rachaprince: This method is different from updateToLatestRelease, because it should be called only after `checkForUpdate` returns a truthy value. If it is called when no release is available, we'd throw the exception.

We'll share the API proposal with you over chat

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionApi.java
```diff
@@ -0,0 +1,137 @@
...
+   */
+  @NonNull
+  public abstract Task<Optional<AppDistributionRelease>> checkForUpdate();
+
+  /**
+   * Updates app to the latest release. If the latest release is an APK, downloads the binary and
+   * starts an installation If the latest release is an AAB, directs the tester to the Play app to
+   * complete the download and installation.
+   *
+   * @throws an UPDATE_NOT_AVAIALBLE exception if
```
rachaprince, line 48: nit: this comment seems incomplete

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionApi.java
```diff
@@ -0,0 +1,137 @@
...
+// limitations under the License.
+
+package com.google.firebase.appdistribution;
+
+import androidx.annotation.NonNull;
+import com.google.android.gms.tasks.Task;
+import java.util.Optional;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class FirebaseAppDistributionApi {
```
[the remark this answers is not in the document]
rachaprince: Yes, the original idea was to change it to a concrete class in the future. But your comment prompted us to revisit our decision and we've changed this to a concrete class without the Api suffix

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionApi.java
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Updates the app to the latest release, if one is available. Returns the release information or
+   * an empty Optional if no update is found. Performs the following actions: 1. If tester is not
+   * signed in, presents the tester with a Google sign in UI 2. Checks if a newer release is
+   * available. If so, presents the tester with a confirmation dialog to begin the download. 3. For
+   * APKs, downloads the binary and starts an installation intent. 4. For AABs, directs the tester
+   * to the Play app to complete the download and installation.
+   */
+  @NonNull
+  public abstract Task<Optional<AppDistributionRelease>> updateToLatestRelease();
```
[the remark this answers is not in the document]
rachaprince: Done

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+
+  /** Returns true if the App Distribution tester is signed in */
+  @NonNull
+  public boolean isTesterSignedIn() {
+    return false;
+  }
+
+  /** Signs out the App Distribution tester */
+  public void signOutTester() {}
+
```
pranavrajgopal, line 139: move this to a new file

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+    // The release notes for this build
+    @NonNull
+    public String getReleaseNotes();
+
+    // The binary type for this build
+    @NonNull
+    public BinaryType getBinaryType();
+  }
+
```
pranavrajgopal, line 158: move this outside of the FirebaseAppDistribution class (not sure if this should be in its own file). @vkryachko thoughts?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+    return Tasks.forResult(null);
+  }
+
+  /**
+   * Interface to subscribe to status updates on the release update process. Called by updateApp.
+   */
+  public interface UpdateProgressListener {
+    public void onProgressUpdate(@NonNull UpdateProgress updateProgress);
+  }
+
```
pranavrajgopal, line 83: Move to its own file.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+   * @throws an UPDATE_NOT_AVAIALBLE exception if no new release is cached from checkForUpdate
+   * @param updateProgressListener a callback function invoked as the update progresses
+   */
+  @NonNull
+  public Task<Void> updateApp(@Nullable UpdateProgressListener updateProgressListener) {
+    return Tasks.forResult(null);
+  }
+
+  /**
+   * Interface to subscribe to status updates on the release update process. Called by updateApp.
```
pranavrajgopal, line 78: Move to its own file

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+
+    // The total number of bytes to download for the APK.
+    // Returns -1 if called on an AAB.
+    @NonNull
+    public long getApkTotalBytesToDownload();
+
+    @NonNull
+    public UpdateStatus getUpdateStatus();
+  }
+
```
pranavrajgopal, line 98: Move this to its own file

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+   */
+  @NonNull
+  public Task<Void> updateApp(@Nullable UpdateProgressListener updateProgressListener) {
+    return Tasks.forResult(null);
+  }
+
+  /**
+   * Interface to subscribe to status updates on the release update process. Called by updateApp.
+   */
+  public interface UpdateProgressListener {
```
pranavrajgopal, line 80: Add a comment for onProgressUpdate

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,163 @@
...
+    // The number of bytes downloaded so far for the APK.
+    //  Returns -1 if called on an AAB.
+    @NonNull
+    public long getApkBytesDownloaded();
+
+    // The total number of bytes to download for the APK.
+    // Returns -1 if called on an AAB.
+    @NonNull
+    public long getApkTotalBytesToDownload();
+
```
pranavrajgopal, line 94: Add a comment to getUpdateStatus

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionRegistrar.java
```diff
@@ -15,16 +15,24 @@
...
 import androidx.annotation.Keep;
+import com.google.firebase.FirebaseApp;
 import com.google.firebase.components.Component;
 import com.google.firebase.components.ComponentRegistrar;
+import com.google.firebase.components.Dependency;
+import com.google.firebase.platforminfo.LibraryVersionComponent;
 import java.util.Arrays;
 import java.util.List;
 
-/** @hide */
```
rachaprince: I'm curious why removed the hide annotation here? It seems like other Registrar classes have it, and @hide should prevent this class from showing up in the API docs, which I think is what we want. (https://stackoverflow.com/q/17035271)

rachaprince:
/test check-changed
/test device-check-changed
//test macrobenchmark

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionRegistrar.java
pranavrajgopal: haha this one is on me. I wasn't sure what the hide annotation does so I asked Andrew to remove it. Sorry about that!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AppDistributionRelease.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.google.firebase.appdistribution;
+
+import androidx.annotation.NonNull;
+
+/**
+ * Interface for AppDistributonRelease object returned by checkForUpdate() and
+ * updateToLatestRelease()
+ */
+public interface AppDistributionRelease {
```
vkryachko, line 23: This looks like a POJO to me, why do we need this to be an interface rather than a `public final class`? Do we anticipate different implementations with non-trivial logic as opposed to just treating this class as a "bag" of values?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.android.gms.tasks.Tasks;
+import com.google.firebase.FirebaseApp;
+import org.jetbrains.annotations.Nullable;
+
+public class FirebaseAppDistribution {
+
+  private final FirebaseApp firebaseApp;
+
+  /** Constructor for FirebaseAppDistribution */
+  public FirebaseAppDistribution(@NonNull FirebaseApp firebaseApp) {
```
vkryachko, line 28: Do we want to allow direct instantiation by developers? usually firebase sdk constructors are package private

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateProgress.java
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.appdistribution;
+
+import androidx.annotation.NonNull;
+
+/** Interface to get download progress for APKs and the status of the update. Used in updateApp. */
+public interface UpdateProgress {
```
vkryachko, line 20: same question here: why an interface rather than a POJO?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Updates app to the latest release. If the latest release is an APK, downloads the binary and
+   * starts an installation If the latest release is an AAB, directs the tester to the Play app to
+   * complete the download and installation.
+   *
+   * @throws an UPDATE_NOT_AVAIALBLE exception if no new release is cached from checkForUpdate
+   * @param updateProgressListener a callback function invoked as the update progresses
+   */
+  @NonNull
+  public Task<Void> updateApp(@Nullable UpdateProgressListener updateProgressListener) {
```
vkryachko, line 70: This API looks a bit suspicious, on the one hand it's returning a task indicating that it will succeed in the future, otoh it also takes a listener to monitor progress.
We have a pattern for such APIs, e.g. storage uses `CancellableTask` that allows to listen to progress updates. It would be beneficial to do something similar here. cc @schmidt-sebastian

pranavrajgopal: Could you speak more about why returning a task is at odds with providing a progress listener? The idea behind a progress listener was to provide customers the flexibility of listening to progress updates (in our case APK download progress). Something like the `CancellableTask` is reasonable but we don't want to introduce update cancellation functionality yet.

pranavrajgopal: To add more context to the API - `updateApp` is a Task<Void> because we didn't think there is any actionable data to be returned to the customer if that task succeeds. But customers might be interested in knowing the progress of the current update (maybe to present a custom UI)

rachaprince: We discussed this offline. We plan to update this method to return an UpdateTask<UpdateState>, and follow the same pattern as the CancellableTask (but without providing cancellation functionality)

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateProgress.java
rachaprince: Updated to a POJO

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AppDistributionRelease.java
rachaprince: Done, this should be a POJO

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AppDistributionRelease.java
```diff
@@ -17,23 +17,44 @@
 import androidx.annotation.NonNull;
 
 /**
- * Interface for AppDistributonRelease object returned by checkForUpdate() and
+ * Data class for AppDistributionRelease object returned by checkForUpdate() and
  * updateToLatestRelease()
  */
-public interface AppDistributionRelease {
+public final class AppDistributionRelease {
```
vkryachko, line 23: non-blocking nit: you may want to add `equals()` and `hashCode()`, or use an `@AutoValue` altogether.

vkryachko (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AppDistributionRelease.java
rachaprince: Good call. We will add these in a follow-up PR

PR #2757, opened 2021-06-23
Implement GetAppCheckToken API

rosalyntan:
Implement API for 3P developers to request a Firebase App Check token as well as observe changes to the token. This will enable developers to protect their own (non-Firebase) backends with the FAC token.

### DIFF
--- a/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
+++ b/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
@@ -46,6 +46,7 @@ public class DefaultFirebaseAppCheck extends FirebaseAppCheck {
   private final Provider<UserAgentPublisher> userAgentPublisherProvider;
   private final Provider<HeartBeatInfo> heartBeatInfoProvider;
   private final List<AppCheckTokenListener> appCheckTokenListenerList;
+  private final List<AppCheckListener> appCheckListenerList;
   private final StorageHelper storageHelper;
   private final TokenRefreshManager tokenRefreshManager;
   private final Clock clock;
@@ -65,6 +66,7 @@ public class DefaultFirebaseAppCheck extends FirebaseAppCheck {
     this.userAgentPublisherProvider = userAgentPublisherProvider;
     this.heartBeatInfoProvider = heartBeatInfoProvider;
     this.appCheckTokenListenerList = new ArrayList<>();
+    this.appCheckListenerList = new ArrayList<>();
     this.storageHelper =
         new StorageHelper(firebaseApp.getApplicationContext(), firebaseApp.getPersistenceKey());
     this.tokenRefreshManager =
@@ -110,7 +112,8 @@ public class DefaultFirebaseAppCheck extends FirebaseAppCheck {
   public void addAppCheckTokenListener(@NonNull AppCheckTokenListener listener) {
     checkNotNull(listener);
     appCheckTokenListenerList.add(listener);
-    tokenRefreshManager.onListenerCountChanged(appCheckTokenListenerList.size());
+    tokenRefreshManager.onListenerCountChanged(
+        appCheckTokenListenerList.size() + appCheckListenerList.size());
     // If there is a token available, trigger the listener with the current token.
     if (hasValidToken()) {
       listener.onAppCheckTokenChanged(
@@ -122,7 +125,28 @@ public class DefaultFirebaseAppCheck extends FirebaseAppCheck {
   public void removeAppCheckTokenListener(@NonNull AppCheckTokenListener listener) {
     checkNotNull(listener);
     appCheckTokenListenerList.remove(listener);
-    tokenRefreshManager.onListenerCountChanged(appCheckTokenListenerList.size());
+    tokenRefreshManager.onListenerCountChanged(
+        appCheckTokenListenerList.size() + appCheckListenerList.size());
+  }
+
+  @Override
+  public void addAppCheckListener(@NonNull AppCheckListener listener) {
+    checkNotNull(listener);
+    appCheckListenerList.add(listener);
+    tokenRefreshManager.onListenerCountChanged(
+        appCheckTokenListenerList.size() + appCheckListenerList.size());
+    // If there is a token available, trigger the listener with the current token.
+    if (hasValidToken()) {
+      listener.onAppCheckTokenChanged(cachedToken);
+    }
+  }
+
+  @Override
+  public void removeAppCheckListener(@NonNull AppCheckListener listener) {
+    checkNotNull(listener);
+    appCheckListenerList.remove(listener);
+    tokenRefreshManager.onListenerCountChanged(
+        appCheckTokenListenerList.size() + appCheckListenerList.size());
   }
 
   @NonNull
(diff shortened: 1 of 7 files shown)


appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
```diff
@@ -110,7 +112,8 @@ public void resetAppCheckState() {
   public void addAppCheckTokenListener(@NonNull AppCheckTokenListener listener) {
     checkNotNull(listener);
     appCheckTokenListenerList.add(listener);
-    tokenRefreshManager.onListenerCountChanged(appCheckTokenListenerList.size());
+    tokenRefreshManager.onListenerCountChanged(
```
malcolmdeck, line 115: I'm assuming that we're intentional about the fact that external listeners cause auto-refreshing, right?

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
```diff
@@ -46,6 +46,7 @@
   private final Provider<UserAgentPublisher> userAgentPublisherProvider;
   private final Provider<HeartBeatInfo> heartBeatInfoProvider;
   private final List<AppCheckTokenListener> appCheckTokenListenerList;
+  private final List<AppCheckListener> appCheckListenerList;
```
malcolmdeck, line 49: It occurs to me that we could, in theory, have both of these implement an interface and then we don't have all of this tricky logic to deal with both lists everywhere. Instead, we'd have one list, and then a helper Factory class to deal with constructing the right thing to shove into each instance. 

That being said, it's not clear to me if the drawbacks to that that approach outweigh the benefits, I just wanted to flag that there was an option to keep the messiness contained to a single place if we wanted.

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
rosalyntan: Yeah, I think that's what makes sense, but happy to discuss if you disagree!

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
```diff
@@ -137,33 +161,63 @@ public void removeAppCheckTokenListener(@NonNull AppCheckTokenListener listener)
...
+      return Tasks.forException(new FirebaseException("No AppCheckProvider installed."));
+    }
     return fetchTokenFromProvider();
   }
 
   /** Fetches an {@link AppCheckTokenResult} via the installed {@link AppCheckProvider}. */
-  Task<AppCheckTokenResult> fetchTokenFromProvider() {
-    return appCheckProvider
-        .getToken()
+  Task<AppCheckTokenResult> fetchTokenResultFromProvider() {
```
malcolmdeck, line 180: Splitting this method in two seems like a code smell - the inner one is for Tokens, and the outer one is for TokenResults, but then you're still constructing TokenResults in the inner one in order to trigger the listener. Is there a reason (testing maybe?) that we've done it this way? I agree that splitting the listeners into the places where their insides are constructed would also be error-prone, which is why I wonder why we've done the split at all.

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultTokenRefresher.java
```diff
@@ -94,20 +93,12 @@ private long getNextRefreshMillis() {
   }
 
   private void onRefresh() {
-    Task<AppCheckTokenResult> task = firebaseAppCheck.fetchTokenFromProvider();
-    task.addOnCompleteListener(
-        new OnCompleteListener<AppCheckTokenResult>() {
+    Task<AppCheckToken> task = firebaseAppCheck.fetchTokenFromProvider();
```
malcolmdeck, line 96: Nit: you probably don't need to declare the variable `task` - you could just call:
firebaseAppCheck.fetchTokenFromProvider().addOnFailureListener(...

malcolmdeck (commented):
I added a bunch of comments - this might be the best way to do things, but it has some vestiges of an implementation in Auth that I regret, so I've left some thoughts.

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultTokenRefresher.java
rosalyntan: Done.

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
rosalyntan: I'm not sure I understand this suggestion -- do you mean have both `AppCheckListener` and `AppCheckTokenListener` implement some other `Listener` interface, and then just have a `List<Listener>` here?

The `onAppCheckTokenChanged` methods of `AppCheckListener` and `AppCheckTokenListener` take in different parameters (`AppCheckToken` vs. `AppCheckTokenResult`), so I don't think that would work.

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
rosalyntan: The split is because the `getAppCheckToken` method needs to return an `AppCheckToken` while the `getToken` method needs to return an `AppCheckTokenResult`. The `fetchTokenFromProvider` method should do all the heavy-lifting for getting the token, updating the stored state, and notifying listeners. All `fetchTokenResultFromProvider` does is wrap the result into an `AppCheckTokenResult` for Firebase SDKs to consume.

Would it be more clear for me to get rid of the `fetchTokenResultFromProvider` method, and handle wrapping the token into an `AppCheckTokenResult` directly within `getToken`?

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
malcolmdeck: Right, that's what you would need a helper/factory for, do deal with parseling the token into the right format for triggering the listener. Then all of the complexity of having 2 types of listeners only lives in that one place. Whether or not that's preferable is up to you

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
malcolmdeck: It's just weird that you have:
- a place for marshalling things into TokenResults
- a place for internal heavy lifting, which also marshalls things into TokenResults
This would be resolved if you had a helper method that takes in a list of listeners of both types and a token and then marshals when necessary. Then you just have:
- a place that takes a token and marshalls it into a TokenResult
- a place that does everything internally, but only manipulates the Token itself. 
That being said, all we'd be doing is hiding the fact that we're duplicating work, so maybe that's not better either. I just know that the whole "have two Listener classes that are almost, but not quite, the same everywhere" is (IMO) just asking for a bug at some point where someone includes one but not both because we haven't isolated the pattern of dealing with both kinds of listeners to a single place. But that might be more trouble than it's worth, I leave it to you. I'm just trying to flag the concern - if I thought I had a clearly better way to do it, I'd say so, but I don't.

malcolmdeck (approved):

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
rosalyntan: Ack, for this PR I'll just inline the `AppCheckTokenResult` wrapping into `getToken` so we don't have two separate methods that do nearly the same thing.

In a follow-up PR, I'll do some refactoring to wrap the two lists of listeners into an internal class, so the logic to deal with the listeners can be encapsulated in one place.

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultFirebaseAppCheck.java
rosalyntan: See comment below.

PR #2763, opened 2021-06-25
Use manual foreground detection to trigger network reconnect

schmidt-sebastian:
This is a follow up to https://github.com/firebase/firebase-android-sdk/pull/2693, which used Android's built-in BackgroundDetector to fast-forward network reconnects when an app entered the foreground. Unfortunately, the BackgroundDetector is pretty conservative and only fires if it knows for certain that the app was in the background and now is in the foreground. This leads to some suppressed notifications. This PR just hooks directly into the notification API and always raises the foreground notification.

Logs:

```
2021-06-25 13:30:40.601 4251-4665/com.google.firebase.example.fireeats I/Firestore: (23.0.2) [OnlineStateTracker]: Could not reach Cloud Firestore backend. Connection failed 1 times. Most recent error: Status{code=UNAVAILABLE, description=Unable to resolve host foo, cause=java.lang.RuntimeException: java.net.UnknownHostException: Unable to resolve host "foo": No address associated with hostname
...
    This typically indicates that your device does not have a healthy Internet connection at the moment. The client will operate in offline mode until it is able to successfully connect to the backend.
2021-06-25 13:30:40.605 4251-4665/com.google.firebase.example.fireeats I/Firestore: (23.0.2) [ExponentialBackoff]: Backing off for 8220 ms (base delay: 7593 ms, delay with jitter: 8372 ms, last attempt: 152 ms ago)
...
2021-06-25 13:30:46.836 4251-4251/com.google.firebase.example.fireeats I/Firestore: (23.0.2) [AndroidConnectivityMonitor]: App has entered the foreground.
2021-06-25 13:30:46.843 4251-4665/com.google.firebase.example.fireeats I/Firestore: (23.0.2) [RemoteStore]: Restarting streams for network reachability change.
2021-06-25 13:30:46.845 4251-4665/com.google.firebase.example.fireeats I/Firestore: (23.0.2) [WatchStream]: (f298b8) Performing stream teardown
2021-06-25 13:30:46.866 4251-4665/com.google.firebase.example.fireeats I/Firestore: (23.0.2) [WatchStream]: (f298b8) Stream is open
```

Fixes https://github.com/firebase/firebase-android-sdk/issues/2637

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,8 @@
-# Unreleased
+# Unreelased
+- [changed] Increases the aggressiveness of network retires when an app's
+  visibility status changes.
+
+# 23.0.1
 - [changed] The SDK now tries to immediately establish a connection to the
   backend when the app enters the foreground.
 
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,6 +1,6 @@
-# Unreelased
+# Unreleased
 - [changed] Increases the aggressiveness of network retires when an app's
-  visibility status changes.
+  foreground status changes.
 
 # 23.0.1
 - [changed] The SDK now tries to immediately establish a connection to the
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
@@ -17,6 +17,8 @@ package com.google.firebase.firestore.remote;
 import static com.google.firebase.firestore.util.Assert.hardAssert;
 
 import android.annotation.TargetApi;
+import android.app.Activity;
+import android.app.Application;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
@@ -24,9 +26,11 @@ import android.content.IntentFilter;
 import android.net.ConnectivityManager;
 import android.net.Network;
 import android.os.Build;
+import android.os.Bundle;
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import com.google.android.gms.common.api.internal.BackgroundDetector;
 import com.google.firebase.firestore.util.Consumer;
+import com.google.firebase.firestore.util.Logger;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -36,8 +40,9 @@ import java.util.List;
  * <p>Implementation note: Most of the code here was shamelessly stolen from
  * https://github.com/grpc/grpc-java/blob/master/android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
  */
-public final class AndroidConnectivityMonitor
-    implements ConnectivityMonitor, BackgroundDetector.BackgroundStateChangeListener {
+public final class AndroidConnectivityMonitor implements ConnectivityMonitor {
+
+  private static final String LOG_TAG = "AndroidConnectivityMonitor";
 
   private final Context context;
   @Nullable private final ConnectivityManager connectivityManager;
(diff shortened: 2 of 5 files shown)


schmidt-sebastian:
Looks like this broke a test.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
```diff
@@ -78,34 +83,63 @@ private void configureNetworkMonitoring() {
...
+
+              @Override
+              public void onActivityStarted(@NonNull Activity activity) {}
+
+              @Override
+              public void onActivityResumed(@NonNull Activity activity) {
+                // Only raise the foreground notification if the same activity when to the
+                // background before. This prevents notifications when an app switches between
+                // activities.
+                if (activity == lastActivity) {
```
schmidt-sebastian, line 117: @samtstern Does this code make sense? I would like to detect when an app goes into the background and then back into the foreground. I assume that in this case, we will get a notification with the same activity, whereas in the normal app lifecycle we will receive this notification with a different activity.

Do you also happen to know if the memory address here will always be the same?

samtstern: I have a feeling `activity == lastActivity` is going to be too simple, as Android destroys activities all the time. That's why you need to do all the `savedInstanceState` stuff to handle device rotation.

But I don't know the more robust way to handle `Activity` comparison.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-# Unreleased
+# Unreelased
```
dconeybe, line 1: typo: "Unreelased"

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
...
+- [changed] Increases the aggressiveness of network retires when an app's
+  visibility status changes.
```
dconeybe, line 3: Is "visibility" the right term here? Would it make more sense to just say "foreground status"?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
dconeybe: Moreover, holding a strong reference to an `Activity` leaks the `Activity` instance, causing a memory leak. As Sam mentions, the `Activity` instances would compare unequal even if it was the same activity but some "configuration change" caused the `Activity` to be destroyed and re-created, such as a screen rotation.

What if you instead compared the activity's `Intent` for equality? Holding a strong reference to an `Intent` is not a memory leak and it more-or-less describes the `Activity`. It's not perfect, because, in theory, you can start a new `Activity` with the exact same `Intent` overtop of itself. But I think that situation is fairly rare.

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
schmidt-sebastian: I changed the implementation to also hook into the component callbacks, which offer a notification when an app's UI becomes hidden.  This is also used here: cs/piper///depot/google3/java/com/google/android/gmscore/integ/client/common/src/com/google/android/gms/common/api/internal/BackgroundDetector.java;l=140

The code is now very similar to BackgroundDetector, but it also fires if the state transition is not 100% certain. In my testing, it works to detect the transitions, whereas BackgroundDetector sometimes misses a transition. It is likely better to fire more often then to miss a transition.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
```diff
@@ -78,34 +86,72 @@ private void configureNetworkMonitoring() {
...
-              context.unregisterReceiver(networkReceiver);
-            }
-          };
+      unregisterRunnable = () -> context.unregisterReceiver(networkReceiver);
     }
   }
 
   private void configureBackgroundStateListener() {
-    BackgroundDetector.getInstance().addListener(this);
+    Application applicationContext = (Application) context.getApplicationContext();
```
dconeybe, line 100: nit: `applicationContext` can be renamed to `application` since it is being casted to an instance of `Application`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AndroidConnectivityMonitor.java
```diff
@@ -78,34 +86,72 @@ private void configureNetworkMonitoring() {
...
+        new Application.ActivityLifecycleCallbacks() {
+          @Override
+          public void onActivityCreated(@NonNull Activity activity, Bundle savedInstanceState) {}
+
+          @Override
+          public void onActivityStarted(@NonNull Activity activity) {}
+
+          @Override
+          public void onActivityResumed(@NonNull Activity activity) {
+            if (inBackground.compareAndSet(true, false)) {
```
dconeybe, line 116: Consider putting the body of `onActivityResumed()` into `onActivityCreated()` as well, so that the "offline" state is updated as early as possible (this is what `BackgroundDetector` does). Even copy it into `onActivityStarted()` too.

dconeybe (changes requested):

dconeybe (approved):

PR #2770, opened 2021-06-28
Set up firebase encoders in transport-runtime

James201311:
Set up firebase encoders in transport-runtime, it will be used to generate builder class and provide encoding methods for client metrics proto file.


transport/transport-runtime/code-gen-cfg.textproto
```diff
@@ -0,0 +1,5 @@
+# proto-file: firebase-android-sdk/transport/transport-runtime/src/main/proto/client_analytics.proto
+# proto-message: ClientMetrics
+
+vendor_package: "com.google.android.datatransport.runtime"
+include: "firebase.transport.ClientMetrics"
```
vkryachko, line 5: nit: add new line at the end

transport/transport-runtime/transport-runtime.gradle
```diff
@@ -47,6 +79,11 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
     sourceSets {
+        main {
+            proto {
+                srcDir 'src/main/proto'
```
vkryachko, line 84: why is this needed? afaict it should be the default even if you don't specify it

James201311: Adding the above line will stop the android studio compiling about the syntax error.

vkryachko: sgtm

vkryachko (approved):

PR #2774, opened 2021-06-29
Defined Client Analytics Store interface.

James201311:
Client Analytics Store interface will be used as  a Persistent Layer
It is responsible for storing, updating, and retrieving client analytics data.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/ClientAnalyticsStore.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/ClientAnalyticsStore.java
@@ -1,0 +1,47 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.android.datatransport.runtime.scheduling.persistence;
+
+import com.google.android.datatransport.runtime.firebase.transport.ClientMetrics;
+import com.google.android.datatransport.runtime.firebase.transport.LogEventDropped.Reason;
+
+/**
+ * Persistent Layer
+ *
+ * <p>Responsible for storing, updating, and retrieving client analytics data.
+ */
+public interface ClientAnalyticsStore {
+  /**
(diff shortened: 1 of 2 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/ClientAnalyticsStore.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.google.android.datatransport.runtime.firebase.transport.ClientMetrics;
+import com.google.android.datatransport.runtime.firebase.transport.LogEventDropped.Reason;
+
+/**
+ * Persistent Layer
+ *
+ * <p>Responsible for storing, updating, and retrieving client analytics data.
+ */
+public interface ClientAnalyticsStore {
```
vkryachko, line 25: imo the term "analytics" is very overloaded, can we use something more concrete like `ClientHealthMetricsStore` instead?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/ClientAnalyticsStore.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Record log event dropped.
+   *
+   * @param eventsDroppedCount number of events dropped
+   * @param reason reason why events are dropped
+   * @param logSource log source of the dropped events
+   */
+  void recordLogEventDropped(long eventsDroppedCount, Reason reason, String logSource);
```
vkryachko, line 33: Curious to know when this method will be called, i.e. will `eventsDroppedCount` ever be greater than `1`?

James201311: It will check if there is an existing row for that `reason` and `logSource`, if there is, it will increment the count value, if not, it will insert a new row.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/ClientAnalyticsStore.java
James201311: Will change that.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/ClientAnalyticsStore.java
vkryachko: I get that, but my question is: will we ever pass anything other than `1` as input to this method?

James201311: yes, we will. eg. if we trying to delete events which is older than 1 week, we would delete multiple events at that time.

vkryachko (approved):

PR #2775, opened 2021-06-29
Inject Context into SQLiteEventStore

James201311:
Inject Context into SQLiteEvent Store
 - Context will be used to get application package name for client analytics purpose

### DIFF
--- a/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/SpyEventStoreModule.java
+++ b/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/SpyEventStoreModule.java
@@ -22,6 +22,7 @@ import com.google.android.datatransport.runtime.time.Clock;
 import com.google.android.datatransport.runtime.time.Monotonic;
 import com.google.android.datatransport.runtime.time.WallTime;
 import dagger.Binds;
+import dagger.Lazy;
 import dagger.Module;
 import dagger.Provides;
 import javax.inject.Named;
@@ -41,7 +42,7 @@ public abstract class SpyEventStoreModule {
       @Monotonic Clock clock,
       EventStoreConfig config,
       SchemaManager schemaManager,
-      @Named("PACKAGE_NAME") String packageName) {
+      @Named("PACKAGE_NAME") Lazy<String> packageName) {
     return spy(new SQLiteEventStore(wallClock, clock, config, schemaManager, packageName));
   }
 
--- a/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/SpyEventStoreModule.java
+++ b/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/SpyEventStoreModule.java
@@ -17,7 +17,6 @@ package com.google.android.datatransport.runtime.scheduling.persistence;
 import static org.mockito.Mockito.spy;
 
 import android.content.Context;
-import androidx.test.platform.app.InstrumentationRegistry;
 import com.google.android.datatransport.runtime.synchronization.SynchronizationGuard;
 import com.google.android.datatransport.runtime.time.Clock;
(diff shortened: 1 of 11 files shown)


transport/transport-runtime/src/test/java/com/google/android/datatransport/runtime/scheduling/persistence/SchemaManagerTest.java
```diff
@@ -70,12 +72,14 @@
       EventStoreConfig.DEFAULT.toBuilder().setLoadBatchSize(5).setEventCleanUpAge(HOUR).build();
 
   private final TestClock clock = new TestClock(1);
+  Context context = InstrumentationRegistry.getInstrumentation().getContext();
```
vkryachko, line 75: `private final`?

vkryachko (commented):
pls use `ApplicationProvider.getApplicationContext()` instead of `InstrumentationRegistry` since it's more portable. throughout

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -63,18 +64,21 @@
...
+  private final Context context;
 
   @Inject
   SQLiteEventStore(
       @WallTime Clock wallClock,
       @Monotonic Clock clock,
       EventStoreConfig config,
-      SchemaManager schemaManager) {
+      SchemaManager schemaManager,
+      Context context) {
```
vkryachko, line 75: Actually, taking a step back, looks like context is only needed to get packageName. Consider injecting the package name directly instead of injecting `Context`[1].

[1] https://github.com/google/guice/wiki/InjectOnlyDirectDependencies

Can we do something like the following?
```suggestion
      @PackageName String packageName) {
```

See https://github.com/google/guice/wiki/BindingAnnotations for details.

transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/SpyEventStoreModule.java
```diff
@@ -39,8 +40,9 @@ static SQLiteEventStore sqliteEventStore(
       @WallTime Clock wallClock,
       @Monotonic Clock clock,
       EventStoreConfig config,
-      SchemaManager schemaManager) {
-    return spy(new SQLiteEventStore(wallClock, clock, config, schemaManager));
+      SchemaManager schemaManager,
+      @Named("PACKAGE_NAME") String packageName) {
```
vkryachko, line 44: Let's use a custom annotation instead of `@Named`, to quote [this page](https://github.com/google/guice/wiki/BindingAnnotations#named):

> Since the compiler can't check the string, we recommend using `@Named` sparingly. Defining your own purpose-built annotations provides better type-safety.

Here's an example of such an annotation: https://github.com/firebase/firebase-android-sdk/blob/7c1afeb999ba1452f6d38014cb394ee3d44f2bc4/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/time/Monotonic.java#L19-L20

James201311: The reason i choose `@Named` is that `EventSoreModuule` used `@Named("SQLITE_DB_NAME")` to provide `dbName`, just trying to be consist with it.

vkryachko: ok, sgtm then

vkryachko, line 44: Relatedly, can we inject a `Lazy<String>` instead of the string directly? We don't want to query the package name at startup, but rather only do so when it's needed

vkryachko (approved):

PR #2777, opened 2021-06-29
Update alternative-root-project to remove java formatting.

rlazo:


alternative-root-project.gradle
```diff
@@ -55,12 +57,4 @@ ext {
...
 
-configure(subprojects) {
-    afterEvaluate {
-        tasks.verifyGoogleJavaFormat {
-            exclude '**/build/classes/**'
-        }
-    }
-}
-
 apply plugin: MultiProjectReleasePlugin
```
vkryachko, line 60: Can we try to just apply this plugin to the main build file? it looks like it's the only difference from build.gradle

davidmotson (approved):

PR #2779, opened 2021-06-30
Implemented ClientAnalyticsStore

James201311:
This PR implemented the following three methods:
-`recordLogEventDropped`
  - this method will be called to record the number of events dropped for what reason and its original destination
  - It firstly check if there is an existing row for that `log_source` and `reason`. If so, it will increment the count of events_dropped_count; if not, it will create a new row with given attributes.

-`loadClientMetrics`
  - this method will return an full-populated `ClientMetrics`
  - In order to be fully populated:
    1. it populates persisted fields, including `log_source`, `reason`, and `eventsDroppedCount`. 
    2. it populates with start time(got from `global_log_event_state`) and end time(current time)
    3. it populates none-Persisted fields, including AppNamespace, current cache size, and max cache size.

-`resetClientMetrics`
 - It delete every rows in `log_event_dropped table`, and update `last_metrics_upload_ms` in `global_log_event_state` table to current time.
 - This method will be called after client analytics metrics was successfully sent to the server.


transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStoreTest.java
```diff
@@ -445,4 +493,200 @@ public void loadActiveContexts_whenTwoContextsWithOneAvailable_shouldReturnIt()
...
+        db ->
+            tryWithCursor(
+                db.rawQuery("SELECT 1 FROM log_event_dropped", new String[] {}),
+                cursor -> cursor.getCount() <= 0));
+  }
+
+  private static final Correspondence<LogSourceMetrics, LogSourceMetrics>
+      CLIENT_METRICS_CORRESPONDENCE =
+      Correspondence.from(
+          SQLiteEventStoreTest::compareLogSourceMetricsByFields, "compare by fields");
```
James201311, line 672: I used Fuzzy Truth to compare them by fields, since generated Proto Object doesn't override `equals` method.

vkryachko: Yeah, we have a bug for that b/182387176. I wonder if it would be easier to compare them via their encoded representation. i.e. via `Arrays.equals(actual.toByteArray(), expected.toByteArray())`?

This way we won't have to update this correspondence implementation if we ever add new fields. wdyt?

James201311: Another challenge is that, we used `List` to represent an collection of `LogSourceMetrics`, but write statement doesn't guarantee to write them in order, and we don't care about their order when comparing them, which could cause tests flaky.

vkryachko: The order of repeated fields `must` be preserved according to the spec[1]. If it not the case it's a bug, can you double check that it is indeed an issue?

[1] https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules

James201311: The issue is when we load the data from the db table, which is not necessarily the order we added them.

vkryachko: I see, makes sense

James201311:
> @James201311: The following test **failed**, say `/retest` to rerun them all:
> 
> Test name	Commit	Details	Rerun command
> check-changed	[106a101](https://github.com/firebase/firebase-android-sdk/commit/106a101097919165f180564b41c9b87ac7fe9d1f)	[link](https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/2779/check-changed/1410245526074429443/)	`/test check-changed`
> Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).

seems like `:encoders:protoc-gen-firebase-encoders` failed, but they passed tests locally, wdyt?

vkryachko:
> > @James201311: The following test **failed**, say `/retest` to rerun them all:
> > Test name	Commit	Details	Rerun command
> > check-changed	[106a101](https://github.com/firebase/firebase-android-sdk/commit/106a101097919165f180564b41c9b87ac7fe9d1f)	[link](https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/2779/check-changed/1410245526074429443/)	`/test check-changed`
> > Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md). If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).
> 
> seems like `:encoders:protoc-gen-firebase-encoders` failed, but they passed tests locally, wdyt? @vkryachko

Gradle could not pull deps from the network, rerunning should fix the issue

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -530,6 +537,145 @@ private static Encoding toEncoding(@Nullable String value) {
...
+
+    return inTransaction(
+        db ->
+            tryWithCursor(
+                db.rawQuery(query, new String[] {}),
+                cursor -> {
+                  while (cursor.moveToNext()) {
+                    String logSource = cursor.getString(0);
+                    LogEventDropped.Reason reason =
+                        LogEventDropped.Reason.values()[cursor.getInt(1)];
```
vkryachko, line 600: Using `values()` is not accurate as enum value indexes are not the same as the protobuf "number"(I mean it happens to work now but is fragile).

Looks like codegen is missing this functionality, created b/192450103.

We should either fix b/192450103 or come up with a workaround by e.g. creating a mapping from ints to Reason values.

James201311: having that functionality would be great, just did some research on `values()` as this [doc](https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.9.2) states: `values()` returns an array containing the constants of this enum type, **in the order** they're declared. And the firebase encoder generates those enum in order and the number of each enum is assigned from 0, plus the proto is also ordered from 0. In that case, this could be accurate. since they are correct, shall we keep it in that way and come back to change it until b/192450103 solved?

vkryachko: The issue is that nothing is preventing me from adding values to the proto out of order, e.g.

```proto
enum Foo {
  UNKNOWN = 0;
  BAR = 2;
  BAZ = 1;
}
```

This is why I would not recommend relying on the order which just `happens` to be identical to the order of java enums.

Because it is fragile, I suggest we choose the `ProtoUtil` way that you described or even just a private static method inside this class with a `switch(number)`, since this is the only place that needs this mapping

James201311: sg, will do that

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -531,6 +538,161 @@ private static Encoding toEncoding(@Nullable String value) {
...
+   */
+  @Override
+  public void recordLogEventDropped(
+      long eventsDroppedCount, LogEventDropped.Reason reason, String logSource) {
+    if (!isLogEventDroppedExist(reason, logSource)) {
+      ContentValues metrics = new ContentValues();
+      metrics.put("log_source", logSource);
+      metrics.put("reason", reason.getNumber());
+      metrics.put("events_dropped_count", eventsDroppedCount);
+      inTransaction(
```
vkryachko, line 556: afaict there's is a race condition here.

We are checking if the row exists in one transaction and then insert in another, I think this method needs to be atomic as a whole rather than use multiple transactions.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -531,6 +538,161 @@ private static Encoding toEncoding(@Nullable String value) {
...
+      return LogEventDropped.Reason.CACHE_FULL;
+    } else if (number == LogEventDropped.Reason.PAYLOAD_TOO_BIG.getNumber()) {
+      return LogEventDropped.Reason.PAYLOAD_TOO_BIG;
+    } else if (number == LogEventDropped.Reason.MAX_RETRIES_REACHED.getNumber()) {
+      return LogEventDropped.Reason.MAX_RETRIES_REACHED;
+    } else if (number == LogEventDropped.Reason.INVALID_PAYLOD.getNumber()) {
+      return LogEventDropped.Reason.INVALID_PAYLOD;
+    } else if (number == LogEventDropped.Reason.SERVER_ERROR.getNumber()) {
+      return LogEventDropped.Reason.SERVER_ERROR;
+    } else {
```
vkryachko, line 589: This whole block is a good candidate for a `switch()`, any reason you're not using it here?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -531,6 +538,161 @@ private static Encoding toEncoding(@Nullable String value) {
...
+      throw new IllegalArgumentException(
```
vkryachko, line 590: Not sure if it's better to throw of fallback to "UNKNOWN" with a log warning. For example I can't see how we can recover from a state where the db contains an unknown reason and we always throw when trying to read it. wdyt?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -531,6 +538,161 @@ private static Encoding toEncoding(@Nullable String value) {
...
+    for (Map.Entry<String, List<LogEventDropped>> entry : metricsMap.entrySet()) {
+      clientMetricsBuilder.addLogSourceMetrics(
+          LogSourceMetrics.newBuilder()
+              .setLogSource(entry.getKey())
+              .setLogEventDroppedList(entry.getValue())
+              .build());
+    }
+  }
+
+  private void setTimeWindow(ClientMetrics.Builder clientMetricsBuilder, long currentTime) {
```
vkryachko, line 653: looks like this method is very "side-effect"y, where it takes a bunch of arguments purely to mutate them. can we change this to instead return the `TimeWindow` object instead?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -531,6 +538,161 @@ private static Encoding toEncoding(@Nullable String value) {
...
+    inTransaction(
+        db ->
+            tryWithCursor(
+                db.rawQuery(
+                    "SELECT last_metrics_upload_ms FROM global_log_event_state", new String[] {}),
+                cursor -> {
+                  while (cursor.moveToNext()) {
```
vkryachko, line 660: Do we expect there to ever be more than 1 row? if not, should we add something like `limit 1`?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
James201311: Great point, i will do that.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
James201311: we cannot use `switch` here, since `LogEventDropped.Reason.INVALID_PAYLOD.getNumber()` is not an constant expression. If we go back to use `switch`, we will have to do `case 0:` `case 1:`, but i feel using `LogEventDropped.Reason.INVALID_PAYLOD.getNumber()` with `if else` is more clear.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
James201311: nice catch!!

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -548,27 +548,23 @@ private static Encoding toEncoding(@Nullable String value) {
...
-          db -> {
-            db.execSQL(query, new String[] {logSource, Integer.toString(reason.getNumber())});
-            return null;
-          });
-    }
+    String sql =
+        "INSERT INTO log_event_dropped (log_source, reason, events_dropped_count) "
+            + "VALUES(?, ?, ?) "
+            + "ON CONFLICT(log_source, reason) "
+            + "DO UPDATE SET events_dropped_count = events_dropped_count + ?";
```
vkryachko, line 555: It's really cool to see upsert, but I am not sure we can afford to use it. i.e. https://www.sqlite.org/lang_UPSERT.html mentions that it was added in version 3.24.0. While we support devices with level all the way down to 16+ https://stackoverflow.com/a/4377116

So we would have to do two queries like you used to do, but place them in one transaction.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
vkryachko: sgtm

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
James201311: didn't realize API 26 and API 27 [link](https://developer.android.com/reference/android/database/sqlite/package-summary) are using a bit older version of SQLite, but i will change it.

vkryachko (approved):

PR #2789, opened 2021-07-02
Restore state when appcheck is used without auth.

jsdt:
In the case where `authToken == null` and `appCheckToken != null`, this would never call `restoreState`. This fixes that case by calling `sendAppCheckTokenHelper` (that function also handles the case where `appCheckToken == null`.


schmidt-sebastian (approved):

schmidt-sebastian (commented):
Do you mind adding an entry to https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/CHANGELOG.md so we know what changed for the next release?

PR #2793, opened 2021-07-05
Extend TransportBackend to have a default method

James201311:


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -396,6 +397,11 @@ public BackendResponse send(BackendRequest request) {
     }
   }
 
+  @Override
+  public boolean shouldUploadClientHealthMetric(TransportContext transportContext) {
+    return transportContext.getExtras() != null;
```
James201311, line 402: >  CCT backend supports 2 different endpoints
>  We route to CCT backend if extras are null and to LegacyFlg otherwise.
>  This (anti-) pattern should not be required for other backends

[Reference](https://github.com/firebase/firebase-android-sdk/blob/2bb33648c3c3fae8fea3551226d098a133b9a82f/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L351)


CCT and FLG are unified into one `TransportBackend`, it sends to FLG if the transportContext's extras is not null.

PR #2797, opened 2021-07-05
Extend TransportBackend to have a default method

James201311:
Extend `TransportBackend` interface to have a default method `shouldUploadClientHealthMetrics`

`shouldUploadClientHealthMetrics`  indicate whether data sent through `TransportBackend` with
`TransportContext` should be recorded as client health metrics and upload to Flg server.

### DIFF
--- a/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
+++ b/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
@@ -36,6 +36,7 @@ import com.google.android.datatransport.cct.internal.NetworkConnectionInfo;
 import com.google.android.datatransport.cct.internal.QosTier;
 import com.google.android.datatransport.runtime.EncodedPayload;
 import com.google.android.datatransport.runtime.EventInternal;
+import com.google.android.datatransport.runtime.TransportContext;
 import com.google.android.datatransport.runtime.backends.BackendRequest;
 import com.google.android.datatransport.runtime.backends.BackendResponse;
 import com.google.android.datatransport.runtime.backends.TransportBackend;
@@ -394,6 +395,11 @@ final class CctTransportBackend implements TransportBackend {
       Logging.e(LOG_TAG, "Could not make request to the backend", e);
       return BackendResponse.transientError();
     }
+  }
+
+  @Override
+  public boolean shouldUploadClientHealthMetrics(TransportContext transportContext) {
+    return transportContext.getExtras() != null;
   }
 
   @VisibleForTesting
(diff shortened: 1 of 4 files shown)


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -396,6 +397,11 @@ public BackendResponse send(BackendRequest request) {
     }
   }
 
+  @Override
+  public boolean shouldUploadClientHealthMetrics(TransportContext transportContext) {
+    return transportContext.getExtras() != null;
+  }
```
James201311, line 403: >  CCT backend supports 2 different endpoints
>  We route to CCT backend if extras are null and to LegacyFlg otherwise.
>  This (anti-) pattern should not be required for other backends

[Reference](https://github.com/firebase/firebase-android-sdk/blob/2bb33648c3c3fae8fea3551226d098a133b9a82f/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java#L351)


CCT and FLG are unified into one `TransportBackend`, it sends to FLG if the transportContext's extras is not null.

James201311: Therefore, in order to only record events sent to Flg, `CctTransportBackend` overrides `shouldUploadClientHealthMetrics`, and
- return true if it sends request to Flg(transportContext's extras is not null)
- return false if it sends request to Cct(transportContext's extras is null)

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/TransportBackend.java
```diff
@@ -15,9 +15,20 @@
...
 
   BackendResponse send(BackendRequest backendRequest);
+
+  /**
+   * This method indicate whether data sent through {@link TransportBackend} with {@link
+   * TransportContext} should be recorded as client health metrics and upload to Flg server.
+   *
+   * <p>By default, it returns false, and events will not be recorded and uploaded to Flg server.
+   */
+  default boolean shouldUploadClientHealthMetrics(TransportContext transportContext) {
```
vkryachko, line 31: To future proof this, a few  suggestions:

1. let's not return a plain boolean but rather return an `UploadOptions` object that has a boolean property
2. let's name the method a bit more generically, i.e.

```java
default UploadOptions getUploadOptions(TransportContext transportContext) {
  return UploadOptions.none();
}
```

The thinking behind it is that it will make it easier to add future `options`. i.e. we won't have introduce new methods to this interface but just add more properties to UploadOptions.

James201311: Great point! It definitely makes it more scalable.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/backends/TransportBackend.java
```diff
@@ -15,9 +15,20 @@
...
 import com.google.android.datatransport.runtime.EventInternal;
+import com.google.android.datatransport.runtime.TransportContext;
 
 public interface TransportBackend {
   EventInternal decorate(EventInternal event);
 
   BackendResponse send(BackendRequest backendRequest);
+
+  /**
+   * This method indicate whether data sent through {@link TransportBackend} with {@link
```
vkryachko, line 26: Can you update the javadoc?

vkryachko (approved):
LGTM, with a javadoc nit

PR #2799, opened 2021-07-06
Inject backend registry into SQLiteEventStore

James201311:
Inject `BaclendRegistry` into `SQLiteEventStore`
- `BaclendRegistry` will be used to get `TransportBackend` through `backendName` to determine its upload options.


James201311:
https://github.com/firebase/firebase-android-sdk/pull/2801
enable us to determin if we should record and upload client metrics. Therefore, this pr is no longer needed.

PR #2802, opened 2021-07-06
Record log event dropped due to cache full

James201311:
Before persisting the events to database, it firstly check if the storage limit has been reached
- if so, it will drop the events and make a call to `ClientHealthMetricsStore` to `recordLogEventDropped` for the reason of `CACHE_FULL`.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
@@ -114,6 +114,8 @@ public class SQLiteEventStore
               // TODO(vkryachko): come up with a more sophisticated algorithm for limiting disk
               // space.
               if (isStorageAtLimit()) {
+                recordLogEventDropped(
+                    1, LogEventDropped.Reason.CACHE_FULL, event.getTransportName());
                 return -1L;
               }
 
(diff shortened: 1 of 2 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -114,6 +114,8 @@ public PersistedEvent persist(TransportContext transportContext, EventInternal e
               // TODO(vkryachko): come up with a more sophisticated algorithm for limiting disk
               // space.
               if (isStorageAtLimit()) {
+                recordLogEventDropped(
```
vkryachko, line 117: Weren't you planning to check the `transportContext` if we want to log or not?

James201311: we could record all log events, and the log events dropped count won't be aggregated, since events all come with log_source. Besides, i don't think those extra metrics will take too much space, record them could be a bonus?
we will only use `transportContext` to determine if we want to upload.

vkryachko: sgtm, let's make sure that ingestion will not fail in the face of "unknown" log sources.

vkryachko (approved):

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
James201311: It will be ok, as long as the logsource is registered in `LogSource` at [here](google3/wireless/android/play/playlog/proto/clientanalytics.proto). afaik, any client needs to register their own logsource if they plan to send data to cct.
`UNKNOWN` logsource itself is registered, so it would be fine.
Any non-registered logsource could possible be here?

PR #2803, opened 2021-07-06
Record logevent dropped due to too old

James201311:
Before old events are deleted, we will make a call to `ClientHealthMetricsStore` to `recordLogEventDropped` for the reason of `MESSAGE_TOO_OLD`.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
@@ -352,7 +352,27 @@ public class SQLiteEventStore
   public int cleanUp() {
     long oneWeekAgo = wallClock.getTime() - config.getEventCleanUpAge();
     return inTransaction(
-        db -> db.delete("events", "timestamp_ms < ?", new String[] {String.valueOf(oneWeekAgo)}));
+        db -> {
+          String query =
+              "SELECT COUNT(*), transport_name FROM events "
+                  + "INNER JOIN transport_contexts ON (events.context_id=transport_contexts._id) "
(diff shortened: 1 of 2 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -352,7 +352,27 @@ public void recordNextCallTime(TransportContext transportContext, long timestamp
   public int cleanUp() {
     long oneWeekAgo = wallClock.getTime() - config.getEventCleanUpAge();
     return inTransaction(
-        db -> db.delete("events", "timestamp_ms < ?", new String[] {String.valueOf(oneWeekAgo)}));
+        db -> {
+          String query =
+              "SELECT COUNT(*), transport_name FROM events "
+                  + "INNER JOIN transport_contexts ON (events.context_id=transport_contexts._id) "
+                  + "WHERE timestamp_ms < ? "
+                  + "GROUP BY transport_name";
```
vkryachko, line 360: Why do we need a join here? can't we just remove it?

In general I think there is risk that this query could be very expensive. Can you run some `explain` queries against the database and provide the output?

James201311: Originally, i was planning to check the transportContext if we want to log or not, in that case, we will use 

```
            SELECT COUNT(*), transport_name, backend_name, priority, extras FROM events 
            INNER JOIN transport_contexts ON (events.context_id=transport_contexts._id) 
            WHERE timestamp_ms < ? 
            GROUP BY transport_name, backend_name, priority, extras
```
in that case, we will not need to join those two tables, will remove it.

vkryachko (approved):

PR #2814, opened 2021-07-08
Record reason unknown

James201311:
Currently, we have four types of Network Response Status: **OK**,  **TRANSIENT_ERROR**, 
 **FATAL_ERROR**,  **INVALID_PAYLOAD**

CctTransportBackend will return 
- **OK** if the network response code is 200
- **TRANSIENT_ERROR** if the network response code is greater than or equal to 500, or equal to 404.
- **INVALID_PAYLOAD** if the network response code is 400
- **FATAL_ERROR** if the network response code is anything other than above, including 100-199, 201-399, 401-403, 405-499 

If LogEvents are dropped due to **FATAL_ERROR**, we call `ClientHealthMetricsStore` to `recordLogEventDropped` for the reason of `REASON_UNKNOWN`;


vkryachko (commented):
I don't think we should be recording `unknown`. This enum value is present only by convention and should be unused, see https://google.aip.dev/126

Issue #2805, opened 2021-07-07
loadBundle() with UTF8 chars throw error on android


aBuder:
#### Steps to reproduce:

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

#### Relevant Code:

```
// TODO(you): code here to reproduce the problem
```

**Describe the bug**
loadBundle() function works fine on iOS and throws an error on android. I have the special character `Ö` in the text of the Cloud Firestore document. If I remove this character `Ö` it will work on android. It looks like UTF-8 chars not working.

### Screenshot Cloud Firestore Document
<img width="577" alt="Bildschirmfoto 2021-07-02 um 17 18 27" src="https://user-images.githubusercontent.com/1265139/124295822-91d28f80-db59-11eb-9e79-597233d95478.png">

Via an external tool, we create a bundle file of Cloud Firestore collections. If I  load the bundle on iOS it works fine, but on Android, I get this error:

```log
Exception has occurred.
FirebaseException ([cloud_firestore/load-bundle-error] Client specified an invalid argument. Note that this differs from failed-precondition. invalid-argument indicates arguments that are problematic regardless of the state of the system (e.g., an invalid field name).)
```

### Bundle file content

```txt
[...]

PR #2820, opened 2021-07-12
Fix an issue with loading bundles with multi-byte character

wu-hui:
Both `BundleReader` and the testing `BundlerBuilder` should be counting UTF-8 bytes, instead of Java `char` count.

This is a fix for: https://github.com/firebase/firebase-android-sdk/issues/2805


firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -153,29 +152,50 @@ private int indexOfOpenBracket() {
     }
   }
 
+  private class ReadJsonResult {
```
dconeybe, line 155: drive-by comment: this class should be `static`, and probably `final`.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -153,29 +152,50 @@ private int indexOfOpenBracket() {
...
       if (buffer.remaining() == 0 && !pullMoreData()) {
         throw abort("Reached the end of bundle when more data was expected.");
       }
 
       int read = Math.min(remaining, buffer.remaining());
-      json.append(buffer, 0, read);
-      buffer.position(buffer.position() + read);
+      byte[] bytes = new byte[read];
+      buffer.get(bytes);
+      json.append(new String(bytes, charset));
```
dconeybe, line 192: drive-by comment: This is dangerous because `bytes` could end in the middle of a multi-byte character. You could possibly use a class like `java.nio.charset.CharsetDecoder` to handle this properly. 

https://docs.oracle.com/javase/7/docs/api/java/nio/charset/CharsetEncoder.html

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: Done.

dconeybe: The change doesn't fix the issue. For example, consider the situation where the last byte of `bytes` specified to `new String(bytes, charset)` is the first byte of a multi-byte character, and the second byte of that character would be the first byte of the `bytes` in the next iteration; this would probably either fail decoding with an exception or would produce garbage.

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: This is the bytes the bundle tells the SDK that it is a UTF8 string. If what you described happens, it means the bundle is invalid, and an exception is exactly what we want here.

dconeybe: IIUC the code reads the bundle's bytes in "chunks". Wouldn't there be arbitrary breaks in the "chunks" such that a multibyte character could span two chunks?

dconeybe: But now, what if `bytesToRead` > `BUFFER_CAPACITY`? IIUC, with the updated code, if the buffer fills up then `pullMoreData()` will return `false` and `readJson()` will incorrectly throw `IllegalArgumentException`.

I have another idea. What if you go back to the original implementation but change `StringBuilder` to `ByteArrayOutputStream`. You can then accumulate the bytes in the `ByteArrayOutputStream` and, once all bytes are read, you can decode the entire byte sequence, avoiding the multi-byte-character-spanning-a-chunk issue.

wu-hui: Yeah, that is better.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,10 +1,18 @@
-# Unreleased
-- [changed] Increases the aggressiveness of network retires when an app's
+# 23.0.3
+- [fixed] Fixed an issue when loading a data bundle with
+  multi-bytes unicode character leads to failures.
```
dconeybe, line 3: nit: "multi-bytes unicode character" -> "multi-byte Unicode characters". The "U" in "Unicode" should be uppercase and the plural needs to be tweaked.

wu-hui: Done.

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -35,17 +35,17 @@
   protected static final int BUFFER_CAPACITY = 1024;
 
   private final BundleSerializer serializer;
-  private final InputStreamReader dataReader;
+  private final InputStream bundleInputStream;
   private final Charset charset = Charset.forName("UTF-8");
```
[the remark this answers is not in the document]
schmidt-sebastian: Unfortunately, this was only added with API level 19.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -101,8 +101,9 @@ private BundleElement readNextElement() throws IOException, JSONException {
       return null;
     }
 
-    String json = readJsonString(Integer.parseInt(lengthPrefix));
-    bytesRead += lengthPrefix.length() + json.getBytes(charset).length;
+    int jsonStringByteCount = Integer.parseInt(lengthPrefix);
+    String json = readJsonString(jsonStringByteCount);
+    bytesRead += lengthPrefix.getBytes(charset).length + jsonStringByteCount;
```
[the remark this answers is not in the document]
schmidt-sebastian: This should just be the lengthPrefix, which should not have UTF-8 as it is just a number.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -133,9 +134,9 @@ private BundleElement readNextElement() throws IOException, JSONException {
       throw abort("Reached the end of bundle when a length string is expected.");
     }
 
-    char[] c = new char[nextOpenBracket];
+    byte[] c = new byte[nextOpenBracket];
     buffer.get(c);
-    return new String(c);
+    return charset.decode(ByteBuffer.wrap(c)).toString();
```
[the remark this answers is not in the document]
schmidt-sebastian: Note that `indexOfOpenBracket` returns the char-index of the first bracket. Using a byte buffer here is not correct. but likely won't make a difference since this part of the input should be ASCII.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -157,25 +158,30 @@ private int indexOfOpenBracket() {
...
     while (remaining > 0) {
       if (buffer.remaining() == 0 && !pullMoreData()) {
         throw abort("Reached the end of bundle when more data was expected.");
       }
 
+      // `read` is the number of bytes guaranteed to exist in `this.buffer` after the above
+      // call to `pullMoreData`. Copy them to `jsonBytes` and advance `this.buffer`'s position.
       int read = Math.min(remaining, buffer.remaining());
-      json.append(buffer, 0, read);
+      jsonBytes.write(buffer.array(), buffer.position(), read);
```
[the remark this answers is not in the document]
schmidt-sebastian: More over, `buffer` is a CharBuffer. CharBuffer deals with UTF-8 direcltly:

```
            String unicode = "\uD83D\uDE0A";
            byte[] b = unicode.getBytes();
            CharBuffer buffer = CharBuffer.allocate(10);
            new InputStreamReader(new ByteArrayInputStream(b)).read(buffer);
            StringBuilder result = new StringBuilder();
            buffer.flip();
            System.out.println(buffer.remaining());
            result.append(buffer);
            System.out.println(result);
```

Prints:
```
2
😊
```

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/BundleTest.java
```diff
@@ -238,18 +239,22 @@ private void verifyQueryResults() throws ExecutionException, InterruptedExceptio
...
     QuerySnapshot limitSnapshot = Tasks.await(limitQuery.get(Source.CACHE));
-    assertEquals(asList(map("bar", 2L, "k", "b")), querySnapshotToValues(limitSnapshot));
+    assertEquals(asList(map("bar", 2L, "k", "\uD83D\uDE0A")), querySnapshotToValues(limitSnapshot));
 
     Query limitToLastQuery = Tasks.await(db.getNamedQuery("limit-to-last"));
     QuerySnapshot limitToLastSnapshot = Tasks.await(limitToLastQuery.get(Source.CACHE));
     assertEquals(asList(map("bar", 1L, "k", "a")), querySnapshotToValues(limitToLastSnapshot));
   }
 
+  private int getUTF8BytesCount(String s) {
```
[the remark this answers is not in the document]
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: It's only supported after KitKat, which I think is why we use string name here.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/BundleTest.java
```diff
@@ -71,7 +72,7 @@
             + "{\"seconds\":1000,\"nanos\":9999},\"exists\":true}}",
         "{\"document\":{\"name\":\"projects/{projectId}/databases/(default)/documents/coll-1/b\","
             + "\"createTime\":{\"seconds\":1,\"nanos\":9},\"updateTime\":{\"seconds\":1,"
-            + "\"nanos\":9},\"fields\":{\"k\":{\"stringValue\":\"b\"},\"bar\":"
+            + "\"nanos\":9},\"fields\":{\"k\":{\"stringValue\":\"\uD83D\uDE0A\"},\"bar\":"
```
schmidt-sebastian, line 75: Note that this test passes against `master` as well.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -185,9 +186,12 @@ private String readJsonString(int length) throws IOException {
    */
   private boolean pullMoreData() throws IOException {
     buffer.compact();
-    int read = dataReader.read(buffer);
+    int bytesToRead = Math.min(bundleInputStream.available(), buffer.remaining());
```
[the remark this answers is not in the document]
schmidt-sebastian: We need a test that hits this issue.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -185,9 +186,12 @@ private String readJsonString(int length) throws IOException {
...
+    byte[] bytes = new byte[bytesToRead];
```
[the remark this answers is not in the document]
schmidt-sebastian: s/Consider reading/Please read

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: This is only for the length string, which is typically 3 to 4 bytes. IMO, it's not worth for the optimization.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: See above.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
dconeybe: Ahh nevermind. I'll resolve this comment.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: Similarly, I feel better to avoid performance optimization without proof. The end result is some byte counting, which is bad for readability.

The other place we do this is acceptable, because there is no good alternative..

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
dconeybe: FYI `buffer` was changed from `CharBuffer` to `ByteBuffer` in one of the recent commits.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
schmidt-sebastian: Got it, but the existing code uses it throughout and creates the initial input stream using UTF-8. It should work (and the test passes), so I am a bit concerned that the bug is somewhere else.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/BundleTest.java
wu-hui: Because the bundles built here also count chars, not bytes.

wu-hui: See other change in this test, it uses byte count, which fails the tests.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -191,12 +192,23 @@ private String readJsonString(int bytesToRead) throws IOException {
...
+    // forward.
+    if (available == 0) {
+      bytesToRead = buffer.remaining();
+    }
     byte[] bytes = new byte[bytesToRead];
     int bytesRead = bundleInputStream.read(bytes);
-    buffer.put(bytes);
+    boolean readSuccess = bytesRead > 0;
+    if (readSuccess) {
+      buffer.put(bytes, 0, bytesRead);
```
dconeybe, line 207: Could you make the call to `buffer.put(bytes, 0, bytesRead)` unconditional? Or is specifying `bytesRead==0` going to cause problems?

dconeybe (approved):
I approve, but please ensure that Sebastian's comments are resolved.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: Yeah, the tests pass because the test bundle builder also count characters, not bytes.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/BundleTest.java
schmidt-sebastian: Sorry, I just saw this flaw in my logic. So the bug is that we try to read more data than available.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: To reply to Sebastian: I am not sure how to do this without mocking, `available()` is stated to be an estimate, which is pretty accurate for most implementations.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
schmidt-sebastian: You are doubling the amount of memory consumed here. Gil pushed back heavily against this in the original PR.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: `bytesRead` can be -1, that is the problem.

dconeybe: Ahh I see. Makes sense.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -34,18 +34,19 @@
...
   @Nullable BundleMetadata metadata;
-  private CharBuffer buffer;
+  private ByteBuffer buffer;
   long bytesRead;
 
   public BundleReader(BundleSerializer serializer, InputStream data) {
     this.serializer = serializer;
-    dataReader = new InputStreamReader(data, charset);
-    buffer = CharBuffer.allocate(BUFFER_CAPACITY);
+    bundleInputStream = data;
```
schmidt-sebastian, line 48: Can you rename the argument to `bundleInputStream` as well? We could also drop the prefix similar to `serializer`.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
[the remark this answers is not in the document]
wu-hui: I still think this is premature..but done anyways.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -133,9 +135,9 @@ private BundleElement readNextElement() throws IOException, JSONException {
       throw abort("Reached the end of bundle when a length string is expected.");
     }
 
-    char[] c = new char[nextOpenBracket];
+    byte[] c = new byte[nextOpenBracket];
```
schmidt-sebastian, line 138: Nit: s/c/b

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
```diff
@@ -185,9 +192,24 @@ private String readJsonString(int length) throws IOException {
...
     buffer.compact();
-    int read = dataReader.read(buffer);
+
+    int available = bundleInputStream.available();
+    int bytesToRead = Math.min(available, buffer.remaining());
+    // `available` is an estimation, we still try to read if the estimation is 0, to move things
+    // forward.
+    if (available == 0) {
+      bytesToRead = buffer.remaining();
+    }
```
schmidt-sebastian, line 202: Can we always read buffer.remaining?

schmidt-sebastian (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/bundle/BundleReader.java
wu-hui: Yes, since we are reading into the buffer directly now!

schmidt-sebastian (approved):
LGTM

PR #2821, opened 2021-07-12
Include ClientMetrics into BachedLogRequest

James201311:
Added an additional `EventInteral` to `BackEndRequest` to send `ClientMetrics`


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -43,6 +47,7 @@
 public class Uploader {
 
   private static final String LOG_TAG = "Uploader";
+  private static final String CLIENT_HEALTH_METRICS_LOG_SOURCE = "1710";
```
vkryachko, line 50: It would be better to call it by name, since it has fewer compatibility issues, can we do that?

vkryachko (approved):

PR #2822, opened 2021-07-12
Reset Client Health Metrics, if it is too old.

James201311:
Reset CLient Health Metrics, if it's one week old.


vkryachko (commented):
I don't recall this discussion, is it actually desired to delete events older than 1 week?

James201311:
The plan was to remove old metrics to avoid it taking up too much storage space, since the storage size will be linear to the number of log sources, and its size is negligible. We will not have this implementation.

PR #2823, opened 2021-07-12
Add global db storage limit

James201311:
Add `MAX_DB_GLOBAL_STORAGE_SIZE_IN_BYTES` to store the data which needs to be beyond the current limit, such as recording cache full records in `ClientHealthMetrics`

### DIFF
--- a/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/TestEventStoreModule.java
+++ b/transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/TestEventStoreModule.java
@@ -24,6 +24,7 @@ import javax.inject.Named;
 @Module
 public abstract class TestEventStoreModule {
   private static final long MAX_DB_STORAGE_SIZE_IN_BYTES = 10 * 1024 * 1024;
+  private static final long MAX_DB_GLOBAL_STORAGE_SIZE_IN_BYTES = 12 * 1024 * 1024;
   private static final int LOAD_BATCH_SIZE = 200;
   private static final int LOCK_TIME_OUT_MS = 0;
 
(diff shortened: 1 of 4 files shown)


transport/transport-runtime/src/androidTest/java/com/google/android/datatransport/runtime/scheduling/persistence/TestEventStoreModule.java
```diff
@@ -24,13 +24,15 @@
 @Module
 public abstract class TestEventStoreModule {
   private static final long MAX_DB_STORAGE_SIZE_IN_BYTES = 10 * 1024 * 1024;
+  private static final long MAX_DB_GLOBAL_STORAGE_SIZE_IN_BYTES = 12 * 1024 * 1024;
```
vkryachko, line 27: I think it would be clearer to have a dedicated limit value of 2MB with a name along the lines of `maxHealthMetricsStorageSize`, wdyt?

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/persistence/SQLiteEventStore.java
```diff
@@ -587,6 +587,9 @@ public void recordLogEventDropped(
       long eventsDroppedCount, LogEventDropped.Reason reason, String logSource) {
     inTransaction(
         db -> {
+          if (isStorageAtGlobalLimit()) {
```
vkryachko, line 590: It looks like the `isStorageAtLimit` method will act differently now, i.e. it will account health metrics, effectively reducing the size to 8-10MB(depending on how many health metric events we have), are we ok with that?

vkryachko (commented):
See some comments below, and also I am curious if we need to worry about limiting the size of metrics storage? i.e. afaict we store constant amount of info per log source, so the storage size will be linear to the number of log sources present in the app, whose upper bound may be low enough to be negligible, wdyt?

James201311:
> See some comments below, and also I am curious if we need to worry about limiting the size of metrics storage? i.e. afaict we store constant amount of info per log source, so the storage size will be linear to the number of log sources present in the app, whose upper bound may be low enough to be negligible, wdyt?

Yes, I agree. 
since the storage size will be linear to the number of log sources, and its size is negligible. I don't think this implementation will bring us benefits, shall we just close this pr? @vkryachko

vkryachko:
+1 to not pursuing this further

PR #2825, opened 2021-07-13
Rollback #2785, which made bom generation and crashlytics smoke test fail.

yifanyang:


yifanyang:
@davidmotson @vkryachko 

The problems are:
- `make-bom` task is also configured to use sdk gradle [images/ci/build_bom.sh](https://team.git.corp.google.com/firebase-android-core/test-infra/+/refs/heads/master/images/ci/build_bom.sh#25)
- #2737 had the smoke test run before #2785, but was merged after it, and is now failing with [some error](https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/2825/smoke-tests/1414739764166266884/)

Despite of the crashlytics errors, to fix the bom task, I will need to build and push a new CI image. It's probably easier to rollback at this moment. Can you help me approve this PR?

The current approach of decoupling sdk and smoke test gradle requires below changes to be pushed (or reverted) at the same time:
- smoke test src change in this repo
- firehorn runner change in its docker image (_already rolled back now_)
- build_bom script change in the CI image

And not sure if I'm still missing some use case. I'll try to think of a cleaner way to do this.

vkryachko (approved):

PR #2830, opened 2021-07-14
Adding AppDistributionReleaseInternal and AAB release identification

pranavrajgopal:
* Added release identification for AABs
* Added new AppDistributionReleaseInternal class that represents the response from the server
* refactored some tests

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -385,7 +385,8 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
       long currentInstalledVersionCode =
           getInstalledAppVersionCode(firebaseApp.getApplicationContext());
 
-      if (isNewerBuildVersion(retrievedLatestRelease) && !isInstalledRelease(retrievedLatestRelease)) {
+      if (isNewerBuildVersion(retrievedLatestRelease)
+          && !isInstalledRelease(retrievedLatestRelease)) {
         return convertToAppDistributionRelease(retrievedLatestRelease);
       } else {
         // Return null if retrieved latest release is older or currently installed
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -42,6 +42,8 @@ import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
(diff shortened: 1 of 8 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -166,28 +168,26 @@ public void onFailure(@NonNull @NotNull Exception e) {
     Task<InstallationTokenResult> installationAuthTokenTask =
         firebaseInstallationsApi.getToken(true);
 
-    Tasks.whenAllComplete(installationIdTask, installationAuthTokenTask)
```
pranavrajgopal, line 169: noticed that we use `whenAllComplete`. This should actually be `whenAllSuccess` since we want it to fail if the installation task fails

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -169,28 +171,26 @@ public void onFailure(@NonNull Exception e) {
...
-                            firebaseApp.getOptions().getApiKey(),
-                            installationTokenResult.getToken());
-
-                    updateOnUiThread(
-                        new Runnable() {
-                          @Override
-                          public void run() {
-                            checkForUpdateTaskCompletionSource.setResult(latestRelease);
-                          }
-                        });
```
pranavrajgopal, line 193: Updated getLatestRelease to throw an exception instead of directly updating the checkForUpdate completion error. This way, we handle the error and success completion in one place.

rachaprince (commented):
This history seems to contain changes that should already be in the branch checkforupdate-mvp. 

I wonder if we need to rebase after the force push?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
rachaprince: good catch!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/AppDistributionReleaseInternal.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /** The release notes for this build */
+  @Nullable
+  public abstract String getReleaseNotes();
+
+  /** The binary type for this build */
+  @NonNull
+  public abstract BinaryType getBinaryType();
+
+  @Nullable
```
rachaprince, line 54: nit: add comments to new fields

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/ReleaseIdentificationUtils.java
```diff
@@ -0,0 +1,43 @@
+// Copyright 2021 Google LLC
```
rachaprince, line 1: I like this restructuring

rachaprince (approved):

rachaprince (approved):

PR #2835, opened 2021-07-16
FAD API error handling

pranavrajgopal:
Adding API error handling and better error messages

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -47,7 +47,6 @@ import com.google.firebase.appdistribution.internal.AppDistributionReleaseIntern
 import com.google.firebase.appdistribution.internal.ReleaseIdentificationUtils;
 import com.google.firebase.installations.FirebaseInstallationsApi;
 import com.google.firebase.installations.InstallationTokenResult;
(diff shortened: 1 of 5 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionException.java
```diff
@@ -58,17 +60,6 @@
   @NonNull private final Status status;
   @Nullable private final AppDistributionRelease release;
```
pranavrajgopal, line 62: Removing these because we should always pass a message to the FirebaseAppDistributionException since its customer facing.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -362,7 +366,8 @@ public void onFailure(@NonNull Exception e) {
           @Override
           public void onClick(DialogInterface dialogInterface, int i) {
             setSignInTaskCompletionError(
-                new FirebaseAppDistributionException(AUTHENTICATION_CANCELED));
+                new FirebaseAppDistributionException(
+                    "Tester cancelled authentication flow", AUTHENTICATION_CANCELED));
```
rachaprince, line 370: Should this string also be in the constants class?

rachaprince (approved):

PR #2838, opened 2021-07-16
Add environment flag to control indexing

schmidt-sebastian:
We are starting with development for Firestore's long awaited indexing support. This PR adds a environment variable that controls whether indexing is enabled in the final build. It is meant to replace the hard-coded flag as we do not want to accidentally change it to true and check it in to master.

### DIFF
--- a/firebase-firestore/firebase-firestore.gradle
+++ b/firebase-firestore/firebase-firestore.gradle
@@ -96,6 +96,12 @@ android.libraryVariants.all { variant ->
         variant.buildConfigField("boolean", "USE_EMULATOR_FOR_TESTS", "true")
     }
 
+    Boolean enableIndexing = System.env.FIRESTOR_ENABLE_INDEXING ?: false
+    if (enableIndexing) {
+        variant.buildConfigField("boolean", "ENABLE_INDEXING", "true")
+    } else {
+        variant.buildConfigField("boolean", "ENABLE_INDEXING", "false")
+    }
 }
 
 configurations.all {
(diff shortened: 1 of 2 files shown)


schmidt-sebastian:
cc @thebrianchen @wu-hui

firebase-firestore/firebase-firestore.gradle
```diff
@@ -96,6 +96,12 @@ android.libraryVariants.all { variant ->
         variant.buildConfigField("boolean", "USE_EMULATOR_FOR_TESTS", "true")
     }
 
+    Boolean enableIndexing = System.env.FIRESTOR_ENABLE_INDEXING ?: false
```
vkryachko, line 99: I don't feel strongly, but looks like above you are using a gradle `property`(`./gradlew -PuseProdBackendForTests=1 ...`) to control emulator behavior, would it make sense to do the same here for consistency?

schmidt-sebastian: I think we used the Gradle property there to make it easy to configure the Emulator for test runs on CI. If I change this new environment variable to a Gradle property, is that something I can control via `local.properties`? If so, it might actually be preferred as I then won't have to start AS in a context that includes the environment variable.

vkryachko:
I think `local.properties` is an Android Gradle plugin thing and is not used by gradle itself automatically. So you have two options:
1. use `gradle.properties` instead(making sure not to include this change in your commits
2. read `local.properties` yourself(handling file not found gracefully) i.e.

```groovy
new Properties().with {properties ->
    file("locals.properties").withReader {reader ->
        properties.load(reader)
    }
    println properties['firestoreEnableIndexing']
}
```

firebase-firestore/firebase-firestore.gradle
schmidt-sebastian: Updated to read from local.properties.

vkryachko (approved):

PR #2842, opened 2021-07-19
FAD Implement updateApp MVP for AABs

andrewbibiloni:
- Add implementation for updateApp for AABs 
- Create UpdateTaskImpl class to return in updateApp 
- Add unit tests

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -393,6 +393,8 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
       AppDistributionReleaseInternal retrievedLatestRelease =
           firebaseAppDistributionTesterApiClient.fetchLatestRelease(fid, appId, apiKey, authToken);
 
+      this.appDistributionReleaseInternal = retrievedLatestRelease;
+
       long currentInstalledVersionCode =
           getInstalledAppVersionCode(firebaseApp.getApplicationContext());
 
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -501,7 +501,12 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
                 firebaseApp.getApplicationContext()));
   }
 
-  private void redirectToPlayForAabUpdate(String downloadUrl) {
+  private void redirectToPlayForAabUpdate(String downloadUrl)
+      throws FirebaseAppDistributionException {
+    if (downloadUrl == null) {
+      throw new FirebaseAppDistributionException(
+          "Download URL not found.", FirebaseAppDistributionException.Status.NETWORK_FAILURE);
+    }
     Intent updateIntent = new Intent(Intent.ACTION_VIEW);
     Uri uri = Uri.parse(downloadUrl);
     updateIntent.setData(uri);
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -344,7 +344,6 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
     } else {
       // If we can't launch a chrome view try to launch anything that can handle a URL.
       Intent browserIntent = new Intent(Intent.ACTION_VIEW, uri);
-      ResolveInfo info = currentActivity.getPackageManager().resolveActivity(browserIntent, 0);
       browserIntent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
       browserIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
       currentActivity.startActivity(browserIntent);
@@ -422,9 +421,6 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
           firebaseAppDistributionTesterApiClient.fetchLatestRelease(fid, appId, apiKey, authToken);
 
       this.appDistributionReleaseInternal = retrievedLatestRelease;
-
-      long currentInstalledVersionCode =
-          getInstalledAppVersionCode(firebaseApp.getApplicationContext());
 
       if (isNewerBuildVersion(retrievedLatestRelease)
           && !isInstalledRelease(retrievedLatestRelease)) {
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -236,10 +236,11 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
     updateAppCancellationSource = new CancellationTokenSource();
     updateAppTaskCompletionSource =
         new TaskCompletionSource<>(updateAppCancellationSource.getToken());
+    Context context = firebaseApp.getApplicationContext();
 
     if (appDistributionReleaseInternal == null) {
       throw new FirebaseAppDistributionException(
-          "No new release available. Try calling checkForUpdate",
+          context.getString(R.string.no_update_available),
           FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE);
     }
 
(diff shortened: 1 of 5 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateTaskImpl.java
```diff
@@ -0,0 +1,123 @@
...
+class UpdateTaskImpl extends UpdateTask {
+
+  @NonNull private final Task<UpdateState> task;
+  @Nullable private OnProgressListener listener = null;
+
+  public UpdateTaskImpl(@NonNull Task<UpdateState> task) {
+    this.task = task;
+  }
+
+  public void updateProgress(@NonNull UpdateState updateState) {
```
rachaprince, line 35: should this be public? Do we want customers to be calling this method from their code? Perhaps it should be package private?

rachaprince (commented):
Looks good so far!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateTaskImpl.java
andrewbibiloni: Yep I think package private makes more sense

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateTaskImpl.java
```diff
@@ -0,0 +1,123 @@
...
+import com.google.android.gms.tasks.Task;
+import java.util.concurrent.Executor;
+
+/** Implementation of UpdateTask, the return type of updateApp. */
+class UpdateTaskImpl extends UpdateTask {
+
+  @NonNull private final Task<UpdateState> task;
+  @Nullable private OnProgressListener listener = null;
+
+  public UpdateTaskImpl(@NonNull Task<UpdateState> task) {
```
rachaprince, line 31: I think this constructor should also be package private

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -473,4 +500,14 @@ private boolean isInstalledRelease(AppDistributionReleaseInternal latestRelease)
             ReleaseIdentificationUtils.extractInternalAppSharingArtifactId(
                 firebaseApp.getApplicationContext()));
   }
+
+  private void redirectToPlayForAabUpdate(String downloadUrl) {
+    Intent updateIntent = new Intent(Intent.ACTION_VIEW);
+    Uri uri = Uri.parse(downloadUrl);
```
rachaprince, line 506: Since downloadUrl is nullable, we should probably add a check to make sure this string is properly formatted (or at least, not null) before redirecting to play

rachaprince (approved):
Nice work! I only had two small comments

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -221,8 +226,31 @@ public void onFailure(@NonNull Exception e) {
...
+      updateAppCancellationSource.cancel();
+    }
+
+    updateAppCancellationSource = new CancellationTokenSource();
+    updateAppTaskCompletionSource =
+        new TaskCompletionSource<>(updateAppCancellationSource.getToken());
+    Context context = firebaseApp.getApplicationContext();
+
+    if (appDistributionReleaseInternal == null) {
+      throw new FirebaseAppDistributionException(
```
pranavrajgopal, line 242: Since context is only used here you can inline `firebaseApp.getApplicationContext().getString` here

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -70,6 +70,11 @@
   private CancellationTokenSource checkForUpdateCancellationSource;
   private final Executor checkForUpdateExecutor;
 
+  private TaskCompletionSource<UpdateState> updateAppTaskCompletionSource = null;
+  private CancellationTokenSource updateAppCancellationSource;
+
```
pranavrajgopal, line 75: I would rename this to be more meaningful. Maybe `cachedLatestRelease` or `currentLatestRelease`?

pranavrajgopal: Open to other names too :) but appDistributionReleaseInternal is kinda ambiguous

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
andrewbibiloni: I thought that maybe I would need it again when implementing the APK part so I left it there for now.  is it better to inline it now and change it back if need be, or inline it later if it ends up not being needed again?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -473,4 +500,19 @@ private boolean isInstalledRelease(AppDistributionReleaseInternal latestRelease)
...
+    if (downloadUrl == null) {
+      throw new FirebaseAppDistributionException(
+          "Download URL not found.", FirebaseAppDistributionException.Status.NETWORK_FAILURE);
+    }
+    Intent updateIntent = new Intent(Intent.ACTION_VIEW);
+    Uri uri = Uri.parse(downloadUrl);
+    updateIntent.setData(uri);
+    updateIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+    currentActivity.startActivity(updateIntent);
+    updateAppTaskCompletionSource.setResult(
```
pranavrajgopal, line 515: We should probably use `AutoValue` for updateState (similar to AppDistributionRelease)

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -221,8 +226,31 @@ public void onFailure(@NonNull Exception e) {
...
+          context.getString(R.string.no_update_available),
+          FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE);
+    }
+
+    if (appDistributionReleaseInternal.getBinaryType() == BinaryType.AAB) {
+      redirectToPlayForAabUpdate(appDistributionReleaseInternal.getDownloadUrl());
+    } else {
+      throw new UnsupportedOperationException("Not yet implemented.");
+    }
+
```
pranavrajgopal, line 252: I might have confused myself a little on the UpdateTask method and how that's going to work for APKs. lets sync on this at some point.

pranavrajgopal: Also sidenote, a customer could call `addProgressListener` on the UpdateTask and they should get `UpdateStatus.REDIRECTED_TO_PLAY` as the response. This might require some rework to store the UpdateTaskImpl as a class variable instead of `updateAppTaskCompletionSource` and call the updateProgress method

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
pranavrajgopal: Ahh okay. Its not a big deal. You can leave this as is.

pranavrajgopal (approved):
Nice work!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -418,6 +446,10 @@ AppDistributionRelease getLatestReleaseFromClient(
     }
   }
```
pranavrajgopal, line 448: nit: Might be worth setting the @VisibileForTesting annotation on this method.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -223,8 +230,33 @@ public void onFailure(@NonNull Exception e) {
...
+        && !updateAppTaskCompletionSource.getTask().isComplete()) {
+      updateAppCancellationSource.cancel();
+    }
+
+    updateAppCancellationSource = new CancellationTokenSource();
+    updateAppTaskCompletionSource =
+        new TaskCompletionSource<>(updateAppCancellationSource.getToken());
+    Context context = firebaseApp.getApplicationContext();
+    this.updateTask = new UpdateTaskImpl(updateAppTaskCompletionSource.getTask());
+
```
pranavrajgopal, line 245: I just realized that this is throwing an exception. Is there a particular reason we are throwing an exception and not setting this as a task failure for UpdateTask? This feels inconsistent with what we have been doing with other methods.

PR #2843, opened 2021-07-19
Add schema for field indexing

schmidt-sebastian:
See go/firestore-composite-index

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
@@ -176,7 +176,7 @@ class SQLiteSchema {
 
     if (fromVersion < INDEXING_SUPPORT_VERSION && toVersion >= INDEXING_SUPPORT_VERSION) {
       Preconditions.checkState(Persistence.INDEXING_SUPPORT_ENABLED);
-      createLocalDocumentsCollectionIndex();
+      createFieldIndex();
     }
   }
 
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -336,24 +336,32 @@ private void createV1RemoteDocumentCache() {
...
-  private void createLocalDocumentsCollectionIndex() {
+  private void createFieldIndex() {
+    // Create two tables to support document indexing. `index_configuration` holds the
+    // configuration for all currently active indices. All entries apply for all users and it is
+    // not possible to only enable indices for a subset of users. `field_index` holds the index
+    // values. If there are pending mutations that affect an indexed field, an additional index
+    // entry is created per mutated field.
     ifTablesDontExist(
-        new String[] {"collection_index"},
+        new String[] {"index_configuration", "field_index"},
```
wu-hui, line 346: Looks like other tables use plural.

wu-hui: And what about `field_index_entries`, too long?

wu-hui (commented):
Please add a simple test to verify tables are created.

Also, not related to code review..when we are sure the schemas are not going to change, we can just release it to the users, is that the plan?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -336,24 +336,32 @@ private void createV1RemoteDocumentCache() {
         });
   }
 
-  // TODO(indexing): Put the schema version in this method name.
-  private void createLocalDocumentsCollectionIndex() {
+  private void createFieldIndex() {
+    // Create two tables to support document indexing. `index_configuration` holds the
+    // configuration for all currently active indices. All entries apply for all users and it is
+    // not possible to only enable indices for a subset of users. `field_index` holds the index
+    // values. If there are pending mutations that affect an indexed field, an additional index
```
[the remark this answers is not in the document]
schmidt-sebastian: I think this should just be "an entry", can you double check?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: I don't have a strong opinion. I made this "index_entries", which is shorter but still contains your suggestion. It also uses the same prefix as "index_configuration", which can be nice for grouping.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -176,7 +176,7 @@ void runMigrations(int fromVersion, int toVersion) {
 
     if (fromVersion < INDEXING_SUPPORT_VERSION && toVersion >= INDEXING_SUPPORT_VERSION) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -336,24 +336,32 @@ private void createV1RemoteDocumentCache() {
         });
   }
 
-  // TODO(indexing): Put the schema version in this method name.
-  private void createLocalDocumentsCollectionIndex() {
+  private void createFieldIndex() {
+    // Create two tables to support document indexing. `index_configuration` holds the
```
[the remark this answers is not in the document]
schmidt-sebastian: Moved to method comment, expanded it a bit and improved formatting.

schmidt-sebastian (commented):
Thanks for the review. I also added a very basic test.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -336,24 +336,37 @@ private void createV1RemoteDocumentCache() {
...
-  // TODO(indexing): Put the schema version in this method name.
-  private void createLocalDocumentsCollectionIndex() {
+  /**
+   * Creates the necessary tables to support document indexing.
+   *
+   * <p>The `index_configuration` table holds the configuration for all indices. Entries in this
+   * table apply for all users. It is not possible to only enable indices for a subset of users.
+   *
+   * <p>The `index_entries` table holds the index values themselves. An index value is created for
+   * every indexed field. If there are pending mutations that affect an indexed field, an additional
```
wu-hui, line 346: This is not technically true is it? This suggests if there is a composite index with 2 fields, there will be two entries in this table.

How about `An entry is created for each value of the indexed field(s)...`

wu-hui (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: You are correct. I changed it to say " An index value is created for each field combination that matches a configured index."

Thanks for catching.

PR #2844, opened 2021-07-20
FAD Implement updateToLatestRelease MVP 

andrewbibiloni:
- Implement updateToLatestRelease 
- Add method to set exception for updateApp

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/Constants.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/Constants.java
@@ -32,7 +32,7 @@ class Constants {
 
     public static final String TIMEOUT_ERROR = "Failed to fetch releases due to timeout";
 
-    public static final String UPDATE_CANCELLED = "Update Cancelled.";
+    public static final String UPDATE_CANCELLED = "Update cancelled";
 
     public static final String UNKNOWN_ERROR = "Unknown Error";
   }
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/Constants.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/Constants.java
@@ -32,6 +32,8 @@ class Constants {
 
     public static final String TIMEOUT_ERROR = "Failed to fetch releases due to timeout";
 
+    public static final String UPDATE_CANCELLED = "Update Cancelled.";
+
     public static final String UNKNOWN_ERROR = "Unknown Error";
   }
 }
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -140,12 +140,39 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
 
     // todo: when persistence is implemented and user already signed in, skip signInTester
     signInTester()
-        .addOnSuccessListener(unused -> showUpdateDialogIfNewReleaseAvailable())
+        .addOnSuccessListener(
+            unused -> {
+              checkForUpdate()
+                  .addOnSuccessListener(
+                      latestRelease -> {
+                        if (latestRelease != null) {
+                          showUpdateAlertDialog(latestRelease);
+                        } else {
+                          updateToLatestReleaseTaskCompletionSource.setResult(null);
+                        }
+                      })
+                  .addOnFailureListener(
+                      e -> {
+                        if (e instanceof FirebaseAppDistributionException) {
+                          setUpdateToLatestReleaseTaskCompletionError(
+                              (FirebaseAppDistributionException) e);
+                        } else {
+                          setUpdateToLatestReleaseTaskCompletionError(
+                              new FirebaseAppDistributionException(
+                                  Constants.ErrorMessages.NETWORK_ERROR, NETWORK_FAILURE));
+                        }
+                      });
+            })
         .addOnFailureListener(
-            e ->
+            e -> {
+              if (e instanceof FirebaseAppDistributionException) {
+                setUpdateToLatestReleaseTaskCompletionError((FirebaseAppDistributionException) e);
+              } else {
                 setUpdateToLatestReleaseTaskCompletionError(
                     new FirebaseAppDistributionException(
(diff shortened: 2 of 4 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/Constants.java
```diff
@@ -32,6 +32,8 @@
 
     public static final String TIMEOUT_ERROR = "Failed to fetch releases due to timeout";
 
+    public static final String UPDATE_CANCELLED = "Update Cancelled.";
```
rachaprince, line 35: Nit: "Update cancelled." to match  the casing of the other strings

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -535,4 +553,65 @@ private void redirectToPlayForAabUpdate(String downloadUrl)
...
+              if (latestRelease != null) {
+                showUpdateAlertDialog(latestRelease);
+              } else {
+                updateToLatestReleaseTaskCompletionSource.setResult(null);
+              }
+            })
+        .addOnFailureListener(
+            e ->
+                setUpdateToLatestReleaseTaskCompletionError(
+                    new FirebaseAppDistributionException(
```
rachaprince, line 570: If checkForUpdate fails with a FirebaseAppDistributionException, can we set the exception on updateToLatestRelease to the same exception, instead of overwriting the exception?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -535,4 +553,65 @@ private void redirectToPlayForAabUpdate(String downloadUrl)
...
+        context.getString(R.string.update_yes_button),
+        (dialogInterface, i) -> {
+          try {
+            updateApp()
+                .addOnSuccessListener(
+                    updateState ->
+                        updateToLatestReleaseTaskCompletionSource.setResult(latestRelease))
+                .addOnFailureListener(
+                    e ->
+                        setUpdateToLatestReleaseTaskCompletionError(
```
rachaprince, line 595: Same question here - if this error is a FirebaseAppDistributionException from updateApp, can we propagate that error instead of overwritting it?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -124,26 +129,25 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
...
-            })
+        .addOnSuccessListener(unused -> showUpdateDialogIfNewReleaseAvailable())
         .addOnFailureListener(
-            new OnFailureListener() {
-              @Override
-              public void onFailure(@NonNull Exception e) {
-                taskCompletionSource.setException(e);
-              }
-            });
+            e ->
```
pranavrajgopal, line 145: If you reset the exception here, the underlying AppDistributionException gets hidden. I would check if `e` is an instance of FirebaseAppDistribution and set the error as `e`. And maybe by default you can set the error to be `                    new FirebaseAppDistributionException(
                        Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE)`

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
pranavrajgopal: oops added the same comment above :) Probably worth doing the same thing for sign in.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -124,26 +129,25 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
...
 
+    // todo: when persistence is implemented and user already signed in, skip signInTester
     signInTester()
-        .addOnSuccessListener(
-            new OnSuccessListener<Void>() {
-              @Override
-              public void onSuccess(Void unused) {
-                taskCompletionSource.setResult(null);
-              }
-            })
```
pranavrajgopal, line 137: I think there might be a lot of method nesting here. I feel we should move up the checkForUpdate call in here and maybe add the alert dialog in a different method.

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -363,17 +376,82 @@ public void updateAppTask_whenNoReleaseAvailable_throwsError() throws Exception
...
+        firebaseAppDistribution.updateToLatestRelease();
+    Task<AppDistributionRelease> updateToLatestReleaseTask2 =
+        firebaseAppDistribution.updateToLatestRelease();
+
+    assertTrue(updateToLatestReleaseTask1.isCanceled());
+    assertFalse(updateToLatestReleaseTask2.isComplete());
+  }
+
+  @Test
+  public void updateToLatestRelease_whenNewReleaseAvailable_showsUpdateDialog() throws Exception {
```
rachaprince, line 406: nit: change this test name to whenNewAabReleaseAvailable

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -363,17 +376,82 @@ public void updateAppTask_whenNoReleaseAvailable_throwsError() throws Exception
...
+    assertTrue(signInDialog.isShowing());
+    signInDialog.getButton(AlertDialog.BUTTON_POSITIVE).performClick();
+
+    firebaseAppDistribution.onActivityCreated(mockSignInResultActivity, mockBundle);
+    firebaseAppDistribution.onActivityResumed(activity);
+
+    // update flow
+    assertTrue(ShadowAlertDialog.getLatestDialog() instanceof AlertDialog);
+    AlertDialog latestDialog = (AlertDialog) ShadowAlertDialog.getLatestDialog();
+    assertEquals(latestDialog, signInDialog);
```
rachaprince, line 453: another option would be assertFalse(alertDialog.isShowing())

rachaprince (approved):

Issue #2840, opened 2021-07-18
ConcurrentModificationException in MetaDataStore.keysDataToJson for Crashlytics


amal:
### Environment

  * Android Studio version: Arctic Fox | 2020.3.1 Beta 5
  * Firebase Component: **Crashlytics**
  * Component version: **18.1.0**

### Problem

Normal usage of Crashlytics **v18.1.0** leads to error in logcat:

```stacktrace
E/FirebaseCrashlytics: Error serializing key/value metadata.
    java.util.ConcurrentModificationException
        at java.util.HashMap$HashIterator.nextNode(HashMap.java:1441)
        at java.util.HashMap$EntryIterator.next(HashMap.java:1475)
        at java.util.HashMap$EntryIterator.next(HashMap.java:1473)
        at java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet$1.next(Collections.java:1708)
        at java.util.Collections$UnmodifiableMap$UnmodifiableEntrySet$1.next(Collections.java:1701)
        at org.json.JSONObject.<init>(JSONObject.java:138)
        at com.google.firebase.crashlytics.internal.common.MetaDataStore.keysDataToJson(MetaDataStore.java:157)
        at com.google.firebase.crashlytics.internal.common.MetaDataStore.writeKeyData(MetaDataStore.java:91)
        at com.google.firebase.crashlytics.internal.common.CrashlyticsController$8.call(CrashlyticsController.java:484)
        at com.google.firebase.crashlytics.internal.common.CrashlyticsController$8.call(CrashlyticsController.java:480)
        at com.google.firebase.crashlytics.internal.common.CrashlyticsBackgroundWorker$3.then(CrashlyticsBackgroundWorker.java:105)
        at com.google.android.gms.tasks.zzc.run(com.google.android.gms:play-services-tasks@@17.2.1:3)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at com.google.firebase.crashlytics.internal.common.ExecutorUtils$1$1.onRun(ExecutorUtils.java:64)
        at com.google.firebase.crashlytics.internal.common.BackgroundPriorityRunnable.run(BackgroundPriorityRunnable.java:27)
        at java.lang.Thread.run(Thread.java:920)
```

#### Steps to reproduce:

[...]

PR #2851, opened 2021-07-23
Fixed ConcurrentModificationException from setCustomKeys

mrichards:
Exception would be caught & logged when making rapid calls to setCustomKeys.

Updated KeysMap#getKeys() to return a defensive copy of keys, and
reduced overhead inside the synchronized blocks when updating keys.

Internal bug b/194422907, Github issue #2840


samedson (approved):
Shipit!

PR #2856, opened 2021-07-26
Add IndexConfiguration

schmidt-sebastian:
This PR adds an API to configure indices and the plumbing that is needed to store the indices in SQLLite. The API itself is for internal use only and will only be used during development. The final API to enable indices might look very different as it has not yet gone through review.

The PR also adds some very basic tests. Since it is not yet possible to determine whether an index has been enabled, the tests are somewhat limited in scope.


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
```diff
@@ -287,4 +289,31 @@ public BundledQuery decodeBundledQuery(com.google.firestore.bundle.BundledQuery
 
     return new BundledQuery(target, limitType);
   }
+
+  public Index encodeFieldIndex(FieldIndex fieldIndex) {
+    Index.Builder index = Index.newBuilder();
+    // The Mobile SDKs treat all indices as collectiong group indices.
```
wu-hui, line 295: Can you explain the reason for this?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -65,4 +70,33 @@ public void addToCollectionParentIndex(ResourcePath collectionPath) {
             });
     return parentPaths;
   }
+
+  @Override
+  public void addFieldIndex(FieldIndex index) {
+    int currentMax =
+        db.query("SELECT MAX(index_id) FROM index_configuration")
+            .firstValue(
+                new Function<Cursor, Integer>() {
```
wu-hui, line 79: Can this be just a lambda?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndex.java
```diff
@@ -0,0 +1,132 @@
...
+      if (o == null || getClass() != o.getClass()) return false;
+
+      Segment segment = (Segment) o;
+      if (!fieldPath.equals(segment.fieldPath)) return false;
+      return kind == segment.kind;
+    }
+
+    @Override
+    public int hashCode() {
+      int result = fieldPath.hashCode();
```
wu-hui, line 78: `kind` should also be part of the hash here?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndex.java
```diff
@@ -0,0 +1,132 @@
...
+ * <p>Every index is associated with a collection. The definition contains a list of fields and the
+ * indexes sorting order (which can be {@link Segment.Kind#ASCENDING}, {@link
+ * Segment.Kind#DESCENDING} or {@link Segment.Kind#CONTAINS} for ArrayContains/ArrayContainsAn
+ * queries.
+ *
+ * <p>Unlike the backend, the SDK does not differentiate between collection or collection
+ * group-scoped indices. Every index can be used for both single collection and collection group
+ * queries.
+ */
+public class FieldIndex implements Iterable<FieldIndex.Segment> {
```
wu-hui, line 34: Make it final if we don't see it being inherited, same for `Segment`.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndex.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @NonNull
+  @Override
+  public Iterator<Segment> iterator() {
+    return segments.iterator();
+  }
+
+  /** Returns a new field index with additional index segment. */
+  public FieldIndex withComponent(FieldPath fieldPath, Segment.Kind kind) {
```
wu-hui, line 109: `Component` seems to be a term out of nowhere, how about `NewField`?

wu-hui (changes requested):

wu-hui (approved):

PR #2857, opened 2021-07-26
FAD Add SignIn persistence to SDK 

andrewbibiloni:
- Add sign in persistence through SignInStorage class using SharedPreferences 
- Implement isTesterSignedIn and signOutTester
- Fix tests affected by persistence 
- Add signIn checks in updateToLatestRelease, updateApp, and checkForUpdate 
- Add unit tests

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -31,6 +31,7 @@ import com.google.android.gms.common.internal.Preconditions;
 import com.google.android.gms.tasks.CancellationTokenSource;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.appdistribution.internal.AppDistributionReleaseInternal;
 import com.google.firebase.installations.FirebaseInstallationsApi;
@@ -53,15 +54,19 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
       null;
   private CancellationTokenSource updateToLatestReleaseCancellationSource;
 
+  private SignInStorage signInStorage;
+
   /** Constructor for FirebaseAppDistribution */
   @VisibleForTesting
   FirebaseAppDistribution(
       @NonNull FirebaseApp firebaseApp,
       @NonNull TesterSignInClient testerSignInClient,
-      @NonNull CheckForUpdateClient checkForUpdateClient) {
+      @NonNull CheckForUpdateClient checkForUpdateClient,
+      @NonNull SignInStorage signInStorage) {
     this.firebaseApp = firebaseApp;
     this.testerSignInClient = testerSignInClient;
     this.checkForUpdateClient = checkForUpdateClient;
+    this.signInStorage = signInStorage;
   }
 
   /** Constructor for FirebaseAppDistribution */
@@ -72,7 +77,8 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
         firebaseApp,
         new TesterSignInClient(firebaseApp, firebaseInstallationsApi),
         new CheckForUpdateClient(
-            firebaseApp, new FirebaseAppDistributionTesterApiClient(), firebaseInstallationsApi));
+            firebaseApp, new FirebaseAppDistributionTesterApiClient(), firebaseInstallationsApi),
+        new SignInStorage(firebaseApp.getApplicationContext()));
   }
 
   /** @return a FirebaseAppDistribution instance */
@@ -113,32 +119,21 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
     updateToLatestReleaseTaskCompletionSource =
         new TaskCompletionSource<>(updateToLatestReleaseCancellationSource.getToken());
 
-    // todo: when persistence is implemented and user already signed in, skip signInTester
-    signInTester()
-        .addOnSuccessListener(
-            unused -> {
-              checkForUpdate()
-                  .addOnSuccessListener(
-                      latestRelease -> {
-                        if (latestRelease != null) {
-                          showUpdateAlertDialog(latestRelease);
-                        } else {
-                          updateToLatestReleaseTaskCompletionSource.setResult(null);
-                        }
-                      })
-                  .addOnFailureListener(
-                      e ->
-                          setUpdateToLatestReleaseErrorWithDefault(
-                              e,
-                              new FirebaseAppDistributionException(
-                                  Constants.ErrorMessages.NETWORK_ERROR, NETWORK_FAILURE)));
-            })
-        .addOnFailureListener(
-            e ->
-                setUpdateToLatestReleaseErrorWithDefault(
-                    e,
-                    new FirebaseAppDistributionException(
-                        Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE)));
+    if (isTesterSignedIn()) {
+      checkForNewReleaseAndUpdateIfAvailable();
+    } else {
+      signInTester()
+          .addOnSuccessListener(
+              unused -> {
+                checkForNewReleaseAndUpdateIfAvailable();
+              })
+          .addOnFailureListener(
+              e ->
+                  setUpdateToLatestReleaseErrorWithDefault(
+                      e,
+                      new FirebaseAppDistributionException(
+                          Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE)));
+    }
 
     return updateToLatestReleaseTaskCompletionSource.getTask();
   }
@@ -156,17 +151,19 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
    */
   @NonNull
   public Task<AppDistributionRelease> checkForUpdate() {
-    return this.checkForUpdateClient
-        .checkForUpdate()
-        .continueWith(
-            task -> {
-              // Calling task.getResult() on a failed task will cause the Continuation to fail
-              // with the original exception. See:
-              // https://developers.google.com/android/reference/com/google/android/gms/tasks/Continuation
-              AppDistributionReleaseInternal latestRelease = task.getResult();
-              setCachedLatestRelease(latestRelease);
-              return convertToAppDistributionRelease(latestRelease);
-            });
+    if (isTesterSignedIn()) {
+      return getCheckForUpdateTask();
+    } else {
+      return signInTester()
+          .continueWithTask(
+              task -> {
+                if (task.isSuccessful()) {
+                  return getCheckForUpdateTask();
+                } else {
+                  return Tasks.forException(task.getException());
+                }
+              });
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -248,7 +252,6 @@ public void updateToLatestRelease_whenNewAabReleaseAvailable_showsUpdateDialog()
     firebaseAppDistribution.onActivityResumed(activity);
 
     // Update flow
-    verify(mockTesterSignInClient, times(1)).signInTester(activity);
```
andrewbibiloni, line 251: Removed this because with new signin checks, this will either be 0 if already signed in and 2 if not, because it will be triggered for both the updateToLatestRelease check and the checkForUpdate that is called within updateToLatestRelease in the context of testing.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/SignInStorage.java
```diff
@@ -0,0 +1,40 @@
...
+  private static final String SIGNIN_PREFERENCES_NAME = "FirebaseAppDistributionSignInStorage";
+  private static final String SIGNIN_TAG = "firebase_app_distro_signin";
+
+  private final SharedPreferences signInSharedPreferences;
+
+  SignInStorage(Context applicationContext) {
+    this.signInSharedPreferences =
+        applicationContext.getSharedPreferences(SIGNIN_PREFERENCES_NAME, Context.MODE_PRIVATE);
+  }
+
```
pranavrajgopal, line 32: nit: This is very minor but `signInStorage.store` feels a bit redundant. Maybe rename this as `setSignInStatus`?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -180,44 +177,51 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
...
+      return this.updateTask;
     }
 
-    return this.updateTask;
+    // if user is not signedIn
+    return new UpdateTaskImpl(
+        Tasks.forException(
+            new FirebaseAppDistributionException(
+                Constants.ErrorMessages.NOT_FOUND_ERROR,
+                FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE)));
```
rachaprince, line 213: I know I suggested UPDATE NOT AVAILABLE earlier, but maybe this should be an authentication error instead

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -180,44 +177,51 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
...
-    } else {
-      if (cachedLatestRelease.getBinaryType() == BinaryType.AAB) {
-        redirectToPlayForAabUpdate(cachedLatestRelease.getDownloadUrl());
+      updateAppCancellationSource = new CancellationTokenSource();
+      updateAppTaskCompletionSource =
+          new TaskCompletionSource<>(updateAppCancellationSource.getToken());
+      Context context = firebaseApp.getApplicationContext();
+      this.updateTask = new UpdateTaskImpl(updateAppTaskCompletionSource.getTask());
+
+      if (cachedLatestRelease == null) {
```
rachaprince, line 192: Can we move this check earlier in the function? Then we can return early if the cachedLatestRelease is null, create a UpdateTaskImpl error with Tasks.forException, and save the overhead of creating a completion sources for a task that will fail

rachaprince: If you move the sign in check early, the flow could be:
 if (!isTesterSignedIn(){
// ...
}
if (cachedLatestRelease == null) { 
// ...
}
// else tester is signed in and cachedLatestRelease is non-null
....

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -180,44 +178,49 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
...
     if (cachedLatestRelease == null) {
-      setUpdateAppTaskCompletionError(
-          new FirebaseAppDistributionException(
-              context.getString(R.string.no_update_available),
-              FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE));
+      return new UpdateTaskImpl(
+          Tasks.forException(
+              new FirebaseAppDistributionException(
+                  Constants.ErrorMessages.NOT_FOUND_ERROR, UPDATE_NOT_AVAILABLE)));
     } else {
```
rachaprince, line 192: nit: because the two if statements above contain return statements, you don't actually need this else

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -114,10 +116,13 @@ public void setup() throws Exception {
                 .setApiKey(TEST_API_KEY)
                 .build());
 
+    // this is spy instead of mock to be able to test isTesterSingedIn functionality
+    signInStorage = spy(new SignInStorage(firebaseApp.getApplicationContext()));
```
rachaprince, line 120: I'm surprised the mock doesn't work, since we are passing in the signInStorage class into the constructor. What happens when you mock this class?

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -205,19 +210,32 @@ public void checkForUpdate_whenCheckForUpdateSucceeds_returnsRelease() throws Ex
     assertEquals(TEST_RELEASE_NEWER_AAB_INTERNAL, firebaseAppDistribution.getCachedLatestRelease());
   }
 
+  @Test
+  public void checkForUpdate_whenSignedIn_doesNotCallSignInTester() {
+    Mockito.doReturn(true).when(signInStorage).getSignInStatus();
+    firebaseAppDistribution.onActivityResumed(activity);
```
rachaprince, line 216: this line is duplicated below

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
rachaprince: Hmmm. We should be able to reset the mock to 0 (if that doesn't happen automatically) And it should still be called at most one time (checkforupdate should not call it if the tester is already signed in)

It's possible to have the mock storage client return false the first time and then true the second. Let me know if you'd like to discuss in person :)

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -233,9 +251,23 @@ public void updateAppTask_whenAabReleaseAvailable_redirectsToPlay() throws Excep
...
+    assertTrue(task.getException() instanceof FirebaseAppDistributionException);
+    FirebaseAppDistributionException e = (FirebaseAppDistributionException) task.getException();
+    assertEquals(Constants.ErrorMessages.AUTHENTICATION_ERROR, e.getMessage());
+    assertEquals(AUTHENTICATION_FAILURE, e.getErrorCode());
+  }
+
   @Test
   public void updateToLatestRelease_whenNewAabReleaseAvailable_showsUpdateDialog()
       throws Exception {
+
```
rachaprince, line 270: let's specify a return value for the istesterSignedIn method so it's clear what the expected signIn behavior is

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -304,6 +335,7 @@ public void updateToLatestRelease_whenSignInCancelled_checkForUpdateNotCalled()
 
   @Test
   public void updateToLatestRelease_whenSignInFailed_checkForUpdateNotCalled() {
+    Mockito.doReturn(false).when(signInStorage).getSignInStatus();
```
rachaprince, line 338: nit: you can make this a static import so you don't need the Mockito prefix in front of "doReturn"

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -340,5 +372,30 @@ public void updateToLatestRelease_whenCheckForUpdateFails_updateAppNotCalled() t
...
+    firebaseAppDistribution.onActivityResumed(activity);
+    when(mockCheckForUpdateClient.checkForUpdate())
+        .thenReturn(Tasks.forResult(TEST_RELEASE_NEWER_AAB_INTERNAL));
+    Mockito.doReturn(true).when(signInStorage).getSignInStatus();
+    firebaseAppDistribution.updateToLatestRelease();
+    verify(mockTesterSignInClient, never()).signInTester(any());
+  }
+
+  @Test
+  public void isTesterSignedIn_afterSuccessfulSignIn_returnsTrue() {
```
rachaprince, line 386: Are these the tests that are requiring the spy instead of mock? Another way to test this is to assert that the storage method setSignInStatus is called with a true (or false) boolean.

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -117,12 +116,12 @@ public void setup() throws Exception {
                 .build());
 
     // this is spy instead of mock to be able to test isTesterSingedIn functionality
-    signInStorage = spy(new SignInStorage(firebaseApp.getApplicationContext()));
+    //    signInStorage = spy(new SignInStorage(firebaseApp.getApplicationContext()));
```
rachaprince, line 119: nit: remove comments

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -267,7 +265,7 @@ public void updateApp_whenNotSignedIn_throwsError() throws Exception {
   @Test
   public void updateToLatestRelease_whenNewAabReleaseAvailable_showsUpdateDialog()
       throws Exception {
-
+    when(mockSignInStorage.getSignInStatus()).thenReturn(false).thenReturn(true);
```
rachaprince, line 268: Can we add a comment explaining why we change the return value of the mock?

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -377,25 +377,24 @@ public void updateToLatestRelease_whenSignedIn_doesNotCallSignInTester() {
...
     when(mockCheckForUpdateClient.checkForUpdate())
         .thenReturn(Tasks.forResult(TEST_RELEASE_NEWER_AAB_INTERNAL));
-    Mockito.doReturn(true).when(signInStorage).getSignInStatus();
+    when(mockSignInStorage.getSignInStatus()).thenReturn(true);
     firebaseAppDistribution.updateToLatestRelease();
     verify(mockTesterSignInClient, never()).signInTester(any());
   }
 
   @Test
   public void isTesterSignedIn_afterSuccessfulSignIn_returnsTrue() {
```
rachaprince, line 386: After the update to these tests I think we should change the names of the tests. We're testing the behavior of the signIn method to make sure it interacts with the storage class. so maybe something like signInTester_afterSuccessfulSignIn_setsSignInStatus?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/SignInStorage.java
```diff
@@ -0,0 +1,40 @@
...
+package com.google.firebase.appdistribution;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+/** Class that handles storage for App Distribution SignIn persistence. */
+public class SignInStorage {
+
+  private static final String SIGNIN_PREFERENCES_NAME = "FirebaseAppDistributionSignInStorage";
+  private static final String SIGNIN_TAG = "firebase_app_distro_signin";
```
rachaprince, line 24: nit: we don't usually use app distro externally. Can we change this to app_distribution?

rachaprince (approved):
Looks good! Just a few small comments

PR #2871, opened 2021-07-30
FAD refactor update logic into UpdateAppClient

andrewbibiloni:
- create client class for updateApp logic 
- move relevant tests to UpdateAppClientTest
- add mockUpdateAppClient to FADtest

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
@@ -14,15 +14,12 @@
 
 package com.google.firebase.appdistribution;
 
-import static com.google.firebase.appdistribution.FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE;
-
 import android.app.Activity;
 import android.content.Intent;
 import android.net.Uri;
 import androidx.annotation.NonNull;
 import com.google.android.gms.tasks.CancellationTokenSource;
 import com.google.android.gms.tasks.TaskCompletionSource;
-import com.google.android.gms.tasks.Tasks;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.appdistribution.internal.AppDistributionReleaseInternal;
 
@@ -43,16 +40,8 @@ public class UpdateAppClient {
       @NonNull AppDistributionReleaseInternal latestRelease, @NonNull Activity currentActivity)
       throws FirebaseAppDistributionException {
 
-    if (latestRelease == null) {
-      return new UpdateTaskImpl(
-          Tasks.forException(
-              new FirebaseAppDistributionException(
-                  Constants.ErrorMessages.NOT_FOUND_ERROR, UPDATE_NOT_AVAILABLE)));
-    }
-
-    if (updateAppTaskCompletionSource != null
-        && !updateAppTaskCompletionSource.getTask().isComplete()) {
-      updateAppCancellationSource.cancel();
+    if (this.updateTask != null && !updateTask.isComplete()) {
+      return this.updateTask;
     }
 
     updateAppCancellationSource = new CancellationTokenSource();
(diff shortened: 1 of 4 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -0,0 +1,98 @@
...
+import android.app.Activity;
+import android.content.Intent;
+import android.net.Uri;
+import androidx.annotation.NonNull;
+import com.google.android.gms.tasks.CancellationTokenSource;
+import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.android.gms.tasks.Tasks;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.appdistribution.internal.AppDistributionReleaseInternal;
+
```
pranavrajgopal, line 28: nit: add a comment describing the class

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -0,0 +1,99 @@
...
+  public UpdateAppClient(@NonNull FirebaseApp firebaseApp) {
+    this.firebaseApp = firebaseApp;
+  }
+
+  @NonNull
+  public UpdateTask getUpdateTask(
+      @NonNull AppDistributionReleaseInternal latestRelease, @NonNull Activity currentActivity)
+      throws FirebaseAppDistributionException {
+
+    if (latestRelease == null) {
```
rachaprince, line 46: Should this check be in FirebaseAppDistribution? I think if we have the @NonNull annotation on the latestRelease parameter, it might make sense to verify that the cachedLatestRelease is nonnull before calling this method

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -0,0 +1,99 @@
...
+      return new UpdateTaskImpl(
+          Tasks.forException(
+              new FirebaseAppDistributionException(
+                  Constants.ErrorMessages.NOT_FOUND_ERROR, UPDATE_NOT_AVAILABLE)));
+    }
+
+    if (updateAppTaskCompletionSource != null
```
rachaprince, line 53: Instead of cancelling the task, let's just return the existing task like Vlad suggested

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -271,6 +242,8 @@ public void updateToLatestRelease_whenNewAabReleaseAvailable_showsUpdateDialog()
     AppDistributionReleaseInternal latestRelease = TEST_RELEASE_NEWER_AAB_INTERNAL.build();
     when(mockCheckForUpdateClient.checkForUpdate()).thenReturn(Tasks.forResult(latestRelease));
     firebaseAppDistribution.setCachedLatestRelease(latestRelease);
+    when(mockUpdateAppClient.getUpdateTask(latestRelease, activity))
+        .thenReturn(new UpdateTaskImpl(Tasks.forResult(any())));
```
rachaprince, line 246: I don't think this should use an argument matcher in the return of the mock. I think this should return a Tasks.ForResult(UpdatTaskImpl(Tasks.forResult(<UpdateStatus>))

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
andrewbibiloni: Yep makes more sense outside given the annotation

rachaprince (approved):

PR #2878, opened 2021-08-02
Check For Update Refactor and tests

pranavrajgopal:
* Refactoring checkForUpdate client to use onSuccessTask instead of task completion source to prevent race conditions
* Adding unit tests to validated task response

PAIR: @rachaprince

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
@@ -14,8 +14,6 @@
 
 package com.google.firebase.appdistribution;
 
-import static com.google.firebase.appdistribution.FirebaseAppDistributionException.Status.AUTHENTICATION_FAILURE;
-
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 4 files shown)


rachaprince (approved):
Nice!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
```diff
@@ -59,56 +51,59 @@
...
+                            fid,
+                            firebaseApp.getOptions().getApplicationId(),
+                            firebaseApp.getOptions().getApiKey(),
+                            installationTokenResult.getToken());
+                    return Tasks.forResult(latestRelease);
+                  } catch (FirebaseAppDistributionException ex) {
+                    return Tasks.forException(ex);
+                  }
+                })
+            .continueWithTask(
```
rachaprince, line 98: I wonder if we could just call

`.continueWith( task -> task.getResult())`

According to https://developers.google.com/android/reference/com/google/android/gms/tasks/Continuation#public-methods, to "To propagate failure from the completed Task call Task.getResult() and allow the RuntimeExecutionException to propagate. The RuntimeExecutionException will be unwrapped such that the Task returned by Task.continueWith(Continuation) or Task.continueWithTask(Continuation) fails with the original exception."

rachaprince: Since getLatestReleaseFromClient already fails with a NETWORK_ERROR, we might be able to just propagate it

rachaprince: I tried the `task.getResult()` approach locally, and the tests failed with `java.lang.Exception: Main looper has queued unexecuted runnables.` So please disregard this suggestion

PR #2881, opened 2021-08-02
FAD updateApp for APK MVP 

andrewbibiloni:
- add implementation to update and install app 
- add InstallActivity class 
- add provider_paths.xml and manifest permissions 
- download functionality follows the pattern found here: https://github.com/FirebasePrivate/appdistro-android-app/blob/master/app-distribution/app/src/main/java/dev/firebase/appdistribution/DownloadUtils.kt#L193


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -79,10 +109,185 @@ private void redirectToPlayForAabUpdate(String downloadUrl, Activity currentActi
...
+  public Task<File> getDownloadTask(String downloadUrl) {
+    if (downloadTaskCompletionSource != null
+        && !downloadTaskCompletionSource.getTask().isComplete()) {
+      downloadCancellationTokenSource.cancel();
+    }
+
+    downloadCancellationTokenSource = new CancellationTokenSource();
+    downloadTaskCompletionSource =
+        new TaskCompletionSource<>(downloadCancellationTokenSource.getToken());
+
```
rachaprince, line 153: maybe set download.pending on the progresslistener before starting the download?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -79,10 +109,185 @@ private void redirectToPlayForAabUpdate(String downloadUrl, Activity currentActi
     this.updateTask.updateProgress(updateState);
   }
 
+  public void updateApk(String downloadUrl, Activity currentActivity) {
+    getDownloadTask(downloadUrl)
+        .addOnSuccessListener(
+            file -> {
+              install(file.getPath(), currentActivity)
+                  .addOnSuccessListener(
```
rachaprince, line 117: this success listener executes on the main thread by default, unless we pass in an executor. Is this what we want?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -79,10 +109,185 @@ private void redirectToPlayForAabUpdate(String downloadUrl, Activity currentActi
...
+    if (e instanceof FirebaseAppDistributionException) {
+      setUpdateAppTaskCompletionError((FirebaseAppDistributionException) e);
+    } else {
+      setUpdateAppTaskCompletionError(defaultFirebaseException);
+    }
+  }
+
+  private void updateProgressOnMainThread(
+      long totalBytes, long downloadedBytes, UpdateStatus status) {
+    downloadHandler.post(
```
rachaprince, line 289: I am unfamiliar with this post method. Could you explain it to me?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/InstallActivity.java
```diff
@@ -0,0 +1,71 @@
...
+    Uri apkUri =
+        FileProvider.getUriForFile(
+            getApplicationContext(),
+            getApplicationContext().getPackageName() + ".provider",
+            new File(path));
+    intent.setDataAndType(apkUri, APK_MIME_TYPE);
+    intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+    this.mStartForResult.launch(intent);
+  }
+
```
pranavrajgopal, line 68: I think we should avoid doing this. This was fine for the hackweek project but we need something better here. One way to handle this would be to have a "installationComplete" method on FirebaseAppDistribution class and then get the instance of that class using `FirebaseAppDistribution.getInstance()`

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -79,10 +109,185 @@ private void redirectToPlayForAabUpdate(String downloadUrl, Activity currentActi
     this.updateTask.updateProgress(updateState);
   }
 
+  public void updateApk(String downloadUrl, Activity currentActivity) {
+    getDownloadTask(downloadUrl)
+        .addOnSuccessListener(
+            file -> {
+              install(file.getPath(), currentActivity)
```
pranavrajgopal, line 116: This is really clean!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/InstallActivity.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    this.mStartForResult =
+        registerForActivityResult(
+            new ActivityResultContracts.StartActivityForResult(),
+            activityResult -> {
+              if (activityResult.getResultCode() == Activity.RESULT_OK) {
+                installTaskCompletionSource.setResult(null);
+              } else {
```
pranavrajgopal, line 47: We should also handle Installation cancelled here

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -79,10 +109,185 @@ private void redirectToPlayForAabUpdate(String downloadUrl, Activity currentActi
...
+                                .setApkTotalBytesToDownload(0)
+                                .setApkBytesDownloaded(file.length())
+                                .setUpdateStatus(UpdateStatus.INSTALLED)
+                                .build();
+                        updateTask.updateProgress(updateState);
+                        updateAppTaskCompletionSource.setResult(updateState);
+                      })
+                  .addOnFailureListener(
+                      e ->
+                          setUpdateAppTaskCompletionError(
```
pranavrajgopal, line 130: This will mask all errors during install and always return INSTALLATION_FAILURE. Might be worth just setting it to `e`

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
```diff
@@ -79,10 +109,185 @@ private void redirectToPlayForAabUpdate(String downloadUrl, Activity currentActi
     this.updateTask.updateProgress(updateState);
   }
```
pranavrajgopal, line 111: Since updateApk logic is pretty complex, it might be worth refactoring all of that in an UpdateApkClient

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
andrewbibiloni: changed it to use e if its firebase exception and install failure as default similar to updatetolatestrelease

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateAppClient.java
andrewbibiloni: From what I've read post is basically analogous to thread.start(), using the thread that the handler is attached to which in this case is the main/UI thread. So essentially it 'posts' the runnable on the main thread's message queue

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
```diff
@@ -0,0 +1,290 @@
...
+
+  private void setUpdateAppErrorWithDefault(
+      Exception e, FirebaseAppDistributionException defaultFirebaseException) {
+    if (e instanceof FirebaseAppDistributionException) {
+      setUpdateAppTaskCompletionError((FirebaseAppDistributionException) e);
+    } else {
+      setUpdateAppTaskCompletionError(defaultFirebaseException);
+    }
+  }
+
```
pranavrajgopal, line 277: nit: might be worth renaming this to postProgress or postDownloadProgress

andrewbibiloni: called it postUpdateProgress since that's the new name for updateState we are gonna change

rachaprince (approved):
This is awesome work! Way to execute on a difficult problem 🎉

PR #2886, opened 2021-08-04
FAD refactor UpdateApp and UpdateTaskImpl

andrewbibiloni:
- remove task completion source for updateApp
- refactor UpdateTaskImpl to have task and completion source created in constructor and implement setResult and setException

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -15,7 +15,6 @@
 package com.google.firebase.appdistribution;
 
 import static com.google.firebase.appdistribution.FirebaseAppDistributionException.Status.AUTHENTICATION_FAILURE;
-import static com.google.firebase.appdistribution.FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE;
 
 import android.app.Activity;
 import android.app.AlertDialog;
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -45,6 +45,7 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
 
   private Task<Void> cachedUpdateToLatestReleaseTask;
   private Task<AppDistributionRelease> cachedCheckForUpdateTask;
+  private UpdateTaskImpl cachedUpdateAppTask;
   private AppDistributionReleaseInternal cachedLatestRelease;
   private final SignInStorage signInStorage;
 
@@ -174,22 +175,30 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
   @NonNull
   public UpdateTask updateApp() throws FirebaseAppDistributionException {
 
+    if (cachedUpdateAppTask != null && !cachedUpdateAppTask.isComplete()) {
+      return cachedUpdateAppTask;
+    }
+
+    cachedUpdateAppTask = new UpdateTaskImpl();
+
     if (!isTesterSignedIn()) {
-      return new UpdateTaskImpl(
-          Tasks.forException(
-              new FirebaseAppDistributionException(
-                  Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE)));
+      cachedUpdateAppTask.setException(
+          new FirebaseAppDistributionException(
+              Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE));
+      return cachedUpdateAppTask;
     }
 
     AppDistributionReleaseInternal cachedRelease = getCachedLatestRelease();
     if (cachedRelease == null) {
-      return new UpdateTaskImpl(
-          Tasks.forException(
-              new FirebaseAppDistributionException(
-                  Constants.ErrorMessages.NOT_FOUND_ERROR, UPDATE_NOT_AVAILABLE)));
+      cachedUpdateAppTask.setException(
+          new FirebaseAppDistributionException(
+              Constants.ErrorMessages.NOT_FOUND_ERROR, UPDATE_NOT_AVAILABLE));
+      return cachedUpdateAppTask;
     }
 
-    return this.updateAppClient.getUpdateTask(cachedLatestRelease, currentActivity);
+    this.updateAppClient.performUpdate(cachedUpdateAppTask, cachedLatestRelease, currentActivity);
+
+    return cachedUpdateAppTask;
   }
 
   /** Returns true if the App Distribution tester is signed in */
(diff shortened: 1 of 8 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -174,22 +175,30 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
   @NonNull
   public UpdateTask updateApp() throws FirebaseAppDistributionException {
```
pranavrajgopal, line 177: Feels a little like this update app logic is split between FirebaseAppDistribution class and UpdateAppClient. Why did we change the client method from `getUpdateTask` to `performUpdate`?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
```diff
@@ -47,57 +44,50 @@
...
   public void updateApk(
-      @NonNull String downloadUrl,
-      @NonNull Activity currentActivity,
       @NonNull UpdateTaskImpl updateTask,
-      @NonNull TaskCompletionSource updateAppTaskCompletionSource) {
-    this.updateTask = updateTask;
-    this.updateAppTaskCompletionSource = updateAppTaskCompletionSource;
-    downloadApk(downloadUrl)
+      @NonNull String downloadUrl,
+      @NonNull Activity currentActivity) {
```
pranavrajgopal, line 58: Similar to the comment above, feels a bit weird that updateApk is taking in a task instead of returning a task.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
andrewbibiloni: It changed to performUpdate (now updateApp) because the UpdateTask gets new'd up in the FirebaseAppDistribution class and gets passed in rather than created in the client so we can do the null/iscomplete check before the signin and null release check, and we thought that its more clear that it is the updateTask passed in that is the one that is being used for listeners/setting result. We should chat about this

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
```diff
@@ -0,0 +1,169 @@
...
+    when(mockFile.getPath()).thenReturn(TEST_URL);
+    when(mockFile.length()).thenReturn(TEST_FILE_LENGTH);
+
+    this.updateApkClient = Mockito.spy(new UpdateApkClient(firebaseApp));
+  }
+
+  @Test
+  public void updateApk_whenDownloadFails_setsNetworkError() throws Exception {
+    UpdateTaskImpl updateTask = new UpdateTaskImpl();
+    updateApkClient.updateApk(updateTask, TEST_URL, activity);
```
rachaprince, line 95: Why does this fail? I don't see any explicit failure returned by mocks in the tests

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
```diff
@@ -47,57 +45,52 @@
...
                         FirebaseAppDistributionException.Status.DOWNLOAD_FAILURE)));
   }
 
-  private @NonNull Task<File> downloadApk(@NonNull String downloadUrl) {
+  @VisibleForTesting
+  @NonNull
+  Task<File> downloadApk(@NonNull String downloadUrl, UpdateTaskImpl updateTask) {
     if (downloadTaskCompletionSource != null
         && !downloadTaskCompletionSource.getTask().isComplete()) {
       downloadCancellationTokenSource.cancel();
```
rachaprince, line 96: Why do we cancel this task instead of returning the same one?

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+  @Test
+  public void updateApk_whenInstallSuccessful_setsResult() throws Exception {
+    List<UpdateProgress> progressEvents = new ArrayList<>();
+    UpdateTaskImpl updateTask = new UpdateTaskImpl();
+    doReturn(Tasks.forResult(mockFile)).when(updateApkClient).downloadApk(TEST_URL, updateTask);
+    updateTask.addOnProgressListener(progressEvents::add);
+
+    updateApkClient.updateApk(updateTask, TEST_URL, activity);
+    latch.await(1000, TimeUnit.MILLISECONDS);
```
rachaprince, line 113: maybe consider adding a comment to explain why we need to call an await here?

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
```diff
@@ -0,0 +1,169 @@
...
+    assertTrue(updateTask.isSuccessful());
+  }
+
+  @Test
+  public void updateApk_whenInstallCancelled_setsError() throws Exception {
+    UpdateTaskImpl updateTask = new UpdateTaskImpl();
+    doReturn(Tasks.forResult(mockFile)).when(updateApkClient).downloadApk(TEST_URL, updateTask);
+
+    updateApkClient.updateApk(updateTask, TEST_URL, activity);
+    latch.await(1000, TimeUnit.MILLISECONDS);
```
rachaprince, line 132: Do we call latch.countDown() anywhere? If not, could we try Thread.sleep() for simplicity?

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
```diff
@@ -0,0 +1,169 @@
...
+    assertFalse(updateTask.isSuccessful());
+    assertTrue(updateTask.getException() instanceof FirebaseAppDistributionException);
+    FirebaseAppDistributionException e =
+        (FirebaseAppDistributionException) updateTask.getException();
+    assertEquals("Installation failed with result code: " + RESULT_FAILED, e.getMessage());
+    assertEquals(FirebaseAppDistributionException.Status.INSTALLATION_FAILURE, e.getErrorCode());
+  }
+
+  @Test
+  public void downloadTask_whenCalledMultipleTimes_getsCancelled() {
```
rachaprince, line 161: Added this question to the class under test too, but should this return the same download task instead of cancelling?

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
andrewbibiloni: Yep Thread.sleep() works and I don't countdown, going to change it.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
andrewbibiloni: I think it should be the same one to be consistent with the changes we made to the other completion source parts, I'll update the logic

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
andrewbibiloni: Fails due to the httpsConnection inputStream being null since the Test URL isn't a real download URL, going to mock the httpsconnection to make it explicit

rachaprince (approved):

PR #2861, opened 2021-07-27
Add Watch Conformance Tests

schmidt-sebastian:
This PR adds the conformance test runners. The tests are run as part of the integration test suite.

PR #2889, opened 2021-08-05
Increase connected check timeout for Firestore to 30m

schmidt-sebastian:
This change seems to be required to get the conformance tests to pass - see https://github.com/firebase/firebase-android-sdk/pull/2861

### DIFF
--- a/firebase-firestore/firebase-firestore.gradle
+++ b/firebase-firestore/firebase-firestore.gradle
@@ -20,6 +20,10 @@ plugins {
 firebaseLibrary {
     testLab.enabled = true
     publishSources = true
+    testLab {
+        enabled = true
+        timeout = '30m'
+    }
 }
 
 protobuf {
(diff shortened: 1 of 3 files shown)


firebase-firestore/firebase-firestore.gradle
```diff
@@ -20,6 +20,10 @@ plugins {
 firebaseLibrary {
     testLab.enabled = true
```
yifanyang, line 21: This line now becomes duplicate.

yifanyang (approved):

firebase-firestore/firebase-firestore.gradle
schmidt-sebastian: Fixed

yifanyang:
I see the frustration being accumulating here... Let me try to figure out why those 3 checks keep failing, and get back to you then.

schmidt-sebastian:
Yeah, I think they will never pass. FWIW, the connectedCheck is the only one that should be affected and all flakes are from RTDB. I have yet to be able to load the test result from checkChanged. I think the problem here is that it runs all tests for all products (as expected).

PR #2892, opened 2021-08-06
FAD reintroduce terminal states to UpdateStatus 

andrewbibiloni:
- Add back terminal states for UpdateStatus
- made the updateTaskImpl passed in to the updateApk method a class variable (cachedUpdateTask) inside updateApkClient since it is reused a lot and setInstallationResult would not be able to access it otherwise 
- Set terminal states in updateProgress in updateApkClient 
- Update APK tests to test for correct status being set

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
@@ -164,6 +164,7 @@ class UpdateApkClient {
       // check that file is actual JAR
       new JarFile(apkFile);
     } catch (Exception e) {
+      postUpdateProgress(totalSize, bytesDownloaded, UpdateStatus.DOWNLOAD_FAILED);
       setDownloadTaskCompletionError(
           new FirebaseAppDistributionException(
               Constants.ErrorMessages.NETWORK_ERROR,
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
@@ -46,6 +46,9 @@ class UpdateApkClient {
   private final Executor downloadExecutor;
   private TaskCompletionSource<Void> installTaskCompletionSource;
   private final FirebaseApp firebaseApp;
+  private long bytesDownloaded = 0;
+  private long totalFileSizeBytes = 0;
+  private UpdateTaskImpl cachedUpdateTask;
 
   public UpdateApkClient(@NonNull FirebaseApp firebaseApp) {
     this.downloadExecutor = Executors.newSingleThreadExecutor();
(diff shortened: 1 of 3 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
```diff
@@ -57,51 +60,51 @@ public void updateApk(
...
             file ->
                 install(file.getPath(), currentActivity)
                     .addOnSuccessListener(
                         unused -> {
-                          postUpdateProgress(
-                              updateTask, file.length(), file.length(), UpdateStatus.DOWNLOADED);
+                          postUpdateProgress(file.length(), file.length(), UpdateStatus.INSTALLED);
                           updateTask.setResult();
                         })
                     .addOnFailureListener(
```
rachaprince, line 74: If we changed this to
.OnSuccessTask(unused -> {
                          postUpdateProgress(file.length(), file.length(), UpdateStatus.INSTALLED);
                          updateTask.setResult();
})
.continueWith( task -> if (task.isComplete() && !task.isSuccessful()) {
        // check the exception status on the task.getException and call postUpdateProgress
      }
      
    Then we might be able to remove the class variables. I don't actually think we need the number of bytes downloaded for installation errors, because if we got to installation we know the download completed so we can just set both totalfilesizebytes and bytes downloaded to  file.size()

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
```diff
@@ -57,51 +60,51 @@ public void updateApk(
...
       @NonNull Activity currentActivity) {
 
-    downloadApk(downloadUrl, updateTask)
+    this.cachedUpdateTask = updateTask;
+    downloadApk(downloadUrl)
         .addOnSuccessListener(
             downloadExecutor,
             file ->
                 install(file.getPath(), currentActivity)
                     .addOnSuccessListener(
```
rachaprince, line 69: Does the onSuccess listener ever run? Or does the successful install kill the process before it executes?

pranavrajgopal: ha that's a great point. Its code that would never run. but might be worth leaving it in there anyway? maybe?

rachaprince: I guess it feels weird that we have an UpdateStatus in our API (`INSTALLED`) that we can never actually return.

Maybe we should remove that state?

andrewbibiloni: Tested it out and it kills the process, so it seems like it might be impossible to set the Install terminals state

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
rachaprince: Update: we chatted  and feel that the cachedTask is fine, but we'd prefer not to store the downloadedbytes as class variables and split logic so that sometimes we use the downloaded bytes from the method parameters while other times we use the downloadedbytes from the class variable

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
```diff
@@ -90,13 +88,16 @@ public void setup() {
 
   @Test
   public void updateApk_whenDownloadFails_setsNetworkError() throws Exception {
+    List<UpdateProgress> progressEvents = new ArrayList<>();
     UpdateTaskImpl updateTask = new UpdateTaskImpl();
+    updateTask.addOnProgressListener(progressEvents::add);
```
rachaprince, line 93: Should this test assert that the progressListener gets called with download failure?

rachaprince (approved):

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/UpdateApkClientTest.java
andrewbibiloni: forgot to remove that, it shouldn't since this failure happens before the download is set to pending

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/UpdateApkClient.java
```diff
@@ -57,51 +59,47 @@ public void updateApk(
       @NonNull String downloadUrl,
       @NonNull Activity currentActivity) {
 
-    downloadApk(downloadUrl, updateTask)
+    this.cachedUpdateTask = updateTask;
```
rachaprince, line 62: I'm realizing now that this might make testing a little more difficult. If we want to test a method other than updateApp, we'll need to make sure this is non-null

andrewbibiloni: True, maybe we could add a method like 
`@VisibleForTesting
  void setCachedUpdateTask(UpdateTaskImpl updateTask) {
    this.cachedUpdateTask = updateTask;
  }`  or just go back to propagating the taks

rachaprince: Both seem reasonable to me :) I'd say go with whatever makes most sense to you

PR #2900, opened 2021-08-11
Add Query Planner

schmidt-sebastian:
This adds a Query Planner to Firestore. This is implemented a bit differently than in the Prototype. The goal is to check all available indices for a collection against the Query Planner and determine which index gives the best match. The number of indices applicable to a given document should be fairly low, and by performing the query planning in one place, the code is isolated and can be tested independently.

The query planner supports both regular queries and merge joins. Merge joins can be used for queries that only have equalities, in which case the query planner tries to match an index against all possible field combinations. There is some inline documentation that should explain the rationale.

I also simplified the index definitions since I realized that we do not need to differentiate between Ascending and Descending indices. Instead, we now only have ORDERED indices and CONTAINS.

Most of the test cases are shamelessly stole from the backend (cs/piper///depot/google3/javatests/com/google/cloud/datastore/backend/storage/query/FirestoreQueryPlannerTest.java). They are drastically simplified though since the SDKs query planner only picks an index and not the strategy that is used for execution.


firebase-firestore/src/main/java/com/google/firebase/firestore/model/QueryPlanner.java
```diff
@@ -0,0 +1,317 @@
...
+   */
+  private FieldIndex processMergeJoin(FieldIndex index) {
+    FieldIndex bestMatch = new FieldIndex(index.getCollectionId());
+
+    if (memoizedFilterPermutations == null) {
+      // Build a set of filters that represent all possible combinations (all subsets in all
+      // orders).
+      // For a simple query with two equality clauses this means that we return four different
+      // filter combinations: filter1, filter2, filter1+filter2 and filter2+filter1
+      // Note that while this is potentially expensive, in reality the number of filters should be
```
wu-hui, line 160: Maybe backend can handle this, for SDK this still make me nervous. I know 8 filters must be rare, may never happen, but 8! == 40320. This will consume a large number of CPU cycles to compute, and memory.

Can we set a hard cap, that we would only compute full sets for the first N (7 likely) filters? This should be enough to get a good performance..and we can sleep better too.

wu-hui (changes requested):
First round pass, i think overall this looks good except the permutation part makes me nervous. I need to do a second round for readability.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/QueryPlannerTest.java
```diff
@@ -0,0 +1,439 @@
...
+            .filter(filter("a1", "==", "a"))
+            .filter(filter("a2", ">", "b"))
+            .orderBy(orderBy("a2", "asc"));
+    QueryPlanner queryPlanner = new QueryPlanner(queriesMultipleFilters.toTarget());
+
+    FieldIndex index =
+        new FieldIndex("collId")
+            .withAddedField(field("a1"), FieldIndex.Segment.Kind.ORDERED)
+            .withAddedField(field("a2"), FieldIndex.Segment.Kind.ORDERED);
+    assertEquals(index, queryPlanner.getMatchingPrefix(index));
```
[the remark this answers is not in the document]
schmidt-sebastian: This was just a test if you really understand the test helpers. You passed. :)

Fixed here and elsewhere.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/QueryPlannerTest.java
```diff
@@ -0,0 +1,439 @@
...
+            .orderBy(orderBy("bar"))
+            .orderBy(orderBy("__name__", "desc"));
+    validateServesTarget(
+        q,
+        "foo",
+        FieldIndex.Segment.Kind.ORDERED,
+        "bar",
+        FieldIndex.Segment.Kind.ORDERED,
+        "__name__",
+        FieldIndex.Segment.Kind.ORDERED);
```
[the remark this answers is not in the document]
schmidt-sebastian: It actually does serve the index since the orderBys are only used to filter documents, not to enforce order.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/QueryPlannerTest.java
```diff
@@ -0,0 +1,439 @@
...
+    FieldIndex expectedIndex = getSegments(field, kind, fieldsAndKind);
+    QueryPlanner queryPlanner = new QueryPlanner(query.toTarget());
+    FieldIndex actualIndex = queryPlanner.getMatchingPrefix(expectedIndex);
+    assertEquals(expectedIndex, actualIndex);
+  }
+
+  private void validateDoesNotServeTarget(
+      Query query, String field, FieldIndex.Segment.Kind kind, Object... fieldsAndKind) {
+    FieldIndex expectedIndex = getSegments(field, kind, fieldsAndKind);
+    QueryPlanner queryPlanner = new QueryPlanner(query.toTarget());
```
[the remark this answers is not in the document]
schmidt-sebastian: Target has some benefits here, as we can serve limitToLast queries with an existing index.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/QueryPlannerTest.java
```diff
@@ -0,0 +1,439 @@
...
+    FieldIndex actualIndex = queryPlanner.getMatchingPrefix(expectedIndex);
+    assertEquals(0, actualIndex.segmentCount());
+  }
+
+  private FieldIndex getSegments(
```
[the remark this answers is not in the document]
schmidt-sebastian: Sounds good.

schmidt-sebastian (commented):
Thanks for the review!

firebase-firestore/src/main/java/com/google/firebase/firestore/model/QueryPlanner.java
```diff
@@ -0,0 +1,317 @@
...
+ *             <td>a CONTAINS, b ORDERED</td>
+ *        </tr>
+ *        <tr>
+ *             <td>where("a", "array-contains", "a").orderBy("b")</td>
+ *             <td>a CONTAINS</td>
+ *        </tr>
+ *     </tbody>
+ * </table>
+ */
+public class QueryPlanner {
```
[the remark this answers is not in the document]
wu-hui: :)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndexMatcher.java
```diff
@@ -0,0 +1,155 @@
...
+import com.google.firebase.firestore.core.Target;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A light query planner for Firestore.
```
wu-hui, line 31: Needs an update here.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/FieldIndexMatcherTest.java
```diff
@@ -0,0 +1,478 @@
...
+            .orderBy(orderBy("__name__"));
+    validateServesTarget(
+        q,
+        "fff",
+        FieldIndex.Segment.Kind.ORDERED,
+        "bar",
+        FieldIndex.Segment.Kind.ORDERED,
+        "__name__",
+        FieldIndex.Segment.Kind.ORDERED);
+    // The field index can be used regardless of the order of segments since we only use the orderBy
```
wu-hui, line 331: I see. 

So `orderBy` here serves the purpose of `fieldExists`. And this is different from using no index at all, because that would include documents without these fields as well.

Then what about an index of `(fff, xxx, bar, __name__)`, this would match with `fff`, but would only include documents having `xxx`, which seems to be wrong?

What about `(bar, __name__)`, this would be useful, but we need to do postfiltering to remove documents without `fff`. Is this understanding correct?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndexMatcher.java
schmidt-sebastian: I added "light". The explanation in the section below describes the functionality better - but I think the term "query planner" is better for a high level understanding. If you think this is misleading fight back and I will change it :)

firebase-firestore/src/test/java/com/google/firebase/firestore/model/FieldIndexMatcherTest.java
schmidt-sebastian: Your understanding is correct. I think we can keep most of our logic, but must change how we handle "index subsets". Given your example, I don't think we can actually serve a documents from an index that indexes more fields than the target has. This further simplifies this PR as we can now return a simple boolean to indicate whether an Index matches a target.

Thanks for catching this flaw! I think all of this should be caught by the conformance test suite when we put everything together, but it is always better to catch things early.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndexMatcher.java
wu-hui: Sounds good.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/FieldIndexMatcherTest.java
wu-hui: Maybe I missed it, can you add a test for `(fff, xxx, bar, __name__)` to make sure this returns false?

Also, I don't think we should resolve it here, but index with a super set fields cannot serve a target seems a bit unfortunate. Can we consider indexing documents with absent fields?

schmidt-sebastian: Added one more test.

We might have to add an additional index if we decide to index documents with missing fields since we will no longer be able to guarantee `O(result set)` performance even for queries that do have their own index.

wu-hui: Yeah, we need to decide what to do here. But this PR LGTM.

wu-hui (approved):

PR #2903, opened 2021-08-12
FAD Update Notifications Manager to indicate download progress

andrewbibiloni:
- Add FirebaseAppDistributionNotificationsManager to handle notification logic and show progress
- Add flag to check if updateApp is being called from the basic configuration
- Fix and add unit tests

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -188,9 +188,7 @@ public class FirebaseAppDistribution implements Application.ActivityLifecycleCal
       return updateTask;
     }
 
-    UpdateTask updateTask =
-        this.updateAppClient.updateApp(cachedLatestRelease, showDownloadInNotificationManager);
-    return updateTask;
+    return this.updateAppClient.updateApp(cachedLatestRelease, showDownloadInNotificationManager);
   }
 
   /** Returns true if the App Distribution tester is signed in */


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -47,6 +47,7 @@
   private UpdateTaskImpl cachedUpdateAppTask;
   private AppDistributionReleaseInternal cachedLatestRelease;
   private final SignInStorage signInStorage;
+  private boolean basicConfiguration = false;
```
rachaprince, line 50: Does this need to be a class variable? I'm worried about how we always need to remember to set this back to false. It feels like it could be error prone.

Could we just overload updateApp so that internally, we call it with `showProgressInNotificationManager=true`, while the public method calls it with 

so we'd have two updateApps:

private updateApp(boolean ....) {
  // same as now
} 

public update app(){
  updateApp(false)
}

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -182,7 +183,10 @@ public synchronized UpdateTask updateApp() {
       return updateTask;
     }
 
-    return this.updateAppClient.updateApp(cachedLatestRelease, currentActivity);
+    UpdateTask updateTask =
+        this.updateAppClient.updateApp(cachedLatestRelease, currentActivity, basicConfiguration);
```
rachaprince, line 187: What are your thoughts about renaming basicConfiguration to something that more clearly indicates that it is a feature flag? My first thought is `showDownloadInNotificationManager`

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    return iconId;
+  }
+
+  /** Adaptive icons cause a crash loop in the Notifications Manager. See b/69969749. */
+  private boolean isAdaptiveIcon(int iconId) {
+    try {
+      Drawable icon = ContextCompat.getDrawable(firebaseApp.getApplicationContext(), iconId);
+      if (VERSION.SDK_INT > Build.VERSION_CODES.O && icon instanceof AdaptiveIconDrawable) {
```
rachaprince, line 125: This is my mistake, but this should probably >=. Version 0 is version 26

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
andrewbibiloni: Agreed, changed it and also added the updateApp overload

rachaprince (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -0,0 +1,136 @@
...
+import android.graphics.drawable.AdaptiveIconDrawable;
+import android.graphics.drawable.Drawable;
+import android.os.Build;
+import android.os.Build.VERSION;
+import android.util.Log;
+import androidx.annotation.VisibleForTesting;
+import androidx.core.app.NotificationCompat;
+import androidx.core.content.ContextCompat;
+import com.google.firebase.FirebaseApp;
+
```
pranavrajgopal, line 32: We might have some naming inconsistencies here. Do we need the `FirebaseAppDistribution` prefix here? Feels a bit superflous.

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -242,7 +242,7 @@ public void updateToLatestRelease_whenNewAabReleaseAvailable_showsUpdateDialog()
     AppDistributionReleaseInternal latestRelease = TEST_RELEASE_NEWER_AAB_INTERNAL.build();
     when(mockCheckForUpdateClient.checkForUpdate()).thenReturn(Tasks.forResult(latestRelease));
     firebaseAppDistribution.setCachedLatestRelease(latestRelease);
-    when(mockUpdateAppClient.updateApp(latestRelease, activity)).thenReturn(new UpdateTaskImpl());
```
pranavrajgopal, line 245: We should add a test to check the logic to only show notifications for the basic configuration

pranavrajgopal (approved):

PR #2905, opened 2021-08-13
Implement AAB updateTask completion on app resume

pranavrajgopal:


rachaprince:
Seems like there might be some complier error, but LGTM! Much cleaner this way

rachaprince (approved):

andrewbibiloni (approved):

PR #2906, opened 2021-08-13
Fix update app synchronization issues

pranavrajgopal:
Making sure we don't use the same cached update task in the `UpdateAppClient` and the `UpdateApkClient` to avoid cross-class synchronization issues


rachaprince (approved):

rachaprince:
Seems like there might be some kind of complier error, but LGTM! Much cleaner this way

andrewbibiloni (approved):

PR #2910, opened 2021-08-16
FAD Add release codeHash verification for APKs

andrewbibiloni:
- Add releaseIdentifierStorage to store Zip hash to external codehash from backend 
- add calculateApkInternalCodeHash to ReleaseIdentificationUtils 
- add cachedCodeHashes concurrent map to checkForUpdate client to map file path and last modified date to intermediate zip hash (to only calculate zip hash once per update) 
- add external codeHash check to isInstalledRelease in checkForUpdateClient
- map ziphash to external hash after download in updateApkClient
- unit tests for isInstalledRelease, fix affected unit tests


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
```diff
@@ -135,11 +147,27 @@ private boolean isNewerBuildVersion(AppDistributionReleaseInternal latestRelease
...
   }
 
-  private boolean isInstalledRelease(AppDistributionReleaseInternal latestRelease) {
+  @VisibleForTesting
+  boolean isInstalledRelease(AppDistributionReleaseInternal latestRelease) {
     if (latestRelease.getBinaryType().equals(BinaryType.APK)) {
-      // TODO(rachelprince): APK codehash verification. For now assume
-      // the release is identical unless the build version is different
-      return true;
+      Context context = firebaseApp.getApplicationContext();
```
rachaprince, line 153: This block of code is getting pretty long. Maybe we should extract a method for lines 153-170

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
```diff
@@ -135,11 +147,27 @@ private boolean isNewerBuildVersion(AppDistributionReleaseInternal latestRelease
...
+      try {
+        metadataPackageInfo =
+            context
+                .getPackageManager()
+                .getPackageInfo(context.getPackageName(), PackageManager.GET_META_DATA);
+      } catch (PackageManager.NameNotFoundException e) {
+        return false;
+      }
+
+      String externalCodeHash = extractApkCodeHash(metadataPackageInfo);
```
rachaprince, line 164: can all of these blocks of code go in the try block? 

That way we could declare metadataPackageInfo inline instead of above the try block

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
```diff
@@ -135,11 +147,27 @@ private boolean isNewerBuildVersion(AppDistributionReleaseInternal latestRelease
...
+
+      if (externalCodeHash == null) {
+        return false;
+      }
+
+      return externalCodeHash.equals(latestRelease.getCodeHash());
```
rachaprince, line 170: Could we add a few comments here explaining what we are comparing?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
```diff
@@ -165,4 +193,16 @@ private long getInstalledAppVersionCode(Context context) throws FirebaseAppDistr
...
+
+  @VisibleForTesting
+  String extractApkCodeHash(PackageInfo packageInfo) {
+    File sourceFile = new File(packageInfo.applicationInfo.sourceDir);
+
+    String key =
+        String.format(
+            Locale.ENGLISH, "%s.%d", sourceFile.getAbsolutePath(), sourceFile.lastModified());
+
+    cachedCodeHashes.putIfAbsent(key, calculateApkInternalCodeHash(sourceFile));
```
rachaprince, line 205: Does this `calculateApkInternalCodeHash(sourceFile)` run every time? I re-read the documentation for putIfAbsent, and I am concerned it might call the calculateApkInternalCodeHash even if the key exists in the map.

Could we add a test to make sure we don't call calculateApkInternalCodeHash if the file exists in the cached hash map?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ReleaseIdentifierStorage.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private final SharedPreferences releaseIdentifierSharedPreferences;
+
+  ReleaseIdentifierStorage(Context applicationContext) {
+    this.releaseIdentifierSharedPreferences =
+        applicationContext.getSharedPreferences(
+            RELEASE_IDENTIFIER_PREFERENCES_NAME, Context.MODE_PRIVATE);
+  }
+
+  void setCodeHashMap(String internalCodeHash, String externalCodeHash) {
```
rachaprince, line 33: One thought: I wonder if we should pass in a more strongly-typed argument for the second parameter. I'm concerned that we are passing in two Strings, and that one of us may later on accidentally swap internal and external codehash.

If we instead pass in AppDistributionReleaseInternal as the second parameter, (and later call `release.codeHash`) we could avoid a caller accidentally swapping the two. I believe NAUTM types this function this way.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/ReleaseIdentificationUtils.java
```diff
@@ -40,4 +50,70 @@ public static String extractInternalAppSharingArtifactId(@NonNull Context appCon
...
+    Log.v(TAG, String.format("Calculating release id for %s", file.getPath()));
+    Log.v(TAG, String.format("File size: %d", file.length()));
+
+    long start = System.currentTimeMillis();
+    long entries = 0;
+    String zipFingerprint = null;
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      ArrayList<Byte> checksums = new ArrayList<>();
+      ZipFile zis = new ZipFile(file);
```
rachaprince, line 65: Could we add a comment explaining why we hash the zip file?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
andrewbibiloni: After writing a unit test it does recalculate it apparently, I am going to rewrite the logic so that it does it only once

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ReleaseIdentifierStorage.java
andrewbibiloni: Yeah that makes sense, will do that

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/ReleaseIdentificationUtils.java
pranavrajgopal: +1 I think high-level comments of what we do in this class or this method would be really useful for the future

pranavrajgopal (approved):

firebase-app-distribution/firebase-app-distribution.gradle
```diff
@@ -48,8 +48,8 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     testImplementation "org.robolectric:robolectric:$robolectricVersion"
     testImplementation "com.google.truth:truth:$googleTruthVersion"
-    testImplementation 'org.mockito:mockito-core:2.25.0'
-    androidTestImplementation "org.mockito:mockito-android:2.25.0"
+    testImplementation 'org.mockito:mockito-inline:3.4.0'
```
rachaprince, line 51: curious, why did this import change?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/CheckForUpdateClient.java
```diff
@@ -150,24 +150,7 @@ private boolean isNewerBuildVersion(AppDistributionReleaseInternal latestRelease
...
-      }
-
-      String externalCodeHash = extractApkCodeHash(metadataPackageInfo);
-
-      if (externalCodeHash == null) {
-        return false;
-      }
-
-      return externalCodeHash.equals(latestRelease.getCodeHash());
+      return latestAndInstalledReleaseSameCodeHash(latestRelease);
```
rachaprince, line 153: Since this method returns a boolean, we should probably follow the convention of asking a question as the name.

For example:
hasSameCodeHashAsInstallledRelease()

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/ReleaseIdentificationUtils.java
```diff
@@ -62,6 +62,8 @@ public static String calculateApkInternalCodeHash(@NonNull File file) {
     try {
       MessageDigest digest = MessageDigest.getInstance("SHA-256");
       ArrayList<Byte> checksums = new ArrayList<>();
+      // hash the ZipFile to reuse existing checksum data for an intermediate hash which gets mapped
+      // to the codeHash returned from the release backend
```
rachaprince, line 66: Can we update this comment to include the why behind using the zip? (In other words, can we answer the question of why we don't just calculate the codeHash returned from the backend?)

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/CheckForUpdateClientTest.java
```diff
@@ -311,4 +315,24 @@ public void isInstalledRelease_whenCodeHashesNotEqual_returnsFalse() {
     doReturn(TEST_CODEHASH_2).when(checkForUpdateClient).extractApkCodeHash(any());
     assertFalse(checkForUpdateClient.isInstalledRelease(TEST_RELEASE_NEWER_APK));
   }
+
+  @Test
+  public void extractApkCodeHash_ifKeyInCachedCodeHashes_doesNotRecalculateZipHash() {
+
+    try (MockedStatic mockedReleaseIdentificationUtils =
```
rachaprince, line 322: we usually don't have a try block in tests. Does this throw an exception?

firebase-app-distribution/firebase-app-distribution.gradle
andrewbibiloni: Was changed because mocking and verifying static methods was not available in the versions that were currently there, changed it to be able to use it for caclulateApkInternalCodeHash

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/CheckForUpdateClientTest.java
andrewbibiloni: Based it on this: https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#48 where they use a try resource block to make sure the mocking of the static method gets closed and is contained to that block

rachaprince: ooo nice. TIL :)

PR #2928, opened 2021-08-24
Make progress even if AppCheck validation fails

schmidt-sebastian:


firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -1146,6 +1145,9 @@ private void sendAppCheckTokenHelper(final boolean restoreStateAfterComplete) {
               }
             }
           }
+          if (restoreStateAfterComplete) {
+            restoreState();
```
maksymmalyhin, line 1149: @schmidt-sebastian Let me double check understanding of the fix. 

What happens:
- RTDB SDK sends the dummy FAC token
- server responds with "not-ok" status but keep the connection open
- the state of the connection is not restored so it is incorrect and prevents other requests from sending

What the fix does:
- state restore is run even in the case "not-ok" response status, so unless the connection is closed the requests can come through.

Does it look accurate? Also, may calling `restoreState()` lead to unintentional behaviour when the connection is closed by the server when the project is enforced?

maksymmalyhin (approved):
LGTM, thank you for the fix! A couple questions just to check my understanding.

Also, is it possible to add tests for this case?

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
schmidt-sebastian: Your understanding is correct.

I also verified what happens if enforcement is enabled. The writes do not get sent, instead the connection is dropped immediately.

```
2021-08-24 14:50:41.646 12502-13043/com.google.firebase.quickstart.database D/Connection: conn_0 - Got control message: {t=s, d=Invalid appcheck token}
2021-08-24 14:50:41.646 12502-13043/com.google.firebase.quickstart.database D/Connection: conn_0 - Connection shutdown command received. Shutting down...
2021-08-24 14:50:41.647 12502-13043/com.google.firebase.quickstart.database W/PersistentConnection: pc_0 - Firebase Database connection was forcefully killed by the server. Will not attempt reconnect. Reason: Invalid appcheck token
2021-08-24 14:50:41.647 12502-13043/com.google.firebase.quickstart.database D/PersistentConnection: pc_0 - Connection interrupted for: server_kill
2021-08-24 14:50:41.648 12502-13043/com.google.firebase.quickstart.database D/Connection: conn_0 - closing realtime connection
2021-08-24 14:50:41.648 12502-13043/com.google.firebase.quickstart.database D/WebSocket: ws_0 - websocket is being closed
2021-08-24 14:50:41.648 12502-13043/com.google.firebase.quickstart.database D/PersistentConnection: pc_0 - Got on disconnect due to OTHER
2021-08-24 14:50:41.650 12502-13043/com.google.firebase.quickstart.database D/WebSocket: ws_0 - WebSocket reached EOF.
2021-08-24 14:50:41.652 12502-13043/com.google.firebase.quickstart.database D/WebSocket: ws_0 - WebSocket error.
    com.google.firebase.database.tubesock.WebSocketException: IO Exception
        at com.google.firebase.database.tubesock.WebSocketWriter.runWriter(WebSocketWriter.java:159)
        at com.google.firebase.database.tubesock.WebSocketWriter.access$000(WebSocketWriter.java:30)
        at com.google.firebase.database.tubesock.WebSocketWriter$1.run(WebSocketWriter.java:47)
        at java.lang.Thread.run(Thread.java:919)
     Caused by: java.net.SocketException: socket is closed
        at com.android.org.conscrypt.ConscryptFileDescriptorSocket$SSLOutputStream.write(ConscryptFileDescriptorSocket.java:618)
        at java.nio.channels.Channels$WritableByteChannelImpl.write(Channels.java:453)
        at com.google.firebase.database.tubesock.WebSocketWriter.writeMessage(WebSocketWriter.java:138)
        at com.google.firebase.database.tubesock.WebSocketWriter.runWriter(WebSocketWriter.java:152)
        at com.google.firebase.database.tubesock.WebSocketWriter.access$000(WebSocketWriter.java:30) 
        at com.google.firebase.database.tubesock.WebSocketWriter$1.run(WebSocketWriter.java:47) 
        at java.lang.Thread.run(Thread.java:919) 
2021-08-24 14:50:41.652 12502-13043/com.google.firebase.quickstart.database D/WebSocket: ws_0 - closed
```

maksymmalyhin: I guess it means that error handling at lines 1124-1145 is not actually needed, but it seems to be minor.

schmidt-sebastian: Hm. Should the SDK re-connect right away in case this is a temporary failure (as outlined in the comment)? It does not re-connect right now.

maksymmalyhin: In the logs you posted there is a message ` Firebase Database connection was forcefully killed by the server. Will not attempt reconnect. Reason: Invalid appcheck token`. It looks like valid behaviour. I guess it's handled at a different place in the class?

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -559,12 +559,17 @@ public void onDisconnect(Connection.DisconnectReason reason) {
 
   @Override
   public void onKill(String reason) {
-    logger.warn(
-        "Firebase Database connection was forcefully killed by the server. Will not attempt"
-            + " reconnect. Reason: "
-            + reason);
+    if (reason.equals("Invalid AppCheck token")) {
+      logger.warn("Detected invalid AppCheck token.Reconnecting");
```
maksymmalyhin, line 563: I wonder if we should increment and check `invalidAppCheckTokenCount` here as well to avoid an infinite reconnection loop in case e.g. when an App Check provider is not configured?

PR #2929, opened 2021-08-24
Implement mutation squashing

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java
```diff
@@ -0,0 +1,119 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.model.mutation;
+
+import static com.google.firebase.firestore.util.Assert.fail;
+import static com.google.firebase.firestore.util.Assert.hardAssert;
+
+import com.google.common.collect.Sets;
```
schmidt-sebastian, line 20: Please do not use Guava APIs as we eventually want to shed all Guava dependencies.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Utility class to compute a mutation from an original document, and a mutated version. There are
+ * likely multiple mutations happened in between, hence "squash".
+ */
+public class MutationSquash {
```
schmidt-sebastian, line 34: I think we are going to have a lot of fun with the name discussion for this one - do you have alternative suggestions? At the very least this class should end with "s" (Collections, Arrays, our own Values class) as it provides a bunch of static utility methods rather than encapsulated state

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java
```diff
@@ -0,0 +1,119 @@
...
+  public enum Type {
```
schmidt-sebastian, line 35: Needs documentation.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java
```diff
@@ -0,0 +1,119 @@
...
+    hardAssert(
+        original.getKey().equals(finalDoc.getKey()), "Squashing documents with different keys.");
+
+    Map<String, Value> originalFields = original.getData().getFieldsMap();
+    Map<String, Value> finalFields = finalDoc.getData().getFieldsMap();
+    HashSet<FieldPath> mask = new HashSet<>();
+    HashMap<String, Value> values = new HashMap<>();
+
+    // Field updates
+    Sets.SetView<String> intersection =
```
schmidt-sebastian, line 91: While this works, I think we are doing some extra work here. We apply the patch mutation to get the final document state, and then we go back to reverse engineer the diff. Do you think we can simplify this we we add this functionality to the Mutation class itself? Something like:

```
class Mutation {
   void applyToLocalView();
   void applyToRemoteDocument();
   Mutation mergeIntoExistingMutation(Mutation target);
}

File: firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java [hunk 5]
```diff
@@ -0,0 +1,119 @@
...
+      case None:
+        return null;
+      default:
+        fail("Unrecognized squash mutation type");
+    }
+    return null;
+  }
+
+  private static SetMutation toSetMutation(MutableDocument original, MutableDocument finalDoc) {
+    hardAssert(
```
schmidt-sebastian, line 68: You can move this assert to the callsite and thus share the code for all methods.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
```diff
@@ -140,6 +141,19 @@ public void applyToLocalView(MutableDocument document, Timestamp localWriteTime)
     document
         .convertToFoundDocument(getPostMutationVersion(document), document.getData())
         .setHasLocalMutations();
+
+    if (squashType == MutationSquash.Type.Patch) {
```
schmidt-sebastian, line 145: If we have a separate method to compute the final mutation then we likely won't need code like this or even the `MutationSquash.Type` distinction. Instead, we can just return an existing Mutation type that by itself is enough to tell the mutations apart.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    MutableDocument setDoc = doc("collection/key", 0, data);
+    MutableDocument original = setDoc.clone();
+
+    Timestamp now = Timestamp.now();
+    Mutation set = setMutation("collection/key", map("bar", "bar-value"));
+    set.applyToLocalView(setDoc, now, MutationSquash.Type.None);
+
+    Mutation squashed = MutationSquash.toMutation(original, setDoc, MutationSquash.Type.Set);
+    squashed.applyToLocalView(original, now, MutationSquash.Type.None);
+    assertEquals(original, setDoc);
```
schmidt-sebastian, line 61: Can you add a helper here? FWIW this looks very much like `verifyRoundtrip` in our serializer test.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    Mutation delete = new DeleteMutation(key("collection/key"), Precondition.NONE);
+    delete.applyToLocalView(doc, now, MutationSquash.Type.None);
+
+    now = Timestamp.now();
+    Mutation patch = patchMutation("collection/key", map("foo.bar", "new-bar-value"));
+    patch.applyToLocalView(doc, now, MutationSquash.Type.None);
+
+    Mutation squashed = MutationSquash.toMutation(original, doc, MutationSquash.Type.Delete);
+    squashed.applyToLocalView(original, now, MutationSquash.Type.None);
+    assertEquals(doc, original);
```
schmidt-sebastian, line 110: `original` should go first - but I think you can just encode the final state here. It's a deleted doc, right? Your test hides this logic.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+
+    now = Timestamp.now();
+    Mutation patch =
+        mergeMutation(
+            "collection/key", map("foo.bar", "new-bar-value"), Arrays.asList(field("foo.bar")));
+    patch.applyToLocalView(doc, now, MutationSquash.Type.None);
+
+    Mutation squashed = MutationSquash.toMutation(original, doc, MutationSquash.Type.Patch);
+    squashed.applyToLocalView(original, now, MutationSquash.Type.None);
+    assertEquals(doc, original);
```
schmidt-sebastian, line 130: Same comment as above.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    List<MutableDocument> docs =
+        Lists.newArrayList(
+            doc("collection/key", 0, map("foo", "foo-value", "bar.baz", 1)),
+            deletedDoc("collection/key", 0),
+            unknownDoc("collection/key", 0));
+    List<Mutation> mutations =
+        Lists.newArrayList(
+            setMutation("collection/key", map("bar", "bar-value")),
+            new DeleteMutation(key("collection/key"), Precondition.exists(true)),
+            new DeleteMutation(key("collection/key"), Precondition.exists(true)),
```
schmidt-sebastian, line 144: Is there a difference between these?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+import static com.google.firebase.firestore.testutil.TestUtil.key;
+import static com.google.firebase.firestore.testutil.TestUtil.map;
+import static com.google.firebase.firestore.testutil.TestUtil.mergeMutation;
+import static com.google.firebase.firestore.testutil.TestUtil.patchMutation;
+import static com.google.firebase.firestore.testutil.TestUtil.setMutation;
+import static com.google.firebase.firestore.testutil.TestUtil.unknownDoc;
+import static org.junit.Assert.assertEquals;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
```
schmidt-sebastian, line 29: Note that the usage here is fine since this is a test-only class.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    caseNumber += runPermutationTests(docs, Lists.newArrayList(mutations));
+
+    // There are 4! * 3 cases
+    assertEquals(72, caseNumber);
+  }
+
+  @Test
+  public void testSquashMutationByCombinationsAndPermutations() {
+    List<MutableDocument> docs =
+        Lists.newArrayList(
```
schmidt-sebastian, line 159: Optional: Is this different from `Arrays.of()`, which we usually use?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+
+    int caseNumber = 0;
+    for (int subsetSize = 0; subsetSize <= mutations.size(); ++subsetSize) {
+      Set<Set<Mutation>> combinations = Sets.combinations(Sets.newHashSet(mutations), subsetSize);
+      for (Set<Mutation> combination : combinations) {
+        caseNumber += runPermutationTests(docs, Lists.newArrayList(combination));
+      }
+    }
+
+    // There are (0! + 6*1! + 15*2! + 20*3! + 15*4! + 6*5! + 6!) * 3 = 5871 cases.
```
schmidt-sebastian, line 186: It is not clear to me where those numbers come from.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    assertEquals(5871, caseNumber);
+  }
+
+  private int runPermutationTests(List<MutableDocument> docs, List<Mutation> mutations) {
+    Timestamp now = Timestamp.now();
+    int caseNumber = 0;
+    List<List<Mutation>> permutations = generatePermutations(Lists.newArrayList(mutations));
+    for (MutableDocument doc : docs) {
```
schmidt-sebastian, line 194: s/doc/originalDoc ?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+      for (List<Mutation> permutation : permutations) {
+        MutableDocument document = doc.clone();
+        MutableDocument docCopy = document.clone();
```
schmidt-sebastian, line 197: Maybe rename to Source/Target and initialize both with originalDoc.clone?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+        for (Mutation mutation : permutation) {
+          type = mutation.applyToLocalView(document, now, type);
+        }
+        Mutation squashed = MutationSquash.toMutation(docCopy, document, type);
+        if (squashed != null) {
+          squashed.applyToLocalView(docCopy, now, MutationSquash.Type.None);
+        }
+        assertEquals(getDescription(caseNumber, permutation, squashed), document, docCopy);
+
+        caseNumber += 1;
```
schmidt-sebastian, line 208: `++caseNumber` - but I wonder if this is needed since there is no logic here that would change the expected state. We should definitely have something like this if the case number is dependent on the logic in the client, but it seems to just be the number of iterations in the loop.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    builder.append("mutations:\n");
+    for (Mutation mutation : mutations) {
+      builder.append(mutation.toString() + "\n");
+    }
+    builder.append("\n");
+    builder.append("squashed:\n");
+    builder.append(squashed == null ? "null" : squashed.toString());
+    builder.append("\n\n");
+
+    return builder.toString();
```
schmidt-sebastian, line 226: Thank you.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+
+        caseNumber += 1;
+      }
+    }
+    return caseNumber;
+  }
+
+  private String getDescription(int caseNumber, List<Mutation> mutations, Mutation squashed) {
+    StringBuilder builder = new StringBuilder();
+    builder.append("MutationSquash test (" + caseNumber + ") failed with:\n");
```
schmidt-sebastian, line 216: FWIW I am not sure if the case number is helpful here (what does it mean if test 3342 failed?) The rest of the description should be sufficient to trace back to the original test.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+
+    int caseNumber = 0;
+    caseNumber += runPermutationTests(docs, Lists.newArrayList(mutations));
+
+    // There are 4! * 3 cases
+    assertEquals(72, caseNumber);
+  }
+
+  @Test
+  public void testSquashMutationByCombinationsAndPermutations() {
```
schmidt-sebastian, line 157: What is the runtime of this test?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+    builder.append("\n");
+    builder.append("squashed:\n");
+    builder.append(squashed == null ? "null" : squashed.toString());
+    builder.append("\n\n");
+
+    return builder.toString();
+  }
+
+  private <E> List<List<E>> generatePermutations(List<E> mutations) {
```
schmidt-sebastian, line 229: E == Mutation?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,250 @@
...
+    E first = mutations.remove(0);
+    List<List<E>> returnValue = Lists.newArrayList();
+    // Generate permutation with first element removed
+    List<List<E>> permutations = generatePermutations(mutations);
+    for (List<E> permutation : permutations) {
+      // Insert `first` to each possible position of the permutation.
+      for (int index = 0; index <= permutation.size(); index++) {
+        List<E> permutationCopy = Lists.newArrayList(permutation);
+        permutationCopy.add(index, first);
+        returnValue.add(permutationCopy);
```
schmidt-sebastian, line 245: You passed the Coding & Algorithms interview :)

schmidt-sebastian (commented):
The logic looks correct, but I do have some high level comments. Let me know what you think and if we need to discuss

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java
wu-hui: I opted for simplicity by trying to avoid doing merge while applying mutations (otherwise we need to handle merge(set, patch), merge(patch, patch), etc..I started that way and found it hard to get to a correct implementation), and also felt `applyToLocalView()` then call `mergeIntoExistingMutation` a bit too verbose.

If the main concern is performance, I think we can pass a FieldMask along the application chain, and only pick those fields to build a final mutation.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
wu-hui: There is likely still some if/else cases, like merge a patchWithMerge to a delete, or a patch to a set?

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
wu-hui: 8 seconds.

wu-hui (commented):
You are quick!

I addressed the most important comment, let's work this one out before I address others.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquash.java
schmidt-sebastian: I think I have two concerns: 
- You are essentially duplicating the mutation class hierarchy here and providing separate implementations for each mutation type (ignoring verify). This hides the coupling between the two implementations, even though these are actually tightly coupled. 
- The patch implementation here should be able to re-use private logic in PatchMutations. Wouldn't a PatchMutation applied to an empty document result in a squashed diff?

PR #2933, opened 2021-08-25
squash alternative

wu-hui:


wu-hui:
@schmidt-sebastian 

Sorry, I changed the implementation again, please take another look. I do think this version looks better.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ArrayTransformOperation.java
```diff
@@ -78,6 +80,17 @@ public int hashCode() {
     return result;
   }
 
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    sb.append("[");
+    for (Value element : elements) {
+      sb.append(element + "; ");
```
schmidt-sebastian, line 88: FWIW I usually use the canonical ID for this.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ArrayTransformOperation.java
```diff
@@ -133,4 +156,44 @@ protected Value apply(@Nullable Value previousValue) {
       return Value.newBuilder().setArrayValue(result).build();
     }
   }
+
+  /**
+   * An array transform operation that delegate its work to an internal list of other array
```
schmidt-sebastian, line 161: delegates

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ArrayTransformOperation.java
```diff
@@ -133,4 +156,44 @@ protected Value apply(@Nullable Value previousValue) {
...
+   * transform operations.
+   *
+   * <p>When a list of array transform operations being applied, the order of application needs to
```
schmidt-sebastian, line 164: Grammar

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ArrayTransformOperation.java
```diff
@@ -133,4 +156,44 @@ protected Value apply(@Nullable Value previousValue) {
...
+    void addTransform(ArrayTransformOperation transform) {
+      operations.add(transform);
+    }
+
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder();
+      sb.append("Internal operations: ");
+      for (ArrayTransformOperation operation : operations) {
+        sb.append("{" + operation + " }");
```
schmidt-sebastian, line 184: Do we need a comma and spaces here?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/ArrayTransformOperation.java
```diff
@@ -133,4 +156,44 @@ protected Value apply(@Nullable Value previousValue) {
...
+
+  /**
+   * An array transform operation that delegate its work to an internal list of other array
+   * transform operations.
+   *
+   * <p>When a list of array transform operations being applied, the order of application needs to
+   * be preserved, and they are not be squashed into one simple mutation. This class exists such
+   * that the squashed array transform operation still has the same order.
+   */
+  static class ArrayTransformList extends ArrayTransformOperation {
```
schmidt-sebastian, line 168: I feel there might be better name if you can fit one of the patterns here - https://en.wikipedia.org/wiki/Software_design_pattern

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquasher.java
```diff
@@ -0,0 +1,249 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Squashes multiple mutations applied on a document into a final mutation. It is guaranteed when
+ * the final mutation is applied on a document, it is as if the entire mutation list has been
+ * applied.
+ */
+public class MutationSquasher {
```
schmidt-sebastian, line 38: If we leave Mutations as is, then I don't think we should have this as a separate class. I think that we can use this design if we also combine the Mutation hierarchy into one class. Is that something that you can prototype first? We could then base this PR off this change, which would make it much cleaner.

I will try to find some time to experiment with moving the logic into the individual mutation classes as  well.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquasher.java
```diff
@@ -0,0 +1,249 @@
...
+  /** Squashes the given mutation */
+  public void squash(Mutation mutation) {
+    if (mutation instanceof SetMutation) {
+      currentSquashed = mutation;
+      exists = true;
+    } else if (mutation instanceof DeleteMutation) {
+      currentSquashed = mutation;
+      exists = false;
+    } else if (currentSquashed == null) {
+      squashPatchMutationWithNull((PatchMutation) mutation);
```
schmidt-sebastian, line 66: It is not obvious to me why this is a Patch (but I believe you). 

I am also a bit confused why this switches from checking the mutation type to checking `currentSquashed`.

BTW - what do you think of just calling these "overlays" instead of squashed mutations?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquasher.java
```diff
@@ -0,0 +1,249 @@
...
+      squashPatchMutationWithDelete((DeleteMutation) currentSquashed, (PatchMutation) mutation);
+    } else if (currentSquashed instanceof SetMutation) {
+      squashPatchMutationWithSet((SetMutation) currentSquashed, (PatchMutation) mutation);
+      exists = true;
+    } else {
+      squashPatchMutationWithPatch((PatchMutation) currentSquashed, (PatchMutation) mutation);
+    }
+  }
+
+  private void squashPatchMutationWithNull(PatchMutation mutation) {
```
schmidt-sebastian, line 77: Do we need to compute an overlay here? Should we not be rejecting the null+Patch case?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquasher.java
```diff
@@ -0,0 +1,249 @@
...
+    if (!mutation.getPrecondition().isValidFor(doc)) {
+      return;
+    }
+
+    PatchMutation patchWithNoTransform =
+        new PatchMutation(
+            mutation.getKey(), mutation.getValue(), mutation.getMask(), mutation.getPrecondition());
+    patchWithNoTransform.applyToLocalView(doc, Timestamp.now());
+
+    // A Patch on a Delete turns into Set.
```
schmidt-sebastian, line 103: Is this case not also invalid? A patch on a delete should get rejected, or am I misunderstanding?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquasher.java
```diff
@@ -0,0 +1,249 @@
...
+      FieldTransform previous, FieldTransform newTransform) {
+    if (previous == null) {
+      return newTransform;
+    }
+
+    if (newTransform.getOperation() instanceof ServerTimestampOperation) {
+      return newTransform;
+    }
+
+    if (newTransform.getOperation() instanceof ArrayTransformOperation) {
```
schmidt-sebastian, line 193: Do you think this logic can live in the individual transforms?

schmidt-sebastian (commented):
Sorry, this deserves a much more thorough review. I am leaving you some hasty comments in case you want to work on this tomorrow morning (or push back on them...).

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/MutationSquasher.java
wu-hui: Yeah.

PR #2950, opened 2021-09-02
Initial base classes for index backfill

thebrianchen:


firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -143,6 +145,10 @@ public FirestoreClient(
           if (gcScheduler != null) {
             gcScheduler.stop();
           }
+
+          if (indexScheduler != null) {
```
schmidt-sebastian, line 149: Should this scheduler always run? We may also need to backfill indices that were configured after client startup.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -0,0 +1,110 @@
...
+
+import androidx.annotation.Nullable;
+import com.google.firebase.firestore.util.AsyncQueue;
+import java.util.concurrent.TimeUnit;
+
+public class IndexBackfiller {
+  /** How long we wait to try running LRU GC after SDK initialization. */
+  private static final long INITIAL_GC_DELAY_MS = TimeUnit.MINUTES.toMillis(1);
+  /** Minimum amount of time between GC checks, after the first one. */
+  private static final long REGULAR_GC_DELAY_MS = TimeUnit.MINUTES.toMillis(5);
```
schmidt-sebastian, line 25: FWIW I think we should be much more aggressive.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -0,0 +1,110 @@
...
+    public int getEntriesAdded() {
+      return entriesAdded;
+    }
+
+    public int getEntriesRemoved() {
+      return entriesRemoved;
+    }
+  }
+
+  public class Scheduler implements StartStopScheduler {
```
schmidt-sebastian, line 56: I am trying to figure out what functionality a StartStopSchedule offers over a Scheduler, but I am not convinced I will ever find out :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -0,0 +1,110 @@
...
+    private final AsyncQueue asyncQueue;
+    private final LocalStore localStore;
+    private boolean hasRun = false;
+    @Nullable private AsyncQueue.DelayedTask gcTask;
```
schmidt-sebastian, line 60: `gc` for `GeneralIndexBaCkfiller` :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -0,0 +1,110 @@
...
+      if (gcTask != null) {
+        gcTask.cancel();
+      }
+    }
+
+    private void scheduleBackfill() {
+      long delay = hasRun ? REGULAR_GC_DELAY_MS : INITIAL_GC_DELAY_MS;
+      gcTask =
+          asyncQueue.enqueueAfterDelay(
+              AsyncQueue.TimerId.GARBAGE_COLLECTION,
```
schmidt-sebastian, line 83: Please use different timerId.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfillerDelegate.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.firebase.firestore.index.IndexEntry;
+import javax.annotation.Nullable;
+
+// TODO: Update with IndexEntry signatures once they've been fleshed out.
+/**
+ * Persistence layers intending to perform index backfills should implement this interface. This
+ * interface defines the operations that the Index Backfiller needs from the persistence layer.
+ */
+public interface IndexBackfillerDelegate {
```
schmidt-sebastian, line 25: Do we need the indirection with the delegate?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerDelegate.java
```diff
@@ -0,0 +1,66 @@
...
+        1,
+        "TEST-BLOB".getBytes(),
+        "sample-uid",
+        "sample-documentId");
+    ;
+  }
+
+  @Override
+  @Nullable
+  public IndexEntry getIndexEntry(int indexId) {
```
schmidt-sebastian, line 54: I don't think this belongs here. You should only be writing. We also need to figure out how to index partial document updates, but even for those I don't think we should have a public API.

schmidt-sebastian: I take this back if you want to use this for testing - but then this should be marked as such.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/MemoryIndexBackfillerTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.google.firebase.firestore.local;
+
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest = Config.NONE)
+public class MemoryIndexBackfillerTest extends IndexBackfillerTestCase {
```
schmidt-sebastian, line 23: This test will not test anything, will it? We can probably just have a single SQLLite backfill test.

schmidt-sebastian (commented):
This looks good, but it could be simplified by making it a bit less generic. We should also figure out how to run this optionally when the indexing flag is enabled.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: Updated to 30 seconds and every minute. We can tune this number once we can run benchmarks.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
thebrianchen: In my mind, the scheduler should always run, but whether the scheduler schedules anything is another question. We can change this as necessary as we build more functionality out.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: Renamed the base class to Scheduler. To be honest, I'm not sure why the LruGarbageCollector's Scheduler used the abstract base class to begin with since it's not shared with any other class.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: 🤯

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: done.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfillerDelegate.java
thebrianchen: Removed the delegate. Initially, I thought we would need to support MemoryPersistence in case the user disables persistence in their app, but I realize we can work around that.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerDelegate.java
thebrianchen: I'll leave this here for the base prototype, and add more features to it as we build it out.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/MemoryIndexBackfillerTest.java
thebrianchen: Done.

thebrianchen (commented):
thanks for the early stage review!

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -143,6 +145,10 @@ public FirestoreClient(
           if (gcScheduler != null) {
             gcScheduler.stop();
           }
+
+          if (indexScheduler != null && Persistence.INDEXING_SUPPORT_ENABLED) {
```
schmidt-sebastian, line 149: Do we need to check the flag here? The scheduler would not be available if the flag is off.

The name is also a bit generic. I wonder if we should include "backfill" in the scheduler name, which reduces the mental scope from "all of indexing" to "backfill index data".

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
```diff
@@ -264,6 +271,10 @@ private void initialize(Context context, User user, FirebaseFirestoreSettings se
     if (gcScheduler != null) {
       gcScheduler.start();
     }
+
+    if (indexScheduler != null && Persistence.INDEXING_SUPPORT_ENABLED) {
```
schmidt-sebastian, line 275: I think we should guard the scheduler assignment based on this flag (line 265). Ideally, if indexing is off, none of the logic should even be loaded. It otherwise becomes hard to reason about the system - e.g. the constructor code of the scheduler needs to be aware that indexing might be on or off.

firebase-firestore/src/main/java/com/google/firebase/firestore/index/IndexEntry.java
```diff
@@ -0,0 +1,49 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.index;
+
+/**
+ * Represents an index entry saved by the SDK in the local storage. Temporary placeholder, since
+ * we'll probably serialize the indexValue right away rather than store it.
```
schmidt-sebastian, line 19: Please add "TODO(indexing)".

Note that we already committed `FieldIndex`. In https://github.com/firebase/firebase-android-sdk/pull/2963, FieldIndex also gets an `indexId`.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -0,0 +1,152 @@
...
+                scheduleBackfill();
+              });
+    }
+  }
+
+  public BackfillScheduler newScheduler(AsyncQueue asyncQueue, LocalStore localStore) {
+    return new BackfillScheduler(asyncQueue, localStore);
+  }
+
+  // TODO: Figure out which index entries to backfill.
```
schmidt-sebastian, line 108: `TODO(indexing)`

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Rule public TestName name = new TestName();
+
+  private SQLitePersistence persistence;
+  private IndexBackfiller backfiller;
+
+  @Before
+  public void setUp() {
+    persistence = PersistenceTestHelpers.createSQLitePersistence();
+    ;
```
schmidt-sebastian, line 57: Tzzzz.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -0,0 +1,152 @@
...
+  public Results backfill() {
+    int numIndexesWritten = 0;
+    int numIndexesRemoved = 0;
+    return new Results(/* hasRun= */ true, numIndexesWritten, numIndexesRemoved);
+  }
+
+  public void addIndexEntry(IndexEntry entry) {
```
schmidt-sebastian, line 115: These methods would probably also be private if it was not for your tests. They should also be marked `@VisibleForTesting` if that is true.

All of the test-only methods should also be package-private.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerTest.java
```diff
@@ -0,0 +1,81 @@
...
+  @After
+  public void tearDown() {
+    persistence.shutdown();
+  }
+
+  @Test
+  public void addAndRemoveIndexEntry() {
+    IndexEntry testEntry =
+        new IndexEntry(1, "TEST_BLOB".getBytes(), "sample-uid", "sample-documentId");
+    backfiller.addIndexEntry(testEntry);
```
schmidt-sebastian, line 70: `addIndex` and related methods should be scoped to a SQLite transaction and not be run without it.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FirestoreClient.java
thebrianchen: Guarded based on the flag and persistence being enabled.

schmidt-sebastian (approved):

PR #2955, opened 2021-09-03
Squash in Mutations

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
```diff
@@ -142,6 +143,32 @@ public void applyToLocalView(MutableDocument document, Timestamp localWriteTime)
         .setHasLocalMutations();
   }
 
+  @Override
+  public Mutation squash(
+      Mutation baseMutation, MutableDocument document, Timestamp localWriteTime) {
+    if (!getPrecondition().isValidFor(document)) {
```
wu-hui, line 149: precondition should be tested against `document` + `baseMutation`, isn't it?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
```diff
@@ -142,6 +143,32 @@ public void applyToLocalView(MutableDocument document, Timestamp localWriteTime)
...
+      return baseMutation;
+    }
+
+    Map<FieldPath, Value> transformResults = localTransformResults(localWriteTime, document);
```
wu-hui, line 153: Same here, transforms need to be applied on `document` + `baseMutation`.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
```diff
@@ -142,6 +143,32 @@ public void applyToLocalView(MutableDocument document, Timestamp localWriteTime)
...
+        baseMutation.getValue() != null ? baseMutation.getValue() : new ObjectValue();
+    mergedPatch.setAll(getPatch());
+    mergedPatch.setAll(transformResults);
+
+    if (baseMutation.getMask() == null) {
+      // TODO: Use strictest preconditions or assert we only use exists
+      return new SetMutation(getKey(), mergedPatch, getPrecondition());
+    } else {
+      HashSet<FieldPath> mergedMaskSet = new HashSet<>(baseMutation.getMask().getMask());
+      mergedMaskSet.addAll(mask.getMask());
```
wu-hui, line 165: `mergedMaskSet` should also include fields from transforms.

wu-hui (commented):
Left some questions, just to make sure if I understand your intention correctly.

Overall I think this will work too, and is probably going to be simpler than my implementation of squashing transforms.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
schmidt-sebastian: I don't think so (but I could be wrong). `baseMutation` is a terrible name (as we are already using it somewhere else in this client with a different meaning), but this `baseMutation` should already have passed all preconditions, otherwise it would not have changed during the last call to `squash`.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
schmidt-sebastian: I actually think we need to do two different things:

- First round, we use the document
- Second round, we use the base mutation

So ideally we would change `localTransformResults` to take an ObjectValue instead of the document.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
schmidt-sebastian: That is correct.

PR #2956, opened 2021-09-03
FAD Changing names to be more consistent with API council change request

mannyjimenez0810:
Renaming a lot of methods, files, and variables to be more in line with other changes.


rachaprince (approved):
Thanks for running with this refactor!

PR #2959, opened 2021-09-07
Merge all mutations into one class

wu-hui:
This PR turns all different mutations into one class, using an enum to differentiate mutation types. This is similar to MutableDocument, and will make subsequent changes to mutation squashing cleaner.


firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -89,10 +170,120 @@ public Precondition getPrecondition() {
...
+              + ", value="
+              + value
+              + "}";
+        }
+      case VERIFY:
+        {
+          return "Mutation{ type=VERIRY; " + keyAndPreconditionToString();
+        }
+    }
+    hardAssert(false, "Unreachable");
```
schmidt-sebastian, line 244: FWIW I think it would be perfectly fine to just use the toString() from Patch here and return null-values where applicable.

PR #2960, opened 2021-09-07
Remove index_value from primary key of index_entries

thebrianchen:
We can remove `index_value` from the composite key since every `index_id` should correspond to a single `index_value`.


schmidt-sebastian (approved):
Pushed a build file change that should fix deviceCheckChanged.

PR #2962, opened 2021-09-08
Overlay on top of mutation list

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -191,17 +191,51 @@ static SnapshotVersion getPostMutationVersion(MutableDocument document) {
    * @return A map of fields to transform results.
    */
   protected Map<FieldPath, Value> localTransformResults(
-      Timestamp localWriteTime, MutableDocument mutableDocument) {
+      Timestamp localWriteTime, MutableDocument mutableDocument, Mutation mutation) {
     Map<FieldPath, Value> transformResults = new HashMap<>(fieldTransforms.size());
     for (FieldTransform fieldTransform : fieldTransforms) {
       TransformOperation transform = fieldTransform.getOperation();
       Value previousValue = mutableDocument.getField(fieldTransform.getFieldPath());
+      if (mutation != null) {
```
schmidt-sebastian, line 199: I wonder if we need this if we pass ObjectValue to this method instead of MutableDocument+mutation. Would this work?

```
diff --git a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
index f5ea818e..934868a6 100644
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
@@ -191,19 +191,11 @@ public abstract class Mutation {
    * @return A map of fields to transform results.
    */
   protected Map<FieldPath, Value> localTransformResults(
-      Timestamp localWriteTime, MutableDocument mutableDocument, Mutation mutation) {
+      Timestamp localWriteTime, ObjectValue currentValue) {
     Map<FieldPath, Value> transformResults = new HashMap<>(fieldTransforms.size());
     for (FieldTransform fieldTransform : fieldTransforms) {
       TransformOperation transform = fieldTransform.getOperation();
-      Value previousValue = mutableDocument.getField(fieldTransform.getFieldPath());
-      if (mutation != null) {
-        FieldUpdate update = getFieldUpdate(fieldTransform.getFieldPath());
-        if (update.type == FieldUpdate.Type.DELETE) {
-          previousValue = null;
-        } else if (update.type == FieldUpdate.Type.SET) {
-          previousValue = update.value;
-        }
-      }
+      Value previousValue = currentValue.get(fieldTransform.getFieldPath());
       transformResults.put(
           fieldTransform.getFieldPath(), transform.applyToLocalView(previousValue, localWriteTime));
     }
diff --git a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
index ed93bb97..06eb8fad 100644
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
@@ -14,6 +14,8 @@
 
 package com.google.firebase.firestore.model.mutation;
 
+
 import com.google.firebase.Timestamp;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.FieldPath;
@@ -134,10 +136,12 @@ public final class PatchMutation extends Mutation {
       return;
     }
 
-    Map<FieldPath, Value> transformResults = localTransformResults(localWriteTime, document, null);
     ObjectValue value = document.getData();
     value.setAll(getPatch());
+
+    Map<FieldPath, Value> transformResults = localTransformResults(localWriteTime, value);
     value.setAll(transformResults);
+
     document
         .convertToFoundDocument(getPostMutationVersion(document), document.getData())
         .setHasLocalMutations();
@@ -145,7 +149,7 @@ public final class PatchMutation extends Mutation {
 
   @Override
   public Mutation squash(
-      Mutation baseMutation, MutableDocument document, Timestamp localWriteTime) {
+          @Nullable Mutation baseMutation, MutableDocument document, Timestamp localWriteTime) {
     MutableDocument preconditionTest =
         (baseMutation != null && (baseMutation instanceof DeleteMutation))
             ? MutableDocument.newNoDocument(getKey(), document.getVersion())
@@ -156,7 +160,7 @@ public final class PatchMutation extends Mutation {
     }
 
     Map<FieldPath, Value> transformResults =
-        localTransformResults(localWriteTime, document, baseMutation);
+        localTransformResults(localWriteTime, baseMutation.getValue());
 
     ObjectValue mergedPatch =
         (baseMutation != null && baseMutation.getValue() != null)
```

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
```diff
@@ -146,27 +146,53 @@ public void applyToLocalView(MutableDocument document, Timestamp localWriteTime)
   @Override
   public Mutation squash(
       Mutation baseMutation, MutableDocument document, Timestamp localWriteTime) {
-    if (!getPrecondition().isValidFor(document)) {
+    MutableDocument preconditionTest =
+        (baseMutation != null && (baseMutation instanceof DeleteMutation))
```
schmidt-sebastian, line 150: Is there some way we can make baseMutation non-nullable? We could use a VerifyMutation that does not verify anything for example, an empty merge with an exist precondition or a new EmptyMutation class. Ideally, this would allow us to always have the correct state in "document" which would mean that we do not have to merge preconditions.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
```diff
@@ -191,17 +191,51 @@ static SnapshotVersion getPostMutationVersion(MutableDocument document) {
...
+          previousValue = update.value;
+        }
+      }
       transformResults.put(
           fieldTransform.getFieldPath(), transform.applyToLocalView(previousValue, localWriteTime));
     }
     return transformResults;
   }
 
+  protected static class FieldUpdate {
```
schmidt-sebastian, line 213: This is pretty elegant but maybe we don't need this if we use an ObjectValue rather than MutableDocument+Mutation strategy as described above.

schmidt-sebastian (commented):
Thanks for playin with this. Some more ideas for simplification - some or all of which might not work.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
wu-hui: I think this only works when `baseMutation` is `Set`. When it is `Patch`, the transforms need to be applied on top of fields from `document` + `baseMutation`.

Passing `getValue` from `baseMutation` will missing the field values we need from `document`, and the transforms will be applied as if those fields are missing if `baseMutation.getValue()` contains no such fields.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
wu-hui: See above.

wu-hui (commented):
Thanks. I am adding tests back to validate this impl...and cleaning it up.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,497 @@
...
+    patch.applyToLocalView(doc, now);
+
+    Mutation squashed = emptyMutation("collection/key");
+    squashed = set.squash(squashed, original, now);
+    squashed = patch.squash(squashed, original, now);
+    squashed.applyToLocalView(original, now);
+    assertEquals(doc, original);
+  }
+
+  // TODO(overlay): This test fails because deleting nested fields like bar.baz leaves an empty bar
```
wu-hui, line 251: @schmidt-sebastian This currently fails, but I do not think this is an issue, we should discuss this.

schmidt-sebastian: I think this is an issue. This changes the behavior - we should be left with an empty map. I don't (yet) know how to solve this though.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Mutation.java
schmidt-sebastian: I am trying to get this to work. I think I am closeish but I still have more test failures than before. Will play around with it more tomorrow. See https://github.com/firebase/firebase-android-sdk/pull/2966 for a broken draft.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,449 @@
...
+        patchMutation(
+            "collection/key",
+            map("foo", "foo-patched-value", "bar.baz", FieldValue.serverTimestamp()));
+
+    squashRoundTrips(doc, set, patch);
+  }
+
+  @Test
+  public void testSquashFieldDeletionOfNestedField() {
+    MutableDocument doc = doc("collection/key", 0, map("bar.baz", 1));
```
schmidt-sebastian, line 163: This does not created a nested object. Rather it creates an object with a single key of `bar.baz`.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,449 @@
...
+          // still remain (which is expected).
+          // Consider mutation (foo.bar 1), then mutation (foo.bar delete()).
+          // This leaves the final result (foo, {}). Despite `doc` has the correct result,
+          // `foo` is not in `mask`, and the resulting mutation will miss `foo` completely without
+          // this logic.
+          if (value == null && path.length() > 1) {
+            path = path.popLast();
+          }
+          objectValue.set(path, docValue.get(path));
+          maskSet.add(path);
```
schmidt-sebastian, line 440: This code snippet should be unit tested explicitly once moved to production code.

I also just figured out why my original proposal was working: `ObjectValue.set('a.b', delete())` creates field 'a', which it should not, but it made my original implementation work.

schmidt-sebastian (commented):
This looks good! Most of the comments are nits about a repeated pattern that we should be able to fix easily.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
schmidt-sebastian: Hm, I think the problem that caused my test to fail was that the both the original and the new doc did not have an entry for "bar". Not sure how this happened, but this points to another thing we need to do - we should convert most test to compare against a hard-coded end state rather than a computed end state.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,449 @@
...
+import java.util.Set;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+/** Tests MutationSquasher */
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest = Config.NONE)
+public class MutationSquashTest {
```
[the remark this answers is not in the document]
wu-hui: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,449 @@
...
+  @Test
+  public void testSquashOneSetMutation() {
+    Map<String, Object> data = map("foo", "foo-value", "baz", "baz-value");
+    squashRoundTrips(
```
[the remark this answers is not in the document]
wu-hui: I think with current implementation, `overlay` is a better name.

firebase-firestore/src/test/java/com/google/firebase/firestore/model/MutationSquashTest.java
```diff
@@ -0,0 +1,449 @@
...
+          Value value = docValue.get(path);
+          // If it is deleting a nested field, we take the immediate parent as the mask used to
+          // construct resulting mutation.
+          // Justification: Nested fields can create parent fields implicitly, if they are
+          // deleted in later mutations, there is no trace that the empty parent fields should
+          // still remain (which is expected).
+          // Consider mutation (foo.bar 1), then mutation (foo.bar delete()).
+          // This leaves the final result (foo, {}). Despite `doc` has the correct result,
+          // `foo` is not in `mask`, and the resulting mutation will miss `foo` completely without
+          // this logic.
```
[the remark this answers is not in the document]
wu-hui: Manually applied because code moved.

wu-hui (commented):
Thanks for the review!

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/FieldMask.java
```diff
@@ -27,13 +28,30 @@
...
-    return new FieldMask(mask);
+    return new FieldMask(mask, Scope.BY_MASK);
+  }
+
+  public static FieldMask allFieldsMask() {
+    return new FieldMask(new HashSet<FieldPath>(), Scope.ALL_FIELDS);
+  }
+
+  public static FieldMask emptyMask() {
+    return new FieldMask(new HashSet<FieldPath>(), Scope.NONE);
```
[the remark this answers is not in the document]
schmidt-sebastian: I am still trying to get it to work without this change (will continue tomorrow).

Reasons I advocate against it:
- It exposes a lot of state on the FieldMask class, some of which is mutually exclusive and only sensical if the consumer understand the underlying states. I wish we could somehow redesign this so that the mask type is hidden from the API.
- It breaks `getMask()`
- It breaks `covers()` (which can be fixed)

We would also need to fix equals+hashCode. 

I will report back tomorrow if I can come up with a solution.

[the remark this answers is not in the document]
schmidt-sebastian: Follow up PR: https://github.com/firebase/firebase-android-sdk/pull/2983

Let me know what you think.

Note that the changes I made uncover a versioning issue that exists in this PR as well (there is a comment in the code changes explaining it).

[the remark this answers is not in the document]
wu-hui: LGTM for the follow up, and thanks for doing it!

Re: version issue, this could be fixed by having a version update in the final mutation. We do need to serialize this as well. Thoughts?

wu-hui:
Pushed a new commit trying to fix the version issue.

Note I have to change the test `testAppliesPatchWithMergeToDocuments` to use `deletedDoc(version:None)` to make it pass. This fails without the new logic added in this PR if the version is `1`, because post mutation version for `NoDocument` is always `0`.

schmidt-sebastian (commented):
Can you rebase against `master`?

wu-hui:
> Can you rebase against `master`?

Done..and thanks for the fix!

schmidt-sebastian (approved):
Thank you for working through this! This looks great.

PR #2963, opened 2021-09-08
Index Evaluation

schmidt-sebastian:
This PR adds the code that is needed to add documents to the Index and to run queries against the Index.

Couple changes from previous PRs:
- The schema now uses collection_groups and no longer collection_ids. This was true before but it also now reflected in the names.
- Since we use collection groups, the schema actually needs to contain the relative path of the document rather than just the document ID.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
@@ -110,8 +110,8 @@ final class SQLiteIndexManager implements IndexManager {
 
   @Override
   public void addIndexEntries(Document document) {
-    ResourcePath documentPath = document.getKey().getPath();
-    String collectionGroup = documentPath.getSegment(documentPath.length() - 2);
+    DocumentKey documentKey = document.getKey();
+    String collectionGroup = documentKey.getCollectionGroup();
     db.query(
             "SELECT index_id, index_proto FROM index_configuration WHERE collection_group = ? AND active = 1")
         .binding(collectionGroup)
@@ -127,10 +127,10 @@ final class SQLiteIndexManager implements IndexManager {
                 if (values == null) return;
 
                 if (Logger.isDebugEnabled()) {
-                  Logger.warn(
+                  Logger.debug(
                       TAG,
                       "Adding index values for document '%s' to index '%s'",
-                      documentPath,
+                      documentKey,
                       fieldIndex);
                 }
 
@@ -144,7 +144,7 @@ final class SQLiteIndexManager implements IndexManager {
                           + "document_name) VALUES(?, ?, ?)",
                       indexId,
                       encoded,
-                      documentPath.canonicalString());
+                      documentKey.toString());
                 }
               } catch (InvalidProtocolBufferException e) {
                 throw fail("Invalid index: " + e);
@@ -176,6 +176,8 @@ final class SQLiteIndexManager implements IndexManager {
     if (fieldIndex == null) return null;
 
     Bound lowerBound = target.getLowerBound(fieldIndex);
+    String lowerBoundOp = lowerBound.isBefore() ? ">=" : ">"; // `startAt()` versus `startAfter()`
+
     @Nullable Bound upperBound = target.getUpperBound(fieldIndex);
 
     if (Logger.isDebugEnabled()) {
@@ -200,13 +202,15 @@ final class SQLiteIndexManager implements IndexManager {
           lowerBoundValues.size() == upperBoundValues.size(),
           "Expected upper and lower bound size to match");
 
+      String upperBoundOp = upperBound.isBefore() ? "<" : "<="; // `endBefore()` versus `endAt()`
+
       // TODO(indexing): To avoid reading the same documents multiple times, we should ideally only
       // send one query that combines all clauses.
       for (int i = 0; i < lowerBoundValues.size(); ++i) {
         db.query(
                 String.format(
                     "SELECT document_name from index_entries WHERE index_id = ? AND index_value %s ? AND index_value %s ?",
-                    lowerBound.isBefore() ? ">=" : ">", upperBound.isBefore() ? "<=" : "<"))
+                    lowerBoundOp, upperBoundOp))
             .binding(fieldIndex.getIndexId(), lowerBoundValues.get(i), upperBoundValues.get(i))
             .forEach(
                 row -> result.add(DocumentKey.fromPath(ResourcePath.fromString(row.getString(0)))));
@@ -218,7 +222,7 @@ final class SQLiteIndexManager implements IndexManager {
         db.query(
                 String.format(
                     "SELECT document_name from index_entries WHERE index_id = ? AND index_value %s  ?",
-                    lowerBound.isBefore() ? ">=" : ">"))
+                    lowerBoundOp))
             .binding(fieldIndex.getIndexId(), lowerBoundValue)
             .forEach(
                 row -> result.add(DocumentKey.fromPath(ResourcePath.fromString(row.getString(0)))));
@@ -259,15 +263,14 @@ final class SQLiteIndexManager implements IndexManager {
                 throw fail("Failed to decode index: " + e);
               }
             });
-    ;
 
     if (activeIndices.isEmpty()) {
       return null;
     }
 
     // Return the index with the most number of segments
-    Collections.sort(activeIndices, (l, r) -> Integer.compare(r.segmentCount(), l.segmentCount()));
-    return activeIndices.get(0);
+    return Collections.max(
+        activeIndices, (l, r) -> Integer.compare(l.segmentCount(), r.segmentCount()));
   }
 
   /**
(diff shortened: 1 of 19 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
```diff
@@ -14,15 +14,210 @@
...
+import java.util.stream.Collectors;
+import org.junit.BeforeClass;
+import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
 
 @RunWith(RobolectricTestRunner.class)
 @Config(manifest = Config.NONE)
 public class SQLiteIndexManagerTest extends IndexManagerTestCase {
```
wu-hui, line 43: Can we add more tests:

* Queries on mixed type document fields
* Collection group queries
* Returns null when no index is found
* Queries where no doc can be found.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -87,6 +105,237 @@ public void addFieldIndex(FieldIndex index) {
...
+        return null;
+      }
+      values.add(field);
+    }
+    return values;
+  }
+
+  @Override
+  @Nullable
+  public Set<DocumentKey> getDocumentsMatchingTarget(Target target) {
```
wu-hui, line 162: Can we add some logging here? It might come in handy troubleshooting performance issues going forward.

wu-hui (changes requested):
Overall good. 

I will do another review for comments and code structure later.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: Added logging.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
schmidt-sebastian: Thanks for the suggestions. All added.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -87,6 +108,259 @@ public void addFieldIndex(FieldIndex index) {
...
+  /**
+   * Returns the list of values for all fields indexed by the provided field index. Returns {@code
+   * null} if one or more of the fields are not set.
+   */
+  @Nullable
+  private List<Value> extractFieldValue(Document document, FieldIndex fieldIndex) {
+    List<Value> values = new ArrayList<>();
+    for (FieldIndex.Segment segment : fieldIndex) {
+      Value field = document.getField(segment.getFieldPath());
+      if (field == null) {
```
wu-hui, line 164: If not already, we should have a test where a field is set to null explicitly.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -87,6 +108,259 @@ public void addFieldIndex(FieldIndex index) {
...
+  /**
+   * Returns an index that can be used to serve the provided target. Returns {@code null} if no
+   * index is configured.
+   */
+  private @Nullable FieldIndex getMatchingIndex(Target target) {
+    TargetIndexMatcher targetIndexMatcher = new TargetIndexMatcher(target);
+    String collectionGroup =
+        target.getCollectionGroup() != null
+            ? target.getCollectionGroup()
+            : target.getPath().getLastSegment();
```
wu-hui, line 241: I am confused here, if we use `getLastSegment` here, would it cause conflicts? I was under the impression that `collection_group` is full path?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -87,6 +108,259 @@ public void addFieldIndex(FieldIndex index) {
...
+              }
+            });
+    ;
+
+    if (activeIndices.isEmpty()) {
+      return null;
+    }
+
+    // Return the index with the most number of segments
+    Collections.sort(activeIndices, (l, r) -> Integer.compare(r.segmentCount(), l.segmentCount()));
```
wu-hui, line 269: Can we use `Collections.max` instead?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -87,6 +108,259 @@ public void addFieldIndex(FieldIndex index) {
...
+          lowerBoundValues.size() == upperBoundValues.size(),
+          "Expected upper and lower bound size to match");
+
+      // TODO(indexing): To avoid reading the same documents multiple times, we should ideally only
+      // send one query that combines all clauses.
+      for (int i = 0; i < lowerBoundValues.size(); ++i) {
+        db.query(
+                String.format(
+                    "SELECT document_name from index_entries WHERE index_id = ? AND index_value %s ? AND index_value %s ?",
+                    lowerBound.isBefore() ? ">=" : ">", upperBound.isBefore() ? "<=" : "<"))
```
wu-hui, line 209: Should the order of operator for upper bound be reversed? 

I am always confused with `isBefore()`, so I could be wrong. But if I am right, please add a test for this.

wu-hui: Actually there are tests for this, so the code is right I guess. Can you please add some comment to explain?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
wu-hui: Thanks! Can we add one more: nested fields?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: I myself am confused about this all the time. It takes a long time to understand - and it looks like I got it wrong here as well. The tests were passing because I invalidated invalid assumptions. This was obvious when I added a small comment, which strongly hinted that I was doing something wrong.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: No, the collectionGroup is the name of the parent collection.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: Probably not since it is likely not available at our API level.

Checking...

Wow, I stand corrected. Java 7 for the win!

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
schmidt-sebastian: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -176,6 +176,8 @@ public void addIndexEntries(Document document) {
     if (fieldIndex == null) return null;
 
     Bound lowerBound = target.getLowerBound(fieldIndex);
+    String lowerBoundOp = lowerBound.isBefore() ? ">=" : ">"; // `startAt()` versus `startAfter()`
```
wu-hui, line 179: Not related to this PR, but what do you think about changing `isBefore` to `isIncluded`?

This way we can do `lower.isIncluded ? ">=" : ">"` and `upper.isIncluded ? "<=" : "<"`. We need to change this everywhere and still keep the API for `startAt` and `startAfter` though.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
wu-hui: I cannot find the new test for this one.

wu-hui (approved):
There is one test I cannot find, but LGTM anyways.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: Brian is also confused about this (so am I). I will send a follow-up PR to do a rename.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: `addsDocuments()` is meant to test it. It adds a document that does not have the field being indexed. Note that a field that is set to `null` is being indexed (and its value is `Value.setNullValue()`).

PR #2965, opened 2021-09-09
Fix bug where `cacheSizeBytes` was sometimes not initialized correctly in `FirebaseFirestoreSettings`

dconeybe:
Fix a bug where the "copy constructor" for `FirebaseFirestoreSettings.Builder` forgot to copy the `cacheSizeBytes` field.

The problematic code is:
https://github.com/firebase/firebase-android-sdk/blob/e41b233118beb9f06cd330f4f1239f880df3eb84/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestoreSettings.java#L55-L60


schmidt-sebastian (approved):
Thanks! I am glad we only did this for the property that controls mostly invisible behavior :)

PR #2966, opened 2021-09-09
Mrschmidt/trysomemore

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/PatchMutation.java
```diff
@@ -142,6 +145,54 @@ public void applyToLocalView(MutableDocument document, Timestamp localWriteTime)
...
+  @Override
+  public Mutation squash(Mutation previousMutation, Timestamp localWriteTime) {
+    hardAssert(previousMutation != null, "previousMutation cannot be null");
+
+    if (!getPrecondition().isValidFor(previousMutation)) {
+      return previousMutation;
+    }
+
+    Map<FieldPath, Value> transformResults =
+        localTransformResults(localWriteTime, previousMutation.getValue());
```
wu-hui, line 157: I think this is problematic..transforms need to be applied on document + previousMutation isn't it?

schmidt-sebastian: I changed this one more and pushed the latest version. The PR is passing tests now.

wu-hui: After some experiments, I think I understand how this works..

But this is similar to my original approach, mutating the document along the way, and also keep track of mutateds field in the squashed mutation. At this point, it will be even simpler to just use `applyToLocalView` and build a `FieldMask` along the way if it is an all patch squash. Then we could do this to get a mutation after all mutations has been applied.

```
FieldMask mask = empty();
mask = mutation1.applyToLocalView(..., mask);
if(mask == null) { // Set or Delete
  squashedMutation = // build from doc
} else {
  squashedMutation = // build from doc and mask
}
```

schmidt-sebastian: The end goal should be that we remove one of the two methods. We should already be able to call `squash` from `applyLocalMutations` by simply ignoring the new return type.

Note that this does mean that we need to rewrite the test, as we cannot compare the result of the two operations anymore.

PR #2968, opened 2021-09-10
Fireperf: create a random delay before initiating a remote config fetch. 

leotianlizhan:
**Rationale**:
This is required to ensure that a remote notification does not trigger multiple config fetches from remote config at the same time. Internal bug reference: b/187985523

iOS version of the same change https://github.com/firebase/firebase-ios-sdk/pull/8593

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
@@ -27,10 +27,10 @@ import com.google.firebase.remoteconfig.FirebaseRemoteConfig;
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigValue;
 import com.google.firebase.remoteconfig.RemoteConfigComponent;
 import java.util.Map;
-import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
@@ -50,12 +50,12 @@ public class RemoteConfigManager {
       TimeUnit.HOURS.toMillis(12);
   private static final long FETCH_NEVER_HAPPENED_TIMESTAMP_MS = 0;
   private static final long MIN_APP_START_CONFIG_FETCH_DELAY_MS = 5000;
-  private static final int RANDOM_APP_START_CONFIG_FETCH_DELAY_MS = 25000;
+  private static final long RANDOM_APP_START_CONFIG_FETCH_DELAY_MS = 25000;
 
   private final ConcurrentHashMap<String, FirebaseRemoteConfigValue> allRcConfigMap;
   private final Executor executor;
-  private final long appStartTime;
-  private final long appStartConfigFetchDelay;
+  private final long appStartTimeInMs;
+  private final long appStartConfigFetchDelayInMs;
 
   private long firebaseRemoteConfigLastFetchTimestampMs = FETCH_NEVER_HAPPENED_TIMESTAMP_MS;
 
@@ -79,20 +79,22 @@ public class RemoteConfigManager {
         executor,
         firebaseRemoteConfig,
         MIN_APP_START_CONFIG_FETCH_DELAY_MS
-            + new Random().nextInt(RANDOM_APP_START_CONFIG_FETCH_DELAY_MS));
+            + ThreadLocalRandom.current().nextLong(RANDOM_APP_START_CONFIG_FETCH_DELAY_MS));
   }
 
   @VisibleForTesting
   RemoteConfigManager(
-      Executor executor, FirebaseRemoteConfig firebaseRemoteConfig, long fetchDelay) {
+      Executor executor,
+      FirebaseRemoteConfig firebaseRemoteConfig,
+      long appStartConfigFetchDelayInMs) {
     this.executor = executor;
     this.firebaseRemoteConfig = firebaseRemoteConfig;
     this.allRcConfigMap =
         firebaseRemoteConfig == null
             ? new ConcurrentHashMap<>()
             : new ConcurrentHashMap<>(firebaseRemoteConfig.getAll());
-    this.appStartTime = getCurrentSystemTimeMillis();
-    this.appStartConfigFetchDelay = fetchDelay;
+    this.appStartTimeInMs = getCurrentSystemTimeMillis();
+    this.appStartConfigFetchDelayInMs = appStartConfigFetchDelayInMs;
   }
 
   /** Gets the singleton instance. */
@@ -313,7 +315,7 @@ public class RemoteConfigManager {
     if (allRcConfigMap.isEmpty()) { // Initial fetch.
       syncConfigValues(firebaseRemoteConfig.getAll());
     }
-    if (passesRandomAppStartDelay() && shouldFetchAndActivateRemoteConfigValues()) {
+    if (shouldFetchAndActivateRemoteConfigValues()) {
       triggerFirebaseRemoteConfigFetchAndActivateOnSuccessfulFetch();
     }
   }
@@ -360,6 +362,22 @@ public class RemoteConfigManager {
     return firebaseRemoteConfig != null;
   }
 
+  /** Returns true if a RC fetch should be made, false otherwise. */
+  private boolean shouldFetchAndActivateRemoteConfigValues() {
+    long currentTimeInMs = getCurrentSystemTimeMillis();
+    return hasAppStartConfigFetchDelayElapsed(currentTimeInMs)
+        && hasLastFetchBecomeStale(currentTimeInMs);
+  }
+
+  /**
+   * Delay fetch by some random time since app start. This is to prevent b/187985523.
+   *
+   * @return true if the random delay has elapsed, false otherwise
+   */
+  private boolean hasAppStartConfigFetchDelayElapsed(long currentTimeInMs) {
+    return (currentTimeInMs - appStartTimeInMs) >= appStartConfigFetchDelayInMs;
+  }
+
   // We want to fetch once when the app starts and every 12 hours after that.
   // The reason we maintain our own timestamps and do not use FRC's is because FRC only updates
   // the last successful fetch timestamp AFTER successfully fetching - which might mean that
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -292,7 +306,7 @@ private void triggerRemoteConfigFetchIfNecessary() {
     if (allRcConfigMap.isEmpty()) { // Initial fetch.
       syncConfigValues(firebaseRemoteConfig.getAll());
```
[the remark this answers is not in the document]
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -69,14 +73,24 @@ private RemoteConfigManager() {
         );
   }
 
-  @VisibleForTesting
   RemoteConfigManager(Executor executor, FirebaseRemoteConfig firebaseRemoteConfig) {
+    this(
+        executor,
+        firebaseRemoteConfig,
+        MIN_APP_START_CONFIG_FETCH_DELAY + new Random().nextInt(25000));
```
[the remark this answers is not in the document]
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -48,9 +49,12 @@
   private static final long TIME_AFTER_WHICH_A_FETCH_IS_CONSIDERED_STALE_MS =
       TimeUnit.HOURS.toMillis(12);
   private static final long FETCH_NEVER_HAPPENED_TIMESTAMP_MS = 0;
+  private static final long MIN_APP_START_CONFIG_FETCH_DELAY = 5000;
```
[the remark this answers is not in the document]
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -292,7 +306,7 @@ private void triggerRemoteConfigFetchIfNecessary() {
     if (allRcConfigMap.isEmpty()) { // Initial fetch.
       syncConfigValues(firebaseRemoteConfig.getAll());
     }
-    if (shouldFetchAndActivateRemoteConfigValues()) {
+    if (passesRandomAppStartDelay() && shouldFetchAndActivateRemoteConfigValues()) {
```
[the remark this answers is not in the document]
leotianlizhan: `triggerRemoteConfigFetchIfNecessary()` gets called every time a `getRemoteConfigValue(key)` gets called, which is every time we query something like `getFloat(key)`. From my understanding from reading the code, remote config is not fetched until `FirebaseRemoteConfig.fetch()` is called. Even the `firebaseRemoteConfig.getAll()` 1 line above doesn't fetch, it should just populate `allRcConfigMap` with default values. However I am not entirely sure on that because the comment does say "Initial fetch" after all (could you or Visu confirm my understanding?). So it should be ok to be called within 5 seconds, because there are default values. Whenever another query (like `getFloat(key)`) for an RC value happens after the app start delay, the entire `allRcConfigMap` will be updated.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -69,14 +74,25 @@ private RemoteConfigManager() {
...
+    this(
+        executor,
+        firebaseRemoteConfig,
+        MIN_APP_START_CONFIG_FETCH_DELAY_MS
+            + new Random().nextInt(RANDOM_APP_START_CONFIG_FETCH_DELAY_MS));
+  }
+
+  @VisibleForTesting
+  RemoteConfigManager(
+      Executor executor, FirebaseRemoteConfig firebaseRemoteConfig, long fetchDelay) {
```
visumickey, line 87: Lets call the new variable "appStartConfigFetchDelayInMs". Basically, we are looking at a delay from app start before starting to fetch. Lets have a name to represent that. The current name includes a notion that we are requesting a delay when this instance is created (This instance could get created much later than app start).

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -69,14 +74,25 @@ private RemoteConfigManager() {
...
     this.executor = executor;
     this.firebaseRemoteConfig = firebaseRemoteConfig;
     this.allRcConfigMap =
         firebaseRemoteConfig == null
             ? new ConcurrentHashMap<>()
             : new ConcurrentHashMap<>(firebaseRemoteConfig.getAll());
+    this.appStartTime = getCurrentSystemTimeMillis();
```
visumickey, line 94: Is this the app start time? Should this be the time at which the application's onCreate method got called?

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -48,9 +49,13 @@
   private static final long TIME_AFTER_WHICH_A_FETCH_IS_CONSIDERED_STALE_MS =
       TimeUnit.HOURS.toMillis(12);
   private static final long FETCH_NEVER_HAPPENED_TIMESTAMP_MS = 0;
+  private static final long MIN_APP_START_CONFIG_FETCH_DELAY_MS = 5000;
+  private static final int RANDOM_APP_START_CONFIG_FETCH_DELAY_MS = 25000;
 
   private final ConcurrentHashMap<String, FirebaseRemoteConfigValue> allRcConfigMap;
   private final Executor executor;
+  private final long appStartTime;
+  private final long appStartConfigFetchDelay;
```
visumickey, line 58: Can we suffix this with the cardinality of the variable? Eg. appStartConfigFetchDelayInMs? This would make the code more readable and makes it easy to understand for anyone working with the code in the future.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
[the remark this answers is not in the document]
visumickey: That is correct. Config fetch gets initiated only when a config lookup happens. Until then, all we do is to sync with the cache.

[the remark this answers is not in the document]
visumickey: I'm generally not too happy with the method name here. `passRandomAppStartDelay()`. The naming convention of method should be more towards making it read it like a real sentence in the context of the caller. For eg: `hasAppStartConfigFetchDelayElapsed()`.

There are a couple of suggestions I have.
1. Rename the method like I had mentioned before `hasAppStartConfigFetchDelayElapsed()`.
2. Consider making this method called inside `shouldFetchAndActivateRemoteConfigValues()`. The `shouldFetchAndActivate...` method already takes care of validating if a fetch can be made. This initial delay is also a part of the same process. So, consider putting them together.

I would prefer option 2 if that is easier to make.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -351,6 +372,10 @@ private boolean shouldFetchAndActivateRemoteConfigValues() {
         > TIME_AFTER_WHICH_A_FETCH_IS_CONSIDERED_STALE_MS;
   }
 
+  private boolean passesRandomAppStartDelay() {
```
visumickey, line 375: Though a private method, can we add some documentation about what we can expect from that method?

visumickey (commented):
Thanks for making this change @leotianlizhan. Glad to see the changes. I have few comments. PTAL as it permits.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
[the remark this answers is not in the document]
leotianlizhan: done using option 2

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -48,9 +49,13 @@
   private static final long TIME_AFTER_WHICH_A_FETCH_IS_CONSIDERED_STALE_MS =
       TimeUnit.HOURS.toMillis(12);
   private static final long FETCH_NEVER_HAPPENED_TIMESTAMP_MS = 0;
+  private static final long MIN_APP_START_CONFIG_FETCH_DELAY_MS = 5000;
+  private static final int RANDOM_APP_START_CONFIG_FETCH_DELAY_MS = 25000;
```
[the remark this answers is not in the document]
leotianlizhan: Changed to long. Also changed from using `new Random().nextInt()` to `ThreadLocalRandom.current().nextLong()` as per https://stackoverflow.com/a/363692

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
leotianlizhan: I did this originally because I was using your iOS PR as a reference, and I assumed that `+ (void)load` was equivalent to constructor, but now I realize it's not, my bad. 

Correct me if I'm wrong but I don't think we currently have a variable or constant anywhere for the application start time. 
Would it be good enough if I made `appStartTime` a static variable in `RemoteConfigManager` and initialize it as it is declared? That way the time should be when the class is initialized, which should be when `public static void getInstance()` is called for the first time (if you trace it backwards it should be called in `FirebasePerfProvider.attachInfo`.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
leotianlizhan: done

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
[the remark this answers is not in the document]
visumickey: Hmm.. Sounds good!

visumickey (approved):
This is great! Thanks for making the changes Leo.

How are you planning to test this? It is important to do some manual testing to make sure we are able fetch the configs in less than 30 seconds and greater than 5 seconds. If you are able to ascertain that, please feel free to merge this change.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
visumickey: This is a tricky one. There is a separate place where the app start time is measured. As much as a possible, we should see if we can rely on that for the start time of the application. This is the valid app start time you should be considering to use.

Please use this measurement instead of creating a new one in RemoveConfigManager.

leotianlizhan: found it and done

leotianlizhan:
> This is great! Thanks for making the changes Leo.
> 
> How are you planning to test this? It is important to do some manual testing to make sure we are able fetch the configs in less than 30 seconds and greater than 5 seconds. If you are able to ascertain that, please feel free to merge this change.

yes I've done some manual testing and it looks like it works

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
visumickey: Sounds good, thanks!

visumickey (approved):

PR #2972, opened 2021-09-13
FAD fixing file provider path bug

mannyjimenez0810:


pranavrajgopal (approved):

firebase-app-distribution/src/main/AndroidManifest.xml
```diff
@@ -44,12 +44,12 @@
 
         <provider
             android:name="androidx.core.content.FileProvider"
-            android:authorities="${applicationId}.provider"
+            android:authorities="${applicationId}.appdistro.fileprovider"
```
rachaprince, line 47: Sorry for the late review, but I think we should change this to `firebaseappdistribution.fileprovider` because we don't use `appdistro` externally, and the firebase prefix helps customers quickly identify this authority as part of the Firebase ecosystem (and aligns with the name of the main folder for this SDK)

firebase-app-distribution/src/main/AndroidManifest.xml
```diff
@@ -44,12 +44,12 @@
...
             android:exported="false"
             android:grantUriPermissions="true">
             <meta-data
                 android:name="android.support.FILE_PROVIDER_PATHS"
-                android:resource="@xml/provider_paths" />
+                android:resource="@xml/appdistro_provider_paths" />
```
rachaprince, line 52: same thing here, I think this should be `firebase_app_distribution_provider_path`

PR #2977, opened 2021-09-15
Add workflow to update Firebase C++ SDK deps when Android SDK "latest released version" changes

jonsimantov:
Whenever the released version of an Android SDK library changes (via latestReleasedVersion= changing in any library's gradle.properties), trigger the C++ SDK to do an Android dependency update, pulling latest deps from gmaven.

This workflow requests a token that has access to firebase-cpp-sdk and uses that token to trigger the C++ update-dependencies workflow.

(iOS equivalent is here: https://github.com/firebase/firebase-ios-sdk/pull/8654)


jonsimantov:
Note to reviewer: For this to work properly, it requires that the normal workflow be that this repo's "Update released versions" PRs are merged AFTER the bits are live for download on gmaven. Can you confirm that's the case?

Thanks!

vkryachko (approved):

PR #2979, opened 2021-09-15
Tighten the query bounds

schmidt-sebastian:
This PR implements type clamping. A query for foo < 3 should only return integers, so we know both a min value and a max value.


firebase-firestore/src/main/java/com/google/firebase/firestore/index/FirestoreIndexValueWriter.java
```diff
@@ -34,7 +34,7 @@
   public static final int INDEX_TYPE_BOOLEAN = 10;
   public static final int INDEX_TYPE_NAN = 13;
   public static final int INDEX_TYPE_NUMBER = 15;
-  public static final int INDEX_TYPE_TIMESTAMP = 10;
+  public static final int INDEX_TYPE_TIMESTAMP = 20;
```
schmidt-sebastian, line 37: This was a typo.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
```diff
@@ -432,4 +457,64 @@ public static Value refValue(DatabaseId databaseId, DocumentKey key) {
             .build();
     return value;
   }
+
+  public static Value getFirstValue(Value.ValueTypeCase valueTypeCase) {
```
thebrianchen, line 461: s/getFirstValue/getLowestValue?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
```diff
@@ -432,4 +457,64 @@ public static Value refValue(DatabaseId databaseId, DocumentKey key) {
...
+      case ARRAY_VALUE:
+        return Value.newBuilder().setArrayValue(ArrayValue.getDefaultInstance()).build();
+      case MAP_VALUE:
+        return Value.newBuilder().setMapValue(MapValue.getDefaultInstance()).build();
+      default:
+        throw new IllegalArgumentException("Unknown value type: " + valueTypeCase);
+    }
+  }
+
+  public static @Nullable Value getNextValue(Value.ValueTypeCase valueTypeCase) {
```
thebrianchen, line 493: s/getNextValue/getHighestValue or getHighestPossibleValue?

NextValue is a bit confusing here. Alternatively, just add some code comments please.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
```diff
@@ -143,14 +143,14 @@ public void testStartAfterFilter() {
   }
 
   @Test
-  public void testEndAtFilter() {
+  public void testEndBeforeFilter() {
```
thebrianchen, line 146: I'm a little confused about the `before` in `Bound`. Based on the naming of this test, if `before == true`, `Bound` does not include the `endAt` value? However, it seems like it's flipped for `startAt` and `startAfter`. 

I find a it harder to reason about before/after since it's flipped for whether it's a startAt, or endAt. Is it possible to change it `before` to an inclusive/exclusive model instead?

thebrianchen (approved):
lgtm with some small suggestions

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
schmidt-sebastian: Will rename.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
schmidt-sebastian: Looks like I forgot to push the commit with the comments. I agree that this is not a great name - however this method returns an exclusive value, i.e. one higher than the max. Do you have a name suggestion that conveys that? I need your input here :)

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
schmidt-sebastian: It's super confusing, but it uses the existing backend terminology:

startAt => before is true (cursor starts right before the provided value)
startAfter => before is false (cursor starts right after the provided value)
endBefore => before is false (cursor stops right before the provided value)
endAt => before is true (cursor stops right after the provided value)

I am confused by this myself, hence the test name here is wrong.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -126,20 +131,23 @@ public Bound getLowerBound(FieldIndex fieldIndex) {
     // Go through all filters to find a value for the current field segment
     for (FieldIndex.Segment segment : fieldIndex) {
       Value lowestValue = Values.NULL_VALUE;
+
       for (Filter filter : filters) {
         if (filter.getField().equals(segment.getFieldPath())) {
           FieldFilter fieldFilter = (FieldFilter) filter;
+
+          Value newValue = null;
```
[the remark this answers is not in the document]
schmidt-sebastian: Will update.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
thebrianchen: Some ideas: getHighestValueExclusive, getUpperBoundExclusive, getExclusiveUpperBound

I think pairing "exclusive" with a "bound" makes more sense. Or you can leave the current name and just have comments.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
thebrianchen: oh man this still doesn't make too much sense... Still can't wrap my mind around how endBefore has before set to false in order to mean that the cursor stops before??

Is this documented anywhere? optional nice thing you can do is encapsulate the before bound stuff into some tester objects :)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
schmidt-sebastian: Changed to getLowerBound() and getUpperBound(). The comments still state whether it is exclusive or not.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
schmidt-sebastian: Will clean up in a follow up.

PR #2980, opened 2021-09-15
Add X-Android-Package and X-Android-Cert headers.

rosalyntan:
Add `X-Android-Package` and `X-Android-Cert` headers to network calls.

[Google-internal bug: b/199533228]

### DIFF
--- a/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
+++ b/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
@@ -16,9 +16,14 @@ package com.google.firebase.appcheck.internal;
 
 import static com.google.android.gms.common.internal.Preconditions.checkNotNull;
 
+import android.content.Context;
+import android.content.pm.PackageManager.NameNotFoundException;
+import android.util.Log;
 import androidx.annotation.IntDef;
 import androidx.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
```diff
@@ -130,6 +145,23 @@ public AppCheckTokenResponse exchangeAttestationForAppCheckToken(
     }
   }
 
+  /** Gets the Android package's SHA-1 fingerprint. */
+  private String getFingerprintHashForPackage() {
+    byte[] hash;
+
+    try {
+      hash = AndroidUtilsLight.getPackageCertificateHashBytes(context, context.getPackageName());
```
vkryachko, line 153: nit: looks like this method is deprecated at head and `getNewestAvailablePackageCertificateRawBytes` is suggested instead, please check if it's ok to use instead.

rosalyntan: I tried updating the `com.google.android.gms:play-services-base` dependency to the latest version (17.6.0), but it looks like `getNewestAvailablePackageCertificateRawBytes` is not available (weirdly, only `getPackageCertificateHashBytes` is available, even though it's marked deprecated in the latest released version).

vkryachko (approved):

PR #2991, opened 2021-09-21
FAD Fixing file provider conflict

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionFileProvider.java
```diff
@@ -0,0 +1,5 @@
+package com.google.firebase.appdistribution;
```
rachaprince, line 1: I think this file needs the copy write text at the top

PR #2993, opened 2021-09-21
Remove special handling in Mutations

schmidt-sebastian:
I have been thinking more about our mutation code. I don't actually think there are any user visible mutations changes if a latency-compensated delete keeps the current version and has "localMutations" set. The only changes I see are in the local store tests, but the integration tests are unchanged.

Let me know if you can think of any problems.

This would further simplify https://github.com/firebase/firebase-android-sdk/pull/2962


schmidt-sebastian:
cc @thebrianchen

wu-hui (approved):
I can neither think of anything..it looks like this logic exists at the beginning of github repo, and now nobody knows the initial reason.

But if all tests are still passing, and we cannot think of anything..I think it is fine to delete it.

PR #2999, opened 2021-09-23
Fad logs file provider

mannyjimenez0810:
Correcting bug where the fileprovider name did not match with what the manifest had.


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -219,13 +217,17 @@ private String getFingerprintHashForPackage(Context context) {
       hash = AndroidUtilsLight.getPackageCertificateHashBytes(context, context.getPackageName());
 
       if (hash == null) {
-        Log.e(TAG, "Could not get fingerprint hash for package: " + context.getPackageName());
+        LogWrapper.getInstance()
```
rachaprince, line 220: I may have mispoke earlier. Can we remove this log and the log below it on line 225? These are the headers private we send to the OnePlatform API to validate the API Key, so I'm not sure we should log them. 

Instead, let's add a log wrapper around the hash log generated in the ReleaseIdentificationUtils

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -14,13 +14,11 @@
```
rachaprince, line 14: Can we log the id of newRelease in this file? Maybe around line 118?

rachaprince (changes requested):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionTesterApiClient.java
mannyjimenez0810: I have logged the apk hash for the release not sure if that is what you meant by id.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionTesterApiClient.java
mannyjimenez0810: The only problem is that logwrapper is package private. I have created a new logwrapper inside internal for the time being to be able to wrap the logs generated. It made sense to create a new log wrapper since its in a new package therefore the tag will be different being "FirebaseAppDistributionInternal" instead of "FirebaseAppDistribution"

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/LogWrapper.java
```diff
@@ -0,0 +1,58 @@
+// Copyright 2021 Google LLC
```
rachaprince, line 1: Rather than duplicating this class, I'd suggest making the Logger class public. Crashlytics and Fireperf both have public logger classes

rachaprince (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/ReleaseIdentificationUtils.java
```diff
@@ -90,20 +90,23 @@ public static String calculateApkHash(@NonNull File file) {
...
+                TAG
+                    + String.format(
+                        "Long id calculation time %d ms and %d entries for %s",
+                        elapsed, entries, file.getPath()));
       }
 
-      Log.v(TAG, String.format("Finished calculating %d entries in %d ms", entries, elapsed));
-      Log.v(TAG, String.format("%s hashes to %s", file.getPath(), zipFingerprint));
+      LogWrapper.getInstance()
+          .v(TAG + String.format("Finished calculating %d entries in %d ms" + entries, elapsed));
```
rachaprince, line 107: should this be a comma instead of a + sign?

mannyjimenez0810: YES! great catch

PR #3000, opened 2021-09-23
AppCheck for Firestore

ehsannas:


firebase-firestore/src/test/java/com/google/firebase/firestore/remote/MockDatastore.java
```diff
@@ -26,6 +27,7 @@
 import com.google.firebase.firestore.model.mutation.MutationResult;
 import com.google.firebase.firestore.remote.WatchChange.WatchTargetChange;
 import com.google.firebase.firestore.spec.SpecTestCase;
+import com.google.firebase.firestore.testutil.EmptyAppCheckTokenProvider;
```
schmidt-sebastian, line 30: I don't see this class in the PR, but I bet that we can combine EmptyAppCheckTokenProvider and EmptyCredentialsProvider.

ehsannas: oops. it seems that the newly-added files weren't added to my commit, sorry. I've added them now.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
```diff
@@ -48,13 +49,15 @@
...
             .add(Dependency.deferred(InternalAuthProvider.class))
+            .add(Dependency.deferred(InternalAppCheckTokenProvider.class))
             .add(Dependency.optional(FirebaseOptions.class))
             .factory(
                 c ->
                     new FirestoreMultiDbComponent(
                         c.get(Context.class),
                         c.get(FirebaseApp.class),
                         c.getDeferred(InternalAuthProvider.class),
+                        c.getDeferred(InternalAppCheckTokenProvider.class),
```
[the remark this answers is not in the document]
ehsannas: I'll try this

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/AppCheckTokenProvider.java
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.auth;
+
+/** A CredentialsProvider has a method to fetch an authorization token. */
```
schmidt-sebastian, line 17: Comment needs update.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,73 @@
...
+  public FirebaseAppCheckTokenProvider(
+      Deferred<InternalAppCheckTokenProvider> deferredAppCheckTokenProvider) {
+    deferredAppCheckTokenProvider.whenAvailable(
+        provider -> {
+          internalAppCheckTokenProvider = provider.get();
+          try {
+            Task<AppCheckTokenResult> pendingResult =
+                internalAppCheckTokenProvider.getToken(/* forceRefresh= */ false);
+            AppCheckTokenResult result =
+                Tasks.await(pendingResult, MAXIMUM_TOKEN_WAIT_TIME_MS, TimeUnit.MILLISECONDS);
```
schmidt-sebastian, line 55: 1) This looks like it might block app initialization.
2) Since `appCheckToken` is initialized asynchronously, I wonder if we even need to block. Can we just chain off the task?
3) We need to confirm that we retry with backoff if the AppCheck token is missing. Due to the initialization delay, we might initially send a request without an AppCheck token.
4) Does the access to AppCheck token need to be synchronized?

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/AppCheckTokenProvider.java
ehsannas: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: IIUC _if_ an app using AppCheck, we should avoid sending any requests to the backend without the AppCheck token. Otherwise the request would be considered malicious. So:

I have modified the initialization phase to asynchronously set the token when it is fetched.
I also updated the `getCurrentAppCheckToken` method to block and fetch a token if one is not already fetched. This could potentially block the very first time any grpc call is made (will not block if the asynchronous token setting during initialization is done before the first grpc call).

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,117 @@
...
+          }
+          Task<AppCheckTokenResult> pendingResult =
+              internalAppCheckTokenProvider.getToken(/* forceRefresh= */ false);
+          if (pendingResult == null) {
+            return;
+          }
+          pendingResult.addOnCompleteListener(
+              task -> {
+                synchronized (this) {
+                  if (appCheckToken != null) {
```
schmidt-sebastian, line 65: I am not sure if the new memory model in Android allows this callback to fire multiple times, but even if it does, it is probably not harmful to use the new token (which should be stable).

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,117 @@
...
+                    AppCheckTokenResult result = task.getResult();
+                    if (result.getError() != null) {
+                      Logger.warn(
+                          LOG_TAG,
+                          "Error getting App Check token; using placeholder token instead. Error: "
+                              + result.getError());
+                    }
+                    appCheckToken = result.getToken();
+                  } else {
+                    Logger.warn(
```
schmidt-sebastian, line 78: Should we clear the token?

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,117 @@
...
+        Logger.warn(
+            LOG_TAG,
+            "Error getting App Check token; using placeholder token instead. Error: "
+                + result.getError());
+      }
+      appCheckToken = result.getToken();
+    } catch (ExecutionException | InterruptedException | TimeoutException e) {
+      Logger.warn(LOG_TAG, "Unexpected error getting App Check token: " + e);
+    }
+    return appCheckToken;
```
schmidt-sebastian, line 115: If I understand this correctly, you are repeating the steps from the constructor here so you can block on the token result. You don't have to do do this - you should create a new instance variable that is just the Task. You can create a new task using [TaskCompletionSource](https://developers.google.com/android/reference/com/google/android/gms/tasks/TaskCompletionSource), which you can then fulfill in the constructor and block on here.

You should also not block in a synchronized method. The scope of your synchronization needs to be much narrower (but note that you won't have to use synchronization at all if you use Tasks).

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: Done. This was very useful. Thanks for your guidance.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,104 @@
...
+
+/** FirebaseAppCheckTokenProvider uses Firebase AppCheck to get an AppCheck token. */
+public final class FirebaseAppCheckTokenProvider extends AppCheckTokenProvider {
+
+  private static final String LOG_TAG = "FirebaseAppCheckTokenProvider";
+
+  private static final int MAXIMUM_TOKEN_WAIT_TIME_MS = 30000;
+
+  /** The AppCheck token string. May be null if the token has not been retrieved yet. */
+  @Nullable private String appCheckToken;
```
schmidt-sebastian, line 37: You should not need this. You can always just get the token result. 

If you do keep using it, you would need to lock any access.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,104 @@
...
+
+    deferredAppCheckTokenProvider.whenAvailable(
+        provider -> {
+          InternalAppCheckTokenProvider internalAppCheckTokenProvider = provider.get();
+          if (internalAppCheckTokenProvider == null) {
+            return;
+          }
+          Task<AppCheckTokenResult> pendingResult =
+              internalAppCheckTokenProvider.getToken(/* forceRefresh= */ false);
+          if (pendingResult == null) {
```
schmidt-sebastian, line 60: This should never be null.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,104 @@
...
+            return;
+          }
+          TaskCompletionSource<String> taskCompletionSource = new TaskCompletionSource<>();
+          getTokenTask = taskCompletionSource.getTask();
```
schmidt-sebastian, line 64: You can set this right as you enter the constructor. Then `getTokenTask` becomes non-null and final.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Nullable
+  public String getCurrentAppCheckToken() {
+    if (appCheckToken != null) {
+      return appCheckToken;
+    }
+
+    if (getTokenTask == null) {
+      return null;
+    }
```
schmidt-sebastian, line 94: These should not be needed. Note that calling `Tasks.await()` on a completed task has almost no overhead - https://source.corp.google.com/piper///depot/google3/java/com/google/android/gmscore/integ/client/tasks/src/com/google/android/gms/tasks/Tasks.java;l=198

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: Good point. Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
ehsannas: Done.

ehsannas:
I'm seeing a test failure (`testRestartFirestoreLeadsToNewInstance`) because a timeout occurs [here](https://github.com/firebase/firebase-android-sdk/blob/ac3276ce7d5f5181b1c419731495eb88a98537af/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java#L1107).

The reason is that the [callback function](https://github.com/firebase/firebase-android-sdk/blob/e8c5cf5025366a1c5943e04c19cbe9683d68fca8/firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java#L50) given to `deferredAppCheckTokenProvider.whenAvailable` is never triggered. Trying to figure out why.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,89 @@
...
+  /** The Task for retrieving the AppCheck token. */
+  private final Task<String> getTokenTask;
+
+  /** Creates a new FirebaseAppCheckTokenProvider. */
+  public FirebaseAppCheckTokenProvider(
+      Deferred<InternalAppCheckTokenProvider> deferredAppCheckTokenProvider) {
+    TaskCompletionSource<String> taskCompletionSource = new TaskCompletionSource<>();
+    getTokenTask = taskCompletionSource.getTask();
+
+    if (deferredAppCheckTokenProvider == null) {
```
schmidt-sebastian, line 45: This should also never be null, should it? If you think it should in fact be nullable please make the argument `@Nullable`.

firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAppCheckTokenProvider.java
```diff
@@ -0,0 +1,89 @@
...
+      taskCompletionSource.setResult(null);
+      return;
+    }
+
+    deferredAppCheckTokenProvider.whenAvailable(
+        provider -> {
+          InternalAppCheckTokenProvider internalAppCheckTokenProvider = provider.get();
+          if (internalAppCheckTokenProvider == null) {
```
schmidt-sebastian, line 53: This also does not seem to be nullable. I checked the documentation and other callsites.

schmidt-sebastian (commented):
I think the test is revealing a real problem. When AppCheck is not available, we always wait  30s. I am not sure what to do about that - I think we should follow up in the email thread.

PR #3001, opened 2021-09-23
Write index entries on backfill

thebrianchen:


schmidt-sebastian:
cc @wu-hui

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -103,13 +131,105 @@ public BackfillScheduler newScheduler(AsyncQueue asyncQueue, LocalStore localSto
...
+    // TODO(indexing): Track the number of entries written and process more field indexes if
+    // we're under the maximum.
+    int lastArrayIndex = Math.min(maxFieldIndexesToProcess, fieldIndexQueue.size());
+    List<FieldIndex> fieldIndexesToProcess = fieldIndexQueue.subList(0, lastArrayIndex);
+
+    // Create a map of each collection group to all the FieldIndexes on it.
+    Map<String, ArrayList<FieldIndex>> collectionToFieldIndexes = new HashMap<>();
+
+    // Map of field index to the most recent snapshot version written.
+    Map<FieldIndex, SnapshotVersion> fieldIndexToNewestVersion = new HashMap<>();
```
schmidt-sebastian, line 169: Do we need to store the snapshot version separately? The FIeldIndex already has a snapshot version which we could just update. This map would then likely change to `Map<IndexId, FieldIndex>`.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -103,13 +131,105 @@ public BackfillScheduler newScheduler(AsyncQueue asyncQueue, LocalStore localSto
...
+
+    for (FieldIndex fieldIndex : fieldIndexesToProcess) {
+
+      // Use the lowest version number as the base.
+      lastReadTime =
+          lastReadTime != null
+              ? fieldIndex.getVersion().compareTo(lastReadTime) < 0
+                  ? fieldIndex.getVersion()
+                  : lastReadTime
+              : fieldIndex.getVersion();
```
schmidt-sebastian, line 179: This can likely be simplified if you directly update the FieldIndex version.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -103,13 +131,105 @@ public BackfillScheduler newScheduler(AsyncQueue asyncQueue, LocalStore localSto
...
+      String collectionGroup = fieldIndex.getCollectionGroup();
+
+      if (collectionToFieldIndexes.containsKey(collectionGroup)) {
```
schmidt-sebastian, line 182: Should we add an API to IndexManager that returns all the collection groups that have active indices? We could rewrite the logic as such:

```
List<String> collectionGroups = indexManger.getCollectionGroupsWIthIndices();
Collections.shuffle(collectionGroups): // Don't always want to start with the same group
for (String collectionGroup: collectionGroups)  {
    List<FieldIndex> indices = indexManager.getIndices(collectionGroup);
    ImmutableSortedMap<DocumentKey, Document> matchingDocuments = ...
    
    // add index
    // Break after x number of docs. 
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -133,35 +153,51 @@ public void addIndexEntries(Document document) {
...
-                      encoded,
-                      documentKey.toString());
-                }
               } catch (InvalidProtocolBufferException e) {
                 throw fail("Invalid index: " + e);
               }
             });
   }
 
+  public int addIndexEntry(
```
schmidt-sebastian, line 164: Please add JavaDoc.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -133,35 +153,51 @@ public void addIndexEntries(Document document) {
...
+      Document document,
+      Collection<FieldIndex> fieldIndexes,
+      @Nullable Map<FieldIndex, SnapshotVersion> fieldIndexToSnapshotVersion) {
```
schmidt-sebastian, line 167: You can likely remove this map too if you update the version in the FieldIndex direclty.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerTest.java
```diff
@@ -49,12 +58,20 @@ public static void afterClass() {
...
   @Before
   public void setUp() {
     persistence = PersistenceTestHelpers.createSQLitePersistence();
+    indexManager = (SQLiteIndexManager) persistence.getIndexManager();
     backfiller = persistence.getIndexBackfiller();
+    CountingQueryEngine queryEngine = new CountingQueryEngine(new DefaultQueryEngine());
+    localStore = new LocalStore(persistence, queryEngine, User.UNAUTHENTICATED);
+
+    //    writeIndexConfigs();
+    //    addInitialTestDocuments();
```
schmidt-sebastian, line 74: Please remove.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerTest.java
```diff
@@ -81,4 +165,16 @@ public void addAndRemoveIndexEntry() {
           assertNull(entry);
         });
   }
+
+  void addFieldIndex(String collectionId, String fieldName) {
+    indexManager.addFieldIndex(
```
schmidt-sebastian, line 170: Does this need to run inside a transaction?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerTest.java
```diff
@@ -81,4 +165,16 @@ public void addAndRemoveIndexEntry() {
...
+        new FieldIndex(collectionId)
+            .withAddedField(field(fieldName), FieldIndex.Segment.Kind.ORDERED));
+  }
+
+  private void addDocumentToCache(String path, String field, SnapshotVersion readTime) {
```
schmidt-sebastian, line 175: Nit: This is just called "addDoc" in other places.

schmidt-sebastian (commented):
Thanks! I think we can simplify this a bit. Let me know if you agree.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
thebrianchen: Removed.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
thebrianchen: done.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -103,13 +131,105 @@ public BackfillScheduler newScheduler(AsyncQueue asyncQueue, LocalStore localSto
...
+    // TODO(indexing): Handle field indexes that are removed by the user.
+    fetchNewFieldIndexes();
+
+    int numIndexesWritten = writeIndexEntries(localStore);
     int numIndexesRemoved = 0;
     return new Results(/* hasRun= */ true, numIndexesWritten, numIndexesRemoved);
   }
 
+  private void fetchNewFieldIndexes() {
+    List<FieldIndex> fieldIndexes = indexManager.getFieldIndexes(lastFieldIndexId + 1);
```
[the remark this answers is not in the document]
thebrianchen: Storing the indexes in memory on the index manager would be nice. Added a TODO to add that in a separate PR. 

However, a separate issue is how the IndexBackfiller can still maintain the queue ordering. Currently the index backfiller moves field indexes that it has finished processing to the back of the queue. Even if we get all the indexes directly from the index manager, we'll still need to get the indexes that we've already processed (unless we use a random approach).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -103,13 +131,105 @@ public BackfillScheduler newScheduler(AsyncQueue asyncQueue, LocalStore localSto
...
+
+  /** Writes index entries based on the FieldIndexQueue. Returns the number of entries written. */
+  private int writeIndexEntries(LocalStore localStore) {
+    int numIndexesWritten = 0;
+    SnapshotVersion lastReadTime = null;
+
+    // TODO(indexing): Track the number of entries written and process more field indexes if
+    // we're under the maximum.
+    int lastArrayIndex = Math.min(maxFieldIndexesToProcess, fieldIndexQueue.size());
+    List<FieldIndex> fieldIndexesToProcess = fieldIndexQueue.subList(0, lastArrayIndex);
```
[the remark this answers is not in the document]
thebrianchen: Can you expand on that? If we have fewer documents than index entries, wouldn't it make it sense to cap on index entries?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: Done. We can remove the map completely in this case since we already have the list of field indexes that we processed here.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: I realize that since we're processing the collection groups in order, we can move the logic down to fetch the lowest version for all fieldindexes mapped to a collection group. 

For example, if we have multiple field indexes that map to a collection group, we can load documents starting from the lowest snapshot version present on the field indexes.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
thebrianchen: This is a viable option, but getting collection groups in random to fill could result in certain field indexes to not be backfilled for a while. Alternatively, using a queue implementation doesn't require that much more overhead, and allows backfiller to make incremental progress each time.

When a user removes an index, does it become inactive? What happens if the user adds the same index again?

thebrianchen (commented):
thanks for the review!

schmidt-sebastian:
Summary from offline discussion:
- We store the "last index update time" on the collection groups table
- We process the collection groups in order of update time, oldest first
- We fetch the documents from the collection group in order of read time
- Once we read a document, we always add all index entries. This avoids re-reading documents during the next index update.
- Once we hit the cap, we advance the update time on the collection group table

PR #3002, opened 2021-09-24
Fix IN filters

schmidt-sebastian:
The current implementation for IN filters has a bug. 

For `where('foo', 'in', [1,3])` we currently run:

```
foo >= 1 && foo <= 1 OR foo >= 1 && foo <= 3 OR foo >= 3 && foo <= 1 OR foo >= 3 && foo <= 3
```

This also returns "2".

We should only run:

```
foo >= 1 && foo <= 1 OR foo >= 3 && foo <= 3
```


schmidt-sebastian:
cc @thebrianchen

wu-hui (approved):

PR #3007, opened 2021-09-27
DocumentOverlays schema and scaffolding

wu-hui:
Creates a new SQLite table "document_overlays" to save overlays for documents that are modified locally. An overlay is just a `Mutation`.

### DIFF
--- a/firebase-firestore/firebase-firestore.gradle
+++ b/firebase-firestore/firebase-firestore.gradle
@@ -106,12 +106,14 @@ android.libraryVariants.all { variant ->
     } catch (FileNotFoundException e) {
     }
 
+    // TODO(Indexing): Delete below once indexing is shipped.
     if (localProps['firestoreEnableIndexing']) {
         variant.buildConfigField("boolean", "ENABLE_INDEXING", "true")
     } else {
         variant.buildConfigField("boolean", "ENABLE_INDEXING", "false")
     }
 
+    // TODO(Overlay): Delete below once overlay is shipped.
     if (localProps['firestoreEnableOverlay']) {
         variant.buildConfigField("boolean", "ENABLE_OVERLAY", "true")
     } else {
--- a/firebase-firestore/firebase-firestore.gradle
+++ b/firebase-firestore/firebase-firestore.gradle
@@ -111,6 +111,12 @@ android.libraryVariants.all { variant ->
     } else {
         variant.buildConfigField("boolean", "ENABLE_INDEXING", "false")
     }
+
+    if (localProps['firestoreEnableOverlay']) {
+        variant.buildConfigField("boolean", "ENABLE_OVERLAY", "true")
+    } else {
+        variant.buildConfigField("boolean", "ENABLE_OVERLAY", "false")
+    }
 }
 
 configurations.all {
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
@@ -255,7 +255,8 @@ public class IntegrationTestUtil {
     // This unfortunately is a global setting that affects existing Firestore clients.
     Logger.setLogLevel(logLevel);
 
-    // TODO: Remove this once this is ready to ship.
+    // TODO: Remove below once this is ready to ship.
+    Persistence.OVERLAY_SUPPORT_ENABLED = true;
     Persistence.INDEXING_SUPPORT_ENABLED = true;
 
     Context context = ApplicationProvider.getApplicationContext();
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
@@ -255,7 +255,7 @@ public class IntegrationTestUtil {
     // This unfortunately is a global setting that affects existing Firestore clients.
     Logger.setLogLevel(logLevel);
 
-    // TODO: Remove below once this is ready to ship.
+    // TODO(Overlay): Remove below once this is ready to ship.
     Persistence.OVERLAY_SUPPORT_ENABLED = true;
     Persistence.INDEXING_SUPPORT_ENABLED = true;
 
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlay.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlay.java
@@ -21,20 +21,20 @@ import com.google.firebase.firestore.model.mutation.Mutation;
 /**
  * Provides methods to read and write document overlays.
  *
- * <p>An overlay is a saved {@code Mutation}, that gives a local view of a document, when applied to
+ * <p>An overlay is a saved {@code Mutation}, that gives a local view of a document when applied to
  * the remote version of the document.
  */
-public interface DocumentOverlays {
+public interface DocumentOverlay {
   /**
    * Gets the saved overlay mutation for the given user and the given document key. Returns null if
    * there is no overlay for the user and document combination.
    */
   @Nullable
-  Mutation getOverlayMutation(String uid, DocumentKey key);
+  Mutation getOverlay(DocumentKey key);
 
   /** Saves the given mutation as overlay for the user and document key. */
-  void saveOverlayMutation(String uid, DocumentKey key, Mutation mutation);
+  void saveOverlay(DocumentKey key, Mutation mutation);
 
   /** Removes the overlay associated for the given user and document key. */
-  void removeOverlayMutation(String uid, DocumentKey key);
+  void removeOverlay(DocumentKey key);
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlay.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlay.java
@@ -21,20 +21,20 @@ import com.google.firebase.firestore.model.mutation.Mutation;
 /**
  * Provides methods to read and write document overlays.
  *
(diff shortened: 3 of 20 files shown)


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java
```diff
@@ -255,7 +255,8 @@ public static FirebaseFirestore testFirestore(
     // This unfortunately is a global setting that affects existing Firestore clients.
     Logger.setLogLevel(logLevel);
 
-    // TODO: Remove this once this is ready to ship.
+    // TODO: Remove below once this is ready to ship.
```
schmidt-sebastian, line 258: Please use "TODO(Overlay)"

firebase-firestore/firebase-firestore.gradle
```diff
@@ -111,6 +111,12 @@ android.libraryVariants.all { variant ->
     } else {
         variant.buildConfigField("boolean", "ENABLE_INDEXING", "false")
     }
+
+    if (localProps['firestoreEnableOverlay']) {
```
schmidt-sebastian, line 115: Should we add a TODO to remove this in the future?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlays.java
```diff
@@ -0,0 +1,40 @@
...
+package com.google.firebase.firestore.local;
+
+import androidx.annotation.Nullable;
+import com.google.firebase.firestore.model.DocumentKey;
+import com.google.firebase.firestore.model.mutation.Mutation;
+
+/**
+ * Provides methods to read and write document overlays.
+ *
+ * <p>An overlay is a saved {@code Mutation}, that gives a local view of a document, when applied to
```
schmidt-sebastian, line 24: Remove comma before "when".

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlays.java
```diff
@@ -0,0 +1,40 @@
...
+ * the remote version of the document.
+ */
+public interface DocumentOverlays {
```
schmidt-sebastian, line 27: Classes like this should be in singular form. Java uses plural form for classes with a bunch of static helpers (Utils.java, Helpers.java, Values.java, Collections.java)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlays.java
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Gets the saved overlay mutation for the given user and the given document key. Returns null if
+   * there is no overlay for the user and document combination.
+   */
+  @Nullable
+  Mutation getOverlayMutation(String uid, DocumentKey key);
```
schmidt-sebastian, line 33: Please use higher level User object. It makes for a more obvious API (but also see comment about dropping this argument).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlays.java
```diff
@@ -0,0 +1,54 @@
...
+
+package com.google.firebase.firestore.local;
+
+import androidx.annotation.Nullable;
+import com.google.firebase.firestore.model.DocumentKey;
+import com.google.firebase.firestore.model.mutation.Mutation;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MemoryDocumentOverlays implements DocumentOverlays {
```
schmidt-sebastian, line 23: Random naming thought (feel free to ignore): I think `MemoryOverlay` would read nicer.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlays.java
```diff
@@ -0,0 +1,54 @@
...
+  private Map<String, Map<DocumentKey, Mutation>> overlays = new HashMap<>();
+
+  @Nullable
+  @Override
+  public Mutation getOverlayMutation(String uid, DocumentKey key) {
```
schmidt-sebastian, line 28: The way UserIds are handled in the MutationQueue is different. We have a single MutationQueue per uses.  That means that these methods wouldn't need to take the User ID at all.

It also means that you wold not have to implement user switching in this class. Overlays should be destroyed once a user signs out, and not be available once the user signs back in again (to match current Memory-persistence behavior). I noticed that you did not implement User ID switching - so you are already on the right track :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/Persistence.java
```diff
@@ -50,8 +50,12 @@
 public abstract class Persistence {
   static final String TAG = Persistence.class.getSimpleName();
 
+  /** Temporary setting for enabling document overlays. */
+  // TODO(Overlay): Remove this.
+  public static boolean OVERLAY_SUPPORT_ENABLED = BuildConfig.ENABLE_OVERLAY;
+
   /** Temporary setting for enabling indexing-specific code paths while in development. */
-  // TODO: Remove this.
+  // TODO(Indexing): Remove this.
```
schmidt-sebastian, line 58: Thanks

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlays.java
schmidt-sebastian: Maybe also s/getOverlayMutation/getOverlay (if you do do this, then here and below)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlays.java
```diff
@@ -0,0 +1,70 @@
...
+    String path = EncodedPath.encode(key.getPath());
+    return db.query("SELECT overlay_mutation FROM document_overlays WHERE uid = ? AND path = ?")
+        .binding(uid, path)
+        .firstValue(
+            row -> {
+              if (row != null) {
+                try {
+                  return serializer.decodeMutation(row.getBlob(0));
+                } catch (InvalidProtocolBufferException e) {
+                  throw fail("Overlay mutations failed to parse: %s", e);
```
schmidt-sebastian, line 45:
```suggestion
                  throw fail("Overlay failed to parse: %s", e);
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -51,8 +51,10 @@
    */
   static final int VERSION = 12;
 
+  static final int OVERLAY_SUPPORT_VERSION = VERSION + 1;
+
   // TODO(indexing): Remove this constant and increment VERSION to enable indexing support
-  static final int INDEXING_SUPPORT_VERSION = VERSION + 1;
+  static final int INDEXING_SUPPORT_VERSION = VERSION + 2;
```
schmidt-sebastian, line 57:
```suggestion
  static final int INDEXING_SUPPORT_VERSION = OVERLAY_SUPPORT_VERSION + 1;
```

firebase-firestore/src/test/java/com/google/firebase/firestore/local/DocumentOverlaysTestCase.java
```diff
@@ -0,0 +1,119 @@
...
+  @Test
+  public void testCanReadSavedOverlay() {
+    Mutation m = patchMutation("coll/doc1", map("foo", "bar"));
+    overlays.saveOverlayMutation("Alice", key("coll/doc1"), m);
+
+    assertEquals(m, overlays.getOverlayMutation("Alice", key("coll/doc1")));
+  }
+
+  @Test
+  public void testCannotReadSavedOverlayByOtherUser() {
```
schmidt-sebastian, line 87: Note: This test would not be relevant once we move the User ID handling to Persistence.java

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlay.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Provides methods to read and write document overlays.
+ *
+ * <p>An overlay is a saved {@code Mutation}, that gives a local view of a document when applied to
+ * the remote version of the document.
+ */
+public interface DocumentOverlay {
+  /**
+   * Gets the saved overlay mutation for the given user and the given document key. Returns null if
```
schmidt-sebastian, line 29: Needs update (remove mention of user)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
```diff
@@ -326,4 +327,13 @@ public FieldIndex decodeFieldIndex(
...
   }
+
+  public Mutation decodeMutation(byte[] blob) throws InvalidProtocolBufferException {
+    Write mutation = Write.parseFrom(blob);
+    return rpcSerializer.decodeMutation(mutation);
+  }
+
+  public byte[] encodeMutation(Mutation mutation) {
+    return rpcSerializer.encodeMutation(mutation).toByteArray();
+  }
```
schmidt-sebastian, line 338: I just scrolled through the file and noticed that none of the other methods here encode and decode to and from bytes. Rather, this is done at the callsite. This actually has some benefit as this "detaches" this encoder from the underlying storage layer. While SQLite uses `byte[]`, the memory implementation uses the Proto directly. With that, is it possible to convert to bytes in the SQLite implementation of the document overlay?

schmidt-sebastian (approved):
Thanks for addressing the feedback. Some final thoughts, but looks good.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalSerializer.java
wu-hui: Makes sense.

PR #3008, opened 2021-09-28
Move app start trace build logic into a non-main thread.

jeremyjiang-dev:
Summary:
- This PR moves the [app start trace build logic](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java#L203-L233) into another thread. 
- The rest of the PR is about formatting old code.

This effort will cut down the trace build time during app startup from ~30ms to a few ms.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
@@ -25,6 +25,7 @@ import androidx.annotation.Nullable;
 import com.google.android.gms.common.util.VisibleForTesting;
 import com.google.firebase.perf.logging.AndroidLogger;
 import com.google.firebase.perf.provider.FirebasePerfProvider;
+import com.google.firebase.perf.session.PerfSession;
 import com.google.firebase.perf.session.SessionManager;
 import com.google.firebase.perf.transport.TransportManager;
 import com.google.firebase.perf.util.Clock;
@@ -64,6 +65,33 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
 
   private static volatile AppStartTrace instance;
   private static ExecutorService executorService;
+
+  private boolean isRegisteredForLifecycleCallbacks = false;
+  private final TransportManager transportManager;
+  private final Clock clock;
+  private Context appContext;
+  /**
+   * The first time onCreate() of any activity is called, the activity is saved as launchActivity.
+   */
+  private WeakReference<Activity> launchActivity;
+  /**
+   * The first time onResume() of any activity is called, the activity is saved as appStartActivity
+   */
+  private WeakReference<Activity> appStartActivity;
+
+  /**
+   * If the time difference between app starts and creation of any Activity is larger than
+   * MAX_LATENCY_BEFORE_UI_INIT, set mTooLateToInitUI to true and we don't send AppStart Trace.
+   */
+  private boolean isTooLateToInitUI = false;
+
+  private Timer appStartTime = null;
+  private Timer onCreateTime = null;
+  private Timer onStartTime = null;
+  private Timer onResumeTime = null;
+
+  private PerfSession startSession;
+  private boolean isStartedFromBackground = false;
 
   /**
    * Called from onCreate() method of an activity by instrumented byte code.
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -191,51 +213,58 @@ public synchronized void onActivityResumed(Activity activity) {
...
+    traceMetricBuilder
+        .setName(Constants.TraceNames.ON_RESUME_TRACE_NAME.toString())
         .setClientStartTimeUs(onStartTime.getMicros())
         .setDurationUs(onStartTime.getDurationMicros(onResumeTime));
-    subtraces.add(temp.build());
+    subtraces.add(traceMetricBuilder.build());
 
     metric
         .addAllSubtraces(subtraces)
         .addPerfSessions(SessionManager.getInstance().perfSession().build());
```
leotianlizhan, line 265: is this thread-safe? Would it make sense to add `synchronized` to `SessionManager.getInstance()`? But `transportManager.log` 2 lines down also calls `SessionManager.getInstance()` in a different thread already, so I'm not sure if it's already thread-safe.

jeremyjiang-dev: Good catch! I stored SessionManager.getInstance().perfSession() into a class variable.

visumickey:
@jeremyjiang-dev Can you add some details on what your PR does? The description of the PR shows the outcomes and not so much on what has changed.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -94,41 +130,29 @@ static AppStartTrace getInstance(TransportManager transportManager, Clock clock)
...
+          instance =
+              new AppStartTrace(
+                  transportManager,
+                  clock,
+                  new ThreadPoolExecutor(
+                      CORE_POOL_SIZE,
+                      MAX_POOL_SIZE,
+                      /* keepAliveTime= */ 10,
+                      TimeUnit.SECONDS,
+                      new LinkedBlockingQueue<>()));
```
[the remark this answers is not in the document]
jeremyjiang-dev: Great suggestions! A bounded LinkedBlockingQueue could save some memory since the capacity would be Integer#MAX_VALUE if not specified. I set the capacity to 1 because we will log _as once for an application run.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -94,41 +130,29 @@ static AppStartTrace getInstance(TransportManager transportManager, Clock clock)
...
         if (instance == null) {
-          instance = new AppStartTrace(transportManager, clock);
+          instance =
+              new AppStartTrace(
+                  transportManager,
+                  clock,
+                  new ThreadPoolExecutor(
+                      CORE_POOL_SIZE,
+                      MAX_POOL_SIZE,
+                      /* keepAliveTime= */ 10,
```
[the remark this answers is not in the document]
jeremyjiang-dev: Good catch! Done.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -191,51 +215,57 @@ public synchronized void onActivityResumed(Activity activity) {
...
-                + startTime.getDurationMicros(onResumeTime)
+                + this.appStartTime.getDurationMicros(onResumeTime)
                 + " microseconds");
 
+    // Log the app start trace in a non-main thread.
+    executorService.execute(this::logAppStartTrace);
+
+    if (isRegisteredForLifecycleCallbacks) {
+      // After AppStart trace is logged, we can unregister this callback.
+      unregisterActivityLifecycleCallbacks();
```
visumickey, line 233: Just out of curiousity - Can this happen on non-main thread? Just thinking if we can delay this for later on a non-main thread.

visumickey (approved):
Please make sure to test your changes to see a successful logging of the app start trace.

🚀

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
jeremyjiang-dev: We cannot. If we delay this, it is possible that the app enters onActivityResumed multiple times before the executor executes so that we might log the startup time multiple times.

jeremyjiang-dev:
> Please make sure to test your changes to see a successful logging of the app start trace.
> 
> 🚀

Thanks for reviewing the PR! I have tested and saw the logging of the app start trace.

leotianlizhan (approved):
LGTM

PR #3016, opened 2021-09-30
Draft: AppCheck callback

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -506,7 +517,7 @@ private void handleWatchStreamClose(Status status) {
   public boolean canUseNetwork() {
     // PORTING NOTE: This method exists mostly because web also has to take into account primary
     // vs. secondary state.
-    return networkEnabled;
+    return networkEnabled && appCheckToken != null;
```
ehsannas, line 520: Not sure if I understand this :( 
If the user's application does not use the AppCheck module, then `appCheckToken` will always be null and we can never use the network.

schmidt-sebastian: My thought on this was that we could use an empty token here. If AppCheck is not available, the listener fires once with an empty token.

PR #3017, opened 2021-09-30
Use more of the backend's ordering implementation

schmidt-sebastian:
This PR adds a bunch of test cases that verify the behavior of indexing when compared with a source of truth (the Java SDK - https://gist.github.com/schmidt-sebastian/4eab4300fa5d36d6c2eaefb6adc17b13).

To make these tests work, a bunch of things change:

- Ordering is now strictly enforced. The index entries are ordered by the direction specified in the index. This means that there are now two index formats (ascending, descending) instead of just one for directional queries.
- Like the backend, the SDK now only supports a single ArrayContains/ArrayContainsAny filter (for cosmetic rather than semantic reasons, can be lifted if the backend lifts its limit)
- Lower bounds can now be nullable too as they essentially act as upper bounds when the direction is DESCENDING
- All variable length index values are now terminated by a truncation indicator (which is always NO_TRUNCATION since the SDK does not truncate)
- All index values are separated by an INFINITY separator (as discussed in http://goto.google.com/firestore-storage-format#encodings)


schmidt-sebastian:
cc @thebrianchen

PR #3018, opened 2021-10-01
Fireperf: remove the redundant PerfSession.create during cold start

leotianlizhan:
b/201549215

Best solution would be to not create a `PerfSession` in `SessionManager`'s constructor. But worried if `SessionManager` is garbage collected, would it be recreated without a `PerfSession`? 

Basically the question is: does Android garbage collect singletons (without killing entire app process)? I think not, I think Android will just kill the entire App process, which wouldn't be a problem. But I couldn't be conclusive.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java
@@ -53,6 +53,8 @@ public class SessionManager extends AppStateUpdateHandler {
 
   private SessionManager() {
     this(GaugeManager.getInstance(), PerfSession.create(), AppStateMonitor.getInstance());
+    // Solution 0
+    this(GaugeManager.getInstance(), null, AppStateMonitor.getInstance());
   }
 
   @VisibleForTesting
@@ -106,21 +108,33 @@ public class SessionManager extends AppStateUpdateHandler {
    * @see PerfSession#isVerbose()
    */
   public void updatePerfSession(ApplicationProcessState currentAppState) {
-    synchronized (clients) {
-      perfSession = PerfSession.create();
+    // Solution 1
+    if (!appStateMonitor.isColdStart()) {
+      synchronized (clients) {
+        perfSession = PerfSession.create();
 
-      for (Iterator<WeakReference<SessionAwareObject>> i = clients.iterator(); i.hasNext(); ) {
-        SessionAwareObject callback = i.next().get();
-        if (callback != null) {
-          callback.updateSession(perfSession);
-        } else {
-          // The object pointing by WeakReference has already been garbage collected.
-          // Remove it from the Set.
-          i.remove();
+        for (Iterator<WeakReference<SessionAwareObject>> i = clients.iterator(); i.hasNext(); ) {
+          SessionAwareObject callback = i.next().get();
+          if (callback != null) {
+            callback.updateSession(perfSession);
+          } else {
+            // The object pointing by WeakReference has already been garbage collected.
+            // Remove it from the Set.
+            i.remove();
+          }
         }
       }
     }
 
+    logGaugeMetadataIfCollectionEnabled(currentAppState);
+    startOrStopCollectingGauges(currentAppState);
+
+    // Solution 2
+    logMetadataAndStartOrStopCollectingGauges(currentAppState);
+  }
+
+  // Solution 2
+  public void logMetadataAndStartOrStopCollectingGauges(ApplicationProcessState currentAppState) {
     logGaugeMetadataIfCollectionEnabled(currentAppState);
     startOrStopCollectingGauges(currentAppState);
   }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/provider/FirebasePerfProvider.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/provider/FirebasePerfProvider.java
@@ -76,7 +76,10 @@ public class FirebasePerfProvider extends ContentProvider {
     // possible.
     // There is code in SessionManager that prevents us from resetting the session twice in case
     // of app cold start.
+    // Solution 1
     SessionManager.getInstance().updatePerfSession(ApplicationProcessState.FOREGROUND);
+    // Solution 2
+    SessionManager.getInstance().logMetadataAndStartOrStopCollectingGauges(ApplicationProcessState.FOREGROUND);
   }
 
   /** Called before {@link Application#onCreate()}. */


firebase-perf/src/test/java/com/google/firebase/perf/metrics/AppStartTraceTest.java
```diff
@@ -228,14 +229,24 @@ public void testStartFromBackground() {
   }
 
   @Test
-  public void testFirebasePerfProviderOnAttachInfoUpdatesPerfSession() {
+  public void testFirebasePerfProviderOnAttachInfo_initializesGaugeCollection() {
+    com.google.firebase.perf.session.PerfSession originalPerfSession =
```
jeremyjiang-dev, line 233: Why do you need to save the originalPerfSession and set it in the end?

leotianlizhan: Otherwise `AppStartTraceTest.testLaunchActivity()` would fail because `trace.onActivityResumed` calls `SessionManager.getInstance().perfSession().build()`, which needs a real perfSession.

jeremyjiang-dev: Unit tests should not depend on each other. Could you create a clearInstance() method so that every test will initialize a new sessionManager
Reference: https://github.com/firebase/firebase-android-sdk/blob/59e027ae2943d51e061f0fd6bf6208fd5e895267/firebase-perf/src/test/java/com/google/firebase/perf/application/AppStateMonitorTest.java#L89

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java
```diff
@@ -125,6 +125,17 @@ public void updatePerfSession(ApplicationProcessState currentAppState) {
     startOrStopCollectingGauges(currentAppState);
   }
 
+  /**
+   * Initial start of gauge collection. This should be called in ContentProvider.attachInfo during
+   * cold-start, because we want to start gauge collection as early as possible. This assumes {@link
+   * PerfSession} was already initialized a moment ago by getInstance(). Unlike updatePerfSession,
+   * this does not reset the perfSession.
+   */
+  public void initializeGaugeCollection() {
```
visumickey, line 134: By this change, we would separate SessionID creation and gaugeCollection initiation separately.

Earlier, with updatePerfSession, a new SessionID will be generated and that sessionID would be associated with all the traces and network requests that go out to the devices. With this change, though there is a sessionID, traces/network requests association to the sessionID would be lost.

Reference: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java#L115

Here is where the sessionID update is propagated to the rest of the SDK and all other traces would record this sessionID. With this new change, this update would not happen and this would make traces/network requests (active at that time - most likely app start trace) not have any session associated with it.

Can you double check if that is the case?

leotianlizhan: This change only separate SessionID creation and gaugeCollection initiation for `FirebasePerfProvider.attachInfo`. For any `clients.updateSession(perfSession)` to be called, it has to be first added to `clients` through `registerForSessionUpdates`, and finding callers of that shows only `Trace` and `NetworkRequestMetricBuilder` registers themselves. During cold start, I doubt any Trace/NetworkTrace will be created before our `FirebasePerfProvider.attachInfo`. (`AppStartTrace` is not a `Trace`, and probably no one has network request in a ContentProvider). 

I guess it is possible that a user could've set up a high priority content provider and has a Trace in that. If you think this is possible, then I can simply add that block of code that updates `clients`, and just leave out the `PerfSession.create()`

visumickey: Sounds good. thanks for clarifying Leo. Makes sense.

firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java
```diff
@@ -125,6 +125,17 @@ public void updatePerfSession(ApplicationProcessState currentAppState) {
...
+    logGaugeMetadataIfCollectionEnabled(ApplicationProcessState.FOREGROUND);
+    startOrStopCollectingGauges(ApplicationProcessState.FOREGROUND);
```
visumickey, line 136: This API `startOrStopCollectingGauges` does not check if we are already collecting gauges or not. So, would there be a possibility where we might be triggering collection from two separate timers? For eg:

calling 
```
startOrStopCollectingGauges(ApplicationProcessState.FOREGROUND)
startOrStopCollectingGauges(ApplicationProcessState.FOREGROUND)
```

this could trigger two collections initiated at the same time. Can we make this atomic to ensure that we do not collect gauges twice at the same time?

Reference: 
https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java#L160

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/session/gauges/CpuGaugeCollector.java#L116

visumickey: I'm guessing that the following blocks should take care of that:

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/session/gauges/CpuGaugeCollector.java#L126-L130

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/session/gauges/MemoryGaugeCollector.java#L83-L89

But, still can you confirm if we are not opening up two separate schedulers? Once confirmed, we are good to go!

leotianlizhan: Yes, in fact, it will stop earlier than those 2 block of code. In `GaugeManager.startCollectingGauges(...)`, it checks `this.sessionId`, if it is not null, then call `stopCollectingGauges()`. It isn't obvious when you look at the variable name and type but `GaugeManager.sessionId` acts as a flag for if we are already collecting gauges.

[This line](https://github.com/firebase/firebase-android-sdk/blob/c91adbebfbc8c5366edef712ba66293eebdb5338/firebase-perf/src/main/java/com/google/firebase/perf/session/gauges/GaugeManager.java#L129) sets `GaugeManager.sessionId` when gauge collection starts.

[This line](https://github.com/firebase/firebase-android-sdk/blob/c91adbebfbc8c5366edef712ba66293eebdb5338/firebase-perf/src/main/java/com/google/firebase/perf/session/gauges/GaugeManager.java#L120) stops gauge collection because `sessionId` was already set. `stopCollectingGauges()` also [resets `sessionId` back to null](https://github.com/firebase/firebase-android-sdk/blob/c91adbebfbc8c5366edef712ba66293eebdb5338/firebase-perf/src/main/java/com/google/firebase/perf/session/gauges/GaugeManager.java#L216).

visumickey: Thanks for confirming @leotianlizhan. Interesting part is the GaugeManager reseting sessionId to null when stopCollectingGauges(). Ideally gauge manager should not worry about reseting the sessionID to null. That is a refactor/cleanup for later.

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/session/SessionManager.java
visumickey: The concern is the source of truth. GaugeManager would have an instance of sessionID which could be null, but the sessionManager would have an instance of sessionID which is not null. This is a recipe of confusion and possibly inconsistencies. As much as possible, we should avoid multiple source of truths. So the callout for the cleanup.

PR #3019, opened 2021-10-01
Check if debug log is enabled before generate dashboard URL

jeremyjiang-dev:
The PR will cut down the cost of generateDashboardURL and save ~10ms if debug log is disabled.


visumickey (approved):
Wow! This is a steal. But something to think about is the cost of logging a message. Are we doing something wrong when we log messages? Worth pursuing with ACore/Android team to make sure they are aware of this.

leotianlizhan (approved):
LGTM!

PR #3028, opened 2021-10-05
AppCheck integration with Firestore

ehsannas:
This is the same as #3027, but AppCheck token change listeners have been removed.


ehsannas:
@google-oss-bot /retest

PR #3036, opened 2021-10-06
Add safety checks to taskutils; dismiss dialog

rachaprince:


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TaskUtils.java
```diff
@@ -35,4 +36,29 @@
     }
     return task;
   }
+
```
pranavrajgopal, line 39: This is nice! we can use this to cleanup other checks like this in a bunch of places as we encounter them in future CLs)

Issue #2647, opened 2021-05-07
FirebaseModelDownloader throws 403 error with API Keys that have app restrictions


kabeer-uber:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 4.1.3
  * Firebase Component: FirebaseModelDownloader
  * Component version: 23.0.0

### [REQUIRED] Step 3: Describe the problem

getModel() with FirebaseModelDownloader throws a 403 error with Google API keys that have android app restrictions.

#### Steps to reproduce:

Create any project and add android app restrictions, use this API key with the FirebaseModelDownloader project to trigger ML model downlaods

![Screen Shot 2021-05-06 at 6 48 29 PM](https://user-images.githubusercontent.com/21314710/117386430-c00a5a80-ae9b-11eb-8f3c-3fc7b0a89c1f.png)

API keys without android app restrictions work, hence changing to None does not throw a 403 error.

![Screen Shot 2021-05-06 at 6 41 50 PM](https://user-images.githubusercontent.com/21314710/117386171-44a8a900-ae9b-11eb-8cdc-29388db19f35.png)

#### Relevant Code:

Code culprit is here as we aren't adding **X-Android-Package** and **X-Android-Cert** in the headers and adding only API key.

https://github.com/firebase/firebase-android-sdk/blob/976ea2fb443bbf7b5828f988275110b6b3b4b281/firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java#L170

```
 @NonNull
  public Task<CustomModel> getCustomModelDetails(
      String projectNumber, String modelName, String modelHash) {
    try {
      URL url =
          new URL(String.format(DOWNLOAD_MODEL_REGEX, downloadHost, projectNumber, modelName));
      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
      connection.setConnectTimeout(CONNECTION_TIME_OUT_MS);
      connection.setRequestProperty(ACCEPT_ENCODING_HEADER_KEY, GZIP_CONTENT_ENCODING);
      connection.setRequestProperty(CONTENT_TYPE, APPLICATION_JSON);
      if (modelHash != null && !modelHash.isEmpty()) {
        connection.setRequestProperty(IF_NONE_MATCH_HEADER_KEY, modelHash);
      }

      Task<InstallationTokenResult> installationAuthTokenTask =
          firebaseInstallations.getToken(false);
[...]

PR #3083, opened 2021-10-27
Support for API key restrictions in FirebaseModelDownloader

kabeer-uber:
Details are provided in this issue description
https://github.com/firebase/firebase-android-sdk/issues/2647

The changes here are at parity with the changes shown below and follow the same pattern:
https://github.com/firebase/firebase-android-sdk/blob/ac3276ce7d5f5181b1c419731495eb88a98537af/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java#L246

There is one caveat that in the referenced **google-play-services-basement-17.4.0** the method below is deprecated, but to keep parity of the pattern with the rest of the codebase I have continued to use it here too. 
`AndroidUtilsLight.getPackageCertificateHashBytes`

### Manual Testing

Was able to export library to mavenLocal() and test it in this simple app
https://github.com/kabeer-uber/firebase-ml-bug-2647

changes work as expected with API restrictions. This also allows us to restrict our keys in our Android app.

As these specific changes are not tested across the other usages, kept the same unit testing parity as the changes are pretty straightfoward.

### DIFF
--- a/firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java
+++ b/firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java
@@ -14,9 +14,14 @@
 
 package com.google.firebase.ml.modeldownloader.internal;
 
+import android.content.Context;
+import android.content.pm.PackageManager;
 import android.util.JsonReader;
 import android.util.Log;
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.android.gms.common.util.AndroidUtilsLight;
+import com.google.android.gms.common.util.Hex;
 import com.google.android.gms.common.util.VisibleForTesting;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.Tasks;
@@ -73,6 +78,8 @@ public class CustomModelDownloadService {
   static final String INSTALLATIONS_AUTH_TOKEN_HEADER = "X-Goog-Firebase-Installations-Auth";
 
   @VisibleForTesting static final String API_KEY_HEADER = "x-goog-api-key";
+  @VisibleForTesting static final String X_ANDROID_PACKAGE_HEADER = "X-Android-Package";
+  @VisibleForTesting static final String X_ANDROID_CERT_HEADER = "X-Android-Cert";
 
   @VisibleForTesting
   static final String DOWNLOAD_MODEL_REGEX = "%s/v1beta2/projects/%s/models/%s:download";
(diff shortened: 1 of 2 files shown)


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java
```diff
@@ -409,4 +424,23 @@ private String getErrorStream(HttpURLConnection connection) {
...
+
+      if (hash == null) {
+        Log.e(TAG, "Could not get fingerprint hash for package: " + context.getPackageName());
+        return null;
+      } else {
+        return Hex.bytesToStringUppercase(hash, /* zeroTerminated= */ false);
+      }
+    } catch (PackageManager.NameNotFoundException e) {
+      Log.e(TAG, "No such package: " + context.getPackageName(), e);
+      return null;
```
vkryachko, line 443: Should we be adding the header if the result is null?

firebase-ml-modeldownloader/src/test/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadServiceTest.java
```diff
@@ -166,7 +169,7 @@ public void downloadService_noHashSuccess() {
                     .withBody(RESPONSE_BODY)));
 
     CustomModelDownloadService service =
-        new CustomModelDownloadService(
+        new CustomModelDownloadService(ApplicationProvider.getApplicationContext(),
```
vkryachko, line 172: Can you add tests in this file to verify that the headers are set correctly?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java
kabeer-uber: @vkryachko It's a good question, probably the backend responds the same way whether null or missing header. I was being consistent with the codebase here, which didn't have the null check.

https://github.com/firebase/firebase-android-sdk/blob/ac3276ce7d5f5181b1c419731495eb88a98537af/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java#L246

Null, check would probably be more efficient on the wire, i.e not send a header if it doesn't exist. 

I can update it if you prefer the minor efficiency over the minor consistency reason. Let me know.

kabeer-uber: I also realized since I am using the static function `AndroidUtilsLight.getPackageCertificateHashBytes(context, context.getPackageName())`, it would likely be easier to test with supporting null, and just check the presence of header.

as also done here.
https://github.com/firebase/firebase-android-sdk/blob/ac3276ce7d5f5181b1c419731495eb88a98537af/firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java#L182

firebase-ml-modeldownloader/src/test/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadServiceTest.java
kabeer-uber: Will do

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java
kabeer-uber: @vkryachko I have addressed the comment in the update, thanks.

firebase-ml-modeldownloader/src/test/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadServiceTest.java
kabeer-uber: @vkryachko I have added tests please review when you get a chance.

vkryachko (approved):

PR #3087, opened 2021-10-29
User-scoped indices

schmidt-sebastian:
This PR changes the IndexManager to use the signed in User for documents with mutations. 

It requires a bunch of changes:
- IndexManager is now user-scoped and gets replaced when the user changes.
- IndexBackfiller and RemoteDocumentCache have setters to replace the IndexManager under the hood
- Misc cleanup that I tried to describe via comments.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
@@ -110,7 +110,10 @@ public final class LocalStore implements BundleCallback {
   private final Persistence persistence;
 
   /** Manages the list of active field and collection indices. */
-  private final IndexManager indexManager;
+  private IndexManager indexManager;
+
+  /** Manages field index backfill. */
+  private final IndexBackfiller indexBackfiller;
 
   /** The set of all mutations that have been sent but not yet been applied to the backend. */
   private MutationQueue mutationQueue;
@@ -145,7 +148,11 @@ public final class LocalStore implements BundleCallback {
   /** Used to generate targetIds for queries tracked locally. */
   private final TargetIdGenerator targetIdGenerator;
 
-  public LocalStore(Persistence persistence, QueryEngine queryEngine, User initialUser) {
+  public LocalStore(
+      Persistence persistence,
+      IndexBackfiller indexBackfiller,
+      QueryEngine queryEngine,
+      User initialUser) {
     hardAssert(
         persistence.isStarted(), "LocalStore was passed an unstarted persistence implementation");
     this.persistence = persistence;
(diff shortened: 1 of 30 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexBackfillerTest.java
```diff
@@ -221,28 +227,6 @@ public void testBackfillWritesUntilCap() {
     assertEquals("coll1", collectionGroups.get(1));
   }
 
-  @Test
-  public void testAddAndRemoveIndexEntry() {
```
schmidt-sebastian, line 225: As far as I can tell, this test only tests test-only methods.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -197,6 +209,11 @@ private void startMutationQueue() {
     localDocuments =
         new LocalDocumentsView(remoteDocuments, mutationQueue, documentOverlayCache, indexManager);
     queryEngine.setLocalDocumentsView(localDocuments);
+    queryEngine.setIndexManager(indexManager);
+
+    remoteDocuments.setIndexManager(indexManager);
```
thebrianchen, line 214: Do we need a test to check that the index manager is swapped when the user changes?

thebrianchen (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: We do - but I am not sure how to add them yet. The user change is only tested via Spec tests and we could do see the affect of switching the user once we also update the Query logic to use indexing by default. I added a TODO.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java
```diff
@@ -69,14 +69,16 @@
...
 
-  MemoryMutationQueue(MemoryPersistence persistence) {
+  MemoryMutationQueue(MemoryPersistence persistence, MemoryIndexManager indexManager) {
     this.persistence = persistence;
     queue = new ArrayList<>();
 
     batchesByDocumentKey = new ImmutableSortedSet<>(emptyList(), DocumentReference.BY_KEY);
     nextBatchId = 1;
     lastStreamToken = WriteStream.EMPTY_STREAM_TOKEN;
+    this.indexManager = indexManager;
```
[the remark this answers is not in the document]
schmidt-sebastian: The instance of IndexManager needs to be shared between all components, so that we share the same collection group index table. This is why I passed it in - but we do already use the same instance for all memory-backed implementations, regardless of user. It makes a ton of sense to just grab a reference here. 
It is semantically the same, but much more readable. Thanks for the suggestion.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -34,12 +34,16 @@
 
   /** Underlying cache of documents and their read times. */
   private ImmutableSortedMap<DocumentKey, Pair<MutableDocument, SnapshotVersion>> docs;
+  /** Manages the collection group index. */
+  private IndexManager indexManager;
```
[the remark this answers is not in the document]
schmidt-sebastian: Added assert.

thebrianchen (approved):
thanks for the explanations!

PR #3097, opened 2021-11-01
Generalize the UnityVersionProvider into DevelopmentPlatformProvider.

mrober:
This will allow us to provide development platform info for other platforms in the future.


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/DevelopmentPlatformProvider.java
```diff
@@ -0,0 +1,64 @@
...
+import static com.google.firebase.crashlytics.internal.common.CommonUtils.getResourcesIdentifier;
+
+import android.content.Context;
+import androidx.annotation.Nullable;
+
+/** Provider for the development platform info. */
+public class DevelopmentPlatformProvider {
+  public static final String UNITY_PLATFORM = "Unity";
+
+  private static final String UNITY_EDITOR = "com.google.firebase.crashlytics.unity_version";
```
mrichards, line 26: Is there a reason this is called `UNITY_EDITOR` instead of something like `UNITY_VERSION_FIELD` ?

mrichards (approved):
One quick q; LGTM either way.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/DevelopmentPlatformProvider.java
mrober: That was the name in the original Unity version provider. But UNITY_VERSION_FIELD is better, renaming.

PR #3099, opened 2021-11-02
Fixes some bugs with the new state managment

rachaprince:
- Removes executor option in the lifecycle notifier because this causes callbacks to execute in an unexpected order
- Sets a succesful signin result when SignInResultAction is created

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TesterSignInClient.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TesterSignInClient.java
@@ -83,7 +83,8 @@ class TesterSignInClient {
 
   @VisibleForTesting
   void onActivityCreated(Activity activity) {
-    // We call finish() in the onCreate method of the SignInResultActivity, so we must set the result
+    // We call finish() in the onCreate method of the SignInResultActivity, so we must set the
+    // result
     // of the signIn Task in the onActivityCreated callback
     if (activity instanceof SignInResultActivity) {
       LogWrapper.getInstance().v("Sign in completed");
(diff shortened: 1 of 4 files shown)


mannyjimenez0810 (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TesterSignInClient.java
```diff
@@ -75,22 +76,24 @@
...
+    lifecycleNotifier.addOnActivityCreatedListener(this::onActivityCreated);
     lifecycleNotifier.addOnActivityStartedListener(this::onActivityStarted);
     lifecycleNotifier.addOnActivityDestroyedListener(this::onActivityDestroyed);
   }
 
-  private void onActivityDestroyed(Activity activity) {
-    this.dismissAlertDialog();
-  }
-
   @VisibleForTesting
```
pranavrajgopal, line 84: Out of curiosity why can't this happen onActivityStarted? Also would be great if you could add a comment explaining why we have different things happening onActivityCreated and onActivityStarted

rachaprince: The flow I noticed was SignInResultActivity Created -> MainActivityCreated -> MainActivityStarted.

I am not sure why SignInResultActivity was not started before MainActivity. Any theories as to why that might happen? 

This logic lived in onCreate originally, so I assumed my implementation was buggy, and we had some reason for putting in onCreate in the original implementation

pranavrajgopal (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TesterSignInClient.java
rachaprince: nvm, I see why. We call finish() in the onCreate of SignInResultActivity. 

I'll add a comment :)

PR #3105, opened 2021-11-04
Rename package to avoid namespace conflicts with gradle plugin

rachaprince:
The SDK FirebaseAppDistribution package name currently conflicts with our gradle plugin, which can cause difficulties importing the SDK class in an IDE.

We also plan to make the gradle classes package private, but this SDK package rename will help customers on older versions of the gradle plugin.


pranavrajgopal (approved):
💯

mannyjimenez0810 (approved):

PR #3111, opened 2021-11-08
Use sequence numbers instead of read time in collection group

thebrianchen:
Instead of using read times to track the ordering of collection groups in the `collection_group_update_times` table, use a sequence counter instead.

Looking for naming suggestions:
- collection_group_indexing
- collection_group_sequence_numbers
- collection_group_sequences


firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -81,9 +80,15 @@
...
+  /**
+   * Returns the next collection group to update. Returns null after reaching the provided max
+   * sequence number.
+   */
   @Nullable
-  String getNextCollectionGroupToUpdate(Timestamp lastUpdateTime);
+  String getNextCollectionGroupToUpdate(int currentMaxSequenceNumber);
+
+  /** Returns the highest sequence number in the collection groups table. */
+  int getMaxCollectionGroupSequenceNumber();
```
schmidt-sebastian, line 91: getIndexBackfillSequenceNumber? I think it might be more interesting to say what the sequence number is for.

I am 50/50 on whether we need this method in our interface though.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -140,11 +139,14 @@ public Results backfill() {
   /** Writes index entries until the cap is reached. Returns the number of documents processed. */
   private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
     int documentsProcessed = 0;
-    Timestamp startingTimestamp = Timestamp.now();
+    int currentMaxSequenceNumber = indexManager.getMaxCollectionGroupSequenceNumber();
 
+    // TODO(indexing): Handle pausing and resuming from the correct document if backfilling hits the
+    // max doc limit while processing docs for a certain read time.
```
schmidt-sebastian, line 145: That's a very big todo :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -55,6 +54,7 @@
 /** A persisted implementation of IndexManager. */
 final class SQLiteIndexManager implements IndexManager {
   private static final String TAG = SQLiteIndexManager.class.getSimpleName();
+  private static final int NEW_COLLECTION_GROUP_SEQUENCE_NUMBER = 0;
```
schmidt-sebastian, line 57: Should we use -1 here? 0 poses the question whether we have indexed the first entry (with a zero-based index).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -208,9 +206,7 @@ private void updateFieldIndex(FieldIndex index) {
     }
 
     // Store that this collection group will updated.
-    // TODO(indexing): Store progress with a counter rather than a timestamp. If using a timestamp,
-    // use the read time of the last document read in the loop.
-    setCollectionGroupUpdateTime(nextCollectionGroup[0], Timestamp.now());
+    setCollectionGroup(nextCollectionGroup[0], getMaxCollectionGroupSequenceNumber() + 1);
```
schmidt-sebastian, line 209: Hm, this should only be set after we are done with the update.

schmidt-sebastian (commented):
I think this PR needs to wait until we have a holistic view on how we keep track of mutation batches and partial updates (e.g. we processed 10 out of 100 documents for a single read time). I am no longer certain that sequence numbers are the right solution here, so it would be good to confirm that they are before we proceed.

thebrianchen (commented):
Made some revisions:
- Mark collection group as completed only after indexing all documents in the collection.
- Removed `getMaxCollectionGroupSequenceNumber()` from index manager and instead let the backfiller track when all collection groups have been processed.
- Added test to verify that collection groups that are partially indexed are processed first.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -169,8 +174,15 @@ private int writeEntriesForCollectionGroup(
...
         localDocumentsView.getDocumentsMatchingQuery(query, earliestUpdateTime);
 
-    Queue<Document> oldestDocuments = getOldestDocuments(documents, entriesRemainingUnderCap);
+    Queue<Document> oldestDocuments = getOldestDocuments(documents, documentsRemaining);
     indexManager.updateIndexEntries(oldestDocuments);
+
+    // Mark the collection group as fully indexed if all documents in the collection have been
+    // indexed during this backfill iteration.
+    if (documentsRemaining >= documents.size()) {
+      indexManager.markCollectionGroupIndexed(collectionGroup);
```
schmidt-sebastian, line 183: This is different from how I implemented it. I always increment the sequence number regardless of whether I am done or not - it's a tiny bit simpler but may not be as optimal as what you are doing. We should discuss.

thebrianchen:
Closing the PR since this was incorporated into #3153

PR #3112, opened 2021-11-08
Release Overlay

wu-hui:
Performance test result: https://github.com/firebase/firebase-android-sdk/pull/3161


schmidt-sebastian (approved):
Woohoo!

Please verify that the schema upgrades works as intended using an actual app. Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/local/Persistence.java
```diff
@@ -52,7 +52,7 @@
 
   /** Temporary setting for enabling document overlays. */
   // TODO(Overlay): Remove this.
-  public static boolean OVERLAY_SUPPORT_ENABLED = BuildConfig.ENABLE_OVERLAY;
+  public static boolean OVERLAY_SUPPORT_ENABLED = true;
```
mikehardy, line 55: I understand I am just some random external person looking at PRs (why am I even here 😆 ?) but curious why not remove the entire symbol if the feature is shipping. Seems it should not be referenced anymore as there is no way to toggle it

wu-hui: Ha, you are right this should be removed.

We are keeping it for now because there are more work associated with fully removing the old code, some old logic is no longer required and should be removed. But we want to keep the release PR small, that that work will be in a followup PR.

But thanks for the extra pair of eyes!

wu-hui:
@schmidt-sebastian PLAAL.

Without the new commit, upgrade will actually crash, because `data_migrations` table is not created. Thanks for reminding me to check with an actual App.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -152,23 +154,22 @@ private void udpateViews(int targetId, boolean fromCache) {
...
     SnapshotVersion version = version(documentVersion);
-    MutationResult mutationResult =
-        new MutationResult(
-            version,
-            transformResult != null
-                ? Collections.singletonList(TestUtil.wrap(transformResult))
-                : Collections.emptyList());
+    List<MutationResult> mutationResults =
+        transformResult == null
+            ? singletonList(new MutationResult(version, emptyList()))
```
schmidt-sebastian, line 162: We usually refer to these method using the class name: ` Arrays.singletonList`.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -963,9 +964,8 @@ public void testReadsAllDocumentsForInitialCollectionQueries() {
     localStore.executeQuery(query, /* usePreviousResults= */ true);
 
     assertRemoteDocumentsRead(/* byKey= */ 0, /* byQuery= */ 2);
-    if (!Persistence.OVERLAY_SUPPORT_ENABLED) {
```
schmidt-sebastian, line 966: Not a blocker, but we should ideally remove all statements like this (including the flag) or leave them all here. If we leave the flag, we should be able to turn it off during development and have the tests pass.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -203,7 +203,9 @@ private void recalculateAndSaveOverlays(Map<DocumentKey, MutableDocument> docs)
     // along the way.
     for (MutationBatch batch : batches) {
       for (DocumentKey key : batch.getKeys()) {
-        FieldMask mask = batch.applyToLocalView(docs.get(key), masks.get(key));
+        FieldMask mask =
+            masks.containsKey(key) ? masks.get(key) : FieldMask.fromSet(new HashSet<>());
```
schmidt-sebastian, line 207: Can we create an EMPTY_MASK sentinel? The missing mask is the default case and we create two objects each time.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -152,23 +151,25 @@ private void udpateViews(int targetId, boolean fromCache) {
     notifyLocalViewChanges(viewChanges(targetId, fromCache, asList(), asList()));
   }
 
-  private void acknowledgeMutation(long documentVersion, @Nullable Object transformResult) {
+  private void acknowledgeMutationWithTransform(long documentVersion, Object... transformResult) {
```
schmidt-sebastian, line 157: acknowledgeMutationWithTransformResults ? (or leave as before)

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -152,23 +151,25 @@ private void udpateViews(int targetId, boolean fromCache) {
...
     SnapshotVersion version = version(documentVersion);
-    MutationResult mutationResult =
-        new MutationResult(
-            version,
-            transformResult != null
-                ? Collections.singletonList(TestUtil.wrap(transformResult))
-                : Collections.emptyList());
+    List<MutationResult> mutationResults =
+        Collections.singletonList(new MutationResult(version, emptyList()));
+    if (transformResult != null) {
```
schmidt-sebastian, line 162: This array should never be null. You can also use Java8 streams to simplify this logic.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
wu-hui: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
wu-hui: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
wu-hui: It is null when `acknowledgeMutation` is called.

schmidt-sebastian: You should call it with no arguments instead of `null`.

schmidt-sebastian: Bump :)

wu-hui: Fixed.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -155,23 +153,27 @@ private void udpateViews(int targetId, boolean fromCache) {
...
+    List<MutationResult> mutationResults =
+        Collections.singletonList(new MutationResult(version, emptyList()));
+
+    if (transformResult.length != 0) {
+      mutationResults = new ArrayList<>();
+      for (Object o : transformResult) {
+        mutationResults.add(
+            new MutationResult(version, Collections.singletonList(TestUtil.wrap(o))));
+      }
+    }
```
schmidt-sebastian, line 169:
```suggestion
    List<MutationResult> mutationResults =
        Arrays.stream(transformResult)
            .map(r -> new MutationResult(version, Collections.singletonList(TestUtil.wrap(r))))
            .collect(Collectors.toList());
```

schmidt-sebastian: (needs some handling for empty list)

wu-hui: done

schmidt-sebastian (approved):

PR #3118, opened 2021-11-09
Add ability to delete indexes

schmidt-sebastian:
This PR diffs the index configuration that the user provides with the currently active definition and removes indexes that are no longer used

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
@@ -791,12 +791,12 @@ public final class LocalStore implements BundleCallback {
         "Configure indexes",
         () -> {
           List<FieldIndex> existingIndexes = new ArrayList<>(indexManager.getFieldIndexes());
-          Collections.sort(existingIndexes, FieldIndex.SEMANTIC_EQUALITY_OPERATOR);
+          Collections.sort(existingIndexes, FieldIndex.SEMANTIC_COMPARATOR);
           Iterator<FieldIndex> existingIt = existingIndexes.iterator();
           @Nullable FieldIndex existingValue = advanceIterator(existingIt);
 
           List<FieldIndex> updatedIndexes = new ArrayList<>(newFieldIndexes);
-          Collections.sort(updatedIndexes, FieldIndex.SEMANTIC_EQUALITY_OPERATOR);
+          Collections.sort(updatedIndexes, FieldIndex.SEMANTIC_COMPARATOR);
           Iterator<FieldIndex> updatedIt = updatedIndexes.iterator();
           @Nullable FieldIndex updatedValue = advanceIterator(updatedIt);
 
@@ -805,7 +805,7 @@ public final class LocalStore implements BundleCallback {
             boolean updated = false;
 
             if (existingValue != null && updatedValue != null) {
-              int cmp = FieldIndex.SEMANTIC_EQUALITY_OPERATOR.compare(existingValue, updatedValue);
+              int cmp = FieldIndex.SEMANTIC_COMPARATOR.compare(existingValue, updatedValue);
               if (cmp < 0) {
                 deleted = true;
               } else if (cmp > 0) {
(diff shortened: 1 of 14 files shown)


schmidt-sebastian:
cc @thebrianchen

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -778,13 +779,43 @@ public void saveNamedQuery(NamedQuery namedQuery, ImmutableSortedSet<DocumentKey
...
-        "Configure indices",
+        "Configure indexes",
         () -> {
-          // TODO(indexing): Disable no longer active indices
-          for (FieldIndex fieldIndex : fieldIndices) {
-            indexManager.addFieldIndex(fieldIndex);
+          Collection<FieldIndex> existingIndexes = indexManager.getFieldIndexes();
+
+          // Delete indices that are no longer used.
+          for (FieldIndex existingIndex : existingIndexes) {
```
wu-hui, line 794: I am a bit worried about performance here. With a 30 indexes setup, this could easily lead up to 1800 (or 1000 realistically) operations. 

I want to suggest to explore something like a bloom filter to help. But I realize this might also be premature optimization. So how about adding a TODO to followup?

wu-hui (commented):
LGTM on what is in the PR. 

I only left one comment, let me know if you want to simply add a TODO, I'll approve in that case.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
wu-hui: Also, can the two loop be merged, such that we only call `matchesConstraints` half times?

schmidt-sebastian: I rewrote the entire thing to run in `O(n log n)`. Can you take another look?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -778,13 +782,52 @@ public void saveNamedQuery(NamedQuery namedQuery, ImmutableSortedSet<DocumentKey
...
+          Iterator<FieldIndex> updatedIt = updatedIndexes.iterator();
+          @Nullable FieldIndex updatedValue = advanceIterator(updatedIt);
+
+          while (existingValue != null || updatedValue != null) {
+            boolean deleted = false;
+            boolean updated = false;
+
+            if (existingValue != null && updatedValue != null) {
+              int cmp = FieldIndex.SEMANTIC_COMPARATOR.compare(existingValue, updatedValue);
+              if (cmp < 0) {
```
wu-hui, line 810: It is not obvious immediately why comparison result leads to `deleted` or `updated`. Can we add some comment? Ideally a high level comment for the method, and some implementation comment to explain details.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -778,13 +782,52 @@ public void saveNamedQuery(NamedQuery namedQuery, ImmutableSortedSet<DocumentKey
...
+          Iterator<FieldIndex> existingIt = existingIndexes.iterator();
+          @Nullable FieldIndex existingValue = advanceIterator(existingIt);
+
+          List<FieldIndex> updatedIndexes = new ArrayList<>(newFieldIndexes);
+          Collections.sort(updatedIndexes, FieldIndex.SEMANTIC_COMPARATOR);
+          Iterator<FieldIndex> updatedIt = updatedIndexes.iterator();
+          @Nullable FieldIndex updatedValue = advanceIterator(updatedIt);
+
+          while (existingValue != null || updatedValue != null) {
+            boolean deleted = false;
```
wu-hui, line 805: I think they can be named `deleteExistingIndex` and `writeNewIndex`?

schmidt-sebastian: I actually realized yesterday that I need to re-use this functionality in a follow up PR, so I changed this method to be more generic and moved it to Util.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
schmidt-sebastian: I added comments to the more generic implementation, as well as tests.

wu-hui (approved):
LGTM!

PR #3122, opened 2021-11-10
Fireperf: fix incorrect `sessionId`s for `_fs` and `_bs` traces

leotianlizhan:
b/204362742

## TLDR
`_fs` and `_bs` traces are using the next sessionId instead of the current sessionId because we call `updateAppState` ([line 176](https://github.com/firebase/firebase-android-sdk/pull/3122/files#diff-eb968bff71b0337966fe331ee130c0ad812d5fed5e8b15c7e676a565bb0074c5L176)) before sending `_fs` or `_bs` traces ([line 177](https://github.com/firebase/firebase-android-sdk/pull/3122/files#diff-eb968bff71b0337966fe331ee130c0ad812d5fed5e8b15c7e676a565bb0074c5L177)).  One of the callbacks of `updateAppState` is `SessionManager` which updates the sessionId, thus `_fs` and `_bs` uses the next session's sessionId. 

This fix simply changes the order of methods, delaying `updateAppState` until session trace is sent.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -174,7 +174,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
         // no more activity in foreground, app goes to background.
         stopTime = clock.getTime();
         sendSessionLog(Constants.TraceNames.FOREGROUND_TRACE_NAME.toString(), resumeTime, stopTime);
-        // order is important to avoid b/204362742
+        // order is important to transport _fs before triggering a sessionId change b/204362742
         updateAppState(ApplicationProcessState.BACKGROUND);
       }
     }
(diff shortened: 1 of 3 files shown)


jeremyjiang-dev (commented):
Thanks for finding the bug and creating a fix! The changes look good to me. 

Can you add some tests to make sure the session IDs are correctly generated?

jeremyjiang-dev (approved):
Thanks for identifying and fixing this!!!

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -173,8 +173,9 @@ public synchronized void onActivityStopped(Activity activity) {
       if (activityToResumedMap.isEmpty()) {
         // no more activity in foreground, app goes to background.
         stopTime = clock.getTime();
-        updateAppState(ApplicationProcessState.BACKGROUND);
         sendSessionLog(Constants.TraceNames.FOREGROUND_TRACE_NAME.toString(), resumeTime, stopTime);
+        // order is important to avoid b/204362742
+        updateAppState(ApplicationProcessState.BACKGROUND);
```
visumickey, line 178: Can you also qualify the reasoning for this ordering? For example: Complete _fs trace before triggering a change in the application state as that would lead to a new sessionID change.

visumickey (approved):
Thanks for making this change.

PR #3124, opened 2021-11-10
Fix heartbeat reporting for App Check.

rosalyntan:


appcheck/firebase-appcheck/src/test/java/com/google/firebase/appcheck/internal/NetworkClientTest.java
```diff
@@ -234,11 +233,8 @@ public void exchangeAttestation_cannotRetry_throwsException() {
 
   private void verifyRequestHeaders() {
     String userAgent = ((DefaultFirebaseAppCheck) FirebaseAppCheck.getInstance()).getUserAgent();
-    String heartBeatCode =
-        ((DefaultFirebaseAppCheck) FirebaseAppCheck.getInstance()).getHeartbeatCode();
 
     verify(mockHttpUrlConnection).setRequestProperty(X_FIREBASE_CLIENT, userAgent);
-    verify(mockHttpUrlConnection).setRequestProperty(X_FIREBASE_CLIENT_LOG_TYPE, heartBeatCode);
```
vkryachko, line 241: Would it be feasible to test both cases: when the heartbeat is `NONE`, and when it's not?

rosalyntan: Yeah, I wasn't sure how to test the heartbeat, besides mocking out `FirebaseApp` and `DefaultFirebaseAppCheck`. Is there a way I can provide a fake/mock `HeartBeatInfo` to `FirebaseAppCheck` without mocking `FirebaseApp` and `FirebaseAppCheck`?

vkryachko: looks like the fact that NetworkClient takes `FirebaseApp` is what makes things hard to test, i.e. most of the stuff that is going on in `NetworkClient`s constructor is kind of going against:
https://github.com/google/guice/wiki/InjectOnlyDirectDependencies

I would suggest to create another constructor to NetworkClient of the form:

`NetworkClient(Context context, FirebaseOptions options, Provider<HeartBeatInfo>, Provider<UserAgentPublisher>)`, this way you would only have to mock `UAPublisher` and `HeartBeatInfo`. To preserve backwards compat you can keep the existing constructor and just delegate to this new one.

rosalyntan: Done.

vkryachko (approved):
LGTM, assuming all tests pass

PR #3130, opened 2021-11-11
Update crashpad.

mrober:
Pulled in crashpad, and fixed the build. Tested on api 26 and api 30.


kmandrika (approved):

mrichards (approved):
Changes look fine, though we should do some extensive manual testing before the next release.

PR #3132, opened 2021-11-11
Update document index entries

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/index/IndexEntry.java
```diff
@@ -14,44 +14,53 @@
...
+import com.google.auto.value.AutoValue;
+import com.google.firebase.firestore.model.DocumentKey;
+import com.google.firebase.firestore.util.Util;
 
-  public byte[] getArrayValue() {
-    return arrayValue;
-  }
+/** Represents an index entry saved by the SDK in its local storage. */
+@AutoValue
+public abstract class IndexEntry implements Comparable<IndexEntry> {
```
schmidt-sebastian, line 27: Made this an AutoValue so that we have hash/equals. I am also planning to keep this class so I removed the TODO that says it is a placeholder.

cc @thebrianchen

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -52,9 +52,6 @@
    */
   List<ResourcePath> getCollectionParents(String collectionId);
 
-  /** Updates the index entries for the given document. */
-  void handleDocumentChange(@Nullable Document oldDocument, @Nullable Document newDocument);
```
schmidt-sebastian, line 56: All updates will be handled by `updateIndexEntries`.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Util.java
```diff
@@ -211,6 +212,34 @@ public static void crashMainThread(RuntimeException exception) {
...
+    return left == null ? (right == null ? 0 : -1) : 1;
+  }
+
+  public static <T extends Comparable<T>> int nullSafeCompare(@Nullable T left, @Nullable T right) {
+    return nullSafeCompare(left, right, Comparable::compareTo);
+  }
+
+  public static int compareByteArrays(byte[] left, byte[] right) {
+    int size = Math.min(left.length, right.length);
+    for (int i = 0; i < size; i++) {
```
schmidt-sebastian, line 229: This code is identical with the implementation right below.

schmidt-sebastian:
cc @thebrianchen

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -304,84 +325,83 @@ private FieldIndex getPostUpdateIndex(FieldIndex baseIndex, SnapshotVersion newR
...
     } else {
-      addSingleEntry(document, fieldIndex.getIndexId(), /* arrayValue= */ null, directionalValue);
-      return true;
+      result.add(
+          IndexEntry.create(
+              fieldIndex.getIndexId(),
+              document.getKey(),
+              uid,
+              directionalValue,
+              /* arrayValue= */ new byte[] {}));
```
schmidt-sebastian, line 358: Using empty byte array instead of `null` makes queries easier.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
```diff
@@ -299,6 +301,20 @@ public void testLimitAppliesOrdering() {
...
+  public void testIndexEntriesAreUpdated() {
+    SQLiteIndexManager sqLiteIndexManager = (SQLiteIndexManager) indexManager;
+    sqLiteIndexManager.addFieldIndex(
+        fieldIndex("coll", "value", FieldIndex.Segment.Kind.ASCENDING));
+    Query query = query("coll").orderBy(orderBy("value"));
+
+    addDoc("coll/doc1", map("value", true));
+    verifyResults(query, "coll/doc1");
+
+    addDocs(doc("coll/doc1", 1, map()), doc("coll/doc2", 1, map("value", true)));
```
wu-hui, line 314: Question, is this PR supposed to handle document deletion? What happens if we call `addDocs(deletedDoc("coll/doc2"))` after this line?

wu-hui (commented):
the code looks great..I have one question on deleted doc.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
schmidt-sebastian: I added a test for this but this already works as is as deleted docs will not have any fields, and as such they will not have index entries.

wu-hui: Thanks, that is what I suspected.

wu-hui (approved):

PR #3133, opened 2021-11-12
Fireperf: create an activity with fragments in `dev-app` for testing

leotianlizhan:
b/208271486: while working on Fragment performance measurement, we need an activity with multiple fragments to test on, in the dev-app.

## Current changes
This adds a button to the main activity that starts a new `FragmentActivity` which uses `Navigation` component with 3 fragments. 

## Testing Listeners
There are also preliminary listeners added to the `FragmentActivity.onCreate()` that we could possibly utilize in our performance measurements, namely `getSupportFragmentManager().registerFragmentLifecycleCallbacks(...)` and `Navigation.findNavController(...).addOnDestinationChangedListener(...)`.

Filter by keyword "Fragment" in logcat and navigate around to see the behaviour of each lifecycle callback. 

## Observation
Navigation component uses something called `NavHostFragment` as a container to swap different fragments into the view. But `NavHostFragment` itself is a `Fragment`, so perhaps we want to exclude that from measurements.

### DIFF
--- a/firebase-perf/dev-app/dev-app.gradle
+++ b/firebase-perf/dev-app/dev-app.gradle
@@ -96,9 +96,6 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.multidex:multidex:2.0.1'
     implementation 'com.google.android.gms:play-services-tasks:17.2.0'
-
-    // 3rd party Deps
-    implementation 'com.squareup.okhttp3:okhttp:4.9.0'
     implementation 'androidx.appcompat:appcompat:1.3.1'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.1'
     implementation 'androidx.vectordrawable:vectordrawable:1.1.0'
@@ -107,6 +104,10 @@ dependencies {
     implementation 'androidx.navigation:navigation-fragment:2.3.5'
     implementation 'androidx.navigation:navigation-ui:2.3.5'
     implementation 'com.google.android.material:material:1.4.0'
+
+    // 3rd party Deps
+    implementation 'com.squareup.okhttp3:okhttp:4.9.0'
+    implementation 'com.github.bumptech.glide:glide:4.12.0'
 
     // Integration Test Deps
     androidTestImplementation 'junit:junit:4.13.1'
--- a/firebase-perf/dev-app/dev-app.gradle
+++ b/firebase-perf/dev-app/dev-app.gradle
@@ -99,6 +99,14 @@ dependencies {
 
     // 3rd party Deps
     implementation 'com.squareup.okhttp3:okhttp:4.9.0'
+    implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation 'androidx.constraintlayout:constraintlayout:2.1.1'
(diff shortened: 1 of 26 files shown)


jeremyjiang-dev (commented):
The empty fragments have very low number of total frames. I think the reason is that the layout is empty and static. Can you add some gifs or pictures to the fragments so that we can have a large number of frames rendered for testing purposes? You may refer to the [codelab](https://firebase.google.com/codelabs/measure-android-view-performance#0) about how to add gifs using Glide library.

leotianlizhan:
> The empty fragments have very low number of total frames. I think the reason is that the layout is empty and static. Can you add some gifs or pictures to the fragments so that we can have a large number of frames rendered for testing purposes? You may refer to the [codelab](https://firebase.google.com/codelabs/measure-android-view-performance#0) about how to add gifs using Glide library.

I added Gifs in all 3 fragments using Glide library. Total frames increased by a lot, but slow/frozen frames remain low. Should I look into using a recyclerview with a lot of items (or each item is a gif) to force it to produce more slow frames?

jeremyjiang-dev:
> > The empty fragments have very low number of total frames. I think the reason is that the layout is empty and static. Can you add some gifs or pictures to the fragments so that we can have a large number of frames rendered for testing purposes? You may refer to the [codelab](https://firebase.google.com/codelabs/measure-android-view-performance#0) about how to add gifs using Glide library.
> 
> I added Gifs in all 3 fragments using Glide library. Total frames increased by a lot, but slow/frozen frames remain low. Should I look into using a recyclerview with a lot of items (or each item is a gif) to force it to produce more slow frames?

Thanks for adding the Gifs! They look great :) I don't have strong preference on the number of slow/frozen frames. It's up to you if you want to produce more slow frames.

visumickey (approved):
Change look good to me. But there are few suggestions for optimizing this further.

1. All the viewModels (Dashboard, Home, Notification) are exactly the same with minor changes. Can we make all these 3 model derive from a common model with just the overrides for the text?
2. All the fragments are very much the same. Should we make all the fragments slightly different where we could have some of the fragments with a good performance and some with not so good performance?

These are not blockers. But something to consider for future iterations.

leotianlizhan:
> Change look good to me. But there are few suggestions for optimizing this further.
> 
> 1. All the viewModels (Dashboard, Home, Notification) are exactly the same with minor changes. Can we make all these 3 model derive from a common model with just the overrides for the text?
> 2. All the fragments are very much the same. Should we make all the fragments slightly different where we could have some of the fragments with a good performance and some with not so good performance?
> 
> These are not blockers. But something to consider for future iterations.

For 1, I think in a real app the ViewModels are most likely totally different, because each fragment is serving different content and layout in a real app. So it makes sense that they don't inherit a common ViewModel. It is just that in our case of testapp they are all super similar. 

For 2, maybe we can swap `frament_notification`'s gif with a very laggy recyclerview? Should I do that as part of this PR or leave it for the future?

visumickey:
> For 2, maybe we can swap `frament_notification`'s gif with a very laggy recyclerview? Should I do that as part of this PR or leave it for the future?

If it not much of a change, can we do that as a part of this PR?

visumickey (approved):

jeremyjiang-dev (approved):

PR #3137, opened 2021-11-15
Move FirestoreRegistrar into the public API

daymxn:
It is currently not possible to draw a dependency on Firestore via components, outside the scope of this library. I'm fairly certain that this is all that's blocking it- but await feedback from others.

@vkryachko @schmidt-sebastian

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
@@ -37,7 +37,6 @@ import java.util.List;
  * @hide
  */
 @Keep
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class FirestoreRegistrar implements ComponentRegistrar {
   @Override
   @Keep


schmidt-sebastian:
This LGTM but I defer final approval to Vlad.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirestoreRegistrar.java
```diff
@@ -37,7 +37,6 @@
  * @hide
  */
 @Keep
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
vkryachko, line 40: You don't want to depend on the registrar itself, if you wanted to depend on Firestore through components you would have to take a dependency on its `FirestoreMultiDbComponent`(which is not `java public`), however that component has a very unrestricted API, i.e. you can get an instance of any database by name. But the problem is that firestore only supports one database per GCP project, so exposing that method publicly is not safe.

It might be ok to depend on firestore through `FirebaseFirestore.getInstance(app)` for now. The alternative would be to create a dedicated component that would return the Firestore instance for the given app, something like:

```
class FirestoreComponent {
  FirebaseFirestore get();
}
```

But it may be too much overhead for now

PR #3138, opened 2021-11-15
Remove shared index entries

schmidt-sebastian:
This PR makes it so that all index entries are scoped to a user, which removes shared index entries for documents that are not mutated. For apps that use multiple users this means that we will keep more index entries, but in the general case it means that we keep one less (instead of keeping the unmodified value + the modified value for single user indices, we now just keep one).

This means that each index now has a different backfill state per user. If a new user signs in, we need to start backfill from zero. We likely need an additional data structure for that.


schmidt-sebastian:
cc @wu-hui

thebrianchen (approved):
lgtm for merging this into your other branch

PR #3143, opened 2021-11-17
Enforce consistency in creating / managing Crashlytics files

mrichards:
Simplifies file access for the Crashlytics SDKs, by having the FileStore
object manage creation of all File (and directory) references.

The FileStore now maintains a consistent directory hierarchy for all
internal Crashlytics files. A side benefit is that session-specific
files are all located in one subdirectory, so cleaning up session files
is less error-prone. This fixes a known bug where some log & keys files
were left behind indefinitely.

Co-authored-by: Matt Robertson <mrober@google.com>


mrober (approved):

vladimirov001:
After this changes crashlitics don't send logs with crash. I've tried to debug it, there is some problem with getWorkingFileForSession(), seems like needs additional call of mkdirs() somewhere.

mrichards:
@vladimirov001 We haven't been able to reproduce that issue. Can you provide more specifics about when you're logging & crashing, the device & api level, etc? Or if you have a minimum reproduceable example, that would be great.

Log messages are serialized async, so it is possible if the crash occurs immediately after the log call, the log won't make it into the report. If you have an example where the same code worked before this change vs it not working now, please let us know.

PR #3147, opened 2021-11-18
Make FieldIndex an AutoValue

schmidt-sebastian:
This changes FIeldIndex to be an AutoValue with the goal to make it easier to extend and add new properties.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -44,6 +44,7 @@ import com.google.firebase.firestore.model.DatabaseId;
 import com.google.firebase.firestore.model.FieldIndex;
 import com.google.firebase.firestore.model.FieldPath;
 import com.google.firebase.firestore.model.ResourcePath;
+import com.google.firebase.firestore.model.SnapshotVersion;
 import com.google.firebase.firestore.remote.FirestoreChannel;
 import com.google.firebase.firestore.remote.GrpcMetadataProvider;
 import com.google.firebase.firestore.util.AsyncQueue;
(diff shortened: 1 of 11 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/model/FieldIndexTest.java
```diff
@@ -29,132 +28,64 @@
 @Config(manifest = Config.NONE)
 public class FieldIndexTest {
 
-  @Test
```
schmidt-sebastian, line 32: These tests are not needed anymore since equals() is generated.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -94,13 +94,15 @@ public void start() {
         .forEach(
             row -> {
               try {
+                int indexId = row.getInt(0);
+                String collectionGroup = row.getString(1);
+                List<FieldIndex.Segment> segments =
+                    serializer.decodeFieldIndexSegments(Index.parseFrom(row.getBlob(2)));
+                SnapshotVersion updateTime =
+                    new SnapshotVersion(new Timestamp(row.getLong(3), row.getInt(4)));
```
thebrianchen, line 102: Seconds are stored in SQLite as `INTEGER`, does it auto-convert to long?

thebrianchen (approved):
super clean, love this!

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -322,21 +323,24 @@ public FirebaseApp getApp() {
...
+              segments.add(FieldIndex.Segment.create(fieldPath, FieldIndex.Segment.Kind.CONTAINS));
             } else if ("ASCENDING".equals(field.optString("order"))) {
-              fieldIndex = fieldIndex.withAddedField(fieldPath, FieldIndex.Segment.Kind.ASCENDING);
+              segments.add(FieldIndex.Segment.create(fieldPath, FieldIndex.Segment.Kind.ASCENDING));
             } else {
-              fieldIndex = fieldIndex.withAddedField(fieldPath, FieldIndex.Segment.Kind.DESCENDING);
+              segments.add(
+                  FieldIndex.Segment.create(fieldPath, FieldIndex.Segment.Kind.DESCENDING));
             }
-            parsedIndices.add(fieldIndex);
```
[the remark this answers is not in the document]
schmidt-sebastian: It adds an the same index multiple times if it has more than one field.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
schmidt-sebastian: INTEGER means non-floating number of any storage size for SQLite: https://www.sqlite.org/datatype3.html

PR #3149, opened 2021-11-18
Startup tracing: remove `FirebasePerfProvider`

leotianlizhan:


firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -432,7 +434,8 @@ protected FirebaseApp(Context applicationContext, String name, FirebaseOptions o
             .addComponentRegistrar(new FirebaseCommonRegistrar())
             .addComponent(Component.of(applicationContext, Context.class))
             .addComponent(Component.of(this, FirebaseApp.class))
-            .addComponent(Component.of(options, FirebaseOptions.class));
+            .addComponent(Component.of(options, FirebaseOptions.class))
+            .addComponent(Component.of(new StartupTime(options.startupTime), StartupTime.class));
```
vkryachko, line 438: Note that `options.startupTime` may be invalid, i.e. it can be set to `Instant.NEVER` if initialization is not happening at app startup. Can we check for that and add this component only if `startupTime.isValid() == true`?

This would also mean that you would have to change your component deps to `optional()`, how feasible is it to change?

firebase-common/src/main/java/com/google/firebase/time/StartupTime.java
```diff
@@ -0,0 +1,13 @@
+package com.google.firebase.time;
+
+public class StartupTime {
```
vkryachko, line 3: AutoValue?

Issue #3084, opened 2021-10-28
Crash on TransportManager Firebase Perf


stevenfredian:
* Android Studio version: _____
  * Firebase Component: Performance
  * Component version: 19.0.11

**Describe the problem**
We are getting following crash on Crashlytics for good number of users with Android OS 8.1.0 only, since we upgrade bom version 26.0.0 to 26.2.0.

Could you please help us, since its happening on multiple users.

<img width="1104" alt="Screen Shot 2021-10-28 at 09 46 57" src="https://user-images.githubusercontent.com/15241939/139179779-c67d3829-52bb-4969-b64f-c689267a7e12.png">

<img width="330" alt="Screen Shot 2021-10-28 at 10 10 51" src="https://user-images.githubusercontent.com/15241939/139179787-28254785-1586-46ae-9ea8-12d03345761a.png">

We already tried several ways, but we could'nt reproduce it.

stevenfredian:
Thanks for the feedback, @argzdev . 
Unfortunately we cannot replicate the issue

stevenfredian:
Thank you, @argzdev .

stevenfredian:
Thank you, @argzdev .

We are using version 30.

We've already try, but we couldn't reproduce it on our codebase, eventhough with 8.1.0 os devices.
But we will try to reproduce with MCVE one.
I'll let you know as soon as possible.

stevenfredian:
Hi @argzdev , unfortunately we cannot reproduce it with emulator and our existing device.

argzdev:
Hello @stevenfredian, thanks for trying to reproduce the issue. Just to confirm, did your users still encounter this type of issue after using the latest sdk?

stevenfredian:
Hi @argzdev , i check on latest bom version (29.0.0), still using Firebase Performance 20.0.3 which has no changes on related function.
So i guess users will still encounter this issue as well.

stevenfredian:
Hi @argzdev , could you please elaborate the current status?

visumickey:
@stevenfredian Looks like the changes made by @argzdev is not submitted due to CI restrictions. I've made a similar change and will push this in quick to rollout the fix.

argzdev:
Hi @stevenfredian, the fix has been merged and will be included in the next release. I can't provide any specific timeline to when this feature will be available, but you may keep an eye out on our [release notes](https://firebase.google.com/support/release-notes/android) or open a new ticket once you still encounter the issue.

[...]

PR #3151, opened 2021-11-19
String format with decimal issue in Android 8.1.0

argzdev:
Speculative fix for issue #3084. NullPointerException is encountered specifically on Android version 8.1.0, this is due to String.format inserted with Double values.

The proposed fix is to use DecimalFormat to format the values before being inserted in String.format


visumickey (approved):
Since this is speculative fix, the changes LGTM. The changes are harmless even if the fix does not resolve the issue. We can merge as soon as the PR goes green.

I'm working to see what can be done for the CLA.

argzdev:
Hi @visumickey, thank you for the approval. Unfortunately, after discussing with my team, it looks like there was an issue with the author email(personal email) I used when committing the change, as a result this might cause an issue with the CLA. So in order to avoid this, I created another PR with the correct details:

https://github.com/firebase/firebase-android-sdk/pull/3176

I'll close this now. Thank you!

PR #3153, opened 2021-11-19
Store per-user state for indexes

schmidt-sebastian:
This PR moves the "how updated is the index?" state from the index table to a per-user table and adds a sequence number. This allows the backfiller to process all index entries independently for each user.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
@@ -15,12 +15,12 @@
 package com.google.firebase.firestore.local;
 
 import androidx.annotation.Nullable;
-import com.google.firebase.Timestamp;
 import com.google.firebase.firestore.core.Target;
 import com.google.firebase.firestore.model.Document;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.FieldIndex;
 import com.google.firebase.firestore.model.ResourcePath;
+import com.google.firebase.firestore.model.SnapshotVersion;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
@@ -87,14 +87,22 @@ public interface IndexManager {
   /** Returns the documents that match the given target based on the provided index. */
   Set<DocumentKey> getDocumentsMatchingTarget(FieldIndex fieldIndex, Target target);
 
-  /** Returns the next collection group to update. */
+  /**
+   * Returns the next collection group to update. Returns {@code null} if no group exists or the
+   * next collection group is already at {@code maxSequenceNumber}.
+   */
   @Nullable
-  String getNextCollectionGroupToUpdate(Timestamp lastUpdateTime);
+  String getNextCollectionGroupToUpdate(long maxSequenceNumber);
+
+  /** Updates the sequence number for the collection group and sets its latest read time. */
+  void updateCollectionGroup(String collectionGroup, long sequenceNumber, SnapshotVersion readTime);
 
   /**
    * Updates the index entries for the provided documents and corresponding field indexes until the
-   * cap is reached. Updates the field indexes in persistence with the latest read time that was
-   * processed.
+   * cap is reached.
    */
   void updateIndexEntries(Collection<Document> documents);
+
+  /** Returns the largest currently used sequence number (as used by the backfiller). */
+  long getHighestSequenceNumber();
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
@@ -87,22 +87,16 @@ public interface IndexManager {
   /** Returns the documents that match the given target based on the provided index. */
   Set<DocumentKey> getDocumentsMatchingTarget(FieldIndex fieldIndex, Target target);
 
-  /**
-   * Returns the next collection group to update. Returns {@code null} if no group exists or the
-   * next collection group is already at {@code maxSequenceNumber}.
-   */
+  /** Returns the next collection group to update. Returns {@code null} if no group exists. */
   @Nullable
-  String getNextCollectionGroupToUpdate(long maxSequenceNumber);
+  String getNextCollectionGroupToUpdate();
 
-  /** Updates the sequence number for the collection group and sets its latest read time. */
-  void updateCollectionGroup(String collectionGroup, long sequenceNumber, SnapshotVersion readTime);
+  /** Sets the collection group's latest read time. */
+  void updateCollectionGroup(String collectionGroup, SnapshotVersion readTime);
 
   /**
    * Updates the index entries for the provided documents and corresponding field indexes until the
    * cap is reached.
    */
   void updateIndexEntries(Collection<Document> documents);
-
-  /** Returns the largest currently used sequence number (as used by the backfiller). */
-  long getHighestSequenceNumber();
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
@@ -91,7 +91,13 @@ public interface IndexManager {
   @Nullable
   String getNextCollectionGroupToUpdate();
 
-  /** Sets the collection group's latest read time. */
+  /**
+   * Sets the collection group's latest read time.
+   *
+   * <p>This method updates the read time for all field indices for the collection group and
+   * increments their sequence number. Subsequent calls to {@link #getNextCollectionGroupToUpdate()}
+   * will return a different collection group (unless only one collection group is configured).
+   */
   void updateCollectionGroup(String collectionGroup, SnapshotVersion readTime);
 
   /**
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
@@ -102,7 +102,8 @@ final class SQLiteIndexManager implements IndexManager {
     // on how up to date the index is.
     db.query(
             "SELECT index_id, sequence_number, read_time_seconds, read_time_nanos "
-                + "FROM index_state WHERE uid = ?").binding(uid)
+                + "FROM index_state WHERE uid = ?")
+        .binding(uid)
         .forEach(
             row -> {
               int indexId = row.getInt(0);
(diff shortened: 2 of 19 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -365,45 +364,39 @@ private void createV1RemoteDocumentCache() {
    */
   private void createFieldIndex() {
     ifTablesDontExist(
-        new String[] {"index_configuration", "index_entries"},
+        new String[] {"index_configuration", "index_state", "collection_state", "index_entries"},
```
thebrianchen, line 367: remove `collection_state`

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -365,45 +364,39 @@ private void createV1RemoteDocumentCache() {
...
-                  + "update_time_nanos INTEGER, "
                   + "PRIMARY KEY (index_id))");
 
-          // The index entries table only has a single primary index. `array_value` should be set
-          // for all queries.
+          // Per index per user state to track the backfill state for each index
+          db.execSQL(
+              "CREATE TABLE index_state ("
+                  + "uid TEXT, "
+                  + "index_id INTEGER, " // Name of the collection group.
```
thebrianchen, line 381: you mean the index id of the field index right?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -87,14 +87,16 @@
   /** Returns the documents that match the given target based on the provided index. */
   Set<DocumentKey> getDocumentsMatchingTarget(FieldIndex fieldIndex, Target target);
 
-  /** Returns the next collection group to update. */
+  /** Returns the next collection group to update. Returns {@code null} if no group exists. */
   @Nullable
-  String getNextCollectionGroupToUpdate(Timestamp lastUpdateTime);
+  String getNextCollectionGroupToUpdate();
+
+  /** Sets the collection group's latest read time. */
```
thebrianchen, line 94: Update comment to include the fact that it updates all field indexes' index state that match the collection group + include the part where it updates the sequence counter as well

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -147,117 +169,51 @@ public void addToCollectionParentIndex(ResourcePath collectionPath) {
...
         nextIndexId,
         index.getCollectionGroup(),
-        encodeSegments(index),
-        index.getUpdateTime().getTimestamp().getSeconds(),
-        index.getUpdateTime().getTimestamp().getNanoseconds());
-
-    // TODO(indexing): Use a 0-counter rather than the timestamp.
-    setCollectionGroupUpdateTime(index.getCollectionGroup(), SnapshotVersion.NONE.getTimestamp());
-
+        encodeSegments(index));
```
thebrianchen, line 184: When we add a new field index, we will want to create an index_state entry right away if it doesn't exist. Otherwise, the index only lives in the memoized memory table and not the persistence table. We currently fetch the index configurations and merge it with the stored index_states, but could we simplify the logic if we wrote new field indexes to the index_state table right away?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -365,45 +364,39 @@ private void createV1RemoteDocumentCache() {
...
+                  + "sequence_number INTEGER, " // Specifies the order of updates
+                  + "read_time_seconds INTEGER, " // Time of last index backfill update
```
thebrianchen, line 383: nit: be more specific with comment: last processed RDC read time of last backfill update 

currently, there's some confusion with which time is being stored

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -365,45 +364,39 @@ private void createV1RemoteDocumentCache() {
...
+                  + "read_time_nanos INTEGER, "
+                  + "PRIMARY KEY (uid, index_id))");
+
+          // The index entries stores the encoded entries for all fields. The table only has a
```
thebrianchen, line 387: nit: The index entries table*

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -365,45 +364,39 @@ private void createV1RemoteDocumentCache() {
...
-    ifTablesDontExist(
-        new String[] {"collection_group_update_times"},
-        () -> {
           db.execSQL(
-              "CREATE TABLE collection_group_update_times ("
-                  + "collection_group TEXT, " // Name of the collection group.
-                  + "update_time_seconds INTEGER," // Time of last index backfill update
-                  + "update_time_nanos INTEGER,"
-                  + "PRIMARY KEY (collection_group))");
+              "CREATE INDEX read_time ON remote_documents(read_time_seconds, read_time_nanos)");
```
thebrianchen, line 399: Can we just add the read_time index on the table with out prepopulating the fields?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
```diff
@@ -365,45 +364,39 @@ private void createV1RemoteDocumentCache() {
    */
   private void createFieldIndex() {
     ifTablesDontExist(
-        new String[] {"index_configuration", "index_entries"},
+        new String[] {"index_configuration", "index_state", "collection_state", "index_entries"},
         () -> {
-          // TODO(indexing): Do we need to store a different update time per user? We need to ensure
-          // that we index mutations entries for all users.
+          // Global configuration for all existing indices
```
thebrianchen, line 369: s/indices/field indexes

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -88,21 +95,36 @@ public void start() {
...
     }
 
+    Map<Integer, FieldIndex.IndexState> indexStates = new HashMap<>();
+
+    // Fetch all index states if persisted for the user.
     db.query(
-            "SELECT index_id, collection_group, index_proto, update_time_seconds, "
-                + "update_time_nanos FROM index_configuration")
+            "SELECT index_id, sequence_number, read_time_seconds, read_time_nanos "
+                + "FROM index_state WHERE uid = ?")
```
thebrianchen, line 103: Don't you need a binding for uid? Surprised the tests pass.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -147,117 +169,51 @@ public void addToCollectionParentIndex(ResourcePath collectionPath) {
   public void addFieldIndex(FieldIndex index) {
     hardAssert(started, "IndexManager not started");
 
-    int nextIndexId = memoizedMaxId + 1;
+    int nextIndexId = memoizedMaxIndexId + 1;
```
thebrianchen, line 172: I am confused about how we initialize memoizedIndexId. It's currently initialized to 1, but where do we read the current max value from the table? FWIU, we're setting new field indexes to indexid 0 across restarts.

I have the same question for memoizedMaxSequenceNumber. I'm not sure where we're reading the initial value from persistence.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -88,21 +95,36 @@ public void start() {
...
+
+    // Fetch all index states if persisted for the user.
     db.query(
-            "SELECT index_id, collection_group, index_proto, update_time_seconds, "
-                + "update_time_nanos FROM index_configuration")
+            "SELECT index_id, sequence_number, read_time_seconds, read_time_nanos "
+                + "FROM index_state WHERE uid = ?")
+        .forEach(
+            row -> {
+              int indexId = row.getInt(1);
```
thebrianchen, line 106: row.getInt(0) maps to indexid

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -88,21 +95,36 @@ public void start() {
...
             row -> {
               try {
                 int indexId = row.getInt(0);
                 String collectionGroup = row.getString(1);
                 List<FieldIndex.Segment> segments =
                     serializer.decodeFieldIndexSegments(Index.parseFrom(row.getBlob(2)));
-                SnapshotVersion updateTime =
-                    new SnapshotVersion(new Timestamp(row.getLong(3), row.getInt(4)));
-
+                FieldIndex.IndexState indexState =
```
thebrianchen, line 122: optional: can you add a comment here that describes why we need this?

Initially I thought it was strange that we didn't just set the index state directly when we created a field index, and instead had this misdirection and check. After thinking about it, i realized that a new user might have not have an index state created for the existing field index, which is why we need this, but I'm afraid I'll forget this by next week.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
thebrianchen: Update, realized we take care of this in memoizeIndex(). Could you document that somewhere? The call to `memoizeIndex` is hidden  deep inside `start()`, and it'd be helpful to clarify that the memoized max counter values are only calculated when we memoize.

However, this also creates a potential problem. What happens if a user's max index id or seq number is not the global max for the table? won't we start overriding the existing counter values? I think it is safer to check for the memoized max ids as separate SELECT MAX calls from the table during start().

thebrianchen (changes requested):
First pass. Haven't looked at the tests yet, but had some initial feedback for the implementation + some questions.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
thebrianchen: Not sure if it is within scope for this PR, but we need a TODO somewhere to write tests for user changes.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteSchema.java
schmidt-sebastian: Done

thebrianchen (approved):
lgtm, thanks for explaining some of the questions i had

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -663,14 +609,27 @@ private boolean isInFilter(Target target, FieldPath fieldPath) {
...
+      FieldIndex updatedIndex =
+          FieldIndex.create(
+              fieldIndex.getIndexId(),
+              fieldIndex.getCollectionGroup(),
+              fieldIndex.getSegments(),
+              FieldIndex.IndexState.create(memoizedMaxSequenceNumber, readTime));
+      db.execute(
+          "REPLACE INTO index_state (index_id, uid,  sequence_number, "
+              + "read_time_seconds, read_time_nanos) VALUES(?, ?, ?, ?, ?)",
+          collectionGroup,
```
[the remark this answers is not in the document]
schmidt-sebastian: Holy !$#1! So there was a test for this that should have caught this - the `testChangeUser()`. The only problem is that it called `updateCollectionGroup` with the first index. When we then call `start()` we call `row.getInt()`. The value is actually `coll1`, but `getInt()` returns `0`, which happens to be the right value. I changed the test to modify the second index instead.

Thanks so much for spotting!

PR #3157, opened 2021-11-20
Don't initialize tracer until device is unlocked.

vkryachko:


firebase-datatransport/src/main/java/com/google/firebase/datatransport/TransportRegistrar.java
```diff
@@ -22,21 +22,23 @@
...
         Component.builder(TransportFactory.class)
             .add(Dependency.required(Context.class))
             .factory(
                 c -> {
                   TransportRuntime.initialize(c.get(Context.class));
                   return TransportRuntime.getInstance().newFactory(CCTDestination.LEGACY_INSTANCE);
                 })
-            .build());
+            .build(),
+        LibraryVersionComponent.create("fire-transport", BuildConfig.VERSION_NAME));
```
yifanyang, line 42: Is this related to this pull request?

yifanyang (approved):

firebase-datatransport/src/main/java/com/google/firebase/datatransport/TransportRegistrar.java
vkryachko: Not exactly, but it makes us to measure how long firelog takes to start, will keep it for now.

PR #3161, opened 2021-11-23
[DO NOT SUBMIT] Overlay performance tests

wu-hui:
_All data are from the average of three runs._

As expected, with overlay the query performance stays constant as the number of local mutations grows; while without overlay, the query performance grows linearly as the number of local mutations grows.

**Without Overlay**

Write 10,000 mutations, each document has 1 mutation: 11614ms
Run 100 queries: **3932ms**

Write 40,000 mutations, each document has 4 mutations: 57618ms
Run 100 queries: **7987ms**

Write 70,000 mutations, each document has 7 mutations: 108599ms
Run 100 queries: **11798ms**

Write 100,000 mutations, each document has 10 mutations: 168217ms
Run 100 queries: **15406ms**

**With Overlay**

Write 10,000 mutations, each document has 1 mutation: 15327ms
Run 100 queries: **4493ms**

Write 40,000 mutations, each document has 4 mutations: 67575ms
Run 100 queries: **4278ms**

Write 70,000 mutations, each document has 7 mutations: 123661ms
Run 100 queries: **4767ms**

Write 100,000 mutations, each document has 10 mutations: 185487ms
Run 100 queries: **4699ms**

### DIFF
--- a/firebase-firestore/firebase-firestore.gradle
+++ b/firebase-firestore/firebase-firestore.gradle
@@ -168,6 +168,7 @@ dependencies {
     annotationProcessor 'com.google.auto.value:auto-value:1.6.5'
 
     testImplementation 'junit:junit:4.12'
+    testImplementation 'com.github.javafaker:javafaker:1.0.2'
     testImplementation 'androidx.test:core:1.2.0'
     testImplementation 'org.mockito:mockito-core:2.25.0'
     testImplementation ("org.robolectric:robolectric:$robolectricVersion") {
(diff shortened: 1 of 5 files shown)


wu-hui:
Note: there is also an increase about 30% on mutation write time, because the SDK now writes more data with local mutations. Things might improve when we remove base mutations, we should re-run this test when they are removed.

wu-hui:
RAW log attache
[performance.txt](https://github.com/firebase/firebase-android-sdk/files/7585293/performance.txt)
d.

PR #3166, opened 2021-11-23
Guard mutable fields with synchronization blocks or methods

rachaprince:
Makes our synchronization patterns more consistent by:

- guarding mutable fields with either a synchronization block or method
- if the object is written asynchronously by our codebase, we use a synchronized block to prevent memory inconsistencies after the calling method returns. 
-avoiding the use of synchronized blocks within synchronized methods, because this increases the chances of deadlock

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
@@ -186,7 +186,7 @@ public class FirebaseAppDistribution {
    * sign in UI
    */
   @NonNull
-  public Task<AppDistributionRelease> checkForNewRelease() {
+  public synchronized Task<AppDistributionRelease> checkForNewRelease() {
     if (cachedCheckForNewReleaseTask != null && !cachedCheckForNewReleaseTask.isComplete()) {
       LogWrapper.getInstance().v("Response in progress");
       return cachedCheckForNewReleaseTask;
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
@@ -23,6 +23,7 @@ import android.app.AlertDialog;
 import android.content.Context;
 import androidx.annotation.GuardedBy;
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.common.internal.Preconditions;
 import com.google.android.gms.tasks.Task;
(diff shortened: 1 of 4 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -126,8 +131,8 @@ public static FirebaseAppDistribution getInstance(@NonNull FirebaseApp app) {
    * to complete the download and installation.
    */
   @NonNull
-  public synchronized UpdateTask updateIfNewReleaseAvailable() {
-    synchronized (updateTaskLock) {
+  public UpdateTask updateIfNewReleaseAvailable() {
```
rachaprince, line 134: I spoke offline with @mannyjimenez0810 and decided to remove the synchronized methods in this class because:

- Using a synchronized block within a synchronized method increases the risk of deadlock
- I'm not sure how much benefit the synchronized method gives us in our use case. The developer can only create one task per-public method because we return the same task if the task is not complete.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/InstallApkClient.java
```diff
@@ -81,7 +81,7 @@ void handleAppResume(Activity activity) {
     }
   }
 
-  Task<Void> installApk(String path) {
+  synchronized Task<Void> installApk(String path) {
```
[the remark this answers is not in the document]
rachaprince: Nope! This was a leftover from a previous draft. Good catch

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -180,7 +186,7 @@ public synchronized UpdateTask updateIfNewReleaseAvailable() {
    * sign in UI
    */
   @NonNull
-  public synchronized Task<AppDistributionRelease> checkForNewRelease() {
+  public Task<AppDistributionRelease> checkForNewRelease() {
     if (cachedCheckForNewReleaseTask != null && !cachedCheckForNewReleaseTask.isComplete()) {
```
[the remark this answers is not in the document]
rachaprince: Yes, another good catch.

I wonder if we could refactor this so we only store cachedCheckForNewRelease in the checkForNewReleaseClient. It feels strange that we store a reference to the task in two places

mannyjimenez0810 (approved):

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
pranavrajgopal: 💯

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/TesterSignInClient.java
```diff
@@ -133,16 +139,24 @@ private void onActivityDestroyed(Activity activity) {
               ErrorMessages.APP_BACKGROUNDED,
               FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE));
     }
-    signInTaskCompletionSource = new TaskCompletionSource<>();
+
+    synchronized (signInTaskLock) {
```
pranavrajgopal, line 143: There is a possible race condition of two threads calling this simultaneously. In that case both threads might bypass the "isCurrentlySigningIn" check above and the second thread might overwrite the first thread and probably induce a weird UX condition. Its probably rare but since we removed method synchronization, it's possible. 

I think a fix for this is to make sure that isSigningIn and re-assignment of the mutable variable to be in the same synchronization block.

pranavrajgopal: If you look at `cachedUpdateIfNewReleaseTask` above, we do something similar where the check and the assignment happens in the same synchronization block.

pranavrajgopal (approved):

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/TesterSignInClient.java
rachaprince: Great catch! Updated

PR #3167, opened 2021-11-23
Fad removing update app functionality to firebaseAppDistribution class

mannyjimenez0810:


firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -220,7 +223,8 @@ public void updateToNewRelease_whenNewAabReleaseAvailable_showsUpdateDialog() {
     AppDistributionReleaseInternal newRelease = TEST_RELEASE_NEWER_AAB_INTERNAL.build();
     when(mockCheckForNewReleaseClient.checkForNewRelease()).thenReturn(Tasks.forResult(newRelease));
     firebaseAppDistribution.setCachedNewRelease(newRelease);
-    when(mockUpdateAppClient.updateApp(newRelease, true)).thenReturn(new UpdateTaskImpl());
+    //  when(firebaseAppDistribution.updateApp(true)).thenReturn(new UpdateTaskImpl());
```
rachaprince, line 226: nit: remove comment

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -220,7 +223,8 @@ public void updateToNewRelease_whenNewAabReleaseAvailable_showsUpdateDialog() {
...
+    doReturn(new UpdateTaskImpl()).when(firebaseAppDistribution).updateApp(true);
```
rachaprince, line 227: Instead of mocking the updateApp method in firebaseAppDistribution, could we set the return value of the mock updateAAB client? Mocking the updateApp method  means that we don't test the code in the class under test

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -376,7 +380,7 @@ public void updateToNewRelease_receiveProgressUpdateFromUpdateApp() {
     firebaseAppDistribution.setCachedNewRelease(newRelease);
 
     UpdateTaskImpl mockTask = new UpdateTaskImpl();
-    when(mockUpdateAppClient.updateApp(newRelease, true)).thenReturn(mockTask);
+    when(firebaseAppDistribution.updateApp(true)).thenReturn(mockTask);
```
rachaprince, line 383: Same request here, can we mock the return value of the AAB client?

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/UpdateAppClientTest.java
```diff
@@ -1,144 +0,0 @@
-// Copyright 2021 Google LLC
```
rachaprince, line 1: A lot of tests were deleted here. Should we move any of them to the FirebaseAppDistributionTest class? I'm particularly interested in preserving the test coverage provided by updateAppTask_whenNoReleaseAvailable_throwsError

rachaprince (changes requested):

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/UpdateAppClientTest.java
mannyjimenez0810: Added in the tests back in. Decided to keep all of them for now.

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -415,4 +436,69 @@ public void taskCancelledOnScreenRotation() {
...
+    firebaseAppDistribution.setCachedNewRelease(newRelease);
+
+    UpdateTask updateTask1 = firebaseAppDistribution.updateApp(true);
+    UpdateTask updateTask2 = firebaseAppDistribution.updateApp(true);
+
+    assertEquals(updateTask1, updateTask2);
+  }
+
+  @Test
+  public void updateApp_whenCalledMultipleTimesWithApk_returnsSameUpdateTask() {
```
rachaprince, line 453: I think this test and the won above it might belong in the updateAAB and updateAPK clients, respectively. I don't see a code path in FirebaseAppDistribution where we check for a cached update task

rachaprince (approved):

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
mannyjimenez0810: Ah since this was in the updateApp tests I moved it over. I didn't see updateApp checking for a cached update task either. Will move the tests to reflect this.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -232,17 +237,35 @@ public UpdateTask updateApp() {
    * Overloaded updateApp with boolean input showDownloadInNotificationsManager. Set to true for
    * basic configuration and false for advanced configuration.
    */
-  private UpdateTask updateApp(boolean showDownloadInNotificationManager) {
+  public UpdateTask updateApp(boolean showDownloadInNotificationManager) {
```
mannyjimenez0810, line 240: Just wanted to call out that this is public now. Since updateApp was public as a class and now that functionality is here it makes sense to me to make it public. Thoughts?

rachaprince: this should still be private. This method overload with the boolean parameter should only be called by our SDK. External packages should call the overload above with no parameters.

I don't think UpdateApp should have been a public class. That was an oversight on our part. The public/private audit going forward will probably make most classes package private

mannyjimenez0810: That makes sense. Will change back.

pranavrajgopal (approved):
Looks great! (agree with Rachel on the private / public comment :) )

PR #3169, opened 2021-11-23
Random changes for Indexing

schmidt-sebastian:
Some changes I made during debugging.


firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -65,29 +68,6 @@ public void setIndexManager(IndexManager indexManager) {
     this.indexManager = indexManager;
   }
 
-  public static class Results {
-    private final boolean hasRun;
-
-    private final int documentsProcessed;
```
schmidt-sebastian, line 71: FYI I don't think we will ever differentiate between documentsWritten and documentsDeleted as they are handled interchangeably by IndexManager.

thebrianchen (approved):

PR #3173, opened 2021-11-24
Update remote_documents table to add dedicated parent_path column

schmidt-sebastian:
The current table design reads all documents from subcollections as we scan a collection path. This makes it impossible to fetch the first N documents by using "Limit N", which we would like to do for the Index backfill.


schmidt-sebastian:
I am working on performance data but don't have anything for now.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
```diff
@@ -57,15 +58,16 @@ public void add(MutableDocument document, SnapshotVersion readTime) {
...
     MessageLite message = serializer.encodeMaybeDocument(document);
 
     db.execute(
         "INSERT OR REPLACE INTO remote_documents "
-            + "(path, read_time_seconds, read_time_nanos, contents) "
-            + "VALUES (?, ?, ?, ?)",
-        path,
+            + "(collection_path, document_id, read_time_seconds, read_time_nanos, contents) "
+            + "VALUES (?, ?, ?, ?, ?)",
+        collectionForKey(documentKey),
```
wu-hui, line 69: These are repeated enough that I think we can add them to `DocumentKey`, like `key.collection()` and `key.documentId()` ?

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteSchemaTest.java
```diff
@@ -571,6 +571,35 @@ public void rewritesCanonicalIds() {
             });
   }
 
+  @Test
+  public void rewritesDocumentKeys() {
```
wu-hui, line 575: I'd like to also have a test where the number of documents to migrate exceeds the batch size, to make sure batching works.

wu-hui (commented):
Looks good. 

How was the performance of the migration?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
schmidt-sebastian: Done

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteSchemaTest.java
schmidt-sebastian: Done

schmidt-sebastian:
I had to update the PR to add a dedicated parent_path column but keep the other column as is. This allows schema downgrades, as older clients expect `path` to exist. The small additional benefit that this brings is that the schema migration is now a lot less performance intensive and should be okay to ship as we have done in the past. I will provide performance data by the end of the day.

schmidt-sebastian:
FYI:

```
Migration from version 12 to 13 took 408 milliseconds
```

This is for 1775 docs.

schmidt-sebastian:
FYI there is a new version of this PR: https://github.com/firebase/firebase-android-sdk/pull/3192/files

PR #3175, opened 2021-11-25
String format with decimal issue in Android 8.1.0

argzdev:
Speculative fix for issue #3084. NullPointerException is encountered specifically on Android version 8.1.0, this is due to String.format inserted with Double values.

The proposed fix is to use DecimalFormat to format the values before being inserted in String.format


argzdev:
Apologies, incorrect usage author when committing, I'll revise this and send another pull request. This is so we won't have any issues with the CLA.

PR #3180, opened 2021-11-25
Add qualifier support to firebase components.

vkryachko:
Details: go/firebase-component-qualifiers


emilypgoogle (approved):

firebase-components/src/main/java/com/google/firebase/components/Qualified.java
```diff
@@ -0,0 +1,64 @@
+// Copyright 2021 Google LLC
```
rlazo, line 1: 2022

firebase-components/src/main/java/com/google/firebase/components/Qualified.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public static <T> Qualified<T> qualified(Class<? extends Annotation> qualifier, Class<T> type) {
+    return new Qualified<>(qualifier, type);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
```
rlazo, line 42: `instanceof` could save the null check, but I'm ok either way since the class is final.

firebase-components/src/main/java/com/google/firebase/components/Qualified.java
```diff
@@ -0,0 +1,64 @@
...
+  public int hashCode() {
+    int result = type.hashCode();
+    result = 31 * result + qualifier.hashCode();
+    return result;
+  }
+
+  @Override
+  public String toString() {
+    if (qualifier == Unqualified.class) {
+      return type.getName();
```
rlazo, line 60: Hard code it to `Unqualified.class.getName()` maybe?

rlazo (changes requested):

firebase-components/src/main/java/com/google/firebase/components/Qualified.java
vkryachko: ack

firebase-components/src/main/java/com/google/firebase/components/Qualified.java
vkryachko: The idea here is to "hide" the fact that `Unqualified` exists, so any unqualified component would render as `TypeName`, while qualified ones would render as `@Hello TypeName`, wdyt?

rlazo (approved):

PR #3182, opened 2021-11-26
Fix perf's javadoc.

vkryachko:


VinayGuthal (approved):

visumickey:
Thanks for making the change @vkryachko

PR #3185, opened 2021-11-29
WIP: OR Queries

ehsannas:


ehsannas:
rebased.

ehsannas:
`Task :firebase-app-distribution:apiInformation FAILED`... Looks like tot is broken

ehsannas:
posted a fix via https://github.com/firebase/firebase-android-sdk/pull/3213

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+        if (filter.matches(doc)) {
+          return true;
+        }
+      }
+      return false;
+    }
+  }
+
+  @Override
+  public String getCanonicalId() {
```
ehsannas, line 129: Sebastian commented that I should updated this function to be consistent with existing canonical IDs. I haven't done that yet.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+
+  /**
+   * Performs a depth-first search to find and return the first FieldFilter in the composite filter
+   * that satisfies the condition. Returns null if none of the FieldFilters satisfy the condition.
+   */
+  public FieldFilter firstFieldFilterWhere(FieldFilterCondition condition) {
+    for (Filter filter : filters) {
+      if (filter instanceof FieldFilter && condition.check(((FieldFilter) filter))) {
+        return (FieldFilter) filter;
+      } else if (filter instanceof CompositeFilter) {
```
schmidt-sebastian, line 89: Can we change this logic so that it does not have to do an instanceof check? Every Filter could have a `firstFieldFilterWhere` style method. Most will just return `this` or `null`.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null || !(o instanceof CompositeFilter)) {
+      return false;
+    }
+    CompositeFilter other = (CompositeFilter) o;
+    // Note: This comparison requires order of filters in the list to be the same, and it does not
+    // remove duplicate subfilters from each composite filter. It is therefore way less expensive.
+    return operator == other.operator && filters.equals(other.filters);
```
schmidt-sebastian, line 156: This is interesting and has some wide-reaching implications:

`where(or(filterA, filterB))` is now different from `where(or(filterB, filterA))`. This means that we will now create two different targets for these queries, which adds additional costs for the customer. It would be nice if we could get the canonicalId and the `equals()` to match in this case.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+    for (Filter filter : filters) {
+      result += filter.hashCode();
+    }
+    return result;
+  }
+
+  /**
+   * Returns a new composite filter that contains all filter from `this` plus all the given filters
+   */
+  public CompositeFilter addFilters(List<Filter> otherFilters) {
```
schmidt-sebastian, line 172: `withAddedFilters` since you are returning a new object

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+          // compositeSubfilter: (B | C)
+          // Result: (A | B | C)
+          newSubfilters.addAll(compositeSubfilter.getFilters());
+        } else {
+          // Filter: (A | (B & C))
+          // compositeSubfilter: (B & C)
+          // Result: (A | (B & C))
+          newSubfilters.add(subfilter);
+        }
+      }
```
schmidt-sebastian, line 233: Would removing this code change anything about the final DNF?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FieldFilter.java
```diff
@@ -15,37 +15,81 @@
...
+
+  /**
+   * Creates a new filter that compares fields and values. Only intended to be called from
+   * Filter.create(). This is a more specialized constructor that takes a parsed Value.
+   */
   protected FieldFilter(FieldPath field, Operator operator, Value value) {
     this.field = field;
     this.operator = operator;
     this.value = value;
+    this.valueObject = null;
```
schmidt-sebastian, line 86: As stated elsewhere, I hope that we can refactor this so the `valueObject` stays in the API layer.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -134,6 +134,9 @@ public boolean isCollectionGroupQuery() {
   /**
    * Returns true if this query does not specify any query constraints that could remove results.
    */
+  // TODO(ehsann): If we do allow composite filters with zero subfilters, then this should be
+  // updated.
+  // TODO(ehsann): How do we handle conditions that are always true? (e.g. age > 0 || age <= 0).
```
schmidt-sebastian, line 139: We should not write any code to optimize this.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -51,6 +53,11 @@
   private final @Nullable Bound startAt;
   private final @Nullable Bound endAt;
 
+  // DNF (Disjunctive Normal Form) representation of this target's filters.
+  // Each element in this list is an AND filter of FieldFilters.
+  // The query result is the disjunction (OR) of each of these AND filters.
+  private final List<CompositeFilter> dnf;
```
schmidt-sebastian, line 59: The Target is essentially the proto representation. We should not store the DNF here.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -75,14 +76,29 @@
...
+   *
+   * @param target The target for which we are looking for an index.
+   * @param andFilter The AND filter within the target's constraints for which we're looking for an
+   *     index. This can be null if the target has no filters.
+   * @return An index that can be used to serve the provided target's conjunction filter. Returns
+   *     {@code null} if no index is configured.
    */
   @Nullable
-  FieldIndex getFieldIndex(Target target);
+  FieldIndex getFieldIndex(Target target, @Nullable CompositeFilter andFilter);
```
schmidt-sebastian, line 88: The reason this method exists is just so that the QueryEngine can quickly short-circuit Index-based execution if no index is defined. The callsite is:

```
FieldIndex i = getFIeldIndex();
if (i != null) {...}
```

It seems like this now adds some complication as we need to pass more state between `getFieldIndex` and `getDocumentsMatchingTarget`. We should probably revisit how this works to minimize these interactions (i.e. add a `hasMatchingIndex()` method or make `getDocumentsMatchingTarget()` nullable).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexedQueryEngine.java
```diff
@@ -69,18 +71,64 @@ public void setIndexManager(IndexManager indexManager) {
...
+    }
+
+    // If there's no limit constraint, we can return all the results.
+    // If the DNF contained only 1 term, the SQLite query has enforced the limit, and we can return
+    // all the results.
+    if (target.getLimit() == -1 || target.getDnf().size() == 1) {
+      return result;
+    }
+
+    // If there's a limit constraint, we should perform all branches of the query (as done above),
```
schmidt-sebastian, line 105: The limit handling should be pushed into SQLite.

schmidt-sebastian (commented):
Some high level comments sprinkled with random nits. Overall, the functionality is in a good place. It probably makes sense to split this up into multiple PRs soon so that we can review smaller units in greater detail and get them merged.

firebase-firestore/src/main/java/com/google/firebase/firestore/Filter.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  @NonNull
+  public static Filter arrayContainsAny(@NonNull FieldPath fieldPath, @Nullable Object value) {
+    return FieldFilter.create(
+        fieldPath.getInternalPath(), FieldFilter.Operator.ARRAY_CONTAINS_ANY, value);
+  }
+
+  @NonNull
+  public static Filter inList(@NonNull String field, @Nullable Object value) {
```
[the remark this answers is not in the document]
ehsannas: Acknowledged

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -388,6 +389,26 @@ public Query whereNotIn(@NonNull FieldPath fieldPath, @NonNull List<? extends Ob
     return whereHelper(fieldPath, Operator.NOT_IN, values);
```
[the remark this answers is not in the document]
ehsannas: yeah. Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -388,6 +389,26 @@ public Query whereNotIn(@NonNull FieldPath fieldPath, @NonNull List<? extends Ob
...
+   * TODO(ehsann): describe.
+   *
+   * @param filters
+   * @return
+   */
+  @NonNull
+  public Query where(@NonNull Filter... filters) {
+    // If no filters were provided, we can return the current query.
+    // TODO (ehsann): Maybe we shouldn't accept zero filters.
+    if (filters.length == 0) {
```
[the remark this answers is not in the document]
ehsannas: Acknowledged

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -400,51 +421,18 @@ public Query whereNotIn(@NonNull FieldPath fieldPath, @NonNull List<? extends Ob
...
+    FieldFilter filter = FieldFilter.create(fieldPath.getInternalPath(), op, value);
+    return new Query(query.filter(parseFieldFilterValue(filter)), firestore);
   }
 
   /**
    * Parses the given documentIdValue into a ReferenceValue, throwing appropriate errors if the
    * value is anything other than a DocumentReference or String, or if the string is malformed.
    */
-  private Value parseDocumentIdValue(Object documentIdValue) {
+  private Value parseDocumentIdValue(
```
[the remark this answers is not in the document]
ehsannas: yeah, this should be reverted. I moved this code to Filter and then moved it back here, and forgot to change the signature back to its original form.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -481,110 +469,80 @@ private Value parseDocumentIdValue(Object documentIdValue) {
...
-      throw new IllegalArgumentException(
-          "Invalid Query. '"
-              + op.toString()
-              + "' filters support a maximum of 10 elements in the value array.");
+  /**
+   * Parses the given filter's value from Object type to Value type.
+   *
+   * @return A new field filter that contains the parsed Value.
+   */
+  private FieldFilter parseFieldFilterValue(FieldFilter filter) {
```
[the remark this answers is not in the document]
ehsannas: I'll try this suggestion.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -652,7 +610,7 @@ private Query orderBy(
           "Invalid query. You must not call Query.endAt() or Query.endBefore() before "
               + "calling Query.orderBy().");
     }
-    validateOrderByField(fieldPath);
```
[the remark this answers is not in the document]
ehsannas: yes, this check is moved to core/Query.java. Generally there was a bunch of validation checks that really didn't belong to the API layer and should have been done in core/Query.java I believe. The addition of new filters made it more necessary to move them. I'll carve out the code moves as separate PRs that we can easily review.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+import androidx.annotation.Nullable;
+import com.google.firebase.firestore.Filter;
+import com.google.firebase.firestore.model.Document;
+import com.google.firestore.v1.StructuredQuery;
+import com.google.firestore.v1.StructuredQuery.CompositeFilter.Operator;
+import java.util.ArrayList;
+import java.util.List;
+
+interface FieldFilterCondition {
+  boolean check(FieldFilter field);
```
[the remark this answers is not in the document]
ehsannas: Ack. Will use `Function`.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/** Represents a filter that is the conjunction or disjunction of other filters. */
+public class CompositeFilter extends Filter {
```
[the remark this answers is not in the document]
ehsannas: Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+  private final List<Filter> filters;
+  //  private final boolean isAnd;
```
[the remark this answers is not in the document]
ehsannas: +1

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+
+  public List<Filter> getFilters() {
+    return filters;
+  }
+
+  public Operator getOperator() {
+    return operator;
+  }
+
+  public boolean isAnd() {
```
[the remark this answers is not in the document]
ehsannas: +1

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -0,0 +1,347 @@
...
+  public boolean isOr() {
+    // TODO(ehsann): Replace with Operator.OR.
+    return operator == Operator.OPERATOR_UNSPECIFIED;
+  }
+
+  /**
+   * Returns true if this is an AND filter and all the filters in this composite filter are
+   * FieldFilters (i.e. it does not contain any other composite filters).
+   */
+  public boolean isFlatAndFilter() {
```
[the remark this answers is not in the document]
ehsannas: Sure. I do see some value in having a completely separate class that handles DNF transformations.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
ehsannas: We should discuss this

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
ehsannas: Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
ehsannas: yeah, it's not possible to reach DNF form if associativity is not applied at every iteration. Without this code the final form may not be DNF and if it is, it may contain nested OR conditions which we don't want.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/FieldFilter.java
ehsannas: Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
ehsannas: Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -200,34 +203,68 @@ public FieldPath getFirstOrderByField() {
...
         FieldFilter fieldfilter = (FieldFilter) filter;
         if (fieldfilter.isInequality()) {
           return fieldfilter.getField();
         }
+      } else if (filter instanceof CompositeFilter) {
+        CompositeFilter compositeFilter = (CompositeFilter) filter;
+        FieldFilter found = compositeFilter.getInequalityFilter();
+        if (found != null) {
+          return found.getField();
+        }
```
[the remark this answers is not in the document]
ehsannas: Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -200,34 +203,68 @@ public FieldPath getFirstOrderByField() {
...
         if (operators.contains(filterOp)) {
           return filterOp;
         }
+      } else if (filter instanceof CompositeFilter) {
+        CompositeFilter compositeFilter = (CompositeFilter) filter;
+        FieldFilter fieldFilter =
+            compositeFilter.firstFieldFilterWhere(f -> operators.contains(f.getOperator()));
+        if (fieldFilter != null) {
+          return fieldFilter.getOperator();
+        }
```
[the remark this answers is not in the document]
ehsannas: Acknowledged.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
ehsannas: I see. Thinking about this more, I think we should be able to push it to SQLiteIndexManager.java as that's the only place where it's really needed.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
ehsannas: Yes, you got it exactly right. I think I know how to simplify it now.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexedQueryEngine.java
ehsannas: Acknowledged.

ehsannas (commented):
Thanks a lot for going through this PR. The PR is larger than I originally wanted it to be. For the most part I agree with your comments. I will start slicing this PR into smaller, more contained changes that can be reviewed/landed more easily. Coming soon to GitHub....

Thanks!

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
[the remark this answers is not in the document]
ehsannas: @schmidt-sebastian I think this wouldn't be a good idea because `FieldFilter` and `CompositeFilter` must inherit from `Filter`. If we add field and operator and value (as `Object`) in `com.google.firebase.fiterstore.Filter`, then `FieldFilter` and `CompositeFilter` will have those members too.

In this PR:
```
Filter Class (no members)
^    ^
|    |
|    FieldFilter Class. Members:
|    Field
|    Op
|    value (Value)
|    valueObject (Object)
|
CompositeFilter Class. Members:
Filter[]
logicType
```

If we keep `Object` in the API layer:
```
Filter Class. Members: Field, Op, valueObject (Object)
^    ^
|    |
|    FieldFilter Class. Members:
|    Field(by inheritance)
|    Op(by inheritance)
|    valueObject (Object) (by inheritance)
|    value (Value)
|
CompositeFilter Class. Members:
Field(by inheritance)
Op(by inheritance)
valueObject (Object) (by inheritance)
Filter[]
logicType
```

Another alternative is to create a third Filter class (E.g. `com.google.firebase.fiterstore.core.Filter`)
```
Filter Class (no members)
^   ^    ^
|   |    |
|   |    FieldFilter Class. Members:
|   |    Field
|   |    Op
|   |    value (Value)
|   | 
|   |
|   CompositeFilter Class. Members:
|   Filter[]
|   logicType
|
com.google.firebase.fiterstore.core.Filter Class. Members:
Field
Op
value (Object)
```
This is what I did initially (and sadly chose `UnqualifiedFilter` as the name instead of `core.Filter`).

What do you think? Did I misunderstand your suggestion?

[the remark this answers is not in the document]
schmidt-sebastian: I think the Web API has a good precedent for this: 

The API layer only knows about the `Object`: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/lite-api/query.ts#L132
This then gets translated to `Value` when we construct the internal Filter.

There is some duplication here (since you need to keep the filter operator) but the two classes are distinct from one another.

My main goal is to get the `Object` out of our core layer and into the API layer. If you add an hierarchy here that shares some members you might not achieve this goal.

PR #3196, opened 2021-12-01
Audit of classes in firebase app distribution to minimize user access

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/UpdateApkClient.java
```diff
@@ -35,7 +35,9 @@
 import java.util.jar.JarFile;
 import javax.net.ssl.HttpsURLConnection;
 
-/** Client class that handles updateApp functionality for APKs in {@link UpdateAppClient}. */
+/**
+ * Client class that handles updateApp functionality for APKs in {@link FirebaseAppDistribution}.
+ */
 class UpdateApkClient {
```
rachaprince, line 41: There are some constructors and methods in this class that are public and I think they could be package private

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/UpdateAabClient.java
```diff
@@ -29,7 +29,9 @@
 import java.util.concurrent.Executors;
 import javax.net.ssl.HttpsURLConnection;
 
-/** Client class that handles updateApp functionality for AABs in {@link UpdateAppClient}. */
+/**
+ * Client class that handles updateApp functionality for AABs in {@link FirebaseAppDistribution}.
+ */
 class UpdateAabClient {
```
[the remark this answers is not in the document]
rachaprince: I saw that this was marked as resolved, but it seems like there are still public methods in this class. I am curious if there is some reason we can't make the methods package private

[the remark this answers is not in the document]
mannyjimenez0810: Jumped the gun a bit. I marked resolved but didn't push the changes, should be in now.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ReleaseIdentificationUtils.java
```diff
@@ -29,7 +29,7 @@
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
-public final class ReleaseIdentificationUtils {
+final class ReleaseIdentificationUtils {
   private static final String TAG = "ReleaseIdentification";
   private static final int BYTES_IN_LONG = 8;
```
rachaprince, line 35: I think the methods in this class should also be package-private

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/SignInResultActivity.java
```diff
@@ -25,6 +25,8 @@
 public class SignInResultActivity extends AppCompatActivity {
   private static final String TAG = "SignInResultActivity:";
 
+  public SignInResultActivity() {}
```
rachaprince, line 28: Do we need this? Seems like the class is public already, so the default constructor will be public too

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/SignInResultActivity.java
```diff
@@ -25,6 +25,8 @@
 public class SignInResultActivity extends AppCompatActivity {
```
lfkellogg, line 25: Does this activity need to be public?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/InstallActivity.java
```diff
@@ -25,13 +25,16 @@
...
+ * Activity opened during installation in {@link FirebaseAppDistribution} after APK download is
+ * finished.
  */
-public class InstallActivity extends AppCompatActivity {
+class InstallActivity extends AppCompatActivity {
   private static final String TAG = "InstallActivity: ";
   private boolean unknownSourceEnablementInProgress = false;
   private boolean installInProgress = false;
 
+  public InstallActivity() {}
```
lfkellogg, line 36: Why do we need a public default constructor here?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/SignInResultActivity.java
mannyjimenez0810: This was a left over from trying to make it package-private will remove

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/SignInResultActivity.java
mannyjimenez0810: I believe it does because it is in the manifest. I tried making it package private and the app would close on start up without any error,  but changing it back to public fixes that issue. 

 https://github.com/firebase/firebase-android-sdk/blob/dd1c33608f3dfa143317b5652659989848742b47/firebase-app-distribution/src/main/AndroidManifest.xml#L37

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/SignInResultActivity.java
lfkellogg: I'm pretty sure even if the class is marked as public, the constructor won't be public unless it's also marked as such.

rachaprince: good to know! thanks!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/SignInResultActivity.java
lfkellogg: Ah! That makes sense. Thanks for the clarification.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/InstallActivity.java
mannyjimenez0810: This can be ignored. This must also be public mistake on my part.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ReleaseIdentificationUtils.java
mannyjimenez0810: Good call. I thought since the class is package-private it was fine but its better to be consistent.

lfkellogg (approved):
LGTM!

rachaprince (approved):
Nice work!

rachaprince:
Looking at the files generated by api.txt, I have a few questions about the visibility of some classes and methods:

@mannyjimenez0810 do you know if we could reduce the visibility of any of the following?:

- FirebaseAppDistributionLifeCycleNotifier
- FirebaseAppDistributionFileProvider
- UpdateProgress.Builder

(It might be worth doing a search for "public" in the app distribution package to see if there are any others we missed)

mannyjimenez0810:
@rachaprince Great Question.

FirebaseAppDistributionLifeCycleNotifier
This was an oversight on my part. Will fix in a future PR. 

FirebaseAppDistributionFileProvider
I tried making this also package private but I believe due to it being in the manifest I was unable to. 
See here https://github.com/firebase/firebase-android-sdk/blob/8bc2ab4f3bc7caeeabe6ea336b721f53b3abe4dd/firebase-app-distribution/src/main/AndroidManifest.xml#L49

I get this exception  java.lang.RuntimeException: Unable to get provider com.google.firebase.app.distribution.FirebaseAppDistributionFileProvider: java.lang.IllegalAccessException: java.lang.Class<com.google.firebase.app.distribution.FirebaseAppDistributionFileProvider> is not accessible from java.lang.Class<android.app.AppComponentFactory>
    

UpdateProgress.Builder
Another good catch. I saw we return UpdateProgress so I didn't look at the methods as carefully as I should have. Will also fix this in an upcoming PR.

Thanks for catching these. Will do a search for public in the package to see if theres other pieces I missed.

PR #3198, opened 2021-12-01
remove heartbeat unused apis

VinayGuthal:


firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -80,7 +80,6 @@ private DefaultHeartBeatInfo(Context context, Set<HeartBeatConsumer> consumers)
     return HeartBeat.NONE;
   }
 
-  @Override
   public Task<List<HeartBeatResult>> getAndClearStoredHeartBeatInfo() {
```
vkryachko, line 83: why is this method still needed? can't it be removed?

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatInfo.java
```diff
@@ -114,7 +113,6 @@ private DefaultHeartBeatInfo(Context context, Set<HeartBeatConsumer> consumers)
         });
   }
 
-  @Override
   public Task<Void> storeHeartBeatInfo(@NonNull String heartBeatTag) {
```
vkryachko, line 116: ditto

vkryachko (approved):
LGTM assuming tests pass

PR #3201, opened 2021-12-02
Optimize query execution by deferring query evaluation

schmidt-sebastian:
Our current query pipeline is essentially as such:

- The RemoteDocumentCache executes a query by scanning all documents in the collection. It reads all documents and removes documents that do not match the query.
- The LocalDocumentsView then fetches all mutations that possibly apply.
- If we fetch a mutation but don't have a document, we go back to the RemoteDocumentCache to fetch the documents that are missing - but these are the same documents that we filtered out in the initial step.

I think the idea behind this was that the RemoteDocumentCache uses the Query to filter the document it reads, but this code now lives in IndexManager. This PR therefore removes the "Query" functionality from the RemoteDocumentCache and instead just returns all documents from the collection.


firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
```diff
@@ -84,9 +85,9 @@ QueryEngine getSubject() {
 
   /**
    * Returns the number of documents returned by the RemoteDocumentCache's
-   * `getDocumentsMatchingQuery()` API (since the last call to `resetCounts()`)
+   * `getAll()` API (since the last call to `resetCounts()`)
    */
-  int getDocumentsReadByQuery() {
+  int getDocumentsReadByColllection() {
```
dconeybe, line 90: nit: Too many "L" characters in "getDocumentsReadByColllection"

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
```diff
@@ -84,9 +85,9 @@ QueryEngine getSubject() {
...
     return documentsReadByQuery[0];
```
dconeybe, line 91: Should the `documentsReadByQuery` instance variable be renamed to `documentsReadByCollection`, with only 2 L's? :P

firebase-firestore/src/main/java/com/google/firebase/firestore/local/RemoteDocumentCache.java
```diff
@@ -80,19 +79,13 @@
...
    *
-   * <p>Implementations may return extra documents if convenient. The results should be re-filtered
-   * by the consumer before presenting them to the user.
-   *
-   * <p>Cached entries for non-existing documents have no bearing on query results.
-   *
-   * @param query The query to match documents against.
+   * @param collection The collection to read.
    * @param offset The read time and document key to start scanning at (exclusive).
    * @return The set of matching documents.
```
dconeybe, line 86: nit: Update the return value javadocs to specify that the returned Map is a newly-created Map that the caller is free to modify. Previously, when it was `ImmutableSortedMap`, this was clear but now that it returns a mutable `Map` it's not clear without looking at the implementation whether or not it's a copy.

dconeybe, line 86: The return value javadocs say "The set of matching documents" but the word "matching" doesn't make sense anymore since it returns all documents. Please update.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
```diff
@@ -229,23 +226,9 @@ public MutableDocument get(DocumentKey documentKey) {
...
-    ImmutableSortedMap<DocumentKey, MutableDocument> matchingDocuments =
-        DocumentCollections.emptyMutableDocumentMap();
-    for (MutableDocument document : allDocuments.values()) {
-      if (document.isFoundDocument() && query.matches(document)) {
-        matchingDocuments = matchingDocuments.insert(document.getKey(), document);
-      }
-    }
-    return matchingDocuments;
+  public Map<DocumentKey, MutableDocument> getAll(
+      final ResourcePath collection, IndexOffset offset) {
```
dconeybe, line 230: nit: `final` on the `ResourcePath` argument isn't strictly required anymore. Can it be removed?

dconeybe (changes requested):
Looks good. Just a few small comments.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/RemoteDocumentCache.java
schmidt-sebastian: Good point. I adjusted the comment.

schmidt-sebastian: Removed

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
schmidt-sebastian: Removed (now it fits into one line!)

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
schmidt-sebastian: Lllol.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
schmidt-sebastian: Yes, thanks for catching.

dconeybe (approved):

PR #3202, opened 2021-12-02
Add mutation batch id to index backfill

thebrianchen:
This PR adds mutation batch id support to the index backfiller. All documents in the same mutation batch are processed, even if it goes over the limit (the alternative is to add an offset for mutations, but batches are limited to 500 writes).

Still todo:
- Current implementation performs a collection group query with documents + overlay, but local mutations are not written to the collection_groups table that support collection group queries. This means that the current implementation only supports local mutations that already have a remote document in the same collection.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/index/IndexEntry.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/index/IndexEntry.java
@@ -20,6 +20,7 @@ import static com.google.firebase.firestore.util.Util.nullSafeCompare;
 import com.google.auto.value.AutoValue;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.util.Util;
+import java.util.Comparator;
 
 /** Represents an index entry saved by the SDK in its local storage. */
 @AutoValue
@@ -53,4 +54,9 @@ public abstract class IndexEntry implements Comparable<IndexEntry> {
 
     return nullSafeCompare(getArrayValue(), other.getArrayValue(), Util::compareByteArrays);
   }
+
+  public static final Comparator<IndexEntry> SEMANTIC_COMPARATOR =
+      (left, right) -> {
+        return left.compareTo(right);
+      };
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlayCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlayCache.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.firestore.local;
 
+import android.util.Pair;
 import androidx.annotation.Nullable;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.ResourcePath;
@@ -54,4 +55,14 @@ public interface DocumentOverlayCache {
    *     change past `sinceBatchId` are returned.
    */
   Map<DocumentKey, Mutation> getOverlays(ResourcePath collection, int sinceBatchId);
+
+  /**
+   * Returns a mapping of overlays containing the largest batch id for each overlay mutation.
+   *
+   * @param collection The collection path to get the overlays for.
+   * @param sinceBatchId The minimum batch ID to filter by (exclusive). Only overlays that contain a
+   *     change past `sinceBatchId` are returned.
+   */
+  Map<DocumentKey, Pair<Integer, Mutation>> getOverlaysWithBatchId(
+      ResourcePath collection, int sinceBatchId);
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
@@ -137,23 +137,56 @@ public class IndexBackfiller {
 
   /** Writes entries for the fetched field indexes. */
   private int writeEntriesForCollectionGroup(
-      LocalDocumentsView localDocumentsView, String collectionGroup, int entriesRemainingUnderCap) {
+      LocalDocumentsView localDocumentsView,
+      String collectionGroup,
+      int documentsRemainingUnderCap) {
+    int documentsProcessed = 0;
     Query query = new Query(ResourcePath.EMPTY, collectionGroup);
 
     // Use the earliest offset of all field indexes to query the local cache.
-    IndexOffset existingOffset = getExistingOffset(indexManager.getFieldIndexes(collectionGroup));
+    Collection<FieldIndex> fieldIndexes = indexManager.getFieldIndexes(collectionGroup);
+    IndexOffset existingOffset = getExistingOffset(fieldIndexes);
 
     // TODO(indexing): Use limit queries to only fetch the required number of entries.
-    // TODO(indexing): Support mutation batch Ids when sorting and writing indexes.
     ImmutableSortedMap<DocumentKey, Document> documents =
         localDocumentsView.getDocumentsMatchingQuery(query, existingOffset);
 
-    List<Document> oldestDocuments = getOldestDocuments(documents, entriesRemainingUnderCap);
+    List<Document> oldestDocuments = getOldestDocuments(documents, documentsRemainingUnderCap);
     indexManager.updateIndexEntries(oldestDocuments);
+    documentsProcessed += oldestDocuments.size();
 
     IndexOffset newOffset = getNewOffset(oldestDocuments, existingOffset);
+
+    // Start indexing mutations only after catching up for read time. This must be done separately
+    // since the backfill's first pass is in read time order rather than mutation batch id order.
+    int documentsRemaining = documentsRemainingUnderCap - oldestDocuments.size();
+    if (documentsRemaining > 0) {
+      int earliestBatchId = getEarliestBatchId(fieldIndexes);
+      newOffset =
+          IndexOffset.create(newOffset.getReadTime(), newOffset.getDocumentKey(), earliestBatchId);
+      Map<Document, Integer> documentToBatchId =
+          localDocumentsView.getDocumentsBatchIdsMatchingCollectionGroupQuery(query, newOffset);
+
+      // Write index entries for documents touched by local mutations and update the offset to the
+      // new largest batch id.
+      if (!documentToBatchId.isEmpty()) {
+        List<Document> documentsByBatchId =
+            getDocumentsByBatchId(documentToBatchId, documentsRemaining);
+
+        // Largest batch id is the last element in documentsByBatchId since it's sorted in ascending
+        // order by batch id.
+        int newLargestBatchId =
+            documentToBatchId.get(documentsByBatchId.get(documentsByBatchId.size() - 1));
+        newOffset =
+            IndexOffset.create(
+                newOffset.getReadTime(), newOffset.getDocumentKey(), newLargestBatchId);
+        indexManager.updateIndexEntries(documentsByBatchId);
+        documentsProcessed += documentsByBatchId.size();
+      }
+    }
+
     indexManager.updateCollectionGroup(collectionGroup, newOffset);
-    return oldestDocuments.size();
+    return documentsProcessed;
   }
 
   /** Returns the lowest offset for the provided index group. */
@@ -169,7 +202,8 @@ public class IndexBackfiller {
   }
 
   /**
-   * Returns the offset for the index based on the newly indexed documents.
+   * Returns the read time and document key offset for the index based on the newly indexed
+   * documents.
    *
    * @param documents a list of documents sorted by read time and key (ascending)
    * @param currentOffset the current offset of the index group
@@ -199,6 +233,40 @@ public class IndexBackfiller {
             IndexOffset.create(l.getReadTime(), l.getKey())
                 .compareTo(IndexOffset.create(r.getReadTime(), r.getKey())));
     return oldestDocuments.subList(0, Math.min(count, oldestDocuments.size()));
+  }
+
+  /**
+   * Returns up to {@code count} documents sorted by batch id ascending, except in cases where the
+   * mutation batch id contains more documents. This method will always return all documents in a
+   * mutation batch, even if it goes over the limit.
+   *
+   * <p>For example, if count = 2, and we have: (docA, batch1), (docB, batch2), and (docC, batch2),
+   * all three documents will be returned since docC is also part of batch2.
+   */
+  private List<Document> getDocumentsByBatchId(
+      Map<Document, Integer> documentToBatchId, int count) {
+    List<Document> oldestDocuments = new ArrayList<>(documentToBatchId.keySet());
+    Collections.sort(
+        oldestDocuments, (l, r) -> documentToBatchId.get(l).compareTo(documentToBatchId.get(r)));
+    int i = Math.min(count, oldestDocuments.size()) - 1;
+
+    // Include all documents that match the last document's batch id.
+    int lastBatchId = documentToBatchId.get(oldestDocuments.get(i));
+    while (i < oldestDocuments.size()
+        && lastBatchId == documentToBatchId.get(oldestDocuments.get(i))) {
+      ++i;
+    }
+    return oldestDocuments.subList(0, i);
+  }
+
+  /** Returns the earliest batch id from the specified field indexes. */
+  private int getEarliestBatchId(Collection<FieldIndex> fieldIndexes) {
+    int lowestBatchId = (int) Double.POSITIVE_INFINITY;
+    for (FieldIndex fieldIndex : fieldIndexes) {
+      lowestBatchId =
+          Math.min(fieldIndex.getIndexState().getOffset().getLargestBatchId(), lowestBatchId);
+    }
+    return lowestBatchId;
   }
 
   @VisibleForTesting
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
@@ -17,6 +17,7 @@ package com.google.firebase.firestore.local;
 import static com.google.firebase.firestore.model.DocumentCollections.emptyDocumentMap;
 import static com.google.firebase.firestore.util.Assert.hardAssert;
 
+import android.util.Pair;
 import androidx.annotation.VisibleForTesting;
 import com.google.firebase.Timestamp;
 import com.google.firebase.database.collection.ImmutableSortedMap;
@@ -316,7 +317,32 @@ class LocalDocumentsView {
     }
   }
 
-  /** Queries the remote documents and overlays by doing a full collection scan. */
+  /**
+   * Returns a mapping of documents that match the provided query and offset mapped to the
+   * corresponding batch id.
+   */
+  // TODO(indexing): Support adding local mutations to collection group table.
+  public Map<Document, Integer> getDocumentsBatchIdsMatchingCollectionGroupQuery(
+      Query query, IndexOffset offset) {
+    hardAssert(
+        query.getPath().isEmpty(),
+        "Currently we only support collection group queries at the root.");
+    String collectionId = query.getCollectionGroup();
+    Map<Document, Integer> results = new HashMap<>();
+    List<ResourcePath> parents = indexManager.getCollectionParents(collectionId);
+
+    // Perform a collection query against each parent that contains the collectionId and
+    // aggregate the results.
+    for (ResourcePath parent : parents) {
+      Query collectionQuery = query.asCollectionQueryAtPath(parent.append(collectionId));
+      Map<Document, Integer> documentToBatchIds =
+          getDocumentsMatchingCollectionQueryWithBatchId(collectionQuery, offset);
+      ;
+      results.putAll(documentToBatchIds);
+    }
+    return results;
+  }
+
   private ImmutableSortedMap<DocumentKey, Document>
       getDocumentsMatchingCollectionQueryFromOverlayCache(Query query, IndexOffset offset) {
     ImmutableSortedMap<DocumentKey, MutableDocument> remoteDocuments =
@@ -325,16 +351,7 @@ class LocalDocumentsView {
 
     // As documents might match the query because of their overlay we need to include all documents
     // in the result.
-    Set<DocumentKey> missingDocuments = new HashSet<>();
-    for (Map.Entry<DocumentKey, Mutation> entry : overlays.entrySet()) {
-      if (!remoteDocuments.containsKey(entry.getKey())) {
-        missingDocuments.add(entry.getKey());
-      }
-    }
-    for (Map.Entry<DocumentKey, MutableDocument> entry :
-        remoteDocumentCache.getAll(missingDocuments).entrySet()) {
-      remoteDocuments = remoteDocuments.insert(entry.getKey(), entry.getValue());
-    }
+    remoteDocuments = updateRemoteDocumentsWithOverlayDocuments(remoteDocuments, overlays.keySet());
 
     // Apply the overlays and match against the query.
     ImmutableSortedMap<DocumentKey, Document> results = emptyDocumentMap();
@@ -346,6 +363,53 @@ class LocalDocumentsView {
       // Finally, insert the documents that still match the query
       if (query.matches(docEntry.getValue())) {
         results = results.insert(docEntry.getKey(), docEntry.getValue());
+      }
+    }
+
+    return results;
+  }
+
+  /** Updates the provided remote documents to include documents in the overlay. */
+  private ImmutableSortedMap<DocumentKey, MutableDocument>
+      updateRemoteDocumentsWithOverlayDocuments(
+          ImmutableSortedMap<DocumentKey, MutableDocument> remoteDocuments,
+          Set<DocumentKey> overlayDocumentKeys) {
+    Set<DocumentKey> missingDocuments = new HashSet<>();
+    for (DocumentKey key : overlayDocumentKeys) {
+      if (!remoteDocuments.containsKey(key)) {
+        missingDocuments.add(key);
+      }
+    }
+    for (Map.Entry<DocumentKey, MutableDocument> entry :
+        remoteDocumentCache.getAll(missingDocuments).entrySet()) {
+      remoteDocuments = remoteDocuments.insert(entry.getKey(), entry.getValue());
+    }
+    return remoteDocuments;
+  }
+
+  /** Queries the remote documents and overlays by doing a full collection scan. */
+  public Map<Document, Integer> getDocumentsMatchingCollectionQueryWithBatchId(
+      Query query, IndexOffset offset) {
+    ImmutableSortedMap<DocumentKey, MutableDocument> remoteDocuments =
+        remoteDocumentCache.getAllDocumentsMatchingQuery(query, offset);
+    Map<DocumentKey, Pair<Integer, Mutation>> overlays =
+        documentOverlayCache.getOverlaysWithBatchId(query.getPath(), offset.getLargestBatchId());
+
+    // As documents might match the query because of their overlay we need to include all documents
+    // in the result.
+    remoteDocuments = updateRemoteDocumentsWithOverlayDocuments(remoteDocuments, overlays.keySet());
+
+    // Apply the overlays and match against the query.
+    Map<Document, Integer> results = new HashMap<>();
+    for (Map.Entry<DocumentKey, MutableDocument> docEntry : remoteDocuments) {
+      int batchId = overlays.get(docEntry.getKey()).first;
+      Mutation overlay = overlays.get(docEntry.getKey()).second;
+      if (overlay != null) {
+        overlay.applyToLocalView(docEntry.getValue(), null, Timestamp.now());
+      }
+      // Finally, insert the documents that still match the query
+      if (query.matches(docEntry.getValue())) {
+        results.put(docEntry.getValue(), batchId);
       }
     }
 
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
@@ -105,4 +105,32 @@ public class MemoryDocumentOverlayCache implements DocumentOverlayCache {
 
     return result;
   }
+
+  @Override
+  public Map<DocumentKey, Pair<Integer, Mutation>> getOverlaysWithBatchId(
+      ResourcePath collection, int sinceBatchId) {
+    Map<DocumentKey, Pair<Integer, Mutation>> result = new HashMap<>();
+
+    int immediateChildrenPathLength = collection.length() + 1;
+    DocumentKey prefix = DocumentKey.fromPath(collection.append(""));
+    Map<DocumentKey, Pair<Integer, Mutation>> view = overlays.tailMap(prefix);
+
+    for (Map.Entry<DocumentKey, Pair<Integer, Mutation>> entry : view.entrySet()) {
+      DocumentKey key = entry.getKey();
+      if (!collection.isPrefixOf(key.getPath())) {
+        break;
+      }
+      // Documents from sub-collections
+      if (key.getPath().length() != immediateChildrenPathLength) {
+        continue;
+      }
+
+      Pair<Integer, Mutation> batchIdToOverlay = entry.getValue();
+      if (batchIdToOverlay.first > sinceBatchId) {
+        result.put(entry.getKey(), batchIdToOverlay);
+      }
+    }
+
+    return result;
+  }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java
@@ -16,6 +16,7 @@ package com.google.firebase.firestore.local;
 
 import static com.google.firebase.firestore.util.Assert.fail;
 
+import android.util.Pair;
 import androidx.annotation.Nullable;
 import com.google.firebase.firestore.auth.User;
 import com.google.firebase.firestore.model.DocumentKey;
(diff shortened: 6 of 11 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -325,27 +351,65 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+      if (query.matches(docEntry.getValue())) {
+        results = results.insert(docEntry.getKey(), docEntry.getValue());
+      }
+    }
+
+    return results;
+  }
+
+  /** Updates the provided remote documents to include documents in the overlay. */
+  private ImmutableSortedMap<DocumentKey, MutableDocument>
```
thebrianchen, line 373: I tried refactoring out shared code between `getDocumentsMatchingCollectionQueryFromOverlayCache` and `getDocumentsCollectionQueryWithBatchId`, but I think it can probably be cleaner.

firebase-firestore/src/main/java/com/google/firebase/firestore/index/IndexEntry.java
```diff
@@ -53,4 +54,9 @@ public int compareTo(IndexEntry other) {
 
     return nullSafeCompare(getArrayValue(), other.getArrayValue(), Util::compareByteArrays);
   }
+
+  public static final Comparator<IndexEntry> SEMANTIC_COMPARATOR =
+      (left, right) -> {
+        return left.compareTo(right);
+      };
```
schmidt-sebastian, line 61:
```suggestion
  public static final Comparator<IndexEntry> SEMANTIC_COMPARATOR = IndexEntry::compareTo;
```

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Util.java
```diff
@@ -298,25 +297,6 @@ public static StringBuilder repeatSequence(
...
-   *
-   * @param before The elements that exist in the original set.
-   * @param after The elements to diff against the original set.
-   * @param onAdd A function to invoke for every element that is part of `after` but not `before`.
-   * @param onRemove A function to invoke for every element that is part of `before` but not
-   *     `after`.
-   */
-  public static <T extends Comparable<T>> void diffCollections(
-      SortedSet<T> before, SortedSet<T> after, Consumer<T> onAdd, Consumer<T> onRemove) {
-    diffCollections(before.iterator(), after.iterator(), before.comparator(), onAdd, onRemove);
```
schmidt-sebastian, line 317: This overload is actually different from the one we now default to. This method takes sorted data - the method you are calling now reorders the data (and copies it) even when already sorted. I think you solve the problem you encountered - that `SortedSet<T>.comparator()` can return `null` - by creating a default comparator based on `T.compareTo()` if the set's comparator is null.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlayCache.java
```diff
@@ -54,4 +55,14 @@
...
+
+  /**
+   * Returns a mapping of overlays containing the largest batch id for each overlay mutation.
+   *
+   * @param collection The collection path to get the overlays for.
+   * @param sinceBatchId The minimum batch ID to filter by (exclusive). Only overlays that contain a
+   *     change past `sinceBatchId` are returned.
+   */
+  Map<DocumentKey, Pair<Integer, Mutation>> getOverlaysWithBatchId(
+      ResourcePath collection, int sinceBatchId);
```
schmidt-sebastian, line 67: I am surprised we don't have a high-level class that contains both the mutation and the batch ID for Overlays. We should probably create one and combine this method with the existing method just above. (cc @wu-hui)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -137,23 +137,56 @@ private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
...
-    List<Document> oldestDocuments = getOldestDocuments(documents, entriesRemainingUnderCap);
+    List<Document> oldestDocuments = getOldestDocuments(documents, documentsRemainingUnderCap);
     indexManager.updateIndexEntries(oldestDocuments);
+    documentsProcessed += oldestDocuments.size();
 
     IndexOffset newOffset = getNewOffset(oldestDocuments, existingOffset);
+
+    // Start indexing mutations only after catching up for read time. This must be done separately
+    // since the backfill's first pass is in read time order rather than mutation batch id order.
+    int documentsRemaining = documentsRemainingUnderCap - oldestDocuments.size();
```
schmidt-sebastian, line 162: Can you move this to a helper method for better encapsulation? We might also want to do the same for the RemoteDocumentCache code.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -137,23 +137,56 @@ private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
...
+      // new largest batch id.
+      if (!documentToBatchId.isEmpty()) {
+        List<Document> documentsByBatchId =
+            getDocumentsByBatchId(documentToBatchId, documentsRemaining);
+
+        // Largest batch id is the last element in documentsByBatchId since it's sorted in ascending
+        // order by batch id.
+        int newLargestBatchId =
+            documentToBatchId.get(documentsByBatchId.get(documentsByBatchId.size() - 1));
+        newOffset =
```
schmidt-sebastian, line 180: Can you move this into `getNewOffset`? (as a final call at the end of the method)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -137,23 +137,56 @@ private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
...
     indexManager.updateIndexEntries(oldestDocuments);
+    documentsProcessed += oldestDocuments.size();
 
     IndexOffset newOffset = getNewOffset(oldestDocuments, existingOffset);
+
+    // Start indexing mutations only after catching up for read time. This must be done separately
+    // since the backfill's first pass is in read time order rather than mutation batch id order.
+    int documentsRemaining = documentsRemainingUnderCap - oldestDocuments.size();
+    if (documentsRemaining > 0) {
+      int earliestBatchId = getEarliestBatchId(fieldIndexes);
```
schmidt-sebastian, line 164: I was hoping we would have one existing offset for both the scan of the RDC and the mutations - but I now realize that is not the case. We could change getExistingOffset() to return an offset with the lowest SnapshotVersion+Key as well as the lowest batch ID (which might be from a different index). What do you think?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -137,23 +137,56 @@ private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
...
+            getDocumentsByBatchId(documentToBatchId, documentsRemaining);
+
+        // Largest batch id is the last element in documentsByBatchId since it's sorted in ascending
+        // order by batch id.
+        int newLargestBatchId =
+            documentToBatchId.get(documentsByBatchId.get(documentsByBatchId.size() - 1));
+        newOffset =
+            IndexOffset.create(
+                newOffset.getReadTime(), newOffset.getDocumentKey(), newLargestBatchId);
+        indexManager.updateIndexEntries(documentsByBatchId);
```
schmidt-sebastian, line 183: If I follow correctly, we are now calling this twice for a document if a document is returned from both RDC and the mutation queue. My hunch is that we may even need to move all mutation ID processing to `getDocumentsMatchingQuery` in LocalDocumentsView to not scan the collections twice.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -201,6 +235,40 @@ private IndexOffset getNewOffset(List<Document> documents, IndexOffset currentOf
     return oldestDocuments.subList(0, Math.min(count, oldestDocuments.size()));
   }
 
+  /**
+   * Returns up to {@code count} documents sorted by batch id ascending, except in cases where the
+   * mutation batch id contains more documents. This method will always return all documents in a
```
schmidt-sebastian, line 240: It sounds like all bets are off when the batch ID contains more documents - maybe rewrite the intro to make it more clear that this only applies to the count.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -316,7 +317,31 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
     }
   }
 
-  /** Queries the remote documents and overlays by doing a full collection scan. */
+  /**
+   * Returns a mapping of documents that match the provided query and offset mapped to the
+   * corresponding batch id.
+   */
+  // TODO(indexing): Support adding local mutations to collection group table.
+  public Map<Document, Integer> getDocumentsBatchIdsMatchingCollectionGroupQuery(
```
schmidt-sebastian, line 325: Your code would be cleaner if this method returned `Map<BatchId, List<Document>>`.

The name is also quite a mouthful - and the method should probably just accept a `String collectionGroup` rather than a query as it ignores all other query constraints.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -201,6 +235,40 @@ private IndexOffset getNewOffset(List<Document> documents, IndexOffset currentOf
...
+    while (i < oldestDocuments.size()
+        && lastBatchId == documentToBatchId.get(oldestDocuments.get(i))) {
+      ++i;
+    }
+    return oldestDocuments.subList(0, i);
+  }
+
+  /** Returns the earliest batch id from the specified field indexes. */
+  private int getEarliestBatchId(Collection<FieldIndex> fieldIndexes) {
+    int lowestBatchId = (int) Double.POSITIVE_INFINITY;
```
schmidt-sebastian, line 264: How about Integer.MAX_VALUE? :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
schmidt-sebastian: We are probably at a point where we should just remove the non-overlay code.

@wu-hui

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
```diff
@@ -105,4 +105,32 @@ public void removeOverlaysForBatchId(int batchId) {
 
     return result;
   }
+
+  @Override
+  public Map<DocumentKey, Pair<Integer, Mutation>> getOverlaysWithBatchId(
```
schmidt-sebastian, line 110: Please find a way to remove this duplication - even if we do not add an Overlay class.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java
```diff
@@ -111,4 +112,34 @@ public void removeOverlaysForBatchId(int batchId) {
 
     return result;
   }
+
+  @Override
+  public Map<DocumentKey, Pair<Integer, Mutation>> getOverlaysWithBatchId(
```
schmidt-sebastian, line 117: Same comment as above.

schmidt-sebastian (commented):
Some early feedback (did not look at test). Looks pretty good, but we may want to think about how we can remove the need to read documents multiple times.

See also for some changes that may affect this: https://github.com/firebase/firebase-android-sdk/pull/3193

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexBackfillerTest.java
```diff
@@ -276,6 +287,129 @@ public void testBackfillUsesLatestReadTimeForEmptyCollections() {
...
+    assertEquals(1, documentsProcessed);
+    SortedSet<IndexEntry> before = indexManager.getExistingIndexEntries(docA.getKey(), fieldIndex);
+
+    // Update doc to new remote version with new value.
+    removeDoc(docA);
+    addDoc("coll1/docA", "foo", version(40), 5);
+    documentsProcessed = backfiller.backfill();
+    assertEquals(1, documentsProcessed);
+    SortedSet<IndexEntry> after = indexManager.getExistingIndexEntries(docA.getKey(), fieldIndex);
+    assertNotEquals(before.first().getDirectionalValue(), after.first().getDirectionalValue());
```
schmidt-sebastian, line 390: Instead of relying on getExistingIndexEntries, can you just the public API and invoke `getDocumentsMatchingTarget` with two different targets (foo > 10 and foo > 20 for example). Before the update, you can verify that one of them matches and then after the update two match.

Same comment applies below.

schmidt-sebastian (commented):
Tests look great - but we probably do not need to use the private API.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexBackfillerTest.java
thebrianchen: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Util.java
thebrianchen: Good catch, thanks. Added the code back and added a null check.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -317,36 +327,208 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+   * <p>If no documents are found, returns an empty list and an offset with the latest read time in
+   * the remote document cache.
+   *
+   * @param collectionGroup The collection group for the documents.
+   * @param offset The offset to index into.
+   * @param updateMemoizedResults Whether to load documents and mutations from persistence.
+   * @return A pair containing the next offset and a list of documents that follow the provided
+   *     offset.
+   */
+  public Pair<IndexOffset, List<Document>> getNextDocumentsAndOffsetForCollectionGroup(
```
schmidt-sebastian, line 351: This would be way simpler if Document contained its BatchId. Here is a strawman proposal: https://gist.github.com/schmidt-sebastian/198aa9a0c44b928abe5fc8e5238b584e

This basically means that we can remove all IndexOffset computation from LocalDocumentsView and keep it in the IndexBackfiller.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -317,36 +327,208 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+      String collectionGroup, IndexOffset offset, boolean updateMemoizedResults) {
+
+    if (updateMemoizedResults) {
```
schmidt-sebastian, line 354: I am not 100% convinced that we need to memoize results (and if we do need to do so, we need to discuss what this means if the user changes).

My expected code flow would be as such:

- Fetch 50 documents from RDC.
- If less than 50 documents, fetch next documents from MutationQueue.
- Combine the two results by applying mutations
- If we really care, we can fetch more documents if we are still below 50.

schmidt-sebastian (commented):
Some high level feedback. Let me know if you want to discuss.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlayCache.java
```diff
@@ -47,11 +48,14 @@
...
+   * provided collection.
    *
    * @param collection The collection path to get the overlays for.
    * @param sinceBatchId The minimum batch ID to filter by (exclusive). Only overlays that contain a
    *     change past `sinceBatchId` are returned.
+   * @return Mapping of each document key in the collection to its largest batch id and overlay
+   *     mutation.
    */
-  Map<DocumentKey, Mutation> getOverlays(ResourcePath collection, int sinceBatchId);
+  Map<DocumentKey, Pair<Integer, Mutation>> getOverlays(ResourcePath collection, int sinceBatchId);
```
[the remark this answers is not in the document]
thebrianchen: Created a separate overlay class.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -128,32 +119,42 @@ private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
...
-    // TODO(indexing): Support mutation batch Ids when sorting and writing indexes.
-    ImmutableSortedMap<DocumentKey, Document> documents =
-        localDocumentsView.getDocumentsMatchingQuery(query, existingOffset);
+    Collection<FieldIndex> fieldIndexes = indexManager.getFieldIndexes(collectionGroup);
+    IndexOffset startingOffset = getExistingOffset(fieldIndexes);
+
+    // Represents documents indexed and the updated offset post-update.
+    Pair<IndexOffset, List<Document>> pair =
+        localDocumentsView.getNextDocumentsAndOffsetForCollectionGroup(
+            collectionGroup, startingOffset, /* updateMemoizedResults=` */ true);
```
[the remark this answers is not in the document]
thebrianchen: Used new approach that does now.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -128,32 +119,42 @@ private int writeIndexEntries(LocalDocumentsView localDocumentsView) {
...
+    IndexOffset newOffset;
+    List<Document> documentsToIndex;
+    do {
+      newOffset = pair.first;
+      documentsToIndex = pair.second;
+      indexManager.updateIndexEntries(documentsToIndex);
+      documentsProcessed += documentsToIndex.size();
+      pair =
+          localDocumentsView.getNextDocumentsAndOffsetForCollectionGroup(
+              collectionGroup, newOffset, /* updateMemoizedResults= */ false);
```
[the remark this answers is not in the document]
thebrianchen: N/A now

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
thebrianchen: Implemented, thanks for starting

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
thebrianchen: Implemented.

thebrianchen:
Main changes:
- Documents are no longer memoized.
- LDV still calculates the new offset and returns the result as pair since it's cleaner than passing in the set of document keys in 2 separate calls.
- `Overlay` class added.
- Mutations now have the largest batch id

/test check-changed
/test device-check-changed

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -157,25 +157,24 @@ private IndexOffset getExistingOffset(Collection<FieldIndex> fieldIndexes) {
         lowestOffset = fieldIndex.getIndexState().getOffset();
       }
     }
-    return lowestOffset == null ? IndexOffset.NONE : lowestOffset;
+    lowestOffset = lowestOffset == null ? IndexOffset.NONE : lowestOffset;
+
+    // Add earliest batch id to offset
+    int earliestBatchId = getEarliestBatchId(fieldIndexes);
```
schmidt-sebastian, line 163: Can you inline this call? You are iterating over the list of field indexes twice.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -108,6 +115,14 @@ Document getDocument(DocumentKey key) {
     return getLocalViewOfDocuments(docs, new HashSet<>());
   }
 
+  /** Gets the local view of the next {@code count} documents based on their read time. */
+  ImmutableSortedMap<DocumentKey, Document> getDocuments(
```
schmidt-sebastian, line 119: Do we still need this method? If so, we should move it back to its original spot to reduce diff.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
     return results;
   }
 
-  /** Queries the remote documents and overlays by doing a full collection scan. */
+  /**
+   * Given a collection group, returns the next documents that follow the provided offset, along
+   * with an updated offset.
+   *
+   * <p>This method returns documents in order of read time from the provided offset, followed by
```
schmidt-sebastian, line 267: Nit: The returned documents are sorted by key. They are only consumed by read time.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+   * documents with mutations in order of batch id from the offset. Since all documents in a batch
+   * are returned together, the total number of documents returned can exceed {@code count}.
+   *
+   * <p>If no documents are found, returns an empty list and an offset with the latest read time in
```
schmidt-sebastian, line 271: s/list/map

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+   * the remote document cache.
+   *
+   * @param collectionGroup The collection group for the documents.
+   * @param offset The offset to index into.
+   * @param count The number of documents to return
+   * @return A pair containing the next offset and a list of documents that follow the provided
```
schmidt-sebastian, line 277: s/list/map

It might also make sense to expand on what "next offset" means.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+   *     offset.
+   */
+  public Pair<IndexOffset, ImmutableSortedMap<DocumentKey, Document>> getNextDocumentsAndOffset(
+      String collectionGroup, IndexOffset offset, int count) {
+    ImmutableSortedMap<DocumentKey, Document> documents =
+        getDocuments(collectionGroup, offset, count);
+
+    // TODO(indexing): Store largest batch id on Document class so we can avoid fetching documents
+    // a second time.
+    ImmutableSortedMap<Integer, List<Document>> batchIdToDocuments =
```
schmidt-sebastian, line 287: Does this TODO encapsulate changing this method to follow the following scheme:

- Fetch next from RDC
- Fetch next from OC
- Fetch missing documents from RDC
- Fetch missing overlays from OC
- Combine documents with overlays.
- Return

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+    }
+
+    return new Pair<>(newOffset, documents);
+  }
+
+  /**
+   * Returns a mapping of batch id to documents that match the provided query sorted by batch id
+   * ascending order.
+   */
+  // TODO(indexing): Support adding local mutations to collection group table.
```
schmidt-sebastian, line 311: I actually looked this up the other day but I think we are doing this already:

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryMutationQueue.java#L155

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteMutationQueue.java#L218

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+  public ImmutableSortedMap<Integer, List<Document>>
+      getDocumentsBatchIdsMatchingCollectionGroupQuery(String collectionGroup, IndexOffset offset) {
```
schmidt-sebastian, line 313: This name is quite the handful but it should go away once we add the batch ID to the `Document` class.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+        ImmutableSortedMap.Builder.emptyMap(StandardComparator.getComparator(Integer.class));
+    List<ResourcePath> parents = indexManager.getCollectionParents(collectionGroup);
+
+    // Perform a collection query against each parent that contains the collectionId and
+    // aggregate the results.
+    for (ResourcePath parent : parents) {
+      ResourcePath collectionId = parent.append(collectionGroup);
+      Query collectionQuery = Query.atPath(collectionId);
+      Map<Document, Integer> documentToBatchIds =
+          getDocumentsInOverlayCacheByBatchId(collectionQuery, offset);
```
schmidt-sebastian, line 324: Does this fetch all overlays? Does this need to be limited by count?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+        getDocuments(collectionGroup, offset, count);
+
+    // TODO(indexing): Store largest batch id on Document class so we can avoid fetching documents
+    // a second time.
+    ImmutableSortedMap<Integer, List<Document>> batchIdToDocuments =
+        getDocumentsBatchIdsMatchingCollectionGroupQuery(collectionGroup, offset);
+    IndexOffset newOffset = getNewOffset(documents, offset);
+
+    // Start updating by mutation batch id if under limit.
+    while (documents.size() < count && !batchIdToDocuments.isEmpty()) {
```
schmidt-sebastian, line 292: The count and the limit processing should be part of `getDocumentsBatchIdsMatchingCollectionGroupQuery`

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -251,16 +260,91 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+          getDocumentsInOverlayCacheByBatchId(collectionQuery, offset);
+      for (Document document : documentToBatchIds.keySet()) {
+        int batchId = documentToBatchIds.get(document);
+        if (!results.containsKey(batchId)) {
+          List<Document> documents = new ArrayList<>();
+          documents.add(document);
+          results = results.insert(batchId, documents);
+        } else {
+          results.get(batchId).add(document);
+        }
```
schmidt-sebastian, line 333: This does fewer lookups in fewer lines:

```
val entry = results.get(batchId);
if (entry == null) {
  entry = new {...} 
  result.insert(batchID, entry);
}
entry.add(document)
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -269,16 +353,68 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
     // Apply the overlays and match against the query.
     ImmutableSortedMap<DocumentKey, Document> results = emptyDocumentMap();
     for (Map.Entry<DocumentKey, MutableDocument> docEntry : remoteDocuments.entrySet()) {
-      Mutation overlay = overlays.get(docEntry.getKey());
-      if (overlay != null) {
-        overlay.applyToLocalView(docEntry.getValue(), null, Timestamp.now());
+      Overlay overlay = overlays.get(docEntry.getKey());
+      Mutation mutation = overlay != null ? overlay.getMutation() : null;
+      if (mutation != null) {
```
schmidt-sebastian, line 358: Can we do:

```
if (overlay != null) {
   overlay.getMutation().apply(...)
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -269,16 +353,68 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+    return results;
+  }
+
+  /**
+   * Returns a mapping of documents to their largest batch ids based on the earliest batch id in the
+   * provided offset.
+   */
+  public Map<Document, Integer> getDocumentsInOverlayCacheByBatchId(
+      Query query, IndexOffset offset) {
+    Map<DocumentKey, MutableDocument> remoteDocuments = new HashMap<>();
```
schmidt-sebastian, line 376: Nit: These are not remote documents, they are local docs.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -269,16 +353,68 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+    Map<DocumentKey, Overlay> overlays =
+        documentOverlayCache.getOverlays(query.getPath(), offset.getLargestBatchId());
+
+    // Apply overlays to empty mutable document map to generate base.
+    for (Map.Entry<DocumentKey, Overlay> entry : overlays.entrySet()) {
+      remoteDocuments.put(entry.getKey(), MutableDocument.newInvalidDocument(entry.getKey()));
```
schmidt-sebastian, line 382: Do we not need to fetch the documents from the RDC?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -269,16 +353,68 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+    Map<Document, Integer> results = new HashMap<>();
+    for (Map.Entry<DocumentKey, MutableDocument> docEntry : remoteDocuments.entrySet()) {
+      Overlay overlay = overlays.get(docEntry.getKey());
+      int batchId = overlay.getLargestBatchId();
+      Mutation mutation = overlay.getMutation();
+      if (mutation != null) {
+        mutation.applyToLocalView(docEntry.getValue(), null, batchId, Timestamp.now());
+      }
+      // Finally, insert the documents that still match the query.
+      if (query.matches(docEntry.getValue())) {
```
schmidt-sebastian, line 395: I think we should do this once at the end of `getNextDocumentsAndOffset`.

FWIW, we might not need to do this at all if we pass a collection group here instead of the query. None of the IndexBackfill methods need a query.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndex.java
```diff
@@ -164,10 +175,17 @@ public static IndexOffset fromDocument(Document document) {
...
+     */
+    public abstract int getLargestBatchId();
+
     public int compareTo(IndexOffset other) {
       int cmp = getReadTime().compareTo(other.getReadTime());
       if (cmp != 0) return cmp;
-      return getDocumentKey().compareTo(other.getDocumentKey());
+      cmp = getDocumentKey().compareTo(other.getDocumentKey());
+      if (cmp != 0) return cmp;
+      return Integer.signum(getLargestBatchId() - other.getLargestBatchId());
```
schmidt-sebastian, line 188: Integer.compare? (for noobs like me)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/MutableDocument.java
```diff
@@ -182,6 +184,10 @@ public SnapshotVersion getReadTime() {
     return readTime;
   }
 
+  public int getMutationBatchID() {
```
schmidt-sebastian, line 187: This doesn't seem used but I wish it was :)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/mutation/Overlay.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.google.firebase.firestore.model.mutation;
+
+/**
+ * Representation of an overlay computed by Firestore.
+ *
+ * <p>Holds information about a mutation and the largest batch id in Firestore when the mutation was
+ * created.
+ */
+public class Overlay {
```
schmidt-sebastian, line 23: You probably want to do one of the following:

- Make this an AutoValue
- Make this a Kotlin value class
- Add equals/hashCode

schmidt-sebastian (commented):
This is a huge step in the right direction. We should talk about the refactor to add (and use) the BatchId in the Document class. I see that you added the code to populate it - maybe it will now be easy enough to also use it? It would allow you to stop passing `Pair<integer, document>` types.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -269,16 +353,68 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+      }
+    }
 
     return results;
   }
+
+  /** Returns the offset for the index based on the newly indexed documents. */
+  private IndexOffset getNewOffset(
+      ImmutableSortedMap<DocumentKey, Document> documents, IndexOffset currentOffset) {
+    if (documents.isEmpty()) {
```
schmidt-sebastian, line 406: FWIW This should eventually live in the IndexBackfiller.

schmidt-sebastian: Discussed offline, this probably does not and we need to leave it here.

thebrianchen (commented):
Some comments are now outdated + I don't really remember their original context. Please review this as a rough draft. Will clean up the comments later.

thebrianchen:
Better version being implemented in #3323

PR #3206, opened 2021-12-06
Set DevelopmentPlatform info for Flutter apps

mrober:
Set DevelopmentPlatform info for Flutter apps

Also made DevelopmentPlatform delay the initialization when possible to avoid an additional file operation during start up.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/DevelopmentPlatformProvider.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/DevelopmentPlatformProvider.java
@@ -18,28 +18,22 @@ import static com.google.firebase.crashlytics.internal.common.CommonUtils.getRes
 
 import android.content.Context;
 import androidx.annotation.Nullable;
+import java.io.IOException;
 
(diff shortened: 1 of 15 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/DevelopmentPlatformProvider.java
```diff
@@ -59,6 +53,65 @@ public String getDevelopmentPlatform() {
...
+        return false;
+      }
+      String[] list = context.getAssets().list(path);
+      return list != null && list.length > 0;
+    } catch (IOException ex) {
+      return false;
+    }
+  }
+
+  private DevelopmentPlatform initDevelopmentPlatform() {
```
mrichards, line 82: Does this method need to be synchronized?

mrober: I think it's ok to not synchronize it. Because even if we have two instances, it's read-only so nothing bad will happen.

mrichards (approved):

PR #3208, opened 2021-12-06
Merge the Index Engines

schmidt-sebastian:
This PR removes all Index Engines in favor one of single index engine that does Indexed queries by default, then falls back to index-free if no index is available and then falls back to collection scans.


firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -14,28 +14,224 @@
...
+    return documentAtLimitEdge.hasPendingWrites()
+        || documentAtLimitEdge.getVersion().compareTo(limboFreeSnapshotVersion) > 0;
+  }
+
+  private ImmutableSortedMap<DocumentKey, Document> executeFullCollectionScan(Query query) {
+    if (Logger.isDebugEnabled()) {
+      Logger.debug(LOG_TAG, "Using full collection scan to execute query: %s", query.toString());
+    }
+    return localDocumentsView.getDocumentsMatchingQuery(query, IndexOffset.NONE);
+  }
```
schmidt-sebastian, line 224: All of this code is moved from DefaultQueryEngine.

thebrianchen:
/test device-check-changed

Checking if `canBeRemovedTwice` and `testObservesExistingDocument` still fail.

schmidt-sebastian:
> Checking if `canBeRemovedTwice` and `testObservesExistingDocument` still fail.

Doing the lords work.

(These two tests have been failing a lot since the GRPC update)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -14,28 +14,224 @@
...
+
+  private ImmutableSortedMap<DocumentKey, Document> mergeResults(
```
thebrianchen, line 226: nit: Either add comment or change method name to include the fact that we're merging results with LDV's new documents since the offset

thebrianchen (approved):
lgtm

PR #3209, opened 2021-12-06
Delete index entries for deleted documents

schmidt-sebastian:
This PR deletes the index entries when a document is removed. The entries are deleted right away, but since we don't have to do much computation this is hopefully sufficient.


thebrianchen (approved):
lgtm, looks good!

PR #3225, opened 2021-12-08
Use immutable Documents in MemoryRemoteDocumentCache

schmidt-sebastian:
This is a small refactor that makes it possible to ensure via compile-time checks that we never return a Document directly from the RemoteDocumentCache. Instead, we must call `clone()`.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
@@ -14,10 +14,12 @@
 
 package com.google.firebase.firestore.local;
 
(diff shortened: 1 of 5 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryRemoteDocumentCache.java
```diff
@@ -31,14 +33,14 @@
 final class MemoryRemoteDocumentCache implements RemoteDocumentCache {
 
   /** Underlying cache of documents and their read times. */
-  private ImmutableSortedMap<DocumentKey, MutableDocument> docs;
+  private ImmutableSortedMap<DocumentKey, Document> docs;
```
schmidt-sebastian, line 36: This is the main change.

ehsannas (approved):

PR #3228, opened 2021-12-09
Fad cancel task when dialog closed

rachaprince:
Cancel the ongoing task when the dialog is dismissed or cancelled. Prevents hanging tasks


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -346,19 +346,11 @@ private UpdateTaskImpl showUpdateAlertDialog(AppDistributionRelease newRelease)
     updateDialog.setButton(
         AlertDialog.BUTTON_NEGATIVE,
         context.getString(R.string.update_no_button),
-        (dialogInterface, i) -> {
-          dialogInterface.dismiss();
```
lfkellogg, line 350: I notice all of these `dialogInterface.dismiss()` calls are gone. Were they unnecessary?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/TesterSignInClient.java
```diff
@@ -162,42 +160,38 @@ private AlertDialog getSignInAlertDialog(Activity currentActivity) {
...
-                        LogWrapper.getInstance().e(TAG + "Fid retrieval failed.", e);
-                        setSignInTaskCompletionError(
-                            new FirebaseAppDistributionException(
-                                Constants.ErrorMessages.AUTHENTICATION_ERROR,
-                                AUTHENTICATION_FAILURE,
-                                e));
-                      }
-                    });
-          }
+        (dialogInterface, i) -> {
```
lfkellogg, line 163: Nice :)

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -475,4 +475,20 @@ public void updateApp_withApkReleaseAvailable_returnsSameApkTask() {
 
     assertEquals(updateTask, updateTaskToReturn);
   }
+
+  private UpdateTask updateToNewRelease(AppDistributionReleaseInternal newRelease) {
+    when(mockSignInStorage.getSignInStatus()).thenReturn(true);
+    when(mockCheckForNewReleaseClient.checkForNewRelease()).thenReturn(Tasks.forResult(newRelease));
+    firebaseAppDistribution.setCachedNewRelease(newRelease);
```
lfkellogg, line 482: Why do we have to both mock checkForNewRelease() AND set the cached new release?

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -475,4 +475,20 @@ public void updateApp_withApkReleaseAvailable_returnsSameApkTask() {
 
     assertEquals(updateTask, updateTaskToReturn);
   }
+
+  private UpdateTask updateToNewRelease(AppDistributionReleaseInternal newRelease) {
```
lfkellogg, line 479: I like the consolidation of setup, but this is combining some of the setup with the actual actions being tested. Could we split out the mocking/caching from the `updateIfNewReleaseAvailable()` call?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
lfkellogg: Awesome, thanks!

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
rachaprince: We don't. Good catch. (I think this is dead code from a previous implementation)

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
rachaprince: Good call. Done. After removing the setCachedNewRelease call below, I felt like this wasn't doing enough to justify a separate method.

I moved the signin mocking to the setup method, since most tests aren't testing the signIn logic

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -374,14 +393,6 @@ public void updateToNewRelease_whenCheckForUpdateFails_updateAppNotCalled() {
     assertEquals(FirebaseAppDistributionException.Status.NETWORK_FAILURE, e.getErrorCode());
   }
 
-  @Test
-  public void updateToNewRelease_callsSignInTester() {
```
lfkellogg, line 378: Just checking: did you remove this test because it's only verifying a method call?

lfkellogg (approved):
Looks good! Love the test refactoring.

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
rachaprince: Ah yes, I meant to leave a note here. I felt this test was testing implementation, not behavior, so I removed it

mannyjimenez0810 (approved):

PR #3229, opened 2021-12-09
Handle the case where a tester has access to 0 releases

lfkellogg:
Also refactored the FirebaseAppDistributionTesterApiClient a bit while I was in there, to clean up the error handling and reduce the scope of the try-catch blocks in there, which were large, catching multiple exceptions of unclear origins, etc.


rachaprince (approved):
I think this is only failing because of lint tests. If you run 

`./gradlew :firebase-app-distribution:googleJavaFormat` locally, it should fix the check-changed

Also, you can run  `./gradlew :firebase-app-distribution:check withErrorProne` locally to see the result of check-changed.

lfkellogg:
> I think this is only failing because of lint tests. If you run
> 
> `./gradlew :firebase-app-distribution:googleJavaFormat` locally, it should fix the check-changed
> 
> Also, you can run `./gradlew :firebase-app-distribution:check withErrorProne` locally to see the result of check-changed.

Oh perfect, thanks for the tip! This also made me realize my google-java-format plugin wasn't enabled, all set now 👍

PR #3213, opened 2021-12-07
Fix app distribution api.txt files.

ehsannas:

PR #3232, opened 2021-12-10
Fix API text files for App Distribution

rachaprince:
Based off #3213 . Fixes both the java and kotlin api text files by running

`/gradlew :firebase-app-distribution:generateApiTxtFile` and 
`/gradlew :firebase-app-distribution:ktx:generateApiTxtFile`

### DIFF
--- a/firebase-app-distribution/api.txt
+++ b/firebase-app-distribution/api.txt
@@ -1,14 +1,192 @@
 // Signature format: 2.0
 package com.google.firebase.app.distribution {
-  public class FirebaseAppDistribution{
+
+  public abstract class AppDistributionRelease {
+    ctor public AppDistributionRelease();
+    method @NonNull public static com.google.firebase.app.distribution.AppDistributionRelease.Builder builder();
+    method @NonNull public abstract com.google.firebase.app.distribution.BinaryType getBinaryType();
+    method @NonNull public abstract String getDisplayVersion();
+    method @Nullable public abstract String getReleaseNotes();
+    method @NonNull public abstract long getVersionCode();
+  }
+
+  public abstract static class AppDistributionRelease.Builder {
+    ctor public AppDistributionRelease.Builder();
+    method @NonNull public abstract com.google.firebase.app.distribution.AppDistributionRelease build();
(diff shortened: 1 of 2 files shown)


mannyjimenez0810 (approved):

lfkellogg:
A couple questions:
- Should we wait until [Manny's PR](https://github.com/firebase/firebase-android-sdk/pull/3196) is merged so there will be fewer public classes/methods to document?
- I'm guessing we need to update this periodically as we change the public surface of our API. How will we remember to do this next time?

rachaprince:
Good callout, let's wait for that PR to land. 

The second question is also a good one. It seems like the `api-information` test *should* run on every PR. I wonder if there is a configuration setting that should be fixed in this repo. I'll ask a question to the Firebase Android Core team

ehsannas:
Thanks @rachaprince for fixing this. I didn't have time to work on #3213 any further. I'll close that.

firebase-app-distribution/api.txt
```diff
@@ -1,14 +1,150 @@
...
+    enum_constant public static final com.google.firebase.app.distribution.FirebaseAppDistributionException.Status RELEASE_URL_EXPIRED;
+    enum_constant public static final com.google.firebase.app.distribution.FirebaseAppDistributionException.Status UNKNOWN;
+    enum_constant public static final com.google.firebase.app.distribution.FirebaseAppDistributionException.Status UPDATE_NOT_AVAILABLE;
+  }
+
+  public class FirebaseAppDistributionFileProvider extends androidx.core.content.FileProvider {
+    ctor public FirebaseAppDistributionFileProvider();
+  }
+
+  public class FirebaseAppDistributionLifecycleNotifier implements android.app.Application.ActivityLifecycleCallbacks {
```
rachaprince, line 67: @mannyjimenez0810 out of curiosity, could \ we make this Lifecycle class package private? and perhaps the interfaces as well?

rachaprince:
Looking at the public api files generated by the script, I think there are still some places where we will want to reduce the visibility of methods and classes

rachaprince:
Closing this in favor of #3235, which limits the number of public classes and methods

PR #3235, opened 2021-12-12
Audit of classes making classes as private as possible

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -80,7 +80,7 @@
   }
 
   /** Constructor for FirebaseAppDistribution */
-  public FirebaseAppDistribution(
+  FirebaseAppDistribution(
```
mannyjimenez0810, line 83: This constructor takes in a lifecycle notifier preventing that class from being package private. Is there some reason these constructors needed to be public?

rachaprince: I don't think so. External classes should be calling the `getInstance` function, not the constructor

rachaprince (commented):
Duplicating a comment I left in another doc - 

Great job investigating all these! I know we recently removed the "internal" package prefix, but I think we should bring it back for classes like `SignInResultActivity`, `FirebaseAppDistributionFileProvider`, `SignInStorage` which need to be public, but should not be included in our api.txt file.

For example, here is how remote config handles the storage class being public, with an "internal" prefix: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigStorageClient.java

mannyjimenez0810:
@rachaprince It makes sense to bring back internal for that reason. Will work on that now

mannyjimenez0810:
@rachaprince  Making an internal package and running the same api command it shows up like this. Is this expected?

package com.google.firebase.app.distribution.internal {

  public class FirebaseAppDistributionFileProvider extends androidx.core.content.FileProvider {
    ctor public FirebaseAppDistributionFileProvider();
  }

  public class SignInResultActivity extends androidx.appcompat.app.AppCompatActivity {
    ctor public SignInResultActivity();
    method public void onCreate(@NonNull android.os.Bundle);
  }

  public class SignInStorage {
    ctor public SignInStorage(android.content.Context);
    method public boolean getSignInStatus();
    method public void setSignInStatus(boolean);
  }

}

rachaprince:
@mannyjimenez0810 Great catch. I don't think that's expected. Looking at remote config, I think we need a `package-info.java` file like this one to annotate the internal package as hidden, and prevent it from showing up in our api https://github.com/firebase/firebase-android-sdk/blob/master/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/package-info.java

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/internal/SignInResultActivity.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
```
rachaprince, line 12: Is it possible to also move InstallActivity to this internal package?

mannyjimenez0810: Yes it is. I had to copy logwrapper since we want logwrapper package private and install activity uses a good amount of logging.

firebase-app-distribution/api.txt
```diff
@@ -1,15 +0,0 @@
-// Signature format: 2.0
```
rachaprince, line 1: We should probably regenerate these files instead of deleting the whole thing. It's okay if we end up closing my PR to update the api.txt files because this PR makes those changes obsolete

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/internal/SignInResultActivity.java
rachaprince: My instinct is to delete the package private logwrapper and exclusively the public one in the internal package. Since we don't document anything in internal, I think the customer will be unlikely to accidentally use the internal package. 

What do you think?

rachaprince (approved):

firebase-app-distribution/api.txt
mannyjimenez0810: Sounds good.

mannyjimenez0810:
That sounds good. I don't see any issue with them possibly seeing the logwrapper. Looking at crashlytics seems like they do something similar.

lfkellogg (approved):
Nice!

PR #3237, opened 2021-12-13
Create firestore/Filter.java

ehsannas:
Posting incremental changes towards supporting composite filters and disjunctions.

For the time being, `firestore/Filter.java` is restricted to `Scope.LIBRARY`, and is hidden for api.txt purposes.


firebase-firestore/src/main/java/com/google/firebase/firestore/Filter.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC
+// Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
ehsannas: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/Filter.java
```diff
@@ -0,0 +1,146 @@
...
+    return field;
+  }
+
+  public FieldFilter.Operator getOperator() {
+    return operator;
+  }
+
+  public Object getValue() {
+    return value;
+  }
```
schmidt-sebastian, line 45: I don't think you would want these to be part of the public API. You should annotate them with hide/RestrictTo so that we don't forget when we make the Filter class public.

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -427,9 +425,15 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
               .parseQueryValue(value, op == Operator.IN || op == Operator.NOT_IN);
     }
-    Filter filter = FieldFilter.create(fieldPath.getInternalPath(), op, fieldValue);
+    FieldFilter filter = FieldFilter.create(fieldPath.getInternalPath(), op, fieldValue);
     validateNewFilter(filter);
-    return new Query(query.filter(filter), firestore);
+    return filter;
+  }
+
+  private Query where(Filter filter) {
```
schmidt-sebastian, line 433: Will this eventually be public? If so, we should add a TODO.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Filter.java
```diff
@@ -14,40 +14,12 @@
...
-      return text;
-    }
-  }
-
-  /** Returns the field the Filter operates over. */
-  public abstract FieldPath getField();
-
   /** Returns true if a document matches the filter. */
-  public abstract boolean matches(Document doc);
+  public abstract boolean matches(@NonNull Document doc);
```
schmidt-sebastian, line 22: We don't use NonNull in the SDK internal classes, only the API layer. We do use Nullable and assume NonNull is the default.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -124,7 +124,8 @@ public boolean hasLimit() {
     if (segment == null) return null;
 
     for (Filter filter : filters) {
-      if (filter.getField().equals(segment.getFieldPath())) {
+      if ((filter instanceof FieldFilter)
+          && (((FieldFilter) filter).getField()).equals(segment.getFieldPath())) {
         FieldFilter fieldFilter = (FieldFilter) filter;
```
schmidt-sebastian, line 129: These three lines are used 4 times in this file. Can we add a helper (`getFieldFilters(Path path)` for example)?

schmidt-sebastian (commented):
Basically LGTM

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
ehsannas: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -427,9 +425,16 @@ private Query whereHelper(@NonNull FieldPath fieldPath, Operator op, Object valu
...
               .parseQueryValue(value, op == Operator.IN || op == Operator.NOT_IN);
     }
-    Filter filter = FieldFilter.create(fieldPath.getInternalPath(), op, fieldValue);
+    FieldFilter filter = FieldFilter.create(fieldPath.getInternalPath(), op, fieldValue);
     validateNewFilter(filter);
-    return new Query(query.filter(filter), firestore);
+    return filter;
+  }
+
+  // TODO(ehsann): This method will become public API. Change visibility and add documentation.
```
schmidt-sebastian, line 433: Nit: Maybe "query"/"orquery" if you want to work on this with others :)

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -115,6 +116,17 @@ public boolean hasLimit() {
     return endAt;
   }
 
+  /** Returns the field filters from the given list that target the given field path. */
+  private List<FieldFilter> getFieldFiltersForPath(List<Filter> filters, FieldPath path) {
```
schmidt-sebastian, line 120: These "filters" are always the instance variable, right? Can the argument be dropped?

schmidt-sebastian (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
ehsannas: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
ehsannas: yeah, done.

PR #3244, opened 2021-12-15
Remove base mutations

wu-hui:
Some tests need to be updated, they have document existence change, hence mutation queue is re-applied. IMO I think it is the right thing to do.

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
@@ -1349,8 +1349,8 @@ public abstract class LocalStoreTestCase {
     assertContains(doc("foo/bar", 0, map("sum", 1)).setHasLocalMutations());
 
     applyRemoteEvent(addedRemoteEvent(doc("foo/bar", 1, map("sum", 1337)), asList(2), emptyList()));
-    assertChanged(doc("foo/bar", 1, map("sum", 1)).setHasLocalMutations());
-    assertContains(doc("foo/bar", 1, map("sum", 1)).setHasLocalMutations());
+    assertChanged(doc("foo/bar", 1, map("sum", 1338)).setHasLocalMutations());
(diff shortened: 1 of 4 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -1349,8 +1349,8 @@ public void testHandlesMergeMutationWithTransformThenRemoteEvent() {
     assertContains(doc("foo/bar", 0, map("sum", 1)).setHasLocalMutations());
 
     applyRemoteEvent(addedRemoteEvent(doc("foo/bar", 1, map("sum", 1337)), asList(2), emptyList()));
-    assertChanged(doc("foo/bar", 1, map("sum", 1)).setHasLocalMutations());
-    assertContains(doc("foo/bar", 1, map("sum", 1)).setHasLocalMutations());
+    assertChanged(doc("foo/bar", 1, map("sum", 1338)).setHasLocalMutations());
```
schmidt-sebastian, line 1352: I am surprised that this changed. Are we not using the overlay here? The problem with this change is that it now causes flicker. If "1337" already contains "1" then we will go to 1338 and then 1337.

schmidt-sebastian (commented):
Please figure out why the tests changed - I don't think they should.

We can also remove the baseMutation field from the proto (but make sure to indicate in a comment that the field ID was already used)

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
wu-hui: It changed because the remote update changed existence state, which leads to re-applying mutation queue, without base mutations this time.

schmidt-sebastian: Makes sense. We probably need to come up with a way to avoid this, but I can't think of anything out of the top of my head.

PR #3246, opened 2021-12-15
Remote document reads elision

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -89,12 +91,20 @@ IndexManager getIndexManager() {
...
+    MutableDocument document = null;
+    if (overlay == null) {
+      document = remoteDocumentCache.get(key);
+    } else {
+      // Remote document elision: avoid reading remote documents if their overlay is Set or Delete.
+      if (overlay instanceof SetMutation || overlay instanceof DeleteMutation) {
+        document = MutableDocument.newInvalidDocument(key);
+      } else {
+        document = remoteDocumentCache.get(key);
+      }
```
schmidt-sebastian, line 103: What about:

```suggestion
    MutableDocument document = overlay instanceof PatchMutation ?
             remoteDocumentCache.get(key) :  MutableDocument.newInvalidDocument(key);
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -254,10 +264,19 @@ void recalculateAndSaveOverlays(Set<DocumentKey> documentKeys) {
...
+    // Remote document elision: avoid reading remote documents if their overlay is Set or Delete.
+    Set<DocumentKey> ignore = new HashSet<>();
+    for (Map.Entry<DocumentKey, Mutation> entry : overlays.entrySet()) {
+      if (entry.getValue() instanceof SetMutation || entry.getValue() instanceof DeleteMutation) {
+        ignore.add(entry.getKey());
+      }
+    }
+
+    Map<DocumentKey, MutableDocument> remoteDocuments =
+        remoteDocumentCache.getAll(query.getPath(), offset, ignore);
```
schmidt-sebastian, line 278: I don't think this part of the optimization is quite worth it. We are still reading the elements from the cache. What do you think about not touching `getAll()`?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/InternalDocument.java
```diff
@@ -0,0 +1,33 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.model;
+
+/** Interface exposing underlying document information that is not visible in {@code Document}. */
+public interface InternalDocument extends Document {
```
schmidt-sebastian, line 18: I don't quite understand the need for this indirection. Can you explain your reasoning?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
```diff
@@ -164,15 +166,18 @@ public MutableDocument get(DocumentKey documentKey) {
     if (collections.isEmpty()) {
       return Collections.emptyMap();
     } else if (BINDS_PER_STATEMENT * collections.size() < SQLitePersistence.MAX_ARGS) {
-      return getAll(collections, offset, limit);
+      return getAll(collections, offset, limit, new HashSet<>());
```
schmidt-sebastian, line 169: `Collections.emptySet()` to avoid creating a new object.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
```diff
@@ -29,18 +29,6 @@
   /** The key for this document */
   DocumentKey getKey();
 
-  /**
-   * Returns the version of this document if it exists or a version at which this document was
-   * guaranteed to not exist.
-   */
-  SnapshotVersion getVersion();
```
schmidt-sebastian, line 36: I am not sure where to put this as a comment, but I think we should just set the "version" and "readTime" to `0` in setHasLocalMutations()

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -389,7 +389,8 @@ public void testHandlesSetMutationThenDocument() {
             asList(targetId),
             emptyList()));
     assertChanged(doc("foo/bar", 2, map("foo", "bar")).setHasLocalMutations());
-    assertContains(doc("foo/bar", 2, map("foo", "bar")).setHasLocalMutations());
+    // Version is 0 because of remote document elision
```
schmidt-sebastian, line 392: I don't think we need these comments once we use SnapshotVersion.NONE for all documents with local mutations (plus, I am not sure that the terminology used in this comment makes it obvious what is going on here)

schmidt-sebastian (commented):
Thank you for doing this. Some suggestions to trim this down a bit.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
wu-hui: It might be useful when user goes offline long enough and there are a lot documents with overlays. It is hard to say at what point it is worth it, but it does not hurt performance so let's keep it?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/InternalDocument.java
wu-hui: I guess I wanted a bit type safety, that `getVersion` and `getReadTime` can only be accessed via this interface, instead of setting them to 0. And `DocumentSnapshot` holds a `Document` so no access to version/readtime to user code.

I feel safer this way, comparing to reviewing all places where they are used and reason whether 0 makes sense there. Setting them to 0 also felt "wrong" because we know they are at least the versions they are set now, and setting to 0 gives up that information.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Document.java
wu-hui: See my other comment.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
wu-hui: This IS a bit awkward, but let's decide which way to go first.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
schmidt-sebastian: I'd rather keep it as is. The old way is shorter, easier to understand and does not require porting. The only thing you are saving here is the Proto decoding, which already happens in the background.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/InternalDocument.java
schmidt-sebastian: My concern here is the same as with the other change: It adds a layer of indirection which adds porting work and code size. We already use SnapshotVersion.NONE as the "non-existing" snapshot version in a lot of places in our codebase.

If you want to achieve what you are after, you could make the missing versions nullable - but I really thing we should just continue to use SnapshotVersion.NONE.

I also do not quite understand where the name "Internal" comes from. I think you mean something like "FoundDocument", which we used to have but moved away from.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
schmidt-sebastian: If we keep the comment, we should phrase it differently and remove the term "elision", which requires too much knowledge to understand.

wu-hui: Removed comments.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/InternalDocument.java
wu-hui: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -546,9 +548,10 @@ public void testHandlesPatchMutationThenDocumentThenAck() {
     int targetId = allocateQuery(query);
     applyRemoteEvent(
         addedRemoteEvent(
-            doc("foo/bar", 1, map("it", "base")).setHasLocalMutations(),
+            asList(doc("foo/bar", 1, map("it", "base")).setHasLocalMutations()),
```
schmidt-sebastian, line 551: I think you can undo the change to "addedRemoteEvent". A remote doc should never have local mutations, so some of these tests are wrong.

schmidt-sebastian (approved):

PR #3252, opened 2021-12-20
Remove support for multiple sdk instances in Firebase App Distribution

rachaprince:
Remove support for multiple instances of the sdk. Only support the default Firebase app


mannyjimenez0810 (approved):
Seems theres one failing check. Great work!

PR #3253, opened 2021-12-20
Clean up executors and add clarifying comments

rachaprince:
* Change the name of executors to better describe their purpose
* Add clarifying comments


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/UpdateApkClient.java
```diff
@@ -180,11 +179,7 @@ private void makeApkDownloadRequest(
   }
 
   private void downloadToDisk(
-      InputStream input,
-      long totalSize,
-      String fileName,
-      AppDistributionReleaseInternal newRelease,
```
rachaprince, line 186: The newRelease parameter was unused

mannyjimenez0810: nice catch!

mannyjimenez0810 (approved):

PR #3254, opened 2021-12-20
Crashlytics efficient keys

mrichards:
Improve efficiency of `setCustomKeys(...)` by adding marker bits to the keys map, so redundant data is not re-serialized. Also improved task scheduling so only one serialization task is in the queue at a time.


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/UserMetadata.java
```diff
@@ -0,0 +1,242 @@
...
+   * with key has changed. This method will also create and schedule a background Task to serialize
+   * the updated data, if necessary.
+   */
+  public void setCustomKeys(Map<String, String> keysAndValues) {
+    customKeys.setKeys(keysAndValues);
+  }
+
+  /** @return defensive copy of the internal keys. */
+  public Map<String, String> getInternalKeys() {
+    synchronized (internalKeys) {
```
mrichards, line 125: Does this need to be synchronized? The setCustomKeys implementation is not.

mrober (approved):

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/UserMetadata.java
mrichards: No, the `synchronized` block here is redundant. Fixed.

PR #3257, opened 2021-12-21
docs: improve documentation for merge behavior

schmidt-sebastian:
Addresses firebase/firebase-admin-node#987

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
@@ -59,7 +59,8 @@ public final class SetOptions {
 
   /**
    * Changes the behavior of {@code set()} calls to only replace the values specified in its data
-   * argument. Fields omitted from the {@code set()} call will remain untouched.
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
```diff
@@ -59,7 +59,8 @@ public FieldMask getFieldMask() {
 
   /**
    * Changes the behavior of {@code set()} calls to only replace the values specified in its data
-   * argument. Fields omitted from the {@code set()} call will remain untouched.
+   * argument. Fields omitted from the {@code set()} call will remain untouched. If your input sets
+   * any field to an empty map, all nested fields at this location are overwritten.
```
thebrianchen, line 63: optional: use `{@code Map}?

thebrianchen (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/SetOptions.java
schmidt-sebastian: I left it as is as it creates weird capitalization and should also really be a `{@link Map}`, which then is too generic to be useful.

PR #3259, opened 2021-12-21
New build version now includes different version name

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
```diff
@@ -160,7 +160,15 @@ private boolean canInstall(AppDistributionReleaseInternal newRelease)
   private boolean isNewerBuildVersion(AppDistributionReleaseInternal newRelease)
       throws FirebaseAppDistributionException {
     return Long.parseLong(newRelease.getBuildVersion())
-        > getInstalledAppVersionCode(firebaseApp.getApplicationContext());
+            > getInstalledAppVersionCode(firebaseApp.getApplicationContext())
+        || !doesAppVersionNameMatch(newRelease);
```
mannyjimenez0810, line 164: Didn't want to throw the below code in here since it didn't look great, but I am open to putting it back.

lfkellogg, line 164: I don't think this new clause belongs in `isNewerBuildVersion()` since it has nothing to do with the build version. Instead it should be added to the `if` block where this is called, like:
```
      if (isNewerBuildVersion(retrievedNewRelease)
          || !doesAppVersionNameMatch(retrievedNewRelease) // <- here
          || !isSameAsInstalledRelease(retrievedNewRelease)) {
        return retrievedNewRelease;
      } else ...
      ```
      
That more accurately captures the fact that there are three distinct cases where we consider the release to be "new."

File: firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java [hunk 2]
```diff
@@ -160,7 +160,15 @@ private boolean canInstall(AppDistributionReleaseInternal newRelease)
...
+  }
+
+  private boolean doesAppVersionNameMatch(AppDistributionReleaseInternal newRelease)
```
lfkellogg, line 167: To avoid having to negate the return value where this is used, and to improve readability, I'd change this to be `hasDifferentAppVersion` and return the opposite boolean value.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
```diff
@@ -195,6 +203,21 @@ private long getInstalledAppVersionCode(Context context) throws FirebaseAppDistr
     return PackageInfoCompat.getLongVersionCode(pInfo);
   }
 
+  private String getInstalledAppVersionName(Context context)
+      throws FirebaseAppDistributionException {
+    PackageInfo pInfo;
+    try {
+      pInfo = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
+    } catch (PackageManager.NameNotFoundException e) {
+      LogWrapper.getInstance().e(TAG + "Unable to locate Firebase App.", e);
```
lfkellogg, line 212: We're not looking for the "Firebase App" right? Seems like "Package name not found when looking for installed app version name" or something would be more accurate.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
```diff
@@ -195,6 +203,21 @@ private long getInstalledAppVersionCode(Context context) throws FirebaseAppDistr
     return PackageInfoCompat.getLongVersionCode(pInfo);
   }
 
+  private String getInstalledAppVersionName(Context context)
+      throws FirebaseAppDistributionException {
+    PackageInfo pInfo;
+    try {
+      pInfo = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
```
lfkellogg, line 210: Rather than initialize pInfo above, then set it here, and return it after the catch block, just return it here:
```
try {
  return context.getPackageManager().getPackageInfo(context.getPackageName(), 0).versionName;
} ...
```

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
```diff
@@ -118,6 +118,7 @@ public void setup() {
             .setApplicationInfo(applicationInfo)
             .build();
     packageInfo.setLongVersionCode(INSTALLED_VERSION_CODE);
+    packageInfo.versionName = "1.0";
```
lfkellogg, line 121: Please add new test cases that cover the new functionality: version name is different, version name is the same, etc.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
rachaprince: +1

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
rachaprince: We should probably update the error message in the `getInstalledAppVersionCode` method as well

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
mannyjimenez0810: Good point will change.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
mannyjimenez0810: Good point

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
mannyjimenez0810: When the version name is the same it should not return a new release which we have tests for. Will add one for when version name is different

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
mannyjimenez0810: ok

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
mannyjimenez0810: That sounds right. Tweaked the logs to better reflect what is going on.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
```diff
@@ -186,7 +194,8 @@ private long getInstalledAppVersionCode(Context context) throws FirebaseAppDistr
     try {
       pInfo = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
     } catch (PackageManager.NameNotFoundException e) {
```
rachaprince, line 196: Thoughts on extracting the getPackageInfo logic into a method so we could DRY up this code?

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
```diff
@@ -286,6 +287,31 @@ public void handleNewReleaseFromClient_whenNewAabIsAvailable_returnsRelease() th
...
   }
 
+  @Test
+  public void
+      handleNewReleaseFromClient_whenNewReleaseHasDifferentVersionNameThanInstalled_returnsRelease()
+          throws Exception {
+    when(mockFirebaseAppDistributionTesterApiClient.fetchNewRelease(
+            any(), any(), any(), any(), any()))
+        .thenReturn(
+            getTestNewReleaseDifferentVersionName()
```
rachaprince, line 297: nit: extract this to a variable so you don't need to set the builder fields again below

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
```diff
@@ -286,6 +287,31 @@ public void handleNewReleaseFromClient_whenNewAabIsAvailable_returnsRelease() th
...
+                .setDownloadUrl("http://fake-download-url")
+                .setIasArtifactId(TEST_IAS_ARTIFACT_ID)
+                .setBinaryType(BinaryType.AAB)
+                .build());
```
rachaprince, line 301: I think you can do something like `getTestInstalledRelease().setDisplayVersion("New Version Name").....`

That way we don't need to create a second method to generate the expected release. 
(This is based on the testing pattern described here: https://testing.googleblog.com/2018/02/testing-on-toilet-cleanly-create-test.html)

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
mannyjimenez0810: I have made the change to see how it looks. It adds some overall code, but does make it look nicer.

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
mannyjimenez0810: Good call. Changed the other methods to also have this change.

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
mannyjimenez0810: Great catch! Thanks!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
rachaprince: I think you could also add the try/catch statements to the new method so we don't need to duplicate that code

rachaprince (approved):

lfkellogg:
Looks good!

PR #3260, opened 2021-12-21
Fad only show UI for basic config

rachaprince:
* SignIn Dialog now only shows for the `updatetonewreleaseifavailable` method


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -132,41 +135,124 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+              (dialogInterface, i) -> dismissSignInDialogCallback(),
+              (dialogInterface) -> dismissSignInDialogCallback());
+      signInDialog.show();
+    } else {
+      checkForNewRelease() // check for new release will call sign in tester
+          .onSuccessTask(this::handleNewRelease)
+          .addOnFailureListener(this::onUpdateIfNewReleaseFailure);
+    }
+
+    synchronized (updateIfNewReleaseTaskLock) {
```
lfkellogg, line 166: This synchronized block serves no purpose since it immediately returns anyway.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/DialogUtils.java
```diff
@@ -0,0 +1,78 @@
...
+import android.content.DialogInterface.OnCancelListener;
+import android.content.DialogInterface.OnClickListener;
+import androidx.annotation.NonNull;
+import com.google.firebase.FirebaseApp;
+
+class DialogUtils {
+
+  private DialogUtils() {}
+
+  static AlertDialog getSignInDialog(
```
lfkellogg, line 15: These methods don't really need to return the dialog since all the caller does in each case is call `show()` on it. For convenience you might as well just change these to `void showSomeDialog()` methods.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/TesterSignInManager.java
```diff
@@ -143,8 +136,16 @@ private void onActivityDestroyed(Activity activity) {
 
       signInTaskCompletionSource = new TaskCompletionSource<>();
 
-      alertDialog = getSignInAlertDialog(currentActivity);
-      alertDialog.show();
+      firebaseInstallationsApi
```
lfkellogg, line 139: Everything in signInTester() before this used to happen prior to the sign in dialog showing. Now it happens after the sign in dialog shows, right? Are there any implications of that? Seems fine to me but just wondering.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
rachaprince: We still need the block, because we are reading the `cachedUpdatedIfNewReleaseTask`, which is guarded by the `updateIfNewReleaseLock` to prevent race conditions.

The alternative would be to remove or ignore the `@GuardedBy` annotation (which I think would be prone to errors)

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -132,41 +135,124 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+      }
+
+      signInDialog =
+          getSignInDialog(
+              currentActivity,
+              firebaseApp,
+              (dialogInterface, i) ->
+                  checkForNewRelease() // check for new release will call sign in tester
+                      .onSuccessTask(this::handleNewRelease)
+                      .addOnFailureListener(this::onUpdateIfNewReleaseFailure),
```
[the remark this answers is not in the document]
rachaprince: This is complicated by the progress updates we post to the  `cachedUpdateIfNewReleaseTask` of type `UpdateTask` (which is present when we are calling from `updateToNewReleaseIfAvailable`), but there is no UpdateTask when `checkForNewRelease` is called directly.  

These handlers are specific to the basic configuration. 

But perhaps there is a way to overload checkForNewRelease to DRY this up. I will take another look. Let me know if you have ideas/thoughts

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -132,41 +135,124 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+                ErrorMessages.APP_BACKGROUNDED,
+                FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE));
+      }
+
+      signInDialog =
+          getSignInDialog(
+              currentActivity,
+              firebaseApp,
+              (dialogInterface, i) ->
+                  checkForNewRelease() // check for new release will call sign in tester
```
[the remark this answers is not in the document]
rachaprince: This is a great question, and something we've gone back and forth with. 

We chose to have checkForNewRelease call signInTester to keep the methods aligned with the iOS SDK. Granted, the SDKs already deviate, so we could remove signIn from the SDK and be okay with the different behaviors, but this may impact Flutter and other cross-platform frameworks

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -129,12 +133,16 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
       cachedUpdateIfNewReleaseTask = new UpdateTaskImpl();
     }
-    checkForNewRelease()
-        .onSuccessTask(
-            release -> {
+
+    showSignInDialog()
+        .onSuccessTask(unused -> signInTester())
+        .onSuccessTask(unused -> checkForNewRelease())
+        .continueWithTask(
```
lfkellogg, line 142: If you're just calling getResult() and allowing the errors to propagate, then AFAICT you may as well use `onSuccessTask()` here.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -145,28 +151,117 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+
+    TaskCompletionSource<Void> showDialogTask = new TaskCompletionSource<>();
+
+    Activity currentActivity = lifecycleNotifier.getCurrentActivity();
+    if (currentActivity == null) {
+      LogWrapper.getInstance().e("No foreground activity found.");
+      showDialogTask.setException(
+          new FirebaseAppDistributionException(
+              ErrorMessages.APP_BACKGROUNDED,
+              FirebaseAppDistributionException.Status.UPDATE_NOT_AVAILABLE));
```
mannyjimenez0810, line 182: Seems like there are scenarios like this where the status doesn't exactly align with what is going on (here its not that an update isn't available). Our logs do have the more specific case to give the user a better sense of whats going on, but do we want to look at these statuses to make them more specific?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -145,28 +151,117 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+                new FirebaseAppDistributionException(
+                    ErrorMessages.AUTHENTICATION_CANCELED, AUTHENTICATION_CANCELED)));
+
+    signInDialog.setOnCancelListener(
+        dialogInterface ->
+            showDialogTask.setException(
+                new FirebaseAppDistributionException(
+                    ErrorMessages.AUTHENTICATION_CANCELED, AUTHENTICATION_CANCELED)));
+
+    signInDialog.show();
```
mannyjimenez0810, line 209: I see we have a variable for updateDialog do we want something similar for signInDialog?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -294,113 +396,39 @@ AppDistributionReleaseInternal getCachedNewRelease() {
...
     synchronized (updateIfNewReleaseTaskLock) {
       safeSetTaskResult(cachedUpdateIfNewReleaseTask);
     }
-    dismissUpdateDialog();
+    dismissDialogs();
   }
 
-  private void dismissUpdateDialog() {
+  private void dismissDialogs() {
+    if (signInDialog != null && signInDialog.isShowing()) {
```
mannyjimenez0810, line 422: Addition to my comment above about dialogs. Is there a reason why theres some difference in logic for checking to dismiss the dialogs?

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -393,6 +394,59 @@ public void updateToNewRelease_whenCheckForUpdateFails_updateAppNotCalled() {
...
+
+  @Test
+  public void signInTester_whenDialogDismissed_taskFails() {
+    when(mockSignInStorage.getSignInStatus()).thenReturn(false);
+    Task updateTask = firebaseAppDistribution.updateIfNewReleaseAvailable();
+
+    AlertDialog dialog = verifySignInAlertDialog();
+    dialog.getButton(AlertDialog.BUTTON_NEGATIVE).performClick(); // dismiss dialog
+
+    assertFalse(updateTask.isSuccessful());
```
mannyjimenez0810, line 420: Its only repeated once so its not necessary, but we could DRY up this code by putting this in another method to shorten this method and assertTaskFailure

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
rachaprince: This is a really good call. I've added an error message enum to our API and made a note to add it to the API addendum.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
lfkellogg: Ah, OK thanks for checking. I'm fine with leaving it for now.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
rachaprince: I'm not sure I understand... could you point me to the line(s) of code?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
rachaprince: Good catch! I don't think so... I'm not sure why we are using a custom boolean flag `updateDialogShown` instead of relying on the built-in `updateDialog.isShowing()` bool.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
lfkellogg: Agreed, good callout. There are a bunch of places like this where the error status and/or message does not align with the actual situation. I'm currently going through them all as part of the error handling cleanup, so @rachaprince if you want I can handle this in a followup since this was just moved code.

For this case though, looking at the statuses, `AUTHENTICATION_FAILED` might work if we consider "authentication" to be synonymous with "tester sign in."

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -129,10 +133,12 @@ public UpdateTask updateIfNewReleaseAvailable() {
       if (cachedUpdateIfNewReleaseTask != null && !cachedUpdateIfNewReleaseTask.isComplete()) {
         return cachedUpdateIfNewReleaseTask;
       }
-
       cachedUpdateIfNewReleaseTask = new UpdateTaskImpl();
     }
-    checkForNewRelease()
+
+    showSignInDialog()
+        .onSuccessTask(unused -> signInTester())
```
lfkellogg, line 139: Just to be clear: even though checkForNewRelease() calls signInTester(), you're calling it here too just to be explicit?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -78,6 +80,16 @@ Activity getCurrentActivity() {
     }
   }
 
+  Activity getNonNullCurrentActivity() throws FirebaseAppDistributionException {
```
rachaprince, line 83: @lfkellogg FYI if you chose to implement my suggestion on your last PR, this may conflict

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -147,26 +152,110 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+        dialogInterface ->
+            showDialogTask.setException(
+                new FirebaseAppDistributionException(
+                    ErrorMessages.AUTHENTICATION_CANCELED, AUTHENTICATION_CANCELED)));
+
+    signInDialog.show();
+    return showDialogTask.getTask();
+  }
+
+  private Task<Void> showUpdateAlertDialog(AppDistributionRelease newRelease) {
```
lfkellogg, line 209: Sorry if this is annoying but can you move this back to the same spot in the code where it was before? It would make the diff easier to read.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/TesterSignInManager.java
lfkellogg: We chatted this out, seems fine.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
lfkellogg: Actually I think your new "no foreground activity" error seems like a good one to add, could be useful elsewhere. `AUTHENTICATION_FAILED` probably implies that the actual authentication was not allowed, which would be misleading.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionLifecycleNotifier.java
lfkellogg: Thanks for the heads up!

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
```diff
@@ -393,6 +394,59 @@ public void updateToNewRelease_whenCheckForUpdateFails_updateAppNotCalled() {
...
 
+  @Test
+  public void updateToNewRelease_whenTesterIsSignedIn_doesNotOpenDialog() {
+    when(mockSignInStorage.getSignInStatus()).thenReturn(true);
+    when(mockNewReleaseFetcher.checkForNewRelease())
+        .thenReturn(
+            Tasks.forException(
+                new FirebaseAppDistributionException(
+                    Constants.ErrorMessages.NETWORK_ERROR,
+                    FirebaseAppDistributionException.Status.NETWORK_FAILURE)));
```
lfkellogg, line 405: Can you remove this unnecessary mock?

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
lfkellogg: If you want to DRY this up, see [my PR](https://github.com/firebase/firebase-android-sdk/pull/3263/files#diff-691d14245f79be92331994b42451dda4fe8bd5fa297bf081604b7eda370b4fa4R177-R190) for a possible approach.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
mannyjimenez0810: I think this question was answered below. Where we use updateDialogShown instead of the updateDialog.isShowing().

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
lfkellogg: SGTM!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -147,26 +164,59 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
-              setCachedUpdateIfNewReleaseCompletionError(
-                  e,
-                  new FirebaseAppDistributionException(
-                      Constants.ErrorMessages.NETWORK_ERROR,
-                      FirebaseAppDistributionException.Status.NETWORK_FAILURE));
-            });
+        .onSuccessTask(
+            unused ->
+                updateApp(true)
+                    .addOnProgressListener(this::postProgressToCachedUpdateIfNewReleaseTask))
```
lfkellogg, line 170: Please fix this formatting

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -195,6 +284,13 @@ public UpdateTask updateIfNewReleaseAvailable() {
                 })
             .addOnFailureListener(
                 e -> {
+                  postProgressToCachedUpdateIfNewReleaseTask(
+                      UpdateProgress.builder()
+                          .setApkBytesDownloaded(UNKNOWN_RELEASE_FILE_SIZE)
+                          .setApkFileTotalBytes(UNKNOWN_RELEASE_FILE_SIZE)
+                          .setUpdateStatus(UpdateStatus.NEW_RELEASE_CHECK_FAILED)
+                          .build());
+
```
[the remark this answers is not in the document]
lfkellogg: Oh I see, it's because the `NEW_RELEASE_CHECK_FAILED` vs. the installation failing? OK in that case I'm fine either way.

lfkellogg (approved):

lfkellogg:
Looks like you just need to run `./gradlew :firebase-app-distribution:googleJavaFormat`

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionTest.java
rachaprince: Addressed this feedback in https://github.com/firebase/firebase-android-sdk/pull/3286

mannyjimenez0810 (approved):

lfkellogg (approved):

PR #3261, opened 2021-12-22
Include internal keys even if no custom keys were set.

mrober:
This will fix the issue with Unity metadata being excluded in some cases.


mrichards (approved):
LGTM, nice test! :)

PR #3262, opened 2021-12-22
Adding clear task flag to fix update app bug

mannyjimenez0810:
The bug in question would occur when a user would update the app and not close the app after updating and opening it using the system dialog that occurs after an update. When they would try to update again the update would fail and the task would hang.

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/internal/InstallActivity.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/internal/InstallActivity.java
@@ -144,8 +144,7 @@ public class InstallActivity extends AppCompatActivity {
 
     // These flags open the installation activity in a new task and to prevent earlier installation
     // tasks from causing future ones to fail we use the clear task flag
-    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-    intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK);
+    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
     LogWrapper.getInstance().v("Kicking off install as new activity");
     startActivity(intent);
   }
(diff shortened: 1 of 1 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/internal/InstallActivity.java
```diff
@@ -143,6 +143,7 @@ private void startAndroidPackageInstallerIntent() {
     }
 
     intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+    intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK);
```
mannyjimenez0810, line 146: I was debating on whether or not to put a comment. I don't feel like it's necessary, but am happy to hear other opinions.

rachaprince: I think a comment could be nice here. Who knows who will be looking at this code a year from now :)

rachaprince (approved):
🎉

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/internal/InstallActivity.java
lfkellogg, line 146: Because these flags are bits, they can be combined like:
```
intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
```

See examples [here](https://source.corp.google.com/search?q=FLAG_ACTIVITY_NEW_TASK.*FLAG_ACTIVITY_CLEAR_TASK&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental).

lfkellogg: Agreed :)

mannyjimenez0810: Ah i see. good thought.

lfkellogg (approved):
Looks great, thanks!

PR #3263, opened 2021-12-22
Improve error handling around downloading and installing

lfkellogg:
- Makes error messages and statuses more relevant and specific
- Catches some additional error cases and removes some unnecessary ones
- Replaces all instances of [log-and-(re)throw](http://go/java-practices/exceptions#log_rethrow), instead setting exception causes
- Injects `HttpsUrlConnectionFactory` to avoid mocking package private class methods
- Replaces `Thread.sleep` in tests with `ExecutorServiceawaitTermination()` to speed up tests
- Adds a couple TODO's I will follow up on

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.app.distribution;
 
+import static com.google.firebase.app.distribution.FirebaseAppDistributionException.Status.DOWNLOAD_FAILURE;
 import static com.google.firebase.app.distribution.FirebaseAppDistributionException.Status.NETWORK_FAILURE;
 import static com.google.firebase.app.distribution.TaskUtils.safeSetTaskException;
 
@@ -27,7 +28,7 @@ import com.google.firebase.app.distribution.internal.InstallActivity;
 import com.google.firebase.app.distribution.internal.LogWrapper;
 import com.google.firebase.app.distribution.internal.SignInResultActivity;
 import java.io.IOException;
-import java.net.URL;
+import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import javax.net.ssl.HttpsURLConnection;
 
@@ -36,6 +37,8 @@ class AabUpdater {
   private static final String TAG = "UpdateAabClient:";
 
   private final FirebaseAppDistributionLifecycleNotifier lifecycleNotifier;
+  private final HttpsUrlConnectionFactory httpsUrlConnectionFactory;
+  private final Executor executor;
 
   @GuardedBy("updateAabLock")
   private UpdateTaskImpl cachedUpdateTask;
@@ -46,12 +49,17 @@ class AabUpdater {
   private final Object updateAabLock = new Object();
 
   AabUpdater() {
-    this(FirebaseAppDistributionLifecycleNotifier.getInstance());
+    this(FirebaseAppDistributionLifecycleNotifier.getInstance(), new HttpsUrlConnectionFactory(), Executors.newSingleThreadExecutor());
   }
 
-  AabUpdater(@NonNull FirebaseAppDistributionLifecycleNotifier lifecycleNotifier) {
+  AabUpdater(
+      @NonNull FirebaseAppDistributionLifecycleNotifier lifecycleNotifier,
+      @NonNull HttpsUrlConnectionFactory httpsUrlConnectionFactory,
+      @NonNull Executor executor) {
     this.lifecycleNotifier = lifecycleNotifier;
+    this.httpsUrlConnectionFactory = httpsUrlConnectionFactory;
     lifecycleNotifier.addOnActivityStartedListener(this::onActivityStarted);
+    this.executor = executor;
   }
 
   @VisibleForTesting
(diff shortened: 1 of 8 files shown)


firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/NewReleaseFetcherTest.java
```diff
@@ -330,8 +330,7 @@ public void isSameAsInstalledRelease_ifApkHashNotPresent_throwsError() {
 
   @Test
   public void extractApkHash_ifKeyInCachedApkHashes_doesNotRecalculateZipHash() {
-
-    try (MockedStatic mockedReleaseIdentificationUtils =
+    try (MockedStatic<ReleaseIdentificationUtils> mockedReleaseIdentificationUtils =
```
lfkellogg, line 333: This fixes a compiler warning.

mannyjimenez0810:
Seems like lint errors you might need to run ./gradlew :firebase-app-distribution:googleJavaFormat

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
```diff
@@ -295,12 +271,11 @@ private void setUpdateTaskCompletionError(FirebaseAppDistributionException e) {
     }
   }
 
-  private void setUpdateTaskCompletionErrorWithDefault(
-      Exception e, FirebaseAppDistributionException defaultFirebaseException) {
+  private void setUpdateTaskCompletionErrorWithDefault(Exception e, String message, Status status) {
```
rachaprince, line 274: I like this signature change a lot

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
```diff
@@ -98,55 +128,36 @@ private void redirectToPlayForAabUpdate(String downloadUrl) {
...
-                          .build());
-                }
-              } else {
-                setUpdateTaskCompletionError(
-                    new FirebaseAppDistributionException(
-                        Constants.ErrorMessages.NETWORK_ERROR,
-                        FirebaseAppDistributionException.Status.DOWNLOAD_FAILURE));
+              Intent updateIntent = new Intent(Intent.ACTION_VIEW);
+              updateIntent.setData(Uri.parse(redirectUrl));
+              updateIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
mannyjimenez0810, line 150: We might need to do some manual testing here to make sure we don't run into the same updateApp bug. This is different since it's redirecting to play, but the NEW_TASK flag makes me think of it.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
```diff
@@ -46,12 +49,17 @@
   private final Object updateAabLock = new Object();
 
   AabUpdater() {
-    this(FirebaseAppDistributionLifecycleNotifier.getInstance());
+    this(FirebaseAppDistributionLifecycleNotifier.getInstance(), new HttpsUrlConnectionFactory(), Executors.newSingleThreadExecutor());
```
mannyjimenez0810, line 52: Interested to know the reasoning for passing in an executor and having the HttpsUrlConnectionFactory class. Seems like the executor is only used once to put it in a variable. Is this providing more consistency in the tests?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -185,7 +185,7 @@ HttpsURLConnection openHttpsUrlConnection(
           ErrorMessages.NETWORK_ERROR, Status.NETWORK_FAILURE, e);
     }
     try {
-      httpsURLConnection.setRequestMethod(REQUEST_METHOD);
+      httpsURLConnection.setRequestMethod(REQUEST_METHOD_GET);
```
rachaprince, line 188: Should this class also use the new HttpConnectionFactory?

rachaprince (commented):
Nice work! Great improvements to the code

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/HttpsUrlConnectionFactory.java
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Google LLC
```
mannyjimenez0810, line 1: I believe ApkUpdater also uses the openHttpsUrlConnection. If this factory is going to be used in AAB I believe it should be used in the APK updater as well.

mannyjimenez0810 (approved):
Great job!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
lfkellogg: Yeah exactly! In both cases, the practical reason is that it makes testing easier. It allows the tests to inject an instance that the test has control over. In the case of `HttpsUrlConnectionFactory`, the test gets to inject a mock. (Previously we were mocking a package-private method of the class under test, which isn't ideal.) In the case of the `Executor`, the test gets to call `.awaitTermination()` so that it can be sure that all tasks have completed before asserting things about the resulting state.

Just in general though, classes should require their dependencies be passed in the constructor instead of instantiating them themselves. That way they don't encapsulate the functionality of their dependencies, and can instead be tested in isolation from them. See https://effectivejava.googleplex.com/third-edition#page=41

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
```diff
@@ -98,55 +128,36 @@ private void redirectToPlayForAabUpdate(String downloadUrl) {
         safeSetTaskException(
```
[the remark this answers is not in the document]
lfkellogg: Great idea! We can use the one you made in your PR :) Maybe in a followup?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/AabUpdater.java
lfkellogg: Ah good point. I added a note to the bug bash doc.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionTesterApiClient.java
lfkellogg: Ah, yeah probably. Maybe that's something we can add to the guiding principles, and update as we go? I think the principle would be: prefer injecting dependencies over mocking methods in the class-under-test.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/HttpsUrlConnectionFactory.java
lfkellogg: It is used in the ApkUpdater. You might need to expand that file as it's hidden right now since the diff is so big.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
```diff
@@ -79,52 +89,43 @@ UpdateTaskImpl updateApk(
...
+        .addOnFailureListener(
+            taskExecutor,
+            e -> {
+              postUpdateProgress(
+                  file.length(),
+                  file.length(),
+                  UpdateStatus.INSTALL_FAILED,
+                  showDownloadNotificationManager);
+              setUpdateTaskCompletionErrorWithDefault(
+                  e, ErrorMessages.APK_INSTALLATION_FAILED, Status.INSTALLATION_FAILURE);
```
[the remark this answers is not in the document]
lfkellogg: Yeah I think so. I'd prefer `ErrorMessages.<value>` personally. Just fixed the one `Constants.ErrorMessages` usage in this PR.

I'd like to get rid of the surrounding Constants class, and honestly I'm hoping after cleaning up the error handling codebase-wide we might want to get rid of the ErrorMessages class altogether. I think it's led to overly generic and often misleading error messages all over the place.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
lfkellogg: Glad to hear it!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/HttpsUrlConnectionFactory.java
mannyjimenez0810: Ah thanks! Thats exactly what happened.

lfkellogg:
This has been rebased on top of https://github.com/firebase/firebase-android-sdk/pull/3274. Please ignore until that PR is merged and this can be rebased on master.

Issue #3265, opened 2021-12-23
DiskReadViolation triggered by setCrashlyticsCollectionEnabled() 


DanielNovak:
Describe your environment

  * Android Studio version: 2020.3.1 Patch 4
  * Firebase Component: Crashlytics
  * Component version: 18.2.6

Describe the problem

Firebase Crashlytics is triggering Android StrictMode DiskReadViolation when `setCrashlyticsCollectionEnabled(true)` is called.

```
StrictMode policy violation; ~duration=351 ms: android.os.strictmode.DiskReadViolation
        at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1659)
        at libcore.io.BlockGuardOs.access(BlockGuardOs.java:74)
        at libcore.io.ForwardingOs.access(ForwardingOs.java:131)
        at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:7715)
        at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:281)
        at java.io.File.exists(File.java:813)
        at android.app.SharedPreferencesImpl.writeToFile(SharedPreferencesImpl.java:738)
        at android.app.SharedPreferencesImpl.access$900(SharedPreferencesImpl.java:59)
        at android.app.SharedPreferencesImpl$2.run(SharedPreferencesImpl.java:672)
        at android.app.SharedPreferencesImpl.enqueueDiskWrite(SharedPreferencesImpl.java:691)
        at android.app.SharedPreferencesImpl.access$100(SharedPreferencesImpl.java:59)
        at android.app.SharedPreferencesImpl$EditorImpl.commit(SharedPreferencesImpl.java:604)
        at com.google.firebase.crashlytics.internal.common.DataCollectionArbiter.storeDataCollectionValueInSharedPreferences(DataCollectionArbiter.java:206)
        at com.google.firebase.crashlytics.internal.common.DataCollectionArbiter.setCrashlyticsDataCollectionEnabled(DataCollectionArbiter.java:92)
        at com.google.firebase.crashlytics.internal.common.CrashlyticsCore.setCrashlyticsCollectionEnabled(CrashlyticsCore.java:257)
        at com.google.firebase.crashlytics.FirebaseCrashlytics.setCrashlyticsCollectionEnabled(FirebaseCrashlytics.java:448)
```

#### Steps to reproduce:

Enable strict mode:
```
        StrictMode.setThreadPolicy(
[...]

PR #3266, opened 2021-12-23
issue_3265_disk_read_violation

argzdev:
Proposed fix by DanielNovak for #3265.

TLDR: DiskReadViolation warning is being triggered when setting `setCrashlyticsCollectionEnabled`

The developer found that the issue is caused by `DataCollectionArbitrer.java` where it's using `commit()` instead of `apply()`.

Relevant code:
```
private static void storeDataCollectionValueInSharedPreferences(SharedPreferences sharedPreferences, Boolean enabled) {
        Editor prefsEditor = sharedPreferences.edit();
        if (enabled != null) {
            prefsEditor.putBoolean("firebase_crashlytics_collection_enabled", enabled);
        } else {
            prefsEditor.remove("firebase_crashlytics_collection_enabled");
        }

        prefsEditor.commit();
    }
```

The key difference of `commit` vs `apply` is that apply is asynchronous. Stack Overflow link [here](https://stackoverflow.com/a/5960732/14129564).

Based on the conversation in SO, changing `commit` to `apply` is safe. We can base it on the Android [documentations](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()).

> As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value.

In this case since we don't use the value for anything else, perhaps this is safe to replace.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/DataCollectionArbiter.java
@@ -203,6 +203,6 @@ public class DataCollectionArbiter {
     } else {
       prefsEditor.remove(FIREBASE_CRASHLYTICS_COLLECTION_ENABLED);
     }
-    prefsEditor.commit();
+    prefsEditor.apply();
   }
 }


mrichards (approved):
Thanks @DanielNovak and @argzdev!

Issue #3249, opened 2021-12-17
Query snapshots not updating after returning to foreground


bswhite1:
### Describe your environment

  * Android Studio version: Artic Fox 2020.3.1 Patch 4
  * Firebase Component: Firestore
  * Component version: Firebase BoM 29.0.0, firestore: 24.0.0. Master as of commit: d0048fdfb9e272f32edc49cd7f64a53cb684cf6e

### Describe the problem
Issue first documented here: https://github.com/FirebaseExtended/flutterfire/issues/7323

Updated problem description:
When the app returns to the foreground after being in the background for several hours, the listeners for QuerySnapshots can return wrong data. It only returns updates to the collection that were made while the app was in the background, or new updates after it is returned to the foreground. Unchanged documents are not returned. A force close and restart of the app corrects the issue.

#### Current debug analysis:

As mentioned above, issue was seen in Firebase BoM 29.0.0, firestore: 24.0.0 
Issue is still seen in master as of this commit: https://github.com/firebase/firebase-android-sdk/commit/d0048fdfb9e272f32edc49cd7f64a53cb684cf6e

I have tracked the issue down to this series of events.

Starting with the log where the app is the background.
App is started up and logs look good.
App is sent to the background and appears to go to sleep.
Eventually it may wake up and log something like this
- [WatchStream]: (b390c6) Stream sending: # com.google.firestore.v1.ListenRequest
     - target_id matches target_id of query experiencing issue.
     - labels { value: "existence-filter-mismatch"}
- RemoteStore.raiseWatchSnapshot (snapshotVersion)
- Creates remoteEvent from watchChangeAggregator.createRemoteEvent(snapshotVersion)
     - Parses through remoteEvent.getTargetMismatches()
          - targetID of problem query is present.
          - this.sendUnwatchRequest(targetId)
          - this.sendWatchRequest(requestTargetData)
     - remoteStoreCallback.handleRemoteEvent(remoteEvent);
          - SyncEngine.handleRemoteEvent (event)
               - LocalStore.applyRemoteEvent (remoteEvent)
                    - targetCache.removeMatchingKeys(change.getRemovedDocuments(), targetId); 
                         - SQLiteTargetCache.removeMatchingKeys  **// This removes the documents tied to the targetId in the targetCache.**
               - emitNewSnapsAndNotifyLocalStore(changes, event)
               - listen snapshot appears to still have all of the documents for query.

I do not get any other events or data past this point. Phone appears to go back to sleep 

[...]

PR #3267, opened 2021-12-23
Clear the lastLimboFreeSnapshot version on existence filter mismatch

schmidt-sebastian:
This fixes a problem where the SDK receives an existence filter mismatch and clears the mapping of documents to targets, but does not receive another snapshot that also clears the snapshot version (i.e. the device went in the background and did was not able to actually process a re-run of the query).

The change that this PR makes is that we always reset the lastLimboFreeSnapshot version to 0 if we get a target mismatch.

Fixes https://github.com/firebase/firebase-android-sdk/issues/3249

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -7,6 +7,8 @@ by opting into a release at
   updates that have not yet been synced with the backend.
 - [changed] Improved performance for queries against collections that contain
   subcollections.
+- [fixed] Fixed an issue that may have caused the SDK to raise incomplete 
+  snapshots when backgrounded during query execution.
 
 # 24.0.0
 - [feature] Added support for Firebase AppCheck.
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -7,8 +7,8 @@ by opting into a release at
   updates that have not yet been synced with the backend.
 - [changed] Improved performance for queries against collections that contain
   subcollections.
-- [fixed] Fixed an issue that may have caused the SDK to raise incomplete 
-  snapshots when backgrounded during query execution.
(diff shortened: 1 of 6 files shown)


bswhite1 (commented):
Should the test case in existence_spec_test.json, or somewhere else, check that the targetData.lastLimboFreeSnapshotVersion is reset to none?

Other than that, the code itself is good. I verified via several test runs and never saw the error.

schmidt-sebastian:
@bswhite1 Thanks for verifying. I also added another test case.

bswhite1 (approved):
Looks good.

What is the easiest way to determine when this is part of a release?
Do you ever use git Tags, or release branches?

schmidt-sebastian:
Please take a look at our release notes to determine what is contained in a release. This PR should be included in the next cut.

We use a private GitHub repository to manage our releases. The tags are not visible here.

firebase-firestore/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ by opting into a release at
   updates that have not yet been synced with the backend.
 - [changed] Improved performance for queries against collections that contain
   subcollections.
+- [fixed] Fixed an issue that may have caused the SDK to raise incomplete
```
dconeybe, line 10: "may have caused" sounds to me like you are unsure that this is the root cause; however, it seems that the root cause is definitively known. Would it make sense to re-word this to imply that the buggy behavior is intermittent?

e.g. Fixed an issue that causes the SDK to sometimes raise incomplete snapshots...

dconeybe (changes requested):

firebase-firestore/CHANGELOG.md
schmidt-sebastian: Good idea. Will update

dconeybe (approved):

PR #3272, opened 2021-12-28
Making firebase installations lazy to initialize

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
```diff
@@ -79,10 +80,10 @@
       return cachedCheckForNewRelease;
     }
 
-    Task<String> installationIdTask = firebaseInstallationsApi.getId();
+    Task<String> installationIdTask = firebaseInstallationsApi.get().getId();
```
lfkellogg, line 83: Is it guaranteed that `firebaseInstallationsApi.get()` will return non-null here? I looks OK according to this [design doc](https://docs.google.com/document/d/1cfBSrBeJLv7LWdtva3MYQRLq-ypjHA7J8_MFAAe6L30/edit#heading=h.va4bklgza3d8) but seeing this example from FirePerf concerns me: [comment](https://github.com/firebase/firebase-android-sdk/blob/6a0b405b3a283687fb39854fb0dbe0f7d7b8e49b/firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java#L108-L119), [null-check](https://github.com/firebase/firebase-android-sdk/blob/6a0b405b3a283687fb39854fb0dbe0f7d7b8e49b/firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java).

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
```diff
@@ -44,7 +45,7 @@
 
   private final FirebaseApp firebaseApp;
   private final FirebaseAppDistributionTesterApiClient firebaseAppDistributionTesterApiClient;
-  private final FirebaseInstallationsApi firebaseInstallationsApi;
+  private final Provider<FirebaseInstallationsApi> firebaseInstallationsApi;
```
lfkellogg, line 48: Change this to `firebaseInstallationsApiProvider`

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/TesterSignInManager.java
```diff
@@ -49,7 +50,7 @@
       "https://appdistribution.firebase.google.com/pub/testerapps/%s/installations/%s/buildalerts?appName=%s&packageName=%s";
 
   private final FirebaseApp firebaseApp;
-  private final FirebaseInstallationsApi firebaseInstallationsApi;
+  private final Provider<FirebaseInstallationsApi> firebaseInstallationsApi;
```
lfkellogg, line 53: Change this to `firebaseInstallationsApiProvider`

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/NewReleaseFetcher.java
mannyjimenez0810: Yea FirePerf does seem to have a null-check that we can also do. I am not entirely sure what cases it would return null. Looking at other parts of the sdk it seems like a check is a good idea. https://github.com/firebase/firebase-android-sdk/blob/6a0b405b3a283687fb39854fb0dbe0f7d7b8e49b/firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java#L340

lfkellogg: The design doc says you can't call `.get()` from the constructor, so I think the way you have it now is good - it's called right before we need it.

After looking at the [integration guide](https://docs.google.com/document/d/1L77viPPs5_TgS97K89YneRBthyp1L7u7E0MquArjU4Q/edit#heading=h.dxjfnxhb6bea) I actually don't think we need the null check. Should we ask the ACore folks though?

lfkellogg:
By the way, I notice that most projects only have `runtimeOnly` dependency on `firebase-installations`: https://github.com/firebase/firebase-android-sdk/blob/6a0b405b3a283687fb39854fb0dbe0f7d7b8e49b/firebase-perf/firebase-perf.gradle#L100

Should we try the same?

mannyjimenez0810:
Yea i see crashlytics also has the runtimeOnly. Will make that change.

firebase-app-distribution/firebase-app-distribution.gradle
```diff
@@ -42,7 +42,7 @@ dependencies {
     implementation project(path: ':firebase-components')
     implementation project(path: ':firebase-installations-interop')
     implementation project(path: ':firebase-common')
-    implementation project(path: ':firebase-installations')
+    runtimeOnly project(':firebase-installations')
```
mannyjimenez0810, line 45: Our gradle has (path: something). I didn't see this in other gradle files so I am not sure why we have that. Decided to copy the runtimeOnly code that other projects have although it doesn't fit with our outline. Should we change our gradle file to remove those (path: ) code snippets not sure if they are necessary

lfkellogg: Yeah let's remove that since it seems unnecessary.

lfkellogg (approved):

PR #3274, opened 2021-12-28
Refactor notification tests to be more focused to the classes under test

lfkellogg:
This should also make `UpdateApkTest` less flaky.

### DIFF
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
@@ -67,10 +67,10 @@ class ApkUpdater {
       @NonNull FirebaseApp firebaseApp,
       @NonNull ApkInstaller apkInstaller,
       @NonNull FirebaseAppDistributionNotificationsManager appDistributionNotificationsManager) {
-    this.appDistributionNotificationsManager = appDistributionNotificationsManager;
     this.taskExecutor = taskExecutor;
     this.firebaseApp = firebaseApp;
     this.apkInstaller = apkInstaller;
+    this.appDistributionNotificationsManager = appDistributionNotificationsManager;
   }
 
   UpdateTaskImpl updateApk(
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
@@ -295,8 +295,7 @@ class ApkUpdater {
     }
   }
 
-  @VisibleForTesting
-  void postUpdateProgress(
+  private void postUpdateProgress(
       long totalBytes, long downloadedBytes, UpdateStatus status, boolean showNotification) {
     synchronized (updateTaskLock) {
       cachedUpdateTask.updateProgress(
--- a/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
+++ b/firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
@@ -41,12 +41,12 @@ class ApkUpdater {
   private static final int UPDATE_INTERVAL_MS = 250;
   private static final String TAG = "UpdateApkClient:";
   private static final String REQUEST_METHOD = "GET";
-  private final FirebaseAppDistributionNotificationsManager appDistributionNotificationsManager;
 
   private TaskCompletionSource<File> downloadTaskCompletionSource;
   private final Executor taskExecutor; // Executor to run task listeners on a background thread
   private final FirebaseApp firebaseApp;
   private final ApkInstaller apkInstaller;
+  private final FirebaseAppDistributionNotificationsManager appDistributionNotificationsManager;
 
   @GuardedBy("updateTaskLock")
   private UpdateTaskImpl cachedUpdateTask;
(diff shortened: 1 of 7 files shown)


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -81,60 +91,21 @@ private NotificationManager createNotificationManager(Context context) {
       channel.setDescription(context.getString(R.string.notifications_channel_description));
       // Register the channel with the system; you can't change the importance
       // or other notification behaviors after this
-      NotificationManager notificationManager = context.getSystemService(NotificationManager.class);
+      NotificationManager notificationManager =
+          (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
```
lfkellogg, line 95: This seems to be the more common way of getting a `NotificationManager`.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
```diff
@@ -41,35 +41,40 @@
...
+      @NonNull FirebaseAppDistributionNotificationsManager appDistributionNotificationsManager) {
+    this.appDistributionNotificationsManager = appDistributionNotificationsManager;
     this.taskExecutor = taskExecutor;
     this.firebaseApp = firebaseApp;
     this.apkInstaller = apkInstaller;
   }
 
   UpdateTaskImpl updateApk(
-      @NonNull AppDistributionReleaseInternal newRelease, boolean showDownloadNotificationManager) {
+      @NonNull AppDistributionReleaseInternal newRelease, boolean showNotification) {
```
mannyjimenez0810, line 77: Curious why the renaming to showNotification? Seems like this is more vague

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionNotificationsManager.java
mannyjimenez0810: Awesome! Do you mind linking to where else you have seen this?

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
```diff
@@ -306,10 +297,7 @@ private void setUpdateTaskCompletionErrorWithDefault(
 
   @VisibleForTesting
   void postUpdateProgress(
```
mannyjimenez0810, line 299: Is this method no longer being tested in this refactor? if so why was that the case and should this then be package-private without the visibleForTesting annotation?

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionNotificationsManagerTest.java
```diff
@@ -14,86 +14,74 @@
...
+            ApplicationProvider.getApplicationContext());
   }
 
   @Test
-  public void getSmallIcon_whenAppIconSet_usesAppIcon() {
-    setupApplicationInfo(R.drawable.test_app_icon);
-    int iconId = notificationsManager.getSmallIcon();
-
-    assertEquals(R.drawable.test_app_icon, iconId);
+  public void updateNotification_withProgress() {
```
mannyjimenez0810, line 51: Awesome new tests!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
```diff
@@ -41,35 +41,40 @@
...
+  @VisibleForTesting
   public ApkUpdater(
       @NonNull Executor taskExecutor,
       @NonNull FirebaseApp firebaseApp,
-      @NonNull ApkInstaller apkInstaller) {
-    this.appDistributionNotificationsManager =
-        new FirebaseAppDistributionNotificationsManager(firebaseApp);
+      @NonNull ApkInstaller apkInstaller,
+      @NonNull FirebaseAppDistributionNotificationsManager appDistributionNotificationsManager) {
+    this.appDistributionNotificationsManager = appDistributionNotificationsManager;
```
mannyjimenez0810, line 70: Curious as to the constructor change. And setting the notification manager should follow the order of the constructor and go at the end.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionNotificationsManager.java
lfkellogg: First of all I just noticed we're already using the String constant method on line 99 below.

Using the String constant is used pretty overwhelmingly throughout the repo: https://github.com/firebase/firebase-android-sdk/search?q=getSystemService

But for the NotificationManager specifically, using the class actually seems more common. So I don't know, it doesn't really matter I guess. I'm inclined to change it to at least make it consistent with line 99.

firebase-app-distribution/src/test/java/com/google/firebase/app/distribution/FirebaseAppDistributionNotificationsManagerTest.java
lfkellogg: Thanks!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
mannyjimenez0810: Sounds good!

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistributionNotificationsManager.java
mannyjimenez0810: Ah good catch making it more consistent

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/ApkUpdater.java
lfkellogg: Ah, yeah good catch. Yeah before this method was only made visible for testing so that the actual notification display functionality could be tested. That functionality is now covered by AppIconSourceTest and FirebaseAppDistributionNotificationsManagerTest.

I've made it private and removed the annotation.

mannyjimenez0810 (approved):

PR #3285, opened 2022-01-04
Showing error when basic config is called in background thread

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -170,6 +178,10 @@ public UpdateTask updateIfNewReleaseAvailable() {
     }
   }
 
+  private boolean notOnMainThread() {
```
mannyjimenez0810, line 181: Method doesn't do much so I can inline it. My initial thought would be that giving it a name could clarify what we are checking.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
```diff
@@ -135,6 +136,13 @@ public UpdateTask updateIfNewReleaseAvailable() {
 
       cachedUpdateIfNewReleaseTask = new UpdateTaskImpl();
     }
+    if (notOnMainThread()) {
```
mannyjimenez0810, line 139: This sets the exception for the task but it doesn't throw the message into logcat. the log statement before it ensures that logcat shows some error message. Did not want to throw in this method because then the user would have to deal with that situation.  

Other thought could be doing this in the TesterSignInManager since the error occurs on the dialog builder.

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
rachaprince: I'd recommend renaming this method so the boolean represents the positive check (isOnMainThread), rather than the negative check. This helps prevent double negatives, which can be tricky to read (More about this stylisitic guide here: go/tott/530)

rachaprince (approved):

firebase-app-distribution/src/main/java/com/google/firebase/app/distribution/FirebaseAppDistribution.java
mannyjimenez0810: I was thinking of my last PR https://github.com/firebase/firebase-android-sdk/pull/3259#discussion_r773464338 In the spirit of not negating the return value thoughts on isOnBackgroundThread?

rachaprince: That sounds good too

PR #3287, opened 2022-01-05
Add keep on ComponentRegistrar to prevent renaming of key lookup

mkj-gram:
The name `com.google.firebase.components.ComponentRegistrar` is used as a key for looking up in the manifest by ComponentDiscovery. If a developer uses `-adaptclassstrings` the string will be renamed and fail to lookup the correct name in the manifest. One such bug was reported at https://issuetracker.google.com/issues/209817768.

The developer should never use `-adaptclassstrings` without specifying a filter but since the code uses a type name for component discovery it is best practice to keep the name to ensure it is not renamed.


rlazo:
@mkj-gram thanks for your contribution! Could you try syncing your branch to head and re-running the tests (commenting /retest)? Not sure what's causing the failures in check-changed.

rlazo:
Also, please wait until https://github.com/firebase/firebase-android-sdk/pull/3384 is in to rebase your branch to make sure the fix is included. Thanks!

rlazo (approved):

rlazo:
Test failing is not related to the changes. Approved. Thanks!

PR #3298, opened 2022-01-10
Removing dependency for activities

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
```diff
@@ -93,7 +97,7 @@ void handleAppResume(Activity activity) {
         cachedInstallApkPath = path;
       } else {
         // only start the install activity if current Activity is in the foreground
-        startInstallActivity(path, currentActivity);
+        startInstallActivity(path);
```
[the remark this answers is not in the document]
mannyjimenez0810: Yes it can!

rachaprince (commented):
LGTM so far!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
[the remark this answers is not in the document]
rachaprince: Can we now start the Install Activity regardless of whether or not the customer's app is in the foreground?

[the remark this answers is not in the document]
mannyjimenez0810: Good question! Currently working on trying to see if there is a set up to make that work. In its current set up if the user backgrounds the app it seems to hang.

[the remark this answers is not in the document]
mannyjimenez0810: It seems like this is not possible due to Install Activity relying on an activity to show unknown sources UI.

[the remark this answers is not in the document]
lfkellogg: That is surprising. I'd think that once the InstallActivity starts, then it's in the foreground and we can show the dialog. That's what happened when you were calling UpdateApp from a background thread, which worked.

What error/stacktrace are you seeing in the logs?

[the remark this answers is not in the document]
mannyjimenez0810: I get a windowLeaked error. I can repro and show the error message. I believe that was because the app itself wasn't in the background, but once you background the app it wants to display the unknown sources UI but since the app is in the background you get this leaked error.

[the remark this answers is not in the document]
lfkellogg: Why does the app get backgrounded? Maybe we could look at it together over VC.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
```diff
@@ -111,6 +111,7 @@ private void startInstallActivity(String path, Activity currentActivity) {
     }
     Intent intent = new Intent(currentActivity, InstallActivity.class);
     intent.putExtra("INSTALL_PATH", path);
+    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
lfkellogg, line 114: Why add this flag?

mannyjimenez0810: Good catch! Leftover from trying to use application context. When using application Context it needs to be a new task but since that was taken out this should also be taken out.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
```diff
@@ -48,18 +50,22 @@
   private AppDistributionReleaseInternal aabReleaseInProgress;
 
   private final Object updateAabLock = new Object();
+  private final Context applicationContext;
```
lfkellogg, line 53: Nit: for brevity, can we just call this `context` everywhere instead of `applicationContext`?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
```diff
@@ -48,18 +50,22 @@
...
 
-  AabUpdater() {
+  AabUpdater(@NonNull FirebaseApp firebaseApp) {
```
lfkellogg, line 55: This constructor should just take a Context object since that's all you need.

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/AabUpdaterTest.java
```diff
@@ -84,7 +98,8 @@ public void setup() throws IOException, FirebaseAppDistributionException {
 
     aabUpdater =
         Mockito.spy(
-            new AabUpdater(mockLifecycleNotifier, mockHttpsUrlConnectionFactory, testExecutor));
+            new AabUpdater(
+                firebaseApp, mockLifecycleNotifier, mockHttpsUrlConnectionFactory, testExecutor));
```
lfkellogg, line 102: If you take my suggestion and require a `Context` here instead of a `FirebaseApp`, then you can just pass in `ApplicationProvider.getApplicationContext()`.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
lfkellogg: Oh interesting. Why does it need to a be a new task when using application context?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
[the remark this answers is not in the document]
lfkellogg: We discussed this in person. Previously, backgrounding the app after starting the basic config would fail the task correctly. With this change it instead hangs the task.

I'm fine punting on this for now.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
mannyjimenez0810: Good thought! I had done this elsewhere before should do it here as well.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
mannyjimenez0810: I did see we used context elsewhere. Context being so vague is the reason I added application to it, but I have no issue making it context.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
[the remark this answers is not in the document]
mannyjimenez0810: To clarify it hangs if we don't depend on the currentActivity at all with checking what to do with InstallActivity.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
mannyjimenez0810: I ran into the issue but this post sums up the issue and the reason for the fix. https://stackoverflow.com/questions/3918517/calling-startactivity-from-outside-of-an-activity-context

lfkellogg: Hmm, ok the fact that you saw that error is making me less confident in my suggestion to use intents + activities. It feels weird to have to start a new task every time.

rachaprince (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
```diff
@@ -193,14 +195,14 @@ private static boolean isResponseSuccess(int responseCode) {
...
-    Context context = firebaseApp.getApplicationContext();
     context.deleteFile(fileName);
     int fileCreationMode =
         VERSION.SDK_INT >= VERSION_CODES.N ? Context.MODE_PRIVATE : Context.MODE_WORLD_READABLE;
     long bytesDownloaded = 0;
     try (BufferedOutputStream outputStream =
-            new BufferedOutputStream(context.openFileOutput(fileName, fileCreationMode));
-        InputStream inputStream = connection.getInputStream()) {
+            new BufferedOutputStream(
+                context.openFileOutput(fileName, fileCreationMode));
```
lfkellogg, line 204: This can go back to being a one-liner.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
```diff
@@ -241,9 +243,11 @@ private void validateJarFile(
 
   private String getApkFileName() {
     try {
-      Context context = firebaseApp.getApplicationContext();
       String applicationName =
-          context.getApplicationInfo().loadLabel(context.getPackageManager()).toString();
+          context
+              .getApplicationInfo()
+              .loadLabel(context.getPackageManager())
+              .toString();
```
lfkellogg, line 250: This can also be a one-liner

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -148,7 +148,6 @@ public UpdateTask updateIfNewReleaseAvailable() {
         return cachedUpdateIfNewReleaseTask;
       }
     }
-
```
lfkellogg, line 151: Nit: can you put this empty line back?

lfkellogg (approved):
Some tiny nits. Looks great!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
mannyjimenez0810: We do this in the AAB updater https://github.com/firebase/firebase-android-sdk/blob/712e5b88c75e6ad8838d21c904a789f37bcfa0a7/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java#L160 With the application context we aren't guaranteed an activity so the flag makes sense to me.

lfkellogg: I think it makes more sense in the AAB updater since we are launching an activity in a different app, Play.

In the case I was proposing, where we would launch a new intent/activity in order to simply show a dialog in the app, we wouldn't want an additional entry in the task backstack for that. It should ideally just continue in the same task.

rachaprince: This error is especially interesting, since we should be starting the install activity as a new task anyway according to the manifest I think https://github.com/firebase/firebase-android-sdk/blob/master/firebase-app-distribution/src/main/AndroidManifest.xml#L34-L35

rachaprince: From the docs:
`singleTask`
The system creates a new task and instantiates the activity at the root of the new task.

firebase-app-distribution/src/test/java/com/google/firebase/appdistribution/AabUpdaterTest.java
```diff
@@ -72,6 +77,15 @@ public void setup() throws IOException, FirebaseAppDistributionException {
...
     FirebaseApp.clearInstancesForTest();
 
+    FirebaseApp firebaseApp =
+        FirebaseApp.initializeApp(
+            ApplicationProvider.getApplicationContext(),
+            new FirebaseOptions.Builder()
+                .setApplicationId(TEST_APP_ID_1)
+                .setProjectId(TEST_PROJECT_ID)
+                .setApiKey(TEST_API_KEY)
+                .build());
```
lfkellogg, line 87: In case you missed my other comment, you don't need to a create a `FirebaseApp` here. You can just inject `ApplicationProvider.getApplicationContext()` into the constructor, below.

mannyjimenez0810: I did miss this comment. Will fix now.

lfkellogg (approved):

PR #3305, opened 2022-01-11
Construct FirebaseAppDistribution on startup

rachaprince:
Our SDK depends on Activity Lifecycle callbacks to manage SDK state. While we do plan to refactor some of our Activity Context dependencies, these lifecycle-dependent state changes are likely out of scope for the current refactor.  

This PR constructs FirebaseAppDistribution on app start so that the activity lifecycle callbacks are always registered before the customer uses the API. This will allow us to remove the following caveat from our docs:

```You must initialize the FirebaseAppDistribution instance in the onCreate() method of anActivity```


mannyjimenez0810 (approved):
Awesome!

PR #3306, opened 2022-01-11
Improve testIgnoresTargetMappingAfterExistenceFilterMismatch

schmidt-sebastian:
This makes the test match reality more as the target has synced keys when it hits the existence filter mismatch.


dconeybe (approved):
Can you just add a description to the PR to explain what the improvements are. Is this just more useful test failure messages?

schmidt-sebastian:
Updated description

PR #3310, opened 2022-01-12
Removing failures when app is backgrounded

mannyjimenez0810:


firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java
```diff
@@ -111,6 +111,39 @@ void onActivityStarted(Activity activity) {
     }
   }
 
+  @NonNull
```
lfkellogg: Just a reminder: the changes to this file can be reverted

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -244,6 +232,11 @@ private boolean isOnBackgroundThread() {
     return this.testerSignInManager.signInTester();
   }
 
+  @NonNull
+  public Task<Void> signInTester(boolean showDialog) {
+    return this.testerSignInManager.signInTester(showDialog);
+  }
```
lfkellogg: Just a reminder: remove this method

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -80,6 +91,25 @@ Activity getCurrentActivity() {
     }
   }
 
+  // TODO (mannyjimenez) Put a lock on this in the future.
```
lfkellogg, line 94: Oh yeah, we should fix this before submitting. I think you can just wrap this entire method body in a `synchronized (lock) {}`, but then you can't call `addOnActivityResumedListener` and `removeOnActivityResumedListener` since they need the lock. Instead you'll have to inline the `this.onActivityResumedListeners.add/remove` calls.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -179,7 +171,10 @@ public UpdateTask updateIfNewReleaseAvailable() {
                 setCachedUpdateIfNewReleaseResult();
                 return Tasks.forResult(null);
               }
-              return showUpdateAlertDialog(release);
+              return lifecycleNotifier
+                  .getForegroundActivity()
+                  .onSuccessTask((activity) -> showUpdateAlertDialog(activity, release));
+              // return showUpdateAlertDialog(release);
```
lfkellogg, line 177: Remove this

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
```diff
@@ -53,16 +61,15 @@ protected void onResume() {
 
     if (!isUnknownSourcesEnabled()) {
       // See comment about install progress above. Same applies to unknown sources UI.
-      if (unknownSourceEnablementInProgress) {
+      if (alertDialog.isShowing()) {
         LogWrapper.getInstance()
             .e(
                 TAG
                     + "Unknown sources enablement was already in progress. It was either canceled or failed");
+        alertDialog.dismiss();
```
lfkellogg, line 69: Ah, does this avoid a "window leaked error" or something?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
```diff
@@ -34,6 +35,13 @@
   private static final String TAG = "InstallActivity: ";
   private boolean unknownSourceEnablementInProgress = false;
```
lfkellogg, line 36: Can you remove this member completely?

lfkellogg (commented):
Looking good so far!

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java
mannyjimenez0810: Yup did a fast follow.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
mannyjimenez0810: Yes it seems that we were not dismissing the dialog as this gets destroyed and so before we finish I decided to dismiss the dialog.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
rachaprince: I'm missing a lot of context, but I am curious if we can call `addOnActivityResumedListener` from within a synchronized block, since the lock is supposed to be reentrant ( A reentrant lock allows a thread that is holding a lock to acquire it again.) Did you test this and find deadlock?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -108,6 +138,18 @@ void addOnActivityDestroyedListener(@NonNull OnActivityDestroyedListener listene
     }
   }
 
+  void addOnActivityResumedListener(@NonNull OnActivityResumedListener listener) {
+    synchronized (lock) {
+      this.onActivityResumedListeners.add(listener);
+    }
+  }
+
+  void removeOnActivityResumedListener(@NonNull OnActivityResumedListener listener) {
```
rachaprince, line 147: Could we add comments or java docs to indicate when we should be removing listeners? Otherwise I feel like we could run into the same memory leaks in the future

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -80,6 +91,25 @@ Activity getCurrentActivity() {
...
+          public void onResumed(Activity activity) {
+            task.setResult(activity);
+            removeOnActivityResumedListener(this);
+          }
+        });
+
+    return task.getTask();
+  }
+
   Activity getNonNullCurrentActivity() throws FirebaseAppDistributionException {
```
rachaprince, line 113: Do we still need this method? Could we remove the `APP_BACKGROUNDED` and `FOREGROUND_ACTIVITY_NOT_AVAILABLE` methods?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
rachaprince: What happens if the Dialog is showing and we call finish()? (for example, if the if statement above is triggered). Do we need to dismiss the dialog in the onDestroy method of this action, too?

rachaprince (commented):
Love this approach! Great work y'all

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
mannyjimenez0810: Good point! That path can be tested and see if deadlock occurs

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
mannyjimenez0810: This is being called before finish so that should be ok. A scenario I am testing now is if the activity gets destroyed.

rachaprince: What if we trigger the finish() on line 50 while the dialog is showing?

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
mannyjimenez0810: removed the method and error codes.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
```diff
@@ -53,23 +60,33 @@ protected void onResume() {
...
       return;
     }
 
     startAndroidPackageInstallerIntent();
   }
 
+  protected void onDestroy() {
+    super.onDestroy();
+    if (alertDialog.isShowing()) {
+      alertDialog.dismiss();
```
mannyjimenez0810, line 82: This prevents the window leak error. My guess is once we deal with the activity and rotations this may be unnecessary, but this will get rid of the error in the meantime.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
rachaprince: Sounds good. @lfkellogg or @mrichards do either of you have a quick explanation? Otherwise we can discuss over VC

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
rachaprince: Would you be able to give me a TL;DR on the window leak error?

rachaprince (approved):

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionException.java
```diff
@@ -42,10 +42,6 @@
     /** Installation canceled */
     INSTALLATION_CANCELED,
 
-    /** No foreground activity available for a given intent */
-    // TODO(rachelprince): add this to API council review
-    FOREGROUND_ACTIVITY_NOT_AVAILABLE,
```
lfkellogg, line 47: Please remove this from from the equivalent file in `firebase-app-distribution-stub` as well, and update that api.txt file. At some point we'll add a submit check to make sure the API surfaces are the same.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
```diff
@@ -53,23 +60,32 @@ protected void onResume() {
...
+      LogWrapper.getInstance()
+          .e(TAG + "Unknown sources enablement canceled. Activity was destroyed");
+      finish();
```
lfkellogg, line 85: Looks like there's no point in calling `finish()` from onDestroy. From the doc for onDestroy: "Perform any final cleanup before an activity is destroyed. This can happen either because the activity is finishing (someone called finish() on it), ..."

So onDestroy is a result of calling finish(). 

https://developer.android.com/reference/android/app/Activity#onDestroy()

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionException.java
mannyjimenez0810: Good call! will do

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
mannyjimenez0810: I was able to confirm this. Checking to see if window leaked error was on my side

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
```diff
@@ -85,7 +85,7 @@ void handleAppResume(Activity activity) {
 
   Task<Void> installApk(String path) {
     synchronized (installTaskLock) {
-      Activity currentActivity = lifeCycleNotifier.getCurrentActivity();
+      Activity currentActivity = lifeCycleNotifier.getForegroundActivity().getResult();
```
lfkellogg: Calling getResult() will throw if there's no foreground activity.

Instead should we pass the activity to this method as an argument? Then wherever this is called, we can chain it all together:

```
lifeCycleNotifier.getForegroundActivity()
  .onSuccessTask(activity -> apkInstaller.installApk(path, activity))
  ...
```

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/internal/InstallActivity.java
```diff
@@ -32,8 +33,14 @@
  */
 public class InstallActivity extends AppCompatActivity {
   private static final String TAG = "InstallActivity: ";
-  private boolean unknownSourceEnablementInProgress = false;
   private boolean installInProgress = false;
+  private AlertDialog alertDialog;
```
lfkellogg, line 37: Can you rename this variable to `enableUnknownSourcesDialog` or something? It would help me understand the logic behind the `if` statement in onResume, for example.

firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
mannyjimenez0810: Good catch! Fixed to implement that change.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkInstallerTests.java
```diff
@@ -65,9 +65,9 @@ public void installApk_currentActivityNotNull_InstallIntentOnCurrentActivity() {
 
   @Test
   public void installApk_currentActivityNull_InstallNotPrompted() {
-    when(mockLifecycleNotifier.getCurrentActivity()).thenReturn(null);
+    when(mockLifecycleNotifier.getForegroundActivity()).thenReturn(Tasks.forResult(null));
```
lfkellogg, line 68: The foreground activity should never be null. We should delete this test case.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
```diff
@@ -83,9 +83,8 @@ void handleAppResume(Activity activity) {
     }
   }
 
-  Task<Void> installApk(String path) {
+  Task<Void> installApk(String path, Activity currentActivity) {
     synchronized (installTaskLock) {
-      Activity currentActivity = lifeCycleNotifier.getCurrentActivity();
       // This ensures that we save the state of the install if the app is backgrounded during
       // APK download
       if (currentActivity == null) {
```
lfkellogg, line 90: We no longer have to account for this case, since `currentActivity` will never be null.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
```diff
@@ -110,26 +112,33 @@ UpdateTaskImpl updateApk(
...
-                  file.length(),
-                  UpdateStatus.INSTALL_FAILED,
-                  showDownloadNotificationManager);
-              setUpdateTaskCompletionErrorWithDefault(
-                  e, ErrorMessages.APK_INSTALLATION_FAILED, Status.INSTALLATION_FAILURE);
-            });
+            activity ->
+                apkInstaller
+                    .installApk(file.getPath(), activity)
+                    .addOnSuccessListener(
```
lfkellogg, line 122: Can we chain this and `addOnFailureListener` below off of `onSuccessTask` instead of `installApk`? It should not make much of a difference except that if the Task returned by getForegroundActivity() did fail for some reason, then we would correctly fail this class's Task. Right now I believe that Task would hang in that case. Maybe we should add a test for that case, even though it's unlikely.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkInstallerTests.java
mannyjimenez0810: talked offline will delete

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkInstaller.java
mannyjimenez0810: will check with manual test but will delete.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
```diff
@@ -169,6 +169,35 @@ public void updateApk_whenInstallSuccessful_setsResult() throws Exception {
     assertThat(updateTask.isSuccessful()).isTrue();
   }
 
+  @Test
+  public void updateApk_getForeGroundActivityFails_setsError() {
+    boolean showNotification = true;
+    doReturn(Tasks.forResult(mockFile))
+        .when(apkUpdater)
+        .downloadApk(TEST_RELEASE, showNotification);
+    TaskCompletionSource<Activity> installTaskCompletionSource = new TaskCompletionSource<>();
```
lfkellogg, line 178: Can you change the name of this to something like `getForegroundActivityCompletionSource`?

lfkellogg (approved):
Looks great just one nit!

PR #3311, opened 2022-01-13
Fireperf: fix screen trace in multi-activity apps

leotianlizhan:
b/210055697

### Cause:
In multi-activity apps, when Activity 1 starts Activity 2, `FrameMetricsAggregator` does not work when `frameMetricsAggregator.add(activity2)` is called before `reset()` of activity1. That is the current behaviour because we use `onStart` and `onStop`. The order of these lifecycle callbacks are well defined in the activity lifecycle documentation, "[Coordinating activities](https://developer.android.com/guide/components/activities/activity-lifecycle#coordinating-activities)" section. 

### Fix:
Move the "stopping of screen traces" from `onActivityStopped` to `onActivityPostPaused`. According to the same [documentation](https://developer.android.com/guide/components/activities/activity-lifecycle#coordinating-activities), the order of these callbacks are well defined, and it is **guaranteed that activity 1's `onPause` will happen before activity 2's `onStart`**. 

The reason `onActivityPostPaused` is picked over `onActivityPaused` is because **`onActivityPaused` is called [at the beginning](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java;l=322) of `onPause`**, and **`onPause` [then calls each fragment's](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java;l=324) `onPause`**. If `reset()` is called in `onActivityPaused`, then the last fragment in the activity won't have a chance to read its frame metrics. 

### Consequence:
Fragment screen traces **must use `onPause`** instead of `onStop` because `frameMetricsAggregator.reset()` is moved from `onStop` to `onPause`.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -324,7 +324,9 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
      */
     try {
       frameMetricsAggregator.remove(activity);
-    } catch (IllegalArgumentException ignored) {}
+    } catch (IllegalArgumentException ignored) {
+      logger.debug("View not hardware accelerated. Unable to collect screen trace.");
+    }
     SparseIntArray[] arr = frameMetricsAggregator.reset();
     if (arr != null) {
       SparseIntArray frameTimes = arr[FrameMetricsAggregator.TOTAL_INDEX];
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -163,10 +163,6 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
 
   @Override
   public synchronized void onActivityStopped(Activity activity) {
-    if (isScreenTraceSupported(activity)) {
-      sendScreenTrace(activity);
-    }
-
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -318,6 +322,11 @@ private void sendScreenTrace(Activity activity) {
      * FrameMetricsAggregator#remove(Activity)} will throw exceptions for hardware acceleration
      * disabled activities.
      */
+    try {
+      frameMetricsAggregator.remove(activity);
```
jeremyjiang-dev, line 326: What will happen if you do not remove the activity and only reset?

leotianlizhan: The way FrameMetricsAggregator works is by [registering a listener](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/Window.java;l=928;drc=master;bpv=1;bpt=1) to the DecorView of a `Window` when you call `FrameMetricAggregator.add(activity)`. Purely for consistency sake I think it is important to remove the listener when we don't want to monitor that activity anymore.

jeremyjiang-dev: Sounds good for consistency

jeremyjiang-dev (approved):

PR #3316, opened 2022-01-14
Use gzip

VinayGuthal:


firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
```diff
@@ -99,7 +101,11 @@
             JSONObject output = new JSONObject();
             output.put("heartbeats", array);
             output.put("version", "2");
-            return Base64.encodeToString(output.toString().getBytes(), Base64.DEFAULT);
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            GZIPOutputStream gzip = new GZIPOutputStream(out);
+            gzip.write(output.toString().getBytes());
+            gzip.close();
+            return Base64.encodeToString(out.toString("UTF-8").getBytes(), Base64.URL_SAFE);
```
carlonzo, line 108: dont you need to close `out` as well?

VinayGuthal: It is safe to not flush() or not close() a ByteArrayOutputStream. it makes no difference to memory usage

vkryachko (approved):

PR #3335, opened 2022-01-20
Showing Dialogs again on screen rotation

mannyjimenez0810:


lfkellogg (commented):
This is a clever way to do this, but I'm not sure it's the best approach in our case. We'd be saving our SDK's state in the Bundle that will actually be presented to the host Activity in onCreate/onRestoreInstanceState. While we could use a very unique key to avoid collisions with the host Activity's own state, it still doesn't feel great to leak our state into their own. I think we should avoid unless it's absolutely necessary. And I don't think it's necessary because unlike the activities, which get destroyed and recreated, the FirebaseAppDistribution instance will live on as long as the app is not killed. So could we just save the state as member variables in FirebaseAppDistribution?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -326,6 +329,14 @@ void onActivityDestroyed(@NonNull Activity activity) {
       // SignInResult is internal to the SDK and is destroyed after creation
       return;
     }
+    if (activity.isChangingConfigurations()) {
+      remakeSignInConfirmationDialog =
+          signInConfirmationDialog != null && signInConfirmationDialog.isShowing();
+      remakeUpdateDialog = (updateDialog != null && updateDialog.isShowing());
```
lfkellogg, line 335: No need for the surrounding parens.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -339,6 +350,16 @@ void onActivityDestroyed(@NonNull Activity activity) {
     }
   }
 
+  void onActivityCreated(Activity activity) {
+    if (remakeSignInConfirmationDialog) {
+      remakeSignInConfirmationDialog = false;
+      signInConfirmationDialog.show();
```
lfkellogg, line 356: I'm kind of surprised this works. I figured we'd need to call `showSignInConfirmationDialog` with the new activity instance.

Will this work in all cases? Might be worth trying on multiple API levels.

lfkellogg (commented):
Is it possible to unit test this?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -64,6 +64,8 @@
   private Task<AppDistributionRelease> cachedCheckForNewReleaseTask;
   private AlertDialog updateDialog;
   private AlertDialog signInConfirmationDialog;
+  private boolean remakeUpdateDialog = false;
```
rachaprince, line 67: Is there any risk of multiple threads reading these values? do these need to be guarded by locks?

rachaprince (commented):
I feel like some sort of test would be helpful for this. Even if we just end up "imitating" the screen rotation by calling the activity lifecycle callbacks. (I believe we've done this before in other tests)

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
mannyjimenez0810: That is a good question. This is only changed in the onDestroyed and the onCreate. I am unsure if other threads will be able to change the value since it is tied to the activity.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
mannyjimenez0810: Hm thats true not sure why it was working. It is a different instance of the activity when it gets recreated it is possible that it was due to launching from the application context but changed to match rachel's PR.

mannyjimenez0810:
Want to discuss the way the test works and see if there is a better option as well as discussing rachel's point about having a sync block. Some of this code will change with the next PR that seems to be doing some similar interactions https://github.com/firebase/firebase-android-sdk/pull/3341/files . Am happy to combine the two or have a follow up with that one.

lfkellogg:
@mannyjimenez0810 looking again at this, I think we can definitely do some sort of basic unit test, like Rachel suggested. Let's chat.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -452,20 +453,36 @@ public void updateToNewRelease_receiveProgressUpdateFromUpdateApp() {
   }
 
   @Test
-  public void taskCancelledOnScreenRotation_whenUpdateDialogShowing() {
+  public void signInConfirmationDialogReappears_whenScreenRotates_whenSignInDialogShowing() {
```
lfkellogg, line 456: Nit, can you rename these test names to use the format:

nameOfMethodCalled_conditionsForThisTest_expectedBehavior?

go/unit-testing-practices#naming

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -452,20 +453,36 @@ public void updateToNewRelease_receiveProgressUpdateFromUpdateApp() {
...
+    when(mockSignInStorage.getSignInStatus()).thenReturn(false);
+    when(activity.isChangingConfigurations()).thenReturn(true);
+
+    UpdateTask updateTask = firebaseAppDistribution.updateIfNewReleaseAvailable();
+
+    // Mimic activity dying
```
lfkellogg, line 462: Maybe "Mimic activity recreated due to screen rotation" ?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -452,20 +453,36 @@ public void updateToNewRelease_receiveProgressUpdateFromUpdateApp() {
...
+
+  @Test
+  public void updateDialogReappears_whenScreenRotates_whenSignInDialogShowing() {
     AppDistributionReleaseInternal newRelease = TEST_RELEASE_NEWER_AAB_INTERNAL.build();
     when(mockNewReleaseFetcher.checkForNewRelease()).thenReturn(Tasks.forResult(newRelease));
+    when(activity.isChangingConfigurations()).thenReturn(true);
 
     UpdateTask updateTask = firebaseAppDistribution.updateIfNewReleaseAvailable();
 
     // Mimic activity dying
```
lfkellogg, line 477: Same here, maybe "Mimic activity recreated due to screen rotation" ?

lfkellogg (approved):
Looks good just some nits!

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -452,20 +453,36 @@ public void updateToNewRelease_receiveProgressUpdateFromUpdateApp() {
...
     firebaseAppDistribution.onActivityDestroyed(activity);
+    firebaseAppDistribution.onActivityCreated(activity);
 
-    assertTaskFailure(updateTask, INSTALLATION_CANCELED, UPDATE_CANCELED);
+    assertAlertDialogShown();
```
lfkellogg, line 481: Oh, can you also add assertions on these tests checking that the task does not get canceled?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
mannyjimenez0810: yea there isn't one method we are calling so wasn't sure what to do with that.

lfkellogg: Ah yeah, to be clear it's more like the method under test, which in this case would be `updateIfNewReleaseAvailable`.

mannyjimenez0810: cool i was leaning towards using the basic config since thats what we call. Thanks!

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
mannyjimenez0810: will do!

lfkellogg (approved):

PR #3336, opened 2022-01-20
Update firebase-appdistribution maven ID in a few last places

lfkellogg:


rachaprince (approved):
Thanks for catching these!

mannyjimenez0810 (approved):

PR #3340, opened 2022-01-21
Await foreground activity during sign in

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TaskUtils.java
```diff
@@ -84,6 +87,50 @@
...
+    static <T1, T2> CombinedTaskResults<T1, T2> create(T1 first, T2 second) {
+      return new AutoValue_TaskUtils_CombinedTaskResults(first, second);
+    }
+  }
+
+  /**
+   * Returns a {@link SuccessContinuation} to be chained off of a {@link Task}, that will run
+   * another task in sequence and combine both results together.
+   *
+   * <p>This is useful when you want to run two tasks and use the results of each, but those tasks
```
rachaprince, line 109: Great docs!

rachaprince (commented):
makes sense so far!

lfkellogg:
Please hold off on reviewing, I think I need to tweak this a bit.

lfkellogg:
OK this is ready for review

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java
```diff
@@ -104,50 +107,68 @@ void onActivityStarted(Activity activity) {
...
-          .addOnSuccessListener(getFidGenerationOnSuccessListener())
           .addOnFailureListener(
-              e -> {
-                LogWrapper.getInstance().e(TAG + "Fid retrieval failed.", e);
-                setSignInTaskCompletionError(
-                    new FirebaseAppDistributionException(
-                        ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE, e));
-              });
+              handleTaskFailure(ErrorMessages.AUTHENTICATION_ERROR, Status.AUTHENTICATION_FAILURE))
+          .onSuccessTask(combineWithResultOf(() -> lifecycleNotifier.getForegroundActivity()))
```
mannyjimenez0810, line 144: Will this be ok in the case where we get the activity but we don't have the lock and once we get the lock the activity changes? 

Foreground activity with no lock-> Lock granted -> Different foreground activity is now there but we continue now that we have the lock.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TaskUtils.java
```diff
@@ -79,14 +87,61 @@
...
+   *
+   * @param secondTaskSource A {@link TaskSource} providing the next task to run
+   * @param <T1> The result type of the first task
+   * @param <T2> The result type of the second task
+   * @return A {@link SuccessContinuation} that will return a new task with result type {@link
+   *     CombinedTaskResults}, combining the results of both tasks
+   */
+  static <T1, T2> SuccessContinuation<T1, CombinedTaskResults<T1, T2>> combineWithResultOf(
+      TaskSource<T2> secondTaskSource) {
+    return firstResult ->
```
mannyjimenez0810, line 139: nice!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java
lfkellogg: Oh great catch! That would break. I'll see what I can do about that.

mannyjimenez0810 (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java
lfkellogg: I made a change in the success listener below, and left some clarifying comments. I think that helps but let me know what you think.

PR #3341, opened 2022-01-21
Host activity interrupted error & dialog dismissal

rachaprince:
This PR makes it so that if the host activity is interrupted and put on pause then we want for it to resume to show a dialog. If there is a new activity that resumes then we cancel the task.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -56,14 +58,18 @@
   @GuardedBy("updateIfNewReleaseTaskLock")
   private UpdateTaskImpl cachedUpdateIfNewReleaseTask;
 
-  private final Object cachedNewReleaseLock = new Object();
-
-  @GuardedBy("cachedNewReleaseLock")
-  private AppDistributionReleaseInternal cachedNewRelease;
+  AtomicReference<AppDistributionReleaseInternal> cachedNewRelease = new AtomicReference();
```
rachaprince, line 61: Since this had a separate lock from the cached task, I thought we could reduce the number of blocks by adding an atomic reference

mannyjimenez0810:
Wanted to get this up before EOD still needs some cleaning before review

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -395,32 +421,42 @@ AppDistributionReleaseInternal getCachedNewRelease() {
...
+    updateConfirmationDialog.setOnCancelListener(
+        dialogInterface -> updateConfirmationDialogClosedCallback(showUpdateDialogTask));
 
-    updateDialog.show();
+    updateConfirmationDialog.show();
 
     return showUpdateDialogTask.getTask();
   }
 
+  private void setRemakeDialogsToFalse() {
```
lfkellogg, line 442: Since we're just using this in one place, and it's already being called from a synchronized block, we might as well just inline it.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -62,11 +62,19 @@
   private AppDistributionReleaseInternal cachedNewRelease;
 
   private Task<AppDistributionRelease> cachedCheckForNewReleaseTask;
-  private AlertDialog updateDialog;
+  private AlertDialog updateConfirmationDialog;
   private AlertDialog signInConfirmationDialog;
-  private boolean remakeUpdateDialog = false;
+  @Nullable private Activity dialogHostActivity = null;
+
+  @GuardedBy("updateIfNewReleaseTaskLock")
```
lfkellogg, line 69: I don't think either of these booleans need to be protected by this lock, or any other lock. They're almost always read or written from the main thread (in lifecycle callbacks), and in the case where they're being set to false in updateIfNewReleaseAvailable, it won't practically affect anything because there's no task in progress at that point.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -322,42 +338,48 @@ public void signOutTester() {
...
-    } else if (remakeUpdateDialog) {
-      remakeUpdateDialog = false;
-      synchronized (cachedNewReleaseLock) {
-        showUpdateAlertDialog(
-            activity, ReleaseUtils.convertToAppDistributionRelease(cachedNewRelease));
-      }
+  @VisibleForTesting
+  void onActivityDestroyed(@NonNull Activity activity) {
+    if (activity == dialogHostActivity) {
+      dialogHostActivity = null;
```
lfkellogg, line 382: On second thought, I think we should add a comment here that this, in part, makes the configuration change case work, since it signals that in onResume we should not bother checking the activity reference.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -393,32 +419,42 @@ AppDistributionReleaseInternal getCachedNewRelease() {
...
+  private void setRemakeDialogsToFalse() {
+    synchronized (updateIfNewReleaseTaskLock) {
+      remakeSignInConfirmationDialog = false;
+      remakeUpdateConfirmationDialog = false;
+    }
+  }
+
+  private void updateConfirmationDialogClosedCallback(TaskCompletionSource showUpdateDialogTask) {
+    synchronized (updateIfNewReleaseTaskLock) {
+      remakeUpdateConfirmationDialog = false;
```
lfkellogg, line 451: We don't need this anymore, right? I'd just drop it and inline this method.

mannyjimenez0810: We don't have this on signIn but it might be nice to have since it does set the remake variable to false when the dialog is cancelled. That may be unnecessary but it makes sense to set it after that since it does not need to be remade at that point.  Curious on your thoughts about that.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
mannyjimenez0810: That is true I agree calling the updateIfNewRelease would not be an issue. I don't see a problem with onPause being called multiple times from different places but if those are all on the main thread there shouldn't be a deadlock issue so as long as we are sure of that we can change it.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
mannyjimenez0810: Will add

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -322,42 +336,48 @@ public void signOutTester() {
...
-    } else if (remakeUpdateDialog) {
-      remakeUpdateDialog = false;
-      synchronized (cachedNewReleaseLock) {
-        showUpdateAlertDialog(
-            activity, ReleaseUtils.convertToAppDistributionRelease(cachedNewRelease));
-      }
+  @VisibleForTesting
+  void onActivityDestroyed(@NonNull Activity activity) {
+    // If the dialogHostActivity is being destroyed it is set to null. This is to ensure that
+    // onResume shows the dialog on a configuration change and not check the activity reference
```
lfkellogg, line 378: "and does not check" ? And please put a `.` on the end of the comment.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -455,4 +479,18 @@ private UpdateTaskImpl getErrorUpdateTask(Exception e) {
...
   }
+
+  private boolean updateIfNewReleaseAvailableIsTaskInProgress() {
+    synchronized (updateIfNewReleaseTaskLock) {
+      return cachedUpdateIfNewReleaseTask != null && !cachedUpdateIfNewReleaseTask.isComplete();
+    }
+  }
+
+  private boolean awaitingSignInDialogConfirmation() {
+    return (updateIfNewReleaseAvailableIsTaskInProgress() && remakeSignInConfirmationDialog);
```
lfkellogg, line 490: Even though it wouldn't make a practical difference, wouldn't it make more sense if this checked if `showSignInDialogTask` was in progress, instead of `cachedUpdateIfNewReleaseTask`?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -455,4 +479,18 @@ private UpdateTaskImpl getErrorUpdateTask(Exception e) {
...
+  }
+
+  private boolean awaitingUpdateDialogConfirmation() {
+    return (updateIfNewReleaseAvailableIsTaskInProgress() && remakeUpdateConfirmationDialog);
```
lfkellogg, line 494: Same here, but `showUpdateDialogTask`

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -480,22 +481,43 @@ public void updateIfNewReleaseAvailable_whenScreenRotates_updateDialogReappears(
...
+    firebaseAppDistribution.onActivityResumed(activity);
 
     assertAlertDialogShown();
     assertFalse(updateTask.isComplete());
   }
 
   @Test
-  public void taskCancelledOnDestroy_whenSignInDialogShowing() {
+  public void
+      updateIfNewReleaseAvailable_signInTaskCancelled_whenSignInDialogShowingAndNewActivityStarts() {
```
lfkellogg, line 492: Nit: for this and the next test, could you change the name of the test to put the "when" clause in the middle of the name, and the expected behavior at the end? E.g. `updateIfNewReleaseAvailable_whenSignInDialogShowingAndNewActivityStarts_signInTaskCancelled`

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -9,7 +10,7 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.AppCompatButton
 import com.google.firebase.appdistribution.AppDistributionRelease
 import com.google.firebase.appdistribution.FirebaseAppDistribution
-import com.google.firebase.appdistribution.FirebaseAppDistributionException
+import com.google.firebase.appdistribution.FirebaseAppDistributionException as FirebaseAppDistributionException1
```
lfkellogg, line 13: Why the `as FirebaseAppDistributionException1`?

lfkellogg (approved):
Some nits, but this looks great!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
mannyjimenez0810: That does make more sense and it shouldn't change anything will change

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
mannyjimenez0810: This was a change rachel made. I wasn't changing anything in testApp and it doesn't impact customers so I decided to leave rachel's wip on the testapp as is.

lfkellogg: Sounds good!

PR #3348, opened 2022-01-25
Allow AAB install to be started from a background thread

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
```diff
@@ -42,29 +43,28 @@
...
+  private final Object updateAabLock = new Object();
+
   @GuardedBy("updateAabLock")
   private UpdateTaskImpl cachedUpdateTask;
 
   @GuardedBy("updateAabLock")
   private AppDistributionReleaseInternal aabReleaseInProgress;
 
-  private final Object updateAabLock = new Object();
-  private final Context context;
```
lfkellogg, line 52: This was unused

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/TestUtils.java
```diff
@@ -22,12 +22,14 @@
...
 
-  static void assertTaskFailure(Task task, Status status, String messageSubstring) {
+  static FirebaseAppDistributionException assertTaskFailure(
+      Task task, Status status, String messageSubstring) {
     assertThat(task.isSuccessful()).isFalse();
     assertThat(task.getException()).isInstanceOf(FirebaseAppDistributionException.class);
     FirebaseAppDistributionException e = (FirebaseAppDistributionException) task.getException();
     assertThat(e.getErrorCode()).isEqualTo(status);
     assertThat(e).hasMessageThat().contains(messageSubstring);
+    return e;
```
mannyjimenez0810, line 32: Why now return a firebase exception?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/AabUpdaterTest.java
```diff
@@ -100,7 +95,7 @@ public void updateAppTask_whenOpenConnectionFails_setsNetworkFailure()
     when(mockHttpsUrlConnectionFactory.openConnection(TEST_URL)).thenThrow(caughtException);
 
     UpdateTask updateTask = aabUpdater.updateAab(TEST_RELEASE_NEWER_AAB_INTERNAL);
-    testExecutor.awaitTermination(1, TimeUnit.SECONDS);
+    testExecutor.awaitTermination(100, TimeUnit.MILLISECONDS);
```
mannyjimenez0810, line 98: Was there some timeout issue that you changed the waiting from 1 second to 100 ms?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TaskUtils.java
```diff
@@ -122,18 +122,35 @@
...
+        secondTaskSource
+            .get()
+            .onSuccessTask(
+                executor,
+                secondResult ->
+                    Tasks.forResult(CombinedTaskResults.create(firstResult, secondResult)));
+  }
+
+  /**
+   * A version of #combineWithResultsOf that runs the listener that combines the results on the main
```
mannyjimenez0810, line 151: nit: Wouldn't this be the given thread since we pass in an executor so it might not be the main thread?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
```diff
@@ -89,7 +93,18 @@ UpdateTaskImpl updateAab(@NonNull AppDistributionReleaseInternal newRelease) {
 
       cachedUpdateTask = new UpdateTaskImpl();
       aabReleaseInProgress = newRelease;
-      redirectToPlayForAabUpdate(newRelease.getDownloadUrl());
+      hasBeenSentToPlayForCurrentTask = false;
+
+      // On a background thread, fetch the redirect URL and open it in the Play app
+      runAsyncInTask(executor, () -> fetchDownloadRedirectUrl(newRelease.getDownloadUrl()))
+          .onSuccessTask(
+              executor,
```
mannyjimenez0810, line 101: Is there a reason for this chain of tasks we use this executor but we don't do this elsewhere?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
```diff
@@ -138,38 +153,30 @@ private static boolean isRedirectResponse(int responseCode) {
...
-        });
+  private void openRedirectUrlInPlay(String redirectUrl, Activity hostActivity) {
+    Intent updateIntent = new Intent(Intent.ACTION_VIEW);
+    updateIntent.setData(Uri.parse(redirectUrl));
+    updateIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+    LogWrapper.getInstance().v(TAG + "Redirecting to play");
+
+    // Launch the intent outside of the synchronized block because we don't want to risk the
+    // activity leaving the foreground while we wait for the lock.
+    hostActivity.startActivity(updateIntent);
```
mannyjimenez0810, line 164: Not saying to move this but wouldn't this get cancelled if the activity leaves the foreground? My understanding is that once we direct them to play if they do not uninstall then install the app all other cases are failure states that should reset the task.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
lfkellogg: Good question. We actually do this inconsistently. For example [ApkUpdater](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java#L101) uses its own executor, while [TesterSignInManager](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java#L145) does not (so callbacks run on the main thread).

In this case since we're already using it for the fetchDownloadRedirectUrl it seemed to make sense to use it everywhere. This also makes testing easier for the reasons we were discussing earlier. Robolectric unit tests run on the main thread, so awaiting Tasks to finish blocks execution of callbacks that also run on the main thread.

I think we should come up with a more consistent approach though.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
lfkellogg: Not sure what you mean, let's chat this out in person.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TaskUtils.java
lfkellogg: This javadoc is for the version below, which does not take an executor.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/AabUpdaterTest.java
lfkellogg: Just to reduce the maximum waiting time of running these tests. 1 second seems like overkill.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/TestUtils.java
lfkellogg: So I can use the returned exception to make assertions about the release (like in `updateAppTask_onAppResume_setsUpdateCancelled`)

mannyjimenez0810: I think this is fine to keep as is. I am not seeing the impact of asserting on the release im aware that was in place before you touched the test, just not sure what the intent with that is. Either way looks good to me.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
lfkellogg: Tried to clarify the comment, let me know what you think!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TaskUtils.java
lfkellogg: Ended up removing this change.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/TestUtils.java
lfkellogg: OK sounds good. I do think there's benefit to asserting that the release was set on the exception. I'll leave it as is for now.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
mannyjimenez0810: nit: very rare -> unlikely or just rare

mannyjimenez0810 (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/AabUpdater.java
lfkellogg: Fair!

PR #3350, opened 2022-01-25
Replace getForegroundActivity with applyToForegroundActivity[Task]

lfkellogg:
This allows the handler to be run _immediately_ after the foreground activity is resumed. When relying on a Task handler, you're actually enqueueing the handler to either the main thread queue or our own executor's queue, and there's no guarantee that the activity is still in the foreground when it's called. While a race condition should be rare, it's better to avoid it if possible.

An alternative would be to pass in a "direct executor" that runs the listener in the same thread that sets the result, but this approach enforces that approach so the caller doesn't have to remember to do that.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
@@ -39,9 +39,6 @@ class FirebaseAppDistributionLifecycleNotifier implements Application.ActivityLi
     void consume(Activity activity);
   }
 
-  /** A {@link SuccessContinuation} that takes an activity and returns a new {@link Task}. */
-  interface ActivityContinuation<T> extends SuccessContinuation<Activity, T> {}
-
   private static FirebaseAppDistributionLifecycleNotifier instance;
   private final Object lock = new Object();
 
@@ -121,7 +118,7 @@ class FirebaseAppDistributionLifecycleNotifier implements Application.ActivityLi
    * Apply a function to a foreground activity, when one is available, returning a {@link Task} that
    * will complete with the result of the Task returned by that function.
    */
-  <T> Task<T> applyToForegroundActivityTask(ActivityContinuation<T> continuation) {
+  <T> Task<T> applyToForegroundActivityTask(SuccessContinuation<Activity, T> continuation) {
     return getForegroundActivity()
         .onSuccessTask(
             // Use direct executor to ensure the consumer is called while Activity is in foreground
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
@@ -20,13 +20,18 @@ import android.os.Bundle;
 import androidx.annotation.GuardedBy;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
-import com.google.android.gms.tasks.Tasks;
 import java.util.ArrayDeque;
 import java.util.Queue;
 
 class FirebaseAppDistributionLifecycleNotifier implements Application.ActivityLifecycleCallbacks {
+
+  /** A functional interface for a function that takes an activity and does something with it. */
+  interface ActivityConsumer {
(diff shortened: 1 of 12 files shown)


firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifierTest.java
```diff
@@ -0,0 +1,99 @@
...
+@RunWith(RobolectricTestRunner.class)
+public class FirebaseAppDistributionLifecycleNotifierTest {
+  private TestActivity activity;
+  private FirebaseAppDistributionLifecycleNotifier lifecycleNotifier;
+
+  static class TestActivity extends Activity {}
+
+  @Before
+  public void setup() {
+    activity = Robolectric.buildActivity(TestActivity.class).create().get();
```
mannyjimenez0810, line 39: Awesome tests!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -20,14 +20,19 @@
...
 import com.google.android.gms.tasks.TaskCompletionSource;
-import com.google.android.gms.tasks.Tasks;
 import java.util.ArrayDeque;
 import java.util.Queue;
 
 class FirebaseAppDistributionLifecycleNotifier implements Application.ActivityLifecycleCallbacks {
 
+  /** A functional interface for a function that takes an activity and does something with it. */
+  interface ActivityConsumer {
+    void consume(Activity activity) throws FirebaseAppDistributionException;
```
mannyjimenez0810, line 33: Looks good! Would like to chat about this consume stuff since I am not familiar with it.

mannyjimenez0810 (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -20,14 +20,28 @@
...
+  /** An {@link Executor} that runs tasks on the current thread. */
+  private static final Executor DIRECT_EXECUTOR = Runnable::run;
+
+  /** A functional interface for a function that takes an activity and does something with it. */
+  interface ActivityConsumer<T> {
+    void consume(Activity activity);
+  }
+
+  /** A {@link SuccessContinuation} that takes an activity and returns a new {@link Task}. */
+  interface ActivityContinuation<T> extends SuccessContinuation<Activity, T> {}
```
lfkellogg, line 43: Leveraging the Task API's SuccessContinuation interface here.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -20,14 +20,28 @@
...
 
 class FirebaseAppDistributionLifecycleNotifier implements Application.ActivityLifecycleCallbacks {
 
+  /** An {@link Executor} that runs tasks on the current thread. */
+  private static final Executor DIRECT_EXECUTOR = Runnable::run;
+
+  /** A functional interface for a function that takes an activity and does something with it. */
+  interface ActivityConsumer<T> {
+    void consume(Activity activity);
+  }
```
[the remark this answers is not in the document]
mrichards: Changing them from `void` to `Void` solves that problem, but if there's no use case for any return values right now, keeping it as `void consume(...)` works for me.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
mrichards: I can't decide if extending `SuccessContinuation` to bind to a specific type makes it cleaner (because there's one less type to worry about) or makes my head hurt (because I have to remember that there's an `is a` relationship here). I'm ok keeping it like this if y'all think it is better. 🙃

mrichards (approved):
There's some tricky APIs in here that are hard to grok (namely `applyToForegroundActivityTask`), but I like that you tried to follow the convention from the Tasks API where possible. Let's give this a shot and see how it goes.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
lfkellogg: Ah, you're right I actually like just keeping it as a SuccessContinuation. We're still enforcing that it accepts an Activity. No need to add another type 👍 

Updated.

PR #3368, opened 2022-01-25
Serve OR queries from the index.

ehsannas:


ehsannas:
the idea is that a DNF transform on a composite filter will result in a condition that looks like
```
condition1 || condition2 || ...
```

where each condition is an **_AND operation of 1 or more FieldFilters_**.

We can then run sub-queries for condition1, condition2, etc, and find the `UNION` of them. In order to simply use the existing code such as TargetIndexMatcher, getArrayValues, getNotInValues, getLowerBound, getUpperBound, etc, we create a new Target object (referred to as "sub-target" in the code) for each sub-query. Each "sub-target" contains only an AND operation of 1 or more FieldFilters (similar to good old queries).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -66,6 +72,12 @@
   private final LocalSerializer serializer;
   private final String uid;
 
+  /**
+   * Maps from a target to its equivalent list of sub-targets. Each sub-target contains only one
+   * term from the target's disjunctive normal form (DNF).
+   */
+  private final Map<Target, List<Target>> targetToDnfSubTargets = new HashMap<>();
```
schmidt-sebastian, line 79: We probably need to find a way to tie this list into our GC algorithm so that we remove the mapping once we remove the target. Probably worth a TODO.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -21,11 +21,14 @@
 import static com.google.firebase.firestore.util.Util.repeatSequence;
 import static java.lang.Math.max;
 
+import android.text.TextUtils;
 import androidx.annotation.Nullable;
+import com.google.common.collect.ObjectArrays;
```
schmidt-sebastian, line 26: We have a goal to kick our Guava dependency, as the Android build doesn't deal well with version conflicts for popular libraries. Hence, we should not directly use Guava and we eventually want to remove it altogether (it is used by GRPC).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -364,41 +450,60 @@ private void deleteIndexEntry(Document document, IndexEntry indexEntry) {
...
+    }
+
+    String queryString =
+        subQueries.size() == 1
+            ?
+            // If there's only one subQuery, just execute the one subQuery.
+            subQueries.get(0)
+            :
+            // Construct "subQuery1 UNION subQuery2 UNION ... LIMIT N"
+            TextUtils.join(" UNION ", subQueries) + " LIMIT " + target.getLimit();
```
schmidt-sebastian, line 504: We need to make sure that we do not pass more than 999 bind elements. The best way to do this probably to validate that the query only contains 100 DNF terms (or what we decide the limit is).

schmidt-sebastian: Can you also add parenthesis to indicate whether the LIMIT belongs to all statements or just the last one?

firebase-firestore/src/main/java/com/google/firebase/firestore/util/LogicUtils.java
```diff
@@ -0,0 +1,54 @@
...
+   *
+   * <p>For instance, for the input: (A || B) && C
+   *
+   * <p>The DNF form is: (A && C) || (B && C)
+   *
+   * <p>The return value is a list with two elements:
+   *
+   * <p>The first element is a composite filter that performs (A && C).
+   *
+   * <p>The second element is a composite filter that performs (B && C).
```
schmidt-sebastian, line 40: Can you reformat to use fewer paragraphs?

firebase-firestore/src/main/java/com/google/firebase/firestore/util/LogicUtils.java
```diff
@@ -0,0 +1,54 @@
...
+   *
+   * @param filter the composite filter to calculate DNF transform for.
+   * @return the terms in the DNF transform.
+   */
+  public static List<Filter> DnfTransform(CompositeFilter filter) {
+    // TODO(orquery): write the DNF transform algorithm here.
```
schmidt-sebastian, line 46: Ah! I was afraid I had to review this now :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
ehsannas: Done

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -74,6 +75,18 @@
...
   Collection<FieldIndex> getFieldIndexes();
 
+  /** Returns whether we can serve the given target from an index. */
+  boolean canServeFromIndex(Target target);
+
+  /**
+   * Iterates over all field indexes that are used to serve the given target, and returns the least
+   * recent offset of them all. Asserts that the target can be served from index.
+   */
+  IndexOffset minOffset(Target target);
```
schmidt-sebastian, line 85: All Java methods should start with a verb: getMinOffset()

You should also align the two comments on their usage of "lowest" and "least recent" (which contrasts min).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -272,6 +285,85 @@ private void updateEntries(
...
+      // If any of the sub-queries cannot be served from the index, the target as a whole cannot be
+      // served from the index.
+      if (getFieldIndex(subTarget) == null) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private IndexOffset getLeastRecentIndexOffset(Collection<FieldIndex> fieldIndexes) {
```
schmidt-sebastian, line 300: This name might still need an update.

schmidt-sebastian (approved):

PR #3374, opened 2022-01-27
Don't deadlock during shutdown()

schmidt-sebastian:
This should fix/improve our flaky Android and Unity CI.


dconeybe (approved):
I've confirmed that this also fixes the flaky failures in the C++ SDK on Android (Googlers can see b/205063697).

PR #3384, opened 2022-01-31
fix ci failures

VinayGuthal:
Adding NO_WRAP and NO_PADDING  to the encoding so that new line character doesn't show up and it doesn't cause e2e tests to fail or smoke tests to fail.


vkryachko (commented):
I does not seem like the commit message adequately describes what is being changed. Can you provide a more detailed explanation?  see go/cl-descriptions

rlazo:
Throughout the code in the repo the combination most commonly used was `Base64.URL_SAFE | Base64.NO_WRAP | Base64.NO_PADDING`, would it be necessary in this case?

rlazo (approved):

PR #3385, opened 2022-01-31
Update `performance-gradle` to `perf-plugin`

jeremyjiang-dev:
I recently updated the folder structure of fireperf plugin in the gradle repo and this needs to be updated.


visumickey (approved):
Thanks for making this change Jeremy. Is this the reason for the recent failure/bugs we received this schedule?

jeremyjiang-dev:
> Thanks for making this change Jeremy. Is this the reason for the recent failure/bugs we received this schedule?

No really. This is for the gradle plugin marker artifact generation where I renamed the directory for the plugin.

leotianlizhan (approved):

PR #3387, opened 2022-02-01
Remove old appdisttro

VinayGuthal:


rlazo (changes requested):
Old appdistro was renamed, no need to delete this.

PR #3397, opened 2022-02-03
Removing sign in tester from checkForNewRelease

mannyjimenez0810:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -252,9 +252,15 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
       return cachedCheckForNewReleaseTask;
     }
+    if (!isTesterSignedIn()) {
+      return Tasks.forException(
+          new FirebaseAppDistributionException(
+              Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE));
+    }
+
     cachedCheckForNewReleaseTask =
-        signInTester()
```
rachaprince, line 256: Could you remove the comment that starts on line 155? With this PR, we no longer need to qualify why we call signIn before calling checkForNewRelease

rachaprince (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
mannyjimenez0810: done

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -252,9 +248,15 @@ public UpdateTask updateIfNewReleaseAvailable() {
       LogWrapper.getInstance().v("Response in progress");
       return cachedCheckForNewReleaseTask;
     }
+    if (!isTesterSignedIn()) {
+      return Tasks.forException(
+          new FirebaseAppDistributionException(
+              Constants.ErrorMessages.AUTHENTICATION_ERROR, AUTHENTICATION_FAILURE));
```
lfkellogg, line 254: I think the status works but I don't think the message is accurate. Could it be something like "Tester not signed in"?

mannyjimenez0810: Is "Failed to authenticate the tester" incorrect? The sentence seems pretty accurate. If the tester is not signed in you cannot authenticate them. This message is also used in getExceptionForHttpResponse in the testerAPiClient and the tester sign in manager. Changing it too "Tester not signed in" could work for those, but it doesn't fit as well as the current message.

lfkellogg: You're right, maybe it's not technically inaccurate, but it's not very helpful. It's true we couldn't authenticate the user, but it's so vague that we're not helping them understand why or resolve the problem.

In the other cases you mention (we get back a 401, or we fail to get a FID) we might not even understand the reasons ourselves, so vague is probably OK. In this case though, we know the reason: the tester wasn't signed in. So I think we should say that in the error message.

mannyjimenez0810: Is your thought to make a new status that is "Tester not signed in" or to change the message that is being shown in the other cases.  In option two the other message is "Failed to authorize the tester". Which I am ok using for the generic cases if we don't make a new message

PR #3399, opened 2022-02-03
Enable global custom attributes on Network Requests

visumickey:


jposuna (approved):
The changes look good. Do you think there's any danger to adding this to network requests? (I.e. why was it not added to network requests from the beginning?)

visumickey:
> 

AFAICT - I don't see any risk in doing this. I spent very less time to debug and make these changes. But I spent a way longer time to understand why this mismatch did happen - and I could not find any clear reason. So, I assume this was an oversight rather than anything else.

jeremyjiang-dev (approved):
The changes LGTM. Have you checked the e2e experience on console? I know this PR is a bit urgent. We need to update docs and api reference later.

jeremyjiang-dev:
Could you also update CHANGELOG?

leotianlizhan (approved):

visumickey:
> Could you also update CHANGELOG?

Done!

PR #3414, opened 2022-02-07
Fix the way of gzipping and base64. This is now consistent with server

VinayGuthal:
Fix the way of gzipping and base64. This is now consistent with server. Tests added to the server side as well.

### DIFF
--- a/firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
+++ b/firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
@@ -102,16 +102,14 @@ public class DefaultHeartBeatController implements HeartBeatController, HeartBea
             JSONObject output = new JSONObject();
             output.put("heartbeats", array);
             output.put("version", "2");
-            try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
-              try (Base64OutputStream b64os =
-                  new Base64OutputStream(
-                      out, Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP)) {
-                try (GZIPOutputStream gzip = new GZIPOutputStream(b64os)) {
-                  gzip.write(output.toString().getBytes("UTF-8"));
-                }
-              }
-              return out.toString("UTF-8");
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+            try (Base64OutputStream b64os =
+                    new Base64OutputStream(
+                        out, Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP);
+                GZIPOutputStream gzip = new GZIPOutputStream(b64os); ) {
+              gzip.write(output.toString().getBytes("UTF-8"));
             }
+            return out.toString("UTF-8");
           }
         });
   }


firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
```diff
@@ -95,19 +96,20 @@
...
             JSONObject output = new JSONObject();
             output.put("heartbeats", array);
             output.put("version", "2");
             ByteArrayOutputStream out = new ByteArrayOutputStream();
-            GZIPOutputStream gzip = new GZIPOutputStream(out);
-            gzip.write(output.toString().getBytes());
+            Base64OutputStream b64os =
+                new Base64OutputStream(out, Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP);
+            GZIPOutputStream gzip = new GZIPOutputStream(b64os);
+            gzip.write(output.toString().getBytes("UTF-8"));
```
vkryachko, line 109: This whole block can be more readable and conventional by using try with resources, something like:

```java
try(ByteArrayOutputStream out = new ByteArrayOutputStream();
    Base64OutputStream(out, Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP);
    GZIPOutputStream gzip = new GZIPOutputStream(b64os)) {
  gzip.write(output.toString().getBytes("UTF-8"));
  return out.toString("UTF-8");
}
```

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatControllerTest.java
```diff
@@ -115,13 +116,17 @@ public void generateHeartBeat_oneHeartBeat()
     heartBeatController
         .getHeartBeatsHeader()
         .addOnCompleteListener(executor, getOnCompleteListener);
-    String expected =
-        Base64.getUrlEncoder()
-            .withoutPadding()
-            .encodeToString(
-                compress(
-                        "{\"heartbeats\":[{\"agent\":\"test-agent\",\"date\":\"[2015-02-03]\"}],\"version\":\"2\"}")
```
vkryachko, line 123: I think it was more readable with a json literal, can we keep it(fixing the date list bug though)

VinayGuthal: Unable to do it bcz of the json array. Even though i assign the value of output.string() it still doesn't work.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
```diff
@@ -95,19 +96,21 @@
...
-            GZIPOutputStream gzip = new GZIPOutputStream(out);
-            gzip.write(output.toString().getBytes());
-            gzip.close();
-            return Base64.encodeToString(
-                out.toString("UTF-8").getBytes(),
-                Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP);
+            try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+              Base64OutputStream b64os =
+                  new Base64OutputStream(out, Base64.URL_SAFE | Base64.NO_PADDING | Base64.NO_WRAP);
+              GZIPOutputStream gzip = new GZIPOutputStream(b64os);
```
vkryachko, line 108: both of these should also be inside the try parentheses.

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatControllerTest.java
```diff
@@ -202,27 +194,32 @@ public void generateHeartBeat_twoHeartBeatsSameUserAgent()
...
+  }
+
+  private static byte[] gzip(String input) {
+    ByteArrayOutputStream byteOutputStream = new ByteArrayOutputStream();
+    try {
+      try (GZIPOutputStream gzipOutputStream = new GZIPOutputStream(byteOutputStream)) {
+        gzipOutputStream.write(input.getBytes(UTF_8));
+      }
+      byte[] gzipped = byteOutputStream.toByteArray();
+      byteOutputStream.close();
```
vkryachko, line 214: Looks like you can use try with resources here as well

VinayGuthal: it complains saying byte[] is not a proper type for try with resources.

rlazo: AFAIK not closing it is OK.

https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html#close()

> Closing a ByteArrayOutputStream has no effect. The methods in this class can be called after the stream has been closed without generating an IOException.

PR #3415, opened 2022-02-07
Adding null checks for instantiating InstrHttpInputStream/InstrHttpOutputStream

### LINKED ISSUE #3406: Firebase Performance Plugin Null Pointer Exception While Measuring Network Performance

other categories use one of these other channels:

    with the firebase tag.
    google group.

  * Android Studio version: Android Studio Dolphin | 2021.3.1 Canary 1
  * Firebase Component: Performance 
  * Component version: 'com.google.firebase:perf-plugin:1.4.1' &&  ('com.google.firebase:firebase-bom:29.0.4')

#### Steps to reproduce: The problem is not reproducible on our end

image

While the 3rd party SDK's relevant code is null-safe, 

com.google.firebase.perf.network.InstrHttpInputStream.close (InstrHttpInputStream.java:67) this line is not. We assess while trying to measure network performance inputStream object closes before measuring.

#### Relevant Code:

Their code's relevant part : 

image

// TODO(you): The bug can not be reproducible on our end 

image

raymondlam:
This should remove the possibility of a null being passed to InstrHttpInputStream/InstrHttpOutputStream.

Fixes #3406

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/network/InstrURLConnectionBase.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/network/InstrURLConnectionBase.java
@@ -140,7 +140,13 @@ class InstrURLConnectionBase {
 
     try {
       final InputStream inputStream = httpUrlConnection.getInputStream();
-      return new InstrHttpInputStream(inputStream, networkMetricBuilder, timer);
+      // Make sure we don't pass in a null into InstrHttpInputStream, since InstrHttpInputStream is
+      // not null-safe.
+      if (inputStream != null) {
+        return new InstrHttpInputStream(inputStream, networkMetricBuilder, timer);
+      }
+      // Only reached when inputStream is null
+      return inputStream;
     } catch (final IOException e) {
       networkMetricBuilder.setTimeToResponseCompletedMicros(timer.getDurationMicros());
       NetworkRequestMetricBuilderUtil.logError(networkMetricBuilder);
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/network/InstrURLConnectionBase.java
```diff
@@ -156,8 +161,13 @@ public long getLastModified() {
...
     try {
-      return new InstrHttpOutputStream(
-          httpUrlConnection.getOutputStream(), networkMetricBuilder, timer);
+      final OutputStream outputStream = httpUrlConnection.getOutputStream();
+      // Make sure we don't pass in a null into InstrHttpOutputStream, since InstrHttpOutputStream
+      // is not null-safe.
+      if (outputStream != null) {
+        return new InstrHttpOutputStream(outputStream, networkMetricBuilder, timer);
+      }
+      return null;
```
jeremyjiang-dev, line 170: Is the caller of `getOutputStream()` okay to deal with null?

raymondlam: This question would apply to both input and output stream.

I suspect that it depends on whether the developers thinks a null may occur with getInputStream (but since the docs don't mention it can be null, devs may not add a check).

However, we use this class to wrap the original url connect and call the underlying object as a delegate.  We shouldn't be returning our instrumented input stream when the original url connection would have returned a null.  If they did not have Firebase Performance, they would have needed to handle the null case anyways.

raymondlam: I could also do `return outputStream` instead of `return null` to showcase that we are not changing the original behavior.

jeremyjiang-dev (approved):

PR #3417, opened 2022-02-07
Add Offset logline

schmidt-sebastian:
This shows how far the index has advanced.


firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexBackfiller.java
```diff
@@ -139,6 +139,7 @@ private int writeEntriesForCollectionGroup(
     indexManager.updateIndexEntries(nextBatch.getDocuments());
 
     IndexOffset newOffset = getNewOffset(existingOffset, nextBatch);
+    Logger.debug(LOG_TAG, "Updating offset: %s", newOffset);
```
ehsannas, line 142: Should be more informative? e.g. add the collection group name in the log message?

ehsannas (approved):

PR #3420, opened 2022-02-08
Parse Overlays in background

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/util/BackgroundQueue.java
```diff
@@ -40,8 +42,12 @@ public void execute(Runnable task) {
...
   /** Wait for all currently scheduled tasks to complete. */
-  public void drain() throws InterruptedException {
-    completedTasks.acquire(pendingTaskCount);
-    pendingTaskCount = 0;
+  public void drain() {
+    try {
+      completedTasks.acquire(pendingTaskCount);
+      pendingTaskCount = 0;
+    } catch (InterruptedException e) {
+      fail("Interrupted while waiting for background task", e);
```
dconeybe, line 50: It's not a good idea to swallow `InterruptedException`. You should at least call `Thread.currentThread().interrupt()` before calling `fail()`.

https://www.baeldung.com/java-interrupted-exception

dconeybe (changes requested):
I did a quick review and had a couple of immediate suggestions. I'll need to sit on it and review it properly tomorrow.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/BackgroundQueue.java
schmidt-sebastian: Added Thread.currentThread().interrupt() (also recommended by other literature)

firebase-firestore/src/test/java/com/google/firebase/firestore/local/DocumentOverlayCacheTestCase.java
```diff
@@ -86,6 +90,13 @@ public void testReturnsNullWhenOverlayIsNotFound() {
     assertNull(cache.getOverlay(key("coll/doc1")));
   }
 
+  @Test
```
dconeybe, line 93: Add a test for calling `getOverlays()` with an empty set.

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/DocumentOverlayCache.java
```diff
@@ -38,6 +39,12 @@
   @Nullable
   Overlay getOverlay(DocumentKey key);
 
+  /**
+   * Gets the saved overlay mutation for the given document keys. Skips keys for which there are no
+   * overlays.
+   */
+  Map<DocumentKey, Overlay> getOverlays(SortedSet<DocumentKey> keys);
```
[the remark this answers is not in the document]
schmidt-sebastian: As discussed offline, I added an assert that ensures that the SortedSet does not have a custom comparator. This avoids copying the keys and post-processing them in this method.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java
```diff
@@ -47,7 +55,49 @@ public Overlay getOverlay(DocumentKey key) {
...
+
+  @Override
+  public Map<DocumentKey, Overlay> getOverlays(SortedSet<DocumentKey> keys) {
+    Map<DocumentKey, Overlay> result = new HashMap<>();
+
+    BackgroundQueue backgroundQueue = new BackgroundQueue();
+    ResourcePath currentCollectionPath = ResourcePath.EMPTY;
+    List<Object> currentDocumentIds = new ArrayList<>();
+    for (DocumentKey key : keys) {
+      if (!currentCollectionPath.equals(key.getCollectionPath())) {
```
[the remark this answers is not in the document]
schmidt-sebastian: First GItHub duplicated my response, and then it ate it:

I added an early exit to `processSingleCollection()` to keep the callsite clean. This avoids extra memory allocations - but note that even without it, no SQL query would have been made.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java
```diff
@@ -47,7 +56,54 @@ public Overlay getOverlay(DocumentKey key) {
...
+    Map<DocumentKey, Overlay> result = new HashMap<>();
+
+    BackgroundQueue backgroundQueue = new BackgroundQueue();
+    ResourcePath currentCollection = ResourcePath.EMPTY;
+    List<Object> accumulatedDocumentIds = new ArrayList<>();
+    for (DocumentKey key : keys) {
+      if (!currentCollection.equals(key.getCollectionPath())) {
+        processSingleCollection(result, backgroundQueue, currentCollection, accumulatedDocumentIds);
+        currentCollection = key.getCollectionPath();
+        accumulatedDocumentIds.clear();
```
dconeybe, line 74: I think you need to assign `accumulatedDocumentIds` to a new object rather than clearing it. This list is specified to `SQLitePersistence.LongQuery` which holds an iterator to its contents. And modifying the list underneath a live iterator is undefined.

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteDocumentOverlayCache.java
schmidt-sebastian: That is a quality code review right here. Thanks for catching this.

schmidt-sebastian: Actually - I think it might have been fine since all processing happens synchronously. But it is more clear with the suggested fix.

PR #3431, opened 2022-02-10
Rc realtime prototype

qdpham13:
Prototype for Realtime RC
Changes include:
- OkHTTP dependencies
- Adding an HTTP/2 client that listens for pushes asynchronously using SSE
- Autofetches on every notification push
- Adds callback interface to use for every notifications
- Automatic HTTP connection management: close in background, open in foreground
- Retry HTTP connection


firebase-config/firebase-config.gradle
```diff
@@ -58,6 +58,9 @@ dependencies {
     implementation 'com.google.firebase:firebase-measurement-connector:18.0.0'
     implementation 'com.google.android.gms:play-services-tasks:17.0.2'
 
+    implementation 'com.squareup.okhttp3:okhttp:4.9.0'
```
vkryachko, line 61: This version is written in kotlin so you might want to downgrade to something like 3.14

firebase-config/firebase-config.gradle
```diff
@@ -58,6 +58,9 @@ dependencies {
...
+    implementation 'com.squareup.okhttp3:okhttp-sse:4.9.0'
```
vkryachko, line 62: Any concerns with this library being marked as experimental in https://github.com/square/okhttp/tree/master/okhttp-sse

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -661,4 +671,51 @@ private static boolean isFetchedFresh(
...
+  }
+
+  /**
+   * Stops automatic realtime handling.
+   * */
+  public void stopAutomaticStreamHandling() {
+    this.automaticRealtimeHandling = false;
+  }
+
+  public void handleAutomaticRealtime(boolean background) {
```
vkryachko, line 710: Why do we care about background? wouldn't it be better to keep the connection open for as long as someone is listening, effectively delegating control of this to the developer?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHTTPClient.java
```diff
@@ -0,0 +1,233 @@
...
+
+        // HTTP Client
+        ConnectionPool connectionPool = new ConnectionPool();
+        this.okHttpClient = new OkHttpClient.Builder()
+                .readTimeout(2, TimeUnit.HOURS)
+                .writeTimeout(2, TimeUnit.HOURS)
+                .connectTimeout(2, TimeUnit.HOURS)
+                .followRedirects(false)
+                .protocols(Collections.singletonList(Protocol.H2_PRIOR_KNOWLEDGE))
+                .retryOnConnectionFailure(true)
```
vkryachko, line 93: Do we know the semantics of this retry functionality when used in conjunction with okhttp-sse? also, given the 2 hour timeouts above, will we retry after that long?

vkryachko (commented):
cc @schmidt-sebastian would appreciate your feedback on this PR

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
schmidt-sebastian: You probably want to re-kick the connection when the app goes in the foreground. But I agree that we should not be tearing down the connection automatically.

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -299,4 +315,26 @@ private static boolean isAbtSupported(FirebaseApp firebaseApp, String namespace)
...
+
+  private static class GlobalBackgroundListener implements BackgroundDetector.BackgroundStateChangeListener {
+    private static AtomicReference<GlobalBackgroundListener> INSTANCE = new AtomicReference<>();
+
+    private static void ensureBackgroundListenerIsRegistered(Context context) {
+      Application application = (Application) context.getApplicationContext();
+      if (INSTANCE.get() == null) {
+        GlobalBackgroundListener globalBackgroundListener = new GlobalBackgroundListener();
+        if (INSTANCE.compareAndSet(null, globalBackgroundListener)) {
+          BackgroundDetector.initialize(application);
```
schmidt-sebastian, line 327: How do we stop the listener? Will there be an API to add/remove the listener?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHTTPClient.java
```diff
@@ -0,0 +1,233 @@
...
+    private final ConfigFetchHandler configFetchHandler;
+    private final FirebaseApp firebaseApp;
+    private final FirebaseInstallationsApi firebaseInstallations;
+    private final Executor executor;
+    private final Context context;
+
+    // Retry parameters
+    private final Timer timer;
+    private final int ORIGINAL_RETRIES = 7;
+    private final long RETRY_TIME = 10000;
```
schmidt-sebastian, line 57: You should include the unit that you are using in the name.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHTTPClient.java
schmidt-sebastian: We need to ensure that we use backoff for these retries.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHTTPClient.java
```diff
@@ -0,0 +1,233 @@
...
+
+    private void setCommonRequestHeaders(Request.Builder request) {
+        request.addHeader(API_KEY_HEADER, this.firebaseApp.getOptions().getApiKey());
+
+        // Headers required for Android API Key Restrictions.
+        request.addHeader(X_ANDROID_PACKAGE_HEADER, context.getPackageName());
+        request.addHeader(X_ANDROID_CERT_HEADER, getFingerprintHashForPackage());
+
+        // Header to denote request is retryable on the server.
+        request.addHeader(X_GOOGLE_GFE_CAN_RETRY, "yes");
```
schmidt-sebastian, line 106: This is missing a lot of Firebase headers (GMPID and others)

firebase-config/firebase-config.gradle
qdpham13: I was actually thinking given the overall size increase from the library (+127%) that we don't try to use HTTP/2 and just use 1.1 so that we can just use the JDK provided HttpUrlConnection class to make a connection. Then we would just use chunked encoding to simulate streaming from server to client. The PR with this idea is here: https://github.com/firebase/firebase-android-sdk/pull/3451

firebase-config/firebase-config.gradle
qdpham13: I was planning on using in the mean time. We could implement our own SSE functionality without much difficulty I would presume since it is pretty straightforward. But like in the previous comment it may be better for us to just use HTTP 1.1 with chunked encoding.

firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
qdpham13: I can't find a way to remove a listener from the BackgroundDetector but I added a flag in FirebaseRemoteConfig that can enable foreground rekicking or disable it.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHTTPClient.java
qdpham13: I was planning on relying more on the retry function that I made that also includes random backoffs, configurable amount of retries, and async scheduling.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHTTPClient.java
qdpham13: Added seconds to param.

PR #3328, opened 2022-01-19
Re-establish streams when App Check token expires.

ehsannas:

PR #3448, opened 2022-02-15
Avoid re-establishing streams for the same app check token.

ehsannas:


ehsannas:
This change should have been done as part of https://github.com/firebase/firebase-android-sdk/pull/3328.

PR #3453, opened 2022-02-16
WIP - Ensure that dialog is shown on UI thread

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -35,7 +37,7 @@
   private static final Executor DIRECT_EXECUTOR = Runnable::run;
 
   /** A functional interface for a function that takes an activity and does something with it. */
-  interface ActivityConsumer<T> {
+  interface ActivityConsumer {
```
lfkellogg, line 40: Noticed that we'd left an unused type param here.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -98,38 +100,50 @@ static synchronized FirebaseAppDistributionLifecycleNotifier getInstance() {
...
-                return Tasks.forResult(null);
-              } catch (Throwable t) {
-                return Tasks.forException(FirebaseAppDistributionException.wrap(t));
-              }
-            });
+    return applyToForegroundActivityTask(
+        activity -> {
+          consumer.consume(activity);
+          return Tasks.forResult(null);
+        });
```
lfkellogg, line 111: I realized that this could just call `applyToForegroundActivityTask`

rachaprince (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -98,30 +98,30 @@ static synchronized FirebaseAppDistributionLifecycleNotifier getInstance() {
...
+   * <p>The continuation function will always be called on the main thread.
    */
   <T> Task<T> applyToForegroundActivityTask(SuccessContinuation<Activity, T> continuation) {
     return getForegroundActivity()
         .onSuccessTask(
-            // Use direct executor to ensure the consumer is called while Activity is in foreground
+            // Use direct executor to ensure the consumer is called while Activity is in
+            // foreground. Also, since the Task returned by getForegroundActivity is always
+            // completed on the main thread, the onSuccessTask callback will also be called on the
+            // main thread.
```
rachaprince, line 124: nice comment

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -455,6 +459,21 @@ public void updateToNewRelease_receiveProgressUpdateFromUpdateApp() {
...
+    when(mockSignInStorage.getSignInStatus()).thenReturn(false);
+    when(activity.isChangingConfigurations()).thenReturn(true);
+
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    Future<UpdateTask> future =
+        executorService.submit(() -> firebaseAppDistribution.updateIfNewReleaseAvailable());
+    TestUtils.awaitAsyncOperations(executorService);
+
+    assertAlertDialogShown();
+    assertFalse(((UpdateTask) future.get()).isComplete());
```
rachaprince, line 474: After showing the dialog, does the next task in updateIfNewReleaseAvail.  execute on the original calling thread? Or do all subsequent tasks in the chain execute on the main thread?

lfkellogg: Good question. Each subsequent callback chained off using onSuccessTask or addOnSuccessListener will execute on whichever executor specified when adding the callback. And since we don't usually pass in any executor, the default is to enqueue them to the main thread executor.

For example here: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java#L170

signInTester will be called on the main thread. To run on a different thread you'd have to pass in an executor, like `.onSuccessTask(executor, unused -> signInTester())`.

mannyjimenez0810: So if they call updateIfNewReleaseAvailable in a background thread we move onto the main thread and continue on the main thread. Since our first work is to show a dialog its as if they didn't even call it from a background thread?

lfkellogg: Kind of. If they call it from the main thread it will show the dialog immediately. If they call it from a background thread, we enqueue a Runnable onto the main thread queue, and that Runnable shows the dialog ([runOnUiThread](https://developer.android.com/reference/android/app/Activity#runOnUiThread(java.lang.Runnable)) does this for us). So we won't interrupt anything that's currently running on the main thread, but it should happen very quickly since things running on the main thread should be fast.

[This is all assuming the case where the app is in the foreground.]

rachaprince: This is very interesting behavior. I could see scenario in which the developer calls updateIfNewReleaseAvail from a background thread and expects the following to happen:

1. queue updateIfNewReleaseAvail. (on bkg thread)
2. showDialog() (on the main thread)
3. checkForNewRelease (on bkg thread)
4. showUpdateDialog() (on main thread)
5. downloadRelease (on bkg thread)

But it sounds like that behavior is not supported via the basic config

lfkellogg: Sorry, to be clear the actual long running work does happen on background threads managed by our SDK:

- [Fetch new release](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/NewReleaseFetcher.java#L111-L112)
- [Download APK](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java#L149-L156)

I was specifically referring to the code executed within the callbacks attached to the tasks. Those should execute very fast because typically they're just creating and returning some new task without waiting for the actual work to finish. So for most of them it doesn't really matter what thread they run on.

lfkellogg:
I'd like to close this in favor of a more targeted approach: https://github.com/firebase/firebase-android-sdk/pull/3453

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
rachaprince: Oh I understand now. Thank you for clarifying!

PR #3454, opened 2022-02-16
Support > and != on the same field

schmidt-sebastian:
This adds the ability to use inequality and equality filters on the same field.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
@@ -14,8 +14,6 @@
 
 package com.google.firebase.firestore.core;
 
-import static com.google.firebase.firestore.model.Values.MAX_VALUE;
-import static com.google.firebase.firestore.model.Values.MIN_VALUE;
 import static com.google.firebase.firestore.model.Values.max;
 import static com.google.firebase.firestore.model.Values.min;
 
@@ -25,10 +23,11 @@ import com.google.firebase.firestore.model.FieldIndex;
 import com.google.firebase.firestore.model.FieldPath;
 import com.google.firebase.firestore.model.ResourcePath;
 import com.google.firebase.firestore.model.Values;
-import com.google.firestore.v1.ArrayValue;
 import com.google.firestore.v1.Value;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 
 /**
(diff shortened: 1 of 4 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -153,23 +152,21 @@ public boolean hasLimit() {
    * Returns the list of values that are used in != or NOT_IN filters. Returns {@code null} if there
    * are no such filters.
    */
-  public @Nullable List<Value> getNotInValues(FieldIndex fieldIndex) {
-    List<Value> values = new ArrayList<>();
-
+  public @Nullable Collection<Value> getNotInValues(FieldIndex fieldIndex) {
```
schmidt-sebastian, line 155: This method now supports a == 1 && a != 2, which now correctly just returns a != 2 instead of the concatenation of both.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -211,13 +208,8 @@ public Bound getLowerBound(FieldIndex fieldIndex) {
...
             break;
           case NOT_EQUAL:
-            filterValue = Values.MIN_VALUE;
-            break;
           case NOT_IN:
-            filterValue =
-                Value.newBuilder()
-                    .setArrayValue(ArrayValue.newBuilder().addValues(MIN_VALUE))
-                    .build();
+            filterValue = Values.MIN_VALUE;
```
schmidt-sebastian, line 212: Not In and Not Equal use the same bound, which makes the comparison logic at the end of this method work as we do not want to compare to an array but to an actual minimum.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -740,8 +740,10 @@ private boolean isInFilter(Target target, FieldPath fieldPath) {
     for (Filter filter : target.getFilters()) {
       if ((filter instanceof FieldFilter) && ((FieldFilter) filter).getField().equals(fieldPath)) {
         FieldFilter.Operator operator = ((FieldFilter) filter).getOperator();
-        return operator.equals(FieldFilter.Operator.IN)
-            || operator.equals(FieldFilter.Operator.NOT_IN);
+        if (operator.equals(FieldFilter.Operator.IN)
```
schmidt-sebastian, line 743: A not-in filter does not have to appear on the first filter for a field. If there are multiple filters for the field path, not in can appear later.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/Values.java
```diff
@@ -41,16 +41,15 @@
   public static final Value NULL_VALUE =
       Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build();
   public static final Value MIN_VALUE = NULL_VALUE;
+  private static final Value MAX_VALUE_TYPE = Value.newBuilder().setStringValue("__max__").build();
```
schmidt-sebastian, line 44: All changes in this file are to support correct ordering for MAX_VALUE.

wu-hui (approved):
Thanks for the comments.

Issue #3459, opened 2022-02-17
Crashlytics report could not be enqueued to DataTransport after library upgrade


danyalberton:
### [READ] Step 1: Are you in the right place?

Yes

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: Bumblebee 2021.1.1
  * Firebase Component: Bom 29.1.0 (crashlytics, crashlytics-ndk, messaging, analytics)
  * Component version: as specified in BoM 29.1.0

#### Issue:
Firebase Crashlytics fails to send any crash, because it crashes internally. It happens after upgrading the library version and it does not fix by themself.

#### Steps to reproduce:

1. Have the library configuration as follow:

```
implementation platform('com.google.firebase:firebase-bom:26.1.1')
implementation 'com.google.firebase:firebase-messaging'
implementation 'com.google.firebase:firebase-analytics'
implementation 'com.google.firebase:firebase-crashlytics'
implementation 'com.google.firebase:firebase-crashlytics-ndk'
```

2. Upgrade to the BoM format as follow:

```
implementation platform('com.google.firebase:firebase-bom:29.1.0')
implementation 'com.google.firebase:firebase-messaging'
implementation 'com.google.firebase:firebase-analytics'
implementation 'com.google.firebase:firebase-crashlytics'
implementation 'com.google.firebase:firebase-crashlytics-ndk'
```

3. Crash the app using random fatal exception.

#### Error log:

The following warning is shown on logcat

```
2022-02-17 11:13:47.703 21697-21877/com.dert.kindertap D/libcrashlytics: Initializing libcrashlytics version 3.2.0
2022-02-17 11:13:47.705 21697-21877/com.dert.kindertap D/libcrashlytics: Initializing native crash handling successful.
2022-02-17 11:13:47.755 21697-21877/com.dert.kindertap W/FirebaseCrashlytics: Crashlytics report could not be enqueued to DataTransport
    android.database.sqlite.SQLiteException: table global_log_event_state already exists (code 1 SQLITE_ERROR): , while compiling: CREATE TABLE global_log_event_state (last_metrics_upload_ms BIGINT PRIMARY KEY)
        at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
        at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:986)
        at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:593)
[...]

PR #3460, opened 2022-02-17
call DROP_GLOBAL_LOG_EVENT_STATE_SQL before inserting during migration

argzdev:
Proposed fix for #3459 by [danyalberton](https://github.com/danyalberton).

The DROP_GLOBAL_LOG_EVENT_STATE_SQL is only called during onDowngrade, which means if somehow the migration fails half way, there is no way of continuing the migration, thus it's best to call DROP_GLOBAL_LOG_EVENT_STATE_SQL whenever creating the table to ensure this table does not exist before upgrade.


danyalberton (approved):
I approve this changes that resolve a problem during library update.

vkryachko (approved):

PR #3464, opened 2022-02-17
Show dialogs on UI thread

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -215,34 +215,40 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
-            showSignInDialogTask.setException(
-                new FirebaseAppDistributionException(
-                    ErrorMessages.AUTHENTICATION_CANCELED, AUTHENTICATION_CANCELED)));
-
-    signInConfirmationDialog.show();
-
+    // We may not be on the main (UI) thread in some cases, specifically if the developer calls
+    // the basic config from the background. If we are already on the main thread, this will
+    // execute immediately.
+    hostActivity.runOnUiThread(
```
mannyjimenez0810, line 223: Can we do the work of creating the dialog not on the UI thread? Can we just have the showing of the dialog on the UI thread?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
```diff
@@ -247,6 +251,19 @@ public void updateToNewRelease_whenNewAabReleaseAvailable_showsUpdateDialog() {
         shadowOf(dialog).getMessage().toString());
   }
 
+  @Test
+  public void updateToNewRelease_fromABackgroundThread_showsUpdateDialog()
```
rachaprince, line 255: (Nit/ No action required) I'm just now noticing that some of these tests use the old name of `updateToNewRelease`, and some use the new name `updateIfNewReleaseAvailable` X)

rachaprince (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
lfkellogg: Good question - we actually can't. `new AlertDialog.Builder(hostActivity).create()` has to be called on the UI thread. I at first just wrapped the `.show()` call and then had to figure out why it was still broken :)

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/FirebaseAppDistributionTest.java
lfkellogg: Ah, good catch. Fixed!

mannyjimenez0810 (approved):

PR #3476, opened 2022-02-22
Make indexing public

schmidt-sebastian:


firebase-common/src/main/java/com/google/firebase/annotations/PreviewApi.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that this object (class, method, etc) is experimental and that both its signature and
+ * implementation are subject to change. An API marked with this annotation provides no guarantee
+ * of API stability or backward-compatibility.
+ */
+@KeepForSdk
+@Target({ElementType.TYPE, ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR})
+public @interface PreviewApi {}
```
vkryachko, line 29: We have a `firebase-annotations` sdk, would it make sense to move it there?

schmidt-sebastian: I can do this, but I would have to add a dependency on `com.google.android.gms:play-services-basement` to use `@KeepForSdk`.

vkryachko: Why do you need `@KeepForSdk`? it has no effect in github

schmidt-sebastian: Okay, removed "KeepForSdk" and moved the class to firebase-annotations. PTAL.

9996142202 (approved):

PR #3485, opened 2022-02-25
`Query.get` should wait for backend connection

IanWyszynski:
Addresses [b/221106730](http://b/221106730). Existing behavior is to throw a "Client is offline" exception when no local value is available. The reason we did this was so that we could fallback to disk cache when the server is unresponsive. 

However, if there is no local persisted value available, which could be the case when an app is just starting up, get will just give up. For consistency with our other operations, get should await a database connection in this case.

Since we want to "try" the server when a persisted value is available, and "await" the server when a persisted value is not available, we have to check for the persisted value before sending any requests to the server.

### DIFF
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
 - [fixed] Fixed a crash that prevented the RTDB SDK from reconnecting to the
   backend if a token refresh attempt was unsuccesful.
+- [fixed] `Query.get` no longer throws "Client is offline" exception when local
+  value is not available, waits for connection instead.
 
 # 20.0.2
 - [fixed] The SDK can now continue to issue writes for apps that send an
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -2,7 +2,7 @@
 - [fixed] Fixed a crash that prevented the RTDB SDK from reconnecting to the
   backend if a token refresh attempt was unsuccesful.
 - [fixed] `Query.get` no longer throws "Client is offline" exception when local
-  value is not available, waits for connection instead.
+  value is not available. Instead, it waits for a backend connection.
 
 # 20.0.2
 - [fixed] The SDK can now continue to issue writes for apps that send an
--- a/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
+++ b/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.database;
 
+import static android.os.AsyncTask.THREAD_POOL_EXECUTOR;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -4556,22 +4557,34 @@ public class QueryTest {
   }
 
   @Test
-  public void testGetThrowsExceptionForEmptyNodeWhenOffline()
-      throws DatabaseException, InterruptedException {
+  public void testGetWaitsForConnection() throws DatabaseException, InterruptedException {
     FirebaseApp app =
         appForDatabaseUrl(IntegrationTestValues.getNamespace(), UUID.randomUUID().toString());
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
     DatabaseReference node = db.getReference().child(db.getReference().push().getKey());
+    long offlinePause = 3000L;
     db.goOffline();
+    THREAD_POOL_EXECUTOR.execute(
+        () -> {
+          try {
+            Thread.sleep(offlinePause);
+          } catch (InterruptedException e) {
+            fail("Exception while pausing for get.");
+          } finally {
+            db.goOnline();
+          }
+        });
     try {
+      long startTime = System.currentTimeMillis();
       Tasks.await(node.get());
+      long elapsed = System.currentTimeMillis() - startTime;
+      // Since get starts after we schedule the re-connect task, it could technically
+      // block for slightly less long than offlinePause, but the difference should be
+      // negligible.
+      assertTrue(elapsed > offlinePause * 0.75);
     } catch (ExecutionException e) {
-      assertEquals(e.getCause().getMessage(), "Client is offline");
-      return;
-    } finally {
-      db.goOnline();
+      fail("get threw an exception: " + e);
     }
-    fail("Client get succeeded even though offline.");
   }
 
   @Test
--- a/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
+++ b/firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
@@ -40,6 +40,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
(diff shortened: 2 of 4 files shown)


firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -428,22 +428,6 @@ public void listen(
             });
     outstandingGets.put(readId, outstandingGet);
 
-    if (!connected()) {
```
IanWyszynski, line 431: Getting rid of this means once we issue a get, it will always be completed. The `Repo` decides whether to use, or discard, the result.

puf:
> For consistency with our other operations, get should await a database connection in this case.

Note that my report is specifically about thew situation during app startup, when I think we should wait for the connection status to stabilize before determining failure. I'm not certain if that applies to other cases too.

In general the `get()` API should follow the connectivity behavior of `once` (when `once` cannot return a locally cached value), since that is the behavior devs are used to.

firebase-database/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 - [fixed] Fixed a crash that prevented the RTDB SDK from reconnecting to the
   backend if a token refresh attempt was unsuccesful.
+- [fixed] `Query.get` no longer throws "Client is offline" exception when local
+  value is not available, waits for connection instead.
```
schmidt-sebastian, line 5:
```suggestion
  value is not available. Instead, it waits for a backend connection.
```

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -4556,22 +4557,34 @@ public void testGetReturnsNullForEmptyNodeWhenOnline()
...
+          }
+        });
     try {
+      long startTime = System.currentTimeMillis();
       Tasks.await(node.get());
+      long elapsed = System.currentTimeMillis() - startTime;
+      // Since get starts after we schedule the re-connect task, it could technically
+      // block for slightly less long than offlinePause, but the difference should be
+      // negligible.
+      assertTrue(elapsed > offlinePause * 0.75);
```
schmidt-sebastian, line 4584: This seems like it has the potential to cause flakes and does not actually add a lot of value. How about doing something like:


```
AtomicBoolean receivedValue = new AtomicBoolean();
  THREAD_POOL_EXECUTOR.execute(
        () -> {
          try {
            Thread.sleep(offlinePause);
            fail("Exception while pausing for get.");
          } finally {
            assertFalse(receivedValue.get());
            db.goOnline();
          }
        });
    try {
      Tasks.await(node.get());
      receivedValue.set(true);
```

firebase-database/src/main/java/com/google/firebase/database/connection/PersistentConnectionImpl.java
```diff
@@ -428,22 +428,6 @@ public void listen(
...
-            if (!outstandingGet.markSent()) {
-              return;
-            }
-            if (logger.logsDebug()) {
-              logger.debug("get " + readId + " timed out waiting for connection");
-            }
-            outstandingGets.remove(readId);
-            source.setException(new Exception("Client is offline"));
-          },
-          GET_CONNECT_TIMEOUT,
```
schmidt-sebastian, line 443: GET_CONNECT_TIMEOUT can be removed. This is the only invocation.

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -522,36 +526,37 @@ public void run() {
...
-                                  query.getRef(),
-                                  IndexedNode.from(serverNode, query.getSpec().getIndex())));
+                    (@NonNull Task<Object> task) -> {
+                      // We do not need to synchronize with the persisted set result callback
+                      // because the run on the same run loop. If the delayed callback above runs
+                      // before this one, this task is a no-op.
+                      if (!task.isSuccessful()) {
+                        if (!persisted.exists()) {
+                          source.setException(
+                              task.getException() != null
```
schmidt-sebastian, line 545: This should never be null if the task failed.

schmidt-sebastian:
> In general the `get()` API should follow the connectivity behavior of `once` (when `once` cannot return a locally cached value), since that is the behavior devs are used to.

@puf - Do you think we should just get rid of the timeout? 3s seems short during app startup and long once the connection has been initialized.

puf:
If `once()` does a timeout it'd be fine to do it for `get()` too. If
`once()` doesn't use ` timeout, I'd prefer to use whatever `once()` uses to
determine whether the connection has stabilized on startup for `get()` too.

schmidt-sebastian:
There is no timeout for `once()` as far as I can tell. We can wait for 3s and fall back to cache if a value is available, otherwise we just wait for for the backend connection.

puf:
"we just wait for for the backend connection."

So `once()` doesn't fire until it has establishes a connection?
Well unless it has a cached value of course, in which case it
fires immediately.

I guess that makes sense, so I'm not sure how to map this to `get()`. It
currently seems to fail even on a good connection when I call `get()` in my
`main` in Flutter, and I'd expect it to only fail there once it explicitly
failed to establish a connection rather than when it not *yet* has been
able to establish a connection.

schmidt-sebastian:
FWIW, FIrestore fails these lookups 30s after the connection attempt, and even at 30 seconds this is a frequent source of bugs.

firebase-database/src/androidTest/java/com/google/firebase/database/QueryTest.java
```diff
@@ -4556,22 +4558,31 @@ public void testGetReturnsNullForEmptyNodeWhenOnline()
...
     FirebaseDatabase db = FirebaseDatabase.getInstance(app);
-    DatabaseReference node = db.getReference().child(db.getReference().push().getKey());
+    DatabaseReference node =
+        db.getReference().child(Objects.requireNonNull(db.getReference().push().getKey()));
     db.goOffline();
+    AtomicBoolean receivedValue = new AtomicBoolean();
+    THREAD_POOL_EXECUTOR.execute(
+        () -> {
+          try {
+            Thread.sleep(3000L);
```
schmidt-sebastian, line 4572: We can now reduce this sleep to something much lower in order not to slow down our tests.

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -522,36 +527,33 @@ public void run() {
...
-                          source.setResult(
-                              InternalHelpers.createDataSnapshot(
-                                  query.getRef(),
-                                  IndexedNode.from(serverNode, query.getSpec().getIndex())));
+                    (@NonNull Task<Object> task) -> {
+                      if (source.getTask().isComplete()) {
+                        return;
+                      }
+                      if (!task.isSuccessful()) {
+                        if (!persisted.exists()) {
```
schmidt-sebastian, line 544: If the task is not successful and we do have a persisted value we should set it right away. We should not wait for 3s.

firebase-database/src/main/java/com/google/firebase/database/core/Repo.java
```diff
@@ -522,36 +527,33 @@ public void run() {
               return;
             }
             serverSyncTree.setQueryActive(query.getSpec());
+            final DataSnapshot persisted = serverSyncTree.persistenceServerCache(query);
+            if (persisted.exists()) {
+              // Prefer the locally persisted value if the server is not responsive.
+              scheduleDelayed(() -> source.trySetResult(persisted), 3000);
```
schmidt-sebastian, line 533: Use a constant for `3000`?

schmidt-sebastian (approved):
Thanks

PR #3488, opened 2022-02-25
Remove need for alternative build script.

rlazo:


alternative-settings.gradle
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Google LLC
```
vkryachko, line 1: Can we remove this file as well? looks like it's not needed anymore

rlazo: Not yet.

build.gradle
```diff
@@ -50,12 +51,19 @@ ext {
     robolectricVersion = '4.3.1'
     protocVersion = '3.14.0'
     javaliteVersion = '3.14.0'
+
+    gMavenRoot = 'https://dl.google.com/dl/android/maven2'
+    firebaseDefaultPreguardFile='oss/default-preguard.txt'
```
vkryachko, line 56: does this path exist? I can't see the oss directory in the repo

rlazo: Removed

rlazo:
Friendly ping.

PR #3489, opened 2022-02-25
Fixing sign in cancellation bug

mannyjimenez0810:
Changing the cancellation of the sign in task from onActivityStarted too onActivityResume. This is because in the original implementation if the user hit back before being redirected to the web the task would hang. This fix makes it so that if the user clicks back into the app at any point the task will be cancelled.


rachaprince (approved):

lfkellogg (approved):
Looks good! I thought about this some more and I like this change should be very safe. I can't think of any weird edge cases. Still makes sense to wait until after the code freeze though.

Issue #1807, opened 2020-07-22
Terrible architecturedesign for FCM messages


bolds07:
### [READ] Step 1: Are you in the right place?

Yes, your people might consider it is a feature request but to me this is clearly a bug due a bad architecture.

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: any
  * Firebase Component: FCM
  * Component version: any

#### Steps to reproduce:

When using FCM to push notifications to android devices, developpers need to deal with a 6 posible scenarios described in this table 
https://firebase.google.com/docs/cloud-messaging/android/receive#handling_messages

the simple fact of having a 6 scenario table for something simple like "push notification" should alread trig an 'alert' in any architect mind, but the thing gets worse when we think about the messages with data. they might be delivered to `onMessageReceived` and might be to the launcher intent. 

so it automatically creates a scenario where the code will be duplicated, two classes will need to deal with same business logic. but this is not the end

follow my logic:
1- app is in background
2- fcm pushes a data+notification message
3- user sees the notification in system tray... and simple ignores it,
 or even has that kind of notification set to NEVER SHOW on android settings

following this scenario the app WILL NEVER RECEIVE THE DATA FROM THAT NOTIFICATION. since the launcher intent from the notification tray will never be clicked

i dont know what kind of android restrictions lead the engineers decide for this such frankstein behavior but at first glance it could be easily fix

if the android-fcm SDK is able to send the data messages to `onMessageReceived` even during background time... why the mixed data+notification could not send the data part also to `onMessageReceived`?

imho everything could be sent to that method and this class could implement a very simple method allowing any developper to foward the notifications to system tray from the `onMessageReceived` method

this would avoid so much code replication and create a less franksteins behavior

carlonzo:
Hi, I would like to add a few points to your post:
- reading a post with the title "TERRIBLE ARCHITECTURE" is totally starting in the wrong way, accusing the engineers who are supposed to come here and help you to have taken terrible architecture decisions. 
- if I was one of them I would have stop reading at the title and close the issue. but I was in the mood of reading something to make a good laugh, so I kept going.
- you start the post saying there is a "bug" due to bad architecture. sadly there is no bug described.
- you complain that there are 6 (6!!) differen scenarios to handle, referring to a table where it shows you 2 scenarios (background and foreground) describing what happens when a notification message is received. 
[...]

Issue #46, opened 2018-09-25
Call FirebaseMessagingService#onMessageReceived, the app in the background.


najeira:
## What feature would you like to see?

Call FirebaseMessagingService#onMessageReceived, the app in the background.

morepork:
Hi, just to clarify, do you mean that for notification messages to call onMessageReceived() at the same time as the notification is shown, or once the user taps on and opens the notification, or something else?

najeira:
before notification. 

I would like to customize notifications using NotificationCompat.BigPictureStyle and NotificationCompat.BigTextStyle.

samtstern:
@najeira I think what you want to do is send a `data` message rather than a `notification` message, which will give you full control over the UI you display.

najeira:
@samtstern Thanks for your reply.

I think that using `data` message to show notifications is bad workaround. On the server side, it is cumbersome to do different processing for iOS and Android. For example, with topic delivery, iOS and Android can not be separated..

julien1619:
Same request here! In our usecase we want to be able to update the badge displayed by some launchers, but, as we are not called when the app is in background, we have to use only the data payload, encapsulate the whole notification data in it, use the iOS custom keys to keep it working like before on iOS, and finally handle the notification display on Android manually.

I can't see any technical reason to not call `onMessageReceived()` everytime, even in background.

samtstern:
> Is there any reason to not provide a means to customize the notifications in the background?

This is a technical restriction that's outside the scope of the Android SDK.  It's really an OS-level decision.  Remember, the FCM message is received by the Google Play services process and your process is not always notified.

I am going to close this issue as there's nothing that can be done in this repo to fix it.  However I will forward on your feedback to the FCM team in case they decided to revisit their decisions on background processing in the future.

najeira:
> This is a technical restriction that's outside the scope of the Android SDK. It's really an OS-level decision. 

I can not agree with this. 

Android OS allows us to customize notifications even if apps is in background.

It is decision of SDK not to call onMessageReceived, It's written in `FirebaseMessagingService#handleIntent`.

Instead, please, please remove `final` from `FirebaseMessagingService#handleIntent`.

Without this, I can not move from GCM to FCM. I am very embarrassed.

najeira:
It is rare to receive notification when an application is in the foreground.  For example, Gmail also displays special notifications. Why are not we allowed?

kroikie:
[...]

PR #3492, opened 2022-02-27
Enable conditional opt out of automatically posting notifications to system tray

### LINKED ISSUE #2639: [messaging] Ability to change the notification before it's sent to the system tray
## What feature would you like to see?

Hi, is there anything on the roadmap to provide the ability to customise the notification before it's sent to the system tray.  I'm a maintainer of [notifee](https://github.com/notifee/react-native-notifee), and we recently added support to do this with iOS, we'd love to be able to do it with Android too.

Currently, the only option there seems to be it is to send data messages. A couple of issues with this method are that you need to send two different types of messages based on if the device is android or iOS, and this also requires to keep track of which the platform the token was registered with. 

In addition to platform-dependent code, there's also different background restrictions for data-only vs notification messages (ref [here](https://firebase.google.com/docs/cloud-messaging/concept-options#setting-the-priority-of-a-message)). 

On iOS, you can alter the notification using a Notification Service Extension where you have a brief amount of time to modify the notification before it's displayed to the user. As well as the `message.apns.aps` key offers a lot of flexibility where you can add a categoryId, sound, threadId e.t.c to the payload. 

One possible solution could be to allow [FirebaseMessagingService.onMessageReceived](https://firebase.google.com/docs/cloud-messaging/android/receive#handling_messages)  to be called for all app states, maybe a check somehow to see if the app does supports this, if not, default to handling with FCM? If this was possible, the one issue I do see with this, is that it could clash with other libraries like `FlutterFire` and `react-native-firebase` as you can only have one FirebaseMessagingService per app. 

Another solution, which I think seems way easier and less intrusive is to have a separate service, like how iOS works, to alter the notification and nothing else.

I can see how this could be a big f/r, and maybe there's reasons that it's not been done?

The other possible solution is to expand the predefined set of user-visible keys for `[message.android](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#androidnotification)`, to include `group` and other `NotificationCompat` properties where possible. 

To summarise, the two pain points I think there is at the moment with Android is:
 - the options allowed via [message.android](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#androidnotification) is limited.
 - no ability to intercept the notification before it's shown when the app is in background
 
## How would you use it?

Tell us how you'd use this feature in your app.

This feature would allow users of this library to send notification messages and customise the notification before it's finally sent to the system tray.

sfuqua:
This change adds an extension point to `FirebaseMessagingService` such that an application can avoid the default behavior where a **display message** (an FCM message with a "notification" payload), received while the app is in the **background**, will automatically post a default-styled Android notification to the system tray without ever calling back into app code.

This is a stab at addressing longstanding community feedback about how `onMessageReceived` is not called for display messages when the app is backgrounded, which can be a point of confusion and frustration:

https://github.com/firebase/firebase-android-sdk/issues/46
https://github.com/firebase/firebase-android-sdk/issues/1807
Fixes #2639 I believe
https://github.com/firebase/quickstart-android/issues/4
https://stackoverflow.com/q/37711082/244184
https://stackoverflow.com/q/37358462/244184

It is desirable as a developer to be able to opt out of this default behavior on a per-notification basis.
This allows an app to leverage the default, convenient behavior "for free" until there's time to customize the display with rich behavior using Android platform features. When that happens, the developer can use this new API to opt-out of the default handling and own their own display based on the message payload.

The proposed API change is a single new overrideable method on `FirebaseMessagingService`, called `shouldUseDefaultNotificationBehavior` - default implementation is just "return true".

If this method is overridden to return false for a given `RemoteMessage`, the current notification logic will be bypassed and the message will fall-through to `onMessageReceived` - the same way "data messages" are handled and the same way display messages while the app is foregrounded get handled.

I tested these changes on:
*  an Android 7/SDK 24 emulator
* a physical Pixel 6 (Android 12/SDK 31)
* a Wear OS 2 (Android 9/SDK 28) emulator 

Behavior is as expected - when my delegate returns false, the default notification is suppressed and `onMessageReceived` is called, as desired. When the delegate returns true, I get default SDK behavior.

# API considerations

* Backwards compatible, maintain current behavior by default but allow "progressive enhancement" by the app
* Leverage existing public types (`RemoteMessage`) instead of exposing the full FCM `Intent` to the app
* Aiming to be as surgical as possible here to unblock developer scenarios while not overhauling the FCM client API surface

Given the history in this space and comments from Firebase folks on linked issues (especially #46) - I feel like I must be missing something on the why of the current SDK design, so looking forward to feedback here! Is there a scenario where this doesn't work and the OS doesn't invoke the SDK, or a reason Google/Firebase would not be okay allowing an app to bypass the current SDK logic for background notifications to configure its own display?

### DIFF
--- a/firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessagingService.java
+++ b/firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessagingService.java
@@ -145,6 +145,22 @@ public class FirebaseMessagingService extends EnhancedIntentService {
   public void onNewToken(@NonNull String token) {}
   ;
 
+  /**
+   * Called when this service is evaluating whether a given RemoteMessage should use standard
+   * behavior when it has a notification payload - that is, automatically show a Notification in the
+   * system tray and do not invoke {@link onMessageReceived} when the app is backgrounded.
+   *
+   * <p>A subclass can override this to return false for specific classes of messages if it wishes
+   * to suppress the default system tray behavior and handle display itself via onMessageReceived.
+   *
+   * @param message The messge object that has been parsed from an incoming push payload.
+   * @return True if default behavior should be used; false if no system tray Notification should be
+   *     displayed and onMessageReceived should be invoked while backgrounded.
+   */
+  public boolean shouldUseDefaultNotificationBehavior(@NonNull RemoteMessage message) {
+    return true;
+  }
+
   /** @hide */
   @Override
   protected Intent getStartCommandIntent(Intent originalIntent) {
@@ -212,7 +228,8 @@ public class FirebaseMessagingService extends EnhancedIntentService {
     // First remove any parameters that shouldn't be passed to the app
     // * The wakelock ID set by the WakefulBroadcastReceiver
     data.remove("androidx.content.wakelockid");
-    if (NotificationParams.isNotification(data)) {
+    RemoteMessage message = new RemoteMessage(data);
+    if (shouldUseDefaultNotificationBehavior(message) && NotificationParams.isNotification(data)) {
       NotificationParams params = new NotificationParams(data);
 
       ExecutorService executor = FcmExecutors.newNetworkIOExecutor();
@@ -232,7 +249,7 @@ public class FirebaseMessagingService extends EnhancedIntentService {
         MessagingAnalytics.logNotificationForeground(intent);
       }
     }
-    onMessageReceived(new RemoteMessage(data));
+    onMessageReceived(message);
   }
 
   private boolean alreadyReceivedMessage(String messageId) {


firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessagingService.java
```diff
@@ -232,7 +249,7 @@ private void dispatchMessage(Intent intent) {
         MessagingAnalytics.logNotificationForeground(intent);
```
sfuqua, line 249: at the moment the change will bypass this `MessagingAnalytics` call, but I think that's probably okay?

if someone uses this new API to bypass default notification logic, the implication is that it should be treated identically to a standard data message (including for analytics)

sfuqua:
@samtstern While the review process kicks off, I'm hoping you might be able to elaborate on some past comments you've made in linked issues, specifically:

https://github.com/firebase/firebase-android-sdk/issues/46#issuecomment-447977454

> > Is there any reason to not provide a means to customize the notifications in the background?
> 
> This is a technical restriction that's outside the scope of the Android SDK. It's really an OS-level decision. Remember, the FCM message is received by the Google Play services process and your process is not always notified.
> 
> I am going to close this issue as there's nothing that can be done in this repo to fix it. However I will forward on your feedback to the FCM team in case they decided to revisit their decisions on background processing in the future.

In my testing (at least as far back as API level 24) and in inspecting code in this repo, the logic in question is owned by this SDK - I feel like I must be missing something as pertains to Play services and how the OS handles incoming messages and passes them off to userland via the `com.google.android.c2dm.intent.RECEIVE` action used internally by this SDK's BroadcastReceiver.

As far as I can tell this is a developer scenario that _maybe_ used to work with GCM (when the app implemented a BroadcastReceiver for that intent themselves) and is now blocked by the current Firebase SDK design, but I can't tell if that's been intentional or just a gap - I can see some folks mentioning similar behavior with GCM:

https://github.com/googlesamples/google-services/issues/166
https://stackoverflow.com/q/35480573/244184

However I am not convinced that this is actual Android OS behavior - I suspect this was an implementation detail of `GcmListenerService`, the same as it is an implementation detail of `FirebaseMessagingService`, and that an app has always been free to write their own `BroadcastReceiver` and handle the push notification Intent however they like - so I don't understand the intention of blocking the behavior in the SDK but am hoping someone can shine some light here.

sfuqua:
Hi @gsakakihara - I'm hoping you or someone else from either the Android or Firebase team can chime in on this PR.

It's a pretty scoped change that unblocks a lot of developer power, but I've seen a lot of confusing history in this space in other issues and it's not clear to me if there's actually Android platform-level issues involved in this scenario, or if it's purely a design choice of the GCM and FCM user-land libraries.

I understand an API change like this will require careful scrutiny from the team, but at a fundamental level, is there something objectionable about this sort of approach to the problem?

See my previous comment with research/questions in this space: https://github.com/firebase/firebase-android-sdk/pull/3492#issuecomment-1053712629

sfuqua:
Hi @gsakakihara - this PR has been in limbo for several months. I don't expect to merge it as-is and I understand an API review is not trivial. However I'm very interested in a deeper understanding of the issues at play here and whether an approach like this would work for the Firebase and Android teams, or whether I'm fundamentally misunderstanding how push notifications work at an Android OS level.

There have been a lot of comments about how "display notifications" are handled automatically, but that _seems_ to be untrue given that both the Firebase and GCM SDKs (which are app code, not special system code, getting invoked as standard `BroadcastReceivers`) _both_ have logic in place to display prefab notifications that call standard, public Android `Notification` and `NotificationChannel` APIs.

If some part of the Android OS _is_ automatically displaying notifications and it's _not_ an SDK issue (in which case, why does the SDK have code to display notifications?), a PR like this one would obviously not be useful.

As a developer the current notification dichotomy on Android is _very_ frustrating, and from the outside looking in it's not clear why it was designed the way that it was. If an "escape hatch" like the one introduced in this PR is possible it would make my life and the lives of many other Android developers much easier.

suhaibkazi (approved):
As per tests i conducted and internal logs, this workaround is the least intrusive and still allows for normal behaviour to take place as required.

eddiecubed:
+1

Can we resolve the conversations in the PR to have this merged?

gsakakihara:
The FCM team has previously looked into allowing developers more control over when and how notifications are displayed by the FCM SDK, but this doesn't fit into the current roadmap at this time. The best solution for now is probably to use a custom build of the SDK with a change like this that will allow you to override the SDK's automatic handling of notifications.

sfuqua:
It might be futile at this point, but I'll offer that I'm frustrated that it took a year and a half to get any comment on this change, and that it has been completely impossible over many years to get the FCM team to offer a straight answer around the automatic behavior of notifications in the SDK.

The SDK's current behavior makes for one of the most perplexing "official" APIs I have ever used on any platform.

FWIW, it is very difficult to use a custom build of the FCM SDK with the rest of the Firebase suite. I spent quite a while trying to get an equivalent of npm's `patch-package` working where I could tamper with your AAR and overwrite just a couple lines of code, or the implementation of a single class, but it's not straightforward. As a result, I think the workaround of extending `FirebaseMessageService` with a hacky override of `handleIntent` is really the only maintainable thing that works today.

RikScheffer:
would be very useful to have this

Vennygis:
Add this first one

vennygis (approved):
Check 1

PR #3499, opened 2022-03-01
Support descending queries

schmidt-sebastian:
I am really struggling to understand why this was broken and not detected by the existing tests. I also wonder whether this is actually the right fix and could use a thorough set of eyes.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
@@ -17,6 +17,7 @@ package com.google.firebase.firestore.core;
 import static com.google.firebase.firestore.model.Values.max;
 import static com.google.firebase.firestore.model.Values.min;
 
+import android.util.Pair;
 import androidx.annotation.Nullable;
 import com.google.firebase.firestore.model.DocumentKey;
 import com.google.firebase.firestore.model.FieldIndex;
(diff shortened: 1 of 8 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
```diff
@@ -99,6 +99,28 @@ public void testOrderByFilter() {
...
+    addDoc("coll/val6", map("a", 3, "b", 3, "c", 6));
+    Query query =
+        query("coll")
+            .filter(filter("a", "==", 2))
+            .filter(filter("b", "==", 2))
+            .filter(filter("c", ">", 2))
+            .orderBy(orderBy("c", "desc"))
+            .orderBy(orderBy("b", "asc"))
+            .orderBy(orderBy("a", "desc"))
+            .limitToFirst(2);
```
ehsannas, line 120: I would remove `limitToFirst` as it's not the intent of the test.

For completion, I'd suggest testing the 4 combinations:
* `c > 2` and `c DESC` (already covered here)
* `c > 2` and `c ASC`
* `c < 5` and `c ASC`
* `c < 5` and `c DESC`

ehsannas: Ideally we should also add some test cases that have `startAt`/`endAt` for each of them

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -185,65 +186,33 @@ public Bound getLowerBound(FieldIndex fieldIndex) {
...
+      // can narrow the scope.
       if (startAt != null) {
         for (int i = 0; i < orderBys.size(); ++i) {
           OrderBy orderBy = this.orderBys.get(i);
           if (orderBy.getField().equals(segment.getFieldPath())) {
             Value cursorValue = startAt.getPosition().get(i);
-            if (max(segmentValue, cursorValue) == cursorValue) {
-              segmentValue = cursorValue;
-              segmentInclusive = startAt.isInclusive();
+            if (max(segmentBound.first, cursorValue) == cursorValue) {
```
ehsannas, line 201: took me a minute to realize why we always do `max` here, and not decide whether we should do `min` or `max` based on the segment kind.
We should add tests that contain startAt and endAt.

schmidt-sebastian: I added tests, which actually showed that this logic is wrong (I compared the result with the Java Server SDK for validation). This is now part of the getAscendingBound() logic.

ehsannas (approved):

PR #3502, opened 2022-03-02
Support sign in redirect when applicationId contains underscore

mannyjimenez0810:


firebase-appdistribution/src/main/AndroidManifest.xml
```diff
@@ -35,6 +35,15 @@
         <activity android:name="com.google.firebase.appdistribution.internal.InstallActivity" android:exported="true" android:launchMode="singleTask"/>
 
         <activity android:name="com.google.firebase.appdistribution.internal.SignInResultActivity" android:exported="true">
+            <intent-filter>
```
rachaprince, line 38: Could we add a comment explaining why we added a second intent filter?

rachaprince (approved):

firebase-appdistribution/src/main/AndroidManifest.xml
```diff
@@ -35,6 +35,15 @@
...
+                <action android:name="android.intent.action.VIEW"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+                <category android:name="android.intent.category.BROWSABLE"/>
+
+                <data android:scheme="appdistribution" android:path="/${applicationId}" android:host="authredirect" />
```
lfkellogg, line 43: Nit: could you rearrange this so that the order reflects the actual order in a URI: scheme, path, host

lfkellogg (approved):
Awesome! Just one nit.

Also, I assume this was manually tested, but this this a good argument for us to set up an integration test or something to cover this behavior.

lfkellogg (changes requested):
Actually, see my comment on cl/432023345

I think we should transition to the new URI, rather than support either depending on whether or not there is an underscore.

rachaprince (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/TesterSignInManager.java
```diff
@@ -45,7 +45,7 @@
 class TesterSignInManager {
   private static final String TAG = "TesterSignInManager:";
   private static final String SIGNIN_REDIRECT_URL =
-      "https://appdistribution.firebase.google.com/pub/testerapps/%s/installations/%s/buildalerts?appName=%s&packageName=%s";
+      "https://appdistribution.firebase.google.com/pub/testerapps/%s/installations/%s/buildalerts?appName=%s&packageName=%s&newRedirectScheme=appdistribution";
```
lfkellogg, line 48: I think `newRedirectScheme=true` would make more sense to me, since we don't actually use the value, and the param name sounds like it should have a boolean assigned to it anyway.

firebase-appdistribution/src/main/AndroidManifest.xml
```diff
@@ -35,12 +35,12 @@
...
 
         <activity android:name="com.google.firebase.appdistribution.internal.SignInResultActivity" android:exported="true">
-            <intent-filter>
+               <intent-filter>
                 <action android:name="android.intent.action.VIEW"/>
                 <category android:name="android.intent.category.DEFAULT"/>
                 <category android:name="android.intent.category.BROWSABLE"/>
 
-                <data android:scheme="appdistribution-${applicationId}" android:host="authredirect" />
+                <data android:scheme="intent" android:host="authredirect" android:path="/appdistribution-${applicationId}"/>
```
lfkellogg, line 43: See my CL comment: https://critique.corp.google.com/cl/432023345/depot/google3/alkali/apps/appdistribution/client/app/create_installation_link/create_installation_link.ts?version=s7#119

firebase-appdistribution/src/main/AndroidManifest.xml
```diff
@@ -40,7 +40,7 @@
                 <category android:name="android.intent.category.DEFAULT"/>
                 <category android:name="android.intent.category.BROWSABLE"/>
 
-                <data android:scheme="appdistribution-${applicationId}" android:host="authredirect" />
+                <data android:scheme="intent" android:host="appdistribution" android:path="/${applicationId" />
```
lfkellogg, line 43:
```suggestion
                <data android:scheme="intent" android:host="appdistribution" android:path="/${applicationId}" />
```

lfkellogg (changes requested):

lfkellogg (approved):

PR #3505, opened 2022-03-03
DO NOT MERGE: Alternative proposal for go/firestore-snapshot-reads

dconeybe:


firebase-firestore/src/main/java/com/google/firebase/firestore/ReadOnlyTransaction.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @NonNull
+  public DocumentSnapshot get(@NonNull DocumentReference documentRef)
+      throws FirebaseFirestoreException {
+    return get(documentRef, null);
+  }
+
+  @NonNull
+  public DocumentSnapshot get(@NonNull DocumentReference documentRef, @Nullable Timestamp readTime)
```
schmidt-sebastian, line 34: FWIW While we could implement this, this breaks the transaction concept of using the same read time across all reads.

dconeybe: Okay, I'll fix this.

schmidt-sebastian: Thanks. We need to ensure that a transaction always returns a consistent view of data.

dconeybe: Done.

dconeybe:
This feature has been put on hold indefinitely

PR #3513, opened 2022-03-08
Add mutability flag to pending intent

rachaprince:
Add a mutability flag to the pending intent to support apps targeting android 12. 

See https://developer.android.com/guide/components/intents-filters#DeclareMutabilityPendingIntent

We use a pending intent because this is sent to the Notifications Manager. See https://developer.android.com/guide/components/intents-filters#PendingIntent

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
@@ -111,6 +111,19 @@ class FirebaseAppDistributionNotificationsManager {
       return null;
     }
     return PendingIntent.getActivity(
-        context, 0, intent, PendingIntent.FLAG_ONE_SHOT | PendingIntent.FLAG_IMMUTABLE);
+        context, 0, intent, getPendingIntentFlags(PendingIntent.FLAG_ONE_SHOT));
+  }
+
+  /**
+   * Adds {@link PendingIntent#FLAG_IMMUTABLE} to a PendingIntent's flags since any PendingIntents
+   * used here don't need to be modified.
+   *
+   * <p>Specifying mutability is required starting at SDK level 31.
+   */
+  private static int getPendingIntentFlags(int baseFlags) {
+    // Only add on platform levels that support FLAG_IMMUTABLE.
+    return Build.VERSION.SDK_INT >= Build.VERSION_CODES.M
+        ? baseFlags | PendingIntent.FLAG_IMMUTABLE
+        : baseFlags;
   }
 }


mannyjimenez0810 (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -110,6 +110,7 @@ private PendingIntent createAppLaunchIntent() {
       LogWrapper.getInstance().w(TAG + "No activity found to launch app");
       return null;
     }
-    return PendingIntent.getActivity(context, 0, intent, PendingIntent.FLAG_ONE_SHOT);
+    return PendingIntent.getActivity(
+        context, 0, intent, PendingIntent.FLAG_ONE_SHOT | PendingIntent.FLAG_IMMUTABLE);
```
lfkellogg, line 114: This makes me wonder if we should avoid using FLAG_IMMUTABLE on older SDK versions: https://github.com/firebase/firebase-android-sdk/blob/21a9f912862bfbd801690a9f88dce22366bec8ac/firebase-messaging/src/main/java/com/google/firebase/messaging/CommonNotificationBuilder.java#L536-L540

lfkellogg: Ah, it looks like FLAG_IMMUTABLE was only added in API level 23 (M) https://developer.android.com/reference/android/app/PendingIntent#FLAG_IMMUTABLE

vkryachko: The issue is that your sdk has misSdkVersion = 16, so it could crash at runtime?

lfkellogg: But the check against `Build.VERSION.SDK_INT` would prevent that crash, right @vkryachko ?

vkryachko: right, but I don't see the check it in the pr, and I thought that in your previous comment you were suggesting that it's safe not to do the check given that this flag is 23+. If we have the check then it shouldn't be a problem.

lfkellogg: Ah, no, sorry. I meant we need to add the check. We'll do that now.

lfkellogg: Done.

mannyjimenez0810 (approved):

PR #3515, opened 2022-03-08
Adding more specific notification messages

mannyjimenez0810:
Notification says "Download failed" if other failure cases occur (e.g. installation cancelled). Download failed is no longer a catch all for other failure cases.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -65,24 +65,24 @@ void updateNotification(long totalBytes, long downloadedBytes, UpdateStatus stat
...
-    } else if (status.equals(UpdateStatus.DOWNLOADED)) {
-      return R.string.download_completed;
-    } else {
-      return R.string.downloading_app_update;
+    switch (status) {
+      case DOWNLOAD_FAILED:
+        return R.string.download_failed;
+      case INSTALL_FAILED:
+        return R.string.install_failed;
+      case INSTALL_CANCELED:
```
mannyjimenez0810, line 73: The UpdateStatus comes from the ApkUpdater class. We don't currently send Install_canceled or new_release_check_failed. I figured we could keep it for now since those are valid UpdateStatuses. 

In the spots we could be more specific it would require a new ENUM, which does not seem worth it. Ex in apk updater.
https://github.com/firebase/firebase-android-sdk/blob/3072bbcdb5416b0c1accda91a9cd9f24cde6063e/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java#L240

Overall adding more enums for specificity doesn't seem worth it since we throw firebase exceptions and getting more specific ex install canceled instead of install failed (link to code block below) also would be nice, but not a low lift.
https://github.com/firebase/firebase-android-sdk/blob/3072bbcdb5416b0c1accda91a9cd9f24cde6063e/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java#L130

Thoughts?

rachaprince: Nice investigation. I agree that more specific error messages don't seem worth the effort.

Especially since new ENUMs would require an API change. 

I think that as long as the category of error is correctly displayed, we can point developers and testers to the other SDK logs to debug

firebase-appdistribution/src/main/res/values/strings.xml
```diff
@@ -25,6 +25,10 @@
   <string name="downloading_app_update">Downloading in-app update...</string>
   <string name="download_completed">Download completed</string>
   <string name="download_failed">Download failed</string>
+  <string name="install_failed">Install Failed</string>
```
rachaprince, line 28: nit: Could we make this casing consistent with the other error messages? (The first letter capitalized, the rest lower case)

rachaprince (approved):

firebase-appdistribution/src/main/res/values/strings.xml
mannyjimenez0810: good catch! thanks!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -65,24 +65,24 @@ void updateNotification(long totalBytes, long downloadedBytes, UpdateStatus stat
...
+        return R.string.install_canceled;
+      case NEW_RELEASE_CHECK_FAILED:
+        return R.string.new_release_check_failed;
+      case UPDATE_CANCELED:
+        return R.string.update_canceled;
+      case DOWNLOADED:
+        return R.string.download_completed;
+      default:
+        return R.string.downloading_app_update;
```
lfkellogg, line 82: It's a little strange to default to "Downloading..." For example why is "Downloading..." the right message when the status is PENDING? https://github.com/firebase/firebase-android-sdk/blob/3072bbcdb5416b0c1accda91a9cd9f24cde6063e/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java#L182

I think ideally the callers to `updateNotification()` would just pass in a string resource ID directly, and then we wouldn't need to do this Enum -> String translation here at all.

mannyjimenez0810: Added in the case for pending to return that it is pending. That should cover all the cases that we use in ApkUpdater. My sense is that will only be in the notification for a small time (even smaller than downloading but that can take more time depending on file size and internet speed). With the next change there shouldn't be a default regardless. 

Why is that ideal? Made the change to pass both, because postUpdateProgress needs the status for the updateProgress builder.

firebase-appdistribution/src/main/res/values/strings.xml
```diff
@@ -25,6 +25,10 @@
   <string name="downloading_app_update">Downloading in-app update...</string>
   <string name="download_completed">Download completed</string>
   <string name="download_failed">Download failed</string>
+  <string name="download_pending">Download failed</string>
```
lfkellogg, line 28: "Download pending"

lfkellogg (approved):

9996142202 (approved):

PR #3518, opened 2022-03-10
Change DocumentOverlayCache.saveOverlays() to require non-null values in the given Map

dconeybe:
The current implementations of `DocumentOverlayCache.saveOverlays()` checks that each value in the given `Map` is non-null, and if it is null it ignores it, as if the key/value pair wasn't even present in the map.

This PR changes the logic to _assume_ that none of the values in the `Map` are null and modifies the call sites to ensure that they avoid putting null values into the map.

This is a slightly cleaner implementation and makes the null behavior more explicit at the call site.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
@@ -50,11 +50,7 @@ public class MemoryDocumentOverlayCache implements DocumentOverlayCache {
     return result;
   }
 
-  private void saveOverlay(int largestBatchId, @Nullable Mutation mutation) {
-    if (mutation == null) {
-      return;
-    }
-
+  private void saveOverlay(int largestBatchId, Mutation mutation) {
     // Remove the association of the overlay to its batch id.
     Overlay existing = this.overlays.get(mutation.getKey());
     if (existing != null) {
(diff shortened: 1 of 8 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
```diff
@@ -50,11 +50,7 @@ public Overlay getOverlay(DocumentKey key) {
     return result;
   }
 
-  private void saveOverlay(int largestBatchId, @Nullable Mutation mutation) {
-    if (mutation == null) {
-      return;
-    }
-
```
ehsannas, line 57: we should assert mutation != null, since `mutation.getKey()` is being used below.

ehsannas: or does removing the `@Nullable` achieve the same thing?

ehsannas (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/MemoryDocumentOverlayCache.java
dconeybe: I added an explicit null check in `saveOverlays()`. @Nullable is just a compile-time check and doesn't affect runtime behavior. I also added a unit test.

Issue #3526, opened 2022-03-11
Messaging fails build with latest 29.2.0 release, 29.1.0 is fine


mikehardy:
### [READ] Step 1: Are you in the right place?

This is the right place

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: current stable, also not relevant for this issue
  * Firebase Component: Messaging
  * Component version: BoM 29.2.0, Messaging is 23.0.1

### [REQUIRED] Step 3: Describe the problem

For some reason the messaging component in BoM 29.2.0 leaks in httpcomponents transitive dependencies which have an APK packaging issue with regard to multiple `META-INF/DEPENDENCIES` files. It looks like it has been a known issue historically as [there is an attempt in your build.gradle to exclude them](https://github.com/firebase/firebase-android-sdk/blob/f103c35b84d85132c5a1859c4cf3ce68bb55961a/firebase-messaging/firebase-messaging.gradle#L103-L105). For some reason that isn't working on 29.2.0, even though it works on 29.1.0 (you can switch between the two BoMs in your messaging quickstart to witness success v failure)

#### Steps to reproduce:

This is me showing you the reproduction using your quickstart-android repository, cloned just now with a visible diff so you can see the changes required to reproduce:

```
mike@osxvm-spare:~/work/Invertase/quickstart-android/messaging (master *) % ./gradlew assembleDebug
> Task :app:mergeDebugJavaResource FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:mergeDebugJavaResource'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.MergeJavaResWorkAction
   > 2 files found with path 'META-INF/DEPENDENCIES' from inputs:
      - /Users/mike/.gradle/caches/modules-2/files-2.1/org.apache.httpcomponents/httpclient/4.5.11/f6d42fee5110c227bac18a550a297e028f2fb21a/httpclient-4.5.11.jar
      - /Users/mike/.gradle/caches/modules-2/files-2.1/org.apache.httpcomponents/httpcore/4.4.13/853b96d3afbb7bf8cc303fe27ee96836a10c1834/httpcore-4.4.13.jar
     Adding a packagingOptions block may help, please refer to
     https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.PackagingOptions.html
     for more information

* Try:
[...]

PR #3533, opened 2022-03-15
Fix fcm dependency

VinayGuthal:


mikehardy (approved):
Nice - I'm only here because I'm the original poster of #3526 - with this change I cannot imagine the google api-client dependency leaking in to the published artifact anymore, thus I cannot imagine the httpcomponents dependencies transitively leaking in either. Should do the trick. I recognize my approval doesn't carry weight but symbolically, :smile:  :+1:

firebase-messaging/src/main/java/com/google/firebase/messaging/ImageDownload.java
```diff
@@ -115,9 +115,16 @@ public Bitmap blockingDownload() throws IOException {
...
-      bytes =
-          ByteStreams.toByteArray(
-              ByteStreams.limit(connectionInputStream, MAX_IMAGE_SIZE_BYTES + 1));
+      ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+
+      int nRead;
+      byte[] data = new byte[MAX_IMAGE_SIZE_BYTES + 1];
+
+      while ((nRead = connectionInputStream.read(data, 0, data.length)) != -1) {
+        buffer.write(data, 0, nRead);
```
rlazo, line 124: This can end up reading several chunks of *up to* `MAX_IMAGE_SIZE_BYTES+1`, which is not equivalent to `ByteStreams.limit`. If you check @ [ByteStreams implementation](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/ByteStreams.java#L744) it limits the *total* value read to be up to `MAX_IMAGE_SIZE_BYTES+1` with a counter. You would need to implement something similar.

VinayGuthal:
> Is there a way to add a previous dependency to continue to use ByteStreams? It used to work without bringing in google-api-client.

@gsakakihara this would introduce a dependency on guava and would be a big regression in terms of size..Guava was vendored in before so it was fine. However now it will cause a lot of bloating

firebase-messaging/src/main/java/com/google/firebase/messaging/ImageDownload.java
VinayGuthal: i think it will be equivalent. bcz the reading is not in the for loop and ByteStreams.limit(connectionInputStream, MAX_IMAGE_SIZE_BYTES + 1)); is getting initialized everytime a function is called so the limit will be set everytime not once

VinayGuthal:
> 



> Is there a way to vendor it in now? Otherwise it seems like that section will need to be fixed since it is definitely a change in behavior.
> 
> Also compileOnly 'com.google.code.findbugs:jsr305:3.0.2' doesn't work, but:
> 
> implementation 'com.google.errorprone:error_prone_annotations:2.11.0' implementation 'com.google.code.findbugs:jsr305:3.0.2'
> 
> does seem to work, but I'm not sure what other effects on size and such that would have.

I think it should fairly be similar between androidx and javax (especially since this is android sdk androidx might be preferred). However if you want to keep it back to javax, i can add implementation 'com.google.code.findbugs:jsr305:3.0.2'  to the gradle and revert the annotation changes

VinayGuthal:
> Is there a way to vendor it in now? Otherwise it seems like that section will need to be fixed since it is definitely a change in behavior.
> 
> Also compileOnly 'com.google.code.findbugs:jsr305:3.0.2' doesn't work, but:
> 
> implementation 'com.google.errorprone:error_prone_annotations:2.11.0' implementation 'com.google.code.findbugs:jsr305:3.0.2'
> 
> does seem to work, but I'm not sure what other effects on size and such that would have.

Vendored in

VinayGuthal:
> Nice - I'm only here because I'm the original poster of #3526 - with this change I cannot imagine the google api-client dependency leaking in to the published artifact anymore, thus I cannot imagine the httpcomponents dependencies transitively leaking in either. Should do the trick. I recognize my approval doesn't carry weight but symbolically, 😄 👍

Thanks a lot @mikehardy for the review :) Your approval does carry weight!! Also thanks for pointing out this issue allowing us to fix this!

gsakakihara (approved):

rlazo (approved):

PR #3549, opened 2022-03-16
Don't use Indexes for outdated limit queries

schmidt-sebastian:
Just like with Index-Free queries, we can't use indexing for limit queries if one of the indexed values contains a non-indexed mutation. The test case should make this case obvious.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
@@ -117,8 +117,15 @@ public class QueryEngine {
 
     ImmutableSortedMap<DocumentKey, Document> indexedDocuments =
         localDocumentsView.getDocuments(keys);
-    return appendRemainingResults(
-        values(indexedDocuments), query, indexManager.getMinOffset(target));
+    IndexOffset offset = indexManager.getMinOffset(target);
+
+    ImmutableSortedSet<Document> previousResults = applyQuery(query, indexedDocuments);
+    if ((query.hasLimitToFirst() || query.hasLimitToLast())
+        && needsRefill(query.getLimitType(), keys.size(), previousResults, offset.getReadTime())) {
+      return null;
+    }
+
+    return appendRemainingResults(values(indexedDocuments), query, offset);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


wu-hui (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -117,8 +117,15 @@ public void initialize(LocalDocumentsView localDocumentsView, IndexManager index
...
     ImmutableSortedMap<DocumentKey, Document> indexedDocuments =
         localDocumentsView.getDocuments(keys);
-    return appendRemainingResults(
-        values(indexedDocuments), query, indexManager.getMinOffset(target));
+    IndexOffset offset = indexManager.getMinOffset(target);
+
+    ImmutableSortedSet<Document> previousResults = applyQuery(query, indexedDocuments);
+    if ((query.hasLimitToFirst() || query.hasLimitToLast())
+        && needsRefill(query.getLimitType(), keys.size(), previousResults, offset.getReadTime())) {
+      return null;
```
ehsannas, line 125: @schmidt-sebastian @wu-hui 

Say we have
```
doc1:  {a: 0, b: 0}
doc2:  {a: 1, b: 1}
doc3:  {a: 1, b: 2}
doc4:  {a: 1, b: 3}
doc5:  {a: 2, b: 4}
```
and that all docs are fully indexed, and no local modifications exist.

if we have a query like `a==1 && b==1 limit 3`, and have the `a ASC` index, the SQLite query will return `doc2, doc3, doc4` because it only reduces the scope to the docs with `a==1`. (we get these three documents regardless of whether the SQL logic applies a LIMIT or not. i.e. regardless of whether there's a partial index).

`indexedDocuments` here is now `doc2, doc3, doc4`.

When `applyQuery` is called, `doc3` and `doc4` are eliminated because they don't meet `b==1`. So `previousResults` is just `doc2`.

Now `needsRefill` returns `true` because size of `indexedDocuments` is 3, but size of `previousResults` is 1. So we return `null` here and the SQLite query results with be ignored and we fall back to full collection scan. This shouldn't be happening.

ehsannas: (verified this by writing a test in SQLiteQueryEngineTest.java)

schmidt-sebastian: Thanks for bringing this up. It looks like we need to prune the list of indexed documents to only include documents that actually match the query. I will look into this.

PR #3565, opened 2022-03-18
Fireperf: fragment lifecycle callbacks

leotianlizhan:
b/208271814

This PR doesn't include tests to unblock the other implementation tasks, tests will be added in another PR. 

EDIT: actually don't see a reason to test because these are just lifecycle callbacks, there's nothing in those callbacks so there's nothing to test.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentMonitor.java
@@ -1,3 +1,17 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package com.google.firebase.perf.application;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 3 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentMonitor.java
```diff
@@ -0,0 +1,71 @@
...
+import androidx.core.app.FrameMetricsAggregator;
+import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentActivity;
+import androidx.fragment.app.FragmentManager;
+import com.google.firebase.perf.logging.AndroidLogger;
+import com.google.firebase.perf.transport.TransportManager;
+import com.google.firebase.perf.util.Clock;
+import com.google.firebase.perf.util.Constants;
+
+public class FragmentMonitor extends FragmentManager.FragmentLifecycleCallbacks {
```
jeremyjiang-dev, line 27: Nit: maybe call it FragmentStateMonitor to match appStateMonitor

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
```diff
@@ -0,0 +1,71 @@
...
+      FrameMetricsAggregator fma) {
+    this.activity = activity;
+    this.clock = clock;
+    this.transportManager = transportManager;
+    this.appStateMonitor = appStateMonitor;
+    this.frameMetricsAggregator = fma;
+  }
+
+  /**
+   * Screen trace name is prefix "_st_" concatenates with Fragment's class name.
```
visumickey, line 49: Can you fix the comment here? Some ambiguity with the language here.

visumickey (approved):

leotianlizhan:
tests failures looks unrelated, seems like it's coming from remote-config, will just merge.

Issue #3571, opened 2022-03-21
firebase-messaging 23.0.2 size increase


carlonzo:
### [REQUIRED] Step 2: Describe your environment

  * Firebase Component: Messaging
  * Component version: 23.0.2

### [REQUIRED] Step 3: Describe the problem

With the new BoM, firebase-messaging has been upgraded from 23.0.0 to 23.0.2 (skipping 23.0.1 as was patched due to a wrong dependency).
 
Comparing the 2 versions, we have seen an noticeable increase in size in resources by 75kb (and code by 153kb).

Here is a comparison of the aars using Diffuse:
```
OLD: firebase-messaging-23.0.0.aar
NEW: firebase-messaging-23.0.2.aar

 AAR      │ old       │ new       │ diff       
──────────┼───────────┼───────────┼────────────
      jar │ 161.8 KiB │ 315.2 KiB │ +153.4 KiB 
 manifest │   2.4 KiB │   2.5 KiB │      +79 B 
      res │  23.3 KiB │    99 KiB │  +75.7 KiB 
    other │ 593.1 KiB │  13.6 KiB │ -579.5 KiB 
──────────┼───────────┼───────────┼────────────
    total │ 780.6 KiB │ 430.3 KiB │ -350.3 KiB 


 JAR     │ old  │ new  │ diff             
─────────┼──────┼──────┼──────────────────
 classes │  121 │   73 │  -48 (+13 -61)   
 methods │ 1139 │ 1036 │ -103 (+171 -274) 
  fields │  445 │  475 │  +30 (+149 -119) 


=================
====   AAR   ====
=================

 size      │ diff       │ path                                                   
───────────┼────────────┼────────────────────────────────────────────────────────
  34.3 KiB │  +34.3 KiB │ + res/raw/gcm_bip.mp3                                  
   5.3 KiB │   +5.3 KiB │ + res/drawable-xxxhdpi-v4/adaptive_icon_foreground.png 
   5.3 KiB │   +5.3 KiB │ + res/drawable-xxxhdpi-v4/icon_with_gradient.png       
   3.9 KiB │   +3.9 KiB │ + res/drawable/gcm_icon.png                            
   3.9 KiB │   +3.9 KiB │ + res/drawable-xxhdpi-v4/adaptive_icon_foreground.png  
   3.9 KiB │   +3.9 KiB │ + res/drawable-xxhdpi-v4/icon_with_gradient.png        
[...]

PR #3574, opened 2022-03-21
Move test resources out from firebase-messaging

carlonzo:
Tests resources were added in production code. This caused the latest release to increase aar size.

This patch moves those resources from main to test sourcepath.

Fixes https://github.com/firebase/firebase-android-sdk/issues/3571


gsakakihara:
fcm_fallback_notification_channel_label in strings.xml needs to stay in main, otherwise this looks good, assuming all tests pass.

carlonzo:
@gsakakihara  oh I see it is referenced by name. Thanks for that. I sent an update. tests are passing locally

gsakakihara:
I can review this once the tests are passing. It looks like unrelated tests that are failing, so it may just need to be rerun.

carlonzo:
@gsakakihara thanks! I've rebased and re run the tests. Should be good now

gsakakihara (approved):

PR #3575, opened 2022-03-21
Fireperf fragments: trace creation and adding custom attributes

leotianlizhan:
b/208271842

Currently there's a validation error:

> W/FirebasePerformance: non-positive totalFrames in screen trace _st_SupportRequestManagerFragment
W/FirebasePerformance: Invalid Trace:_st_SupportRequestManagerFragment
W/FirebasePerformance: Unable to process the PerfMetric (trace metric: _st_SupportRequestManagerFragment (duration: 118339.913ms)) due to missing or invalid values. See earlier log statements for additional information on the specific missing/invalid values.

This is due to no frame metrics for the "total_frames" validation check, because we have not implemented that yet.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -148,8 +148,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
         fragmentActivity
             .getSupportFragmentManager()
             .registerFragmentLifecycleCallbacks(
-                new FragmentStateMonitor(
-                    clock, transportManager, this, frameMetricsAggregator),
+                new FragmentStateMonitor(clock, transportManager, this, frameMetricsAggregator),
                 true);
       }
     }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -149,7 +149,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
             .getSupportFragmentManager()
             .registerFragmentLifecycleCallbacks(
                 new FragmentStateMonitor(
-                    fragmentActivity, clock, transportManager, this, frameMetricsAggregator),
+                    clock, transportManager, this, frameMetricsAggregator),
                 true);
       }
     }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
@@ -90,13 +90,22 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
   }
 
   AppStateMonitor(TransportManager transportManager, Clock clock) {
+    this(
+        transportManager,
+        clock,
+        ConfigResolver.getInstance(),
+        hasFrameMetricsAggregatorClass() ? new FrameMetricsAggregator() : null);
+  }
+
+  AppStateMonitor(
+      TransportManager transportManager,
+      Clock clock,
+      ConfigResolver configResolver,
+      FrameMetricsAggregator frameMetricsAggregator) {
     this.transportManager = transportManager;
     this.clock = clock;
-    configResolver = ConfigResolver.getInstance();
-    hasFrameMetricsAggregator = hasFrameMetricsAggregatorClass();
-    if (hasFrameMetricsAggregator) {
-      frameMetricsAggregator = new FrameMetricsAggregator();
-    }
+    this.configResolver = configResolver;
+    this.frameMetricsAggregator = frameMetricsAggregator;
   }
 
   public synchronized void registerActivityLifecycleCallbacks(Context context) {
@@ -142,7 +151,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
 
   @Override
   public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
-    if (isScreenTraceSupported(activity) && configResolver.isPerformanceMonitoringEnabled()) {
+    if (isScreenTraceSupported() && configResolver.isPerformanceMonitoringEnabled()) {
       if (activity instanceof FragmentActivity) {
         FragmentActivity fragmentActivity = (FragmentActivity) activity;
         fragmentActivity
@@ -159,7 +168,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
 
   @Override
   public synchronized void onActivityStarted(Activity activity) {
-    if (isScreenTraceSupported(activity) && configResolver.isPerformanceMonitoringEnabled()) {
+    if (isScreenTraceSupported() && configResolver.isPerformanceMonitoringEnabled()) {
       // Starts recording frame metrics for this activity.
       /**
        * TODO: Only add activities that are hardware acceleration enabled so that calling {@link
@@ -306,7 +315,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
   /** Stops screen trace right after onPause because of b/210055697 */
   @Override
   public void onActivityPostPaused(@NonNull Activity activity) {
-    if (isScreenTraceSupported(activity)) {
+    if (isScreenTraceSupported()) {
       sendScreenTrace(activity);
     }
   }
@@ -427,10 +436,9 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
   /**
    * Only send screen trace if FrameMetricsAggregator exists.
    *
-   * @param activity The Activity for which we're monitoring the screen rendering performance.
    * @return true if supported, false if not.
    */
-  private boolean isScreenTraceSupported(Activity activity) {
+  protected boolean isScreenTraceSupported() {
     return hasFrameMetricsAggregator;
   }
 
@@ -439,7 +447,7 @@ public class AppStateMonitor implements ActivityLifecycleCallbacks {
    * updated to 26.1.0 (b/69954793), there will be ClassNotFoundException. This method is to check
    * if FrameMetricsAggregator exists to avoid ClassNotFoundException.
    */
-  private boolean hasFrameMetricsAggregatorClass() {
+  private static boolean hasFrameMetricsAggregatorClass() {
     try {
       Class<?> initializerClass = Class.forName(FRAME_METRICS_AGGREGATOR_CLASSNAME);
       return true;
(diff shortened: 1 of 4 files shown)


firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -0,0 +1,86 @@
...
+    doAnswer((Answer<Timer>) invocationOnMock -> new Timer(currentTime)).when(clock).getTime();
+
+    DeviceCacheManager.clearInstance();
+    ConfigResolver.clearInstance();
+
+    ConfigResolver configResolver = ConfigResolver.getInstance();
+    configResolver.setDeviceCacheManager(new DeviceCacheManager(new FakeDirectExecutorService()));
+  }
+
+  @Test
```
visumickey, line 76: Can we add more test cases for this? This is an important part of validation and it would better to be exhaustive about the usecases we check here. Some scenarios that come to my mind:

1. Trace creation and stop for a particular fragment. What if we request for a trace again for the same fragment?
2. How do we verify if the hashmap for the fragment to trace is cleaned up as a part of the activity transition?
3. How do we handle scenarios where the length of the trace name for the fragment is way too high?
4. What happens if the fragment contains characters other than alphabets?

Think of more exhaustive usescases to make sure they are covered in unit tests as well.

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
```diff
@@ -17,28 +17,27 @@
...
+import com.google.firebase.perf.metrics.Trace;
 import com.google.firebase.perf.transport.TransportManager;
 import com.google.firebase.perf.util.Clock;
 import com.google.firebase.perf.util.Constants;
+import java.util.WeakHashMap;
 
 public class FragmentStateMonitor extends FragmentManager.FragmentLifecycleCallbacks {
   private static final AndroidLogger logger = AndroidLogger.getInstance();
-  private final FragmentActivity activity;
+  private final WeakHashMap<Fragment, Trace> fragmentToTraceMap = new WeakHashMap<>();
```
visumickey, line 30: Having a map is a good idea. Can you validate few scenarios for this map when there are recycler views in visibility which can cause more fragments to be created. Are we cleaning up the hashmap at such scale is worth accounting for.

Especially scenarios where the recycler view has fragments of which some of them are in visibility and some of them are not (but already created and ready).

leotianlizhan: androidx ViewPager2 is using Fragments inside of a RecyclerView. If needed, I think we should use that to test this behavior.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
leotianlizhan: > 1. Trace creation and stop for a particular fragment. What if we request for a trace again for the same fragment?  

Could you elaborate on this?
> 2. How do we verify if the hashmap for the fragment to trace is cleaned up as a part of the activity transition?  

This would not be an unit test but an integration test, because it involves everything to be real (no Activity, Fragment mocks). 
> 3, 4 

These can be added as unit tests.

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
jeremyjiang-dev: In theory, every activity will have one fragmentStateMonitor, which will hold all the fragments in the map. The map will be cleaned up after transitioning to another activity. Therefore, I think the map size will not be very large.

leotianlizhan: @visumickey I feel like this is a non-blocking investigation that can be parallelized: I want to create a new buganizer to track this investigation, then just finish the current PR with unit tests, then come back to investigate. Would that be okay?

visumickey: I agree, it is a non-blocking investigation and we can always course correct if needed. For now, lets unblock ourselves and move ahead on this.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
leotianlizhan: Great idea on usecase 2 simulation, I feel like I could add a bunch of tests to previous fixes I've made using that idea.

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
leotianlizhan: created b/226189464 to tracking this

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -0,0 +1,173 @@
...
+  @Mock private AppCompatActivity mockActivity;
+  @Mock private AppCompatActivity mockActivityB;
+  @Mock private AppStateMonitor appStateMonitor;
+  @Mock private FrameMetricsAggregator fma;
+
+  @Captor private ArgumentCaptor<TraceMetric> argTraceMetric;
+
+  private long currentTime = 0;
+  private static final String longFragmentName =
+      "_st_NeverGonnaGiveYouUpNeverGonnaLetYouDownNeverGonnaRunAroundAndDesertYouNeverGonnaMakeYouCryNeverGonnaSayGoodbyeNeverGonnaTellALieAndHurtYou";
```
jeremyjiang-dev, line 68: I like this super long name :)

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -0,0 +1,173 @@
...
+        spy(new AppStateMonitor(mockTransportManager, clock, configResolver, fma));
+    FragmentStateMonitor fragmentMonitor =
+        spy(new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, fma));
+    doReturn(true).when(appStateMonitor).isScreenTraceSupported();
+    doReturn(longFragmentName)
+        .when(fragmentMonitor)
+        .getFragmentScreenTraceName(nullable(Fragment.class));
+
+    fragmentMonitor.onFragmentResumed(mockFragmentManager, mockFragment);
+    fragmentMonitor.onFragmentPaused(mockFragmentManager, mockFragment);
```
jeremyjiang-dev, line 147: What are you checking against? Did I miss anything?

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -0,0 +1,173 @@
...
+    FragmentStateMonitor fragmentMonitor =
+        new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, fma);
+    doReturn(true).when(appStateMonitor).isScreenTraceSupported();
+    // Activity_A onCreate registers FragmentStateMonitor, then:
+    appStateMonitor.onActivityStarted(mockActivity);
+    appStateMonitor.onActivityResumed(mockActivity);
+    fragmentMonitor.onFragmentResumed(mockFragmentManager, mockFragment);
+    appStateMonitor.onActivityPaused(mockActivity);
+    // reset() was not called at the time of fragments collecting its frame metrics
+    verify(fma, times(0)).reset();
```
jeremyjiang-dev, line 166: Test fma.remove() as well?

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
leotianlizhan: No you didn't miss anything, it's still WIP

leotianlizhan: Finished adding checks now.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -428,10 +436,9 @@ private void sendSessionLog(String name, Timer startTime, Timer endTime) {
   /**
    * Only send screen trace if FrameMetricsAggregator exists.
    *
-   * @param activity The Activity for which we're monitoring the screen rendering performance.
    * @return true if supported, false if not.
    */
-  private boolean isScreenTraceSupported(Activity activity) {
+  protected boolean isScreenTraceSupported() {
```
visumickey, line 441: Can FrameMetricsAggregator not exist? It seems that we are forcing to create one as a part of the initializer and so, curious how this helps.

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
```diff
@@ -60,12 +60,44 @@ public void onFragmentResumed(@NonNull FragmentManager fm, @NonNull Fragment f)
...
+    fragmentToTraceMap.put(f, fragmentTrace);
   }
 
   @Override
   public void onFragmentPaused(@NonNull FragmentManager fm, @NonNull Fragment f) {
     super.onFragmentPaused(fm, f);
     // Stop Fragment screen trace
     logger.debug("FragmentMonitor %s.onFragmentPaused ", f.getClass().getSimpleName());
+    if (!fragmentToTraceMap.containsKey(f)) {
+      logger.error(
```
visumickey, line 86: Is this an error or a warning? In this case, I don't think the developer can act on that error. So, I guess this qualifies for a warning.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
leotianlizhan: I think this is checking if the Class itself exists, because the class itself was only added in later Android versions. I only removed the parameter because it is not used at all, I assume it was used before for some other code that got removed.

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
leotianlizhan: Changed to `logger.warn`

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
visumickey: Sounds good.

visumickey (approved):

PR #3576, opened 2022-03-21
Revert "Add Firebase App Distribution to smoke tests"

rachaprince:
Reverts firebase/firebase-android-sdk#3555

Because app-distro has not been included in the BoM, it is causing smoke test failures for some other products. This removes app-distribution from smoke test temporarily. We will add it back when the product is out of beta (and added to the BoM).


yifanyang (approved):
Thanks!

PR #3579, opened 2022-03-22
Log notification open after onActivityCreated() on Android 7.1.

gsakakihara:
https://github.com/firebase/firebase-android-sdk/pull/3432
* Also included delaying checking for notification open logging on Android 7.1 (N MR1/25) to avoid Bundle unparceling race condition.


James201311 (approved):
LGTM

PR #3581, opened 2022-03-23
Extending test-app to have buttons for all sdk methods

mannyjimenez0810:


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -26,108 +34,200 @@ class MainActivity : AppCompatActivity() {
...
+                   .setNeutralButton("Okay") { dialog, _ ->
+                       dialog.dismiss()
+                   }.show()
+           }
+       }
+        }
+
+        updateIfNewReleaseAvailableButtonBackground.setOnClickListener {
+            executorService.execute {
+                Thread.sleep(10000)
```
rachaprince, line 68: why is there a thread sleep here, but not in the other background threads?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -26,108 +34,200 @@ class MainActivity : AppCompatActivity() {
...
+        val signOutButtonBackground = findViewById<AppCompatButton>(R.id.sign_out2)
+        val checkForUpdateButtonBackground = findViewById<AppCompatButton>(R.id.check_for_update2)
+        val updateAppButtonBackground = findViewById<AppCompatButton>(R.id.update_app2)
+
+        /** Button Listeners */
+        updateIfNewReleaseAvailableButton.setOnClickListener {
+                   if (updateTask == null || updateTask?.isComplete == true) {
+           updateTask = firebaseAppDistribution.updateIfNewReleaseAvailable().addOnFailureListener {
+               val ex = it as FirebaseAppDistributionException1
+               Log.d("FirebaseAppDistribution", "MAINACTIVITY:ERROR ERROR. CODE: " + it.errorCode)
```
rachaprince, line 55: Could we extract this failure listener and add it to all the buttons for testing?

firebase-appdistribution/test-app/src/androidTest/java/com/googletest/firebase/appdistribution/testapp/MainActivityTest.java
```diff
@@ -0,0 +1,2 @@
+package PACKAGE_NAME;public class MainActivityTest {
```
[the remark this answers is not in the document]
mannyjimenez0810: We can I wanted to keep it in since I had so much trouble last time with Android Studio, but will remove.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
mannyjimenez0810: I believe the reason here was we wanted to simulate a thread that wouldn't run right away to see if that would also mess with the functionality. Since we are no longer testing that will remove.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -26,108 +34,200 @@ class MainActivity : AppCompatActivity() {
...
-//               progressBar.setProgress(percentage, true)
-//                progressPercent.text = "$percentage %"
-//            }
-//        }
+        checkForUpdateButton.setOnClickListener {
+            firebaseAppDistribution.checkForNewRelease().addOnSuccessListener {
+                setupUI(isSignedIn =  firebaseAppDistribution.isTesterSignedIn, isUpdateAvailable = it != null, release = it)
+            }
+        }
+        checkForUpdateButtonBackground.setOnClickListener {
```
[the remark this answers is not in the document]
mannyjimenez0810: I assume that you mean above line 118 above the setOnClickListener

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -1,23 +1,59 @@
...
+    var updateTask: Task<Void>? = null
+    var release: AppDistributionRelease? = null
+    val executorService: ExecutorService = Executors.newFixedThreadPool(1)
+    lateinit var signInButton: AppCompatButton
+    lateinit var signOutButton: AppCompatButton
+    lateinit var checkForUpdateButton: AppCompatButton
+    lateinit var updateAppButton: AppCompatButton
+    lateinit var updateIfNewReleaseAvailableButton: AppCompatButton
+    lateinit var updateIfNewReleaseAvailableButtonBackground: AppCompatButton
+    lateinit var signInButtonBackground: AppCompatButton
```
mannyjimenez0810, line 31: Tried not to do this line multiple times. But if we initialize it outside of onCreate or other function then the app just crashes. Happy to put this back as it was before since I don't think we save any lines of code this way.

firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -27,7 +27,7 @@ android {
         versionName "1.0"
         versionCode 1
 
-        testInstrumentationRunner "androidx.test.runnecr.AndroidJUnitRunner"
```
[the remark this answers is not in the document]
mannyjimenez0810: It shall no longer be a bug! This one was a doozy.

rachaprince (approved):

PR #3582, opened 2022-03-23
Fix refresh logic for App Check custom providers.

### LINKED ISSUE #3561: AppCheck Custom Token refreshed in infinite loop

    with the firebase tag.
    google group.

  * Android Studio version: Android Studio Chipmunk Beta 4
  * Firebase Component: Firebase AppCheck (Beta)
  * Component version:16.0.0-beta04

We have implemented a cloud function with a custom appcheck token provider in NodeJS. This JWT token we are exposing our android devices. The JWT token produces a claim with seconds value. The firebase android sdk retrieves the expiration time from the token itself and interprets it as milliseconds. The SDK now thinks the token already expired
#### Steps to reproduce:
1. Implement a custom token provider with NodeJS or generate a  JWT token with seconds in the claim "exp" and "iat" in seconds. 
2.  Return this token in your AppCheckProvider wtth the signature  `fun getToken(): Task<AppCheckToken>` and add a log statement
3.  Register the AppCheckProviderFactory
4. Start app and observe the log - it will contain many lines of the log statement of step2 since the sdk thinks the token is expired and the token needs to  be refreshed

#### Relevant Code:
My analysis of the refresh mechanism leads to the following Insight. The problem is actually the function `public void maybeScheduleTokenRefresh(@NonNull AppCheckToken token)` in TokenRefreshManager. It ignores the getExpireTimeMillis() in the AppCheckToken and calls instead `public static DefaultAppCheckToken constructFromRawToken(@NonNull String token)`
```
class MyAppCheckToken(
    private val token: String,
    private val expiration: Long
) : AppCheckToken() {
    override fun getToken(): String {
        return token
    }

    override fun getExpireTimeMillis(): Long {
        return expiration
    }
}

class MyCheckProvider(private val firebaseApp: FirebaseApp) : AppCheckProvider {
    override fun getToken(): Task<AppCheckToken> {
        val taskCompletionSource = TaskCompletionSource<AppCheckToken>()
        /
        try {
            // call to myService to get token from NodeJS Cloudfunction via Admin SDK
            val token = getJwtFromNodeJsAdminSdk()
            // Log Statement
            Timber.d("token recieved")
            taskCompletionSource.setResult(token)
        } catch (e: IOException) {
            taskCompletionSource.setException(e)
        }
        return taskCompletionSource.task
    }
}

class MyAppCheckProviderFactory private constructor() : AppCheckProviderFactory {
    @NonNull
    override fun create(@NonNull firebaseApp: FirebaseApp): AppCheckProvider {
        return MyAppCheckProvider(firebaseApp)
    }

    companion object {
        private var INSTANCE: MyAppCheckProviderFactory? = null
        val instance: MyAppCheckProviderFactory?
            @JvmStatic
            @NonNull
            get() {
                if (INSTANCE == null) {
                    INSTANCE = MyAppCheckProviderFactory()
                }
                return INSTANCE
            }
    }
}

```

rosalyntan:
Fixes #3561.

### DIFF
--- a/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultAppCheckToken.java
+++ b/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultAppCheckToken.java
@@ -41,6 +41,8 @@ public final class DefaultAppCheckToken extends AppCheckToken {
   @VisibleForTesting static final String ISSUED_AT_KEY = "iat";
   @VisibleForTesting static final String EXPIRATION_TIME_KEY = "exp";
 
+  private static final long ONE_SECOND_MILLIS = 1000L;
+
   // Raw token value
   private final String token;
   // Timestamp in MS at which this token was generated
@@ -79,7 +81,7 @@ public final class DefaultAppCheckToken extends AppCheckToken {
           TokenParser.parseTokenClaims(tokenResponse.getAttestationToken());
       long iat = getLongFromClaimsSafely(claimsMap, ISSUED_AT_KEY);
       long exp = getLongFromClaimsSafely(claimsMap, EXPIRATION_TIME_KEY);
-      expiresInMillis = (exp - iat) * 1000L;
+      expiresInMillis = (exp - iat) * ONE_SECOND_MILLIS;
     }
 
     return new DefaultAppCheckToken(tokenResponse.getAttestationToken(), expiresInMillis);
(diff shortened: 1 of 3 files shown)


appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultAppCheckToken.java
```diff
@@ -138,10 +138,10 @@ public static DefaultAppCheckToken constructFromRawToken(@NonNull String token)
     Map<String, Object> claimsMap = TokenParser.parseTokenClaims(token);
     long iat = getLongFromClaimsSafely(claimsMap, ISSUED_AT_KEY);
     long exp = getLongFromClaimsSafely(claimsMap, EXPIRATION_TIME_KEY);
-    long expiresInMillis = exp - iat;
+    long expiresInMillis = (exp - iat) * 1000L;
```
malcolmdeck, line 141: Nit: if might want 1000L as a constant, as are in tests. But, like, eh - not a big deal as is.

rosalyntan: Done.

PR #3583, opened 2022-03-23
Do not apply SQL LIMIT if a partial index is used.

ehsannas:


ehsannas:
b/226168223 for more details

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
```diff
@@ -186,6 +186,55 @@ public boolean servedByIndex(FieldIndex index) {
...
   }
 
+  /**
+   * Returns whether the given index is a partial index for the current target.
+   *
+   * <p>A partial index is one which does not have a segment for every field filter and every
+   * orderBy in the query.
+   */
+  public boolean isPartialIndex(FieldIndex index) {
+    List<FieldFilter> allFilters = new ArrayList<>(equalityFilters);
```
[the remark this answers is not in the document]
schmidt-sebastian: It looks like the answer is "no" since our explicit orderBy only included the inequality segments, and not all equality segments: https://osscs.corp.google.com/firebase-sdk/firebase-android-sdk/+/master:firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java;l=341

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -446,13 +446,18 @@ private void deleteIndexEntry(Document document, IndexEntry indexEntry) {
...
     List<Object> bindings = new ArrayList<>();
+    boolean hasPartialIndexes = false;
 
     for (Target subTarget : getSubTargets(target)) {
       FieldIndex fieldIndex = getFieldIndex(subTarget);
       if (fieldIndex == null) {
         return null;
       }
 
+      if (!hasPartialIndexes && new TargetIndexMatcher(subTarget).isPartialIndex(fieldIndex)) {
```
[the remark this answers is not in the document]
ehsannas: I've adopted the implementation proposed in #3590

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -369,6 +371,20 @@ public Direction getKeyOrder() {
     return this.orderBys.get(this.orderBys.size() - 1).getDirection();
   }
 
+  /** Returns the number of segments of a perfect index for this target. */
+  public Integer getSegmentCount() {
```
schmidt-sebastian, line 375:
```suggestion
  public int getSegmentCount() {
```

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -369,6 +371,20 @@ public Direction getKeyOrder() {
...
+    Set<FieldPath> fields = new HashSet<>();
+    for (Filter filter : filters) {
+      for (FieldFilter subFilter : filter.getFlattenedFilters()) {
+        fields.add(subFilter.getField());
+      }
+    }
+    for (OrderBy orderBy : orderBys) {
+      fields.add(orderBy.getField());
+    }
+    return fields.size();
```
schmidt-sebastian, line 385: I think we need to handle some exclusions:

```
query.where('a', '>', 42).orderBy('a')
```

and

```
query.where('a', '>', 42).where('a', '<', 1337)
```

both have a "perfect" segment count of 1.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -84,17 +85,30 @@
...
   /**
-   * Returns an index that can be used to serve the provided target. Returns {@code null} if no
-   * index is configured.
+   * Returns an index that can be used to serve the provided target, as well as the number of index
+   * segments that matched the target filters and orderBys. Returns {@code null} if no index is
+   * configured.
    */
   @Nullable
-  FieldIndex getFieldIndex(Target target);
+  Pair<FieldIndex, Integer> getFieldIndexAndSegmentCount(Target target);
```
schmidt-sebastian, line 93: This is probably a personal problem that points to some deeper issue about my adolescence, but I really do not like APIs that return `Pair`. I completely understand why you choose to use a `Pair` here and am not suggesting you should change it, but `Pair`s are very nondescriptive and you essentially had to make up for this with documentation. You could achieve a more concise result by returning a wrapper type (albeit at the cost of code size)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -110,17 +111,21 @@ public void initialize(LocalDocumentsView localDocumentsView, IndexManager index
       return null;
     }
 
-    List<DocumentKey> keys = indexManager.getDocumentsMatchingTarget(target);
-    if (keys == null) {
+    Pair<List<DocumentKey>, Boolean> indexResults = indexManager.getDocumentsMatchingTarget(target);
```
schmidt-sebastian, line 114: Is it possible to handle all "usedFullIndex"/limit logic in QueryEngine? Prior to running the query we could possibly determine whether a full index is used and if so, we could drop the limit. 

Maybe we could have something like:

```
enum IndexStatus {
   None,
   Partial,
   Full
}

IndexStatus status = indexManager.canServerUsingIndex(target);

if (status.equals(None)) {
  return null;
}

if (status.equals(Partial)) { 
  // Witty yet descriptive comment
  query = query.limitToFirst(-1);
}
```

firebase-firestore/src/test/java/com/google/firebase/firestore/model/TargetIndexMatcherTest.java
```diff
@@ -570,13 +569,13 @@ private void validateServesTarget(
       Query query, String field, FieldIndex.Segment.Kind kind, Object... fieldsAndKind) {
     FieldIndex expectedIndex = fieldIndex("collId", field, kind, fieldsAndKind);
     TargetIndexMatcher targetIndexMatcher = new TargetIndexMatcher(query.toTarget());
-    assertTrue(targetIndexMatcher.servedByIndex(expectedIndex));
+    assertTrue(targetIndexMatcher.servedByIndex(expectedIndex) > 0);
```
schmidt-sebastian, line 572: assertGreaterThan?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
schmidt-sebastian: FWIW, it also seems like we don't actually use this method in the public API anymore, so we could remove the public declaration and just use `Pair` in SQLiteIndexManager.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
ehsannas: What do you think?

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
ehsannas: My assumption was that if an index can be used to serve a target and the number of index segments equals the number of unique FieldPaths in the target's filters&orderBys, then it is the perfect index for it.

I'm a bit concerned whether this is the right assumption; but I cannot think of any case where this is incorrect. Any thoughts?

For the first case `a>42 orderby a`, the "perfect" segment count of 1 sounds good to me (`A ASC`)
For the second case, it has multiple inequalities and I don't know how that should be served from the index. In theory, the `A ASC` can be used two times to compute the results.

schmidt-sebastian: I think we need to count the number of unique field paths that are used in both `orderBy` and `where`. We do need to special case array-contains and array-contains-any filters, which would require an additional `IndexKind.CONTAINS` segment that is independent from any ordering.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
schmidt-sebastian: It's hard to say where to draw the line here. Your current approach has some benefits in it keeps the logic self contained for SQLite. We can add a different implementation (for the memory cache) for example that handles these queries differently. 
Adding this logic to the Query engine makes this part of the logic part of the main query evaluation path. It brings the concepts of "Partial" indexes from SQLite to the SDK as a hole and it creates a common way to execute limit queries with partial targets. I do think that even a memory-based indexing implementation will run into the same problem, and thus to me it makes sense to generalize this logic and have it in QueryEngine. But I cannot predict the future.

ehsannas: I see your point. I've moved the logic to QueryEngine

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
```diff
@@ -369,6 +373,29 @@ public Direction getKeyOrder() {
...
+          hasArraySegment = true;
+        } else {
+          fields.add(subFilter.getField());
+        }
+      }
+    }
+    for (OrderBy orderBy : orderBys) {
+      fields.add(orderBy.getField());
+    }
+    return fields.size() + (hasArraySegment ? 1 : 0);
```
schmidt-sebastian, line 396: We need to figure out or story for __name__. Do you think we should enforce that indices end with __name__ or should we just assume it is always there? I thought we were already ignoring any trailing __name__ elements when an index is configured, which means that we would have to ignore the element here as well. We are, however, not doing this right now so the user would have to end every index with a __name__ component. Let me know if you want to talk about this, but I think it is important to come up with a decision before proceeding much further.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -32,6 +32,21 @@
  * Collection Group queries.
  */
 public interface IndexManager {
+  /** Represents the index state as it relates to a particular target. */
+  public enum IndexStatus {
```
schmidt-sebastian, line 36: I don't have a suggestion, but we should see if there is a different name since we already have IndexState which stores how up to date an index is.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -110,11 +110,27 @@ public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
...
     }
 
+    if (indexStatus.equals(IndexManager.IndexStatus.PARTIAL)) {
+      // We cannot apply a limit for targets that are served using a partial index.
+      // If a partial index will be used to serve the target, the query may return a superset of
+      // documents that match the target (e.g. if the index doesn't include all the target's
+      // filters), or may return the correct set of documents in the wrong order (e.g. if the index
+      // doesn't include a segment for one of the orderBys). Therefore a limit should not be applied
+      // in such cases.
+      query = query.limitToFirst(-1);
```
schmidt-sebastian, line 127: Can you use Target.NO_LIMIT?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -83,12 +98,8 @@ public interface IndexManager {
...
   IndexOffset getMinOffset(String collectionGroup);
 
-  /**
-   * Returns an index that can be used to serve the provided target. Returns {@code null} if no
-   * index is configured.
-   */
-  @Nullable
-  FieldIndex getFieldIndex(Target target);
+  /** Returns the type of index (if any) that can be used to serve the given target */
+  IndexStatus canServeUsingIndex(Target target);
```
schmidt-sebastian, line 102: I think this is not a good name - it would be if we also passed the Index, but we don't. Maybe something like "hasIndex()"/"hasBackingIndex()"? 

Optional.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
```diff
@@ -524,7 +524,7 @@ public void testArrayContainsDoesNotMatchNonArray() {
   public void testNoMatchingFilter() {
     setUpSingleValueFilter();
     Query query = query("coll").filter(filter("unknown", "==", true));
-    assertNull(indexManager.getFieldIndex(query.toTarget()));
+    assertNull(((SQLiteIndexManager) indexManager).getFieldIndexAndSegmentCount(query.toTarget()));
```
schmidt-sebastian, line 527: Use the new canServeUsingIndex API?

schmidt-sebastian (commented):
Thanks! This looks great. My comments are all just nits at this point.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Target.java
ehsannas: yeah, I think all indexes should have an _implicit_ order by `__name__`, and if the user explicitly adds a segment for it, we can drop it and always assume it implicitly exists.

I will update this code to skip the `__name__` fields.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -83,12 +98,8 @@ public interface IndexManager {
...
   IndexOffset getMinOffset(String collectionGroup);
 
-  /**
-   * Returns an index that can be used to serve the provided target. Returns {@code null} if no
-   * index is configured.
-   */
-  @Nullable
-  FieldIndex getFieldIndex(Target target);
+  /** Returns the type of index (if any) that can be used to serve the given target */
+  TargetIndexType hasIndex(Target target);
```
schmidt-sebastian, line 102: I think this is fine. `IndexType getIndexType(Target target)` would be even finer :)

schmidt-sebastian (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
ehsannas: Done

PR #3588, opened 2022-03-24
Fireperf fragments: sampling

leotianlizhan:
b/225186908

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
@@ -104,7 +104,10 @@ final class RateLimiter {
     return samplingBucketId < validNetworkSamplingBucketIdThreshold;
   }
 
-  /** Returns whether device is allowed to send trace events based on trace sampling rate. */
+  /**
+   * Returns whether device is allowed to send Fragment screen trace events based on Fragment screen
+   * trace sampling rate.
+   */
   private boolean isDeviceAllowedToSendFragmentScreenTraces() {
     float validFragmentSamplingBucketIdThreshold = configResolver.getFragmentSamplingRate();
     return samplingBucketId < validFragmentSamplingBucketIdThreshold;
@@ -112,7 +115,9 @@ final class RateLimiter {
 
   /** Identifies if the {@link PerfMetric} is a Fragment screen trace */
   private boolean isFragmentScreenTrace(PerfMetric metric) {
-    return metric.hasTraceMetric() && metric.getTraceMetric().getName().startsWith(Constants.SCREEN_TRACE_PREFIX) && metric.getTraceMetric().containsCustomAttributes(Constants.ACTIVITY_ATTRIBUTE_KEY);
+    return metric.hasTraceMetric()
+        && metric.getTraceMetric().getName().startsWith(Constants.SCREEN_TRACE_PREFIX)
+        && metric.getTraceMetric().containsCustomAttributes(Constants.ACTIVITY_ATTRIBUTE_KEY);
   }
 
   /**
--- a/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
@@ -104,6 +104,17 @@ final class RateLimiter {
     return samplingBucketId < validNetworkSamplingBucketIdThreshold;
   }
 
+  /** Returns whether device is allowed to send trace events based on trace sampling rate. */
+  private boolean isDeviceAllowedToSendFragmentScreenTraces() {
+    float validFragmentSamplingBucketIdThreshold = configResolver.getFragmentSamplingRate();
+    return samplingBucketId < validFragmentSamplingBucketIdThreshold;
+  }
+
+  /** Identifies if the {@link PerfMetric} is a Fragment screen trace */
+  private boolean isFragmentScreenTrace(PerfMetric metric) {
+    return metric.hasTraceMetric() && metric.getTraceMetric().getName().startsWith(Constants.SCREEN_TRACE_PREFIX) && metric.getTraceMetric().containsCustomAttributes(Constants.ACTIVITY_ATTRIBUTE_KEY);
+  }
+
   /**
    * Check if the {@link PerfMetric} should be rate limited.
    *
@@ -136,6 +147,12 @@ final class RateLimiter {
   boolean isEventSampled(PerfMetric metric) {
     if (metric.hasTraceMetric()
         && !(isDeviceAllowedToSendTraces()
+            || hasVerboseSessions(metric.getTraceMetric().getPerfSessionsList()))) {
+      return false;
+    }
+
+    if (isFragmentScreenTrace(metric)
+        && !(isDeviceAllowedToSendFragmentScreenTraces()
             || hasVerboseSessions(metric.getTraceMetric().getPerfSessionsList()))) {
       return false;
     }


firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
```diff
@@ -398,6 +398,67 @@ public void testDeviceSampling_tracesDisabledButNetworkEnabled() {
     assertThat(limiter.getIsDeviceAllowedToSendTraces()).isFalse();
   }
 
+  @Test
+  public void testDeviceSampling_tracesEnabledButFragmentDisabled() {
```
jeremyjiang-dev, line 402: nit: change the name to `testDeviceSampling_tracesEnabledButFragmentDisabled_dropsFragmentTrace`. A good name describes both the actions that are being taken on a system and the expected outcome.

go/unit-testing-practices

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
```diff
@@ -398,6 +398,67 @@ public void testDeviceSampling_tracesDisabledButNetworkEnabled() {
...
+                TraceMetric.newBuilder()
+                    .setName("_st_TestActivity")
+                    .addAllPerfSessions(Arrays.asList(createNonVerbosePerfSessions())))
+            .build();
+    assertThat(limiter.isFragmentScreenTrace(activityTrace)).isFalse();
+    assertThat(limiter.isEventSampled(activityTrace)).isTrue();
+  }
+
+  @Test
+  public void testDeviceSampling_tracesDisabledButFragmentEnabled() {
```
jeremyjiang-dev, line 438: nit: add expected behavior to the test name

jeremyjiang-dev:
LGTM on check-changed is green

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
```diff
@@ -824,6 +922,32 @@ public void testNetworkRequestsAreNotSampledWhenSessionIsVerboseAndSamplingEnabl
     assertThat(limiter.isEventSampled(network)).isTrue();
   }
 
+  @Test
```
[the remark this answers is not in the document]
leotianlizhan: Done

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
leotianlizhan: Done

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
leotianlizhan: Done

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
```diff
@@ -104,6 +104,22 @@ private boolean isDeviceAllowedToSendNetworkEvents() {
     return samplingBucketId < validNetworkSamplingBucketIdThreshold;
   }
 
+  /**
+   * Returns whether device is allowed to send Fragment screen trace events based on Fragment screen
+   * trace sampling rate.
+   */
+  private boolean isDeviceAllowedToSendFragmentScreenTraces() {
+    float validFragmentSamplingBucketIdThreshold = configResolver.getFragmentSamplingRate();
+    return samplingBucketId < validFragmentSamplingBucketIdThreshold;
```
visumickey, line 113: Is this the same samplingBucketId used for traces? If so, what are the pros and cons in using a different sampling bucketId? I was biased towards creating a new sampling bucketId for fragment traces. But checking if there are any negates around that.

leotianlizhan: Good point. I used the same bucketId because Network and Trace shared the same one as well. But after thinking about this I think we need to change to different bucketId like you said. 

I don't think it makes a big difference in the case of Network and Trace because they are independent, but since fragment is dependent there's a big problem with using the same bucketId.

#### Suppose we use the same bucketId, and suppose trace sampling rate is 40% and fragment sampling rate is 80%.
Then 40% of all devices will be sampled to send traces. However, within that 40%, all devices will send fragment traces, because 0.8 > 0.4, and we are using the same bucketId, so rolling the dice only once. If the dice already landed < 0.4, then it must also be < 0.8. 

```
0    0.1    0.2    0.3    0.4    0.5    0.6    0.7    0.8    0.9    1
             *             |
             *                                         |
```

#### In the independent case, like network and trace, suppose trace sampling rate is 40% and network is 80%.
```
0    0.1    0.2    0.3    0.4    0.5    0.6    0.7    0.8    0.9    1
|         both             |
|         network                                      |
                                                       |     none   |
```

visumickey: Thanks for the exploration Leo. Like we discussed offline, we should go with two separate bucket ids for traces and fragment screen traces. This would enable us to avoid the situation where fragment screen trace sampling flag becomes unusable.

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
```diff
@@ -63,7 +65,13 @@
    * @param capacity token bucket capacity
    */
   public RateLimiter(@NonNull Context appContext, final Rate rate, final long capacity) {
-    this(rate, capacity, new Clock(), getSamplingBucketId(), ConfigResolver.getInstance());
+    this(
+        rate,
+        capacity,
+        new Clock(),
+        getSamplingBucketId(),
```
jeremyjiang-dev, line 72: I think for `getSamplingBucketId()` we need to feed date time as the seed of the Random generator

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
jeremyjiang-dev: Is the test name updated?

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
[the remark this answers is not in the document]
jeremyjiang-dev: Is the test name updated? Could you check all newly added tests to see if they are following the naming convention?

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
visumickey: The `getSamplingBucketId` is a pure random number [Reference](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java#L72-L74). I don't think we need a specific seed.

Am I missing something here?

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
```diff
@@ -78,11 +86,16 @@ static float getSamplingBucketId() {
       final long capacity,
       final Clock clock,
       float samplingBucketId,
+      float fragmentBucketId,
       ConfigResolver configResolver) {
     Utils.checkArgument(
-        0.0f <= samplingBucketId && samplingBucketId < 1.0f,
+        0.0f <= samplingBucketId
```
visumickey, line 92: Recommend separating these checks. Makes it easy to debug with separate log statements (One for trace bucket ID and one for fragment).

visumickey (approved):
LGTM after addressing the unit test comments.

I'm not fully aligned on adding a new parameter to the RateLimiter constructor. This means that for every new metric, we would end up adding a new sampling rate parameter to the rate limiter. But, we are aligning our current implementation with the existing options. So, fine with it for now. But we need to find a better strategy to organize the rate limiter.

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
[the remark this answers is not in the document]
leotianlizhan: Yes, it should be now `isEventSampled_fragmentWithVerboseSessionEnabled_returnsTrue`

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
leotianlizhan: I don't think that's needed. From java.util.random documentation:

> public Random()
Creates a new random number generator. This constructor sets the seed of the random number generator to a value very likely to be distinct from any other invocation of this constructor.

jeremyjiang-dev: Got it! Sounds good!

jeremyjiang-dev (approved):

firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
leotianlizhan: yes, all newly added tests should be using the naming convention

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
leotianlizhan: Done

PR #3590, opened 2022-03-25
Different partial index implementation

schmidt-sebastian:


firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -447,12 +448,17 @@ private void deleteIndexEntry(Document document, IndexEntry indexEntry) {
...
     for (Target subTarget : getSubTargets(target)) {
-      FieldIndex fieldIndex = getFieldIndex(subTarget);
-      if (fieldIndex == null) {
+      Pair<FieldIndex, Integer> pair = doMagic(subTarget);
+      if (pair == null) {
         return null;
       }
 
+      FieldIndex fieldIndex = pair.first;
+      isFullIndex |= fieldIndex.getSegments().size() == pair.second;
```
ehsannas, line 460: `fieldIndex.getSegments().size() == pair.second`
is basically
`pair.first.getSegments().size() == pair.second`

isn't this always `true` ?

schmidt-sebastian: Hm... Do we need a `Target.getSegmentCount()`? I am trying to remove some of the double computation, but maybe I am off on the wrong path.

ehsannas: yeah that's a good idea. I'll adopt this implementation.

PR #3592, opened 2022-03-25
Add frame metrics to fragment traces

jeremyjiang-dev:
The original PR with comments can be found #3584

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
@@ -27,12 +27,12 @@ import com.google.firebase.perf.util.Constants;
  * @hide
  */
 public class FrameMetricsCalculator {
-  public static class FrameMetrics {
+  public static class PerfFrameMetrics {
     int totalFrames = 0;
     int slowFrames = 0;
     int frozenFrames = 0;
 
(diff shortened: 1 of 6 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.firebase.perf.util.Constants;
+
+/**
+ * FrameMetricsCalculator helps calculate total frames, slow frames, and frozen frames from metrics
+ * collected by {@link FrameMetricsAggregator}
+ *
+ * @hide
+ */
+public class FrameMetricsCalculator {
+  public static class FrameMetrics {
```
visumickey, line 30: COuld this be namespaced to something with a "Perf prefix"? FrameMetrics is an Android class name as well. [Reference](https://developer.android.com/reference/android/view/FrameMetrics)

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
jeremyjiang-dev: Done!

jeremyjiang-dev:
Merging the PR. The smoke test failure is related to firestore.

PR #3594, opened 2022-03-28
Adding unit tests to ApkUpdater class

mannyjimenez0810:
ApkUpdater class was missing some unit test coverage on classes such as downloadToDisk and updateApk where we were not testing some paths of success.

Added in testing to verify all of the notifications used in the ApkUpdater class. 
Test the happy path of the updateApk method going through download apk -> makeApkDownloadRequest -> downloadToDisk.

As well as testing the fail case of validateJarFile.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
@@ -246,6 +246,7 @@ class ApkUpdater {
     return bytesDownloaded;
   }
 
+  @VisibleForTesting
   void validateJarFile(File apkFile, long totalSize, boolean showNotification, long bytesDownloaded)
       throws FirebaseAppDistributionException {
     try {
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
@@ -84,6 +84,7 @@ class ApkUpdater {
     this.appDistributionNotificationsManager = appDistributionNotificationsManager;
     this.httpsUrlConnectionFactory = httpsUrlConnectionFactory;
     this.lifeCycleNotifier = lifeCycleNotifier;
+    cachedUpdateTask = new UpdateTaskImpl();
   }
 
   UpdateTaskImpl updateApk(
@@ -159,7 +160,8 @@ class ApkUpdater {
     return downloadTaskCompletionSource.getTask();
   }
 
-  private void makeApkDownloadRequest(
+  @VisibleForTesting
+  void makeApkDownloadRequest(
       @NonNull AppDistributionReleaseInternal newRelease, boolean showNotification)
       throws FirebaseAppDistributionException {
     String downloadUrl = newRelease.getDownloadUrl();
@@ -203,7 +205,8 @@ class ApkUpdater {
     return responseCode >= 200 && responseCode < 300;
   }
 
-  private long downloadToDisk(
+  @VisibleForTesting
+  long downloadToDisk(
       HttpsURLConnection connection, long totalSize, String fileName, boolean showNotification)
       throws FirebaseAppDistributionException {
     context.deleteFile(fileName);
@@ -246,8 +249,8 @@ class ApkUpdater {
     return bytesDownloaded;
   }
 
-  private void validateJarFile(
-      File apkFile, long totalSize, boolean showNotification, long bytesDownloaded)
+  @VisibleForTesting
+  void validateJarFile(File apkFile, long totalSize, boolean showNotification, long bytesDownloaded)
       throws FirebaseAppDistributionException {
     try {
       new JarFile(apkFile).close();
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
@@ -84,7 +84,6 @@ class ApkUpdater {
     this.appDistributionNotificationsManager = appDistributionNotificationsManager;
     this.httpsUrlConnectionFactory = httpsUrlConnectionFactory;
     this.lifeCycleNotifier = lifeCycleNotifier;
-    cachedUpdateTask = new UpdateTaskImpl();
   }
 
   UpdateTaskImpl updateApk(
@@ -96,7 +95,6 @@ class ApkUpdater {
 
       cachedUpdateTask = new UpdateTaskImpl();
     }
-
     downloadApk(newRelease, showNotification)
         .addOnSuccessListener(taskExecutor, file -> installApk(file, showNotification))
         .addOnFailureListener(
@@ -204,8 +202,7 @@ class ApkUpdater {
     return responseCode >= 200 && responseCode < 300;
   }
 
-  @VisibleForTesting
-  long downloadToDisk(
+  private long downloadToDisk(
       HttpsURLConnection connection, long totalSize, String fileName, boolean showNotification)
       throws FirebaseAppDistributionException {
     context.deleteFile(fileName);
@@ -248,7 +245,6 @@ class ApkUpdater {
     return bytesDownloaded;
   }
 
-  @VisibleForTesting
   void validateJarFile(File apkFile, long totalSize, boolean showNotification, long bytesDownloaded)
       throws FirebaseAppDistributionException {
     try {
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
@@ -160,8 +160,7 @@ class ApkUpdater {
     return downloadTaskCompletionSource.getTask();
   }
 
-  @VisibleForTesting
-  void makeApkDownloadRequest(
+  private void makeApkDownloadRequest(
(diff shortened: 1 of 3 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
```diff
@@ -84,6 +84,7 @@ public ApkUpdater(
     this.appDistributionNotificationsManager = appDistributionNotificationsManager;
     this.httpsUrlConnectionFactory = httpsUrlConnectionFactory;
     this.lifeCycleNotifier = lifeCycleNotifier;
+    cachedUpdateTask = new UpdateTaskImpl();
```
mannyjimenez0810, line 87: I am unsure why we were not initializing our update tasks in the constructor, but I added it here since it was needed for a test.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
```diff
@@ -203,7 +205,8 @@ private static boolean isResponseSuccess(int responseCode) {
     return responseCode >= 200 && responseCode < 300;
   }
 
-  private long downloadToDisk(
+  @VisibleForTesting
```
mannyjimenez0810, line 207: Wasn't sure how to test this without this annotation but seemed like the best case without a big refactor.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
```diff
@@ -238,4 +248,37 @@ public void updateApp_whenCalledMultipleTimesWithApk_returnsSameUpdateTask() {
...
+    List<UpdateProgress> progressEvents = new ArrayList<>();
+    updateTask.addOnProgressListener(testExecutor, progressEvents::add);
+
+    assertThat(progressEvents).hasSize(2);
+    assertThat(progressEvents.get(0).getUpdateStatus()).isEqualTo(UpdateStatus.PENDING);
+    assertThat(progressEvents.get(1).getUpdateStatus()).isEqualTo(UpdateStatus.DOWNLOADED);
+  }
+
+  @Test
+  public void downloadToDisk_whenHasInputStream_notificationSetCorrectlyHasCorrectInputLength()
```
mannyjimenez0810: Added two tests.

 One to check the notifications for makeApkDownloadRequest through updateApk. 

Another to test the read and writing for the downloadToDisk method. And one on the negative path making sure the notification set.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
```diff
@@ -170,6 +183,52 @@ public void updateApk_whenInstallSuccessful_setsResult() throws Exception {
     assertThat(updateTask.isSuccessful()).isTrue();
```
rachaprince, line 183: [Not directly related to this PR, since this code was written beforehand] 

The test above (updateApk_whenInstallSuccessful_setsResult) seems like something we should remove. We can't set the result on a successful install because the process is killed

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
```diff
@@ -170,6 +183,52 @@ public void updateApk_whenInstallSuccessful_setsResult() throws Exception {
...
+
+    assertThat(events).hasSize(3);
+    assertThat(events.get(0).getUpdateStatus()).isEqualTo(UpdateStatus.PENDING);
+    assertThat(events.get(1).getUpdateStatus()).isEqualTo(UpdateStatus.DOWNLOADING);
+    assertThat(events.get(1).getApkBytesDownloaded()).isEqualTo(TEST_FILE.length());
+    assertThat(events.get(2).getUpdateStatus()).isEqualTo(UpdateStatus.DOWNLOADED);
+  }
+
+  @Test
+  public void updateApk_invalidJarFile_throwsException() throws IOException {
```
lfkellogg, line 210: This test is named `invalidJarFile` but seems to actually test when the file input (or output?) fails.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
```diff
@@ -170,6 +183,52 @@ public void updateApk_whenInstallSuccessful_setsResult() throws Exception {
...
+    doReturn(new ByteArrayInputStream(TEST_FILE.getBytes()))
+        .when(mockHttpsUrlConnection)
+        .getInputStream();
+    when(mockApkInstaller.installApk(any(), any())).thenReturn(Tasks.forResult(null));
+    FileInputStream fileInputStream =
+        (FileInputStream) ApkUpdaterTest.class.getResourceAsStream("file does not exist");
+    doReturn(fileInputStream).when(mockContext).openFileInput(anyString());
```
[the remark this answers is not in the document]
mannyjimenez0810: For validateJarFile it takes in a File. I am unable to mock that method anymore I decided to pass in a file. We get that file from calling openFileInput on the context. So I had to mock the context so that it would return the correct FileinputStream.

rachaprince (approved):

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
mannyjimenez0810: It tests that the tests fails when there is an invalidJarFile. My understanding was methodUnderTest_conditions_expectedEndState. So here the condition would be calling this with an invalidJarFile.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
mannyjimenez0810: Good point! Willing to delete it since I do not believe it adds to our current tests.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
lfkellogg: But it looks like the condition is really more like "the file doesn't exist" (either when reading or writing, I'm not sure). From below: 
```
FileInputStream fileInputStream =
        (FileInputStream) ApkUpdaterTest.class.getResourceAsStream("file does not exist");
```

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
```diff
@@ -246,8 +246,7 @@ private long downloadToDisk(
     return bytesDownloaded;
   }
 
-  private void validateJarFile(
-      File apkFile, long totalSize, boolean showNotification, long bytesDownloaded)
+  void validateJarFile(File apkFile, long totalSize, boolean showNotification, long bytesDownloaded)
```
lfkellogg, line 249: Please add `@VisibleForTesting`

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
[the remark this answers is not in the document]
lfkellogg: I'm confused because I don't see us calling `openFileInput` anywhere. If it's used by something else under the hood (maybe the `JarFile` constructor?) then we should probably leave a comment here explaining that.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
mannyjimenez0810: There is no input or output happening with that fileInputStream.  It is simply so that this line " File apkFile = context.getFileStreamPath(fileName);" Returns an incorrect jarFile so that when we do " new JarFile(apkFile).close();" it throws an error. 

Either way the file not existing leads to an invalid jar file.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
mannyjimenez0810: Actually I can add the private back in since we no longer mock this. I don't believe the visibleForTesting annotation is required.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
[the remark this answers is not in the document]
mannyjimenez0810: O sorry for the confusion that is an unnecessary mock. Thanks for catching that!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/ApkUpdater.java
mannyjimenez0810: good catch!

lfkellogg (approved):

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/ApkUpdaterTest.java
mannyjimenez0810: Just to clarify the reason this fails is because we lack the validateJarFile mock. We don't pass in an invalid string by default the test will pass in an invalid string when it does. "context.getFileStreamPath(fileName);"

PR #3606, opened 2022-03-31
Update the gradle properties to match the EAP release branch

jeremyjiang-dev:


firebase-perf/ktx/gradle.properties
```diff
@@ -1,2 +1,3 @@
 android.enableUnitTestBinaryResources=true
-
+version=20.0.7-eap
```
rlazo, line 2: You can revert the changes to this file. It's only changed by the tool so there wouldn't be any conflict.

jeremyjiang-dev: Done!

rlazo (approved):

PR #3609, opened 2022-03-31
Fix validation bug for composite filters.

ehsannas:


ehsannas:
FYI @cherylEnkidu

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -364,6 +364,75 @@ public void testCanListenForQueryMetadataChanges() {
     listener2.remove();
   }
 
+  /*
+  // TODO(orquery): Re-enable this test once the `where` API is public.
```
schmidt-sebastian, line 368: Can we find a way to include this test in a non-commented out fashion. Ideas include:

- `@Ignore`
- Making the test call into a package-private method, preferably directly but if that is not possible via something like https://cs.opensource.google/firebase-sdk/firebase-android-sdk/+/master:firebase-firestore/src/androidTest/java/com/google/firebase/firestore/AccessHelper.java?q=accesshelper&ss=firebase-sdk

Commented code usually ages poorly.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/CompositeFilter.java
```diff
@@ -28,6 +28,10 @@
   private final List<Filter> filters;
   private final Operator operator;
 
+  // Memoized list of all field filters that can be found by traversing the tree of filters
+  // contained in this composite filter.
+  private List<FieldFilter> flattenedFilters;
```
schmidt-sebastian, line 33: We usually use the "memoized" prefix for these kind of states.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
ehsannas: I realized I could remove the access modifier (rather than using `protected`) to enable testing and keeping the API bot happy.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -364,6 +372,72 @@ public void testCanListenForQueryMetadataChanges() {
...
+        });
+    // Conflicting operations within a composite filter.
+    assertThrows(
+        IllegalArgumentException.class,
+        () -> {
+          Query query =
+              collection.where(
+                  or(
+                      and(equalTo("a", "b"), inArray("c", Arrays.asList("d", "e"))),
+                      and(equalTo("e", "f"), notInArray("c", Arrays.asList("d", "e")))));
```
schmidt-sebastian, line 410: Hm, I don't think we want to go down this level and provide semantic validation. This is just a huge can of worms - I think we should reject "syntactically" invalid queries but not inspect the actual contents of the query operations.

Here and below.

schmidt-sebastian (approved):

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
ehsannas: I think I don't understand what you mean. We're not throwing the error because `in` and `not-in` have the same list values. We're throwing the error because `in` and `not-in` are conflicting operations.

if we see `or(a>1, b>1)`, we should reject it because it has multiple inequalities.
if we see `or(a in [1,2], b not-in [1,2])`, we should reject because of conflicting ops.
right?

schmidt-sebastian: Okay - if the query cannot be run on the backend no mattter what values we are filtering by then yes, we should reject it. Sorry for the misunderstanding (which was mostly driven by the fact that you used the same field values in for `in` and `notIn`.

ehsannas: yeah that's a bit confusing. I've changed the field values in the test

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -364,6 +372,72 @@ public void testCanListenForQueryMetadataChanges() {
     listener2.remove();
   }
 
+  @Test
+  public void testInvalidQueryFilters() {
```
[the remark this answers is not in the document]
ehsannas: I have moved the test to ValidationTest.java.

schmidt-sebastian (approved):
Thanks for the update and for adding the test message assertions.

PR #3622, opened 2022-04-06
SDK support for logical termination in Firestore BatchGetDocuments

cherylEnkidu:
#no-changelog

For BatchGetDocuments, since the SDK knows the number of document paths in the requests, as soon as the SDK receives all the responses needed, it will fire user callback to save user's waiting time.

### DIFF
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
@@ -39,6 +39,7 @@ import com.google.firestore.v1.FirestoreGrpc;
 import io.grpc.Status;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -174,17 +175,16 @@ public class Datastore {
       builder.addDocuments(serializer.encodeKey(key));
     }
     List<BatchGetDocumentsResponse> responses = new ArrayList<>();
-    TaskCompletionSource<List<MutableDocument>> tcs = new TaskCompletionSource<>();
+    TaskCompletionSource<List<MutableDocument>> completionSource = new TaskCompletionSource<>();
 
     channel.runStreamingResponseRpc(
         FirestoreGrpc.getBatchGetDocumentsMethod(),
         builder.build(),
-        new FirestoreChannel.Listener<BatchGetDocumentsResponse>() {
+        new FirestoreChannel.StreamingListener<BatchGetDocumentsResponse>() {
           @Override
           public void onMessage(BatchGetDocumentsResponse message) {
             responses.add(message);
             if (responses.size() == keys.size()) {
-              callback_fired = true;
               Map<DocumentKey, MutableDocument> resultMap = new HashMap<>();
               for (BatchGetDocumentsResponse response : responses) {
                 MutableDocument doc = serializer.decodeMaybeDocument(response);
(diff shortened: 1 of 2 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -170,21 +173,19 @@ WriteStream createWriteStream(WriteStream.Callback listener) {
...
-            task -> {
-              if (!task.isSuccessful()) {
-                if (task.getException() instanceof FirebaseFirestoreException
-                    && ((FirebaseFirestoreException) task.getException()).getCode()
-                        == FirebaseFirestoreException.Code.UNAUTHENTICATED) {
-                  channel.invalidateToken();
-                }
-              }
+    List<BatchGetDocumentsResponse> responses = new ArrayList<>();
+    TaskCompletionSource<List<MutableDocument>> tcs = new TaskCompletionSource<>();
```
schmidt-sebastian, line 177: Unfortunately, we don't really use abbreviations in Java :) You should make this `completionSource`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -170,21 +173,19 @@ WriteStream createWriteStream(WriteStream.Callback listener) {
...
 
+    channel.runStreamingResponseRpc(
+        FirestoreGrpc.getBatchGetDocumentsMethod(),
+        builder.build(),
+        new FirestoreChannel.Listener<BatchGetDocumentsResponse>() {
+          @Override
+          public void onMessage(BatchGetDocumentsResponse message) {
+            responses.add(message);
+            if (responses.size() == keys.size()) {
+              callback_fired = true;
```
schmidt-sebastian, line 187: This should be `camelCase`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -193,8 +194,26 @@ WriteStream createWriteStream(WriteStream.Callback listener) {
...
-            });
+              tcs.setResult(results);
+            }
+          }
+
+          @Override
+          public void onClose(Status status) {
+            if (status.isOk() && !callback_fired) {
+              callback_fired = true;
+              tcs.setResult(new ArrayList<>());
```
schmidt-sebastian, line 205: `Collections.emptyList()`, which just returns a reference to an already allocated object.

schmidt-sebastian: FWIW you can also remove `callback_fired` and invoke `tcs.trySetResult()`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/Datastore.java
```diff
@@ -193,8 +194,26 @@ WriteStream createWriteStream(WriteStream.Callback listener) {
...
+            } else if (!status.isOk()) {
+              FirebaseFirestoreException exception = exceptionFromStatus(status);
+              if (exception.getCode() == FirebaseFirestoreException.Code.UNAUTHENTICATED) {
+                channel.invalidateToken();
+              }
+              tcs.setException(exception);
```
schmidt-sebastian, line 211: trySetException

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -44,6 +42,16 @@
  */
 public class FirestoreChannel {
 
+  public abstract static class Listener<T> {
```
schmidt-sebastian, line 45: Please use a more descriptive name and add a class comment.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreChannel.java
```diff
@@ -44,6 +42,16 @@
...
+    public Listener() {}
+
+    public void onMessage(T message) {}
+
+    public void onClose(Status status) {}
+
+    public boolean callback_fired = false;
```
schmidt-sebastian, line 52: This can be removed (I think), but even if not it should not be part of the Listener interface. You can move it to the callsite if needed.

schmidt-sebastian (commented):
This looks good! Is it possible to add a test that tests that we return a document if we receive a document followed by an error?

cherylEnkidu:
We don't have unit tests for FirestoreChannel and we cannot have an integration test which returns a document even if it closed by an error.

PR #3625, opened 2022-04-07
Simplify Crashlytics settings

mrichards:
The "app" portion of the settings response is no longer used by
Crashlytics. This commit removes parsing of that section, which
required refactoring how the Task API is used to wait for settings
to return.

It also flattens the package hierarchy for the settings.* packages
and removes some unnecessary interfaces.


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/settings/Settings.java
```diff
@@ -12,86 +12,84 @@
...
-  @Override
-  public SessionSettingsData getSessionData() {
+  public SessionData getSessionData() {
     return sessionData;
   }
 
-  @Override
-  public FeaturesSettingsData getFeaturesData() {
-    return featuresData;
+  public FeatureFlagData getFeatureFlagData() {
```
mrober, line 77: Why do we need these getters when the fields are public already?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/settings/SettingsController.java
```diff
@@ -212,39 +211,33 @@ public Settings getSettings() {
...
                   settings.set(fetchedSettings);
 
-                  // Signal the app settings on any Tasks that already exist, and then replace the
-                  // task so
-                  // that any new callers get the new app settings instead of any old ones.
-                  appSettingsData.get().trySetResult(fetchedSettings.getAppSettingsData());
-                  TaskCompletionSource<AppSettingsData> fetchedAppSettings =
-                      new TaskCompletionSource<>();
-                  fetchedAppSettings.trySetResult(fetchedSettings.getAppSettingsData());
-                  appSettingsData.set(fetchedAppSettings);
```
mrober, line 222: Did you verify that we don't need to do this?

mrober (approved):

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/settings/SettingsController.java
mrichards: It works as expected in the happy path. I found a bug that can occur when there's no network access, but it predates this PR by at least several versions. So I'm fairly confident the above lines are fine, but I'll work it all out when I fix that bug.

PR #3627, opened 2022-04-07
Update App Check SDKs to use `v1` API.

rosalyntan:
Switch the App Check SDKs over to using the `v1` App Check APIs.


malcolmdeck (approved):
LGTM!

PR #3637, opened 2022-04-12
Fireperf: fix unchecked operations in e2e `FirebasePerformanceFragmentScreenTracesTest`

leotianlizhan:
An attempt to fix b/228604783.


jeremyjiang-dev:
Besides changing ActivityAction to ActivityAction<FirebasePerfFragmentsActivity>, did you also rename the class?

Were you able to run the tests locally and verify the data on console?

visumickey (approved):
I highly suspect that this change would not make any difference in the execution of the project. This seems like an harmless change and happy to approve this. But I think we need to dig this deeper to understand what is causing the screen traces to not reach sawmill.

jeremyjiang-dev (approved):

jeremyjiang-dev:
> 

Nice! Do you know if we have data on console when you run `connectedCheck` before this PR? If there is no data, then we are certain that the PR will fix the problem.

PR #3648, opened 2022-04-14
Disable verbose logging in transport-runtime.

### LINKED ISSUE #3441: Firebase Performance Monitoring Console/Logcat Spam
* **Android Studio version:** Android Studio Bumblebee | 2021.1.1 Patch 1 (Build #AI-211.7628.21.2111.8139111)
  * **Firebase Component:** Firebase Performance Monitoring
  * **Component version:** `com.google.firebase:firebase-bom:29.1.0` (`com.google.firebase:firebase-perf:20.0.5`, `com.google.firebase:perf-plugin:1.4.1`)

#### Steps to reproduce:

The steps to reproduce this were rather simple. To reproduce, I simply:
1) Followed the official [getting started guide](https://firebase.google.com/docs/perf-mon/get-started-android) for Android to add Firebase Performance Monitoring to my app.
2) Built/ran my app using a physical device deployment target (Pixel 5 running Android 12).
3) Watched Logcat/console and observed the same two lines (with very minor variation) appearing over and over again, without end.

The two rows that keep appearing are as follows:
```
2022-02-14 16:17:49.179 25682-25745/com.redacted.app.redacted.dev D/TransportRuntime.SQLiteEventStore: Storing event with priority=DEFAULT, name=FIREPERF for destination cct
2022-02-14 16:17:49.182 25682-25745/com.redacted.app.redacted.dev D/TransportRuntime.JobInfoScheduler: Upload for context TransportContext(cct, DEFAULT, MSRodHRwczovL2ZpcmViYXNlbG9nZ1luZy1wYS5nb59nbGVhcGlzLmNvbS92MS9maXJlbG9nL2xlZ2FjeS9iYXRjaGxvZ2xBSXphU3lDY2traUg4aTJaQVJ3T3MxTEV6RktsZDE1YU3HOG96S28=) is already scheduled. Returning...
```

As mentioned, the above rows continue to appear over and over, with approximately less than a second between each occurrence.

I've found loosely related posts on StackOverflow and on [other projects here on GitHub](https://github.com/FirebaseExtended/flutterfire/issues/6140), but no resolution seems apparent.

In order to pre-empt the recommendation to disable logging via `AndroidManifest.xml` `firebase_performance_logcat_enabled` `meta-data`, I've already done so, however it doesn't stop the above from occurring:

```xml
<meta-data
    android:name="firebase_performance_logcat_enabled"
    android:value="false" />
```

Any help with this would be greatly appreciated!

vkryachko:
fixes #3441

### DIFF
--- a/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
+++ b/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
@@ -316,8 +316,8 @@ final class CctTransportBackend implements TransportBackend {
 
     int responseCode = connection.getResponseCode();
     Logging.i(LOG_TAG, "Status Code: %d", responseCode);
-    Logging.d(LOG_TAG, "Content-Type: " + connection.getHeaderField("Content-Type"));
-    Logging.d(LOG_TAG, "Content-Encoding: " + connection.getHeaderField("Content-Encoding"));
+    Logging.d(LOG_TAG, "Content-Type: %s", connection.getHeaderField("Content-Type"));
+    Logging.d(LOG_TAG, "Content-Encoding: %s", connection.getHeaderField("Content-Encoding"));
 
     if (responseCode == 302 || responseCode == 301 || responseCode == 307) {
       String redirect = connection.getHeaderField("Location");


jeremyjiang-dev:
Should we also update the log level for the logging statement that the ticket is mentioning? For example, https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/AlarmManagerScheduler.java#L107

vkryachko:
@jeremyjiang-dev note the change in the implementation of `Logging.d()`, it now checks if debug level is enabled, which it is not by default, see https://developer.android.com/reference/android/util/Log#isLoggable(java.lang.String,%20int)

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -24,30 +24,51 @@ private static String getTag(String tag) {
...
 
   public static void d(String tag, String message) {
-    Log.d(getTag(tag), message);
+    tag = getTag(tag);
+    if (Log.isLoggable(tag, Log.DEBUG)) {
+      Log.d(tag, message);
+    }
   }
 
   public static void d(String tag, String message, Object arg1) {
```
jeremyjiang-dev, line 33: Is this redundant of `public static void d(String tag, String message, Object... args) `.
Same applies to `public static void d(String tag, String message, Object arg1, Object arg2)`

vkryachko: it's not redundant as it avoids an Object[] array allocation, it's common to overload like this for efficiency in the most common logging cases

jeremyjiang-dev: Gotcha!

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/CctTransportBackend.java
```diff
@@ -315,9 +315,9 @@ private HttpResponse doSend(HttpRequest request) throws IOException {
     }
 
     int responseCode = connection.getResponseCode();
-    Logging.i(LOG_TAG, "Status Code: " + responseCode);
-    Logging.i(LOG_TAG, "Content-Type: " + connection.getHeaderField("Content-Type"));
-    Logging.i(LOG_TAG, "Content-Encoding: " + connection.getHeaderField("Content-Encoding"));
+    Logging.i(LOG_TAG, "Status Code: %d", responseCode);
+    Logging.d(LOG_TAG, "Content-Type: " + connection.getHeaderField("Content-Type"));
```
rlazo, line 319: Why not use formatting in these cases?

vkryachko: done

jeremyjiang-dev:
> @jeremyjiang-dev note the change in the implementation of `Logging.d()`, it now checks if debug level is enabled, which it is not by default, see https://developer.android.com/reference/android/util/Log#isLoggable(java.lang.String,%20int)

How should we advise the developer to solve the issue in the ticket. I think the developer needs to set logging level to INFO in order to get rid off the logs in the tickets.

vkryachko:
> > @jeremyjiang-dev note the change in the implementation of `Logging.d()`, it now checks if debug level is enabled, which it is not by default, see https://developer.android.com/reference/android/util/Log#isLoggable(java.lang.String,%20int)
> 
> How should we advise the developer to solve the issue in the ticket. I think the developer needs to set logging level to INFO in order to get rid off the logs in the tickets.

As you can see in the docs I linked, info is the default, nothing "lower" than INFO will be logged and has to be explicitly enabled via `setprop` to be able to see debug messages(which we will do when we need to debug issues going forward while nothing will be logged for our customers by default)

rlazo:
As per the [logging documentation](https://source.android.com/devices/tech/debug/understanding-logging#log-standards) , it's possible to strip blocks to `isLoggable` using proguard rules. If a developer uses this, our wrapper will still be called to create the concatenated TAG (since it's outside the block). Is this something we worry about, performance wise?

rlazo (approved):
My previous comment shouldn't prevent this change from being merged BTW.

vkryachko:
> As per the [logging documentation](https://source.android.com/devices/tech/debug/understanding-logging#log-standards) , it's possible to strip blocks to `isLoggable` using proguard rules. If a developer uses this, our wrapper will still be called to create the concatenated TAG (since it's outside the block). Is this something we worry about, performance wise?

I mean, that's fair. However those rules are not the default and need to be explicitly added to the app configuration. At which point imo it's reasonable for the developer to deal with the side effects of such changes.

Also, I don't think with the link you provided it would be possible to only remove isLoggable without removing Log.d(), iiuc `assumenosideeffects` only informs proguard that the call does not produce side effects, however since the result of `isLoggable` is used in the if statement, proguard would not remove such calls unless all if branches are also determined to have no side effects in them.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -24,30 +24,51 @@ private static String getTag(String tag) {
...
+    tag = getTag(tag);
+    if (Log.isLoggable(tag, Log.INFO)) {
+      Log.i(tag, String.format(message, arg1));
+    }
   }
 
   public static void e(String tag, String message, Throwable e) {
-    Log.e(getTag(tag), message, e);
+    tag = getTag(tag);
+    if (Log.isLoggable(tag, Log.ERROR)) {
```
gevorg-kopalyan, line 63: Log.isLoggable will produce an error because sometimes it can exceed the limit of 23 characters

PR #3653, opened 2022-04-14
Add first-open time to Remote Config fetch request.

richaurora:


firebase-config/firebase-config.gradle
```diff
@@ -78,4 +78,6 @@ dependencies {
     androidTestImplementation 'com.linkedin.dexmaker:dexmaker-mockito:2.25.0'
     androidTestImplementation 'junit:junit:4.12'
     androidTestImplementation "org.skyscreamer:jsonassert:1.5.0"
+
+    api 'joda-time:joda-time:2.10.14'
```
vkryachko, line 82: This is a pretty heavy dependency, especially given that you don't seem to do any arithmetic on it but just rely on its toString() method, I suggest to use the long directly and use SimpleDateFormat to convert it to an iso string.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -79,6 +80,9 @@ public class ConfigFetchHandler {
    */
   @VisibleForTesting static final int HTTP_TOO_MANY_REQUESTS = 429;
 
+  /** First Open time key in GA user-properties. */
```
danasilver, line 83: Might add to the comment that this is a reserved property set by GA (I think that's the case?) per https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -174,6 +177,7 @@ FetchResponse fetch(
       Map<String, String> analyticsUserProperties,
       String lastFetchETag,
       Map<String, String> customHeaders,
+      Instant firstOpenTime,
```
danasilver, line 180: Thinking aloud: it doesn't make sense to merge this in with the other `analyticsUserProperties` since we want to send it as its own request param.

firebase-config/firebase-config.gradle
richaurora: Removed the dependency and replaced it with SimpleDataFormat. Thanks, I definitely struggled a bit with this and was looking for inputs!

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
richaurora: Yes, it is an internal property. Updated the comment to say that it's collected by GA as part of of the predefined "user-dimensions". I called it "user-dimensions" following the recent GA documentation (https://support.google.com/firebase/answer/9268042?hl=en&ref_topic=6317484) but on second thought, maybe something simpler is better: "internal user-property automatically collected by GA" ?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
richaurora: Yes, and the reason we want to send it in its own request param is because analyticsUserProperties in fetch traditionally contains custom user-defined properties only. And we want to keep, and highlight, that distinction. It also lets us explicitly call out the information that we are sending up to the server.

danasilver (approved):
One small comment about a comment. Looks good to me!

sknopf (approved):
LGTM

vkryachko (approved):

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -79,6 +79,12 @@ public class ConfigFetchHandler {
    */
   @VisibleForTesting static final int HTTP_TOO_MANY_REQUESTS = 429;
 
+  /**
+   * First-open-time key name in GA user-properties. First-open time is a predefined user-dimension
```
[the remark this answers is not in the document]
richaurora: Done. :)

PR #3657, opened 2022-04-15
Re-run limit queries without limit if they need to be re-filled

schmidt-sebastian:
This changes the query engine to also drop the limit on limit queries that need to be refilled. It also cleans up some of the Target.getLimit() logic that leads to repetitive code.


schmidt-sebastian:
I am not sure why the test is failing... I am still trying to debug it.

schmidt-sebastian:
Should be fixed now.

ehsannas (approved):

PR #3660, opened 2022-04-18
Skip custom attributes when the length of the key or the value is 0

visumickey:


firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
```diff
@@ -189,6 +189,14 @@ public void testInvalidCustomAttribute() {
     trace.clearCustomAttributes().putCustomAttributes("ga_test", "value");
     assertThat(new FirebasePerfTraceValidator(trace.build()).isValidPerfMetric()).isFalse();
 
+    trace = trace.clone();
+    trace.clearCustomAttributes().putCustomAttributes("key", "");
+    assertThat(new FirebasePerfTraceValidator(trace.build()).isValidPerfMetric()).isFalse();
+
+    trace = trace.clone();
+    trace.clearCustomAttributes().putCustomAttributes("value", "");
```
jeremyjiang-dev, line 197: Do you want to test `trace.clearCustomAttributes().putCustomAttributes("", "value");`

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
```diff
@@ -143,9 +145,9 @@ public static String validateMetricName(@Nullable String str) {
   public static String validateAttribute(@NonNull Map.Entry<String, String> attribute) {
     String key = attribute.getKey();
     String value = attribute.getValue();
-    if (key == null) {
+    if (key == null || key.length() == 0) {
       return "Attribute key must not be null";
```
jeremyjiang-dev, line 149: nit: return "Attribute key must not be null or an empty string.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
```diff
@@ -143,9 +145,9 @@ public static String validateMetricName(@Nullable String str) {
...
-    } else if (value == null) {
+    } else if (value == null || value.length() == 0) {
       return "Attribute value must not be null";
```
jeremyjiang-dev, line 151: nit: return "Attribute key must not be null or an empty string.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
```diff
@@ -630,6 +630,11 @@ private void checkAttribute(@NonNull String key, @NonNull String value) {
       throw new IllegalArgumentException(
           String.format(Locale.ENGLISH, "Trace '%s' has been stopped", name));
     }
+
+    if (key == null || value == null) {
```
jeremyjiang-dev, line 634: Is this redundant? We are already validating the key and value in https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java#L641

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
```diff
@@ -28,6 +28,8 @@
 /** An abstract class providing an interface to validate PerfMetric */
 public abstract class PerfMetricValidator {
 
+  private String format;
```
jeremyjiang-dev, line 31: Is this used?

visumickey: This is not used! Thanks for calling this out. Removed it.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
visumickey: Yes, this is indeed redundant. Removed this.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
visumickey: Done!

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
visumickey: Updated! Done.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
visumickey: Ha! Good catch!

jeremyjiang-dev (approved):

firebase-perf/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Refer [GMaven](https://maven.google.com/web/index.html?q=firebase-perf#com.googl
 
 *   {{fixed}} Fixed a bug where screen traces were not capturing frame metrics for multi-activity apps.
 *   {{feature}} Added support for measuring screen performance metrics for "Activity Fragments" out-of-the-box.
+*   {{fixed}} Exclude custom attributes whose key/value length is 0.
```
jeremyjiang-dev, line 26: nit : Excluded

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
```diff
@@ -169,7 +169,7 @@ public void screenTrace_shouldNotAllowNonPositiveTotalFrames() {
   }
 
   @Test
-  public void testInvalidCustomAttribute() {
+  public void traceValidator_invalidCustomAttriubute_skipsAttribute() {
```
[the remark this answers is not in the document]
visumickey: Fixed it. Let me know how this looks to you!

firebase-perf/src/main/java/com/google/firebase/perf/metrics/HttpMetric.java
```diff
@@ -181,24 +181,21 @@ public void putAttribute(@NonNull String attribute, @NonNull String value) {
...
 
-  private void checkAttribute(@Nullable String attribute, @Nullable String value) {
+  private void checkAttribute(@Nullable String key, @Nullable String value) {
     if (isStopped) {
       throw new IllegalArgumentException(
           "HttpMetric has been logged already so unable to modify attributes");
     }
-    if (attribute == null || value == null) {
-      throw new IllegalArgumentException("Attribute must not have null key or value.");
-    }
```
jposuna, line 191: Why did you get rid of that check? The params are @Nullable but we don't verify they are non-null. Also, why are the params marked as @Nullable if we want them not to be null?

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
```diff
@@ -169,7 +169,7 @@ public void screenTrace_shouldNotAllowNonPositiveTotalFrames() {
   }
 
   @Test
-  public void testInvalidCustomAttribute() {
+  public void traceValidator_invalidCustomAttribute_marksPerfMetricInvalid() {
```
jposuna, line 172: The best practice is to test one behavior per test. We should try to avoid having one monolithic test that asserts so many things.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
```diff
@@ -189,6 +189,14 @@ public void testInvalidCustomAttribute() {
     trace.clearCustomAttributes().putCustomAttributes("ga_test", "value");
     assertThat(new FirebasePerfTraceValidator(trace.build()).isValidPerfMetric()).isFalse();
 
+    trace = trace.clone();
```
jposuna, line 192: Why do we clone instead of creawting a new valid trace metric? E.g. if we made some other modification to the trace that causes it to be invalid, no amount of changing the custom attributes would make it valid, so the tests could be passing incorrectly.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
```diff
@@ -189,6 +189,14 @@ public void testInvalidCustomAttribute() {
...
+    trace.clearCustomAttributes().putCustomAttributes("key", "");
```
raymondlam, line 193: Can we have tests for when key/values are null?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/HttpMetric.java
```diff
@@ -181,24 +181,21 @@ public void putAttribute(@NonNull String attribute, @NonNull String value) {
...
         && customAttributesMap.size() >= Constants.MAX_TRACE_CUSTOM_ATTRIBUTES) {
       throw new IllegalArgumentException(
           String.format(
               Locale.ENGLISH,
               "Exceeds max limit of number of attributes - %d",
               Constants.MAX_TRACE_CUSTOM_ATTRIBUTES));
     }
-    String err =
-        PerfMetricValidator.validateAttribute(new AbstractMap.SimpleEntry<>(attribute, value));
+    String err = PerfMetricValidator.validateAttribute(new AbstractMap.SimpleEntry<>(key, value));
```
raymondlam, line 198: nit: Why do we return an error instead of just throwing an IllegalArgumentException in the validate method? This makes it slightly harder to find where the exception took place.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/HttpMetric.java
raymondlam: +1 Since nulls are not valid, shouldn't it be `@NonNull`

visumickey: This is a redundant check. This is already validated in the `validateAttribute` function and so I chose to remove it to avoid redundancy. Modified to make the param "NonNull". The public facing API is already calling it as "NonNull", this was an internal check since we could trim the attribute to avoid the whitespaces in the front and back.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
visumickey: Added!

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
visumickey: Fair point. I have separate the tests and have got devoid of clone now.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
visumickey: Good point. Separated the tests to individual atomic tests.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/HttpMetric.java
visumickey: Like the idea. This had to be done in many parts of the code. For now, I have made this change just for the custom attributes API.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
```diff
@@ -638,10 +639,7 @@ private void checkAttribute(@NonNull String key, @NonNull String value) {
               "Exceeds max limit of number of attributes - %d",
               Constants.MAX_TRACE_CUSTOM_ATTRIBUTES));
     }
-    String err = PerfMetricValidator.validateAttribute(new AbstractMap.SimpleEntry<>(key, value));
-    if (err != null) {
-      throw new IllegalArgumentException(err);
-    }
+    PerfMetricValidator.validateAttribute(new AbstractMap.SimpleEntry<>(key, value));
```
raymondlam, line 642: nit: we can statically import this.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidator.java
```diff
@@ -180,9 +180,10 @@ private boolean isValidCounterId(@Nullable String counterId) {
 
   private boolean hasValidAttributes(Map<String, String> customAttributes) {
     for (Map.Entry<String, String> entry : customAttributes.entrySet()) {
-      String err = PerfMetricValidator.validateAttribute(entry);
-      if (err != null) {
-        logger.warn(err);
+      try {
+        PerfMetricValidator.validateAttribute(entry);
+      } catch (IllegalArgumentException exception) {
+        logger.warn(exception.getLocalizedMessage());
```
raymondlam, line 186: nit: (preferably in a separate cl) we should have a separate logger.warn that accepts a throwable. This way the stacktrace is still logged.

https://developer.android.com/reference/android/util/Log#w(java.lang.String,%20java.lang.Throwable)

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
```diff
@@ -140,28 +140,30 @@ public static String validateMetricName(@Nullable String str) {
...
+  public static void validateAttribute(@NonNull Map.Entry<String, String> attribute) {
     String key = attribute.getKey();
     String value = attribute.getValue();
-    if (key == null) {
-      return "Attribute key must not be null";
-    } else if (value == null) {
-      return "Attribute value must not be null";
+    if (key == null || key.length() == 0) {
+      throw new IllegalArgumentException("Attribute key must not be null or empty");
+    } else if (value == null || value.length() == 0) {
```
raymondlam, line 148: The `else if` are unnecessarily when you are already throwing/returning for each condition. You can just use `if`.

raymondlam (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/Trace.java
visumickey: Done.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
visumickey: Ha! Obvious! Thanks for the reminder.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidator.java
visumickey: Acknowledged! Good idea and will follow up on this separately.

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -364,10 +364,7 @@ private void checkAttribute(@Nullable String key, @Nullable String value) {
               Constants.MAX_TRACE_CUSTOM_ATTRIBUTES));
     }
 
-    String err = PerfMetricValidator.validateAttribute(new AbstractMap.SimpleEntry<>(key, value));
-    if (err != null) {
-      throw new IllegalArgumentException(err);
-    }
+    PerfMetricValidator.validateAttribute(new AbstractMap.SimpleEntry<>(key, value));
```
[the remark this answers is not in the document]
visumickey: Makes perfect sense to modify the name of the method. Done!

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidator.java
```diff
@@ -180,9 +180,10 @@ private boolean isValidCounterId(@Nullable String counterId) {
 
   private boolean hasValidAttributes(Map<String, String> customAttributes) {
```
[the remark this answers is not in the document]
visumickey: Sounds good! Like the idea, done.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/validator/PerfMetricValidator.java
```diff
@@ -135,33 +134,42 @@ public static String validateMetricName(@Nullable String str) {
   /**
    * Checks whether the given map entry fits key/value constraints for a Trace Attribute.
    *
-   * @param attribute A key/value pair for an Attribute
+   * @param key Key for the Attribute
+   * @param value Value for the Attribute
    * @return null if the entry can be used as an Attribute, if not, an error string explaining why
    *     it can't be used.
    */
   @Nullable
```
jposuna, line 142: Nit: This is no longer `@Nullable` since it's void :)

visumickey: Done!

firebase-perf/src/test/java/com/google/firebase/perf/metrics/validator/FirebasePerfTraceValidatorTest.java
```diff
@@ -169,47 +169,77 @@ public void screenTrace_shouldNotAllowNonPositiveTotalFrames() {
...
+    assertThat(new FirebasePerfTraceValidator(trace.build()).isValidPerfMetric()).isFalse();
+  }
+
+  @Test
+  public void traceValidator_customAttributeWithLongKey_marksPerfMetricInvalid() {
 
     StringBuilder longString = new StringBuilder();
     for (int i = 0; i <= Constants.MAX_ATTRIBUTE_KEY_LENGTH; i++) {
       longString.append("a");
     }
```
jposuna, line 227: FYI, since Java11 you can do "a".repeat(Constants.MAX_ATTRIBUTE_KEY_LENGTH) and it will do the same thing.

visumickey: Done!

leotianlizhan (approved):
Great idea to make errors thrown from the origin so we have better stacktraces for debugging!

PR #3665, opened 2022-04-20
Fireperf: encapsulate `FrameMetricsAggregator` in a new object called `FrameMetricsRecorder`

leotianlizhan:
b/229123892

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
@@ -1,0 +1,148 @@
+package com.google.firebase.perf.application;
+
+import android.app.Activity;
+import android.util.SparseIntArray;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.core.app.FrameMetricsAggregator;
+
+import com.google.android.gms.common.util.VisibleForTesting;
+import com.google.firebase.perf.logging.AndroidLogger;
+import com.google.firebase.perf.metrics.FrameMetricsCalculator;
+import com.google.firebase.perf.metrics.FrameMetricsCalculator.PerfFrameMetrics;
+import com.google.firebase.perf.util.Constants;
+
+import java.util.WeakHashMap;
+
+/**
+ * Provides FrameMetrics data from an Activity's Window. Encapsulates FrameMetricsAggregator.
+ * <p>
+ * Note: each recorder holds a reference to an Activity, so it is very important to dereference
+ * each recorder before the associated Activity is destroyed
+ * </p>
+ * */
+public class FrameMetricsRecorder {
+  private static final AndroidLogger logger = AndroidLogger.getInstance();
+  private final Activity activity;
+  private final FrameMetricsAggregator fma;
+  private final WeakHashMap<Object, PerfFrameMetrics> subTraceMap;
+  private boolean isRecording;
+
+  /**
+   * Creates a recorder for a specific activity
+   * @param activity the activity that the recorder is collecting data from
+   */
+  public FrameMetricsRecorder(Activity activity) {
+    this(activity, new FrameMetricsAggregator(), new WeakHashMap<>(), false);
+  }
+
+  @VisibleForTesting
+  FrameMetricsRecorder(Activity activity, FrameMetricsAggregator frameMetricsAggregator, WeakHashMap<Object, PerfFrameMetrics> subTraceMap, boolean isRecording) {
+    this.activity = activity;
+    this.fma = frameMetricsAggregator;
+    this.subTraceMap = subTraceMap;
+    this.isRecording = isRecording;
+  }
+
+  /**
+   * Starts recording FrameMetrics for the activity window
+   */
+  public void start() {
+    if (isRecording) {
+      logger.error("FrameMetricsAggregator is already recording %s", activity.getClass().getSimpleName());
+      return;
+    }
+    fma.add(activity);
+    isRecording = true;
+  }
+
+  /**
+   * Stops recording FrameMetrics for the activity window
+   * @return FrameMetrics accumulated during the current recording
+   */
+  public PerfFrameMetrics stop() {
+    if (!isRecording) {
+      logger.error("Cannot stop because no recording was started");
+      return null;
+    }
+    isRecording = false;
+    PerfFrameMetrics data = this.snapshot();
+    try {
+      fma.remove(activity);
+    } catch (IllegalArgumentException ignored) {
+      logger.error("View not hardware accelerated. Unable to collect FrameMetrics.");
+    }
+    fma.reset();
+    return data;
+  }
+
+  /**
+   * Starts a sub-trace in the current recording
+   * @param key a UI state to associate this sub-trace with. e.g.) fragment
+   */
+  public void startSubTrace(Object key) {
+    if (!isRecording) {
+      logger.error("Cannot start sub-trace because FrameMetricsAggregator is not recording");
+      return;
+    }
+    if (subTraceMap.containsKey(key)) {
+      logger.error("Cannot start sub-trace because one is already ongoing with the key %s", key.getClass().getSimpleName());
+      return;
+    }
+    subTraceMap.put(key, this.snapshot());
+  }
+
+  /**
+   * Stops the sub-trace associated with the given UI state
+   * @param key the UI state associated with the sub-trace to be stopped
+   * @return FrameMetrics accumulated during this sub-trace
+   */
+  public PerfFrameMetrics stopSubTrace(Object key) {
+    if (!isRecording) {
+      logger.error("Cannot stop sub-trace because FrameMetricsAggregator is not recording");
+      return null;
+    }
+    if (!subTraceMap.containsKey(key)) {
+      logger.error("Sub-trace associated with key %s was not started or does not exist", key.getClass().getSimpleName());
+      return null;
+    }
+    PerfFrameMetrics snapshotStart = subTraceMap.remove(key);
+    PerfFrameMetrics snapshotEnd = this.snapshot();
+    return snapshotEnd.subtract(snapshotStart);
+  }
+
+  /**
+   * Calculate total frames, slow frames, and frozen frames from SparseIntArray[] recorded by {@link
+   * FrameMetricsAggregator}.
+   *
+   * @return the frame metrics
+   */
+  protected @NonNull PerfFrameMetrics snapshot() {
+    SparseIntArray[] arr = this.fma.getMetrics();
+    int totalFrames = 0;
+    int slowFrames = 0;
+    int frozenFrames = 0;
+
+    if (arr != null) {
+      SparseIntArray frameTimes = arr[FrameMetricsAggregator.TOTAL_INDEX];
+      if (frameTimes != null) {
+        for (int i = 0; i < frameTimes.size(); i++) {
+          int frameTime = frameTimes.keyAt(i);
+          int numFrames = frameTimes.valueAt(i);
+          totalFrames += numFrames;
+          if (frameTime > Constants.FROZEN_FRAME_TIME) {
+            // Frozen frames mean the app appear frozen.  The recommended thresholds is 700ms
+            frozenFrames += numFrames;
+          }
+          if (frameTime > Constants.SLOW_FRAME_TIME) {
+            // Slow frames are anything above 16ms (i.e. 60 frames/second)
+            slowFrames += numFrames;
+          }
+        }
+      }
+    }
+    // Only incrementMetric if corresponding metric is non-zero.
+    return new PerfFrameMetrics(totalFrames, slowFrames, frozenFrames);
+  }
+}
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
@@ -49,6 +49,13 @@ public class FrameMetricsCalculator {
     public int getTotalFrames() {
       return totalFrames;
     }
+
+    public PerfFrameMetrics subtract(PerfFrameMetrics b) {
+      int newTotalFrames = this.totalFrames - b.getTotalFrames();
+      int newSlowFrames = this.slowFrames - b.getSlowFrames();
+      int newFrozenFrames = this.frozenFrames - b.getFrozenFrames();
+      return new PerfFrameMetrics(newTotalFrames, newSlowFrames, newFrozenFrames);
+    }
   }
 
   /**
--- a/firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
+++ b/firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
@@ -1,0 +1,236 @@
+package com.google.firebase.perf.application;
+
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.ArgumentMatchers.nullable;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+import android.app.Activity;
+import android.view.WindowManager;
+
+import androidx.core.app.FrameMetricsAggregator;
+
+import com.google.firebase.perf.FirebasePerformanceTestBase;
+import com.google.firebase.perf.metrics.FrameMetricsCalculator.PerfFrameMetrics;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.InOrder;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.robolectric.Robolectric;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.android.controller.ActivityController;
+
+import java.util.List;
+import java.util.WeakHashMap;
+
+/** Unit tests for {@link com.google.firebase.perf.application.FrameMetricsRecorder}. */
+@RunWith(RobolectricTestRunner.class)
+public class FrameMetricsRecorderTest extends FirebasePerformanceTestBase {
+  private Activity activity;
+  private FrameMetricsRecorder recorder;
+
+  @Mock private FrameMetricsAggregator fma;
+  @Mock private PerfFrameMetrics frameMetrics1;
+  @Mock private PerfFrameMetrics frameMetrics2;
+  @Mock private PerfFrameMetrics frameMetrics3;
+
+  @Spy
+  private final WeakHashMap<Object, PerfFrameMetrics> subTraceMap = new WeakHashMap<>();
+
+  @Before
+  public void setUp() {
+    initMocks(this);
+    activity = createFakeActivity(true);
+    recorder = spy(new FrameMetricsRecorder(activity, fma, subTraceMap, false));
+    // stop() depends on snapshot(), so we stub snapshot() to test stop() independently
+    doReturn(null).when(recorder).snapshot();
+  }
+
+  @Test
+  public void eachInstance_usesOneAndTheSameActivityInFrameMetricsAggregator_always() {
+    ArgumentCaptor<Activity> activityCaptor = ArgumentCaptor.forClass(Activity.class);
+    recorder.start();
+    recorder.stop();
+    recorder.start();
+    recorder.stop();
+    verify(fma, times(2)).add(activityCaptor.capture());
+    verify(fma, times(2)).remove(activityCaptor.capture());
+    List<Activity> activities = activityCaptor.getAllValues();
+    Assert.assertTrue(activities.stream().allMatch(a -> a == activity));
+  }
+
+  @Test
+  public void start_callsFrameMetricsAggregator() {
+    verify(fma, times(0)).add(nullable(Activity.class));
+    recorder.start();
+    verify(fma, times(1)).add(nullable(Activity.class));
+  }
+
+  @Test
+  public void start_whileAlreadyStarted_fails() {
+    recorder.start();
+    recorder.start();
+    verify(fma, times(1)).add(nullable(Activity.class));
+  }
+
+  @Test
+  public void start_afterPreviousEnded_succeeds() {
+    recorder.start();
+    recorder.stop();
+    recorder.start();
+    verify(fma, times(2)).add(nullable(Activity.class));
+  }
+
+  @Test
+  public void stop_callsFrameMetricsAggregator() {
+    recorder.start();
+    recorder.stop();
+    verify(fma, times(1)).remove(nullable(Activity.class));
+    verify(fma, times(1)).reset();
+  }
+
+  @Test
+  public void stop_whileNotStarted_fails() {
+    recorder.stop();
+    verify(fma, times(0)).remove(nullable(Activity.class));
+    verify(fma, times(0)).reset();
+    recorder.start();
+    recorder.stop();
+    recorder.stop();
+    verify(fma, times(1)).remove(nullable(Activity.class));
+    verify(fma, times(1)).reset();
+  }
+
+  @Test
+  public void stop_snapshotsFrameMetricsAggregator_beforeReset() {
+    FrameMetricsRecorder recorder = new FrameMetricsRecorder(activity, fma, subTraceMap, true);
+    FrameMetricsRecorder spyRecorder = spy(recorder);
+    // stop() depends on snapshot(), so we stub snapshot() to test stop() independently
+    doReturn(null).when(spyRecorder).snapshot();
+    spyRecorder.stop();
+    InOrder orderVerifier = inOrder(spyRecorder, fma);
+    orderVerifier.verify(spyRecorder).snapshot();
+    orderVerifier.verify(fma).reset();
+  }
+
+  @Test
+  public void startSubTrace_whenNotRecording_fails() {
+    Object uiState = new Object();
+    recorder.startSubTrace(uiState);
+    verify(subTraceMap, times(0)).put(nullable(Object.class), nullable(PerfFrameMetrics.class));
+
+    recorder.start();
+    recorder.stop();
+    recorder.startSubTrace(uiState);
+    verify(subTraceMap, times(0)).put(nullable(Object.class), nullable(PerfFrameMetrics.class));
+  }
+
+  @Test
+  public void startSubTrace_whenSameSubTraceWithGivenKeyIsAlreadyOngoing_fails() {
+    doReturn(frameMetrics1).when(recorder).snapshot();
+    ArgumentCaptor<Object> objectCaptor = ArgumentCaptor.forClass(Object.class);
+    Object uiState1 = new Object();
+    Object uiState2 = new Object();
+
+    recorder.start();
+    recorder.startSubTrace(uiState1);
+    verify(subTraceMap, times(1)).put(objectCaptor.capture(), nullable(PerfFrameMetrics.class));
+    Assert.assertSame(subTraceMap.get(uiState1), frameMetrics1);
+    Assert.assertSame(objectCaptor.getValue(), uiState1);
+    Assert.assertNotSame(objectCaptor.getValue(), uiState2);
+
+    recorder.startSubTrace(uiState1);
+    verify(subTraceMap, times(1)).put(nullable(Object.class), nullable(PerfFrameMetrics.class));
+  }
+
+  @Test
+  public void startSubTrace_whenSucceeds_putsNewEntryInMap() {
+    doReturn(frameMetrics1).when(recorder).snapshot();
+    Object uiState1 = new Object();
+    Object uiState2 = new Object();
+    recorder.start();
+    recorder.startSubTrace(uiState1);
+    Assert.assertSame(subTraceMap.get(uiState1), frameMetrics1);
+
+    doReturn(frameMetrics2).when(recorder).snapshot();
+    recorder.startSubTrace(uiState2);
+    Assert.assertSame(subTraceMap.get(uiState2), frameMetrics2);
+  }
+
+  @Test
+  public void stopSubTrace_whenNotRecording_fails() {
+    Object uiState = new Object();
+    subTraceMap.put(uiState, frameMetrics1);
+    Assert.assertTrue(subTraceMap.containsKey(uiState));
+
+    recorder.stopSubTrace(uiState);
+    verify(subTraceMap, times(0)).remove(nullable(Object.class));
+
+    recorder.start();
+    recorder.stop();
+    recorder.startSubTrace(uiState);
+    verify(subTraceMap, times(0)).remove(nullable(Object.class));
+  }
+
+  @Test
+  public void stopSubTrace_whenNoSubTraceWithGivenKeyExists_fails() {
+    doReturn(frameMetrics1).when(recorder).snapshot();
+    Object uiState1 = new Object();
+    Object uiState2 = new Object();
+    subTraceMap.put(uiState2, frameMetrics2);
+
+    recorder.start();
+    recorder.stopSubTrace(uiState1);
+    verify(subTraceMap, times(0)).remove(nullable(Object.class));
+  }
+
+  @Test
+  public void stopSubTrace_whenSucceeds_removesEntryInMap() {
+    doReturn(frameMetrics2).when(recorder).snapshot();
+    Object uiState1 = new Object();
+    recorder.start();
+    subTraceMap.put(uiState1, frameMetrics1);
+    Assert.assertEquals(1, subTraceMap.size());
+    recorder.stopSubTrace(uiState1);
+
+    Assert.assertEquals(0, subTraceMap.size());
+  }
+
+  @Test
+  public void stopSubTrace_whenSucceeds_returnsDifferenceBetweenSnapshots() {
+    recorder.start();
+    Object uiState1 = new Object();
+    doReturn(frameMetrics1).when(recorder).snapshot();
+    recorder.startSubTrace(uiState1);
+
+    doReturn(frameMetrics2).when(recorder).snapshot();
+    PerfFrameMetrics difference = mock(PerfFrameMetrics.class);
+    doReturn(difference).when(frameMetrics2).subtract(argThat(arg -> arg == frameMetrics1));
+
+    PerfFrameMetrics result = recorder.stopSubTrace(uiState1);
+    Assert.assertSame(difference, result);
+  }
+
+  private static Activity createFakeActivity(boolean isHardwareAccelerated) {
+    ActivityController<Activity> fakeActivityController = Robolectric.buildActivity(Activity.class);
+
+    if (isHardwareAccelerated) {
+      fakeActivityController.get().getWindow().addFlags(WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED);
+    } else {
+      fakeActivityController.get().getWindow().clearFlags(WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED);
+    }
+
+    return fakeActivityController.start().get();
+  }
+}


firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,148 @@
...
+    return snapshotEnd.subtract(snapshotStart);
+  }
+
+  /**
+   * Calculate total frames, slow frames, and frozen frames from SparseIntArray[] recorded by {@link
+   * FrameMetricsAggregator}.
+   *
+   * @return the frame metrics
+   */
+  protected @NonNull PerfFrameMetrics snapshot() {
```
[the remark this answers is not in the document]
leotianlizhan: Yes, I copied that method over here because I realize no other class will use that method once we encapsulate everything related to FMA, including calculations, in the Recorder. I think it's better to make that into a implementation detail hidden away by the recorder (private/protected method). 

I was planning to remove Calculator and make a separate file for `PerfFrameMetrics` later in the 2nd PR, because `AppStateMonitor` and `FragmentStateMonitor` uses the Calculator, and we still need `PerfFrameMetrics`.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  /**
+   * Stops recording FrameMetrics for the activity window
+   * @return FrameMetrics accumulated during the current recording
+   */
+  public PerfFrameMetrics stop() {
+    if (!isRecording) {
+      logger.error("Cannot stop because no recording was started");
+      return null;
```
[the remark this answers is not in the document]
leotianlizhan: The purpose of these checks is to tell us if the class is being used wrong (calling fma.add twice). Reading go/java-practices/null#tolerant, I agree we shouldn't use null, but in this case I think we should throw errors. How do you feel about that solution?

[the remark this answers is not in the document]
jeremyjiang-dev: I think logging error messages is fine because we don't want the app to crash because of the exception we are throwing here. If we don't throw exceptions, the caller needs to know the return value, which we will prefer Optional instead of null.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,148 @@
...
+    isRecording = true;
+  }
+
+  /**
+   * Stops recording FrameMetrics for the activity window
+   * @return FrameMetrics accumulated during the current recording
+   */
+  public PerfFrameMetrics stop() {
+    if (!isRecording) {
+      logger.error("Cannot stop because no recording was started");
```
visumickey, line 66: Should be an error or a warning? Rationale to consider: Anything thrown as an error will show up on the console for the developers using it (even with the default log levels). If this is not disrupting the behavior of the application, recommend using logger levels like warnings, debug and info.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
visumickey: What you mean by throwing errors? Exception and let the applcation crash? I would refrain from that since that would cause the application of the user to fail because of an issue with the SDK.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,148 @@
...
+      return null;
+    }
+    isRecording = false;
+    PerfFrameMetrics data = this.snapshot();
+    try {
+      fma.remove(activity);
+    } catch (IllegalArgumentException ignored) {
+      logger.error("View not hardware accelerated. Unable to collect FrameMetrics.");
```
visumickey, line 74: Should the hardware acceleration be checked before starting to record rather than wait for the stoping the recording?

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+    fma.reset();
+    return data;
+  }
+
+  /**
+   * Starts a sub-trace in the current recording
+   * @param key a UI state to associate this sub-trace with. e.g.) fragment
+   */
+  public void startSubTrace(Object key) {
```
visumickey, line 84: Since this is a generic `Object` type, can we type check internally on what the type of the object is before we start recording the metrics for the Object?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
```diff
@@ -49,6 +49,13 @@ public int getSlowFrames() {
     public int getTotalFrames() {
       return totalFrames;
     }
+
+    public PerfFrameMetrics subtract(PerfFrameMetrics b) {
```
visumickey, line 53: I think need to get a better name for this. "Subtract" does not sound any intuitive here. Some recommendation could be `differenceInFrameMetricsFromSnapshot(PerfFrameMetrics oldSnapshot)`.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
leotianlizhan: specifically throwing `IllegalStateException`. The point is that we let the application crash when we are testing manually or in E2E pipeline, so we are alerted to errors like very intricate FMA interaction causing screen traces to be completely wrong silently that would otherwise go unnoticed for months.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: I make it generic `Object` to be more robust and not just for fragments. Perhaps in the future some sort of Jetpack Compose UI can be associated with a sub-trace, this implementation isn't only allowing fragments, it's to be very generic and more in the perspective of `FrameMetricsAggregator`. FMA doesn't care what you are tagging this subtrace as. From FMA's perspective, we are making a sub-trace (snapshot B - snapshot A), and just giving it a "key" as a name.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
leotianlizhan: Had a small chat with Visu, I realize as a library/SDK it's more important to not crash than to have incorrect feature/function. Both are really bad, but crashing is the bigger evil because it discourages them to use our SDK. But I still believe we need a way to alert for erroneous functionality more obviously than a very quiet log then the program moves on. in this situation you can only discover when some customer look into the specifics like how many frames our FMA reports vs their own FMA reports.  That's way too hidden.

In Java, `assert` will throw exceptions only when `-ea` is passed in to the JVM. This allows you to throw in debug builds but not production builds.

In Android, due to the way ART works, every app starts off by forking a zygote, and that zygote has assert disabled, so `assert` won't work any E2E test running on Android, ever.

Can we make autopush use debug build? Then check `if (BuildConfig.DEBUG)` then throw, otherwise be silent. This way our E2E tests will fail because of exceptions.

jposuna (commented):
Before reviewing the changes themselves, can you explain why we want to have this?

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
jeremyjiang-dev: I think it is fine to only throw exception in debug build. 
I found some code from Crashlytics does this. https://github.com/firebase/firebase-android-sdk/blob/5f7ac459817e30f36c0ff9711f00c4011a0ee683/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsController.java#L449-L456

jeremyjiang-dev:
Could you mark this PR to ready for review and I can take another look?
Would you like to merge this PR to a feature branch instead of master for all the frame metrics refactor work?

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
leotianlizhan: That's awesome! Can we ask Crashlytics folks on if this can be setup to work in the automated E2E pipeline (in TestLab)?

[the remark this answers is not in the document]
jeremyjiang-dev: We can set `android:debuggable=true` in E2E test app's manifest. 
Ref: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CommonUtils.java#L483-L489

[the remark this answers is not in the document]
leotianlizhan: I saw that too but I just wanted to confirm there's no unexpected limitations. 

Also, we probably don't want to change E2E test app until we fix the "Prod-E2E-Uses-Master-E2E-App" oversight (@visumickey found this yesterday)

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: We used to have hardware acceleration checks but it was not working, so we deprecated it.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
leotianlizhan: I thought "subtract" because we are literally subtracting the counts for total/slow/frozen frames. Also snapshot is a concept that should be contained in `FrameMetricsRecorder` imo. `PerfFrameMetrics` shouldn't know anything about snapshot, and its method naming should be whatever this operation does from the perspective of `PerfFrameMetrics`.

leotianlizhan:
> Could you mark this PR to ready for review and I can take another look? Would you like to merge this PR to a feature branch instead of master for all the frame metrics refactor work?

Done. I changed base branch to a new feature branch

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: IMO that would couple the Recorder with other concepts unnecessarily. Recorder is a wrapper of FMA, and it should only be responsible for operations related to FMA. A "sub-trace" only needs a key, it doesn't matter what type it is. It could be any UI object representing that "the activity currently has X UI on screen".

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * <p>Note: each recorder holds a reference to an Activity, so it is very important to dereference
+ * each recorder before the associated Activity is destroyed
+ */
+public class FrameMetricsRecorder {
+  private static final AndroidLogger logger = AndroidLogger.getInstance();
+  private final Activity activity;
+  private final FrameMetricsAggregator fma;
+  private final WeakHashMap<Object, PerfFrameMetrics> subTraceMap;
+  private boolean isRecording = false;
```
jposuna, line 38: I'd recommend making blocks of "types of variables". E.g:
- public static final
- private static final
- public final
- private final
- private

If you separate each "block" by an empty line, it's a lot faster to immediately see if a variable is public/private final, etc because you only need to check which block you're in once, instead of for each variable.  (It's similar to imports vs static imports)

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Provides FrameMetrics data from an Activity's Window. Encapsulates FrameMetricsAggregator.
+ *
+ * <p>Note: each recorder holds a reference to an Activity, so it is very important to dereference
+ * each recorder before the associated Activity is destroyed
+ */
+public class FrameMetricsRecorder {
+  private static final AndroidLogger logger = AndroidLogger.getInstance();
+  private final Activity activity;
+  private final FrameMetricsAggregator fma;
```
jposuna, line 36: I usually try to avoid abbreviating things which aren't typically abbreviated / obvious.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+  private final WeakHashMap<Object, PerfFrameMetrics> subTraceMap;
+  private boolean isRecording = false;
+
+  /**
+   * Creates a recorder for a specific activity
```
jposuna, line 41: Nit: Try to be consistent about ending in a `.` - it's not the end of the world, but if we get consistent about small things like that it'll just naturally be easier to get consistent about other things. Apply throughout.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+  public Optional<PerfFrameMetrics> stop() {
+    if (!isRecording) {
+      logger.error("Cannot stop because no recording was started");
+      return Optional.absent();
+    }
+    Optional<PerfFrameMetrics> data = this.snapshot();
+    try {
+      fma.remove(activity);
+    } catch (IllegalArgumentException err) {
+      logger.debug("View not hardware accelerated. Unable to collect FrameMetrics. %s", err);
```
jposuna, line 84: Is this the only reason we can have an IAE?

jposuna, line 84: I'm surprised that if the message says `Unable to collect FrameMetrics` we are still able to return data.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
jposuna: My 2c is to try to avoid using `Object` and other generic types as much as possible. It makes it easy to misuse APIs. go/java-practices/methods#types

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+    fma.reset();
+    isRecording = false;
+    return data;
+  }
+
+  /**
+   * Starts a sub-trace in the current recording
```
jposuna, line 92: I'm not really sure what a sub-trace is and why it needs to be recorded in this class.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+ * Provides FrameMetrics data from an Activity's Window. Encapsulates FrameMetricsAggregator.
+ *
+ * <p>Note: each recorder holds a reference to an Activity, so it is very important to dereference
+ * each recorder before the associated Activity is destroyed
+ */
+public class FrameMetricsRecorder {
+  private static final AndroidLogger logger = AndroidLogger.getInstance();
+  private final Activity activity;
+  private final FrameMetricsAggregator fma;
+  private final WeakHashMap<Object, PerfFrameMetrics> subTraceMap;
```
jposuna, line 37: Why are you using a WeakHashMap?

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+   * FrameMetricsAggregator}.
+   *
+   * @return the frame metrics
+   */
+  protected Optional<PerfFrameMetrics> snapshot() {
+    if (!isRecording) {
+      logger.warn("No recording has been started.");
+      return Optional.absent();
+    }
+    SparseIntArray[] arr = this.fma.getMetrics();
```
jposuna, line 152: FYI - Object arrays are evil. go/java-practices/choosing-collections#arrays

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,182 @@
...
+    return Optional.of(snapshotEnd.get().subtract(snapshotStart));
+  }
+
+  /**
+   * Calculate total frames, slow frames, and frozen frames from SparseIntArray[] recorded by {@link
+   * FrameMetricsAggregator}.
+   *
+   * @return the frame metrics
+   */
+  protected Optional<PerfFrameMetrics> snapshot() {
```
jposuna, line 147: Why is this protected instead of private?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
jposuna: Can you at least add a javadoc to explain what this method does?

Personally, I don't know why we would be subtracting some metrics from others (as in, what is the use-case?) :)

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
```diff
@@ -0,0 +1,405 @@
...
+public class FrameMetricsRecorderTest extends FirebasePerformanceTestBase {
+  private Activity activity;
+  private FrameMetricsRecorder recorder;
+
+  @Mock private FrameMetricsAggregator fma;
+  @Mock private PerfFrameMetrics frameMetrics1;
+  @Mock private PerfFrameMetrics frameMetrics2;
+  @Mock private PerfFrameMetrics frameMetrics3;
+
+  @Spy private final WeakHashMap<Object, PerfFrameMetrics> subTraceMap = new WeakHashMap<>();
```
jposuna, line 61: This is a code-smell to me. Can we verify things through the public API without accessing internals? go/unit-testing-practices#public-apis

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
```diff
@@ -0,0 +1,405 @@
...
+    verify(fma, times(0)).add(nullable(Activity.class));
+    recorder.start();
+    verify(fma, times(1)).add(nullable(Activity.class));
+  }
+
+  @Test
+  public void start_whileAlreadyStarted_fails() {
+    recorder.start();
+    recorder.start();
+    verify(fma, times(1)).add(nullable(Activity.class));
```
jposuna, line 95: Do we want this to be nullable? Why are we using this argument matcher instead of a different one?

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
```diff
@@ -0,0 +1,405 @@
...
+  public void startSubTrace_whenSameSubTraceWithGivenKeyIsAlreadyOngoing_fails() {
+    doReturn(Optional.of(frameMetrics1)).when(recorder).snapshot();
+    ArgumentCaptor<Object> objectCaptor = ArgumentCaptor.forClass(Object.class);
+    Object uiState1 = new Object();
+    Object uiState2 = new Object();
+
+    recorder.start();
+    recorder.startSubTrace(uiState1);
+    verify(subTraceMap, times(1)).put(objectCaptor.capture(), nullable(PerfFrameMetrics.class));
+    Assert.assertSame(frameMetrics1, subTraceMap.get(uiState1));
```
jposuna, line 162: Use a static import.

jposuna, line 162: Prefer to use the Truth framework consistently rather than mixing frameworks.

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
```diff
@@ -0,0 +1,405 @@
+// Copyright 2022 Google LLC
```
jposuna, line 1: General comment on how this class is being tested: It looks like this class is testing most things through interactions rather than testing behaviors / expected outcomes through public APIs. go/unit-testing-practices#public-apis

I try to write tests from the point of view of a potential caller (i.e. a use-case to use this class). For example, as a caller, I don't care about how this class is implemented - that means that I don't care if a map is called once or twice. What I care about is that the results of my interactions are correct. I'm not sure that the current tests verify that as much as verify implementation details.

To put it differently, these tests are very close to change-detector tests because they are not testing that the class behaves correctly, but rather that the internals of the class have specific interactions.

go/tott/299 has a much better explanation of what I'm trying to say :)

jposuna (changes requested):

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: Yes. Android will try to register a listener on the top level view of the Activity Window to listen for frame times. If the view isn't hardware accelerated, the listener will not be registered but no error will occur. Only when trying to remove it, it will throw IAE when that listener can't be found.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: For testing and mocking in the tests. Ideally things like these implementation detail should be private, but I'm not sure how to test it and mock it if I do that.

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
leotianlizhan: Do you mean all the mocks or just the spy `subtraceMap`?

For `start()` it's hard to verify because it's a `void` function, not sure what's the best way to verify those.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: Good to know. In this case I don't find it actionable because `FrameMetricsAggregator` is a dependency that we can't change.

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
leotianlizhan: Could you elaborate? I'm not sure what you mean in this context.

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
leotianlizhan: That philosophy makes a lot of sense. However I am finding difficulty in achieving those, for example, `start()` and `stop()` are related to each other. `start()` is a `void` function, it does thing so `stop()` will work, but I don't know how to only use outputs to test start individually aside from mocking its implementation detail and seeing behaviors on those mocks (definitely fragile to changes). Similarly `stop()` relies on `start()` in the sense that `stop()` would only work when `start()` has been called prior and correctly, so it'd be hard to test individually without mocking everything in `stop()`.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: Mainly because code in the SDK anywhere that involve Activity / Fragment uses WeakHashMap. I think it's to avoid memory leaks when the class isn't being used properly. If this object is not dereferenced when Activity is no longer visible, since this object holds a reference to that Activity, the Activity won't be collected by the GC. 

The main use of `WeakHashMap` or `WeakReference` I see that people use is for AsyncTasks that uses an Activity, because AsyncTask might execute later when Activity should already be GCed. 

In this case, I will change to `HashMap` because it's not a public API, only we are using this class, so it should be properly dereferenced before Activity is destroyed.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: `FrameMetricsAggregator` collects frame-time from an Activity's Window, for every frame rendered for that Window. Fragment screen trace is only an interval within that same frame-time data from the Window. `sub-trace` may be a bad name but I couldn't think of another word to describe this relationship, open to more suggestions.

This is also the reason I made the parameter `Object`, because from `FrameMetricsAggregator`'s perspective, all it does and all it can do is collect frame-times for an Activity's Window. Fragment screen trace is a "UI state" we arbitrarily added to tag an interval within that frame-time recording. `FrameMetricsAggregator` shouldn't care if you use anything like `Fragment` or other UI objects, all it cares about is that you want a sub-interval of frametimes from time A to time B, and you want to give it a name "Fragment A".

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
leotianlizhan: I guess a followup question is: for methods that are dependent on each other or used in conjunction, is it okay to write tests that tests both at the same time instead of independently?

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: Made it private and test class behaviors instead, following go/unit-testing-practices

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
leotianlizhan: Following go/unit-testing-practices, now testing behaviors not individual methods, so this spy is removed.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
leotianlizhan: On second thought, the recorder class is getting too big and too difficult to test if this is moved over, so reverted it and now `snapshot()` uses `FrameMetricsCalculator`.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
jposuna: Discussed offline. There are 2 main options:

1. Using "fragment" instead of sub-trace throughout this class for now and update it when we have a second use case.
2. Keep it generic, but then we need to add some more javadoc to explain the generic concepts we are using.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * @return FrameMetrics accumulated during the current recording.
+   */
+  public Optional<PerfFrameMetrics> stop() {
+    if (!isRecording) {
+      logger.warn("Cannot stop because no recording was started");
+      return Optional.absent();
+    }
+    if (!subTraceMap.isEmpty()) {
+      logger.warn(
```
visumickey, line 85: Should we return right here? The warnings shows that sub-traces are on-going and recommending to stop the activity screen trace, but we seem to be stopping the metric collection even with this warning.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+    isRecording = false;
+    return data;
+  }
+
+  /**
+   * Starts a sub-trace in the current recording.
+   *
+   * @param key a UI state to associate this sub-trace with. e.g.) fragment
+   */
+  public void startSubTrace(Fragment key) {
```
visumickey, line 106: Can we rename this to `startFragment(Fragment fragment)`. Since the signature is locked, it makes sense to use the same terminology. We can generalize this later. Over engineering is equally challenging to maintain.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+    if (subTraceMap.containsKey(key)) {
+      logger.warn(
+          "Cannot start sub-trace because one is already ongoing with the key %s",
+          key.getClass().getSimpleName());
+      return;
+    }
+    Optional<PerfFrameMetrics> snapshot = this.snapshot();
+    if (!snapshot.isAvailable()) {
+      logger.warn("startSubTrace(%s): snapshot() failed", key.getClass().getSimpleName());
```
visumickey, line 119: This information is not useful for the developers. Can we change this to something the developers can understand? Eg: `startFragment(%s): Unable to capture frame metrics when the fragment was shown on screen`.

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
jposuna: Discussed offline. Summary:

Testing the behavior of the class is what we expect is much more useful than testing the internal interactions of the class. The interactions could be what we think they should be, but they could lead to incorrect outcomes. By testing the behavior we can verify that the class does what we want it to do

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+    Optional<PerfFrameMetrics> snapshotEnd = this.snapshot();
+    if (!snapshotEnd.isAvailable()) {
+      logger.warn("stopSubTrace(%s): snapshot() failed", key.getClass().getSimpleName());
+      return Optional.absent();
+    }
+    return Optional.of(snapshotEnd.get().subtract(snapshotStart));
+  }
+
+  /**
+   * Calculate total frames, slow frames, and frozen frames from SparseIntArray[] recorded by {@link
```
visumickey, line 152: If it is calculate, should we call the method as "calculateMetrics"? Based on what is happening in the function, my recommendation is `snapshotFrameMetrics` and change the comment to `Snapshot frame metrics from Frame Metrics Aggregator`.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/FrameMetricsCalculator.java
```diff
@@ -49,6 +49,19 @@ public int getSlowFrames() {
...
       return totalFrames;
     }
+
+    /**
+     * Subtracts frame-time counts of the argument object (that) from the current object (this).
+     *
+     * @param that the subtrahend PerfFrameMetrics object.
+     * @return difference of this and the argument.
+     */
+    public PerfFrameMetrics subtract(PerfFrameMetrics that) {
```
visumickey, line 59: I still think that this method needs to be renamed to something like `deltaFrameMetricsFromSnapshot(PerfFrameMetrics that)`.

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
```diff
@@ -0,0 +1,261 @@
...
+  public void start_afterPreviousEnded_doesCallFMASuccessfully() {
+    recorder.start();
+    recorder.stop();
+    recorder.start();
+    verify(fma, times(2)).add(any());
+  }
+
+  @Test
+  public void stop_whileNotStarted_returnsEmptyResult() {
+    Optional<PerfFrameMetrics> result;
```
jposuna, line 95: Can we declare it only once it's needed rather than at the top?

firebase-perf/src/test/java/com/google/firebase/perf/application/FrameMetricsRecorderTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+  @Test
+  public void startAndStopSubTrace_whenCalledTwice_ignoresSecondCall() {
+    Fragment fragment = new Fragment();
+    Optional<PerfFrameMetrics> result1;
+    Optional<PerfFrameMetrics> result2;
+
+    // Happens only when we hook on to the same instance of fragment twice. Very unlikely.
+    recorder.start();
+    stubFrameMetricsAggregatorData(fma, new int[][] {{1, 1}, {17, 1}, {800, 1}}); // 3, 2, 1
```
jposuna, line 154: I think it would be helpful to have comments on each of these explaining whether we expect them to be taken into account or not and why

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: I forgot to update the comments. This used to have the logic in `FrameMetricsCalculator` copied over, but now it just uses `FrameMetricsCalculator`, so yes I will update the comment so it's about "snapshotting" and not about "calculating".

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+ * dereference each recorder at or before its Activity's onDestroy. Similar for Fragments.
+ */
+public class FrameMetricsRecorder {
+  private static final AndroidLogger logger = AndroidLogger.getInstance();
+
+  private final Activity activity;
+  private final FrameMetricsAggregator frameMetricsAggregator;
+  private final Map<Fragment, PerfFrameMetrics> subTraceMap;
+
+  private boolean isRecording = false;
```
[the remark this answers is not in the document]
leotianlizhan: I think recording makes more sense because that matches with the class name, and [FrameMetricsAggregator's API also uses the term "recording"](https://developer.android.com/reference/kotlin/androidx/core/app/FrameMetricsAggregator#add(android.app.Activity))

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: This message is meant to be shown to us, not to developers. In addition, `snapshot()` can be called in both `startFragment(...)` and `stopFragment(...)`, so I think it's more clear to just say "snapshot() failed".

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: In this case, the only problem is fragment traces are going to be wrong / fail. Activity trace is still correct, so I didn't return empty and just let it continue.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,174 @@
...
+      FrameMetricsAggregator frameMetricsAggregator,
+      Map<Fragment, PerfFrameMetrics> subTraceMap) {
+    this.activity = activity;
+    this.frameMetricsAggregator = frameMetricsAggregator;
+    this.subTraceMap = subTraceMap;
+  }
+
+  /** Starts recording FrameMetrics for the activity window. */
+  public void start() {
+    if (isRecording) {
```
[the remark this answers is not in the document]
leotianlizhan: I am not sure what value that API would add because I don't see us using it. We are just going to call `start()` in `onActivityStart` and `stop()` in `onActivityStop`, if it fails then we are not doing a symmetrical call and are calling `start()` somewhere else, which is wrong and it should output warning.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: Using "fragment" for now. Will generalize it in the future if needed.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
jposuna: +1 to Leo

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
jposuna: +1 to Leo

jposuna (approved):
Thanks for working on this and spending the time to make the changes!

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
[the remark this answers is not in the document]
visumickey: I'm mixed on this. I have seen usecases where having such an API would be very useful. But in the interest of self-containment, I'm fine to move ahead here.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
visumickey: Sounds good. That makes sense. In that case, since the log message is not actionable for the developer, can we turn the level to "Debug"?

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: discussed offline and changing it to `logger.debug` would avoid developers seeing it, yet we still get the precise information that `snapshot()` failed.

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -0,0 +1,185 @@
...
+   * Stops recording FrameMetrics for the activity window.
+   *
+   * @return FrameMetrics accumulated during the current recording.
+   */
+  public Optional<PerfFrameMetrics> stop() {
+    if (!isRecording) {
+      logger.debug("Cannot stop because no recording was started");
+      return Optional.absent();
+    }
+    if (!fragmentSnapshotMap.isEmpty()) {
```
jeremyjiang-dev, line 91: Do we need to clear the `fragmentSnapshotMap`? If not, when we start another activity and want to call `startFragment(Fragment fragment)`, the map could be non-empty. Therefore, we will not be able to start certain fragment.

leotianlizhan: Good point! Added `fragmentSnapshotMap.clear()`.

jeremyjiang-dev (approved):
Thanks Leo for your hard work! Great work!

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
jposuna: Can you add a test that would trigger this behavior so we don't accidentally reintroduce the bug in the future?

leotianlizhan: Done!

PR #3668, opened 2022-04-20
Fix a bug with limitToLast and cursors

wu-hui:
Fix for android for: https://github.com/firebase/firebase-js-sdk/issues/6134


cherylEnkidu (approved):
This looks great! Thank you :)

PR #3683, opened 2022-04-26
Fireperf: use the new class `FrameMetricsRecorder` in `AppStateMonitor` and `FragmentStateMonitor`

leotianlizhan:
With `FrameMetricsRecorder` taking care of managing `FrameMetricsAggregator`, the need for `AppStateMonitor` to use FrameMetricsAggregator is gone. This PR is to make change to AppStateMonitor to use the `FrameMetricsRecorder` for measuring frame metrics.

As a part of this, the lifecycle event used to measure the start and stop of an activity screen trace is aligned to make it work for all versions of Android.


firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,79 +88,75 @@ public void setUp() {
...
+        "_st_FragmentMonitorMockFragment", monitor.getFragmentScreenTraceName(testFragment));
+  }
+
   @Test
   public void lifecycleCallbacks_differentFrameMetricsCapturedByFma_logFragmentScreenTrace() {
     FragmentStateMonitor monitor =
-        new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, fma);
-    when(fma.getMetrics()).thenReturn(fmaMetrics1);
+        new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, recorder);
+    doReturn(Optional.of(frameCounts1)).when(recorder).stopFragment(any());
```
jeremyjiang-dev, line 123: Is there a specific reasons that you want to use `doReturn..when` instead of `when...thenReturn`? 
go/java-testing/mocking#stubbing-with-when-does-not-always-work suggests "Tip: Use Mockito.doReturn(value).when(mockObject).method(arguments) in cases where when(...) does not work. Avoid it otherwise (style rule, it is harder to read)."
Applies to other places in this test file.

jeremyjiang-dev (commented):
Thanks for making these changes! I wonder if you can test the fragment / activity screen trace in a sample app to see if we can dispatch the events as expected?

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -149,28 +150,64 @@ public void incrementTsnsCount(int value) {
     tsnsCount.addAndGet(value);
   }
 
-  @Override
-  public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+  // Starts tracking the frame metrics for an activity.
+  public void startFrameMonitoring(Activity activity) {
```
jposuna, line 154: Does this need to be public?

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -149,28 +150,64 @@ public void incrementTsnsCount(int value) {
...
+      fragmentActivity
+          .getSupportFragmentManager()
+          .unregisterFragmentLifecycleCallbacks(activityToFragmentStateMonitorMap.remove(activity));
+    }
+  }
+
+  @Override
+  public synchronized void onActivityStarted(Activity activity) {
+    if (isScreenTraceSupported() && configResolver.isPerformanceMonitoringEnabled()) {
+      if (activityToRecorderMap.get(activity) == null) {
```
jposuna, line 191: This is more directly expressed as `activityToRecorderMap.containsKey(activity)`

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -149,28 +150,64 @@ public void incrementTsnsCount(int value) {
...
+    if (activityToFragmentStateMonitorMap.containsKey(activity)) {
+      FragmentActivity fragmentActivity = (FragmentActivity) activity;
+      fragmentActivity
+          .getSupportFragmentManager()
+          .unregisterFragmentLifecycleCallbacks(activityToFragmentStateMonitorMap.remove(activity));
+    }
+  }
+
+  @Override
+  public synchronized void onActivityStarted(Activity activity) {
```
jposuna, line 189: Nit: I'd probably put `onActivityStarted` before `onActivityDestroyed` since it more closely follows the lifecycle, and it's easier to go "chronologically" than jumping forwards and backwards. (And I'd apply this to the rest of the overriden methods too).

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -315,54 +338,25 @@ public void onActivitySaveInstanceState(Activity activity, Bundle outState) {}
...
-  @Override
-  public void onActivityPostPaused(@NonNull Activity activity) {
-    if (isScreenTraceSupported()) {
-      sendScreenTrace(activity);
-    }
-  }
-
   /**
-   * Send screen trace. If hardware acceleration is not enabled, all frame metrics will be zero and
-   * the trace will not be sent.
```
jposuna, line 328: Is this comment no longer true?

firebase-perf/src/test/java/com/google/firebase/perf/application/AppStateMonitorTest.java
```diff
@@ -15,6 +15,7 @@
 package com.google.firebase.perf.application;
```
jposuna: The behavior of AppStateMonitor has changed quite a bit. I'm surprised the tests haven't changed barely at all. I feel like we're probably missing tests (e.g. testing new behaviors), as well as having coverage for the behavior changes that have happened (some things happen in different points of the activity's lifecycle)

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -169,116 +165,101 @@ public void lifecycleCallbacks_onPausedCalledBeforeOnResume_doesNotLogFragmentSc
...
   }
 
+  /** Simulate call order of activity + fragment lifecycle events */
   @Test
   public void lifecycleCallbacks_cleansUpMap_duringActivityTransitions() {
-    // Simulate call order of activity + fragment lifecycle events
-    AppStateMonitor appStateMonitor =
-        spy(new AppStateMonitor(mockTransportManager, clock, configResolver, fma));
+    Bundle savedInstanceState = mock(Bundle.class);
+    AppStateMonitor appStateMonitor = mock(AppStateMonitor.class);
```
jposuna, line 193: Why do we need to create more mocks within a test? Can't we use the one created at the top with `@Mock`?

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -90,22 +94,18 @@ public static AppStateMonitor getInstance() {
...
   AppStateMonitor(TransportManager transportManager, Clock clock) {
-    this(
-        transportManager,
-        clock,
-        ConfigResolver.getInstance(),
-        hasFrameMetricsAggregatorClass() ? new FrameMetricsAggregator() : null);
+    this(transportManager, clock, ConfigResolver.getInstance(), hasFrameMetricsAggregatorClass());
   }
 
   AppStateMonitor(
```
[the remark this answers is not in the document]
visumickey: Done. Makes sense and a good idea.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
visumickey: Currently they are organized in pairs. (Created, Destroyed) and (Started, Stopped). So organizing by lifecycle events might break it a bit for understanding. Some of the operation performed at `onActivityCreated` are negated at `onActivityDestroyed` and similar for start and stop.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
visumickey: This comment is true, but this is not relevant here. This has moved to FrameMetricsRecorder. So removed it.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -315,54 +327,25 @@ public void onActivitySaveInstanceState(Activity activity, Bundle outState) {}
...
+   * Sends the screen trace for the provided activity.
    *
    * @param activity activity object.
    */
   private void sendScreenTrace(Activity activity) {
-    if (!activityToScreenTraceMap.containsKey(activity)) {
-      return;
-    }
     Trace screenTrace = activityToScreenTraceMap.get(activity);
     if (screenTrace == null) {
```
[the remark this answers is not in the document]
visumickey: It is possible if there had been a toggle in the performance Monitoring enabled state at runtime. This check is mainly to protect such situations.

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
```diff
@@ -96,33 +91,18 @@ public void onFragmentPaused(@NonNull FragmentManager fm, @NonNull Fragment f) {
...
-        FrameMetricsCalculator.calculateFrameMetrics(this.frameMetricsAggregator.getMetrics());
-
-    int totalFrames = curPerfFrameMetrics.getTotalFrames() - prePerfFrameMetrics.getTotalFrames();
-    int slowFrames = curPerfFrameMetrics.getSlowFrames() - prePerfFrameMetrics.getSlowFrames();
-    int frozenFrames =
-        curPerfFrameMetrics.getFrozenFrames() - prePerfFrameMetrics.getFrozenFrames();
 
-    if (totalFrames == 0 && slowFrames == 0 && frozenFrames == 0) {
-      // All metrics are zero, no need to send screen trace.
+    Optional<PerfFrameMetrics> data = activityFramesRecorder.stopFragment(f);
```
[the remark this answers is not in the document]
visumickey: Done.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,79 +88,75 @@ public void setUp() {
...
+
+    doNothing().when(recorder).start();
+    doNothing().when(recorder).startFragment(any());
 
-    ConfigResolver configResolver = ConfigResolver.getInstance();
-    configResolver.setDeviceCacheManager(new DeviceCacheManager(new FakeDirectExecutorService()));
-    ConfigResolver spyConfigResolver = spy(configResolver);
-    doReturn(true).when(spyConfigResolver).isPerformanceMonitoringEnabled();
-    this.configResolver = spyConfigResolver;
+    // doReturn(Optional.of(frameCounts1)).when(recorder).stopFragment(any());
```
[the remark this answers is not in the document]
visumickey: Done

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,79 +88,75 @@ public void setUp() {
...
+    // doReturn(Optional.of(frameCounts2)).when(recorder).stopFragment(any());
+
+    activity = createFakeActivity(true);
 
     // fmaMetrics1 should have 1+3+1=5 total frames, 3+1=4 slow frames, and 1 frozen frames.
```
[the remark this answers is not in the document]
visumickey: Done.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,79 +88,75 @@ public void setUp() {
...
-    SparseIntArray sparseIntArray = new SparseIntArray();
-    sparseIntArray.append(1, 1);
-    sparseIntArray.append(17, 3);
-    sparseIntArray.append(800, 1);
-    fmaMetrics1[FrameMetricsAggregator.TOTAL_INDEX] = sparseIntArray;
+    frameCounts1 = new PerfFrameMetrics(9, 5, 3);
```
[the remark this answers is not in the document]
visumickey: Honestly, we have got away from the specific of those frame counts here. All we receive here is just the overall count. So, not sure if we need more details here.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,79 +88,75 @@ public void setUp() {
...
 
     // fmaMetrics2 should have 5+5+4=14 total frames, 5+4=9 slow frames, and 4 frozen frames.
```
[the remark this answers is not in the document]
visumickey: Done

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,79 +88,75 @@ public void setUp() {
...
+  public void fragmentTraceName_validFragment_validFragmentScreenTraceNameGenerated() {
+    FragmentStateMonitor monitor =
+        new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, recorder);
+    Fragment testFragment = new FragmentMonitorMockFragment();
+    Assert.assertEquals(
+        "_st_FragmentMonitorMockFragment", monitor.getFragmentScreenTraceName(testFragment));
+  }
+
   @Test
   public void lifecycleCallbacks_differentFrameMetricsCapturedByFma_logFragmentScreenTrace() {
```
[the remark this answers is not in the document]
visumickey: Good idea! Done.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
visumickey: Doesn't have to be. Marking this as `private`.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
visumickey: Got it. Done!

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
visumickey: Makes sense. Updated it.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
visumickey: Good reference point. Thanks Jeremy!

visumickey: Something I learnt. Mockito.doReturn when used on spied objects, does not make the function call at all. Whereas, Mockito.when will call the function and if there are exceptions, we need to handle that. So had to revert back this change in one of the places.

firebase-perf/src/test/java/com/google/firebase/perf/application/AppStateMonitorTest.java
leotianlizhan: The only lifecycle change is reverting the activity screen trace that is from `onResume` -> `onPostPaused` to `onStart` -> `onStop`. `onStart` -> `onStop` was the original behavior before attempting to fix the b/210055697 that started all of this in the first place. That's why there's not new tests for that.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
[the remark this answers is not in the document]
leotianlizhan: I removed this trace cause the line that says `if (screenTrace == null) {` already handles it, so this check is redundant.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
visumickey: A good thread on that is here - https://stackoverflow.com/questions/20353846/mockito-difference-between-doreturn-and-when

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -90,22 +94,18 @@ public static AppStateMonitor getInstance() {
   }
 
   AppStateMonitor(TransportManager transportManager, Clock clock) {
-    this(
-        transportManager,
-        clock,
-        ConfigResolver.getInstance(),
-        hasFrameMetricsAggregatorClass() ? new FrameMetricsAggregator() : null);
+    this(transportManager, clock, ConfigResolver.getInstance(), hasFrameMetricsAggregatorClass());
```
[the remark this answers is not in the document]
visumickey: Thanks for highlighting this. Was punting on this thinking it is a big change. But now, managed to make the change relatively easily.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -315,54 +327,25 @@ public void onActivitySaveInstanceState(Activity activity, Bundle outState) {}
...
-    } catch (IllegalArgumentException ignored) {
-      logger.debug("View not hardware accelerated. Unable to collect screen trace.");
-    }
-    FrameMetricsCalculator.PerfFrameMetrics perfFrameMetrics =
-        FrameMetricsCalculator.calculateFrameMetrics(frameMetricsAggregator.reset());
-    if (perfFrameMetrics.getTotalFrames() == 0
-        && perfFrameMetrics.getSlowFrames() == 0
-        && perfFrameMetrics.getFrozenFrames() == 0) {
-      // All metrics are zero, no need to send screen trace.
+    Optional<FrameMetricsCalculator.PerfFrameMetrics> perfFrameMetrics =
```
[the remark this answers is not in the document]
visumickey: Good idea, done.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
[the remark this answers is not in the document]
jeremyjiang-dev: Makes sense! Thanks!

jeremyjiang-dev (approved):
LGTM on checks green!

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
```diff
@@ -49,13 +49,17 @@
 public class AppStateMonitor implements ActivityLifecycleCallbacks {
 
   private static final AndroidLogger logger = AndroidLogger.getInstance();
-  private static final String FRAME_METRICS_AGGREGATOR_CLASSNAME =
-      "androidx.core.app.FrameMetricsAggregator";
 
   private static volatile AppStateMonitor instance;
-  private static boolean hasFrameMetricsAggregator = false;
+  private final boolean screenPerformanceRecordingSupported;
```
jposuna, line 54: Should this be moved to the block below (since it's no longer static, and is now final)?

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -88,197 +92,176 @@ public void setUp() {
...
 
   @Test
-  public void lifecycleCallbacks_differentFrameMetricsCapturedByFma_logFragmentScreenTrace() {
+  public void fragmentTraceName_validFragment_validFragmentScreenTraceNameGenerated() {
     FragmentStateMonitor monitor =
-        new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, fma);
-    when(fma.getMetrics()).thenReturn(fmaMetrics1);
+        new FragmentStateMonitor(clock, mockTransportManager, appStateMonitor, recorder);
+    Fragment testFragment = new FragmentMonitorMockFragment();
+    Assert.assertEquals(
```
jposuna, line 119: Can we use the truth framework? (assertThat...)

I'd recommend using it consistently whenever we touch a test. That way eventually all / most tests will be using it.

jposuna (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateMonitor.java
visumickey: Done.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
visumickey: Good idea. Moved out all such asserts to use truth framework.

PR #3689, opened 2022-04-28
Minor refactor of IndexBackfiller to eliminate private property.

tom-andersen:
This is a very minor refactor to simplify code:
- removes 2 lines of code.
- removes 1 private property.
- removes conditional evaluation on every execution.
- removes assignment on every execution.
- adds 1 parameter to private method.

There should be no side effects, since the IndexBackfiller never appears to start after a being stopped. If for some reason the IndexBackfiller should start again, there will simply be a larger delay in subsequent starts.


wu-hui (approved):
Thanks for doing this!

PR #3693, opened 2022-05-02
Save token on a separate thread.

gsakakihara:
* Changed blockingGetToken() to save the token on an Executor thread to avoid doing file IO on the main thread.

### DIFF
--- a/firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
+++ b/firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
@@ -13,6 +13,7 @@
 // limitations under the License.
 package com.google.firebase.messaging;
 
+import static com.google.firebase.messaging.FcmExecutors.newFileIOExecutor;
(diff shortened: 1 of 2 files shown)


willxu-google (approved):

firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
```diff
@@ -195,7 +197,8 @@ static synchronized FirebaseMessaging getInstance(@NonNull FirebaseApp firebaseA
         new GmsRpc(
             firebaseApp, metadata, userAgentPublisher, heartBeatInfo, firebaseInstallationsApi),
         /* taskExecutor= */ newTaskExecutor(),
-        /* fileIoExecutor= */ newInitExecutor());
+        /* initExecutor= */ newInitExecutor(),
+        /* fileExecutor= */ newFileIOExecutor());
```
James201311, line 201: nit: how about renaming it to `newFileExecutor`?

James201311 (approved):
LGTM

PR #3717, opened 2022-05-11
Perf: add "No parent" constant string when parent fragment does not exist

leotianlizhan:
b/231632433

### Change
Add `No parent` constant string as the value of `Parent_fragment` custom attribute when a fragment does not have a parent fragment.

### Rationale
The root fragment does not have a parent fragment. In those cases, the Fireperf console has a jarring UX.

For example, if Fragment `C` is used in 2 places in an app. One as a child of Fragment `P`, and one as a root fragment without parent. The console would display only 1 option for `Parent_fragment`, which is `P`. However in the filter and the graph, you can see some data for `All data`, and when you filter by `Parent_fragment`, you can see some data for fragments with parent `P`. That leads users to think: `C` is only used in one place in my app, and that is as a child of `P`, so why is the data for `All data` and parent `P` different?

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
@@ -66,11 +66,10 @@ public class FragmentStateMonitor extends FragmentManager.FragmentLifecycleCallb
         new Trace(getFragmentScreenTraceName(f), transportManager, clock, appStateMonitor);
     fragmentTrace.start();
 
-    if (f.getParentFragment() != null) {
-      fragmentTrace.putAttribute(
-          Constants.PARENT_FRAGMENT_ATTRIBUTE_KEY,
-          f.getParentFragment().getClass().getSimpleName());
-    }
+    Fragment parent = f.getParentFragment();
+    fragmentTrace.putAttribute(
+        Constants.PARENT_FRAGMENT_ATTRIBUTE_KEY,
+        parent == null ? Constants.NO_PARENT_FRAGMENT : parent.getClass().getSimpleName());
     if (f.getActivity() != null) {
       fragmentTrace.putAttribute(
           Constants.ACTIVITY_ATTRIBUTE_KEY, f.getActivity().getClass().getSimpleName());
--- a/firebase-perf/src/main/java/com/google/firebase/perf/util/Constants.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/util/Constants.java
@@ -58,6 +58,9 @@ public class Constants {
   /** Attribute key for the hosting activity of a fragment screen trace. */
   public static final String ACTIVITY_ATTRIBUTE_KEY = "Hosting_activity";
 
+  /** Attribute value for when the current fragment does not have a parent fragment. */
+  public static final String NO_PARENT_FRAGMENT = "No parent";
+
   /** frames longer than 16 ms are slow frames */
   public static final int SLOW_FRAME_TIME = 16;
 


firebase-perf/src/main/java/com/google/firebase/perf/util/Constants.java
```diff
@@ -58,6 +58,9 @@ public class Constants {
   /** Attribute key for the hosting activity of a fragment screen trace. */
   public static final String ACTIVITY_ATTRIBUTE_KEY = "Hosting_activity";
 
+  /** Attribute value for when the current fragment does not have a parent fragment. */
+  public static final String NO_PARENT_FRAGMENT = "No parent";
```
jeremyjiang-dev, line 62: nit: rename to FRAGMENT_CUSTOM_ATTRIBUTE_VALUE_NO_PARENT_FRAGMENT?

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
```diff
@@ -66,11 +66,10 @@ public void onFragmentResumed(@NonNull FragmentManager fm, @NonNull Fragment f)
...
 
-    if (f.getParentFragment() != null) {
-      fragmentTrace.putAttribute(
-          Constants.PARENT_FRAGMENT_ATTRIBUTE_KEY,
-          f.getParentFragment().getClass().getSimpleName());
-    }
+    Fragment parent = f.getParentFragment();
+    fragmentTrace.putAttribute(
+        Constants.PARENT_FRAGMENT_ATTRIBUTE_KEY,
+        parent == null ? Constants.NO_PARENT_FRAGMENT : parent.getClass().getSimpleName());
```
jeremyjiang-dev, line 72: Could you add a test to check the parent_fragment attribute value when there is no parent fragment?

jeremyjiang-dev (approved):

firebase-perf/src/main/java/com/google/firebase/perf/util/Constants.java
leotianlizhan: how about PARENT_FRAGMENT_ATTRIBUTE_VALUE_NONE?

jeremyjiang-dev: Makes sense to me!

firebase-perf/src/main/java/com/google/firebase/perf/application/FragmentStateMonitor.java
leotianlizhan: Sounds good, added `fragmentTraceCreation_addsSpecialAttributeValue_whenFragmentDoesNotHaveParent` in `FragmentStateMonitorTest`

firebase-perf/src/main/java/com/google/firebase/perf/util/Constants.java
```diff
@@ -58,6 +58,9 @@ public class Constants {
   /** Attribute key for the hosting activity of a fragment screen trace. */
   public static final String ACTIVITY_ATTRIBUTE_KEY = "Hosting_activity";
 
+  /** Attribute value for when the current fragment does not have a parent fragment. */
+  public static final String PARENT_FRAGMENT_ATTRIBUTE_VALUE_NONE = "No parent";
```
visumickey, line 62: nit: Can we rename this as NO_PARENT_FRAGMENT_ATTRIBUTE_VALUE = "No parent"

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
```diff
@@ -243,6 +244,21 @@ public void fragmentTraceCreation_dropsTrace_whenFragmentNameTooLong() {
     verify(mockTransportManager, times(0)).log(any(TraceMetric.class), any());
   }
 
+  @Test
+  public void fragmentTraceCreation_addsSpecialAttributeValue_whenFragmentDoesNotHaveParent() {
```
visumickey, line 248: Can we name this as `fragmentTraceCreation_NoParentFragment_addsNoParentAttributeValue`?

Reference: go/unit-testing-practices#naming

visumickey, line 248: Curious - Do we have tests for custom attributes when there is a parent fragment available?

visumickey (approved):

jeremyjiang-dev (approved):

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
leotianlizhan: `addsNoParentAttributeValue` could be ambiguous (can sound like not adding anything), which is why I used the wording "special attribute value"

leotianlizhan: Good question, I don't believe we do

firebase-perf/src/main/java/com/google/firebase/perf/util/Constants.java
visumickey: Sounds good.

firebase-perf/src/test/java/com/google/firebase/perf/application/FragmentStateMonitorTest.java
visumickey: The new name you proposed makes sense to me.

Issue #3056, opened 2021-10-14
Override behavior of In-App message


garispe:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: Arctic Fox 2020.3.1 Patch 3
  * Firebase Component: In-App Messaging
  * Component version: com.google.firebase:firebase-bom:28.3.0

#### Steps to reproduce:

I created a Firebase In-App messaging campaign from console and I used `addClickListener()` to override behavior of primary button action. When I receive the in-app message and click primary button, my custom listener is invoked correctly but the Firebase SDK opens the link from campaign.

I need to attach dynamic information to the link configured in the Firebase console and then open the composed link.

The problem is in [this line](https://github.com/firebase/firebase-android-sdk/blob/e1f356a87ed506f765df4f4a2f3c93daae44f2aa/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java#L322) because when I override the click behavior, the SDK invokes `launchUriIntent(activity, Uri.parse(action.getActionUrl()));` anyway

#### Relevant Code:

I think that the fix can be something like that:

```java
@Override
public void onClick(View v) {
    if (callbacks != null) {
        Logging.logi("Calling callback for click action");
        callbacks.messageClicked(action);
    } else {    <====== THIS ELSE
        launchUriIntent(activity, Uri.parse(action.getActionUrl()));
    }
   
    notifyFiamClick();
    // Ensure that we remove the displayed FIAM, and ensure that on re-load, the message
    // isn't re-displayed
    removeDisplayedFiam(activity);
    inAppMessage = null;
    callbacks = null;
}
```

In this way, the SDK delegates the behavior to the custom listener

garispe:
Hi @aguatno Thanks for your reply. I can create the corresponding PR but I don't know if it's the expected behavior from Firebase because iOS works in the same way. If this is correct, I'll need to manage this issue locally in my app. Do you think that is an unexpected behavior?

AbdulrhmanKassemSalla:
I faced the same issue 
if I used this listener "messageClicked" I don't want any other behaviour 
@garispe your solution is perfect

garispe:
@AbdulrhmanKassemSalla I already opened a PR to fix this

argzdev:
Hi @garispe, thank you for submitting a PR for this! I was able to test this and it works nicely.

[...]

PR #3719, opened 2022-05-12
FIAM: Invoke launchUriIntent only if needed

garispe:
Related issue: #3056


eldhosembabu:
As mentioned in the issue comment here: https://github.com/firebase/firebase-android-sdk/issues/3056#issuecomment-1127993892 we cannot take in this change for the time being now as this is going to be a change in functionality (and not a bug fix) and will be considered as a new feature from our side which needs evaluation and approval. We will add this to our feature backlog and you might be able to see something similar in our feature releases.

Issue #3716, opened 2022-05-11
Crash in datatransport com.google.android.datatransport.runtime.logging.Logging 


gevorg-kopalyan:
### [READ] Step 1: Are you in the right place?

yes

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: Android Studio Bumblebee | 2021.1.1 Patch 3
  * Firebase Component:  Analytics, Auth, Performance, Dynamic links, Crashlytics, Remote Config, Messaging 
  * Component version: Bom 30.0.0

#### Steps to reproduce:

Getting crashes in google play console but not in Firebase Crashlytics 

```
java.lang.IllegalArgumentException: Log tag "TransportRuntime.Executor" exceeds limit of 23 characters

  at android.util.Log.isLoggable (Native Method)
  at com.google.android.datatransport.runtime.logging.Logging.e (Logging.java:63)
  at com.google.android.datatransport.runtime.SafeLoggingExecutor$SafeLoggingRunnable.run (SafeLoggingExecutor.java:49)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1133)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:607)
  at java.lang.Thread.run (Thread.java:776)
```

In com.google.android.datatransport:transport-runtime Library which firebase datatrasport library is using has a new  Log.isLoggable() check which doesn't check for a tag limit of 23 characters and can produce crash.
Here is the [pull request](https://github.com/firebase/firebase-android-sdk/pull/3648) that added new Log.isLoggable check

darrentaft:
Glad it's not just me! Had to immediately roll back to 29.3.1. Thanks for raising.

argzdev:
Thanks for reporting, @gevorg-kopalyan, @darrentaft. I'll notify our engineers and see what we can do here.

vlad-roid:
This is **bad**, it's hitting affected users repeatedly. So far only 11 people in our app are affected, but each person has been hit by the crash over 50 times! I assume this happens on app trying to auto-start, I can't imagine someone having the patience to try and open an app 50 times.

![image](https://user-images.githubusercontent.com/26101312/168507296-a4aedacc-61a3-4b72-9fa4-729467b6f469.png)

AliA74:
[...]

PR #3728, opened 2022-05-16
Shortened prefix and capped concatenated string

argzdev:
Fixes https://github.com/firebase/firebase-android-sdk/issues/3716.

Android version 7.0 has a limit of 23 characters when using isLoggable. Since this limit cannot be bypassed or changed, the preferred solution is to shorten the prefix and cap the concatenated string.

The concatenated string with prefix would look like this. Limited at 23 chars.
e.g.
```
TRuntime.SQLiteEventSto
TRuntime.CctTransportBa
TRuntime.AlarmManagerSc
TRuntime.JobInfoSchedul
TRuntime.Uploader
TRuntime.Executor
```

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
@@ -14,13 +14,26 @@
 
 package com.google.android.datatransport.runtime.logging;
 
+import android.os.Build;
 import android.util.Log;
 
 public final class Logging {
   private Logging() {}
 
   private static String getTag(String tag) {
-    return "TransportRuntime." + tag;
+    String prefix = "TRuntime.";
+    return prefix + concatTag(prefix, tag);
+  }
+
+  private static String concatTag(String prefix, String tag){
+    int loggableLimit = 23;
+    int remainingSpace = loggableLimit - prefix.length();
+
+    if(android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.N)
+      if(tag.length() > remainingSpace)
+        tag = tag.substring(0, remainingSpace);
+
+    return tag;
   }
 
   public static void d(String tag, String message) {


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -14,13 +14,26 @@
...
+import android.os.Build;
 import android.util.Log;
 
 public final class Logging {
   private Logging() {}
 
   private static String getTag(String tag) {
-    return "TransportRuntime." + tag;
+    String prefix = "TRuntime.";
+    return prefix + concatTag(prefix, tag);
```
rlazo, line 25: function name suggests that concatenation is happening, but you're not actually concatenating anything.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -14,13 +14,26 @@
...
+  }
+
+  private static String concatTag(String prefix, String tag){
+    int loggableLimit = 23;
+    int remainingSpace = loggableLimit - prefix.length();
+
+    if(android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.N)
```
rlazo, line 32: to prevent an unnecessary function call, I'd make the function always cap to max 23, and move the `os.Build` check to `getTag`.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -14,13 +14,26 @@
...
 
+import android.os.Build;
 import android.util.Log;
 
 public final class Logging {
   private Logging() {}
 
   private static String getTag(String tag) {
-    return "TransportRuntime." + tag;
+    String prefix = "TRuntime.";
```
rlazo, line 24: for convenience make this a class level constant

argzdev: Got it

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
argzdev: Got it.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -18,22 +18,24 @@
 import android.util.Log;
 
 public final class Logging {
+  private static final String prefix = "TRuntime.";
```
rlazo, line 21: Constants are usually UPPERCASE

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -18,22 +18,24 @@
...
-    return prefix + concatTag(prefix, tag);
+    if(android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.N)
+      return concatTag(prefix, tag);
+
+    return prefix + tag;
   }
 
   private static String concatTag(String prefix, String tag){
+    String concatText = prefix + tag;
     int loggableLimit = 23;
```
rlazo, line 33: This also looks like a constant. Use a name that reflects that the limit is for N and below, maybe LOGGING_TAG_MAX_SIZE_SDK_N (I'm not that good with names).

argzdev: Got it. I've used the similar format from other constants in other files, where the `MAX` is the starting keyword for the constants.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
argzdev: Noted.

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/logging/Logging.java
```diff
@@ -27,7 +27,7 @@ private static String getTag(String tag) {
     if(android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.N)
       return concatTag(LOG_PREFIX, tag);
 
-    return MAX_LOG_TAG_SIZE_IN_SDK_N + tag;
+    return LOG_PREFIX + tag;
```
rlazo, line 30: I think this was unintended? otherwise also update the constant :-)

argzdev: Yup, sorry about that. I incorrectly used the constant for the limit instead of the prefix.

argzdev: I've updated the constant tag on the previous commit, sorry I created 2 commits.

rlazo (approved):

vkryachko:
Please format the code with `./gradlew :transport:transport-runtime:googleJavaFormat`

PR #3743, opened 2022-05-23
Rc realtime dev

qdpham13:
Adds:

Realtime RC based exceptions
Http stream connection via HttpUrlConnection
Http helper methods that creates request and fills in headers (Taken from ConfigFetchHttpClient)
Template Version Number to ConfigContainer
Unit tests
Auto fetch file - Fetch until we receive new template version or we run out of attempts


firebase-config/api.txt
```diff
@@ -65,6 +65,16 @@ package com.google.firebase.remoteconfig {
     method public int getLastFetchStatus();
   }
 
+  public class FirebaseRemoteConfigRealtimeUpdateFetchException extends com.google.firebase.remoteconfig.FirebaseRemoteConfigException {
+    ctor public FirebaseRemoteConfigRealtimeUpdateFetchException(@NonNull String);
+    ctor public FirebaseRemoteConfigRealtimeUpdateFetchException(@NonNull String, @Nullable Throwable);
+  }
+
+  public class FirebaseRemoteConfigRealtimeUpdateStreamException extends com.google.firebase.remoteconfig.FirebaseRemoteConfigException {
```
vkryachko, line 73: What is the semantic difference between these 2 exceptions? Do we expect devs to understand this difference or is it part of the implementation detail?

qdpham13: The difference is that the fetch exception is meant to indicate to the user that we got a new config update signal but we were unable to retrieve the latest config version that we were expecting. And the second one indicated that the stream was unable to open. We'll let users know about these exceptions so that they can expect them when writing their onError callback. We'll include these exceptions as part of our docs.

rlazo: Just for reference, do you have an example of different actions that would make sense to take depending on which exception is thrown?

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigRealtimeUpdateFetchException.java
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.remoteconfig;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+/**
+ * A Firebase Remote Config issue that is caused by the auto-fetch component of Realtime config
```
rlazo, line 21:
```suggestion
 * A Firebase Remote Config exception caused by the auto-fetch component of Realtime config
```

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigRealtimeUpdateStreamException.java
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.remoteconfig;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+/**
+ * A Firebase Remote Config issue that is caused by the stream component of Realtime config updates.
```
rlazo, line 21:
```suggestion
 * A Firebase Remote Config exception caused by the stream component of Realtime config updates.
```

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -0,0 +1,175 @@
...
+  private void listenForNotifications() {
+    if (httpURLConnection != null) {
+      try {
+        int responseCode = httpURLConnection.getResponseCode();
+        if (responseCode == 200) {
+          InputStream inputStream = httpURLConnection.getInputStream();
+          handleNotifications(inputStream);
+          inputStream.close();
+        } else {
+          for (ConfigUpdateListener listener : eventListener) {
```
rlazo, line 87: Nit: Maybe it would make sense to refactor this block to a method like `broadcastErrorToListeners` since it's used in several places.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -0,0 +1,175 @@
...
+    String message;
+    while ((message = reader.readLine()) != null) {
+      long targetTemplateVersion = configFetchHandler.getTemplateVersionNumber();
+      try {
+        JSONObject jsonObject = new JSONObject(message);
+        if (jsonObject.has("latestTemplateVersionNumber")) {
+          targetTemplateVersion = jsonObject.getLong("latestTemplateVersionNumber");
+        }
+      } catch (JSONException ex) {
+        Log.i(TAG, "Can't get latest config update message.");
```
rlazo, line 116: To be strict, you could get the config, but you couldn't parse it.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  private void autoFetch(int remainingAttempts, long targetVersion) {
+    if (remainingAttempts == 0) {
+      for (ConfigUpdateListener listener : eventListener) {
+        listener.onError(
+            new FirebaseRemoteConfigRealtimeUpdateFetchException(
+                "Unable to fetch latest version."));
+      }
+
```
rlazo, line 131: rm

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -434,6 +435,16 @@ private static ConfigContainer extractConfigs(JSONObject fetchResponse, Date fet
         containerBuilder.withPersonalizationMetadata(personalizationMetadata);
       }
 
+      String templateVersionNumber = null;
+      try {
+        templateVersionNumber = fetchResponse.getString(TEMPLATE_VERSION_NUMBER);
```
rlazo, line 440: Maybe `optString` https://developer.android.com/reference/org/json/JSONObject#optString(java.lang.String,%20java.lang.String)

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigRealtimeUpdateFetchException.java
qdpham13: Done. 
And for the stream exception some things they could do is remove their listener, fetch again, or add a new listener to trigger another round of http stream retries.
And for the fetch exception users could try and fetch again or activate their default configs.

firebase-config/api.txt
qdpham13: For the stream exception some things they could do is remove their listener, fetch again, or add a new listener to trigger another round of http stream retries.
And for the fetch exception users could try and fetch again or activate their default configs.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
qdpham13: Updated comment to include parsing error

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigRealtimeUpdateStreamException.java
qdpham13: Done.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
qdpham13: Just following the convention that was left for the other response objects.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -0,0 +1,179 @@
...
+    executor.execute(
+        new Runnable() {
+          @Override
+          public void run() {
+            listenForNotifications();
+          }
+        });
+  }
+
+  private void propagateErrors(FirebaseRemoteConfigException exception) {
```
vkryachko, line 80: synchronized as well?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -14,32 +14,257 @@
...
 
   @GuardedBy("this")
   private final Set<ConfigUpdateListener> listeners;
 
-  public ConfigRealtimeHttpClient() {
-    listeners = new LinkedHashSet<ConfigUpdateListener>();
+  @GuardedBy("this")
+  private boolean isStreamOpen;
+
+  private HttpURLConnection httpURLConnection;
```
vkryachko, line 66: `@GuardedBy("this")` as well?

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -1039,6 +1043,45 @@ public void personalization_hasMetadata_successful_without_analytics() throws Ex
...
+    verify(mockRealtimeHttpClient).addRealtimeConfigUpdateListener(eventListener);
+    assertThat(registration).isNotNull();
+  }
+
+  @Test
+  public void realtime_addListener_fail() {
+    when(mockRealtimeHttpClient.addRealtimeConfigUpdateListener(null)).thenReturn(null);
+
+    ConfigUpdateListenerRegistration registration =
+        mockRealtimeHttpClient.addRealtimeConfigUpdateListener(null);
```
vkryachko, line 1066: Should we throw instead of returning null?

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -1039,6 +1043,45 @@ public void personalization_hasMetadata_successful_without_analytics() throws Ex
...
+  public void realtime_removeListener_success() {
+    ConfigUpdateListener eventListener = generateEmptyRealtimeListener();
+    when(mockRealtimeHttpClient.addRealtimeConfigUpdateListener(eventListener))
+        .thenReturn(mockRealtimeRegistration);
+
+    ConfigUpdateListenerRegistration registration =
+        mockRealtimeHttpClient.addRealtimeConfigUpdateListener(eventListener);
+    registration.remove();
+
+    assertThat(registration).isNotNull();
```
vkryachko, line 1081: I think this is redundant since otherwise line 1079 would have thrown NullPointerException

vkryachko (commented):
It does not look like the pr is sufficiently covered by tests, i.e. I can't see tests for actual streaming

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class ConfigAutoFetch {
+
+  private static final int FETCH_RETRY = 3;
+
+  private final Set<ConfigUpdateListener> eventListeners;
```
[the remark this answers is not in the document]
qdpham13: Updated. And yeah, with the next PR that includes the retry methods we would do this for them already. This exception would be more of an indication to the user that Realtime won't be able to work at all due to the stream issue. This may enable them to revert to base RC as a fallback or some other mitigation where they are not reliant on realtime.

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
qdpham13: Actually updated the method to have NonNull annotations similar to firestore(https://github.com/firebase/firebase-android-sdk/blob/946624278814ec6d467dd145f45421348e740225/firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentReference.java#L366). 
Went ahead and removed the bad listener test too.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -14,32 +14,257 @@
...
+
+    return httpURLConnection;
+  }
+
+  private synchronized void startAutoFetch() {
+    ConfigUpdateListener retryCallback =
+        new ConfigUpdateListener() {
+          @Override
+          public void onEvent() {
+            pauseRealtime();
```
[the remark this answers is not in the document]
qdpham13: In my next PR I'll introduce the http retry method. So when autofetch and the http stream connection close we want to explicitly close the http connection(.disconnect) using pauseRealtime so that we can start a new one with the http retry method.

firebase-config/api.txt
qdpham13: And yeah, with the next PR that includes the retry methods we would do this for them already. This exception would be more of an indication to the user that Realtime won't be able to work at all due to the stream issue. This may enable them to revert to base RC as a fallback or some other mitigation where they are not reliant on realtime.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: Added

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
qdpham13: Updated

firebase-config/api.txt
qdpham13: Mainly so user's can fall back onto something else and know when to not rely on Realtime. When we retry we're only going to have limited attempts b/c we don't want to use up the devices bandwidth/battery making an unlimited amount of retries. Especially in scenarios when our endpoint is down or the device has no network connection and so it won't successfully connect for a while. Do you think we should just fail quietly and not pass anything back if we can't establish a stream?

qdpham13: Updated. PTAL @vkryachko

qdpham13:
Please review when you get the chance, @vkryachko. Thanks!

qdpham13:
Hey everyone, please take a look at the PR and give a review when you get the chance. If everyone feels like it would be easier to go over it in a meeting we can do that too. Thanks! @vkryachko

vkryachko (approved):

PR #3744, opened 2022-05-24
Upgrade Gradle to 6.9

daymxn:
We're currently utilizing a very old version of Gradle (5.4), wheres the current LTS for Gradle is 7.4.2

In an effort to get the SDK up-to-date, this PR gets Gradle up to 6.9
Moving to 7.x will require some additional work, but this is the first step.

Buganizer tickets: _233146890, 233139769, 233146881_

_233146878_ as well, but that's associated with a different PR that has [already been merged](https://github.com/google/play-services-plugins/pull/227).

### DIFF
--- a/appcheck/firebase-appcheck-debug-testing/api.txt
+++ b/appcheck/firebase-appcheck-debug-testing/api.txt
@@ -1,0 +1,15 @@
+// Signature format: 2.0
+package com.google.firebase.appcheck.debug.testing {
+
+  public final class DebugAppCheckTestHelper {
+    method @NonNull public static com.google.firebase.appcheck.debug.testing.DebugAppCheckTestHelper fromInstrumentationArgs();
+    method public <E extends java.lang.Throwable> void withDebugProvider(@NonNull com.google.firebase.appcheck.debug.testing.DebugAppCheckTestHelper.MaybeThrowingRunnable<E>) throws E;
+    method public <E extends java.lang.Throwable> void withDebugProvider(@NonNull com.google.firebase.FirebaseApp, @NonNull com.google.firebase.appcheck.debug.testing.DebugAppCheckTestHelper.MaybeThrowingRunnable<E>) throws E;
+  }
+
+  public static interface DebugAppCheckTestHelper.MaybeThrowingRunnable<E extends java.lang.Throwable> {
+    method public void run() throws E;
+  }
+
+}
+
--- a/appcheck/firebase-appcheck-debug/api.txt
+++ b/appcheck/firebase-appcheck-debug/api.txt
@@ -1,0 +1,10 @@
+// Signature format: 2.0
+package com.google.firebase.appcheck.debug {
+
+  public class DebugAppCheckProviderFactory implements com.google.firebase.appcheck.AppCheckProviderFactory {
+    method @NonNull public com.google.firebase.appcheck.AppCheckProvider create(@NonNull com.google.firebase.FirebaseApp);
+    method @NonNull public static com.google.firebase.appcheck.debug.DebugAppCheckProviderFactory getInstance();
+  }
+
+}
+
--- a/appcheck/firebase-appcheck-interop/api.txt
+++ b/appcheck/firebase-appcheck-interop/api.txt
@@ -1,0 +1,25 @@
+// Signature format: 2.0
+package com.google.firebase.appcheck {
+
+  public abstract class AppCheckTokenResult {
+    ctor public AppCheckTokenResult();
+    method @Nullable public abstract com.google.firebase.FirebaseException getError();
+    method @NonNull public abstract String getToken();
+  }
+
+}
+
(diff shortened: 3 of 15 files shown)


firebase-appdistribution-api/firebase-appdistribution-api.gradle
```diff
@@ -47,6 +47,7 @@ dependencies {
     testImplementation "com.google.truth:truth:$googleTruthVersion"
     testImplementation 'androidx.test:core:1.2.0'
     testImplementation 'org.mockito:mockito-inline:3.4.0'
+    testImplementation project(path: ':firebase-appdistribution-api')
```
rlazo: is `path:` necessary?

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/UpdateProgress.java
```diff
@@ -24,6 +24,7 @@ public interface UpdateProgress {
    * @returns the number of bytes downloaded, or -1 if called when updating to an AAB or if no new
    *     release is available.
    */
+  @NonNull
```
rlazo, line 27: These two are native types, so they cannot be null.

rlazo:
@yifanyang FYI

firebase-appdistribution-api/firebase-appdistribution-api.gradle
daymxn: Good catch! Not at all, since its the first param. Removed.

smoke-tests/build.gradle
```diff
@@ -67,7 +67,7 @@ dependencies {
   // Firebase
   implementation "com.google.firebase:firebase-analytics"
   implementation "com.google.firebase:firebase-annotations"
-  implementation "com.google.firebase:firebase-appdistribution:16.0.0-beta01"
```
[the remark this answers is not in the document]
daymxn: Ah that makes a lot of sense. Although I'm a bit confused by "The eventual bom used by the smoke test for building the test app is computed based on the public bom". This implies that the bom used by smoke-test is _not_ the actual public bom- is that correct?

smoke-tests/build.gradle
```diff
@@ -67,7 +67,8 @@ dependencies {
   // Firebase
   implementation "com.google.firebase:firebase-analytics"
   implementation "com.google.firebase:firebase-annotations"
-  implementation "com.google.firebase:firebase-appdistribution:16.0.0-beta02"
+  implementation "com.google.firebase:firebase-appdistribution:16.0.0-beta03"
+  implementation "com.google.firebase:firebase-appdistribution-api:16.0.0-beta03"
```
daymxn, line 71: Would you mind explaining the difference between these two artifacts? From reading the source, it seems as though `firbase-appdistribution-api` is utilized for outlining `firbase-appdistribution`, which in turn implements it? I'm assuming it's made that way so that `firbase-appdistribution` can be excluded from release builds for bundle sizes? But even then, it doesn't make sense to me why we have an api variant in the first place.

smoke-tests/build.gradle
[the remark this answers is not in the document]
yifanyang: Sorry, I didn't get the permission right. Can you try this one?

https://storage.cloud.google.com/android-ci/repro.tar

smoke-tests/build.gradle
yifanyang: I don't know the details. I'd imagine there may be use cases where some other product which don't require `app-distro`'s implementation, but does need to depend on some class/interface for `app-distro` to recognize and work with it.

@vkryachko Can you share your understanding here?

vkryachko: The motivation for the sdk split is exactly that, the idea is to never include `firebase-appdistribution` in prod build: [context](https://firebase.devsite.corp.google.com/docs/app-distribution/set-up-alerts?platform=android)

As for why `-api` exists see how much more complicated setup instructions were before the split: cl/451230288

smoke-tests/build.gradle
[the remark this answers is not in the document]
daymxn: This responds exactly as the PR does- any idea why my repo from the main SDK directory doesn't? It builds correctly, even without the explicit inclusion of `-api`.

smoke-tests/build.gradle
daymxn: So If I'm understanding this correctly, `firbase-appdistribution-api` acts as a wrapper of sorts around the actual `firebase-appdistribution`. This allows developers to keep "testing logic" in their code for whenever it may be needed in the future, but also allows the developer to exclude the actual testing artifact when not testing (as it not only affects bundle sizes, but is actually against tos for production since it's auto-updating). Essentially, developers can keep testing logic in their code- as the api just hollows it out, versus needing to scrap it everytime they want to release?

rlazo (approved):

Issue #3740, opened 2022-05-21
release build - R8 reports Missing class com.google.firebase.messaging.TopicOperation$TopicOperations


mikehardy:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: gradle plugin 7.0.4 (that's the thing that contains R8)
  * Firebase Component: Messaging
  * Component version: BoM 30.0.1

#### Steps to reproduce:

Do a release build that includes messaging, on android, with proguard / code shrinking enabled

#### Relevant Code:

Again, with a code shrink issue I don't *think* code is required, I *think* that you only need to include the library (like in a quickstart) and then build for release

```

> Task :app:minifyProdReleaseWithR8
WARNING:Missing classes detected while running R8. Please add the missing classes or apply additional keep rules that are generated in /Users/mike/work/Kullki/ksocialscore/packages/public-app/android/app/build/outputs/mapping/prodRelease/missing_rules.txt.

WARNING:R8: Missing class com.google.firebase.messaging.TopicOperation$TopicOperations (referenced from: void com.google.firebase.messaging.TopicOperation.<init>(java.lang.String, java.lang.String))

```

I don't see a proguard-rules.pro file in the source tree here so I can't say for certain if it is incorrect rules, or just that there should be a proguard file and there isn't one at all

sgjesse:
Looking at https://mvnrepository.com/artifact/com.google.firebase/firebase-messaging/23.0.5, the code for the private `com.google.firebase.messaging.TopicOperation(String, String)` constructor looks like this:
```
  private com.google.firebase.messaging.TopicOperation(java.lang.String, java.lang.String);
    descriptor: (Ljava/lang/String;Ljava/lang/String;)V
    flags: (0x0002) ACC_PRIVATE
    Code:
      stack=3, locals=3, args_size=3
         0: aload_0
[...]

PR #3746, opened 2022-05-25
Add @Retention to TopicOperation.TopicOperations annotation.

gsakakihara:
https://github.com/firebase/firebase-android-sdk/issues/3740
* Added @Retention(RetentionPolicy.SOURCE) to TopicOperation.TopicOperations annotation declaration to avoid having it included in the library.
* Also fixed a typo in FcmExecutors.

### DIFF
--- a/firebase-messaging/src/main/java/com/google/firebase/messaging/TopicOperation.java
+++ b/firebase-messaging/src/main/java/com/google/firebase/messaging/TopicOperation.java
@@ -22,6 +22,8 @@ import androidx.annotation.Nullable;
 import androidx.annotation.StringDef;
 import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.common.internal.Objects;
+import java.lang.annotation.Retention;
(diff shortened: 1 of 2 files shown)


willxu-google (approved):

firebase-messaging/src/main/java/com/google/firebase/messaging/TopicOperation.java
```diff
@@ -117,6 +119,7 @@ public int hashCode() {
   }
 
   @StringDef({TopicOperations.OPERATION_SUBSCRIBE, TopicOperations.OPERATION_UNSUBSCRIBE})
+  @Retention(RetentionPolicy.SOURCE)
   @interface TopicOperations {
     String OPERATION_SUBSCRIBE = "S";
     String OPERATION_UNSUBSCRIBE = "U";
```
James201311, line 125: Will these two values be discarded at compile time if we add “SOURCE” Retention policy?
If so, we might need to pull these two string out.

PR #3753, opened 2022-05-26
Ywmei/serialization dev

ywmei-brt1:


ywmei-brt1:
1. Try to remove the app/src/main/resouce folder from the code package, but found that **unit tests** are depend on `app:generateDebugBuildConfig`, this task is depend on mainfest file, mainfest file is depend on the resource files.
2. Try to remove package.json file, but found that **integration tests* are depend on this file to run local firestore emulator.

PR #3762, opened 2022-05-30
Upgrade kotlin version to 1.6.20

vkryachko:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryType.java
```diff
@@ -29,6 +29,9 @@ public String getFormat() {
   }
 
   public String getComponentName() {
+    if (this == ANDROID) {
```
rlazo, line 32: Add a comment explaining why this workaround is needed.

rlazo: thanks!

PR #3767, opened 2022-05-31
Refactor out ApkHashExtractor

lfkellogg:
This will be useful for in-app feedback because we'll be using the APK hash to identify the release.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkHashExtractor.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+/** Extracts a hash of the installed APK. */
+class ApkHashExtractor {
+
+  private static final String TAG = "ApkHashExtractor:";
+  private static final int BYTES_IN_LONG = 8;
+
+  private final ConcurrentMap<String, String> cachedApkHashes = new ConcurrentHashMap<>();
```
lfkellogg, line 31: Note that this cache used to be static (global). That didn't make sense given that the hash depends on the Context that's provided to each instance. Instead I've pulled out the creation of the ApkHashExtractor to the FirebaseAppDistributionRegistrar so that there is clearly only one of them.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkHashExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+      }
+      zipFingerprint = sb.toString();
+
+    } catch (IOException | NoSuchAlgorithmException e) {
+      LogWrapper.getInstance().v(TAG + "id calculation failed for " + file.getPath());
+      return null;
+    } finally {
+      long elapsed = System.currentTimeMillis() - start;
+      if (elapsed > 2 * 1000) {
+        LogWrapper.getInstance()
```
kaibolay, line 124: maybe log this always and merge the log statements?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ReleaseIdentificationUtils.java
```diff
@@ -21,20 +21,11 @@
...
-import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
-import java.util.Enumeration;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipFile;
 
 final class ReleaseIdentificationUtils {
```
kaibolay, line 25: maybe this deserves to be renamed since it's now only focused on AABs

kaibolay:
> This will be useful for in-app feedback because we'll be using the APK hash to identify the release.

...or the package identifier for AABs. So I'm curious: Why not keep those two related things in `ReleaseIdentificationUtils`?

lfkellogg:
> > This will be useful for in-app feedback because we'll be using the APK hash to identify the release.
> 
> ...or the package identifier for AABs. So I'm curious: Why not keep those two related things in `ReleaseIdentificationUtils`?

The main reason is the caching of the APK hash. That used to be maintained in the NewReleaseFetcher, but we'll want to take advantage of that caching for IAF as well. So now that's all isolated to the ApkHashExtractor.

We could instead have the class be something like "BinaryIdentifierExtractor" and handle the AAB vs. APK logic there, but the APK specific stuff seems like enough of a responsibility on its own. Maybe that could be a separate class that depends on an injected ApkHashExtractor.

kaibolay (approved):

PR #3771, opened 2022-06-01
Refactor FID logic into FirebaseAppDistributionTesterApiClient

lfkellogg:
Refactors the FID/FIS auth token fetching into the API client, since that will be needed by the feedback stuff.

### DIFF
--- a/firebase-appdistribution/firebase-appdistribution.gradle
+++ b/firebase-appdistribution/firebase-appdistribution.gradle
@@ -47,6 +47,7 @@ dependencies {
     implementation project(':firebase-components')
     implementation project(':firebase-installations-interop')
     implementation project(':firebase-common')
+    testImplementation project(path: ':firebase-appdistribution')
     runtimeOnly project(':firebase-installations')
 
     testImplementation 'junit:junit:4.13.2'
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
@@ -114,40 +114,6 @@ class FirebaseAppDistributionTesterApiClient {
         });
   }
 
-  /**
-   * Fetches and returns the name of the installed release given the hash of the installed APK, or
-   * null if it could not be found.
-   */
-  @NonNull
-  Task<String> findReleaseUsingApkHash(String apkHash) {
-    return Tasks.forException(
-        new FirebaseAppDistributionException("Not implemented", Status.UNKNOWN));
-  }
-
-  /**
-   * Fetches and returns the name of the installed release given the IAS artifact ID of the
-   * installed app bundle, or null if it could not be found.
-   */
-  @NonNull
-  Task<String> findReleaseUsingIasArtifactId(String iasArtifactId) {
-    return Tasks.forException(
-        new FirebaseAppDistributionException("Not implemented", Status.UNKNOWN));
-  }
-
-  /** Creates a new feedback from the given text, and returns the feedback name. */
-  @NonNull
-  Task<String> createFeedback(String testerReleaseName, String feedbackText) {
-    return Tasks.forException(
-        new FirebaseAppDistributionException("Not implemented", Status.UNKNOWN));
-  }
-
-  /** Commits the feedback with the given name. */
-  @NonNull
-  Task<Void> commitFeedback(String feedbackName) {
-    return Tasks.forException(
-        new FirebaseAppDistributionException("Not implemented", Status.UNKNOWN));
-  }
-
   private String readResponse(HttpsURLConnection connection)
       throws FirebaseAppDistributionException {
     int responseCode;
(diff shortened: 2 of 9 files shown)


firebase-appdistribution/firebase-appdistribution.gradle
```diff
@@ -47,6 +47,7 @@ dependencies {
     implementation project(':firebase-components')
     implementation project(':firebase-installations-interop')
     implementation project(':firebase-common')
+    testImplementation project(path: ':firebase-appdistribution')
```
lfkellogg, line 50: I needed to add this for my IDE to recognize the classes from my tests. Don't know why.

kaibolay:
> This also refactors the FID/FIS auth token fetching into the API client, since that will be needed by the feedback stuff.

I think it might make sense to focus this PR only on this refactoring. The four stub methods don't really do anything. They can be in a separate PR - or you'll just have a PR which directly adds the implementation (without adding the four stubs)

lfkellogg:
> > This also refactors the FID/FIS auth token fetching into the API client, since that will be needed by the feedback stuff.
> 
> I think it might make sense to focus this PR only on this refactoring. The four stub methods don't really do anything. They can be in a separate PR - or you'll just have a PR which directly adds the implementation (without adding the four stubs)

Good idea, done. Will just implement them in a followup.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -97,12 +129,23 @@ AppDistributionReleaseInternal fetchNewRelease(
...
+  private String makeGetRequest(String path, String token) throws FirebaseAppDistributionException {
+    String url = String.format("https://%s/%s", APP_TESTERS_HOST, path);
+    HttpsURLConnection connection;
+    try {
+      connection = openHttpsUrlConnection(url, token);
+    } catch (IOException e) {
+      throw new FirebaseAppDistributionException(
+          ErrorMessages.NETWORK_ERROR, Status.NETWORK_FAILURE, e);
+    }
+    return readResponse(connection);
```
kaibolay, line 148: this should probably be inside the try block. Then you don't need to catch the IOException inside readResponse() - just declare that method to throw IOException - it'll be handled here

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/AabUpdaterTest.java
```diff
@@ -68,6 +67,7 @@ public class AabUpdaterTest {
 
   private AabUpdater aabUpdater;
   private ShadowActivity shadowActivity;
+  private static ExecutorService testExecutor;
```
kaibolay, line 70: Shouldn't be static since it's initialized in a `@Before` method. But I'm curious why you changed this from the "private static final" that was used before.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -55,38 +68,57 @@ class FirebaseAppDistributionTesterApiClient {
...
+    this(
+        Executors.newSingleThreadExecutor(),
+        firebaseApp,
+        firebaseInstallationsApiProvider,
+        new HttpsUrlConnectionFactory());
   }
 
   @VisibleForTesting
   FirebaseAppDistributionTesterApiClient(
+      @NonNull Executor taskExecutor,
```
kaibolay, line 90: Is this ever anything other than `Executors.newSingleThreadExecutor()`? It seems that both, production and test code, pass in the same thing, but maybe I'm overlooking a caller.

If it's always the same, remove it as a parameter.

lfkellogg: This constructor is here, and `@VisibleForTesting`, because we want to pass in an executor from the test, so that we can await the async work that's enqueued to it.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -97,12 +129,23 @@ AppDistributionReleaseInternal fetchNewRelease(
         connection.disconnect();
```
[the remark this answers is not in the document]
lfkellogg: Done! This also made me realize I could combine the two `catch (IOException e)` in `makeGetRequest()` and `readResponseBody()`.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/AabUpdaterTest.java
lfkellogg: Ah, true, done.

I changed it to re-initialize in the `@Before` because I noticed that only having one static executor was causing state to bleed between tests.

kaibolay (approved):

PR #3776, opened 2022-06-02
Remove MaxPermSize Java Option

martinbonnin:
Removing the `MaxPermSize` option has [it started being ignored in Java8](https://stackoverflow.com/questions/18339707/permgen-elimination-in-jdk-8) and causes Java 17 (and maybe others) to fail with an unrecognized argument error


martinbonnin:
2025 cleanup

PR #3778, opened 2022-06-02
Add in-app feedback requests to FirebaseAppDistributionTesterApiClient

lfkellogg:
Also refactors out low level API-specific HTTP and basic JSON parsing functionality into TesterApiHttpClient

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
@@ -31,11 +31,9 @@ class ErrorMessages {
   static final String AUTHENTICATION_CANCELED = "Tester canceled the authentication flow.";
 
   static final String RELEASE_NOT_FOUND_ERROR =
-      "Release not found. An update was not available for the current tester and app. Make sure that FirebaseAppDistribution#checkForNewRelease returns with a non-null AppDistributionRelease before calling FirebaseAppDistribution#updateApp";
+      "Release not found. An update was not available for the current tester and app. Make sure that checkForNewRelease() returns with a non-null AppDistributionRelease before calling updateApp()";
 
-  static final String NOT_FOUND_ERROR = "Resource not found.";
-
-  static final String NO_RELEASE_ACCESS = "Tester does not have access to the release.";
+  static final String NOT_FOUND_ERROR = "Resource not found (or the tester may not have access).";
 
   static final String TIMEOUT_ERROR =
       "Request timed out. Check the tester's internet connection and try again.";
@@ -48,7 +46,7 @@ class ErrorMessages {
       "Download URL not found. This was a most likely due to a transient condition and may be corrected by retrying.";
 
   static final String HOST_ACTIVITY_INTERRUPTED =
-      "Host activity interrupted while dialog was showing. Try calling FirebaseAppDistribution#updateIfNewReleaseAvailable again.";
+      "Host activity interrupted while dialog was showing. Try calling updateIfNewReleaseAvailable() again.";
 
   static final String APK_INSTALLATION_FAILED =
       "The APK failed to install or installation was canceled by the tester.";
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
@@ -18,7 +18,9 @@ class ErrorMessages {
   static final String NETWORK_ERROR = "Request failed with unknown network error.";
 
   static final String JSON_PARSING_ERROR =
-      "Error parsing service response when checking for new release. This was most likely due to a transient condition and may be corrected by retrying.";
+      "Error parsing service response. This was most likely due to a transient condition and may be corrected by retrying.";
+
+  static final String JSON_SERIALIZATION_ERROR = "Error building request body.";
 
   static final String AUTHENTICATION_ERROR =
       "Failed to authenticate the tester. The tester was either not signed in, or something went wrong. Try signing in again.";
@@ -28,8 +30,12 @@ class ErrorMessages {
 
   static final String AUTHENTICATION_CANCELED = "Tester canceled the authentication flow.";
 
-  static final String NOT_FOUND_ERROR =
-      "Release not found. An update was not available for the current tester and app. Make sure that FirebaseAppDistribution#checkForNewRelease returns with a non-null  AppDistributionRelease before calling FirebaseAppDistribution#updateApp";
+  static final String RELEASE_NOT_FOUND_ERROR =
+      "Release not found. An update was not available for the current tester and app. Make sure that FirebaseAppDistribution#checkForNewRelease returns with a non-null AppDistributionRelease before calling FirebaseAppDistribution#updateApp";
+
+  static final String NOT_FOUND_ERROR = "Resource not found.";
+
+  static final String NO_RELEASE_ACCESS = "Tester does not have access to the release.";
 
   static final String TIMEOUT_ERROR =
       "Request timed out. Check the tester's internet connection and try again.";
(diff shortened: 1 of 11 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -109,62 +96,116 @@ Task<AppDistributionReleaseInternal> fetchNewRelease() {
               String.format(
                   "v1alpha/devices/-/testerApps/%s/installations/%s/releases",
                   firebaseApp.getOptions().getApplicationId(), fid);
-          String responseBody = makeGetRequest(path, token);
-          return parseNewRelease(responseBody);
+          HttpResponse response = testerApiHttpClient.makeGetRequest(path, token);
+          if (response.code() == 404) {
```
kaibolay, line 100: Why is this special and not handled by `getExceptionForHttpResponse()`?
Also, I think 404 should be exclusively returned for "not found" - is this a bug on the server-side? If so, we ought to fix that.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -109,62 +96,116 @@ Task<AppDistributionReleaseInternal> fetchNewRelease() {
...
+        (unused, token) -> {
+          LogWrapper.getInstance().i("Creating feedback for release: " + testerReleaseName);
+          String path = String.format("v1alpha/%s/feedback", testerReleaseName);
+          String requestBody = buildCreateFeedbackBody(feedbackText).toString();
+          HttpResponse response = testerApiHttpClient.makePostRequest(path, token, requestBody);
+          if (response.code() == 403) {
+            throw new FirebaseAppDistributionException(
+                ErrorMessages.NO_RELEASE_ACCESS, Status.AUTHENTICATION_FAILURE);
+          } else if (response.code() == 404) {
+            throw new FirebaseAppDistributionException("Release was not found", Status.UNKNOWN);
```
kaibolay, line 162: Shouldn't there be a `Status.NOT_FOUND`?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -109,62 +96,116 @@ Task<AppDistributionReleaseInternal> fetchNewRelease() {
...
+  /** Creates a new feedback from the given text, and returns the feedback name. */
+  @NonNull
+  Task<String> createFeedback(String testerReleaseName, String feedbackText) {
+    return runWithFidAndToken(
+        (unused, token) -> {
+          LogWrapper.getInstance().i("Creating feedback for release: " + testerReleaseName);
+          String path = String.format("v1alpha/%s/feedback", testerReleaseName);
+          String requestBody = buildCreateFeedbackBody(feedbackText).toString();
+          HttpResponse response = testerApiHttpClient.makePostRequest(path, token, requestBody);
+          if (response.code() == 403) {
```
kaibolay, line 158: why handle this separately? I think `getExceptionForHttpResponse()` would do the same

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
```diff
@@ -109,62 +96,116 @@ Task<AppDistributionReleaseInternal> fetchNewRelease() {
...
+            throw new FirebaseAppDistributionException(
+                ErrorMessages.NO_RELEASE_ACCESS, Status.AUTHENTICATION_FAILURE);
+          } else if (response.code() == 404) {
```
kaibolay, line 161: why handle this separately? I think `getExceptionForHttpResponse()` would do the same

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
lfkellogg: Cleaned this up according to our discussion. All error handling is done in the TesterApiHttpClient, with custom tag prepended to error messages.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
```diff
@@ -28,8 +30,12 @@ class ErrorMessages {
 
   static final String AUTHENTICATION_CANCELED = "Tester canceled the authentication flow.";
 
-  static final String NOT_FOUND_ERROR =
-      "Release not found. An update was not available for the current tester and app. Make sure that FirebaseAppDistribution#checkForNewRelease returns with a non-null  AppDistributionRelease before calling FirebaseAppDistribution#updateApp";
+  static final String RELEASE_NOT_FOUND_ERROR =
+      "Release not found. An update was not available for the current tester and app. Make sure that FirebaseAppDistribution#checkForNewRelease returns with a non-null AppDistributionRelease before calling FirebaseAppDistribution#updateApp";
```
kaibolay, line 34: nit: `FirebaseAppDistribution.checkForNewRelease()` and `FirebaseAppDistribution.updateApp()` are more legible - this is not JavaDoc

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
lfkellogg: RE: NOT_FOUND, currently we do return 404 for missing access cases, like when [the tester doesn't have access to the app](http://google3/java/com/google/firebase/appdistro/boq/testers/v1alpha/service/actions/ListTesterAppInstallationReleasesAction.java;l=278;rcl=452123374). Also if the app doesn't exist we also return 404, but we should return 403 according to https://google.aip.dev/211.

We should fix this on the backend (assuming it won't break any old SDKs, which it might), but in the meantime I think we need to keep treating 404s as `AUTHENTICATION_FAILURE`.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
lfkellogg: This is now handed in the http client.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClient.java
lfkellogg: I've actually changed this back to `AUTHENTICATION_FAILURE` as it was previously. See my other comment as to why.

I also added a code comment in the http client, explaining that we should change the status returned for 404s once we don't return them for missing access.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
lfkellogg: I actually like `checkForNewRelease()` and `updateApp()`!

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClientTest.java
```diff
@@ -54,18 +50,32 @@ public class FirebaseAppDistributionTesterApiClientTest {
   private static final String TEST_API_KEY = "AIzaSyabcdefghijklmnopqrstuvwxyz1234567";
   private static final String TEST_APP_ID_1 = "1:123456789:android:abcdef";
   private static final String TEST_PROJECT_ID = "777777777777";
```
kaibolay, line 52: can you change this to some String. with letters instead of numbers, e.g. `"project-name"` - makes it easier to distinguish from `TEST_PROJECT_NUMBER`

kaibolay (approved):

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionTesterApiClientTest.java
lfkellogg: Good idea, done!

PR #3781, opened 2022-06-03
E2e deflake

leotianlizhan:
1. Is it possible for `flgForceUploadSync()` to get the `Destination` and `Priority` somewhere so it's not hardcoded?
2. Espresso Idling Resource does add quite a bit of code, I wonder if it's not worth the extra maintenance cost and just use `Thread.sleep()`?

### DIFF
--- a/firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceFragmentScreenTracesTest.java
+++ b/firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceFragmentScreenTracesTest.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.testing.fireperf;
 
+import static androidx.test.espresso.Espresso.onIdle;
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.contrib.RecyclerViewActions.scrollToPosition;
 import static androidx.test.espresso.matcher.ViewMatchers.withId;
@@ -46,7 +47,7 @@ public class FirebasePerformanceFragmentScreenTracesTest {
       new ActivityScenarioRule<>(FirebasePerfFragmentsActivity.class);
 
   @Test
-  public void scrollAndCycleThroughAllFragments() throws InterruptedException {
+  public void scrollAndCycleThroughAllFragments() {
     ActivityScenario scenario = activityRule.getScenario();
     scrollRecyclerViewToEnd(HomeFragment.NUM_LIST_ITEMS, R.id.rv_numbers_home);
     scenario.onActivity(new NavigateAction(R.id.navigation_fast));
@@ -54,10 +55,10 @@ public class FirebasePerformanceFragmentScreenTracesTest {
     scenario.onActivity(new NavigateAction(R.id.navigation_slow));
     scrollRecyclerViewToEnd(SlowFragment.NUM_LIST_ITEMS, R.id.rv_numbers_slow);
     assertThat(scenario.getState()).isEqualTo(State.RESUMED);
-    scenario.moveToState(State.STARTED).moveToState(State.CREATED);
-
-    // End Activity screen trace by relaunching the activity to ensure the screen trace is sent.
-    scenario.launch(FirebasePerfFragmentsActivity.class);
+    scenario.moveToState(State.STARTED).moveToState(State.CREATED); // trigger activity screen trace
+    onIdle();
+    // Block until all Fireperf events are sent by Firelog
+    FireperfUtils.flgForceUploadSync();
   }
 
   private void scrollRecyclerViewToEnd(int itemCount, int viewId) {
--- a/firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceScreenTracesTest.java
+++ b/firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceScreenTracesTest.java
@@ -14,10 +14,13 @@
 
 package com.google.firebase.testing.fireperf;
 
+import static androidx.test.espresso.Espresso.onIdle;
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.contrib.RecyclerViewActions.scrollToPosition;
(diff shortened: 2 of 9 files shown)


jeremyjiang-dev (commented):
Thanks for your hard work on the investigation and refactoring!

My concern is that Espresso Idling Resource added some code to the SDK and its only purpose is for testing. I think the ideal way is to integrate idling resource into executors like you found in TikTok framework. Therefore, I am leaning towards finding some lightweight ways to fix the E2E flakiness.

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceFragmentScreenTracesTest.java
```diff
@@ -54,10 +54,11 @@ public void scrollAndCycleThroughAllFragments() throws InterruptedException {
...
     assertThat(scenario.getState()).isEqualTo(State.RESUMED);
-    scenario.moveToState(State.STARTED).moveToState(State.CREATED);
-
-    // End Activity screen trace by relaunching the activity to ensure the screen trace is sent.
-    scenario.launch(FirebasePerfFragmentsActivity.class);
+    scenario.moveToState(State.STARTED).moveToState(State.CREATED); // trigger activity screen trace
+    // Wait for TransportManager and Firelog executors to finish
+    Thread.sleep(5000);
+    // Block until all Fireperf events are sent by Firelog
+    InstrumentationTestUtil.flgForceUploadSync();
```
visumickey, line 61: Would the screen trace be completed by this time? Should we move to a different activity to make sure the current activity trace is completed?

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceScreenTracesTest.java
```diff
@@ -49,7 +51,11 @@ public void scrollRecyclerViewToEnd() {
...
       currItemCount += 5;
     }
-    // End Activity screen trace by switching to another Activity
-    scenario.launch(FirebasePerfScreenTracesActivity.class);
+    assertThat(scenario.getState()).isEqualTo(State.RESUMED);
+    scenario.moveToState(State.STARTED).moveToState(State.CREATED); // trigger activity screen trace
+    // Wait for TransportManager and Firelog executors to finish
+    Thread.sleep(5000);
+    // Block until all Fireperf events are sent by Firelog
+    InstrumentationTestUtil.flgForceUploadSync();
```
visumickey, line 59: Same as above, should we cause another activity state transition to make sure we are dispatching events?

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/InstrumentationTestUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.android.datatransport.cct.CCTDestination;
+import com.google.android.datatransport.runtime.TransportRuntimeTesting;
+
+public class InstrumentationTestUtil {
+  /**
+   * Blocks calling thread until all of Fireperf's persisted events are sent by Firelog. Must be
+   * called after events have persisted.
+   */
+  static void flgForceUploadSync() {
+    TransportRuntimeTesting.forceUpload(CCTDestination.LEGACY_INSTANCE, Priority.DEFAULT);
```
visumickey, line 27: I think the Priority.DEFAULT is once in every 30 seconds. Can we use Priority.HIGHEST to make sure the events get dispatched immediately?

https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/util/PriorityMapping.java#L29

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceFragmentScreenTracesTest.java
leotianlizhan: The line before this is
```scenario.moveToState(State.STARTED).moveToState(State.CREATED); // trigger activity screen trace```
which already triggers screen trace to be recorded. Reading the source code, behind the scenes that method also moves to different activity.

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/InstrumentationTestUtil.java
leotianlizhan: From the last Acore sync, my understanding from Vlad's recommendation to use this was that this Priority is the type of event that we want to force-send. For Fireperf, all events are sent using [`Event.ofData()`](https://github.com/firebase/firebase-android-sdk/blob/58941173b766325b670c76e9ea851feb696e2a23/firebase-perf/src/main/java/com/google/firebase/perf/transport/FlgTransport.java#L56), and that is always [`Priority.DEFAULT`](https://github.com/firebase/firebase-android-sdk/blob/58941173b766325b670c76e9ea851feb696e2a23/transport/transport-api/src/main/java/com/google/android/datatransport/Event.java#L33). @vkryachko for confirmation

jeremyjiang-dev (approved):

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/FirebasePerformanceFragmentScreenTracesTest.java
visumickey: Sounds good! The activity transition was not evidently understandable in the code here. May be a comment here would make it easy for understanding this for future contributions.

firebase-perf/e2e-app/src/androidTest/java/com/google/firebase/testing/fireperf/InstrumentationTestUtil.java
visumickey: A nudge to this comment! Apart from this, rest of the changes look good to me.

visumickey (approved):
Changes look good to me on CI green!

PR #3797, opened 2022-06-08
Refactor IndexBackfiller constructor

tom-andersen:


ehsannas (approved):
This is great. Good job identifying this. One minor nit pointed out below.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/Assert.java
```diff
@@ -32,6 +32,23 @@ public static void hardAssert(boolean condition, String messageFormat, Object...
     }
   }
 
+  /**
+   * Triggers a hard assertion. The null check is guaranteed to be checked at runtime. If the object
+   * is null an AssertionError will be thrown. The string messageFormat will be formatted with the
+   * provided args using {@link String#format(String, Object...)}.
+   *
+   * @param obj The condition to check
```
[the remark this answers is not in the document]
tom-andersen: Fixed

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexBackfillerTest.java
```diff
@@ -83,9 +85,10 @@ public void setUp() {
             persistence.getMutationQueue(User.UNAUTHENTICATED, indexManager),
             documentOverlayCache,
             indexManager);
-    backfiller = new IndexBackfiller(persistence, new AsyncQueue());
-    backfiller.setIndexManager(indexManager);
-    backfiller.setLocalDocumentsView(localDocumentsView);
+    LocalStore localStore = mock(LocalStore.class);
```
wu-hui, line 88: I am still a bit hang up on the fact that local store is no longer final because we need to mock it.

It seems `QueryEngine` depends on `IndexManager` and `LocalDocumentsView`, which we already have. Can we just create the query engine here, and use it to construct a proper local store instance?

In general, I am against introducing mocks unless there is no other way around, here it might be more hassle, but I think the end result is closer to the code that gets run in prod.

wu-hui (commented):
Great refactor, I still want to avoid mocking local store however..

firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexBackfillerTest.java
tom-andersen: To address your concern, I created an interface `UserComponents`. `LocalStore` will implement interface, and the test can just use an anonymous implementation. This will allow LocalStore to remain `final`. I also think this might be a step in the right direction to separate out logic associated with user change and components. Maybe this responsibility should be moved outside LocalStore (already quite large 900 lines class)?

tom-andersen: Followup on offline conversation:

Instantiating LocalStore drags in too many other dependencies. From a _unit_ test perspective, mocks were invented to solve this problem, and avoid having to instantiate the world in order to run a unit test. LocalStore is a textbook definition of something that should be mocked in _unit_ tests.

I agree that we should strive to keep classes final. This has performance benefits, and also helps us more easily reason about our code.

Having a single implementation of an interface is not necessarily a bad thing. Consider the Dependency Inversion Principle from SOLID. The interface describes a contract between two classes. Although LocalStore offers many other methods, within the context of providing components with correct user attached, the interface offers a reduced subset of public methods on LocalStore. The interface doesn't simply repeat the public methods on the class, but instead describes what is required to fulfill the contract with the IndexBackfiller class. This makes the system more loosely coupled. This is exactly what was done with the BundleCallback interface as well.

That said, I have come up with another iteration. Instead of an interface, I have simply required Supplier<> parameters for `IndexManager` and `LocalDocumentsView`.  Unfortunately, since create method that instantiates IndexBackfiller exists in the `com.google.firebase.firestore.core` package, I had to make `com.google.firebase.firestore.local.LocalDocumentsView` public instead of package private.

I am not sure whether this is an improvement over having an interface.

wu-hui (approved):
LGTM!

PR #3806, opened 2022-06-13
[Remote Config] add useEmulator to Remote Config SDK

kroikie:
Add `useEmulator` method to allow Remote Config Emulator usage.

The `useEmulator` method changes the `fetchUrl` to point to the emulator. This is inline with `useEmulator` implementations in other SDKs.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
@@ -582,10 +582,9 @@ public class FirebaseRemoteConfig {
    * @param port the emulator port (for example, 9299)
    * @exception FirebaseRemoteConfigClientException Thrown when useEmulator is called after a fetch.
    */
-  public void useEmulator(@NonNull String host, int port)
-      throws FirebaseRemoteConfigClientException {
+  public void useEmulator(@NonNull String host, int port) {
     if (this.frcMetadata.getInfo().getLastFetchStatus() != LAST_FETCH_STATUS_NO_FETCH_YET) {
-      throw new FirebaseRemoteConfigClientException(
+      throw new IllegalStateException(
           "Cannot connect to emulator after a fetch has been made.");
     }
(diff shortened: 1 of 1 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -242,7 +249,16 @@ private void setUpUrlConnection(
   }
 
   private String getFetchUrl(String projectNumber, String namespace) {
-    return String.format(FETCH_REGEX_URL, projectNumber, namespace);
+    String fetchRegexUrl = FETCH_REGEX_URL;
+    if (emulatedServiceSettings != null) {
+      fetchRegexUrl =
+          "http://"
+              + emulatedServiceSettings.getHost()
```
yuchenshi, line 256: Will this handle IPv6 addresses correctly, like bracket-quoting (`[::1]`)?

kroikie: Yes: [android docs](https://developer.android.com/reference/java/net/HttpURLConnection#ipv6-support).

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -242,7 +249,16 @@ private void setUpUrlConnection(
   }
 
   private String getFetchUrl(String projectNumber, String namespace) {
-    return String.format(FETCH_REGEX_URL, projectNumber, namespace);
+    String fetchRegexUrl = FETCH_REGEX_URL;
```
danasilver, line 252: Could be more direct to return early if emulatedServiceSettings is null and if not, construct and return the emulator url.

kroikie: Done!

PR #3812, opened 2022-06-14
Update the orderBy filtering mechanism for OR queries.

ehsannas:


firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -392,7 +392,12 @@ private boolean matchesFilters(Document doc) {
 
   /** A document must have a value for every ordering clause in order to show up in the results. */
   private boolean matchesOrderBy(Document doc) {
-    for (OrderBy order : explicitSortOrder) {
+    // We must use `getOrderBy` here, not `explicitSortOrder`. For queries without disjunctions,
```
wu-hui, line 395: I am confused by how this change makes the fix. Can you give a better explanation or some example in this comment?

ehsannas: I have attempted to improve this comment with an example. Please let me know if it's still unclear. Thanks

wu-hui (approved):
Yes, much better with an example!

PR #3813, opened 2022-06-14
duplicate initialization during direct boot

yifanyang:
google-services.json: http://gpaste/6726677146107904

Repro steps:
- `./gradlew publishAllToBuildDir` in sdk root dir
- `./gradlew assemble` in `direct-boot-test` dir
- install the direct boot test app on real device
- restart the device without unlocking the device
- `adb -d logcat FirebasePerformance:D FirebaseApp:D yifany:D *:S` to listen to device log
- `adb -d shell am broadcast -a yifany -n "com.example.testapp/com.example.testapp.MyReceiver"` to send a broadcast
- unlock the device
- search for `ConfigResolver initialized` in device log

In my experiment (http://gpaste/6375922266537984), there were multiple entries of `ConfigResolver initialized` from different processes? threads?

**06/17 update**:
Reproduced again on a local Pixel 6. Output from `logcat` and `ps` can be found here (http://gpaste/6610304772341760).
It looks like the  multiple logcat entries of `ConfigResolver initialized` are indeed from different processes, but all these processes are for the same app (`direct.boot`).


yifanyang:
@vkryachko @visumickey 

Hello folks, I have observed some odd behavior of our SDKs during direct boot mode. Can you folks help me take a look at this pull request?

direct-boot-test/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,41 @@
...
+            </intent-filter>
+        </activity>
+
+        <receiver
+            android:name=".MyReceiver"
+            android:exported="true"
+            android:directBootAware="true">
+            <intent-filter>
+                <action android:name="android.intent.action.BOOT_COMPLETED"/>
+                <!--<action android:name="android.intent.action.LOCKED_BOOT_COMPLETED" />-->
```
vkryachko, line 35: What is the point of having this commented out? And why do you need to send the broadcast manually?

yifanyang: This broadcast `LOCKED_BOOT_COMPLETED` is sent right after the device finished reboot. I don't want the app to listen to it, since I would like to examine the device (logcat, ps) even before the app start. 

Also now that I control the timing via manual broadcast, it's easier for me to reason about which log entry comes from what event. Does it make sense?

PR #3816, opened 2022-06-15
Implement collectAndSendFeedback()

lfkellogg:
Adds `FeedbackSender` as a new Component, so that it can be accessed by the FeedbackActivity.

Screencast (submit fails because the upload URL mapping has not rolled out to GFE yet): https://screencast.googleplex.com/cast/NTQ2NDIyNjkzNzE3NjA2NHw0M2YwNjUxYS03Nw

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
@@ -15,18 +15,22 @@
 package com.google.firebase.appdistribution.impl;
 
 import android.graphics.Bitmap;
-import android.graphics.Bitmap.Config;
 import com.google.android.gms.tasks.Task;
-import com.google.android.gms.tasks.Tasks;
 
-/** A class that takes screenshots of the host app. */
-class ScreenshotTaker {
+/** Sends tester feedback to the Tester API. */
+class FeedbackSender {
 
-  private static final Bitmap TEMP_FIXED_BITMAP = Bitmap.createBitmap(400, 400, Config.RGB_565);
+  private final FirebaseAppDistributionTesterApiClient testerApiClient;
 
-  /** Take a screenshot of the running host app. */
-  Task<Bitmap> takeScreenshot() {
-    // TODO(lkellogg): Actually take a screenshot
-    return Tasks.forResult(TEMP_FIXED_BITMAP);
+  FeedbackSender(FirebaseAppDistributionTesterApiClient testerApiClient) {
+    this.testerApiClient = testerApiClient;
+  }
+
+  /** Send feedback text and screenshot to the Tester API for the given release. */
+  Task<Void> sendFeedback(String releaseName, String feedbackText, Bitmap screenshot) {
+    return testerApiClient
+        .createFeedback(releaseName, feedbackText)
+        .onSuccessTask(feedbackName -> testerApiClient.attachScreenshot(feedbackName, screenshot))
+        .onSuccessTask(testerApiClient::commitFeedback);
   }
 }
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
@@ -18,8 +18,6 @@ import android.graphics.Bitmap;
 import androidx.annotation.NonNull;
 import com.google.android.gms.tasks.Task;
 import com.google.firebase.FirebaseApp;
-import com.google.firebase.appdistribution.FirebaseAppDistribution;
-import com.google.firebase.appdistribution.internal.FirebaseAppDistributionProxy;
 
 /** Sends tester feedback to the Tester API. */
 class FeedbackSender {
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
@@ -15,7 +15,11 @@
 package com.google.firebase.appdistribution.impl;
 
 import android.graphics.Bitmap;
+import androidx.annotation.NonNull;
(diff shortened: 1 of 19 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionRegistrar.java
```diff
@@ -46,23 +46,39 @@ public class FirebaseAppDistributionRegistrar implements ComponentRegistrar {
         Component.builder(FirebaseAppDistribution.class)
             .add(Dependency.required(FirebaseApp.class))
             .add(Dependency.requiredProvider(FirebaseInstallationsApi.class))
+            .add(Dependency.required(FeedbackSender.class))
             .factory(this::buildFirebaseAppDistribution)
             // construct FirebaseAppDistribution instance on startup so we can register for
             // activity lifecycle callbacks before the API is called
             .alwaysEager()
             .build(),
+        Component.builder(FeedbackSender.class)
```
rachaprince, line 55: I am not sure I understand why this was added to the Registrar. Is there a reason we want to implement this at a later time instead of on startup?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ScreenshotTaker.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.android.gms.tasks.Tasks;
+
+/** A class that takes screenshots of the host app. */
+class ScreenshotTaker {
+
+  private static final Bitmap TEMP_FIXED_BITMAP = Bitmap.createBitmap(400, 400, Config.RGB_565);
+
+  /** Take a screenshot of the running host app. */
+  Task<Bitmap> takeScreenshot() {
+    // TODO(lkellogg): Actually take a screenshot
```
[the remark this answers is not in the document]
lfkellogg: Nice! We can get an activity using FirebaseAppDistributionActivityLifecycleNotifier.applyToForegroundActivity().

TODO: shorten the name of that class :)

I think we can add that in a followup.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -297,6 +306,43 @@ private UpdateTask updateApp(boolean showDownloadInNotificationManager) {
...
+            taskExecutor,
+            screenshot ->
+                testerSignInManager
+                    .signInTester()
+                    .addOnFailureListener(
+                        taskExecutor,
+                        e ->
+                            LogWrapper.getInstance()
+                                .e("Failed to sign in tester. Could not collect feedback.", e))
+                    .onSuccessTask(taskExecutor, unused -> releaseIdentifier.identifyRelease())
```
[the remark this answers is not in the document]
lfkellogg: I don't think so. I assume we want to take the screenshot ASAP, as opposed to potentially having to wait until they visit the browser to sign in. And then we have to sign them in before trying to identify the release, since that requires an API call. We _could_ look up the binary identifier in parallel to the screenshot or sign in, but I'm not sure it's worth it to complicate the ReleaseIdentifier API by splitting the binary identifier computation and API call.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
```diff
@@ -0,0 +1,45 @@
...
+/** Sends tester feedback to the Tester API. */
+class FeedbackSender {
+
+  private final FirebaseAppDistributionTesterApiClient testerApiClient;
+
+  FeedbackSender(FirebaseAppDistributionTesterApiClient testerApiClient) {
+    this.testerApiClient = testerApiClient;
+  }
+
+  @NonNull
```
jladieu, line 31: I noticed you removed this annotation in UpdateProgress.  How do you decide it's needed here?

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jladieu, line 1: Is this file generated somehow?

jladieu (approved):

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
lfkellogg: No I built this. It's a very basic prototype UI. You can see it in the screencast in the PR description.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
lfkellogg: It wasn't needed in UpdateProgress because those return types (long) were not nullable.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionRegistrar.java
lfkellogg: Great, thanks for taking a look!

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
lfkellogg: Actually I don't think this is necessary either. In this case it's because this is not a publicly exposed API. I'll remove it.

kaibolay (approved):

kaibolay:
> Task :firebase-appdistribution-api:verifyGoogleJavaFormat FAILED

The following files are not formatted properly:
/home/prow/go/src/github.com/firebase/firebase-android-sdk/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java

PR #3824, opened 2022-06-20
Define operator enum for composite filters.

ehsannas:


ehsannas:
@cherylEnkidu FYI

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java
```diff
@@ -695,6 +695,32 @@ StructuredQuery.Filter encodeUnaryOrFieldFilter(FieldFilter filter) {
     return StructuredQuery.Filter.newBuilder().setFieldFilter(proto).build();
   }
 
+  CompositeFilter.Operator encodeCompositeFilterOperator(
```
MarkDuckworth, line 698: Not sure if we have guidelines with respect to this, but I find the readability of `CompositeFilter.Operator` to be challenging compared to `StructuredQuery.CompositeFilter.Operator` or a fully qualified type name. Only in the context of having `CompositeFilter.Operator` in another namespace.

MarkDuckworth (approved):
One comment, but otherwise LGTM.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java
ehsannas: I don't think there's a particular requirement from the style guide, but I agree that it's more readable that way. Done.

PR #3835, opened 2022-06-22
Take actual screenshots when giving feedback

lfkellogg:
In prototype UI, shows thumbnail of screenshot:

screenshot_screenshot

### DIFF
--- a/firebase-appdistribution/firebase-appdistribution.gradle
+++ b/firebase-appdistribution/firebase-appdistribution.gradle
@@ -51,7 +51,7 @@ dependencies {
     runtimeOnly project(':firebase-installations')
 
     testImplementation 'junit:junit:4.13.2'
-    testImplementation "org.robolectric:robolectric:$robolectricVersion"
+    testImplementation "org.robolectric:robolectric:4.8.1"
     testImplementation "com.google.truth:truth:$googleTruthVersion"
     testImplementation 'org.mockito:mockito-inline:3.4.0'
     androidTestImplementation "org.mockito:mockito-android:3.4.0"
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
@@ -15,7 +15,6 @@
 package com.google.firebase.appdistribution.impl;
 
 import android.graphics.Bitmap;
-import android.graphics.BitmapFactory;
 import android.os.Bundle;
 import android.view.View;
 import android.widget.EditText;
@@ -23,11 +22,14 @@ import android.widget.ImageView;
 import android.widget.Toast;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
+import java.io.File;
 
 /** Activity for tester to compose and submit feedback. */
 public class FeedbackActivity extends AppCompatActivity {
 
   private static final String TAG = "FeedbackActivity";
+  private static final int THUMBNAIL_WIDTH = 200;
+  private static final int THUMBNAIL_HEIGHT = 200;
 
   public static final String RELEASE_NAME_EXTRA_KEY =
       "com.google.firebase.appdistribution.FeedbackActivity.RELEASE_NAME";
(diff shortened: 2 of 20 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
```diff
@@ -32,11 +34,18 @@ static FeedbackSender getInstance() {
...
-        .onSuccessTask(feedbackName -> testerApiClient.attachScreenshot(feedbackName, screenshot))
+        .onSuccessTask(feedbackName -> attachScreenshot(feedbackName, screenshot))
         .onSuccessTask(testerApiClient::commitFeedback);
   }
+
+  private Task<String> attachScreenshot(String feedbackName, @Nullable Bitmap screenshot) {
+    if (screenshot == null) {
+      return Tasks.forResult(feedbackName);
+    }
+    return testerApiClient.attachScreenshot(feedbackName, screenshot);
```
kaibolay, line 49: Maybe this should now get the filename of the screenshot instead? If I understand correctly we already have a compressed file - so we should be able to simply write that to the network instead of compressing the bitmap (again)

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -28,20 +31,42 @@ public class FeedbackActivity extends AppCompatActivity {
...
   private String releaseName;
-  private Bitmap screenshot;
+  @Nullable private Bitmap screenshot;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     releaseName = getIntent().getStringExtra(RELEASE_NAME_EXTRA_KEY);
-    screenshot = getIntent().getParcelableExtra(SCREENSHOT_EXTRA_KEY);
+    screenshot = readScreenshot(getIntent().getStringExtra(SCREENSHOT_FILENAME_EXTRA_KEY));
```
kaibolay, line 45: How about creating a thumbnail in `ScreenshotTaker` (in addition to writing the full resolution file)? It should be possible to pass the Thumbnail directly via the intent so it can be displayed in this activity. The full resolution file can then be read in `FirebaseAppDistributionTesterApiClient` without getting read and decompressed here - and then recompressed in the client...

kaibolay (changes requested):

firebase-appdistribution/firebase-appdistribution.gradle
```diff
@@ -51,7 +51,7 @@ dependencies {
     runtimeOnly project(':firebase-installations')
 
     testImplementation 'junit:junit:4.13.2'
-    testImplementation "org.robolectric:robolectric:$robolectricVersion"
+    testImplementation "org.robolectric:robolectric:4.8.1"
```
lfkellogg, line 54: I had to update Robolectric to pick up `setAllowInvalidImageData()`, and following the recommendation in Android SDK chat to just update it for our own local tests.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
lfkellogg: Any displayable bitmap is really too big to be passing via Intent, since they should be kept to "a few kb".

However, I've updated FeedbackAction to read a scaled thumbnail from the png file, which will reduce the memory footprint when starting the action.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ImageUtils.java
```diff
@@ -0,0 +1,67 @@
...
+              "Tried to read image with bad dimensions: %dx%d", targetWidth, targetHeight));
+    }
+    ImageSize imageSize = getImageSize(file);
+    final BitmapFactory.Options options = new BitmapFactory.Options();
+    options.inSampleSize =
+        calculateInSampleSize(imageSize.width(), imageSize.height(), targetWidth, targetHeight);
+    return BitmapFactory.decodeFile(file.getAbsolutePath(), options);
+  }
+
+  private static ImageSize getImageSize(File file) {
```
kaibolay, line 43: nit: This could be moved to ImageSize to act as a factory method. Proposed method name: `read()`

PR #3837, opened 2022-06-23
Rename keep.xml to avoid merge conflict with dynamic features

### LINKED ISSUE #3836: Please add prefix on res/raw/keep.xml file

    with the firebase tag.
    google group.

  * Android Studio version: Any
  * Firebase Component: firebase-common
  * Component version: 20.1.1

After update firebase-common 20.1.1, app build was failed with following log.

```
> Task :app:packageDebugBundle FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:packageDebugBundle'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.PackageBundleTask$BundleToolWorkAction
   > Modules 'base' and 'feature1' contain entry 'res/raw/keep.xml' with different content.

```

My app has [dynamic feature modules](https://developer.android.com/guide/playcore/feature-delivery), and a 3rd party library which is included on a dynamic feature has also `res/raw/keep.xml` 

But firebase-common has [same resource name with different content](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-common/src/main/res/raw/keep.xml), and AGP cannot merge them.

It seems that the resource name does not have to be `keep.xml`,
https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/res/shrinker/usages/ToolsAttributeUsageRecorder.kt#51

But it is introduced as keep.xml in [the guideline documentation](https://developer.android.com/studio/build/shrink-code#keep-resources).
So, many library are providing their rules as `keep.xml` including `firebase-common`

To avoid merge conflict `keep.xml`, please rename it with prefix.

ganadist:
Fix #3836


vkryachko:
@ganadist thanks for the PR

@ricowind do you mind taking a look? does the change look reasonable?

Issue #3826, opened 2022-06-21
auto collection of android_id is not compliance with privacy policy


xyz-fly:
针对android 21 以上
采用的版本是：Bom 30.1.0    firebase-crashlytics  18.2.11

我们在使用该Crashlytics SDK时，我们需要先给用户弹出对话框，展示是否同意隐私权益，如果用户同意后，才可以收集像android ID这样的参数。但弹出隐私协议对话框，还需要用户操作，这个过程需要花费一定时间。但是，app却已经开始调用收集android ID的函数方法。我们希望可以改善这块函数方法，不要在用户未同意隐私权益前，调用这段代码。

我们发现，触发这个收集android id的方法来自于CommonUtils.java的如下函数方法调用：

  public static boolean isEmulator(Context context) {
    final String androidId = Secure.getString(context.getContentResolver(), Secure.ANDROID_ID);
    return Build.PRODUCT.contains(SDK)
        || Build.HARDWARE.contains(GOLDFISH)
        || Build.HARDWARE.contains(RANCHU)
        || androidId == null;
  }

argzdev:
Translation:
> "When we use the Crashlytics SDK, we need to pop up a dialog box for the user to show whether he agrees to the privacy rights. If the user agrees, we can collect parameters such as android ID. However, the pop-up privacy agreement dialog box still requires user operation, and this process takes a certain amount of time. However, the app has already started to call the function method to collect the android ID. We hope to improve this function and do not call this code before the user agrees to the privacy rights."
> "We found that the method that triggers this collection of android id comes from the following function method call in CommonUtils.java:"

Thanks for reporting, @xyz-fly. What you can do is refer to our documentation [here](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=android#enable-reporting). You can initially disable `firebase_crashlytics_collection_enabled` In the application block of your `AndroidManifest.xml` file

```
<meta-data
    android:name="firebase_crashlytics_collection_enabled"
    android:value="false" />
```

Then enable collection through your code when the developer agrees to your privacy rights using: `Firebase.crashlytics.setCrashlyticsCollectionEnabled(true)`


That being said, since this is resolved, I'll be closing this now. Thanks!

xyz-fly:
> Translation:
> 
[...]

PR #3840, opened 2022-06-23
Remove Android ID check heuristic from isEmulator check.

mrober:
Remove Android ID check heuristic from isEmulator check.

Tested with unit tests, the default emulators with Android Studio, and a physical device. There might be more heuristics we can replace this with if needed.


mrober:
#3826

kmandrika (approved):

PR #3844, opened 2022-06-24
Revert "Add support for return values to applyToForegroundActivity()"

lfkellogg:
Reverts firebase/firebase-android-sdk#3831

This will be reapplied to the new feedback feature branch.


lfkellogg:
Let's not do this yet

PR #3852, opened 2022-06-29
Archive of implementation of all the features, not for review

ywmei-brt1:
A demo showing how the decoder is going to work together with the component registrar


firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
```diff
@@ -178,11 +180,31 @@ public <T> T toObject(
     checkNotNull(
         serverTimestampBehavior, "Provided serverTimestampBehavior value must not be null.");
     Map<String, Object> data = getData(serverTimestampBehavior);
-    return data == null
-        ? null
+    // Ktx decoder goes here!
+    MapEncoder ktxMapEncoder = getKtxMapEncoderById();
+    if (data == null) return null;
+    return isKtxSerializable(valueType)
+        ? ktxMapEncoder.decode(data, valueType, getReference())
```
ywmei-brt1, line 187: This is my plan for handling Kotlin decoder.

firebase-firestore/ktx/src/androidTest/java/com/google/firebase/firestore/EncoderDecoderRoundTripTests.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    // This test is intended to fail for demo purpose
+    //
+    // expected: PlainProject(Name=foo, OwnerName=bar, BoolField=true, DocID=null) -> this is the
+    // Kotlin Custom Object before the round trip
+    //
+    // but was : PlainProject(Name=foo, OwnerName=bar, BoolField=true, DocID=123-456-789) -> this is
+    // the object after the round trip, DocID is filled by @KDcoumentId annotation
+    //
+    // This difference proofs that the DocumentSnapshot.toObject() is using the Kotlin decoder under the hood.
```
ywmei-brt1, line 35: There is a sample round trip demo showing the Kotlin encoder/decoder component is working, and @KDocumentId field is ignored during encoding, and is filled by decoder

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
vkryachko: What is the purpose of finding a specific encoder by id? can't you use `supports()` instead? i.e.

```java
for(MapEncoder encoder : availableEncoders) {
  if(encoder.supports(value.class)) {
    return encoder.decode(value);
  }
}
// fallback to default
return CustomClassMapper(convertToCustomClass(data, valueType, getReference());
```

The idea is that the first encoder that claims to support the type is used.

firebase-firestore/src/main/java/com/google/firebase/firestore/encoding/MapEncoder.java
```diff
@@ -27,10 +28,15 @@ public interface MapEncoder {
...
   Map<String, Object> encode(Object value);
 
+  // Decodes a map to custom object
+  <T> T decode(Map<String, Object> data, Class<T> valueType, DocumentReference docRef);
+
   // Uses reflection to determine if this object is suitable to be encoded with the Kotlin
   // serialization plugin
   boolean isAbleToBeEncoded(Object value);
 
+  <T> boolean isAbleToBeDecoded(Class<T> valueType);
```
vkryachko, line 38: Do we need 2 methods? is it possible that some future encoder would want to support only encoding or only decoding? also do we want it to be possible?

How about having just one method that would indicate support for both directions? e.g.

```java
boolean supports(Class<?> cls);
```

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentSnapshot.java
ywmei-brt1: Thank you, I have discussed with Rodrigo, and we are using `supports()` now.

firebase-firestore/src/main/java/com/google/firebase/firestore/encoding/MapEncoder.java
ywmei-brt1: I brought up my concern to Rodrigo that there are some cases, where one group of people saving to firestore, and another group of people get data from firestore. It will be cool/useful if we give customer the option to choose which encoder/decoder they want to use at runtime.  [I do heard that the company Strip is saving their enum field to firestore in BigCamel format, and try to get them back as smallCamel ]

Rodrigo is agree with you that we should not give customer too much freedom, and it makes customer life easier as they have our Kotlin extension lib, they can use the Kotlin serialization style in both encoding and decoding. So eventually, we are following your `supports()` suggestion. Thank you.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/MapEncoderKtxImp.kt
```diff
@@ -29,35 +30,29 @@ class MapEncoderKtxImp : MapEncoder {
...
+        valueType: Class<T>,
+        docRef: DocumentReference
+    ): T {
+        val deserializer = serializer(valueType)
+        val doc: FirestoreDocument = FirestoreDocument(docRef.id, docRef)
+        return decodeFromNestedMap(data, deserializer, doc) as T
+    }
+
+    // a combination of isAbleToBeEncode and isAbleToBeDecoded
+    override fun <T : Any?> supports(valueType: Class<T>): Boolean {
```
ywmei-brt1, line 46: I am using `supports()` now. The interface is much clean now :)

PR #3854, opened 2022-06-29
Increase the Data Transfer read timeout default to 130 seconds

mrober:
There is a deadline mismatch between the client and the server. The client has a deadline of 40 seconds, and server has 2 minutes, causing the client to repeat requests (with backoff), until they are delivered. This will increase the client deadline to 130 seconds.


kmandrika (approved):
LGTM, but please add why this increase is happening to the description of the PR

PR #3855, opened 2022-06-29
update kotlin from 1.3.72 to 1.4.32

thatfiredev:
Since we can't yet upgrade to Kotlin 1.6 or later versions, I'm making small incremental changes to get us there.


thatfiredev:
@google-oss-bot /retest

rlazo (approved):
Kotlindoc still works after this change.

Issue #3864, opened 2022-06-30
java.lang.NumberFormatException: For input string: ""


kabeer-uber:
### [READ] Step 1: Are you in the right place?

Yes

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: **Any**
  * Firebase Component: firebase-ml-modeldownloader
  * Component version: 24.0.2

#### Steps to reproduce:
Note this is an edge case the requirement to hit this edge case are as below:

Be in the process of migrating from [FirebaseModelManager](https://firebase.google.com/docs/reference/android/com/google/firebase/ml/common/modeldownload/FirebaseModelManager) (old version) to [FirebaseModelDownloader](https://firebase.google.com/docs/reference/android/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader) (open source).

Why because they store the same model in the exact same destination. Path (/data/user/0/\<app package name\>/no_backup/com.google.firebase.ml.custom.models/\<firebase persistent key\>/\<model name\>) verified by checking the file paths via both the downloaders. But they store model details in different locations, i.e details in the SharedPreferences cache are in different namespaces. (Open source firebase model downloader stores it in [com.google.firebase.ml.modelDownloader](https://github.com/firebase/firebase-android-sdk/blob/ac3276ce7d5f5181b1c419731495eb88a98537af/firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/SharedPreferencesUtil.java#L40)  and likely old FirebaseModelManager stores it here com.google.firebase.ml.common.modeldownload) 

Two models need to be triggered for download at the same time (It leads to partial information present in the SharedPreferences cache) and with the larger model already downloaded on the device. Smaller model download = faster download time relative to larger model download.

[...]

PR #3865, opened 2022-07-01
Adding file path check to avoid NumberFormatException due to empt path

kabeer-uber:
As the issue https://github.com/firebase/firebase-android-sdk/issues/3864 details, this is an edge case I would like to fix and shouldn't occur in regular occurrence.

An alternative to this fix is to potentially change

CUSTOM_MODEL_ROOT_PATH = "com.google.firebase.ml.custom.models";

to something different than what [FirebaseModelManager](https://firebase.google.com/docs/reference/android/com/google/firebase/ml/common/modeldownload/FirebaseModelManager) was using.


rlazo (approved):
It looks better, thanks!

rlazo:
Thanks again for your contribution @kabeer-uber!

PR #3871, opened 2022-07-06
Support leaf modules in changed modules task.

vkryachko:
The task should return the module even if nothing depends on it since it
likely has tests of its own.


buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/ChangedModulesTask.kt
```diff
@@ -33,7 +33,9 @@ abstract class ChangedModulesTask : DefaultTask() {
             AffectedProjectFinder(project, changedGitPaths.toSet(), listOf()).find().map { it.path }
                 .toSet()
 
-        val result = mutableMapOf<String, MutableSet<String>>()
+        val result = project.rootProject.subprojects.associateTo(mutableMapOf()) {
```
rlazo, line 36: wouldn't `associate` work too? Since you are pre-filling the map with the projects, all elements should be there so the statement below only needs to add elements to the inner `MutableMap`

rlazo (approved):

PR #3873, opened 2022-07-06
Set date format locale to english

nabilsunesara:
This patch is to address issue #3757


thatfiredev:
This has been done in #3877

PR #3875, opened 2022-07-06
Add failsafe check.

qdpham13:
- Add backend failsafe check.
- Add app id and sdk version to request
- add boolean check to response
- [go/realtime-rc-sdk-failsafe](http://go/realtime-rc-sdk-failsafe)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -119,7 +119,8 @@ private void handleNotifications(InputStream inputStream) throws IOException {
     BufferedReader reader = new BufferedReader((new InputStreamReader(inputStream, "utf-8")));
     String message;
     while ((message = reader.readLine()) != null) {
-      if (!message.contains("latestTemplateVersionNumber")) {
+      if (!message.contains("latestTemplateVersionNumber")
+          && !message.contains("featureDisabled")) {
```
rlazo, line 123: Without more context I'd assume that the `featureDisabled` would be checked first to short-cut the rest of the cut. Why is it checked last?

rlazo (commented):
Would it make sense to add tests for this?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
qdpham13: I added the one pager for this change to the description. This flag will always be in the response so this check skips any other messages we may get; it doesn't matter the order of the check.

qdpham13: Also added tests.

PR #3878, opened 2022-07-07
basic decoder

ywmei-brt1:
Basic decoder convert a nested map to a Kotlin custom object.


firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+ * @param nestedObject The nested object (map or list) that that needs to be decoded.
+ * @param docRef The [DocumentReference] where this nested object is obtained from.
+ */
+private abstract class FirestoreAbstractDecoder(
+    private val nestedObject: Any,
+    private val docRef: DocumentReference
+) : AbstractDecoder() {
+
+    object Constants {
+        const val START_INDEX = 0
```
rlazo, line 49: Not sure this `const` enhances readability.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * Returns true if the current element being decoded is not null.
+     *
+     * In case this method returns false, [decodeNull] method will be used instead of [decodeValue]
+     * to decode this null value element.
+     */
+    final override fun decodeNotNullMark(): Boolean {
+        val result = isCurrentDecodeElementNotNull
```
rlazo, line 74: Using flags is brittle, since it's up to the developer to ensure it's updated, which is error prone. Could this value be computed?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+        return enumFieldNames.indexOf(decodedEnumFieldName)
+    }
+
+    /**
+     * Returns true if the current element being decoded is not null.
+     *
+     * In case this method returns false, [decodeNull] method will be used instead of [decodeValue]
+     * to decode this null value element.
+     */
+    final override fun decodeNotNullMark(): Boolean {
```
rlazo, line 73: The implementation here changes the value of the flag, which doesn't seem what the interface expects

https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.encoding/-decoder/decode-not-null-mark.html

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+        val result = isCurrentDecodeElementNotNull
+        isCurrentDecodeElementNotNull = true
+        return result
+    }
+
+    /**
+     * Returns a not null primitive value that is going to be assigned as a field of the decoded
+     * [Serializable] object.
+     *
+     * Note: Firestore saves Int values as Long, and Float values as Double from the server side;
```
rlazo, line 83:
```suggestion
     * Note: Firestore saves [Int] values as [Long], and [Float] values as [Double] from the server side;
```

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+     * therefore, during decoding process, depending on the field's [SerialKind], cast might be
+     * required to convert Long back to Int, and to convert Double back to Float.
```
rlazo, line 85: same here

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * Returns a not null value from [decodeValueList]. This is the value that is going to be
+     * decoded as the field value of the [Serializable] object.
+     *
+     * Note: An [IllegalArgumentException] will be thrown if the [elementIndex] is pointing to a
+     * null value from [decodeValueList].
+     */
+    private fun validateDecodedElementNotNull(): Any =
```
rlazo, line 104: Name doesn't match what the method does. You would expect `validate` methods to return true or false based on whether the validation succeeded.  Maybe `decodeElementNotNullOrThrow` ?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+        decodeValueList.elementAt(elementIndex - 1)
+            ?: throw IllegalArgumentException(
+                "Got a null value while trying to decode a not null field."
+            )
+
+    final override val serializersModule: SerializersModule = EmptySerializersModule
+
+    final override fun beginStructure(descriptor: SerialDescriptor): CompositeDecoder {
+        val innerCompositeObject = getCompositeObject(elementIndex)
```
rlazo, line 113: It's very possible that the object at `elementIndex` is not composite. We, the ones with an object at `elementIndex` **are** composite for sure. Inner object is a better name.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+ * (i.e. the name/type of each property to be encoded, the annotations on each property).
+ *
+ * Based on the descriptor’s information, during the decoding process, a nested map will be send to
+ * the decoder. Decoder will loop through all the key-value pairs in the nested map, use the key to
+ * find the target field from the [SerialDescriptor], then assign the value to the field.
+ *
+ * @param nestedObject The nested map that that needs to be decoded to a @[Serializable] object.
+ * @param docRef The [DocumentReference] where this nested map is obtained from.
+ */
+private class FirestoreMapDecoder(nestedObject: Map<String, Any?>, docRef: DocumentReference) :
```
rlazo, line 158: nestedobject is actually a map.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+    private val nestedMap = nestedObject
+    private val decodeNameList: List<String>
+    override val decodeValueList: List<Any?>
+
+    /**
+     * Returns a list of keys and values that need to be decoded as fields of the [Serializable]
+     * object.
+     */
+    init {
+        val (decodeNameList, decodeValueList) = nestedObject.toList().unzip()
```
rlazo, line 170: here the fact that it's called `nestedObject` makes the code confusing.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+ */
+private class FirestoreMapDecoder(nestedObject: Map<String, Any?>, docRef: DocumentReference) :
+    FirestoreAbstractDecoder(nestedObject, docRef) {
+
+    private val nestedMap = nestedObject
+    private val decodeNameList: List<String>
+    override val decodeValueList: List<Any?>
+
+    /**
+     * Returns a list of keys and values that need to be decoded as fields of the [Serializable]
```
rlazo, line 166: init blocks don't return anything

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+     * Returns the index of the element to be decoded. Index represents a position of the current
+     * element in the serial descriptor element that can be found with
+     * [SerialDescriptor.getElementIndex]. Additional to the element index, this method also returns
+     * [CompositeDecoder.DECODE_DONE] to indicate the decoding process is finished, and returns
+     * [CompositeDecoder.UNKNOWN_NAME] to indicate that the element to be decode is not in
+     * [descriptor].
+     *
+     * @param descriptor The [SerialDescriptor] of the [Serializable] object.
+     */
+    override fun decodeElementIndex(descriptor: SerialDescriptor): Int {
```
rlazo, line 185: no need to prefix all with `decode`

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/TestUtil.kt
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 Google LLC
```
rlazo, line 1: Sync PRs to include the deletion of this code from the non test code.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/UnitTestUtilTests.kt
```diff
@@ -0,0 +1,61 @@
+// Copyright 2022 Google LLC
```
rlazo, line 1: Code in testutil seems trivial enough that this test is not necessary.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
```diff
@@ -0,0 +1,212 @@
...
+package com.google.firebase.firestore.ktx
+
+import com.google.common.truth.Truth.assertThat
+import com.google.firebase.firestore.assertThrows
+import com.google.firebase.firestore.documentReference
+import com.google.firebase.firestore.ktx.serialization.decodeFromMap
+import kotlinx.serialization.Serializable
+import org.junit.Test
+
+class FirestoreMapDecoderTests {
```
rlazo, line 24: runwith

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
```diff
@@ -0,0 +1,212 @@
...
+    private val firestoreDocument = documentReference("abc/1234")
```
rlazo, line 25: do this in the @before method

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -0,0 +1,244 @@
...
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.modules.EmptySerializersModule
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.serializer
+
+/**
+ * A skeleton implementation of a decoder that decodes a nested map (or list) of Firestore supported
+ * types as a @[Serializable] object (or its field).
+ *
+ * Most of the decode* methods have default implementation that delegates decodeValue(value: Any).
```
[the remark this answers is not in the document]
ywmei-brt1: Changed to delegate to!

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: Very interesting, the `const` keyword helps prevent the generated getter in the Java decomplied code:

1. This is the current code's decomplied Java code:
```
public static final int START_INDEX = 0;
```
2. If I remove the `const` keyword, there will be a getter generated for it like a property:
```
      private static final int START_INDEX;
      @NotNull
      public static final FirestoreAbstractDecoder.Constants INSTANCE;

      public final int getSTART_INDEX() {
         return START_INDEX;
      }
      ```

File: build.gradle [hunk 17]
```diff
@@ -16,7 +16,7 @@ import com.google.firebase.gradle.plugins.license.LicenseResolverPlugin
 import com.google.firebase.gradle.MultiProjectReleasePlugin
 
 buildscript {
-    ext.kotlinVersion = '1.6.20'
+    ext.kotlinVersion = '1.3.72'
```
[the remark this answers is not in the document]
ywmei-brt1: Sorry, I missed this from a version compatibility test, I changed this back now.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: Thank you for the new name. I changed to `decodedElementNotNullOrThrow`

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: Actually, when this `beginStructure()` method is triggered, it is guaranteed from the `descriptor`'s meta data that the element need to be decoded is a composite element.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: Changed the verb to `Separates`

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: This is the method name I have to override from the parent class `CompositeDecoder.kt`
https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization.encoding/-composite-decoder/decode-element-index.html

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/TestUtil.kt
ywmei-brt1: I have synced the PRs now, but these two are the test codes I need:
1. `documentReference` is a dummy docRef with firestore set to be `null`, this is needed for unit test this decoder, because decoder's constructor requires a not null `documentReference`;
2. `assertThrows` is the method to check throws with Truth, this is used for both integration test and unit test.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/UnitTestUtilTests.kt
ywmei-brt1: Thank you, I deleted this file. I was not sure if I should unit test test utilities before.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
ywmei-brt1: created a static field for `firestoreDocument` and initialized with @BeforeClass, so I only need to initialize it once for all the tests.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
ywmei-brt1: following the `MapTester`'s example, I am using `@RunWith(RobolectricTestRunner::class)` now.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: Haha, I see now~ thank you.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: this parameter I changed the name to `val nestedMap`, the same parameter for list decoder, I changed to `override val decodeValueList`

firebase-firestore/ktx/src/androidTest/java/com/google/firebase/firestore/FirestoreMapEncoderIntegrationTest.kt
```diff
@@ -90,10 +90,11 @@ class FirestoreMapEncoderIntegrationTest {
             )
 
         for (student in studentList) {
-            assertFailsWith<IllegalArgumentException>(
-                message = "not supported, please use",
-                block = { docRefKotlin.setData(student) }
-            )
+            assertThrows<IllegalArgumentException> {
```
rlazo, line 93: Why the change?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
rlazo: I meant the variables inside :-)

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
rlazo: Unless you need to mock parts of android OS roboelectric is overkill

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapEncoderTests.kt
```diff
@@ -15,12 +15,15 @@
...
+import com.google.firebase.firestore.assertThrows
 import com.google.firebase.firestore.ktx.serialization.encodeToMap
-import kotlin.test.assertFailsWith
 import kotlinx.serialization.Serializable
 import org.junit.Assert.assertTrue
 import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
 
+@RunWith(RobolectricTestRunner::class)
```
rlazo, line 26: Same

firebase-firestore/ktx/src/androidTest/java/com/google/firebase/firestore/FirestoreMapEncoderIntegrationTest.kt
ywmei-brt1: As you suggested before, I am using a high order test util function to use `Truth` under the hood to do this assertThrow check, so this makes it easy for me to debug when it fails. 

The original one uses the android test JUnit library, I have a TODO in my ktx.gradle file to remove this dependency after I merge this decoder PR (because this Truth assertThrow helpper method is only in this decoder PR so far).

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
rlazo: Do using `@RunWith(Junit4::class)` work?

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapEncoderTests.kt
ywmei-brt1: updated with `@RunWith(JUnit4::class)` now.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
ywmei-brt1: Yes, it works. Thank you.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapDecoderTests.kt
```diff
@@ -0,0 +1,226 @@
...
+
+    companion object {
+        lateinit var firestoreDocument: DocumentReference
+
+        @BeforeClass
+        @JvmStatic
+        fun generateDocRef() {
+            firestoreDocument = documentReference("abc/1234")
+        }
+    }
```
rlazo, line 39: What's the reason this is created here instead of in a regular `@Before` method?

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
ywmei-brt1: I think we do not need a getter here, I think Kotlin is recommend us to use `const` whenever we can: 
The advantages of keep this number as a compile time `const` are:
1. the decompiled Java code is shorter,
2. this `const` will be inlined (directly use the primitive value to replace the variable in the compiled code), my understanding is this can reduce the runtime heap

rlazo (approved):

PR #3900, opened 2022-07-14
Ensure Client exists

wu-hui:
Fixes b/Fix 238932261


cherylEnkidu (approved):
Thank you for your fix!

PR #3912, opened 2022-07-18
Update FirebaseMessagingDirectBootReceiver to match FCM's FirebaseIns…

gsakakihara:
…tanceIdReceiver.

* Changed FirebaseMessagingDirectBootReceiver to match the FCM SDK version of FirebaseInstanceIdReceiver, using FCM's methods in order to remove dependency on firebase-iid.
* Cleaned up dependencies.
* Updated minSdkVersion to 19 to match FCM SDK's minSdkVersion.

### DIFF
--- a/firebase-messaging-directboot/src/main/java/com/google/firebase/messaging/directboot/FirebaseMessagingDirectBootReceiver.java
+++ b/firebase-messaging-directboot/src/main/java/com/google/firebase/messaging/directboot/FirebaseMessagingDirectBootReceiver.java
@@ -14,20 +14,20 @@
 package com.google.firebase.messaging.directboot;
 
 import android.content.Context;
-import android.content.Intent;
 import android.util.Log;
-import androidx.legacy.content.WakefulBroadcastReceiver;
-import com.google.android.gms.common.util.concurrent.NamedThreadFactory;
-import com.google.firebase.iid.FcmBroadcastProcessor;
-import com.google.firebase.iid.ServiceStarter;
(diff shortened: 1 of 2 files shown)


firebase-messaging-directboot/src/main/java/com/google/firebase/messaging/directboot/FirebaseMessagingDirectBootReceiver.java
```diff
@@ -49,58 +49,20 @@
...
-          .newSingleThreadExecutor(
-              new NamedThreadFactory("fcm-db-intent-handle"), ThreadPriority.LOW_POWER);
-
+  /** @hide */
   @Override
-  public void onReceive(Context context, Intent intent) {
-    if (intent == null) {
-      return;
-    }
-    if (!ACTION_DIRECT_BOOT_REMOTE_INTENT.equals(intent.getAction())) {
```
willxu-google, line 72: Is this intent preserved somewhere else or is it no longer necessary?

gsakakihara:
It's now implementing CloudMessagingReceiver.onMessageReceive() instead of BroadcastReceiver.onReceive(), which handles all of that now. This is updating to the FCM version of FirebaseInstanceIdReceiver (which FirebaseMessagingDirectBootReceiver should be mirroring, as it should do basically the same as the normal receiver, just for Direct Boot messages). See https://github.com/firebase/firebase-android-sdk/blob/master/firebase-messaging/src/main/java/com/google/firebase/iid/FirebaseInstanceIdReceiver.java.

willxu-google (approved):

PR #3917, opened 2022-07-19
Handle API disabled errors in firebase-appdistribution

lfkellogg:
See [go/fad-sdk-auth-errors](http://go/fad-sdk-auth-errors) for more information.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiDisabledErrorDetails.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiDisabledErrorDetails.java
@@ -1,0 +1,110 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.appdistribution.impl;
+
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+import java.util.ArrayList;
+import java.util.List;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Details about a {@code SERVICE_DISABLED} error returned by the Tester API.
+ *
+ * <p>Error structure is described in the <a
+ * href="https://cloud.google.com/apis/design/errors#error_details">Cloud APIs documentation</a>.
+ */
+@AutoValue
+abstract class TesterApiDisabledErrorDetails {
+
+  @AutoValue
+  abstract static class HelpLink {
+    abstract String description();
+
+    abstract String url();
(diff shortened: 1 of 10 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiHttpClient.java
```diff
@@ -255,6 +252,22 @@ private static FirebaseAppDistributionException getExceptionForHttpResponse(
...
+    // Check if this is an API disabled error
+    TesterApiDisabledErrorDetails apiDisabledErrorDetails =
+        TesterApiDisabledErrorDetails.tryParse(responseBody);
+    if (apiDisabledErrorDetails != null) {
+      String messageWithHelpLinks =
+          String.format(
+              "%s\n\n%s", ErrorMessages.API_DISABLED, apiDisabledErrorDetails.formatLinks());
+      return getException(tag, messageWithHelpLinks, Status.API_DISABLED);
+    }
+
```
jladieu, line 266: Would it be worth logging the responseBody in this case (for example, so the developer could print the logs in `--verbose` or `--debug` somehow?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiDisabledErrorDetails.java
```diff
@@ -0,0 +1,110 @@
...
+      JSONObject detail = details.getJSONObject(i);
+      if (detail.getString("@type").equals(type)) {
+        return detail;
+      }
+    }
+    throw new JSONException("No detail present with type: " + type);
+  }
+
+  static List<HelpLink> parseHelpLinks(JSONArray details) {
+    List<HelpLink> helpLinks = new ArrayList<>();
```
jladieu, line 91: Is `ImmutableList.Builder` an option here instead of the mutable variety?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TesterApiDisabledErrorDetailsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.firebase.appdistribution.impl.TesterApiDisabledErrorDetails.HelpLink;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
```
jladieu, line 27: Great coverage!

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiHttpClient.java
lfkellogg: It would! We actually print the response body out in verbose mode already: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiHttpClient.java#L183

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiDisabledErrorDetails.java
lfkellogg: Unfortunately we'd have to pull in guava. I think it's pretty large which is why the SDKs have avoided pulling in that dependency.

Issue #2014, opened 2020-09-29
ANR


RodolfoGS:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 4.0
  * Firebase Component: Messaging (Database, Firestore, Storage, Functions, etc)
  * Component version: 20.2.4

### [REQUIRED] Step 3: Describe the problem

Appear tons of ANR errors in Play Console related to firebase messaging. And a lot of users have the same problem and are reporting in a closed issue of 2 years ago (2018).

Issue closed that continues active with more than 50 comments and 33 participants https://github.com/firebase/quickstart-android/issues/594

![image](https://user-images.githubusercontent.com/870365/94559403-d0cc6100-0226-11eb-89da-d63849810d57.png)

#### Steps to reproduce:

I can't reproduce, I just see the issues in the play console.

#### Relevant Code:

```
Broadcast of Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x11000010 pkg=xxx cmp=xxx/com.google.firebase.iid.FirebaseInstanceIdReceiver (has extras) }, InvisibleToUser

com.google.firebase.iid.FirebaseInstanceIdReceiver


"main" prio=5 tid=1 Runnable
  | group="main" sCount=0 dsCount=0 flags=4 obj=0x713d3fc8 self=0x788ebc1c00
  | sysTid=22434 nice=0 cgrp=default sched=0/0 handle=0x7890126ed0
  | state=R schedstat=( 310821836 210351148 893 ) utm=8 stm=22 core=0 HZ=100
  | stack=0x7fe8780000-0x7fe8782000 stackSize=8192KB
  | held mutexes= "mutator lock"(shared held)
  at com.google.firebase.internal.DataCollectionConfigStorage.directBootSafe (DataCollectionConfigStorage.java)
[...]

PR #3919, opened 2022-07-19
Finish message broadcast after sending Intent to bound Service.

gsakakihara:
* Changed to finish the message broadcast as soon as the Intent is sent to the FirebaseMessagingService when binding to the Service. This should reduce the chance that the BroadcastReceiver ANRs in certain circumstances.
* Acquired a WakeLock when using bindService() to send an Intent to FirebaseMessagingService to prevent the device from sleeping while processing the Intent now that the broadcast is finished before the Service has finished handling it.
  - Holding the WakeLock for up to 60 seconds for high priority messages to match the behavior for high priority messages when using startService().
  - Holding the WakeLock for up to 9 seconds for normal priority messages to match the previous behavior where the broadcast was not forced to finish for 9 seconds.

### DIFF
--- a/firebase-messaging/src/main/java/com/google/firebase/messaging/FcmBroadcastProcessor.java
+++ b/firebase-messaging/src/main/java/com/google/firebase/messaging/FcmBroadcastProcessor.java
@@ -118,11 +118,14 @@ public class FcmBroadcastProcessor {
     if (Log.isLoggable(TAG, Log.DEBUG)) {
       Log.d(TAG, "Binding to service");
     }
+    if (ServiceStarter.getInstance().hasWakeLockPermission(context)) {
+      WakeLockHolder.sendWakefulServiceIntent(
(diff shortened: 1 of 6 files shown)


James201311 (approved):
LGTM!! 🚢🚢🚢

shashachu:
Do we know if this is related to either of these ANR issues reported? https://github.com/firebase/firebase-android-sdk/issues/2014 https://github.com/firebase/firebase-android-sdk/issues/3052

gsakakihara:
https://github.com/firebase/firebase-android-sdk/issues/3052 No, that one was happening during Application initialization, before the FCM SDK has a chance to do anything.

https://github.com/firebase/firebase-android-sdk/issues/2014 That issue has a lot of different reports that seem to have a number of causes. This would not help the original report for example (the ANR happened during FirebaseAppImpl.getInstance(), which is also before the FCM SDK does anything), but might improve some of the other reports. It's difficult to tell without more information.

PR #3922, opened 2022-07-20
share annotations between java and kotlin

ywmei-brt1:
1. Add kotlin reflection lib to Java lib,
2. Extends java `CustomerClassMapper` to conduct reflection on each field can be converted to kotlin property, to get annotations on properties.
3. Fixed all the ktx test to reflect this change


firebase-firestore/ktx/src/androidTest/java/com/google/firebase/firestore/ComponentRegistrarTest.kt
```diff
@@ -97,4 +98,42 @@ class ComponentRegistrarTest {
...
+        // Kotlin can see this @DocumentId annotation
+        val annotations = DocumentIdObj.serializer().descriptor.getElementAnnotations(0)
+        val annotation = annotations[0]
+        annotation::class.java isAssignableTo DocumentId::class.java
+
+        // Java POJO should also see this annotation and use it to skip the annotated field
+        val docRef = testCollection("pojo").document("456")
+        docRef.withoutCustomMappers { set(DocumentIdObj()) }
+        val pojoMap = waitFor(docRef.get()).data
+        pojoMap shouldBe emptyMap
```
ywmei-brt1, line 116: I have changed the Java custom mapper to make Java code able to capture annotations on Property. This is the test.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -912,7 +915,8 @@ Map<String, Object> serialize(T object, ErrorPath path) {
     }
 
     private void applyFieldAnnotations(Field field) {
-      if (field.isAnnotationPresent(ServerTimestamp.class)) {
+      if (field.isAnnotationPresent(ServerTimestamp.class)
+          || isAnnotationPresentOnProperty(field, ServerTimestamp.class)) {
```
ywmei-brt1, line 919: The trick is to use Kotlin reflection do another round of reflection to get annotations on properties.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -925,13 +929,24 @@ private void applyFieldAnnotations(Field field) {
...
+          || isAnnotationPresentOnProperty(field, DocumentId.class)) {
         Class<?> fieldType = field.getType();
         ensureValidDocumentIdType("Field", "is", fieldType);
         documentIdPropertyNames.add(propertyName(field));
       }
     }
 
+    private boolean isAnnotationPresentOnProperty(
+        Field field, Class<? extends Annotation> annotationType) {
+      KProperty<?> property = ReflectJvmMapping.getKotlinProperty(field);
```
ywmei-brt1, line 942: use Kotlin reflection to get Kotlin property and its annotation from Java field

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.kt
```diff
@@ -38,7 +38,7 @@ import kotlinx.serialization.SerialInfo
 @SerialInfo
 @Retention(RetentionPolicy.RUNTIME)
 @Target(
-// TODO: revert this file back to Java or find out a method to share this file between Java and Ktx.
+    AnnotationTarget.PROPERTY,
     AnnotationTarget.FUNCTION,
     AnnotationTarget.PROPERTY_GETTER,
     AnnotationTarget.PROPERTY_SETTER,
```
thatfiredev, line 44: Do we really need these? I see they were removed from `ServerTimestamp`

ywmei-brt1: Sorry, I missed these two lines, I just removed them. We do not need these as Kotlin serialization annotations on setter and getter will be transparent to the compiler plugin. Thank you.

firebase-firestore/src/main/java/com/google/firebase/firestore/DocumentId.kt
```diff
@@ -37,11 +38,6 @@ import kotlinx.serialization.SerialInfo
  */
 @SerialInfo
 @Retention(RetentionPolicy.RUNTIME)
```
vkryachko, line 40: please remove retention, it defaults to runtime in kotlin, if you prefer to keep it, please switch it to kotlin.Retention

firebase-firestore/src/main/java/com/google/firebase/firestore/ServerTimestamp.kt
```diff
@@ -24,12 +25,6 @@ import kotlinx.serialization.SerialInfo
  */
 @SerialInfo
 @Retention(RetentionPolicy.RUNTIME)
```
vkryachko, line 27: same here

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -912,7 +920,8 @@ Map<String, Object> serialize(T object, ErrorPath path) {
     }
 
     private void applyFieldAnnotations(Field field) {
```
vkryachko, line 922: I don't think that's the right place to check for getters, instead we should modify applyGetterAnnotations, I would really like for someone on the firestore team to take a look tbh.

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -925,13 +934,24 @@ private void applyFieldAnnotations(Field field) {
...
+        Field field, Class<? extends Annotation> annotationType) {
+      String fieldName = propertyName(field);
+      if (ktxGeneratedGetters.containsKey(fieldName)) {
+        return ktxGeneratedGetters.get(fieldName).isAnnotationPresent(annotationType);
+      }
+      return false;
+    }
+
     private void applyGetterAnnotations(Method method) {
       if (method.isAnnotationPresent(ServerTimestamp.class)) {
```
vkryachko, line 956: hm, I don't understand this code too well, but it seems unnecessary to have a dedicated kotlin getter map(above). how about just modifying `applyGetterAnnotations()` with something like the following:



```suggestion
    private static boolean isGetterAnnotationPresent(
        Method method,
        Class<? extends Annotation> annotation) {
      if (method.isAnnotationPresent(annotation)) { return true; }
      try {
        Method getterAnnotationsMethod = method
            .getClass()
            .getMethod(method.getName() + "$annotations");
            return method.isSynthetic() && method.isAnnotationPresent(annotation);
      } catch (NoSuchMethodException ex) {
        return false;
      }
    }
    private void applyGetterAnnotations(Method method) {
      if (isGetterAnnotationPresent(method, ServerTimestamp.class)) {
```

And the same for DocumentId below. this should be the only change required afaict.

But we should definitely check with someone on the firestore team as well.

ywmei-brt1: Thank you for this code, it works with some minor changes. This is way shorter than my `applyFieldAnnotations` version, thank you.

firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -70,7 +70,12 @@ private abstract class FirestoreKtxAbstractDecoder(
         // TODO: Add a EnumNamingProperties parameter, and convert decodedEnumFieldName based on it
         // i.e. case insensitive, snake_case match camelCase, etc
         val enumFieldNames = enumDescriptor.elementNames.toList()
```
rlazo, line 72: You could do

```
val index = enumDescriptor.elementNames.indexOf(decodedEnumFieldName)
        if (index == -1)
            throw IllegalArgumentException(
                "Could not find a match for enum field name of $decodedEnumFieldName."
            )
        return index
```

that way you don't create a list unnecessarily

ywmei-brt1: I didn't realize that Kotlin has `indexOf` buildin for iterables!!! This is so cool. 
 I updated this method, and added a unit test for the throw message. Thank you.

firebase-firestore/ktx/src/androidTest/java/com/google/firebase/firestore/ComponentRegistrarTest.kt
```diff
@@ -97,4 +98,42 @@ class ComponentRegistrarTest {
...
+
+    @Serializable
+    private data class DocumentIdObj(@DocumentId val docId: String = "should be null in Java")
+
+    @Test
+    fun docId_annotation_should_be_shared_between_java_and_kotlin() {
+        // Kotlin can see this @DocumentId annotation
+        val annotations = DocumentIdObj.serializer().descriptor.getElementAnnotations(0)
+        val annotation = annotations[0]
+        assertThat(annotation::class.java).isAssignableTo(DocumentId::class.java)
```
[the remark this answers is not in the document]
ywmei-brt1: done~

firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -932,8 +933,23 @@ private void applyFieldAnnotations(Field field) {
       }
     }
 
+    private static boolean isGetterAnnotationPresent(
+        Method method, Class<? extends Annotation> annotation) {
+      if (method.isAnnotationPresent(annotation)) {
+        return true;
+      }
+      try {
+        Method getterAnnotationsMethod =
```
vkryachko, line 942: Let's add a very detailed comment on what is going on here. what this special method is and why we need it.

vkryachko: Can you also add this url to your comment for reference pls?
https://github.com/JetBrains/kotlin/blob/92109a6bef9e18075996637035b63a95cf76230d/core/compiler.common.jvm/src/org/jetbrains/kotlin/load/java/JvmAbi.kt#L59

ywmei-brt1: comments updated. Thank you.

rlazo (approved):

PR #3930, opened 2022-07-21
Add some more details to firebase-appdistribution/test-app/README.md

lfkellogg:


jladieu (approved):
Thanks for refining this.  I'll try these steps out.

PR #3931, opened 2022-07-21
feat(rtdb): create suspend function to run transactions

thatfiredev:
This is the implementation of go/rtdb-transaction-ktx which proposes an out-of-the-box suspend function for developers to run transactions on the Realtime Database.

### DIFF
--- a/firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
+++ b/firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
@@ -18,12 +18,18 @@ import androidx.annotation.Keep
 import com.google.firebase.FirebaseApp
 import com.google.firebase.components.Component
 import com.google.firebase.components.ComponentRegistrar
+import com.google.firebase.database.DatabaseError
+import com.google.firebase.database.DatabaseReference
 import com.google.firebase.database.DataSnapshot
 import com.google.firebase.database.FirebaseDatabase
 import com.google.firebase.database.GenericTypeIndicator
 import com.google.firebase.database.MutableData
+import com.google.firebase.database.Transaction
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.platforminfo.LibraryVersionComponent
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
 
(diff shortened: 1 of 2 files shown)


firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
```diff
@@ -59,6 +65,39 @@ inline fun <reified T> MutableData.getValue(): T? {
...
+            currentData: DataSnapshot?
+        ) {
+            if (error != null) {
+                continuation.resumeWithException(error.toException())
+            } else {
+                if (currentData != null) {
+                    continuation.resume(currentData)
+                }
+            }
+        }
```
thatfiredev, line 97: I just realized I've ignored 2 things when implementing this:
- the `committed` boolean
- the case where both `error` and `currentData` are null: I assumed it would never happen, but it seems like it can as seen in this test:
    https://github.com/firebase/firebase-android-sdk/blob/fe17099d7a156e22faca77bef41b1b56a5116fd9/firebase-database/src/androidTest/java/com/google/firebase/database/TransactionTest.java#L150-L173

thatfiredev: Alternatives Considered:

1. Have this method return a [`Pair<DataSnapshot, Boolean>`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-pair/) to able to return both the committed boolean and the new data at location. I don't remember seeing any API return a Pair, but its usage isn't so different from destructure declarations which we already have in our SDK.
    ```kotlin
    try {
        val (snapshot, committed) = postRef.runTransaction { mutableData: MutableData ->
             // (actual transaction code omitted for brevity)
             return@runTransaction Transaction.success(mutableData)
        }
        // transaction succeeded and currentData has the current data at the location
    } catch (e: DatabaseException) {
        // handle errors
    }
    ```
1. Only return a `DataSnapshot` in cases where the transaction was successful. Aborted transactions will return a null `DataSnapshot` and won't throw an exception.

thatfiredev: After giving it some thought, I think Alternative 1 makes more sense as it doesn't change the current RTDB transaction behavior.

Alternative 2 can be consider as a behavior change if developers are migrating from the current callback to this new suspend function.

PR #3934, opened 2022-07-22
Update messaging around API_DISABLED status

lfkellogg:
Based on feedback from vlukanovich@

### DIFF
--- a/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionException.java
+++ b/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionException.java
@@ -83,11 +83,11 @@ public class FirebaseAppDistributionException extends FirebaseException {
     /**
      * The Firebase App Distribution Tester API is disabled for this project.
      *
-     * <p>The developer of this app must enable the API in the Google Cloud Console before using the
-     * App Distribution SDK. See the <a
-     * href="https://firebase.google.com/docs/app-distribution/set-up-alerts?platform=android">documentation</a>
-     * for more information. If you enabled this API recently, wait a few minutes for the action to
-     * propagate to our systems and retry.
+     * <p>Before you use the App Distribution SDK in your app, you must enable the API in the Google
+     * Cloud console. For more information, see the <a
+     * href="https://firebase.google.com/docs/app-distribution/set-up-alerts?platform=android">documentation</a>.
+     * If you enabled this API recently, wait a few minutes for the action to propagate to the App
+     * Distribution systems, and retry.
      */
     API_DISABLED,
   }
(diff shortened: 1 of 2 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
```diff
@@ -53,7 +53,7 @@ class ErrorMessages {
       "The APK failed to install or installation was canceled by the tester.";
 
   static final String API_DISABLED =
-      "The App Distribution Tester API is disabled. It must be enabled in the Google Cloud Console. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.";
+      "The App Distribution Tester API is disabled; you must enable it in the Google Cloud console. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.";
```
kaibolay, line 56: Could we directly link to the right place in pantheon here? I've seen that in other error messages.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionException.java
```diff
@@ -83,11 +83,11 @@ public enum Status {
     /**
      * The Firebase App Distribution Tester API is disabled for this project.
      *
-     * <p>The developer of this app must enable the API in the Google Cloud Console before using the
-     * App Distribution SDK. See the <a
-     * href="https://firebase.google.com/docs/app-distribution/set-up-alerts?platform=android">documentation</a>
-     * for more information. If you enabled this API recently, wait a few minutes for the action to
-     * propagate to our systems and retry.
+     * <p>Before you use the App Distribution SDK in your app, you must enable the API in the Google
```
kaibolay, line 86: Could we directly link to the right place in pantheon here? I've seen that in other error messages.

kaibolay (approved):

jladieu (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ErrorMessages.java
lfkellogg: Actually we do! We pull it out of the response and append it to the message: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiHttpClient.java#L261-L263

You can see an example here: [go/fad-iaf-api-disabled](http://go/fad-iaf-api-disabled)

kaibolay: Thanks!

PR #3945, opened 2022-07-27
ktx reflection to get both Java and Ktx annotations

ywmei-brt1:
1. Using Ktx reflection (for @SerialName) and Java reflection (for @DocumentId, and @Servertimestamp) inside of Ktx lib to get both of the target == PROPERTIES and target == FIELD annotations;
2. Removed the Ktx reflection inside of Java lib
3. Converted @DocumentId and @ServerTimestamp back to Java files
4. Currently, this solution not able to handle recursive Generic types.


firebase-firestore/ktx/src/main/kotlin/com/google/firebase/firestore/ktx/serialization/FirestoreMapDecoder.kt
```diff
@@ -270,11 +308,43 @@ private class FirestoreListDecoder(
...
+        currentDecodeElement = Element(decodeValueList[elementIndex], elementIndex, null)
         return elementIndex++
     }
 }
 
+/**
+ * A data class that stores the information of all the annotations (both Java and Ktx) associated
+ * with each field name, and the Java class types associated with each field name.
+ */
+class JavaAndKtxAnnotationMapper(clazz: Class<*>) {
```
ywmei-brt1, line 320: This is the main logic of reflection, I have to use Kotlin reflection to guide Java reflection (java reflection cannot get all the fields from super class, while Kotlin has this native method).

Also need Kotlin reflection to get the @SerialName, so that I can match the java annotations' owner with the elements in `descriptor`.

Need to handle `List<CustomClass>` with some Kotlin trick, Java will erase generic type, While Kotlin can obtain this from `KType` (not `KClass`)

BUT, I am not 100% sure if this method is possible to support recursive generic type or not.

firebase-firestore/ktx/src/test/kotlin/com/google/firebase/firestore/ktx/FirestoreMapEncoderTests.kt
```diff
@@ -234,11 +274,12 @@ class FirestoreMapEncoderTests {
...
 
-        val innerObj = GenericObject("foo")
-        val recursiveObj = GenericObject(innerObj)
-        val encodedRecursiveObj = encodeToMap(recursiveObj)
-        val expectedRecursiveObj = mutableMapOf("value" to mutableMapOf("value" to "foo"))
-        assertThat(encodedRecursiveObj).containsExactlyEntriesIn(expectedRecursiveObj)
+//        TODO: Support obtain annotations from Recurisvie Genetric Objects
+//        val innerObj = GenericObject("foo")
+//        val recursiveObj = GenericObject(innerObj)
+//        val encodedRecursiveObj = encodeToMap(recursiveObj)
```
ywmei-brt1, line 280: This recursive obj is currently thrown. I cannot get the `Class` for the nested Generic Types, but I need this `Class` information to get annotations.

But if we decide to not handle any annotations inside of nested GenetricObjects, I can easily make it pass.

PR #3947, opened 2022-07-27
try to get java annotations

ywmei-brt1:


firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -928,24 +927,23 @@ private void applyFieldAnnotations(Field field) {
         }
         serverTimestamps.add(propertyName(field));
       }
-
-      if (field.isAnnotationPresent(DocumentId.class)
-          || isAnnotationPresentOnProperty(field, DocumentId.class)) {
+      if (field.getAnnotationsByType(DocumentId.class).length != 0) {
```
ywmei-brt1, line 930: this `getAnnotationsByType` requires API 24 (current is 19), and it does not work. I always get `length == 0` here at this line.

PR #3954, opened 2022-07-29
Add workflow definition for health metric tests

yifanyang:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/AffectedProjectFinder.java
```diff
@@ -68,10 +68,10 @@ Set<Project> find() {
 
   private static Set<String> changedPaths(File workDir) {
     try {
-      // works on Prow only.
+      // works on CI only.
       Process process =
           Runtime.getRuntime()
-              .exec("git diff --name-only --submodule=diff HEAD@{0} HEAD@{1}", null, workDir);
+              .exec("git diff --name-only --submodule=diff HEAD^1 HEAD", null, workDir);
```
vkryachko, line 74: What's the motivation for this change? I don't think these are equivalent, afaicr `HEAD@{0} HEAD@{1}` produces a diff based on reflog, while `HEAD^1 HEAD` just diffs the last commit. wouldn't it break the rest of our prow jobs?

yifanyang: The commit [`9ceb2a8`](https://github.com/firebase/firebase-android-sdk/pull/3954/commits/9ceb2a8b31d2133a9ab756737f6b0a5175b13388) in this pull request outputs the git tree and the value of refs `HEAD@{0}`, `HEAD@{1}`, `HEAD^1`, `HEAD`. These are the logs from size job on GHA and Prow respectively:
- GHA (presubmit): https://github.com/firebase/firebase-android-sdk/runs/7635586138?check_suite_focus=true#step:9:72
- Prow (presubmit): https://android-ci.firebaseopensource.com/view/gcs/android-ci/pr-logs/pull/firebase_firebase-android-sdk/3954/binary-size/1554498146170572805/#0:build-log.txt%3A57
- Prow (post-submit): https://android-ci.firebaseopensource.com/view/gcs/android-ci/logs/postsubmit-binary-size/1554498146170572802#0:build-log.txt%3A56

To my observation:
- The GHA (presubmit) log shows that there is no `HEAD@{1}` on GitHub Actions
- The Prow (post-submit) log shows that `HEAD@{0}` and `HEAD@{1}` are always the same

To my understanding:
  - in presubmit case `HEAD@{0}`, `HEAD@{1}` and `HEAD`, `HEAD^1` are equivalent
  - in post-submit case, diff between `HEAD^1` and `HEAD` is what we actually intend to have

Does it make sense?

yifanyang: @vkryachko Can you take another look at this change?

And this time the coverage task for all affected products passed, finally.

vkryachko (approved):

PR #3965, opened 2022-08-03
Realtime rc failsafe

qdpham13:
Add back off flag to Realtime. If enabled, this will prevent Realtime from working until the app restarts. 
Backend will have a new field called featureDisabled added to stream response. If this field is enabled, this will set stop Realtime from working until the next time the app is restarted. The current back off state in the SDK will be represented by the flag  isBackoffEnabled.


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -106,32 +119,43 @@ public void listenForNotifications() {
...
   // Auto-fetch new config and execute callbacks on each new message
   private void handleNotifications(InputStream inputStream) throws IOException {
     BufferedReader reader = new BufferedReader((new InputStreamReader(inputStream, "utf-8")));
     String message;
+    String fullMessage = "";
     while ((message = reader.readLine()) != null) {
-      if (!message.contains("latestTemplateVersionNumber")) {
-        continue;
-      }
+      fullMessage += message;
```
danasilver, line 132: I'm having trouble following the parsing logic. Could we parse the JSON response first, then get the fields we need? Example in fetch: https://github.com/firebase/firebase-android-sdk/blob/c6cc8c5d4a00518033a6f281486c9610b821e1c4/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java#L372-L383

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -106,32 +119,43 @@ public void listenForNotifications() {
...
+
+      if (message.contains("}")) {
+        fullMessage = parseMessage(fullMessage);
+        if (!fullMessage.isEmpty()) {
+          long targetTemplateVersion = configFetchHandler.getTemplateVersionNumber();
+          try {
+            JSONObject jsonObject = new JSONObject(fullMessage);
+            if (jsonObject.has("latestTemplateVersionNumber")) {
```
danasilver, line 140: Would make these strings constants in the class

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -271,12 +283,17 @@ public void onEvent() {
           }
 
           @Override
-          public void onError(Exception error) {}
+          public void onError(Exception error) {
```
danasilver, line 286: Should we check this is the realtime disabled exception?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -288,9 +305,9 @@ synchronized void beginRealtimeHttpStream() {
         resetRetryParameters();
         ConfigAutoFetch configAutoFetch = startAutoFetch(httpURLConnection);
         configAutoFetch.listenForNotifications();
-      } else {
-        retryHTTPConnection();
+        closeRealtimeHttpStream();
       }
+      retryHTTPConnection();
```
danasilver, line 310: Looks like we'll always retry the connection even after closing it above?

danasilver (changes requested):

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -1134,6 +1134,34 @@ public void realtime_stream_listen_fail() throws Exception {
...
 
+  @Test
+  public void realtime_stream_listen_and_failsafe_enabled() throws Exception {
+    when(mockHttpURLConnection.getResponseCode()).thenReturn(200);
+    when(mockHttpURLConnection.getInputStream())
+        .thenReturn(
+            new ByteArrayInputStream(
+                "{ \"featureDisabled\": true }".getBytes(StandardCharsets.UTF_8)));
+    when(mockFetchHandler.getTemplateVersionNumber()).thenReturn(1L);
+    when(mockFetchHandler.fetch(0)).thenReturn(Tasks.forResult(realtimeFetchedContainerResponse));
```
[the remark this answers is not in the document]
qdpham13: Added this check

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: Yeah essentially. configAutoFetch.listenForNotifications is a blocking call as it listens for messages through the stream. As soon as it moves on from that method it means the stream is closed and no messages are left so we'll try to reopen the http connection. retryHTTPConnection has checks to see if the app is in the right condition to retry(i.e no backoff, no current http connection. )

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: As of right now this is only being onError method is only being used when there is a backoff response from the backend.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -70,6 +71,9 @@ public class ConfigRealtimeHttpClient {
   @GuardedBy("this")
   private long httpRetrySeconds;
 
+  @GuardedBy("this")
+  private boolean isBackoffEnabled;
```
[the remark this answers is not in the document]
qdpham13: Updated to use the same naming convention as iOS.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
qdpham13: Updated.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
qdpham13: For our[ parsing](https://github.com/firebase/firebase-android-sdk/blob/6150f7b28cbeff3ba81d98d29bc6419b1a1cd43c/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java#L117), since it's a stream there will be multiple messages coming through our while loop so we can't just end it when we get one message. So what we do is add every incoming message to our string fullMessage string. When we see a message that contains '}' then we know the backend just finished sending a full message. We then call parseMessage which just strips the beginning and end of fullMessage up until it reaches '{' and '}'. Everything between '{' and. '}' should be properly formatted so we don't need to do anything there.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -77,6 +77,19 @@ private synchronized void executeAllListenerCallbacks() {
     }
   }
 
+  private String parseMessage(String message) {
```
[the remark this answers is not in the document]
qdpham13: For our[ parsing](https://github.com/firebase/firebase-android-sdk/blob/6150f7b28cbeff3ba81d98d29bc6419b1a1cd43c/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java#L117), since it's a stream there will be multiple messages coming through our while loop so we can't just end it when we get one message. So what we do is add every incoming message to our string fullMessage string. When we see a message that contains '}' then we know the backend just finished sending a full message. We then call parseMessage which just strips the beginning and end of fullMessage up until it reaches '{' and '}'. Everything between '{' and. '}' should be properly formatted so we don't need to do anything there.

danasilver (approved):
A few minor comments, but this LGTM. Thanks!

PR #3967, opened 2022-08-04
Add an extension point to ComponentRuntime.

vkryachko:
It allows augmenting/replacing components as they are requested from
ComponentRegistrars.

The motivation is to be able to measure initialization time of
certain components and at the same time not to introduce this
measurement logic directly to the components framework.

### DIFF
--- a/firebase-components/src/main/java/com/google/firebase/components/Component.java
+++ b/firebase-components/src/main/java/com/google/firebase/components/Component.java
@@ -15,6 +15,8 @@
 package com.google.firebase.components;
 
 import androidx.annotation.IntDef;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


firebase-components/src/main/java/com/google/firebase/components/Component.java
```diff
@@ -153,6 +168,12 @@ public boolean isValue() {
     return type == ComponentType.VALUE;
   }
 
+  /** Creates a copy of the component with {@link ComponentFactory} replaced. */
+  public Component<T> withFactory(ComponentFactory<T> factory) {
+    return new Component<>(
+        name, providedInterfaces, dependencies, instantiation, type, factory, publishedEvents);
+  }
+
```
yifanyang, line 176: Is this for testing only? Looks like the purpose of `ComponentFactory` is only for product teams to implement. Once a product defined its factory in its registrars (and supply it via the component builder), it never needs to change?

yifanyang (approved):

emilypgoogle (approved):

firebase-components/src/main/java/com/google/firebase/components/Component.java
vkryachko: It's intended to be used outside of tests, for existing use example see: https://github.com/firebase/firebase-android-sdk/blob/da0f5937f964cfb2312f16b408285c44ae05840c/firebase-common/src/main/java/com/google/firebase/monitoring/ComponentMonitoring.java#L80-L84

PR #3968, opened 2022-08-04
Remove OkHttp since it is not used.

tom-andersen:


tom-andersen:
Branches diverged ... remove PR.

PR #3971, opened 2022-08-05
Add info text argument to startFeedback

lfkellogg:
Based on [go/iaf-sdk-api-proposal](http://go/iaf-sdk-api-proposal)

Tested in test app:
image

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
@@ -16,9 +16,11 @@ package com.google.firebase.appdistribution.impl;
 
 import android.graphics.Bitmap;
 import android.os.Bundle;
+import android.text.method.LinkMovementMethod;
 import android.view.View;
 import android.widget.EditText;
 import android.widget.ImageView;
+import android.widget.TextView;
 import android.widget.Toast;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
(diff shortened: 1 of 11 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -33,17 +35,21 @@ public class FeedbackActivity extends AppCompatActivity {
...
   public static final String RELEASE_NAME_EXTRA_KEY =
       "com.google.firebase.appdistribution.FeedbackActivity.RELEASE_NAME";
+  public static final String INFO_TEXT_EXTRA_KEY =
+      "com.google.firebase.appdistribution.FeedbackActivity.INFO_TEXT";
   public static final String SCREENSHOT_FILENAME_EXTRA_KEY =
       "com.google.firebase.appdistribution.FeedbackActivity.SCREENSHOT_FILE_NAME";
 
   private FeedbackSender feedbackSender;
   private String releaseName;
+  private CharSequence infoText;
```
jladieu, line 45: Just curious, how do you choose `CharSequence` instead of a plain ole' `String`?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -306,12 +310,17 @@ private UpdateTask updateApp(boolean showDownloadInNotificationManager) {
   }
 
   @Override
-  public void startFeedback() {
-    startFeedback(Executors.newSingleThreadExecutor());
+  public void startFeedback(int infoTextResourceId) {
+    startFeedbackWithInfoText(firebaseApp.getApplicationContext().getText(infoTextResourceId));
```
jladieu, line 314: Can you chain to `startFeedback(String)` instead of directly to `startFeedbackWithInfoText`, so the implementations are less likely to diverge?

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
```diff
@@ -4,6 +4,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    android:padding="24dp"
```
jladieu, line 7: Is this just generated XML changes?  I kind of just skimmed this.

jladieu (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
lfkellogg: Because that's what `Context.getText()` returns: https://developer.android.com/reference/android/content/Context#getText(int)

`Context.getString()` returns a String but it does not retain any html/styles, which we need.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
lfkellogg: Actually I'm realizing our API should have `startFeedback(CharSequence)`, so yes.

I was trying to model our API after [TextView.setText](https://developer.android.com/reference/android/widget/TextView#setText(java.lang.CharSequence)), which accepts an int or a CharSequence. I thought it was int or String.

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
lfkellogg: These add the info text to our proof-of-concept UX.

PR #3973, opened 2022-08-08
Remove Firebase Predictions from the Firebase installation ID delete …

jenh:
…comment as it was recently shut down; replace with Firebase A/B Testing, which also requires a FID.

### DIFF
--- a/firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
+++ b/firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
@@ -48,9 +48,9 @@ public interface FirebaseInstallationsApi {
   Task<InstallationTokenResult> getToken(boolean forceRefresh);
 
   /**
-   * Async function that deletes this Firebase app installation from Firebase backend. This call
-   * may cause Firebase Cloud Messaging, Firebase Remote Config, Firebase A/B Testing, or
-   * Firebase In-App Messaging to not function properly.
+   * Async function that deletes this Firebase app installation from Firebase backend. This call may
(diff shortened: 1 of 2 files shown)


danasilver (approved):
🚀

jenh:
Great catch--thank you, Rosario! I've updated that file and added.

thatfiredev (approved):

firebase-installations-interop/src/main/java/com/google/firebase/installations/FirebaseInstallationsApi.java
```diff
@@ -49,8 +49,8 @@ public interface FirebaseInstallationsApi {
 
   /**
    * Async function that deletes this Firebase app installation from Firebase backend. This call
-   * would possibly lead Firebase Notification, Firebase RemoteConfig, Firebase Predictions or
-   * Firebase In-App Messaging not function properly.
+   * may cause Firebase Cloud Messaging, Firebase Remote Config, Firebase A/B Testing, or
+   * Firebase In-App Messaging to not function properly.
```
thatfiredev, line 53: The `check-changed` CI check is failing due to the formatting here. Autoformat suggested using this instead:
```
   * Async function that deletes this Firebase app installation from Firebase backend. This call may
   * cause Firebase Cloud Messaging, Firebase Remote Config, Firebase A/B Testing, or Firebase
   * In-App Messaging to not function properly.
```

Notice that the word "may" moved from the second line to the first and "Firebase" moved from the third line to the second.

jenh: Thanks a bunch, that did the trick!

PR #3978, opened 2022-08-09
Enable oss master to private main branch sync on GitHub Actions

yifanyang:


.github/workflows/private-mirror-sync.yml
```diff
@@ -0,0 +1,22 @@
+name: Private Mirror Sync
```
vkryachko, line 1: let's make sure it only runs on firebase/firebase-android-sdk repo.

yifanyang: Done.

vkryachko (approved):

.github/workflows/private-mirror-sync.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: true
+          token: ${{ secrets.GOOGLE_OSS_BOT_TOKEN }}
       - name: Force push HEAD to private repo main branch
         run: |
           git remote add mirror git@github.com:FirebasePrivate/firebase-android-sdk.git
```
vkryachko, line 23: how will `git push` know to use the right credentials?

yifanyang: I just realized that I haven't actually tested it before sending it out for review. But it should work now IIUC.

> how will `git push` know to use the right credentials?

The `actions/checkout` [usage](https://github.com/actions/checkout#usage) section describes one flag `persist-credentials` which defaults to `true`. I think this flag makes following git invocations to use the above configured token.

PR #3984, opened 2022-08-11
Commit code style

martinbonnin:
This is so that Android Studio doesn't auto-fold imports. We might want to add more configuration there.

See https://github.com/firebase/firebase-android-sdk/pull/1252#discussion_r943456561

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,7 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JetCodeStyleSettings>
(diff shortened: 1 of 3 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JetCodeStyleSettings>
+      <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="100" />
```
martinbonnin, line 4: 100 is the biggest value allowed by the settings UI.

martinbonnin:
2025 cleanup

PR #3988, opened 2022-08-12
Fix sync job between OSS and private repository

yifanyang:
The current setup of the workflow is not working:

https://github.com/firebase/firebase-android-sdk/actions/workflows/private-mirror-sync.yml

There are two problems:
- `actions/checkout` checks out a shallow copy of the code, which is not push-able
- the token supplied to `actions/checkout` can only be used for authentication with http but not with ssh

### DIFF
--- a/.github/workflows/private-mirror-sync.yml
+++ b/.github/workflows/private-mirror-sync.yml
@@ -16,9 +16,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          fetch-depth: 0
           submodules: true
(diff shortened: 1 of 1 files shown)


vkryachko (commented):
@paulb777 does this look reasonable? is this similar to how you sync the private repo on ios?

paulb777:
This looks good to me.  The iOS repo uses a Jenkins or Kokoro job that was set up years ago. GHA would be more sustainable. cc: @morganchen12

vkryachko (approved):

PR #3999, opened 2022-08-16
Use kotlin script for buildSrc gradle file.

vkryachko:


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,109 @@
...
+        create("firebaseVendorPlugin") {
+            id = "firebase-vendor"
+            implementationClass = "com.google.firebase.gradle.plugins.VendorPlugin"
+        }
+    }
+})
+
+tasks.withType<Test> {
+    testLogging {
+        // Make sure output from
```
rlazo, line 102: not need to break this comment into so many lines, maybe?

rlazo (approved):

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+    implementation("org.jetbrains.dokka:dokka-android-gradle-plugin:0.9.17-g013-9b8280a")
+    implementation("com.android.tools.build:gradle:3.4.3")
+    implementation("gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.9")
+
+    testImplementation("junit:junit:4.13.2")
+    testImplementation("com.google.truth:truth:1.1.2")
+    testImplementation("commons-io:commons-io:2.6")
+}
+
+gradlePlugin(Action {
```
daymxn, line 69: Theres a gradle extension method on projects that has the action block as the last argument (actually, it's only argument), so Action shouldn't be needed here. That is, this should work:
```
gradlePlugin {
...
}
```

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+dependencies {
+    // Firebase performance plugin, it should be added here because of how gradle dependency
+    // resolution works, otherwise it breaks Fireperf Test Apps.
+    // See https://github.com/gradle/gradle/issues/12286
+    implementation("com.google.firebase:perf-plugin:$perfPluginVersion")
+
+    implementation("com.google.auto.value:auto-value-annotations:1.8.1")
+    annotationProcessor("com.google.auto.value:auto-value:1.6.5")
+    implementation("digital.wup:android-maven-publish:3.6.3")
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.4.32")
```
daymxn, line 47: Since you're using the `kotlin-dsl`, this can actually be reduced to such:
```kotlin
implementation(kotlin("gradle-plugin", "1.4.32"))
```

But thats a design choice, especially since it differs from everything else.

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+    implementation("gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.9")
+
+    testImplementation("junit:junit:4.13.2")
+    testImplementation("com.google.truth:truth:1.1.2")
+    testImplementation("commons-io:commons-io:2.6")
+}
+
+gradlePlugin(Action {
+    plugins {
+        create("licensePlugin") {
```
daymxn, line 71: It doesn't matter too much, but you can replace these `create` calls with `register` calls to fall in line with configuration avoidance. I'm curious if this being defined in buildSrc changes anything regarding configuration avoidance (that is, makes it pointless).

The pro would be that plugins not used wouldn't be created eagerly during builds. Plugins would only be created when they're used and needed.

buildSrc/build.gradle.kts
vkryachko: While it works, studio fails to type check it, let's leave it as is for now

daymxn: > While it works, studio fails to type check it, let's leave it as is for now

Yeah its just a nit, so either way works. Although, I think that may be an issue with your AS cache- because mine works perfectly fine. Or maybe its a bug within AS, I'm currently on `2021.3.1 Beta 4`

daymxn (approved):

buildSrc/build.gradle.kts
vkryachko: hm, it used to not work with AGP 7.2.2, but looks like it's not an issue anymore, fixed

PR #3974, opened 2022-08-09
Upgrade AGP to 7.2.2

vkryachko:

PR #4002, opened 2022-08-17
Use the more common stdlib dep now that we ship java8 bytecode.

vkryachko:
Additionally fix pom dependency type resolution.

### DIFF
--- a/appcheck/firebase-appcheck/ktx/ktx.gradle
+++ b/appcheck/firebase-appcheck/ktx/ktx.gradle
@@ -51,7 +51,7 @@ android {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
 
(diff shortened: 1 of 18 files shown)


rlazo (approved):

vkryachko:
> Starting in Kotlin 1.4, the stdlib dependency is [added by default](https://kotlinlang.org/docs/whatsnew14.html#dependency-on-the-standard-library-added-by-default). I wonder if we have tried removing the dependency from our gradle files?

Yeah I tried that, but it does not work, likely because we have this in our code https://github.com/firebase/firebase-android-sdk/blob/75ea92c7cce64809afa9cc724c10ef1a821dd21a/firebase-common/firebase-common.gradle#L78

I think it confuses the plugin and the dep is not added

thatfiredev (approved):

vkryachko:
@thatfiredev unrelated to this change, looks like appcheck:ktx has a failing test due to a missing KtxRegistrar, can you please take a look?

thatfiredev:
@vkryachko I was hoping that'd be fixed in #3974, but I can open a separate PR to add the Registrar before that one gets merged.

PR #4003, opened 2022-08-17
add FirebaseAppCheckKtxRegistrar to appcheck-ktx

thatfiredev:
This PR should add the missing KtxRegistrar to appcheck-ktx

### DIFF
--- a/appcheck/firebase-appcheck/ktx/src/main/kotlin/com/google/firebase/appcheck/ktx/FirebaseAppCheck.kt
+++ b/appcheck/firebase-appcheck/ktx/src/main/kotlin/com/google/firebase/appcheck/ktx/FirebaseAppCheck.kt
@@ -39,3 +39,12 @@ operator fun AppCheckToken.component1() = token
  * @return the expireTimeMillis of the [AppCheckToken]
  */
 operator fun AppCheckToken.component2() = expireTimeMillis
+
+internal const val LIBRARY_NAME: String = "fire-app-check-ktx"
+
+/** @suppress */
+@Keep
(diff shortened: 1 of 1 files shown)


appcheck/firebase-appcheck/ktx/src/main/kotlin/com/google/firebase/appcheck/ktx/FirebaseAppCheck.kt
```diff
@@ -39,3 +44,12 @@ operator fun AppCheckToken.component1() = token
  * @return the expireTimeMillis of the [AppCheckToken]
  */
 operator fun AppCheckToken.component2() = expireTimeMillis
+
+internal const val LIBRARY_NAME: String = "fire-app-check-ktx"
+
+/** @suppress */
+@Keep
```
vkryachko, line 51: nit: no need for this as keeps propagate from components
https://github.com/firebase/firebase-android-sdk/blob/eb4d215b5d8b7677cdc9e0d6dedf72025f4ad085/firebase-components/proguard.txt#L4-L5

vkryachko (approved):
Thanks for the fix 👍 

One minor nit below

appcheck/firebase-appcheck/ktx/src/main/kotlin/com/google/firebase/appcheck/ktx/FirebaseAppCheck.kt
thatfiredev: Removed

PR #4012, opened 2022-08-19
database-ktx: add callbackFlow for eventlisteners

thatfiredev:
Part of [go/kotlin-flows-firebase-android](http://go/kotlin-flows-firebase-android)

This PR should add callbackFlows for RTDB's `ValueEventListener` and `ChildEventListener`.


firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
```diff
@@ -59,6 +66,56 @@ inline fun <reified T> MutableData.getValue(): T? {
...
+/**
+ * Starts listening to this query and emits its values via a [Flow].
+ *
+ * - When the returned flow starts being collected, a [ValueEventListener] will be attached.
+ * - When the flow completes, the listener will be removed.
+ */
+fun Query.snapshots() = callbackFlow<DataSnapshot> {
+    val listener = addValueEventListener(object : ValueEventListener {
+        override fun onDataChange(snapshot: DataSnapshot) {
+            trySend(snapshot)
```
rlazo, line 78: For https://github.com/firebase/firebase-android-sdk/commit/72354736fdf0318ed4968ddff09d0bfea2f398e2 we went with `trySendBlocking` is there any difference between cases?

thatfiredev: When using `trySendBlocking` Android Studio shows me this warning:

Screenshot 2022-08-19 at 16 26 25

rlazo (approved):

firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
vkryachko: 1. I think it is a false positive in the linter.
2. That said,  it looks like the events are dispatched on the UI thread, so it's not ok to directly `trySendBlocking`:
https://github.com/firebase/firebase-android-sdk/blob/58c167795980b296c737ffce044c29c7ff1926d3/firebase-database/src/main/java/com/google/firebase/database/android/AndroidEventTarget.java#L25

imo instead we can do the same thing we did for firestore, something like
```suggestion
            someExecutor.execute(() -> trySendBlocking(snapshot));
```

thatfiredev: @vkryachko Looks like firestore's background executor uses a ThreadPool with 4 threads:

https://github.com/firebase/firebase-android-sdk/blob/d50a3d1102bcd9aedf84e5bf6929161f0b40bbc2/firebase-firestore/src/main/java/com/google/firebase/firestore/util/Executors.java#L42-L44

https://github.com/firebase/firebase-android-sdk/blob/d50a3d1102bcd9aedf84e5bf6929161f0b40bbc2/firebase-firestore/src/main/java/com/google/firebase/firestore/util/Executors.java#L30

Should we also use a ThreadPool for RTDB?

```suggestion
            Executors.newFixedThreadPool(nThreads = 4).execute {
                trySendBlocking(ChildEvent.Added(snapshot, previousChildName))
            }
```

thatfiredev: @vkryachko You were right, it was a false-positive in the linter. After the AGP7 update, the lint warning went away.

But my question on what executor we'll want to use remains.

vkryachko: Is there a BG executor we could use inside RTDB?

(The best one would be to use a `@BackgroundExecutor` from go/firebase-component-qualifiers, but it's not implemented yet :( )

thatfiredev: Seems like RTDB delegates events to one of these 3 methods:

https://github.com/firebase/firebase-android-sdk/blob/6a566dd17b9f255bfad60311f92230a2d92665b3/firebase-database/src/main/java/com/google/firebase/database/core/Repo.java#L302-L315

I wonder if we could use it to deliver Flow values

vkryachko: Not sure if that is safe, i.e. isn't in happening on the UI thread?

@maneesht what are your thoughts?

thatfiredev: It seems like `scheduleNow()` delegates to `DefaultRunLoop` which uses a `ScheduledThreadPoolExecutor `, so it doesn't block the UI Thread.

https://github.com/firebase/firebase-android-sdk/blob/6a566dd17b9f255bfad60311f92230a2d92665b3/firebase-database/src/main/java/com/google/firebase/database/core/utilities/DefaultRunLoop.java#L65-L69

maneesht: My only concern with this is that we would end up blocking other operations within the `Repo`'s actions. Maybe we should create a new executor solely for this? As RTDB only has one ATM

maneesht: I thought this through a bit more, and let's go ahead and just use the default run loop. If we feel like there's a performance gain, we can always change this in a future patch.

firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
```diff
@@ -59,6 +67,66 @@ inline fun <reified T> MutableData.getValue(): T? {
...
+/**
+ * Starts listening to this query and emits its values via a [Flow].
+ *
+ * - When the returned flow starts being collected, a [ValueEventListener] will be attached.
+ * - When the flow completes, the listener will be removed.
+ */
+fun Query.snapshots() = callbackFlow<DataSnapshot> {
+    val listener = addValueEventListener(object : ValueEventListener {
+        override fun onDataChange(snapshot: DataSnapshot) {
+            repo.scheduleNow {
```
vkryachko, line 79: hm, where is `repo` coming from? looks like it's `protected` in `Query`, I am surprised that it's accessible from here

vkryachko: oh, actually nvm, it looks like there is a getter defined, but still lint complains about `@RestrictTo(LIBRARY_GROUP)`, can you pls update `database/ktx.gradle` like it's done for firestore https://github.com/firebase/firebase-android-sdk/blob/523730a3b4463160490170d9348818a5c403a2eb/firebase-firestore/ktx/ktx.gradle#L20 ?

vkryachko (approved):

firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
thatfiredev: @vkryachko  Done

firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt
thatfiredev: Thanks for looking into it @maneesht !

thatfiredev:
@vkryachko I changed the Flows to be properties instead of functions. Please take a second look

vkryachko (approved):

PR #4015, opened 2022-08-22
Expose Multi-DB, and add tests.

tom-andersen:


firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -104,22 +104,31 @@ public interface InstanceRegistry {
...
     FirebaseApp app = FirebaseApp.getInstance();
     if (app == null) {
       throw new IllegalStateException("You must call FirebaseApp.initializeApp first.");
     }
-    return getInstance(app, DatabaseId.DEFAULT_DATABASE_ID);
+    return app;
+  }
+
+  @NonNull
+  public static FirebaseFirestore getInstance() {
```
wu-hui, line 116: Hmm, it looks like we never had public doc for `getInstance` for Java, can you add them now since it gets slightly more complex?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
```diff
@@ -1131,6 +1135,87 @@ public void testAppDeleteLeadsToFirestoreTerminate() {
...
+    assertNotSame(db1, db3);
+    assertNotSame(db2, db3);
+  }
+
+  @Test
+  public void testNamedDbHaveDifferentPersistence()
+      throws ExecutionException, InterruptedException, TimeoutException {
+    // TODO: Have backend with named databases created beforehand.
+    // Emulator doesn't care if database was created beforehand.
+    if (!isRunningAgainstEmulator()) {
```
wu-hui, line 1176: Use `org.junit.Assume.assumeTrue` will give you nicer report i think.

wu-hui (approved):
Some ultra-nits, LGTM

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
tom-andersen: Documentation added.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
tom-andersen: Done.

wu-hui (changes requested):
As discussed, please make the added interface package-private for now.

tom-andersen:
> As discussed, please make the added interface package-private for now.

Doing this broke other parts that call methods, so I am reverting change. I think we will just have to wait until we can make this public before we merge.

wu-hui (approved):

PR #4023, opened 2022-08-23
Add a new plugin for Dackka

daymxn:
Our current Dokka implementation is old and was due for an upgrade. Thankfully, the folks over at AndroidX have been working on new tooling that integrates directly with Dokka- and provides translations for DevSite. This PR does not include the removal of any old systems, only the addition of the new ones. The java variant of our ref doc generations will remain the same until we are comfortable with the DackkaPlugin.

This PR comes with:

- Various javadoc fixes
- Additional utility methods for buildSrc
- `DackkaGenerationTask` for creating ref docs with dackka
- `FiresiteTransformTask` for translating Devsite output from dackka to Firesite ready output
- `DackkaPlugin` for managing everything dackka related, and providing default configurations

All functionality provided in this PR is annotated, and follows standard practices with newer Gradle versions. There are minor differences that we can not implement until we upgrade our AGP version (per #3974), but I've left comments around them to revisit in the future.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,9 @@ buildscript {
         }
         maven {
             url 'https://storage.googleapis.com/android-ci/mvn/'
+            metadataSources {
+                artifact()
+            }
         }
 
     }
--- a/build.gradle
+++ b/build.gradle
@@ -77,6 +77,9 @@ configure(subprojects) {
         mavenCentral()
         maven {
             url 'https://storage.googleapis.com/android-ci/mvn/'
+            metadataSources {
+                artifact()
+            }
         }
     }
 
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
@@ -68,6 +68,7 @@ abstract class GenerateDocumentationTask @Inject constructor(
     }
 
     private fun constructArguments(): JSONObject {
+        // TODO(b/243675474): Move these to a task input for caching purposes
         val linksMap = mapOf(
             "android" to "https://developer.android.com/reference/kotlin/",
             "google" to "https://developer.android.com/reference/",
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
@@ -52,7 +52,7 @@ abstract class GenerateDocumentationTaskExtension : DefaultTask() {
 /**
  * Task to run Dackka on a project.
  *
- * Since dackka needs to be ran on the command line, we have to organize the arguments for dackka into
+ * Since dackka needs to be run on the command line, we have to organize the arguments for dackka into
  * a json file. We then pass that json file to dackka as an argument.
  *
  * @see GenerateDocumentationTaskExtension
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
@@ -1,0 +1,154 @@
+package com.google.firebase.gradle.plugins
+
+import java.io.File
+import javax.inject.Inject
+import org.gradle.api.DefaultTask
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.provider.SetProperty
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+import org.gradle.process.ExecOperations
+import org.gradle.workers.WorkAction
+import org.gradle.workers.WorkParameters
+import org.gradle.workers.WorkerExecutor
+import org.json.JSONObject
+
+/**
+ * Extension class for [GenerateDocumentationTask].
+ *
+ * Provides public configurations for the task.
+ *
+ * @property dackkaJarFile a [File] of the Dackka fat jar
+ * @property dependencies a list of all dependent jars (the classpath)
+ * @property kotlinSources a list of kotlin source roots
+ * @property javaSources a list of java source roots
+ * @property suppressedFiles a list of files to exclude from documentation
+ * @property outputDirectory where to store the generated files
+ */
+abstract class GenerateDocumentationTaskExtension : DefaultTask() {
+    @get:InputFile
+    abstract val dackkaJarFile: Property<File>
+
+    @get:Input
+    abstract val dependencies: ListProperty<File>
+
+    @get:InputFiles
+    abstract val kotlinSources: ListProperty<File>
+
+    @get:InputFiles
+    abstract val javaSources: ListProperty<File>
+
+    @get:InputFiles
+    abstract val suppressedFiles: ListProperty<File>
+
+    @get:OutputDirectory
+    abstract val outputDirectory: Property<File>
+}
+
+/**
+ * Task to run Dackka on a project.
+ *
+ * Since dackka needs to be ran on the command line, we have to organize the arguments for dackka into
+ * a json file. We then pass that json file to dackka as an argument.
+ *
+ * @see GenerateDocumentationTaskExtension
+ */
+abstract class GenerateDocumentationTask @Inject constructor(
+    private val workerExecutor: WorkerExecutor
+) : GenerateDocumentationTaskExtension() {
+
+    @TaskAction
+    fun build() {
+        val configFile = saveToJsonFile(constructArguments())
+        launchDackka(configFile, workerExecutor)
+    }
+
+    private fun constructArguments(): JSONObject {
+        val linksMap = mapOf(
+            "android" to "https://developer.android.com/reference/kotlin/",
+            "google" to "https://developer.android.com/reference/",
+            "firebase" to "https://firebase.google.com/docs/reference/kotlin/",
+            "coroutines" to "https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/"
+        )
+
+        val jsonMap = mapOf(
+            "moduleName" to "",
+            "outputDir" to outputDirectory.get().path,
+            "globalLinks" to "",
+            "sourceSets" to listOf(mutableMapOf(
+                "sourceSetID" to mapOf(
+                    "scopeId" to "androidx",
+                    "sourceSetName" to "main"
+                ),
+                "sourceRoots" to kotlinSources.get().map { it.path } + javaSources.get().map { it.path },
+                "classpath" to dependencies.get().map { it.path },
+                "documentedVisibilities" to listOf("PUBLIC", "PROTECTED"),
+                "skipEmptyPackages" to "true",
+                "suppressedFiles" to suppressedFiles.get().map { it.path },
+                "externalDocumentationLinks" to linksMap.map { (name, url) -> mapOf(
+                    "url" to url,
+                    "packageListUrl" to "file://${project.rootDir.absolutePath}/kotlindoc/package-lists/$name/package-list"
+                ) }
+            )),
+            "offlineMode" to "true",
+            "noJdkLink" to "true"
+        )
+
+        return JSONObject(jsonMap)
+    }
+
+    private fun saveToJsonFile(jsonObject: JSONObject): File {
+        val outputFile = File.createTempFile("dackkaArgs", ".json")
+
+        outputFile.deleteOnExit()
+        outputFile.writeText(jsonObject.toString(2))
+
+        return outputFile
+    }
+
+    private fun launchDackka(argsFile: File, workerExecutor: WorkerExecutor) {
+        val workQueue = workerExecutor.noIsolation()
+
+        workQueue.submit(DackkaWorkAction::class.java) {
+            args.set(listOf(argsFile.path, "-loggingLevel", "WARN"))
+            classpath.set(setOf(dackkaJarFile.get()))
+            projectName.set(project.name)
+        }
+    }
+}
+
+/**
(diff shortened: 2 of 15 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
```diff
@@ -23,3 +29,45 @@ fun toBoolean(value: Any?): Boolean {
...
 }
+
+/**
+ * Shorthand for [BaseExtension.getBootClasspath]
+ *
+ * This method assumes your project has the [BaseExtension] extension,
+ * as it uses [ExtensionContainer.getByType]
+ */
+val Project.bootClasspath: List<File>
+    get() = extensions.getByType<BaseExtension>().bootClasspath
```
vkryachko, line 40: to be safe would it make sense to do something like the following?
```suggestion
    get() = extensions.findByType<BaseExtension>()?.bootClasspath ?: emptyList()
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
```diff
@@ -23,3 +29,45 @@ fun toBoolean(value: Any?): Boolean {
...
+        }
+
+/**
+ * Fetches the jars of dependencies associated with this configuration through an artifact view.
+ *
+ * Excluding androidx.annotation:annotation artifacts.
+ */
+fun Configuration.getJars() = incoming.artifactView {
+    attributes {
+        // replace value with android-class instead of jar after agp upgrade
```
vkryachko, line 67:
```suggestion
        // TODO(b/241795594): replace value with android-class instead of jar after agp upgrade
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
```diff
@@ -23,3 +29,45 @@ fun toBoolean(value: Any?): Boolean {
...
+        attribute(Attribute.of("artifactType", String::class.java), "jar")
+    }
+    componentFilter {
+        !it.displayName.startsWith("androidx.annotation:annotation:")
+    }
```
vkryachko, line 72: Do you know if this is actually needed? I can't recall why we filter it out.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
```diff
@@ -109,7 +109,7 @@ public void apply(Project project) {
                     .setFreeCompilerArgs(
                         ImmutableList.of("-module-name", kotlinModuleName(project))));
 
-    Dokka.configure(project, android, firebaseLibrary);
```
vkryachko, line 112: Is `Dokka.java` still needed or can it be removed?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -0,0 +1,162 @@
...
+        val dackkaFile = project.provider { project.dackkaConfig.singleFile }
+        val dackkaOutputDirectory = File(project.buildDir, "dackkaDocumentation")
+
+        dackkaJarFile.set(dackkaFile)
+        outputDirectory.set(dackkaOutputDirectory)
+    }
+
+    private fun registerFiresiteTransformTask(project: Project) =
+        project.tasks.register<FiresiteTransformTask>("firesiteTransform") {
+            mustRunAfter("generateDackkaDocumentation")
```
vkryachko, line 130: is must run after the right thing to do as opposed to connecting the task's output to this task's input?

i.e. 
```kotlin
val dackka = project.tasks.register<GenerateDocumentationTask>("generateDackkaDocumentation") {
  // ...
}
project.tasks.register<FiresiteTransformTask>("firesiteTransform") {
  dackkaFiles.set(dackka.flatMap { it. outputDirectory })
}
```

context: https://docs.gradle.org/current/userguide/lazy_configuration.html#working_with_task_dependencies_in_lazy_properties

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -0,0 +1,162 @@
...
+                    javaSources.set(sourcesForJava)
+                    kotlinSources.set(sourcesForKotlin)
+                    suppressedFiles.set(excludedFiles)
+
+                    applyCommonConfigurations()
+                }
+            }
+        }
+
+    // Remove when fixed: b/243534168
```
vkryachko, line 98: nit
```suggestion
    // TODO(b/243534168): Remove when fixed
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -0,0 +1,162 @@
...
+    private fun projectSpecificSources(project: Project) =
+        when (project.name) {
+            "firebase-common" -> {
+                project.project(":firebase-firestore").files("src/main/java/com/google/firebase").toList()
+            }
+            else -> emptyList()
+        }
+
+    // Remove when fixed: b/243534168
```
vkryachko, line 107: nit
```suggestion
    // TODO(b/243534168): Remove when fixed
```

build.gradle
```diff
@@ -77,6 +77,9 @@ configure(subprojects) {
         mavenCentral()
         maven {
             url 'https://storage.googleapis.com/android-ci/mvn/'
+            metadataSources {
```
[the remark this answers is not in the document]
daymxn: Tells gradle, that when looking in this specific repo, to look for artifacts exclusively. Typically, gradle looks for a pom file. It won't pick up the fatjar as such, so having gradle looks for artifacts specifically allows gradle to find it. It also still finds our other jars on the repo.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
```diff
@@ -0,0 +1,154 @@
...
+    abstract val suppressedFiles: ListProperty<File>
+
+    @get:OutputDirectory
+    abstract val outputDirectory: Property<File>
+}
+
+/**
+ * Task to run Dackka on a project.
+ *
+ * Since dackka needs to be ran on the command line, we have to organize the arguments for dackka into
```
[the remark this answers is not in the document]
daymxn: Its funny, cause I had that first- but changed it to `ran` as `run` didn't sound right. I do think `run` is correct though :')

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
vkryachko: Actually, I just tested your chance locally and it appears that this exclude causes nullability issues, i.e. with it all `@Nullable/@NotNull` annotations on java code are ignored and it cause all kotlin doc to have "platform types", i.e. `fun foo(bar: Bar!): Baz!` instead of `fun foo(bar: Bar?): Baz`. I think we want to remove this line(removing this locally fixed the nullability issue for me).

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt
```diff
@@ -0,0 +1,154 @@
...
+                    "sourceSetName" to "main"
+                ),
+                "sourceRoots" to kotlinSources.get().map { it.path } + javaSources.get().map { it.path },
+                "classpath" to dependencies.get().map { it.path },
+                "documentedVisibilities" to listOf("PUBLIC", "PROTECTED"),
+                "skipEmptyPackages" to "true",
+                "suppressedFiles" to suppressedFiles.get().map { it.path },
+                "externalDocumentationLinks" to linksMap.map { (name, url) -> mapOf(
+                    "url" to url,
+                    "packageListUrl" to "file://${project.rootDir.absolutePath}/kotlindoc/package-lists/$name/package-list"
```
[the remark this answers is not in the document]
daymxn: Thats a very good point, done: b/243675474

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FiresiteTransformTask.kt
```diff
@@ -0,0 +1,79 @@
...
+    // We don't actually upload class or index files,
+    // so these headers will throw not found errors if not removed.
+    private fun String.removeClassHeader() =
+        remove(Regex("- title: \"Class Index\"\n {2}path: \".+\"\n\n"))
+    private fun String.removeIndexHeader() =
+        remove(Regex("- title: \"Package Index\"\n {2}path: \".+\"\n\n"))
+
+    // We use a common book for all sdks, wheres dackka expects each sdk to have its own book.
+    private fun String.fixBookPath() =
+        remove(Regex("(?<=setvar book_path ?%})(.+)(?=/_book.yaml\\{% ?endsetvar)"))
```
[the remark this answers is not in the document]
daymxn: This doesn't remove the setvar, thats a look behind regex group. It removes the path from book.yaml links. For example, this:
`{% setvar book_path %}/reference/firestore/_book.yaml{% endsetvar %}`
becomes this:
`{% setvar book_path %}/_book.yaml{% endsetvar %}`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FiresiteTransformTask.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    // Our documentation lives under /docs/reference/ versus the expected /reference/
+    private fun String.fixHyperlinks() =
+        replace(Regex("(?<=href=\")(/)(?=reference/.*\\.html)"), "/docs/")
```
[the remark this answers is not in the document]
daymxn: For sure, done.
b/243674302
b/243674303
b/243674305
b/243673063

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
daymxn: No, I have no idea. Although- it seems to generate just fine. I'll remove it for now, and we can revisit this if we end up needing it removed again.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
daymxn: It can be removed, should I go ahead and remove it?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
daymxn: Thats a really good idea, I didn't even think about that. That also avoids hardcoded strings. +1

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
vkryachko: I think I remember now, in the old dokka having nullability annotations on the classpath would result in annotations literally propagating to the kotlindoc, i.e. it would render with `@NonNull`, so I removed it and it did not cause any issues with kotlin nullability, it looks like it's no longer the case, so yeah let's remove this

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
daymxn: Yeah, but it also could hide a potential bug. If the boot classpath does not exist, then theres an issue. It should always exist- even if its empty. Adding this here would mean it wouldn't throw an error when it doesn't exist- and you may not realize it. What are your thoughts?

vkryachko: That's a good point. otoh this method is a partial function, i.e. its receiver is too permissive, would it make sense to change it to something like `val LibraryExtension.bootClasspath`, this will make sure that any caller has an instance of LibraryExtension, as opposed to allowing to call it on projects that can be non-android projects. also it looks like the only place that is using this has access to a LibraryExtension, wdyt?

build.gradle
```diff
@@ -77,6 +77,9 @@ configure(subprojects) {
...
+                artifact()
+            }
```
vkryachko, line 82: I think you need to do the same in buildSrc tests to make tests pass:
https://github.com/firebase/firebase-android-sdk/blob/13083a2f991b81059ca3fe05764908b38c3d077b/buildSrc/src/test/kotlin/com/google/firebase/gradle/plugins/PublishingPluginTests.kt#L339-L341

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.java
vkryachko: I think we should, but a follow up PR might be better, since this one is already large

daymxn: Agreed 
b/243685887

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
daymxn: Actually, you may be on to something. I've changed it to just use the `LibraryExtension` directly instead.

build.gradle
daymxn: Hmm, lets find out :')

Also, should we add `--stacktrace` to the buildSrc action?

daymxn: Sure enough, that fixed it:)

vkryachko (approved):
👍

build.gradle
vkryachko: I don't think `--stacktrace` would be useful in this case, the problem was with tests failing and these failures can be seen in the build log.

PR #4026, opened 2022-08-24
Add documentation for DackkaPlugin

daymxn:
In accordance with [b/240169898](b/240169898), and more specifcally [b/240169649](b/240169649)- this PR adds a `README.md` and `DackkaPlugin.md` to buildSrc.

`README.md` just serves as a table of contents currently, and should be updated down the line once enough documentation has been created to justify otherwise.

`DackkaPlugin.md` walks through how Dackka and the DackkaPlugin work, as to help those onboarding understand it.

In the future, I'd like to add additional documentation in the repo itself.

_Note: I'm not sure if storing the documentation next to the plugin is the best idea. Maybe in something like `docs/plugins` at the root? I'm not completely sure._


buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
```diff
@@ -0,0 +1,99 @@
...
+an internal service called **Devsite**. Firebase hosts their documentation on a
+variant of Devsite called **Firesite**. You can click [here](https://firebase.google.com/docs/reference) 
+to see how that looks. Essentially, it's just Google's way of decorating (and organizing)
+documentation.
+
+Devsite expects its files to be in a very specific format. Previously, we would
+use an internal variant of metalava called doclava- which allowed us to provide
+sensible defaults as to how the Javadoc should be rendered. Then, we would do
+some further transformations to get the Javadoc output in-line with what
+Devsite expects. This was a lengthy process, and came with a lot of overhead.
```
vkryachko, line 18: Actually doclava and metalava are unrelated, [doclava](https://code.google.com/archive/p/doclava/) is a javadoc doclet that has been abandoned and is currently unmaintained, metalava is a reimplementation of a subset of what doclava's functionality that is unrelated to javadoc generation but rather api surface management.

So I would suggest changing this to something along the lines of
```suggestion
Devsite expects its files to be in a very specific format. Previously, we would
use a custom javadoc doclet doclava- which allowed us to provide
sensible defaults as to how the Javadoc should be rendered. Then, we would do
some further transformations to get the Javadoc output in-line with what
Devsite expects. However doclava does not support kotlindoc and has been unmaintained for many years.
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
```diff
@@ -0,0 +1,99 @@
...
+
+Dackka is an internal solution to that. Dackka provides a devsite plugin for
+Dokka that will handle the job of doclava. Not only does this mean we can cut
+out a huge portion of our transformation systems- but the overhead for maintaining
```
vkryachko, line 22: Which systems are you referring to? iiuc we will still have to run our source through metalava for example

vkryachko (commented):
Overall looks good, thanks for thoroughly documenting it. Left a few comments below.

A meta point: I wonder if it makes sense to move this documentation closer to the source of truth, namely to the kdoc of the DackkaPlugin. That will make it more easier to find and more likely that this documentation will stay up to date. wdyt?

daymxn:
> Overall looks good, thanks for thoroughly documenting it. Left a few comments below.
> 
> 
> 
> A meta point: I wonder if it makes sense to move this documentation closer to the source of truth, namely to the kdoc of the DackkaPlugin. That will make it more easier to find and more likely that this documentation will stay up to date. wdyt?

Hmmm that's a good point. Although the KDOC does already provide a moderate amount of explanation. This doc is supposed to be a bit more in-depth and longer. I think having it live in the kdoc would eat up a lot of lines for something you (typically) expect to be brief. Furthermore, the markdown aspect of it wouldn't be as readable as it is currently.

I do think we should move it someplace else though. Currently, it lives right next to the source- which is fine for now. But should we document more of the plugins, it could get out of control fairly quickly.

fwiw, we _could_ also add a note to the kdoc about this file? So if folks miss it, they know where to look. Thoughts?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
daymxn: We shouldn't have to, not with dackka at least. There's a diagram provided in the docs for the plug-in, but basically dackka handles all that metalava previously provided. Unless I'm missing something?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
daymxn: Hmm, im a bit confused then. Would you mind explaining that some more? What purpose does metalava provide in our doc generation? My understanding is it handles removing classes marked with hidden and provides stub variants of the source?

vkryachko: yeah that's right, that's what metalava does for us, additionally it produces api.txt files for us, this is what doclava does as part of javadoc generation as well.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
vkryachko: hm, does metalava remove hidden superclasses for example?

vkryachko:
> I think having it live in the kdoc would eat up a lot of lines for something you (typically) expect to be brief

I don't quite agree with that, I think it's usually brief since there is not much to say about it, but when there is it's ok and even good to make it detailed. i.e. look at `java.util.List` as one example.

That said I don't feel strongly, but slightly prefer having this documentation inline with the class.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
daymxn: Where do we use api.txt files again? Also, do we have any purpose for the stubs beyond javadoc generation?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
daymxn: define "remove". It doesnt generate documentation for the class, but its still referenced as a parent. It also seems to still generate a link for it.

daymxn: Although, it seems thats a bug. b/243324828

daymxn:
> > I think having it live in the kdoc would eat up a lot of lines for something you (typically) expect to be brief
> 
> I don't quite agree with that, I think it's usually brief since there is not much to say about it, but when there is it's ok and even good to make it detailed. i.e. look at `java.util.List` as one example.
> 
> That said I don't feel strongly, but slightly prefer having this documentation inline with the class.

Hmm, but isn't that because they generate javadocs for it? Because reading in the IDE is not the most pleasant. Or would you rather we generate javadocs for documentation instead of providing markdown files?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
vkryachko: > but its still referenced as a parent

This is not the behavior we get with metalava afaicr, ideally we should preserve current behavior, i.e.:

Source:
```java
/** @hide */ class Base {}
class Derived extends Base {}
```

Docstubs with metalava result:
```java
class Derived {}
```

One example of this that I found is [FirebaseInAppMessagingDisplay](https://firebase.devsite.corp.google.com/docs/reference/kotlin/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay?hl=en)

vkryachko:
> Because reading in the IDE is not the most pleasant

Depends on the IDE I guess, but in Android Studio/IntelliJ it renders nicely.

Screen Shot 2022-08-26 at 9 38 42 AM
y

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.md
daymxn: Ahh I see. Owen does say that that's the behavior they want, but its not current in place per that bug. For the time being, do you just wanna have dackka pull the sources from doc-stubs?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -67,7 +156,7 @@ abstract class DackkaPlugin : Plugin<Project> {
         )
     }
 
-    // TODO(b/243534168): Refactor when fixed, so we no longer need stubs
+    // TODO(b/243954517): Refactor when fixed, so we no longer need stubs
```
vkryachko, line 159: I think this bug just got fixed this morning, but afaicr there was another one related to hiding the hidden superclass, do you recall if we filed it?

vkryachko: I think this is the right one https://issuetracker.google.com/issues/243324828

daymxn: Yeah, the initial bug was [b/243324828](https://issuetracker.google.com/issues/243324828).
Another one was the one you filed (which you're probably thinking of), which is this: [b/243954517](https://b.corp.google.com/issues/243954517).

Both those bugs need to be fixed for us to remove stubs, so we still need the original bug to fixed. It seems like that might be sooner rather than later, shout out to the androidx team. Let me change the comment to reflect that.

vkryachko (approved):

PR #4027, opened 2022-08-25
[Fireperf] Fix `RateLimiter` by avoiding manually setting time

leotianlizhan:
Blocking https://github.com/firebase/firebase-android-sdk/pull/4024.
### Context and why this is bad
b/243686627
TL;DR: don't use `@VisibleForTesting` method for real code, they break encapsulation. In this case, `Timer(long time)` is breaking logic too because it's mixing 2 different timebases together (time-since-epoch and time-since-VM-start). It's a bad method that should not exist even for tests.
### Leftover problem
However if we remove the above, there's a leftover problem from https://github.com/firebase/firebase-android-sdk/pull/2662:
Before https://github.com/firebase/firebase-android-sdk/pull/2662, say there are 0 tokens at first. We can call `rateLimiter.check()` after any amount of time, so something like 1.33 tokens can be generated, but we only use `long` for our `tokenCount`, so it will automatically round to 1, and record current time as `lastTimeTokenReplenished`, then use that token so we are back down to 0. However if next time we call `check()`, 0.67 tokens are generated, then we round to 0, logic before https://github.com/firebase/firebase-android-sdk/pull/2662 would say we have no tokens to use, which is wrong because within the time described in above scenario, 2 tokens should have been generated, and we only used 2 tokens, so there should be enough. 

https://github.com/firebase/firebase-android-sdk/pull/2662 tried to fix this by manually setting `lastTimeTokenReplenished` to the exact time at which a whole-number of tokens are added. In the same example, if 1.33 tokens are generated, then we still round-down and only add 1 to `tokenCount`, but we advance `lastTimeTokenReplenished` by `the time it took to generate 1 token`. This is like time only increment in **ticks, with each tick guaranteed to be a multiple of the time it takes to generate a single token**. This works, but manually setting time is error-prone (this PR/bug is literal proof). 
### Solution
Simply change tokens to `double`. **Guava also uses double for their rate limiter's token counter** https://github.com/google/guava/blob/9c1e5dea4b980202ba003b90fcb64183d42031b3/android/guava/src/com/google/common/util/concurrent/SmoothRateLimiter.java#L313

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
@@ -264,7 +264,7 @@ final class RateLimiter {
     // Token bucket capacity, also the initial number of tokens in the bucket.
     private long capacity;
     // Number of tokens in the bucket.
-    private long tokenCount;
+    private double tokenCount;
 
     private Rate foregroundRate;
     private Rate backgroundRate;
@@ -298,22 +298,16 @@ final class RateLimiter {
      */
     synchronized boolean check(@NonNull PerfMetric metric) {
       Timer now = clock.getTime();
-      long newTokens =
-          Math.max(
-              0,
-              (long)
-                  (lastTimeTokenReplenished.getDurationMicros(now)
-                      * rate.getTokensPerSeconds()
-                      / MICROS_IN_A_SECOND));
-      tokenCount = Math.min(tokenCount + newTokens, capacity);
-      if (newTokens > 0) {
-        lastTimeTokenReplenished =
-            new Timer(
-                lastTimeTokenReplenished.getMicros()
-                    + (long) (newTokens * MICROS_IN_A_SECOND / rate.getTokensPerSeconds()));
+      double newTokens =
+          (lastTimeTokenReplenished.getDurationMicros(now)
+              * rate.getTokensPerSeconds()
+              / MICROS_IN_A_SECOND);
+      if (newTokens > 0.0) {
+        tokenCount = Math.min(tokenCount + newTokens, capacity);
+        lastTimeTokenReplenished = now;
       }
-      if (tokenCount > 0) {
-        tokenCount--;
+      if (tokenCount >= 1.0) {
+        tokenCount -= 1.0;
         return true;
       }
       if (isLogcatEnabled) {
--- a/firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
+++ b/firebase-perf/src/test/java/com/google/firebase/perf/transport/RateLimiterTest.java
@@ -93,34 +93,31 @@ public class RateLimiterTest extends FirebasePerformanceTestBase {
     RateLimiterImpl limiterImpl =
         new RateLimiterImpl(TWO_TOKENS_PER_MINUTE, 2, mClock, mockConfigResolver, NETWORK, false);
     PerfMetric metric = PerfMetric.getDefaultInstance();
-    // clock is 15 seconds, token count is 1.
+    // clock is 30 seconds, count is 2, afterwards is 1
+    currentTime = currentTime.plusSeconds(30);
+    assertThat(limiterImpl.check(metric)).isTrue();
+    // clock is 45 seconds, count is 1.5, afterwards is 0.5
     currentTime = currentTime.plusSeconds(15);
     assertThat(limiterImpl.check(metric)).isTrue();
-    // clock is 30 seconds, count is 1.
+    // clock is 60 seconds, count is 1, afterwards is 0
     currentTime = currentTime.plusSeconds(15);
     assertThat(limiterImpl.check(metric)).isTrue();
-    // clock is 45 seconds, count is 0.
-    currentTime = currentTime.plusSeconds(15);
-    assertThat(limiterImpl.check(metric)).isTrue();
-    // clock is 60 seconds, count is 1
-    currentTime = currentTime.plusSeconds(15);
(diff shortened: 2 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
```diff
@@ -298,22 +298,16 @@ static class RateLimiterImpl {
...
-                  (lastTimeTokenReplenished.getDurationMicros(now)
-                      * rate.getTokensPerSeconds()
-                      / MICROS_IN_A_SECOND));
-      tokenCount = Math.min(tokenCount + newTokens, capacity);
-      if (newTokens > 0) {
-        lastTimeTokenReplenished =
-            new Timer(
-                lastTimeTokenReplenished.getMicros()
-                    + (long) (newTokens * MICROS_IN_A_SECOND / rate.getTokensPerSeconds()));
+      double newTokens =
```
visumickey, line 301: A similar issue would exist in iOS as well. We might need to followup on that to fix it.

visumickey (approved):

jfcong (approved):
Thanks for fixing this!

firebase-perf/src/main/java/com/google/firebase/perf/transport/RateLimiter.java
leotianlizhan: Does iOS also have our own RateLimiter implementation?

leotianlizhan:
Only failing GHA is `Health Metrics / Coverage (push)`, asked Acore, this check is to compute metrics for calculating metrics differences like size and coverage, and it's not required because it runs for every product (not just fireperf), so other products might fail / be flaky. I checked the logs and it indeed failed due to other products, independent of us, so it is okay to merge. 
```
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: FAILURE: Build completed with 3 failures.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 1: Task failed with an exception.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: -----------
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: * What went wrong:
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: Execution failed for task ':firebase-messaging:testDebugUnitTest'.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: > There were failing tests. See the report at: file:///home/runner/work/firebase-android-sdk/firebase-android-sdk/firebase-messaging/build/reports/tests/testDebugUnitTest/index.html
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: * Try:
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: ==============================================================================
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 2: Task failed with an exception.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: -----------
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: * What went wrong:
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: Execution failed for task ':firebase-storage:testDebugUnitTest'.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: > There were failing tests. See the report at: file:///home/runner/work/firebase-android-sdk/firebase-android-sdk/firebase-storage/build/reports/tests/testDebugUnitTest/index.html
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: * Try:
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: ==============================================================================
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 3: Task failed with an exception.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: -----------
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: * What went wrong:
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: Execution failed for task ':firebase-messaging:testReleaseUnitTest'.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: > There were failing tests. See the report at: file:///home/runner/work/firebase-android-sdk/firebase-android-sdk/firebase-messaging/build/reports/tests/testReleaseUnitTest/index.html
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: 
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: * Try:
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
[I] 2022-08-25 02:32:14 +0000 UTC fireci.gradle: ==============================================================================
```

Issue #3774, opened 2022-06-01
Cannot mark as impressed more than one message almost same time


igorjacauna:
* Android Studio version: Arctic Fox 2020.3.1 Patch 4
  * Firebase Component: InAppMessage
  * Component version: 19.1.5

#### Steps to reproduce:

Have a custom Display component for InAppMessage, i.e.:

```java
FirebaseInAppMessaging.getInstance().setMessageDisplayComponent(new MyFirebaseMessaging());
```

#### Relevant Code:

My custom display component has a `HashMap` to handle multiple messages displaying on screen. Like 2 cards embedded on layout.

```java
private final HashMap<String, HashMap> InAppMessages = new HashMap<>();
```

Every time we receive a message we do something like:

```java
@Override
public void displayMessage(@NonNull InAppMessage inAppMessage, @NonNull FirebaseInAppMessagingDisplayCallbacks callbacks) {
    try {
        String campaignId = inAppMessage.getCampaignMetadata().getCampaignId();
    
        // Here we build a object that represents the received message
        HashMap<String, Object> inAppMessageObject = new HashMap<String, Object>();
        inAppMessageObject.put("message", inAppMessage);
        inAppMessageObject.put("callbacks", callbacks);
        
        // Then we put on HashMap this object to be called later
        this.InAppMessages.put(campaignId, inAppMessageObject);
    } catch (Exception e) {
        Log.e("FIAM: Error", e.toString());
    }
}
```

#### The problem

Some times, more than one event is triggered each one with a message assigned:

```java
FirebaseInAppMessaging.getInstance().triggerEvent("main_screen_opened");
FirebaseInAppMessaging.getInstance().triggerEvent("another_main_screen_opened");
``` 

But when we try to mark as impressed only one message are marked. We use something like this to handle multiple messages and mark each as impressed:

```java
public void markAsImpressionDetected(String campaignId) {
    try {
        // We recover from HashMap the message callbacks
        FirebaseInAppMessagingDisplayCallbacks callbacks = (FirebaseInAppMessagingDisplayCallbacks) this.InAppMessages.get(campaignId).get("callbacks");

        // And call 
        callbacks.impressionDetected();
    } catch(Exception e) {
        Log.e("FIAM: Error", e.toString());
    }
}
```

With milliseconds between each call. But always only one of 2 or more messages are marked as impressed. And no error was throw

argzdev:
[...]

PR #4028, opened 2022-08-25
Fix: Cannot mark as impressed more than one message almost same time

eldhosembabu:
https://github.com/firebase/firebase-android-sdk/issues/3774

The Flag variable to check whether impressed or not was a static one and because of which when one FIAM message is marked as impressed, it was affecting other messages until the app restarts.

### DIFF
--- a/firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/DisplayCallbacksImpl.java
+++ b/firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/DisplayCallbacksImpl.java
@@ -73,7 +73,7 @@ public class DisplayCallbacksImpl implements FirebaseInAppMessagingDisplayCallba
   }
 
   @VisibleForTesting
-  boolean hasImpressed(){
+  boolean hasImpressed() {
     return wasImpressed;
   }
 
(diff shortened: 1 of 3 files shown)


vkryachko:
@eldhosembabu can you add tests for this change?

eldhosembabu:
> @eldhosembabu can you add tests for this change?

Done

firebase-inappmessaging/src/main/java/com/google/firebase/inappmessaging/internal/DisplayCallbacksImpl.java
```diff
@@ -72,6 +72,11 @@ public class DisplayCallbacksImpl implements FirebaseInAppMessagingDisplayCallba
     wasImpressed = false;
   }
 
+  @VisibleForTesting
+  boolean hasImpressed() {
+    return wasImpressed;
```
vkryachko, line 77: curious why the method is called `hasImpressed`, while it accesses the variable called `wasImpressed`.

eldhosembabu: Changed the method name to wasImpressed. This is introduced just for writing unit test.

vkryachko (approved):

PR #4030, opened 2022-08-25
Make buildSrc and fireci tests run conditionally

yifanyang:
* Added a custom action for detecting whether certain path patterns are changed in a pull request
* Configure buildSrc tests and fireci tests to run only when their source code or workflow file are modified


yifanyang:
@vkryachko I am not able to make #3863 work.

Per [GitHub doc](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks), I shall create a dummy workflow. However, I couldn't achieve the goal of making it run only when its counterpart workflow would not run, via `paths` and `paths-ignore`.

It seems the approach suggested in the doc would sometimes result in having two identical check posted to the pull request, as in the current list of checks in #3863 and https://github.com/yifanyang/github-actions-test/pull/10.

vkryachko:
hm, I am not sure that the amount of added complexity is worth it, how about we leave the affected checks optional for now?

yifanyang:
@vkryachko 

> hm, I am not sure that the amount of added complexity is worth it, how about we leave the affected checks optional for now?

Or shall we just make buildSrc and fireci tests always run, and make them required. They don't take long to finish anyway.

vkryachko:
I slightly prefer making them not required, it looks like a pattern the ios repo already uses

yifanyang:
> I slightly prefer making them not required, it looks like a pattern the ios repo already uses

Sounds good, then I'll stop working on this for now.

PR #4031, opened 2022-08-25
Remove Dokka Plugin

daymxn:
Per [b/243685887](b/243685887), now that Dackka has been implemented- we can safely remove the old Dokka plugin.

We also don't need to register the DackkaPlugin, as it automatically gets applied via `FirebaseLibrary`. Unless we end up having a use case where consumers want/or need to manually apply the plugin- exposing it this way could open the door to misunderstandings & bugs. This PR removes it to be safe.


buildSrc/build.gradle.kts
```diff
@@ -72,10 +72,6 @@ gradlePlugin {
             id = "LicenseResolverPlugin"
             implementationClass = "com.google.firebase.gradle.plugins.license.LicenseResolverPlugin"
         }
-        register("dackkaPlugin") {
-            id = "DackkaPlugin"
-            implementationClass = "com.google.firebase.gradle.plugins.DackkaPlugin"
-        }
```
vkryachko, line 78: Hm, how come this plugin is being removed?

daymxn: I mentioned it in the pr description, but basically it serves no purpose. I didn't realize you don't need to actually register plugins to use them. This way, people don't accidentally apply the plugin- and potentially cause issues down the line cause of a misunderstanding on how the plugin works. We can always add it back if need be fwiw

vkryachko:
please check buildSrc failures, look legit

buildSrc/build.gradle.kts
```diff
@@ -72,10 +72,6 @@ gradlePlugin {
             id = "LicenseResolverPlugin"
             implementationClass = "com.google.firebase.gradle.plugins.license.LicenseResolverPlugin"
         }
-        register("dackkaPlugin") {
```
vkryachko, line 75: Please also remove the following above:

```
implementation("org.jetbrains.dokka:dokka-android-gradle-plugin:0.9.17-g013-9b8280a")
```

vkryachko (approved):

PR #4038, opened 2022-08-29
Upgrade count upto to use int64

wu-hui:


firebase-firestore/src/proto/google/firestore/v1/query.proto
```diff
@@ -313,7 +313,7 @@ message StructuredAggregationQuery {
       // Requires:
       //
       // * Must be greater than zero when present.
-      int32 up_to = 1;
+      int64 up_to = 1;
```
dconeybe, line 316: This should be `google.protobuf.Int64Value`, no? Or is that the same thing?

https://github.com/googleapis/googleapis/blob/260b98fa50c7e16634262b4c0ade2429cfb18f20/google/firestore/v1/query.proto#L323

dconeybe: It really looks like `int64` and `google.protobuf.Int64Value` are distinct types.

https://github.com/protocolbuffers/protobuf/blob/91cda68a22c999f37aa4671896e6a314bac5881d/src/google/protobuf/wrappers.proto#L69-L75

wu-hui: Fixed.

dconeybe (approved):

PR #4043, opened 2022-08-30
Clean up unused methods.

gsakakihara:
* Cleaned up some unused methods in CommonNotificationBuilder.
* Also changed to explicitly save the result of sendIntent() in FcmBroadcastProcessor but not use it since we're blocking on the result of the Task, and added a comment to that effect.


James201311 (approved):
LGTM!

PR #4044, opened 2022-08-30
add coroutines-play-services as a transitive dep to firebase-common-ktx

thatfiredev:
_(Googlers see [go/suspend-ktx-firebase-android](http://go/suspend-ktx-firebase-android))_


thatfiredev:
I'm slightly concerned with the increased size shown in the bot report above. Seems like adding the library as a transitive dependency adds **~390kB** to every ktx module - to me it is a big bump for a small addition.
I'll check if we can expose only the `.await()` function instead of the whole library to see if it makes any difference in terms of  size.

thatfiredev:
I'm holding this until #3974 gets merged. I think it would be better if we release with the latest version of Kotlin + latest version of kotlinx-coroutines-play-services .

firebase-common/ktx/src/test/kotlin/com/google/firebase/ktx/Tests.kt
```diff
@@ -101,3 +107,36 @@ class KtxTests {
...
+        try {
+            task.await()
+            fail("Task#await should throw an Exception")
+        } catch (e: Exception) {
+            assertThat(e).isInstanceOf(TestException::class.java)
+            assertThat(task.isSuccessful).isFalse()
+        }
+    }
+
+    // TODO(thatfiredev): add a test for CancellationToken once we support Coroutines >= 1.6
```
vkryachko, line 141: curious: what's blocking us from supporting that?

thatfiredev: @vkryachko you were a few minutes ahead of me, hehe.
Using Coroutines 1.6 was throwing errors during compilation because we were using an older version of Kotlin. Now that we're using Kotlin 1.7 it works fine.

thatfiredev: But it seems like none of our APIs support passing a CancellationToken, so I'm not adding a test for that.

vkryachko: yeah, my hope is that tasks become cancellable directly, that way we could get proper cancellation without changes required in all of our API surfaces

vkryachko (approved):

thatfiredev:
@vkryachko Can I merge this?

Also since it bumps firebase-common, does it mean we'll have to release all SDKs? Or only the ones that have methods returning Tasks?
But since it's a fireconf launch, I guess releasing them all wouldn't be so bad.

vkryachko:
Sure, go ahead and merge. Yeah, we will have to release everyone with tasks

thatfiredev:
@vkryachko will do. Thanks!

thatfiredev:
@vkryachko GitHub won't let me merge this, it says "Required statuses must pass before merging", I'm not sure why.

thatfiredev:
@vkryachko I checked the logs and noticed that the failing check was the flaky test which got fixed recently. I just had to sync with master, thanks for pointing me in the right direction.

PR #4051, opened 2022-09-01
Realtime RC Store Template Version

qdpham13:
Save template version to disk in metadata and use variable for templateVersion to make it more similar to iOS (https://github.com/firebase/firebase-ios-sdk/blob/ccef4605c4499cc90646395328874f4e159075fb/FirebaseRemoteConfig/Sources/RCNConfigFetch.m#L78). 
This change to save templateVersion was pushed into iOS recently: https://github.com/firebase/firebase-ios-sdk/pull/10162


qdpham13:
/test smoke-tests
--

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -117,6 +118,7 @@ public ConfigFetchHandler(
     this.frcBackendApiClient = frcBackendApiClient;
     this.frcMetadata = frcMetadata;
     this.customHttpHeaders = customHttpHeaders;
+    this.lastTemplateVersion = getTemplateVersionFromDisk();
```
danasilver, line 121: I'd avoid doing work (esp reading from disk) in the constructor - if it's required we could pass it in. It looks like the instance variable is used to get the latest fetched version in ConfigAutoFetch (rather than being required in this class), though. Does it need to be initialized from a saved value in any event?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -319,6 +321,12 @@ private FetchResponse fetchFromBackend(
               getFirstOpenTime(),
               currentTime);
 
+      if (response.getFetchedConfigs() != null) {
+        // Set template version in memory.
```
danasilver, line 325: nit: prob don't need to say memory to refer to the instance variable. that threw me off at first

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -530,11 +538,14 @@ private Long getFirstOpenTime() {
...
+    // Get templateVersion on disk from metadata.
+    return frcMetadata.getLastTemplateVersion();
+  }
+
   public long getTemplateVersionNumber() {
-    if (fetchedConfigsCache.get() != null && fetchedConfigsCache.get().getResult() != null) {
-      return fetchedConfigsCache.get().getResult().getTemplateVersionNumber();
-    }
-    return 1L;
+    return lastTemplateVersion;
```
danasilver, line 548: If the goal is to provide the last fetched template version, I would encapsulate it in `FetchResponse` with the other fetched data like `lastFetchETag`

qdpham13: None of the fetch response itself isn't kept around after it's invoked. And lastFetchEtag is invoked once to store on disk. I keep the templateVersion as an instance var so that it can be accessed quickly and frequently from Realtime.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
qdpham13: Done

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
qdpham13: Updated to only update when the accessor method is called. And I keep the templateVersion in fetch b/c its more of a part of fetch and the fetch response then anything and it isn't entirely a Realtime change. Also, there are other ways that fetch can be invoked instead of Realtime, in this cases we would also want to keep updating the templateVersion.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
danasilver: Gotcha. We could follow the same pattern as we do with the ETag and read it from metadata when it's needed (even though it might read from disk). That way there's just one source of truth. wdyt?

https://github.com/firebase/firebase-android-sdk/pull/4051/files#diff-d171ba9c7617c7b7a8ac69a1b277477a505e618633831ba0e9d828e609b87097R318

danasilver (approved):
One suggestion, but LGTM

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
qdpham13: I think that makes sense, I updated it to reflect this

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -531,10 +535,7 @@ private Long getFirstOpenTime() {
   }
 
   public long getTemplateVersionNumber() {
-    if (fetchedConfigsCache.get() != null && fetchedConfigsCache.get().getResult() != null) {
-      return fetchedConfigsCache.get().getResult().getTemplateVersionNumber();
-    }
-    return 1L;
+    return frcMetadata.getLastTemplateVersion();
```
danasilver, line 538: Thinking aloud: we might refactor this to just use `frcMetadata.getLastTemplateVersion();` where we need it in the future but `getTemplateVersionNumber()` is already in use, so this maintains the functionality. 👍

danasilver (approved):
LGTM. Mind updating the PR title to be a bit more descriptive before merging?

PR #4052, opened 2022-09-01
Update latest released version.

rlazo:


vkryachko (commented):
I think both app-distro and dynamic-module support are missing(maybe bacause their versions contain `-beta`?

rlazo:
appdistro was not released. Fixed dynamic-module

vkryachko (approved):

Issue #4045, opened 2022-08-31
firebase-perf is not compliant with market policies


yin-haoran:
Getting running processes before privacy agreement is not compliant with market policies

```
com.google.firebase.perf.session.gauges.GaugeMetadataManager.getCurrentProcessName() {
    // activityManager.getRunningAppProcesses()
}
```

argzdev:
Hi @yin-haoran, thanks for reporting. Which market is it non-compliant with? Also can you share a screenshot of the error of the non compliance?

yin-haoran:
> 

Tencent Appstore. 

stack information of the error of the non compliance as follow
```
android.app.ActivityManager.getRunningAppProcesses(ActivityManager.java:3564)<[---com.google.firebase.perf.session.gauges.GaugeMetadataManager.getCurrentProcessName(GaugeMetadataManager.java:3)<---com.google.firebase.perf.session.gauges.GaugeMetadataManager.<init](https://github.com/firebase/firebase-android-sdk/issues/---com.google.firebase.perf.session.gauges.GaugeMetadataManager.getCurrentProcessName(GaugeMetadataManager.java:3)%3C---com.google.firebase.perf.session.gauges.GaugeMetadataManager.%3Cinit)>(GaugeMetadataManager.java:8)<---com.google.firebase.perf.session.gauges.GaugeMetadataManager.<init>(GaugeMetadataManager.java:1)<---com.google.firebase.perf.session.gauges.GaugeManager.initializeGaugeMetadataManager(GaugeManager.java:1)<---com.google.firebase.perf.session.SessionManager.lambda$setApplicationContext$0(SessionManager.java:1)<---com.google.firebase.perf.session.SessionManager.a(Unknown Source:0)<---com.google.firebase.perf.session.a.run(Unknown Source:6)<---java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:457)<---java.util.concurrent.FutureTask.run(FutureTask.java:266)<---java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)<---java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)<---java.lang.Thread.run(Thread.java:764)
```





















































[...]

PR #4057, opened 2022-09-02
Removing getRunningAppProcesses since the process_name isn't used

raymondlam:
getRunningAppProcesses() was also causing compliance issues with Tencent's appstore #4045


jeremyjiang-dev (approved):
Thanks for fixing this!

jeremyjiang-dev:
The failing unit test seems to be flaky

visumickey (approved):
Thanks for making these changes!

PR #4065, opened 2022-09-07
Properly encode http params.

### LINKED ISSUE #4060: "E/FirebaseCrashlytics: Settings request failed. java.io.FileNotFoundException when build versionName contains spaces

    with the firebase tag.
    google group.

  * Android Studio version:
Android Studio Chipmunk | 2021.2.1 Patch 2
Build #AI-212.5712.43.2112.8815526, built on July 10, 2022
Runtime version: 11.0.12+7-b1504.28-7817840 amd64
VM: OpenJDK 64-Bit Server VM by Oracle Corporation
Windows 8.1 6.3
GC: G1 Young Generation, G1 Old Generation
Memory: 4032M
Cores: 8
Registry: external.system.auto.import.disabled=true, debugger.watches.in.variables=false, ide.instant.shutdown=false
Non-Bundled Plugins: com.intellij.ideolog (203.0.30.0), org.jetbrains.kotlin (212-1.7.10-release-333-AS5457.46)

  * Firebase Component: _____ (Database, Firestore, Storage, Functions, etc)
  * 
plugins {
...
    id 'com.google.firebase.crashlytics'
}
    dependencies {
...
        implementation 'com.google.firebase:firebase-analytics'
        implementation 'com.google.firebase:firebase-crashlytics'
        implementation 'com.google.firebase:firebase-messaging'
...
}

  * Component version: _____

       implementation platform('com.google.firebase:firebase-bom:30.4.0')  // Bill Of Materials
...

also:
    dependencies {
        classpath 'com.android.tools.build:gradle:7.2.2'
        classpath 'com.google.gms:google-services:4.3.13'
        // Add the Crashlytics Gradle plugin
        classpath 'com.google.firebase:firebase-crashlytics-gradle:2.9.1'
    }

A very detailed description of the manifestation this problem which I described on StackOverflow, and a workaround that I subsequently discovered, is given at https://stackoverflow.com/questions/73589421/firebase-crashlytics-crash-reports-only-appearing-when-app-runs-in-pre-release-t/73589741#73589741.  

THE ULTIMATE CAUSE, AND BUG, I believe (and I don't know why this works in Google Play Pre-release Testing) is that the project versionName field from the module-level build.gradle is being used by Crashlytics to create the URL for the settings file (or whatever the initial file is that is attempted to be uploaded at the app's startup), without URL-encoding this string to account for spaces in the URL (and possibly limiting the string to the max chars for a GET URL - this part I did not test.)

There is no documentation limiting this versionName to a non-descriptive, no-spaces content, that I have seen, and no documentation of this limitation in the Crashlytics setup documentation that I have read and followed to get set up.

#### Steps to reproduce:

Follow all of the setup steps for adding firebase & crashlytics to an app here: https://firebase.google.com/docs/android/setup?hl=en and here: https://firebase.google.com/docs/crashlytics/get-started?authuser=0&platform=android, verify that the google-services.json is correct. Enable Analytics sharing. Ensure Crashlytics enabled for app on Firebase console. Run the app and force a crash by using a temporary "Test Crash" button added in the main activity, which throws a RuntimeException.  (Run this in a non-VPN environment.)

On startup, the app gets an error "E/FirebaseCrashlytics: Settings request failed. java.io.FileNotFoundException: https://firebase-settings.crashlytics.com/spi/v2/platforms/android/gmp/(snip)&build_version=99&display_version=2022.08.31.99 blah blah blah" and then when I click on my "TEST CRASH" button, I get another error " E/FirebaseCrashlytics: Cannot send reports. Timed out while fetching settings.".

The thing is, the crash reports DO appear for crashes that occur during the Google Play Pre-Release testing upon upload to Google Play (closed testing), including for the "CRASH TEST" button, and then I get crash-free users, logs, stack trace, the whole thing. Because this part is working, it appeared that my setup is correct.

#### Relevant Code:

In the module level build.gradle, I had a descriptive versionName which explains in Google Play the latest summary of changes:

```versionName "2022.08.31.99 Blah, blah blah```

, which looked suspicious in the error "E/FirebaseCrashlytics: Settings request failed. java.io.FileNotFoundException: https://firebase-settings.crashlytics.com/spi/v2/platforms/android/gmp/(snip)&build_version=99&display_version=2022.08.31.99 blah blah blah", since URLs can't have spaces - so I tried changing it to just:

```versionName "2022.08.31.99```

I recompiled it and ran it over USB on the phone, tapped the test crash button and got a crash report in Crashlytics console almost immediately. Backed it out, the problem occurs, put it back, it's working.

mrober:
Properly encode http params. This should resolve #4060 where special characters were not being encoded.

Should we consider using a typical http library instead of this custom built network package?

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/network/HttpGetRequest.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/network/HttpGetRequest.java
@@ -19,7 +19,9 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
 import java.net.URL;
+import java.net.URLEncoder;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -95,7 +97,8 @@ public class HttpGetRequest {
     return new HttpResponse(responseCode, body);
   }
 
-  private String createUrlWithParams(String url, Map<String, String> queryParams) {
+  private String createUrlWithParams(String url, Map<String, String> queryParams)
+      throws UnsupportedEncodingException {
     String queryParamsString = createParamsString(queryParams);
 
     if (queryParamsString.isEmpty()) {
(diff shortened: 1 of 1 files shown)


mrichards (approved):
LGTM. I agree we should consider moving to a built-in network library at some point soon, but this seems like a great stopgap.

PR #4066, opened 2022-09-07
Realtime rc NO_CHANGE bug

qdpham13:
Fix NO_CHANGE issue by removing the Etag whenever the template version is the default value (0).
https://docs.google.com/document/d/14C1z60RlKwri12pnton8noDlITH45sO6OQxCq7jSTbs/edit#heading=h.xgjl2srtytjt
Merge in after templateVersion cache change -> https://github.com/firebase/firebase-android-sdk/pull/4051

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
@@ -166,7 +166,8 @@ public class ConfigFetchHandler {
   }
 
   /**
-   * Starts fetching configs from the Firebase Remote Config server.
+   * Starts fetching configs from the Firebase Remote Config server without the Etag header in the
+   * request.
    *
    * <p>Guarantees consistency between memory and disk; fetched configs are saved to memory only
    * after they have been written to disk.
@@ -203,7 +204,9 @@ public class ConfigFetchHandler {
             executor,
             (cachedFetchConfigsTask) ->
                 fetchIfCacheExpiredAndNotThrottled(
-                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds, true));
+                    cachedFetchConfigsTask,
+                    minimumFetchIntervalInSeconds, /*excludeEtagHeader */
+                    true));
   }
 
   /**
@@ -216,7 +219,7 @@ public class ConfigFetchHandler {
   private Task<FetchResponse> fetchIfCacheExpiredAndNotThrottled(
       Task<ConfigContainer> cachedFetchConfigsTask,
       long minimumFetchIntervalInSeconds,
-      boolean excludeEtagHeaderForRealtime) {
+      boolean excludeEtagHeader) {
     Date currentTime = new Date(clock.currentTimeMillis());
     if (cachedFetchConfigsTask.isSuccessful()
         && areCachedFetchConfigsValid(minimumFetchIntervalInSeconds, currentTime)) {
@@ -261,10 +264,7 @@ public class ConfigFetchHandler {
                     String installationId = installationIdTask.getResult();
                     String installationToken = installationAuthTokenTask.getResult().getToken();
                     return fetchFromBackendAndCacheResponse(
-                        installationId,
-                        installationToken,
-                        currentTime,
-                        excludeEtagHeaderForRealtime);
+                        installationId, installationToken, currentTime, excludeEtagHeader);
                   });
     }
 
@@ -324,14 +324,10 @@ public class ConfigFetchHandler {
    * {@code fetchedConfigsCache}.
    */
   private Task<FetchResponse> fetchFromBackendAndCacheResponse(
-      String installationId,
-      String installationToken,
-      Date fetchTime,
-      boolean excludeEtagHeaderForRealtime) {
+      String installationId, String installationToken, Date fetchTime, boolean excludeEtagHeader) {
     try {
       FetchResponse fetchResponse =
-          fetchFromBackend(
-              installationId, installationToken, fetchTime, excludeEtagHeaderForRealtime);
+          fetchFromBackend(installationId, installationToken, fetchTime, excludeEtagHeader);
       if (fetchResponse.getStatus() != Status.BACKEND_UPDATES_FETCHED) {
         return Tasks.forResult(fetchResponse);
       }
@@ -354,10 +350,7 @@ public class ConfigFetchHandler {
    */
   @WorkerThread
   private FetchResponse fetchFromBackend(
-      String installationId,
-      String installationToken,
-      Date currentTime,
-      boolean excludeEtagHeaderForRealtime)
+      String installationId, String installationToken, Date currentTime, boolean excludeEtagHeader)
       throws FirebaseRemoteConfigException {
     try {
       HttpURLConnection urlConnection = frcBackendApiClient.createHttpURLConnection();
@@ -372,7 +365,7 @@ public class ConfigFetchHandler {
               customHttpHeaders,
               getFirstOpenTime(),
               currentTime,
-              excludeEtagHeaderForRealtime);
+              excludeEtagHeader);
 
       if (response.getFetchedConfigs() != null) {
         // Set template version in metadata to be saved on disk.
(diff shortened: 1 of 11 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
```diff
@@ -182,9 +182,15 @@ FetchResponse fetch(
       String lastFetchETag,
       Map<String, String> customHeaders,
       Long firstOpenTime,
-      Date currentTime)
+      Date currentTime,
+      Boolean excludeEtagHeaderForRealtime)
```
danasilver, line 186: Instead of calling this with `false` when that's the default, we could be more direct/clear in the callers (and set a good default option when using fetch) by creating an overload and one new method for the "exclude etag" case. That'd also make it easier to turn the double negative around for the default case:

```
FetchResponse fetch(...) {
  fetch(..., /* withEtag */ true);
}

FetchResponse fetchWithoutEtag(...) {
  fetch(..., /* withEtag */ false);
}

private FetchResponse fetch(..., includeEtag) {
  /* existing code + etag condition */
}

File: firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java [hunk 2]
```diff
@@ -173,7 +214,9 @@ public Task<FetchResponse> fetch(long minimumFetchIntervalInSeconds) {
    * fetch time and {@link BackoffMetadata} in {@link ConfigMetadataClient}.
    */
   private Task<FetchResponse> fetchIfCacheExpiredAndNotThrottled(
-      Task<ConfigContainer> cachedFetchConfigsTask, long minimumFetchIntervalInSeconds) {
+      Task<ConfigContainer> cachedFetchConfigsTask,
+      long minimumFetchIntervalInSeconds,
+      boolean excludeEtagHeaderForRealtime) {
```
danasilver, line 219: I think it'd be ok to drop the `forRealtime` in this class since that's abstracted away from the fetch here

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -162,7 +162,48 @@ public Task<FetchResponse> fetch(long minimumFetchIntervalInSeconds) {
...
+   *     updates, the {@link FetchResponse}'s configs will be {@code null}.
+   */
+  public Task<FetchResponse> fetchWithoutEtag(long minimumFetchIntervalInSeconds) {
+    return fetchedConfigsCache
+        .get()
+        .continueWithTask(
+            executor,
+            (cachedFetchConfigsTask) ->
+                fetchIfCacheExpiredAndNotThrottled(
+                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds, true));
```
danasilver, line 206: Mind adding an inline comment like `/* excludeEtagHeader */ true`

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -162,7 +162,48 @@ public Task<FetchResponse> fetch(long minimumFetchIntervalInSeconds) {
             executor,
             (cachedFetchConfigsTask) ->
                 fetchIfCacheExpiredAndNotThrottled(
-                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds));
+                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds, false));
+  }
+
+  /**
+   * Starts fetching configs from the Firebase Remote Config server.
```
danasilver, line 169: I might include `without setting the ETag header` up here just cause it's the main distinction

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandlerTest.java
```diff
@@ -736,6 +738,17 @@ public void getInfo_hitsThrottleLimit_throttledStatus() throws Exception {
         .isEqualTo(firstFetchedContainer.getFetchTime());
   }
 
+  @Test
+  public void realtimeFetch_excludeEtag_andSuccess() throws Exception {
```
danasilver, line 742: I'd use `fetchWithoutEtag_` as the prefix since we're testing that method here. Can we assert the ETag header isn't sent?

danasilver (approved):
A few minor comments. LGTM

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -328,14 +328,14 @@ synchronized void beginRealtimeHttpStream() {
   // Pauses Http stream listening
   synchronized void closeRealtimeHttpStream() {
     if (httpURLConnection != null) {
-      this.httpURLConnection.disconnect();
-
       // Explicitly close the input stream due to a bug in the Android okhttp implementation.
       // See github.com/firebase/firebase-android-sdk/pull/808.
       try {
+        this.httpURLConnection.getErrorStream().close();
```
danasilver, line 334: Might have accidentally added this file to the PR?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -162,7 +162,51 @@ public Task<FetchResponse> fetch(long minimumFetchIntervalInSeconds) {
             executor,
             (cachedFetchConfigsTask) ->
                 fetchIfCacheExpiredAndNotThrottled(
-                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds));
+                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds, false));
```
danasilver, line 165: same inline comment here

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java
danasilver, line 186: same thing since this is downstream of the fetch handler... I think this can just be `excludeEtagHeader`

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -1164,6 +1164,23 @@ public void realtime_stream_listen_and_failsafe_disabled() throws Exception {
     verify(mockFetchHandler).fetch(0);
   }
 
+  @Test
+  public void realtime_fetch_exclude_Etag() throws Exception {
```
danasilver, line 1168: maybe something like `realtimeFetch_storedTemplateVersionZero_fetchExcludesEtag`?

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandlerTest.java
```diff
@@ -926,7 +945,8 @@ private void verifyETags(@Nullable String requestETag, String responseETag) thro
             /* lastFetchETag= */ eq(requestETag),
             /* customHeaders= */ any(),
             /* firstOpenTime= */ any(),
-            /* currentTime= */ any());
+            /* currentTime= */ any(),
+            /* excludeEtagHeaderForRealtime */ any());
```
danasilver, line 949: Inline comment updates here too

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -162,7 +162,53 @@ public Task<FetchResponse> fetch(long minimumFetchIntervalInSeconds) {
             executor,
             (cachedFetchConfigsTask) ->
                 fetchIfCacheExpiredAndNotThrottled(
-                    cachedFetchConfigsTask, minimumFetchIntervalInSeconds));
+                    cachedFetchConfigsTask,
+                    minimumFetchIntervalInSeconds, /*excludeEtagHeader */
```
danasilver, line 166: wonder if we can get the formatter to leave this on the right line... no need to worry about it right now

danasilver (approved):

PR #4071, opened 2022-09-09
Revamp test harness for macrobenchmark tests

yifanyang:
* Upgrade Gradle version to 7.5.1
* Upgrade Android Gradle plugin version to 7.2.2
* Support trace collection for [custom event](https://d.android.com/topic/performance/tracing/custom-events)
* Support building and running tests manually


ci/fireci/fireciplugins/macrobenchmark.py
```diff
@@ -78,35 +92,39 @@ def _artifact_key_version(artifact):
...
+  with open('health-metrics/benchmark/config.yaml') as yaml_file:
+    config = yaml.safe_load(yaml_file)
+    for app in config['test-apps']:
+      app['plugins'] = app.get('plugins', [])
+      app['traces'] = app.get('traces', [])
+      app['plugins'].extend(config['common-plugins'])
+      app['traces'].extend(config['common-traces'])
+
+    # Adding an empty android app for baseline comparison
+    config['test-apps'].insert(0, {'sdk': 'baseline', 'name': 'baseline'})
```
vkryachko, line 105: Since we are now measuring Firebase traces only, do we still need this baseline to be present?

yifanyang: Makes sense. Removed it and the reference to it in the README.

vkryachko (approved):

PR #4072, opened 2022-09-09
Make dackka task depend on docStubs task.

vkryachko:
Otherwise no docs are produced for java sdks.

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
@@ -92,6 +92,7 @@ abstract class DackkaPlugin : Plugin<Project> {
                         val sourcesForKotlin = emptyList<File>()
                         val packageLists = fetchPackageLists(project)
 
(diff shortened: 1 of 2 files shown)


rlazo (approved):

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -76,22 +79,24 @@ abstract class DackkaPlugin : Plugin<Project> {
...
                         sources.set(project.provider { sourcesForJava })
                     }
 
                     docsTask.configure {
+                        clientName.set(project.firebaseConfigValue { artifactId })
                         // this will become useful with the agp upgrade, as they're separate in 7.x+
                         val sourcesForKotlin = emptyList<File>()
                         val packageLists = fetchPackageLists(project)
 
+                        if (!isKotlin) dependsOn(docStubs)
```
daymxn, line 99: Nice catch! Ideally we should switch this to `dependsOnAndMustRunAfter` in the future, but it's not breaking for now.

PR #4078, opened 2022-09-12
Add examples for new Flow operators in firestore-ktx

zsmb13:
Available from [24.3.0](https://firebase.google.com/support/release-notes/android#firestore-ktx_v24-3-0)


thatfiredev (approved):
LGTM!

---

For reference: Had an offline discussion with @zsmb13 and agreed that this would be a good reference to point developers to until Flows are properly documented in our public docs.

PR #4082, opened 2022-09-13
Add Javadoc support to the DackkaPlugin

daymxn:
In accordance with [b/245753889](https://b.corp.google.com/issues/245753889), this adds support to the DackkaPlugin to generate Javadoc variants.

More specifically this:
- Removes the `deleteDackkaGeneratedJavaReferences` task
- Refactors `copyOutputToCommonDirectory` to two separate task
- Adds a new task to copy the Javadoc output via `copyJavaDocToCommonDirectory`
- Adds a new task to copy the Kotlindoc output via `copyKotlinDocToCommonDirectory`
- Adds an extension method to the `Copy` task in gradle for copying individual directories
- Fixes the `DackkaPlugin` documentation to reflect the new changes

Keep in mind that the current implementation in this pr requires the project setting `dackkaJavadoc` to be equal to `true`, otherwise Dackka won't copy the Javadoc output to the root directory.

For example, running `./gradlew :firebase-common:kotlindoc` will yield this at the root build directory
```
firebase-kotlindoc/
  kotlin/
```
Wheres running this `./gradlew -PdackkaJavadoc=true :firebase-common:kotlindoc` will yield this at the root build directory
```
firebase-kotlindoc/
  kotlin/
  android/
```

Android contains the Firesite ready Javadoc files for staging.

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
@@ -90,26 +90,23 @@ tasks above). While we do not currently offer any configuration for the Dackka
 plugin, this could change in the future as needed. Currently, the DackkaPlugin
 provides sensible defaults to output directories, package lists, and so forth.
 
-The DackkaPlugin also provides two extra tasks:
-[cleanDackkaDocumentation][registerCleanDackkaDocumentation] and
-[deleteDackkaGeneratedJavaReferences][registerDeleteDackkaGeneratedJavaReferencesTask].
+The DackkaPlugin also provides three extra tasks:
+[cleanDackkaDocumentation][registerCleanDackkaDocumentation],
+[copyJavaDocToCommonDirectory][registerCopyJavaDocToCommonDirectoryTask] and
+[copyKotlinDocToCommonDirectory][registerCopyKotlinDocToCommonDirectoryTask].
 
 _cleanDackkaDocumentation_ is exactly what it sounds like, a task to clean up (delete)
 the output of Dackka. This is useful when testing Dackka outputs itself- and
 shouldn't be apart of the normal flow. The reasoning is that it would otherwise
 invalidate the gradle cache.
 
-_deleteDackkaGeneratedJavaReferences_ is a temporary addition. Dackka generates
-two separate styles of docs for every source set: Java & Kotlin. Regardless of
-whether the source is in Java or Kotlin. The Java output is how the source looks
-from Java, and the Kotlin output is how the source looks from Kotlin. We publish
-these under two separate categories, which you can see here:
-[Java](https://firebase.google.com/docs/reference/android/packages)
-or
-[Kotlin](https://firebase.google.com/docs/reference/kotlin/packages).
-Although, we do not currently publish Java packages with Dackka- and will wait
-until we are more comfortable with the output of Dackka to do so. So until then,
-this task will remove all generate Java references from the Dackka output.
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -234,38 +231,39 @@ abstract class DackkaPlugin : Plugin<Project> {
...
 
-    // If we decide to publish java variants, we'll need to address the generated format as well
-    // TODO(b/243833009): Make task cacheable
-    private fun registerDeleteDackkaGeneratedJavaReferencesTask(project: Project, outputDirectory: Provider<File>) =
-        project.tasks.register<Delete>("deleteDackkaGeneratedJavaReferences") {
-            mustRunAfter("generateDackkaDocumentation")
+    // TODO(b/246593212): Migrate doc files to single directory
+    private fun registerCopyJavaDocToCommonDirectoryTask(project: Project, outputDirectory: Provider<File>) =
+        project.tasks.register<Copy>("copyJavaDocToCommonDirectory") {
+            if (project.rootProject.findProperty("dackkaJavadoc") == "true") {
```
vkryachko, line 242: Is this going to be cache compliant? if not can you make a note of it here with justification for why it's ok?(i.e. that it's a very temporary flag)

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GradleUtils.kt
```diff
@@ -0,0 +1,10 @@
+package com.google.firebase.gradle.plugins
+
+import java.io.File
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.Copy
+
+fun Copy.fromDirectory(directory: Provider<File>) =
+    from(directory) {
+        into(directory.map { it.nameWithoutExtension })
+    }
```
vkryachko, line 10: Not sure what this does, is it copying into the same directory? i.e. what if the directory has no extension?

also, not sure if you need nesting, instead of
```suggestion
fun Copy.fromDirectory(directory: Provider<File>) {
    from(directory)
    into(directory.map { it.nameWithoutExtension })
}
```

daymxn: This is different from 
```
from()
into()
```

If you do `from("daymon/myFolder")` it will copy everything under `myFolder`, not the folder itself.

The general solution to this is as such:

```
from("daymon/myFolder")
into("vlad/myFolder")
```

Where it copies all the contents of `myFolder` into a new `myFolder` under the directory you want. This isn't as straight forward when you have multiple directories that you need to do this for- in one `Copy` task.

This extension method fixes that.

```
    from(myFolder) {
        into(myFolder.nameWithoutExtension)
    }
```

Instead of _just_ copying the contents of the folder, this pastes them into a temp subdirectory (that doesn't exist beyond the task), and returns that temp folder instead. This way, you get the folder AND its contents, versus just the contents. Its a little work-around, abstracted by an extension method (`fromDirectory`).

Furthermore, `nameWithoutExtension` is not specific to files with extensions. This is supposed to be used on directories, which don't actually have extensions. `nameWithoutExtension` is an extension method on `File`- provided by `kotlin.io`. If the file doesn't have an extension, it just returns the file name. FWIW, `nameWithoutExtension` could be removed.

vkryachko: Makes sense, thanks for a detailed explanation.

vkryachko (approved):

rlazo (approved):

PR #4087, opened 2022-09-14
Implemented exponential backoff and max retry with resumable uploads

maneesht:
We have exponential backoff implemented, but not for `uploadChunk`. If `uploadChunk` fails, we check if we are in a recoverable state, and then request the status of the upload, and then retry the upload immediately if the request was successful. 

Instead of immediately retrying the upload, we should wait until the appropriate number of seconds, and then retry again. And we should only do so a max number of times. This PR implements both - exponential backoff and max retries.


firebase-storage/src/test/java/com/google/firebase/storage/TestUtil.java
```diff
@@ -32,6 +32,12 @@
 public class TestUtil {
 
   static FirebaseApp createApp() {
+    // Many tests require you to call the callback on the same thread that was initially
+    // instantiated.
+    // With the 5 second keepalive, after 5 seconds, the thread will get killed and eventually a new
+    // one will be created.
+    // Therefore causing many of the tests to fail
+    StorageTaskScheduler.setCallbackQueueKeepAlive(90, TimeUnit.SECONDS);
```
maneesht, line 40: To add additional context:

We call `attachListeners` https://github.com/firebase/firebase-android-sdk/blob/ed10eb5ebe998dc23ead46419a2b4dcbb5e2482c/firebase-storage/src/testUtil/java/com/google/firebase/storage/TestUploadHelper.java#L49 to set up the listeners and then those listeners call `ControllableSchedulerHelper` to verify the callback thread that it initially was called with is the same one that is called in subsequent events https://github.com/firebase/firebase-android-sdk/blob/ed10eb5ebe998dc23ead46419a2b4dcbb5e2482c/firebase-storage/src/testUtil/java/com/google/firebase/storage/ControllableSchedulerHelper.java#L63
However, in cases where we have long backoffs, this won't be the case, as the thread pool will terminate any threads that are not executed within the keepalive time, which is 5 seconds.
Our alternative is to set the keepAlive, or disable the threads from being killed. A third option would be to remove this verification check.

firebase-storage/src/test/java/com/google/firebase/storage/TestUtil.java
```diff
@@ -148,6 +154,6 @@ static void await(Task<?> task, int timeout, TimeUnit timeUnit) throws Interrupt
    * Tasks to be executed.
    */
   static void await(Task<?> task) throws InterruptedException {
-    await(task, 3, TimeUnit.SECONDS);
+    await(task, 10, TimeUnit.SECONDS);
```
tonyjhuang, line 157: why? 

don't forget to update the javadoc

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
```diff
@@ -130,7 +160,7 @@ public void cantUploadToRoot() throws Exception {
         });
 
     // TODO(mrschmidt): Lower the timeout
-    TestUtil.await(task, 300, TimeUnit.SECONDS);
+    TestUtil.await(task, 7, TimeUnit.MINUTES);
```
tonyjhuang, line 163: this seems at odds with the line comment above😅 why is this necessary?

tonyjhuang (commented):
small comments, overall looks good!

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
```diff
@@ -72,6 +79,13 @@ public class UploadTask extends StorageTask<UploadTask.TaskSnapshot> {
   private volatile Exception mServerException = null;
   private volatile int mResultCode = 0;
   private volatile String mServerStatus;
+  private volatile long maxSleepTime;
+  private static final Random random = new Random();
+  /*package*/ static Sleeper sleeper = new SleeperImpl();
+  /*package*/ static Clock clock = DefaultClock.getInstance();
+  private int sleepTime =
+      0; // TODO(mtewani): Make it so that the send is 0,1,2,4,8,... and start at 0
+  private final int sleepInterval = 1000;
```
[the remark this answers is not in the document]
maneesht: Yes

firebase-storage/src/test/java/com/google/firebase/storage/TestUtil.java
maneesht: This is to compensate for the backoff we just implemented. Updated the doc

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
```diff
@@ -410,6 +430,24 @@ private boolean recoverStatus(boolean withRetry) {
...
+    } catch (InterruptedException e) {
+      Log.w(TAG, "thread interrupted during exponential backoff.");
+
+      Thread.currentThread().interrupt();
+      mServerException = e;
+      return false;
+    }
+    boolean sendRes = send(request);
+    if (sendRes) {
+      sleepTime = 0;
```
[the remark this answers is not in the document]
maneesht: Done.

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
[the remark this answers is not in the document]
tonyjhuang: consider renaming to minimumSleepInterval?

firebase-storage/src/test/java/com/google/firebase/storage/TestUtil.java
maneesht: I think in case the wrong executor gets called by `UploadTask`. 

Here's a quick StackOverflow answer that talks about the tradeoffs:
https://stackoverflow.com/a/18225863

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
maneesht: Ended up changing this to 1 minute. The 7 was a precaution :-)

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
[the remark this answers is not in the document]
maneesht: Done

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
```diff
@@ -72,6 +79,13 @@ public class UploadTask extends StorageTask<UploadTask.TaskSnapshot> {
   private volatile Exception mServerException = null;
   private volatile int mResultCode = 0;
   private volatile String mServerStatus;
+  private volatile long maxSleepTime;
+  private static final Random random = new Random();
+  /*package*/ static Sleeper sleeper = new SleeperImpl();
+  /*package*/ static Clock clock = DefaultClock.getInstance();
+  private int sleepTime =
+      0; // TODO(mtewani): Make it so that the send is 0,1,2,4,8,... and start at 0
```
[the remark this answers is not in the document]
maneesht: Done

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
```diff
@@ -321,15 +337,18 @@ private boolean shouldContinue() {
...
-        tryChangeState(INTERNAL_STATE_FAILURE, false);
+    if (inErrorState) {
+      if (currentTime > deadLine || !recoverStatus(true)) {
+        if (serverStateValid()) {
+          tryChangeState(INTERNAL_STATE_FAILURE, false);
+        }
+        return false;
       }
-      return false;
+      sleepTime = Math.max(sleepTime * 2, sleepTime + (minimumSleepInterval * 2));
```
tonyjhuang, line 350: you probably want

`Math.max(sleepTime * 2, minimumSleepInterval)`

The way this is written, the first sleep time == 2000

tonyjhuang (approved):

firebase-storage/src/main/java/com/google/firebase/storage/UploadTask.java
maneesht: Oh wow. Nice catch! Will change

PR #4095, opened 2022-09-14
Add strict mode testing in firebase-messaging

emilypgoogle:
Part of a larger effort to resolve and prevent strict mode violations for eager components.


gsakakihara:
Just to clarify, this isn't actually resolving any strict mode violations, right? This appears to be adding a new test to prevent any new strict mode violations from being added in the future? If so, could you change the commit message to reflect that? I'm slightly concerned that someone just looking at the commit messages could think that there was a problem before and misdiagnose a strict mode issue, believing that it was caused by FCM and not identifying the actual cause.

emilypgoogle:
Firebase squashes commits so I changed the PR name and it'll use that instead of what I used for my branch

gsakakihara (approved):

PR #4100, opened 2022-09-15
lintVersion: 30.2.2 -> 30.3.0-rc01 

kaibolay:
fix `UsingOnClickInXml` which did not work correctly when running in AGP 7.2.2.


kaibolay:
Better fix: https://github.com/firebase/firebase-android-sdk/pull/4104

PR #4108, opened 2022-09-15
Upgrade to AGP 7.3

thatfiredev:
Not a good idea to merge this now since it will require everyone to use AS Dolphin... But I'm curious to see if the build will succeed.


vkryachko:
You may have to also upgrade our CI infra projects, namely

* https://github.com/firebase/firebase-android-sdk/blob/master/health-metrics/benchmark/template/build.gradle
* https://github.com/firebase/firebase-android-sdk/blob/master/smoke-tests/build.gradle

Maybe more places as well, cc @yifanyang if I missed anything

thatfiredev:
Closing now that we're at AGP 7.4:

https://github.com/firebase/firebase-android-sdk/blob/f419204d771bb5e3417d83239b92503601dbbb42/buildSrc/build.gradle.kts#L68

PR #4110, opened 2022-09-15
Add projectSpecificSources back to the DackkaPlugin

daymxn:
This fixes: [b/246972998](https://b.corp.google.com/issues/246972998)

Removing `projectSpecificSources` was an over-sight, and led to `com.google.firebase.Timestamp` not being generated during doc generation. Adding the method back fixes this.

Additionally, I removed the `isKotlin` check for `excludedFiles` as it was redundant (since the method checks for a specific project, it would be ignored when needed as a by-product).

I also added an additional util method for `dependsOnAndMustRunAfter` and changed our `dependsOn(docStubs)` as such. We need `docStubs` to run _before_ `generateDackkaDocumentation`, and not having `mustRunAfter` was leading to not-so-fun gradle "optimizations".

And some restructuring for readability and consistency :')

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
@@ -168,6 +168,7 @@ abstract class DackkaPlugin : Plugin<Project> {
                     val classpath = compileConfiguration.getJars() + project.javadocConfig.getJars() + project.files(bootClasspath)
 
                     val sourcesForJava = sourceSets.flatMap {
+                        // TODO(b/246984444): Investigate why kotlinDirectories includes javaDirectories
                         it.javaDirectories.map { it.absoluteFile }
                     }
 
@@ -177,13 +178,12 @@ abstract class DackkaPlugin : Plugin<Project> {
                     }
 
                     docsTask.configure {
-                        clientName.set(project.firebaseConfigValue { artifactId })
-                        // this will become useful with the agp upgrade, as they're separate in 7.x+
-                        val sourcesForKotlin = emptyList<File>()
+                        if (!isKotlin) dependsOnAndMustRunAfter(docStubs)
+
+                        val sourcesForKotlin = emptyList<File>() + projectSpecificSources(project)
                         val packageLists = fetchPackageLists(project)
 
-                        if (!isKotlin) dependsOn(docStubs)
-                        val excludedFiles = if (!isKotlin) projectSpecificSuppressedFiles(project) else emptyList()
+                        val excludedFiles = projectSpecificSuppressedFiles(project)
                         val fixedJavaSources = if (!isKotlin) listOf(project.docStubs) else sourcesForJava
 
                         javaSources.set(fixedJavaSources)
@@ -207,10 +207,19 @@ abstract class DackkaPlugin : Plugin<Project> {
         }.toList()
 
     // TODO(b/243534168): Remove when fixed
+    private fun projectSpecificSources(project: Project) =
+        when (project.name) {
+            "firebase-common" -> {
+                project.project(":firebase-firestore").files("src/main/java/com/google/firebase").toList()
+            }
+            else -> emptyList()
+        }
+
+    // TODO(b/243534168): Remove when fixed
     private fun projectSpecificSuppressedFiles(project: Project): List<File> =
         when (project.name) {
             "firebase-common" -> {
-                project.files("${project.docStubs}/com/google/firebase/firestore").toList()
+                project.project(":firebase-firestore").files("src/main/java/com/google/firebase/firestore").toList()
             }
             "firebase-firestore" -> {
                 project.files("${project.docStubs}/com/google/firebase/Timestamp.java").toList()
@@ -226,6 +235,7 @@ abstract class DackkaPlugin : Plugin<Project> {
 
         dackkaJarFile.set(dackkaFile)
         outputDirectory.set(dackkaOutputDirectory)
+        clientName.set(project.firebaseConfigValue { artifactId })
     }
 
     // TODO(b/243833009): Make task cacheable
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -206,11 +206,20 @@ abstract class DackkaPlugin : Plugin<Project> {
             include("**/package-list")
         }.toList()
 
+    // TODO(b/243534168): Remove when fixed
+    private fun projectSpecificSources(project: Project) =
+        when (project.name) {
+            "firebase-common" -> {
+                project.project(":firebase-firestore").files("src/main/java/com/google/firebase").toList()
```
vkryachko, line 213: Isn't "files()" recursive in this case? i.e. won't it include all of firestore's source as opposed to just `Timestamp.java`?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -206,11 +206,20 @@ abstract class DackkaPlugin : Plugin<Project> {
             include("**/package-list")
         }.toList()
 
+    // TODO(b/243534168): Remove when fixed
+    private fun projectSpecificSources(project: Project) =
+        when (project.name) {
```
vkryachko, line 211: Additionally I think you want to subtract `Timestamp.java` from firestore's source, otherwise it runs the risk of overwriting `com/google/firebase/package-summary.html` of firebase-common, Which I think is the case today in prod: https://firebase.google.com/docs/reference/kotlin/com/google/firebase/package-summary

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
```diff
@@ -63,6 +64,14 @@ fun <T : Task, R : Task> T.dependsOnAndMustRunAfter(otherTask: R) {
     dependsOn(otherTask)
 }
 
+/**
+ * Utility method to call [Task.mustRunAfter] and [Task.dependsOn] on the specified task
+ */
+fun <T : Task, R : Task> T.dependsOnAndMustRunAfter(otherTask: Provider<R>) {
+    mustRunAfter(otherTask)
+    dependsOn(otherTask)
```
vkryachko, line 72: doesn't `dependsOn` imply `mustRunAfter` though?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -177,13 +178,12 @@ abstract class DackkaPlugin : Plugin<Project> {
                     }
 
                     docsTask.configure {
-                        clientName.set(project.firebaseConfigValue { artifactId })
-                        // this will become useful with the agp upgrade, as they're separate in 7.x+
-                        val sourcesForKotlin = emptyList<File>()
+                        if (!isKotlin) dependsOnAndMustRunAfter(docStubs)
```
vkryachko, line 181: isn't it cleaner to wire this dependency up via `javaSources.set(docStubs.flatMap { it.output })`?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
daymxn: You would think so, but it doesn't. Sometimes it will run after it, sometimes it will run alongside it. Depending on how gradle decides to optimize things.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
daymxn: Yeah, but you'd be invoking docStubs for `kotlin` modules as well.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
daymxn: We do that in `projectSpecificSuppressedFiles`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
daymxn: Actually, it includes the directory. Dackka doesn't support individual files, only directories. As a work around, we suppress everything but timestamp for common in `projectSpecificSuppressedFiles`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
vkryachko: sure, but it can be conditional right?

```kotlin
javaSources.set(if(isKotlin) sourcesForJava  else docStubs.flatMap { it.output })
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ProjectUtils.kt
vkryachko: I see, in that case inputs/outputs seems like the preferred way to model deps iiuc

daymxn: > I see, in that case inputs/outputs seems like the preferred way to model deps iiuc

Hmm, I was having some issues before regarding this- but I'm unable to replicate them anymore. It was like wiring a dependency didn't implicitly force the depended task to be up-to-date. I'll revert this for now, and we can come back to it should the issue pop back up. Could've just been some dependency being cached incorrectly- which we'll wanna find if that's the case.

vkryachko (approved):

PR #4112, opened 2022-09-16
Make a best effort attempt to flush reports at crash time

mrober:
Make a best effort attempt to flush reports at crash time. This should allow us to upload reports for app initialization crashes.  This will also prevent the app from needing to relaunch to upload the scheduled reports. If the best effort attempt fails, this will still go back to the original behaviour.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSender.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSender.java
@@ -17,9 +17,12 @@ package com.google.firebase.crashlytics.internal.send;
 import android.content.Context;
 import androidx.annotation.NonNull;
 import com.google.android.datatransport.Encoding;
+import com.google.android.datatransport.Priority;
 import com.google.android.datatransport.Transformer;
 import com.google.android.datatransport.Transport;
 import com.google.android.datatransport.cct.CCTDestination;
+import com.google.android.datatransport.runtime.Destination;
+import com.google.android.datatransport.runtime.TransportContext;
 import com.google.android.datatransport.runtime.TransportRuntime;
 import com.google.android.gms.tasks.Task;
 import com.google.firebase.crashlytics.internal.common.CrashlyticsReportWithSessionId;
@@ -46,14 +49,22 @@ public class DataTransportCrashlyticsReportSender {
       (r) -> TRANSFORM.reportToJson(r).getBytes(Charset.forName("UTF-8"));
 
   private final ReportQueue reportQueue;
-  private final Transformer<CrashlyticsReport, byte[]> transportTransform;
+
+  private static final Destination destination =
+      new CCTDestination(CRASHLYTICS_ENDPOINT, CRASHLYTICS_API_KEY);
+  static final TransportContext transportContext =
+      TransportContext.builder()
+          .setBackendName(destination.getName())
+          .setExtras(destination.getExtras())
+          .setPriority(Priority.HIGHEST)
+          .build();
 
   public static DataTransportCrashlyticsReportSender create(
       Context context, SettingsProvider settingsProvider, OnDemandCounter onDemandCounter) {
     TransportRuntime.initialize(context);
     final Transport<CrashlyticsReport> transport =
         TransportRuntime.getInstance()
-            .newFactory(new CCTDestination(CRASHLYTICS_ENDPOINT, CRASHLYTICS_API_KEY))
+            .newFactory(destination)
             .getTransport(
                 CRASHLYTICS_TRANSPORT_NAME,
                 CrashlyticsReport.class,
@@ -61,13 +72,11 @@ public class DataTransportCrashlyticsReportSender {
                 DEFAULT_TRANSFORM);
     ReportQueue reportQueue =
         new ReportQueue(transport, settingsProvider.getSettingsSync(), onDemandCounter);
-    return new DataTransportCrashlyticsReportSender(reportQueue, DEFAULT_TRANSFORM);
+    return new DataTransportCrashlyticsReportSender(reportQueue);
   }
 
-  DataTransportCrashlyticsReportSender(
-      ReportQueue reportQueue, Transformer<CrashlyticsReport, byte[]> transportTransform) {
+  DataTransportCrashlyticsReportSender(ReportQueue reportQueue) {
     this.reportQueue = reportQueue;
-    this.transportTransform = transportTransform;
   }
 
   @NonNull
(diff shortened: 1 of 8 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/ReportQueue.java
```diff
@@ -115,6 +117,24 @@ TaskCompletionSource<CrashlyticsReportWithSessionId> enqueueReport(
...
+                  .upload(DataTransportCrashlyticsReportSender.transportContext, 1, () -> {});
+              latch.countDown();
+            })
+        .start();
+    try {
+      //noinspection ResultOfMethodCallIgnored best effort only
+      latch.await(2, TimeUnit.SECONDS);
+    } catch (InterruptedException unused) {
+
+    }
```
vkryachko, line 135: You shouldn't swallow this exception: https://engdoc.corp.google.com/eng/doc/devguide/java/practices/interruption.md#interruptedexception

Instead maybe mimic awaitUninterruptibly from guava https://github.com/google/guava/blob/94786f7c419041068ad48f04fe3d9fa4a266b6dd/guava/src/com/google/common/util/concurrent/Uninterruptibles.java#L56-L72

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/ReportQueue.java
```diff
@@ -115,6 +117,24 @@ TaskCompletionSource<CrashlyticsReportWithSessionId> enqueueReport(
...
+            () -> {
+              TransportRuntime.getInstance()
+                  .getUploader()
+                  .upload(DataTransportCrashlyticsReportSender.transportContext, 1, () -> {});
+              latch.countDown();
+            })
+        .start();
+    try {
+      //noinspection ResultOfMethodCallIgnored best effort only
+      latch.await(2, TimeUnit.SECONDS);
```
vkryachko, line 132: If you block waiting for the thread to finish, why have a `Thread` at all instead of calling the blocking method directly?

vkryachko (changes requested):

vkryachko:
@emilypgoogle jfyi

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSender.java
```diff
@@ -46,28 +49,34 @@ public class DataTransportCrashlyticsReportSender {
...
-  private final Transformer<CrashlyticsReport, byte[]> transportTransform;
+
+  private static final Destination destination =
+      new CCTDestination(CRASHLYTICS_ENDPOINT, CRASHLYTICS_API_KEY);
+  static final TransportContext transportContext =
+      TransportContext.builder()
+          .setBackendName(destination.getName())
+          .setExtras(destination.getExtras())
+          .setPriority(Priority.HIGHEST)
+          .build();
```
vkryachko, line 60: Hm, for some reason my previous comment on this disappeared, will try to replicate it again:

`TransportContext` as well as `TransportRuntime.getUploader()` are completely internal to firelog and I'd rather we did have teams call them directly. Instead I would suggest to encapsulate this logic with something like the following:

```java
package com.google.android.datatransport.runtime;

@Discouraged(
  message = "TransportRuntime is not a realtime delivery system, don't use unless you absolutely must.")
public final class ForcedSender {
  private ForcedSender() {}

  @WorkerThread
  public static void sendBlocking(Transport<?> transport) {
    TransportContext ctx = castToTransportImplAndGetContextOutOrThrow(transport);
    TransportRuntime.getInstance().getUploader().send(ctx, 1, () -> {});
  }
```

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/ReportQueue.java
mrober: Done

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/ReportQueue.java
mrober: We want this to time out after 2 seconds, so it never leaves the app hanging after the uncaught exception.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/send/DataTransportCrashlyticsReportSender.java
mrober: Done

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ForcedSender.java
```diff
@@ -0,0 +1,26 @@
...
+import android.annotation.SuppressLint;
+import androidx.annotation.Discouraged;
+import androidx.annotation.WorkerThread;
+import com.google.android.datatransport.Priority;
+import com.google.android.datatransport.Transport;
+
+@Discouraged(
+    message =
+        "TransportRuntime is not a realtime delivery system, don't use unless you absolutely must.")
+public final class ForcedSender {
```
vkryachko, line 12: nit
```suggestion
public final class ForcedSender {
  private ForcedSender() {}
```

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ForcedSender.java
```diff
@@ -0,0 +1,26 @@
...
+  @WorkerThread
+  public static void sendBlockingWithPriority(Transport<?> transport, Priority priority) {
```
vkryachko, line 14: nit 
```suggestion
  public static void sendBlocking(Transport<?> transport, Priority priority) {
```

arguments are part of the method signature, so no need to include the priority in the method name

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ForcedSender.java
```diff
@@ -0,0 +1,26 @@
...
+    @SuppressLint("DiscouragedApi")
+    TransportContext context = getTransportContextOrThrow(transport).withPriority(priority);
+    TransportRuntime.getInstance().getUploader().upload(context, 1, () -> {});
```
vkryachko, line 17: `upload()` is asynchronous, you want to use logAndUpdateState instead, see https://github.com/firebase/firebase-android-sdk/blob/fdb88d7c9a072b180bec441612a05337549a3d38/transport/transport-runtime-testing/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/UploadTestSupport.java#L28-L30

vkryachko (commented):
We probably need some tests as well

mrichards (approved):
LGTM!

transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/scheduling/jobscheduling/Uploader.java
```diff
@@ -111,7 +111,7 @@ public void upload(TransportContext transportContext, int attemptNumber, Runnabl
         });
   }
 
-  BackendResponse logAndUpdateState(TransportContext transportContext, int attemptNumber) {
+  public BackendResponse logAndUpdateState(TransportContext transportContext, int attemptNumber) {
```
vkryachko, line 114: Please add `@RestrictTo(LIBRARY_GROUP)` annotation

vkryachko (approved):

PR #4114, opened 2022-09-16
[Fireperf][AASA] send `_experiment_app_start_ttid` trace, controlled by RC flag

leotianlizhan:
b/245947422
b/243540978

During startup, when we want to decide to record this trace or not, we only check disk cache for RC flag because RC is not initialized yet. 

This RC flag will be saved to disk ONLY upon successful RC `fetchAndActivate()`. There's a problem with all other flags in that they are being saved to disk every time they are queried.

### DIFF
--- a/firebase-perf/dev-app/src/main/AndroidManifest.xml
+++ b/firebase-perf/dev-app/src/main/AndroidManifest.xml
@@ -39,7 +39,7 @@
             android:name="fragment_sampling_percentage"
             android:value="100.0" />
         <meta-data
-            android:name="experiment_as_ttid"
+            android:name="experiment_app_start_ttid"
             android:value="true" />
 
         <receiver
--- a/firebase-perf/dev-app/src/main/AndroidManifest.xml
+++ b/firebase-perf/dev-app/src/main/AndroidManifest.xml
@@ -38,6 +38,9 @@
         <meta-data
             android:name="fragment_sampling_percentage"
             android:value="100.0" />
+        <meta-data
+            android:name="experiment_as_ttid"
+            android:value="true" />
 
         <receiver
             android:name=".FirebasePerfTestReceiver"
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
@@ -115,19 +115,9 @@ public class ConfigResolver {
 
   /** Default API to call for whether performance monitoring is currently silent. */
   public boolean isPerformanceMonitoringEnabled() {
-    saveExperimentFlagToDeviceCache();
     Boolean isPerformanceCollectionEnabled = getIsPerformanceCollectionEnabled();
     return (isPerformanceCollectionEnabled == null || isPerformanceCollectionEnabled == true)
         && getIsServiceCollectionEnabled();
-  }
-
-  // TODO: remove after _experiment_as_ttid is finished
-  private void saveExperimentFlagToDeviceCache() {
-    ExperimentTTID config = ExperimentTTID.getInstance();
-    Optional<Boolean> rcValue = getRemoteConfigBoolean(config);
-    if (rcValue.isAvailable()) {
-      deviceCacheManager.setValue(config.getDeviceCacheFlag(), rcValue.get());
-    }
   }
 
   /** Returns whether developers have enabled Firebase Performance event collection. */
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
@@ -24,6 +24,7 @@ import com.google.firebase.perf.BuildConfig;
 import com.google.firebase.perf.config.ConfigurationConstants.CollectionDeactivated;
 import com.google.firebase.perf.config.ConfigurationConstants.CollectionEnabled;
 import com.google.firebase.perf.config.ConfigurationConstants.FragmentSamplingRate;
+import com.google.firebase.perf.config.ConfigurationConstants.ExperimentTTID;
 import com.google.firebase.perf.config.ConfigurationConstants.LogSourceName;
 import com.google.firebase.perf.config.ConfigurationConstants.NetworkEventCountBackground;
 import com.google.firebase.perf.config.ConfigurationConstants.NetworkEventCountForeground;
@@ -760,6 +761,38 @@ public class ConfigResolver {
     // 3. Reads value from cache layer.
     Optional<Float> deviceCacheValue = getDeviceCacheFloat(config);
     if (deviceCacheValue.isAvailable() && isSamplingRateValid(deviceCacheValue.get())) {
+      return deviceCacheValue.get();
+    }
+
+    // 4. Returns default value if there is no valid value from above approaches.
+    return config.getDefault();
+  }
+
+  /** Returns if _experiment_as_ttid should be captured. */
+  public boolean getIsExperimentTTIDEnabled() {
+    // Order of precedence is:
+    // 1. If the value exists in Android Manifest, return this value.
+    // 2. If the value exists through Firebase Remote Config, cache and return this value.
+    // 3. If the value exists in device cache, return this value.
+    // 4. Otherwise, return default value.
+    ExperimentTTID config = ExperimentTTID.getInstance();
+
+    // 1. Reads value in Android Manifest (it is set by developers during build time).
+    Optional<Boolean> metadataValue = getMetadataBoolean(config);
+    if (metadataValue.isAvailable()) {
+      return metadataValue.get();
+    }
+
+    // 2. Reads value from Firebase Remote Config, saves this value in cache layer if valid.
+    Optional<Boolean> rcValue = getRemoteConfigBoolean(config);
+    if (rcValue.isAvailable()) {
+      deviceCacheManager.setValue(config.getDeviceCacheFlag(), rcValue.get());
+      return rcValue.get();
+    }
+
+    // 3. Reads value from cache layer.
+    Optional<Boolean> deviceCacheValue = getDeviceCacheBoolean(config);
+    if (deviceCacheValue.isAvailable()) {
       return deviceCacheValue.get();
     }
 
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
@@ -772,7 +772,7 @@ public class ConfigResolver {
   public boolean getIsExperimentTTIDEnabled() {
     // Order of precedence is:
     // 1. If the value exists in Android Manifest, return this value.
-    // 2. Cannot read value from RC because it's not initalized yet
+    // 2. If the value exists through Firebase Remote Config, cache and return this value.
     // 3. If the value exists in device cache, return this value.
     // 4. Otherwise, return default value.
     ExperimentTTID config = ExperimentTTID.getInstance();
@@ -783,7 +783,13 @@ public class ConfigResolver {
       return metadataValue.get();
     }
 
-    // 2. Cannot read value from RC because it's not initialized yet.
+    // 2. Reads value from Firebase Remote Config, saves this value in cache layer if valid.
+    Optional<Boolean> rcValue = getRemoteConfigBoolean(config);
+    if (rcValue.isAvailable()) {
+      deviceCacheManager.setValue(config.getDeviceCacheFlag(), rcValue.get());
+      return rcValue.get();
+    }
+
     // 3. Reads value from cache layer.
     Optional<Boolean> deviceCacheValue = getDeviceCacheBoolean(config);
     if (deviceCacheValue.isAvailable()) {
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
@@ -23,8 +23,8 @@ import androidx.annotation.VisibleForTesting;
 import com.google.firebase.perf.BuildConfig;
 import com.google.firebase.perf.config.ConfigurationConstants.CollectionDeactivated;
 import com.google.firebase.perf.config.ConfigurationConstants.CollectionEnabled;
-import com.google.firebase.perf.config.ConfigurationConstants.FragmentSamplingRate;
 import com.google.firebase.perf.config.ConfigurationConstants.ExperimentTTID;
+import com.google.firebase.perf.config.ConfigurationConstants.FragmentSamplingRate;
 import com.google.firebase.perf.config.ConfigurationConstants.LogSourceName;
 import com.google.firebase.perf.config.ConfigurationConstants.NetworkEventCountBackground;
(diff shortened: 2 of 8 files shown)


firebase-perf/dev-app/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,9 @@
         <meta-data
             android:name="fragment_sampling_percentage"
             android:value="100.0" />
+        <meta-data
+            android:name="experiment_as_ttid"
```
raymondlam, line 42: I'd prefer experiment_app_start_ttid so that it is more obvious that it is for app start.

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigurationConstants.java
```diff
@@ -661,4 +661,39 @@ protected String getMetadataFlag() {
...
+    }
+
+    @Override
+    protected Boolean getDefault() {
+      return false;
+    }
+
+    @Override
+    protected String getRemoteConfigFlag() {
+      return "fpr_experiment_as_ttid";
```
raymondlam, line 686: Similarly, `as` should be `app_start` here.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -227,11 +272,24 @@ public synchronized void onActivityResumed(Activity activity) {
 
     // Log the app start trace in a non-main thread.
     executorService.execute(this::logAppStartTrace);
+  }
 
-    if (isRegisteredForLifecycleCallbacks) {
-      // After AppStart trace is logged, we can unregister this callback.
-      unregisterActivityLifecycleCallbacks();
```
raymondlam, line 233: Are we still unregistering somewhere if the the experiment is disabled and recordFirstDrawDone never gets called?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -227,11 +272,24 @@ public synchronized void onActivityResumed(Activity activity) {
...
+  private void logColdStart(Timer start, Timer end, PerfSession session) {
+    TraceMetric.Builder metric =
+        TraceMetric.newBuilder()
+            .setName("_experiment_as_ttid")
+            .setClientStartTimeUs(start.getMicros())
+            .setDurationUs(start.getDurationMicros(end));
+
+    TraceMetric.Builder subtrace =
+        TraceMetric.newBuilder()
+            .setName("_experiment_classLoadTime")
```
raymondlam, line 286: What's the purpose of this sub trace?

firebase-perf/src/test/java/com/google/firebase/perf/metrics/AppStartTraceTest.java
```diff
@@ -261,4 +279,38 @@ public void testFirebasePerfProviderOnAttachInfo_initializesGaugeCollection() {
     Assert.assertEquals(oldSessionId, SessionManager.getInstance().perfSession().sessionId());
     verify(mockPerfSession, times(2)).isGaugeAndEventCollectionEnabled();
   }
+
+  @Test
+  @Config(sdk = 26)
+  public void timeToInitialDisplay_isLogged() {
```
raymondlam, line 285: Can you add some comments to clarify what is going on in this test? There is a lot of setup and some verify statements mid way through the test making it hard to follow.

firebase-perf/dev-app/src/main/AndroidManifest.xml
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigurationConstants.java
leotianlizhan: done

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: Great catch! Added it back.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: This captures the old-start, which is class-load of `FirebasePerfProvider.class`. I just want to have more data to study the difference that the new start would have.

firebase-perf/src/test/java/com/google/firebase/perf/metrics/AppStartTraceTest.java
leotianlizhan: Added comments! The setup is quite big, only the last few lines of this test is actually verifying.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -228,12 +274,30 @@ public synchronized void onActivityResumed(Activity activity) {
...
+    if (!isExperimentTTIDEnabled && isRegisteredForLifecycleCallbacks) {
       // After AppStart trace is logged, we can unregister this callback.
       unregisterActivityLifecycleCallbacks();
     }
   }
 
+  private void logColdStart(Timer start, Timer end, PerfSession session) {
+    TraceMetric.Builder metric =
+        TraceMetric.newBuilder()
+            .setName("_experiment_app_start_ttid")
```
jeremyjiang-dev, line 286: nit: define the name as a constant Constants.TraceNames.

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
```diff
@@ -767,6 +768,32 @@ public float getFragmentSamplingRate() {
...
+    // 4. Otherwise, return default value.
+    ExperimentTTID config = ExperimentTTID.getInstance();
+
+    // 1. Reads value in Android Manifest (it is set by developers during build time).
+    Optional<Boolean> metadataValue = getMetadataBoolean(config);
+    if (metadataValue.isAvailable()) {
+      return metadataValue.get();
+    }
+
+    // 2. Cannot read value from RC because it's not initialized yet.
```
jeremyjiang-dev, line 786: Just to make sure my understanding is correct: for the very first app start of an app, we will read the default value.

firebase-perf/dev-app/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,9 @@
         <meta-data
             android:name="fragment_sampling_percentage"
             android:value="100.0" />
+        <meta-data
```
jeremyjiang-dev, line 41: Can you also update the e2e test app so that it will generate this trace periodically? but I think our e2e test app does not generate app start trace. If so, it is out of the scope of this PR.

jeremyjiang-dev (approved):
Looks good to me overall

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
leotianlizhan: Yes, for the very very first without any prior RC fetch, it will use default value.

firebase-perf/dev-app/src/main/AndroidManifest.xml
leotianlizhan: Checked just now and you are right, our E2E test app doesn't generate `_app_start` at all.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: This code is for dark-launch, it will be removed next release, so I didn't add a new official constant to Constants.java. I will definitely do it for the GA launch.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -228,12 +274,30 @@ public synchronized void onActivityResumed(Activity activity) {
     // Log the app start trace in a non-main thread.
     executorService.execute(this::logAppStartTrace);
 
-    if (isRegisteredForLifecycleCallbacks) {
+    if (!isExperimentTTIDEnabled && isRegisteredForLifecycleCallbacks) {
```
visumickey, line 277: What happens if the flag dynamically changes from No to YES (or) vice versa? Would that cause a situation where we try to unregister for callbacks when we have not registered or the other way about?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -228,12 +274,30 @@ public synchronized void onActivityResumed(Activity activity) {
...
+
+    TraceMetric.Builder subtrace =
+        TraceMetric.newBuilder()
+            .setName("_experiment_classLoadTime")
+            .setClientStartTimeUs(FirebasePerfProvider.getAppStartTime().getMicros())
+            .setDurationUs(FirebasePerfProvider.getAppStartTime().getDurationMicros(end));
+
+    metric.addSubtraces(subtrace).addPerfSessions(this.startSession.build());
+
+    transportManager.log(metric.build(), ApplicationProcessState.FOREGROUND_BACKGROUND);
```
visumickey, line 298: Why FOREGROUND_BACKGROUND?

firebase-perf/dev-app/src/main/AndroidManifest.xml
jeremyjiang-dev: IIRC the test skips the onResume() callback

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
raymondlam: sg

raymondlam (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: `_app_start` is `FOREGROUND_BACKGROUND` so I am just following the same behavior.

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
```diff
@@ -767,6 +768,32 @@ public float getFragmentSamplingRate() {
     return config.getDefault();
   }
 
+  /** Returns if _experiment_app_start_ttid should be captured. */
+  public boolean getIsExperimentTTIDEnabled() {
+    // Order of precedence is:
+    // 1. If the value exists in Android Manifest, return this value.
+    // 2. Cannot read value from RC because it's not initalized yet
```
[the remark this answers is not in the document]
leotianlizhan: This is a comment in the same style as other flags if you look into the same file. Other flags has the order laid out like this, I just changed the 2nd step to "Cannot read from RC".

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
```diff
@@ -767,6 +768,32 @@ public float getFragmentSamplingRate() {
     return config.getDefault();
   }
 
+  /** Returns if _experiment_app_start_ttid should be captured. */
+  public boolean getIsExperimentTTIDEnabled() {
+    // Order of precedence is:
```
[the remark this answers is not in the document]
leotianlizhan: This is just Manifest > Device Cache > SDK Default

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: Good question! Since this flag is only read once in the beginning of `AppStartTrace#onActivityResumed`, it's not a problem.
> Scenario 1: experiment disabled, an activity resumed, experiment enabled, another activity resumed.

After the first resume, the unregistering would be called, so the 2nd activity resume would not do anything. TTID isn't recorded for this app start, it will only start working next app start.

> Scenario 2: experiment enabled, an activity resumed, experiment enabled, another activity resumed.

After the first resume, the `AppStartTrace.onResume` variable would be set, so even if the 2nd activity resume triggers, nothing happens because this method will return immediately without sending anything else.

firebase-perf/src/test/java/com/google/firebase/perf/config/RemoteConfigManagerTest.java
```diff
@@ -230,6 +240,29 @@ public void getString_validFrcValue_updatesWithNewValue() {
     assertThat(testRemoteConfigManager.getString("some_key3").isAvailable()).isFalse();
   }
 
+  @Test
+  public void syncConfigValues_savesNewlyFetchedValueToDeviceCache() {
```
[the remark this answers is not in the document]
leotianlizhan: It is already done in this test. `syncConfigValues` simulates fetch success (it is only called after fetch success), so in this test we call `syncConfigValues` twice and verifies that the disk cache value has changed.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -342,6 +341,19 @@ protected void syncConfigValues(Map<String, FirebaseRemoteConfigValue> newlyFetc
         allRcConfigMap.remove(existingKey);
       }
     }
+
+    // Save to device cache upon successful RC fetchAndActivate
```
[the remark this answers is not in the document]
leotianlizhan: Done

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -315,7 +314,7 @@ private void triggerRemoteConfigFetchIfNecessary() {
       return;
     }
     if (allRcConfigMap.isEmpty()) { // Initial fetch.
-      syncConfigValues(firebaseRemoteConfig.getAll());
+      allRcConfigMap.putAll(firebaseRemoteConfig.getAll());
```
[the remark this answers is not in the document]
visumickey: Sounds good. That makes sense.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
visumickey: Sounds good.

For the first scenario - I assume that experiment is disabled?

In the second scenario - I assume that would mean experiment is enabled?

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: Yes, these scenarios show what happens if the app starts with 1 state, then changes after app start.

leotianlizhan:
`Health Metrics / Coverage (push)` fails due to firestorage test failures, so we can merge.

PR #4122, opened 2022-09-20
Add a screenshot-based trigger for in-app feedback

lfkellogg:
Also has some refactoring to, among other things, allow the passing of a screenshot URI to the `startFeedback()` API.

### DIFF
--- a/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
+++ b/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
@@ -16,7 +16,6 @@ package com.google.firebase.appdistribution.internal;
 
 import android.app.Activity;
 import android.net.Uri;
-
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.android.gms.tasks.Continuation;
@@ -86,7 +85,12 @@ public class FirebaseAppDistributionStub implements FirebaseAppDistribution {
   }
 
   @Override
-  public void startFeedback(@NonNull CharSequence infoText, @NonNull Uri screenshotUri) {
+  public void startFeedback(@NonNull int infoTextResourceId, @Nullable Uri screenshotUri) {
+    return;
+  }
+
+  @Override
+  public void startFeedback(@NonNull CharSequence infoText, @Nullable Uri screenshotUri) {
     return;
   }
 
--- a/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
+++ b/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
@@ -15,6 +15,8 @@
 package com.google.firebase.appdistribution.internal;
 
 import android.app.Activity;
+import android.net.Uri;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.android.gms.tasks.Continuation;
@@ -80,6 +82,11 @@ public class FirebaseAppDistributionStub implements FirebaseAppDistribution {
 
   @Override
   public void startFeedback(@NonNull CharSequence infoText) {
+    return;
+  }
+
+  @Override
+  public void startFeedback(@NonNull CharSequence infoText, @NonNull Uri screenshotUri) {
     return;
   }
 
--- a/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
+++ b/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
@@ -58,9 +58,7 @@ public class FirebaseAppDistributionStub implements FirebaseAppDistribution {
   }
 
   @Override
-  public void signOutTester() {
-    return;
-  }
+  public void signOutTester() {}
 
   @NonNull
   @Override
@@ -75,24 +73,16 @@ public class FirebaseAppDistributionStub implements FirebaseAppDistribution {
   }
 
   @Override
-  public void startFeedback(int infoTextResourceId) {
-    return;
-  }
+  public void startFeedback(int infoTextResourceId) {}
 
   @Override
-  public void startFeedback(@NonNull CharSequence infoText) {
-    return;
-  }
+  public void startFeedback(@NonNull CharSequence infoText) {}
 
   @Override
-  public void startFeedback(@NonNull int infoTextResourceId, @Nullable Uri screenshotUri) {
-    return;
-  }
+  public void startFeedback(@NonNull int infoTextResourceId, @Nullable Uri screenshotUri) {}
 
   @Override
-  public void startFeedback(@NonNull CharSequence infoText, @Nullable Uri screenshotUri) {
-    return;
-  }
+  public void startFeedback(@NonNull CharSequence infoText, @Nullable Uri screenshotUri) {}
 
   private static <TResult> Task<TResult> getNotImplementedTask() {
     return Tasks.forException(
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
@@ -26,6 +26,7 @@ import android.widget.TextView;
 import android.widget.Toast;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
+import java.io.IOException;
 
 /** Activity for tester to compose and submit feedback. */
 public class FeedbackActivity extends AppCompatActivity {
@@ -67,7 +68,7 @@ public class FeedbackActivity extends AppCompatActivity {
     Button submitButton = this.findViewById(R.id.submitButton);
     submitButton.setOnClickListener(this::submitFeedback);
 
-    Bitmap thumbnail = readThumbnail();
+    Bitmap thumbnail = screenshotUri == null ? null : readThumbnail();
     if (thumbnail != null) {
       ImageView screenshotImageView = this.findViewById(R.id.thumbnail);
       screenshotImageView.setImageBitmap(thumbnail);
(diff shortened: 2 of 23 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -67,29 +71,30 @@ private void setupView() {
...
     submitButton.setOnClickListener(this::submitFeedback);
 
-    Bitmap thumbnail = readThumbnail();
-    if (thumbnail != null) {
+    try {
+      Bitmap thumbnail = readThumbnail();
       ImageView screenshotImageView = this.findViewById(R.id.thumbnail);
       screenshotImageView.setImageBitmap(thumbnail);
-    } else {
+    } catch (FirebaseAppDistributionException e) {
```
kaibolay, line 78: Instead of using exceptions for flow control, you could have `readThumbnail()` return `null` instead of throwing an exception if there's no screenshot available.

But even simpler: Just check whether `screenshotUri` is null before trying to read the thumbnail

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ImageUtils.java
```diff
@@ -14,51 +14,78 @@
...
+   * @throws FirebaseAppDistributionException if the image could not be read
    */
-  @Nullable
-  public static Bitmap readScaledImage(File file, int targetWidth, int targetHeight) {
+  public static Bitmap readScaledImage(
+      ContentResolver contentResolver, Uri uri, int targetWidth, int targetHeight)
+      throws FirebaseAppDistributionException {
     if (targetWidth <= 0 || targetHeight <= 0) {
-      throw new IllegalArgumentException(
+      throw new FirebaseAppDistributionException(
```
kaibolay, line 62: I think this can remain an `IllegalArgumentException` - it's clearly a programmer error if negative (or zero) values are passed in. I don't think we want to even catch the problem.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/FeedbackTriggers.java
```diff
@@ -0,0 +1,138 @@
...
+
+  private static final String TAG = "FeedbackTriggers";
+  private static final boolean SHOULD_CHECK_IF_PENDING = Build.VERSION.SDK_INT == 29;
+  private static final String[] PROJECTION =
+      SHOULD_CHECK_IF_PENDING
+          ? new String[] {MediaStore.Images.Media.DATA}
+          : new String[] {
+            MediaStore.Images.Media.DATA, android.provider.MediaStore.MediaColumns.IS_PENDING
+          };
+  private final Set<String> seenImages = new HashSet<>();
```
kaibolay, line 48: nit: Set<Uri> ?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/FeedbackTriggers.java
```diff
@@ -0,0 +1,138 @@
...
+                        })
+                    .show();
+              }
+            });
+  }
+
+  void registerScreenshotObserver() {
+    context
+        .getContentResolver()
+        .registerContentObserver(Media.EXTERNAL_CONTENT_URI, true /*notifyForDescendants*/, this);
```
kaibolay, line 94: supernit: I'm partial to the comment coming first with a `=`: `/* notifyForDescendants= */ true`

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -58,10 +62,22 @@ class MainActivity : AppCompatActivity() {
         signInStatus = findViewById<TextView>(R.id.sign_in_status)
         progressPercent = findViewById<TextView>(R.id.progress_percentage)
         progressBar = findViewById<ProgressBar>(R.id.progress_bar)
+
+
+        val thread = HandlerThread("AppDistroFeedbackTrigger")
+        thread.start()
```
kaibolay, line 68: should the be stopped when the activity is destroyed?

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/internal/FirebaseAppDistributionStub.java
```diff
@@ -83,6 +84,16 @@ public void startFeedback(@NonNull CharSequence infoText) {
     return;
   }
 
+  @Override
+  public void startFeedback(@NonNull int infoTextResourceId, @Nullable Uri screenshotUri) {
+    return;
```
kaibolay, line 89: nit: `return` is redundant. Maybe remove throughout this class?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiHttpClient.java
```diff
@@ -123,13 +124,15 @@ JSONObject makePostRequest(String tag, String path, String token, String request
...
-  JSONObject makeUploadRequest(String tag, String path, String token, InputStream inputStream)
+  JSONObject makeUploadRequest(String tag, String path, String token, Uri contentUri)
       throws FirebaseAppDistributionException {
     Map<String, String> extraHeaders = new HashMap<>();
     extraHeaders.put(X_GOOG_UPLOAD_PROTOCOL_HEADER, X_GOOG_UPLOAD_PROTOCOL_RAW);
     extraHeaders.put(X_GOOG_UPLOAD_FILE_NAME_HEADER, X_GOOG_UPLOAD_FILE_NAME);
     RequestBodyWriter requestBodyWriter =
         outputStream -> {
+          InputStream inputStream =
+              firebaseApp.getApplicationContext().getContentResolver().openInputStream(contentUri);
```
kaibolay, line 135: try-with-resources

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TesterApiHttpClientTest.java
```diff
@@ -254,13 +256,15 @@ public void makeUploadRequest_writesRequestBodyAndSetsCorrectHeaders() throws Ex
...
-        ApplicationProvider.getApplicationContext().getFileStreamPath("requestBody.txt");
-    try (FileWriter writer = new FileWriter(testRequestBodyFile)) {
-      writer.write("Test post body");
-    }
 
-    testerApiHttpClient.makeUploadRequest(TAG, TEST_PATH, TEST_AUTH_TOKEN, testRequestBodyFile);
+    ContentResolver contentResolver =
+        ApplicationProvider.getApplicationContext().getContentResolver();
+    ShadowContentResolver shadowContentResolver = shadowOf(contentResolver);
+    ByteArrayInputStream inputStream = new ByteArrayInputStream("Test post body".getBytes(UTF_8));
```
kaibolay, line 263: nit: try-with-resources (it's not really necessary to close a ByteArrayInputStream - especially in tests - but I try to make sure to always close stream - everywhere to avoid accidental resource leaks. I think it's a worthwhile habit.

kaibolay (approved):
Feel free to merge with remaining TODOs - they can be addressed in a follow-up

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/FeedbackTriggers.java
```diff
@@ -0,0 +1,138 @@
...
+import android.util.Log;
+import androidx.activity.ComponentActivity;
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.contract.ActivityResultContracts;
+import androidx.core.content.ContextCompat;
+import com.google.firebase.appdistribution.FirebaseAppDistribution;
+import java.util.HashSet;
+import java.util.Set;
+
+class FeedbackTriggers extends ContentObserver {
```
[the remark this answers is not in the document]
lfkellogg: Done

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TesterApiHttpClientTest.java
```diff
@@ -251,28 +251,33 @@ public void makePostRequest_whenConnectionFails_throwsError() throws Exception {
...
-
-    assertThat(new String(requestBodyOutputStream.toByteArray(), UTF_8))
-        .isEqualTo("Test post body");
-    verify(mockHttpsURLConnection).setDoOutput(true);
-    verify(mockHttpsURLConnection).setRequestMethod("POST");
-    verify(mockHttpsURLConnection).addRequestProperty("Content-Type", "application/json");
-    verify(mockHttpsURLConnection).addRequestProperty("X-Goog-Upload-Protocol", "raw");
-    verify(mockHttpsURLConnection).addRequestProperty("X-Goog-Upload-File-Name", "screenshot.png");
-    verify(mockHttpsURLConnection).disconnect();
+    // Setup test streams. They should be closed but close them anyway just to be safe.
```
kaibolay, line 255: I think the comment is missing some words. Maybe entirely drop the comment entirely or explain that since these are in-memory streams they just get garbage collected and don't strictly need to be closed to avoid resource leaks.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -63,20 +65,29 @@ class MainActivity : AppCompatActivity() {
         progressPercent = findViewById<TextView>(R.id.progress_percentage)
         progressBar = findViewById<ProgressBar>(R.id.progress_bar)
 
+        screenshotTriggerThread = HandlerThread("AppDistroFeedbackTrigger")
+        screenshotTriggerThread.start()
+        screenshotTrigger =
+            ScreenshotDetectionFeedbackTrigger(
+                this,
+                "Here's some terms and conditions",
```
kaibolay, line 73: this should be from the `string` resource file, right?

kaibolay (approved):
Nice. Thanks!

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TesterApiHttpClientTest.java
lfkellogg: Got it. Removed the comment.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
lfkellogg: Yeah probably better to use a resource

PR #4124, opened 2022-09-21
Make FiresiteTransform a Cacheable task

daymxn:
In an effort to make the DackkaPlugin tasks all cache compliant (per [b/243833009](https://b.corp.google.com/issues/243833009)), the only remaining task that was not already cache complaint was FiresiteTransform. This PR fixes that.

More specifically, we now copy the transformed files into a new directory labeled `dackkaTransformedFiles`, versus working on the same files that Dackka outputs. This allows gradle to check if the files have actually changed, and pass an `UP-TO-DATE-CHECK` as needed.

Build time has dropped from ~50 seconds to only ~15 seconds, that's a 333% decrease!

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/FiresiteTransformTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/FiresiteTransformTask.kt
@@ -3,7 +3,11 @@ package com.google.firebase.gradle.plugins
 import java.io.File
 import org.gradle.api.DefaultTask
 import org.gradle.api.provider.Property
+import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.TaskAction
 
 /**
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/FiresiteTransformTask.kt
```diff
@@ -21,10 +25,15 @@ import org.gradle.api.tasks.TaskAction
  *  **Please note:**
  *  This task is idempotent- meaning it can safely be ran multiple times on the same set of files.
  */
+@CacheableTask
```
vkryachko, line 28: Interesting, aren't tasks cacheable by default?

vkryachko (approved):

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FiresiteTransformTask.kt
daymxn: Yeah, but having this explicit annotation forces you to label the annotations properly. For example, without this annotation, you could compile without the `PathSensitive` attribute (which helps solidify *how* the task should be looked at and cached).

vkryachko: T.I.L. 👍

PR #4126, opened 2022-09-22
Make count queries publicly available for use

dconeybe:
This PR changes the visibility of classes and methods that provide "count" queries from default to public, making them available for anyone to use. See #3847 for the implementation of COUNT queries.


dconeybe:
https://github.com/firebase/firebase-android-sdk/pull/4130

PR #4127, opened 2022-09-22
Firestore: Fix COUNT APIs to match the API proposal

dconeybe:
This is a follow-up to #3847 to modify the API surface of the new Query.count() feature to match the API proposal go/firestore-count-api-client-proposal.


dconeybe:
Closing in favor of https://github.com/firebase/firebase-android-sdk/pull/4130

PR #3664, opened 2022-04-19
Tomandersen/transaction options

tom-andersen:
Add TransactionOptions to allow control of maximum attempts to commit before transaction fails.

PR #4130, opened 2022-09-22
Public Count

wu-hui:


dconeybe (approved):
LGTM

dconeybe:
Do you need to update `firebase-firestore/api.txt`, like #3664 did for TransactionOptions?

wu-hui:
> Do you need to update `firebase-firestore/api.txt`, like #3664 did for TransactionOptions?

Yep, thanks for the catch.

dconeybe:
Add a change log entry to `firebase-firestore/CHANGELOG.md`

e.g.
```
- [feature] Added `Query.count()`, which fetches the number of documents in the
  result set without actually downloading the documents (#4130).
```

dconeybe (changes requested):
Add change log entry

dconeybe (approved):

PR #4138, opened 2022-09-22
Add shake for feedback to test app

lfkellogg:


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
```diff
@@ -0,0 +1,10 @@
+package com.googletest.firebase.appdistribution.testapp
+
+import android.app.Application
+
+class AppDistroTestApplication : Application() {
+    override fun onCreate() {
```
kaibolay, line 6: Unrelated to this PR, but should we move more initialization code from `MainActiviry.onCreate()` to here?

Also, I'm curious: Would this also work if we added it to `MainActivity.onCreate()`?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ShakeForFeedback.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.util.Log
+import com.google.firebase.appdistribution.ktx.appDistribution
+import com.google.firebase.ktx.Firebase
+import com.squareup.seismic.ShakeDetector
+
+class ShakeForFeedback private constructor() : ShakeDetector.Listener,
+    Application.ActivityLifecycleCallbacks {
+    private val shakeDetector = ShakeDetector(this)
+
+    override fun hearShake() {
```
kaibolay, line 16: "hear" - interesting terminology, but I guess that's what Square picked...

kaibolay (approved):

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ShakeForFeedback.kt
lfkellogg: Yeah true, you'd think it would be "feelShake" :)

Maybe because it's a "Listener"?

kaibolay: 🤯

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
kaibolay: Ah, makes sense. Let's explore moving the initialization code here in a separate PR once this one is merged.

Also, I think we should restore the button to launch the second activity so we can test those scenarios. Maybe you can do that when adding the radio buttons.

kaibolay (approved):

PR #4143, opened 2022-09-24
Firestore: Re-write API javadocs for COUNT API

dconeybe:


firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
```diff
@@ -1224,10 +1224,17 @@ private void validateHasExplicitOrderByForLimitToLast() {
...
 
   /**
-   * Creates an {@link AggregateQuery} counting the number of documents matching this query.
+   * Returns a query that counts the documents in the result set of this query.
    *
-   * @return An {@link AggregateQuery} object that can be used to count the number of documents in
-   *     the result set of this query.
+   * <p>The returned query, when executed, counts the documents in the result set of this query
+   * <em>without actually downloading the documents</em>. The count is performed on the server and
+   * only the resulting count is downloaded.
```
ehsannas, line 1231: I think this sentence should be removed.. `The count is performed on the server and only the resulting count is downloaded.`

Depending on the SOURCE that is used for `get` call, this may not be true in the future, right?

ehsannas (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/Query.java
dconeybe: Removed. I was trying to figure out how to word it to convey its efficiency. But I think it's probably fine without this staement.

firebase-firestore/src/main/java/com/google/firebase/firestore/AggregateQuery.java
```diff
@@ -70,14 +69,35 @@ public Task<AggregateQuerySnapshot> get(@NonNull AggregateSource source) {
     return tcs.getTask();
   }
 
+  /**
+   * Compares this object with the given object for equality.
+   *
+   * <p>This object is considered "equal" to the other object if and only if all of the following
```
markarndt, line 75: Please double-check whether closing </p> and </li> tags are required here and a few lines below, and in corresponding docstrings in other source files.

dconeybe: I originally had the closings tags in there, but the code formatter actually removed them!

PR #4149, opened 2022-09-27
Specify unique ref tags in Dackka output

daymxn:
In accordance with [b/247570655](https://b.corp.google.com/issues/247570655), this PR extends `FiresiteTransform` to offer configuration for the generated `_reference-head-tags.html` at the top of html files.

We would like to offer a unique ref html file for each set of docs (Java & Kotlin), for SEO purposes etc.,

To accomplish this, without drastically changing the behavior of `FiresiteTransform` - we now expose a task to separate the Java and Kotlin generated sets before transformation- and then combine them after the fact. This required a bit of overhead to accomplish, but will pay off in the long run whenever we decide to combine the output directory of Java and Kotlin generated sets.

To keep things as idiomatic as possible, I started defining the desired output directories at the root of `kotlindoc`'s configuration, and offer them as inputs to the relevant tasks.

There was also a lot of repeat in the creation of sub directories and files- which I split off into extension methods;

`project.fileFromBuildDir()` -> creates a `File` at the project's builddir

`project.childFile()` -> maps a file provider to another file provider as a sub directory

Furthermore, I found some artifacts that don't actually generate any documentation (everything is either `@hidden` or private), and marked them as such as to avoid needing an empty dir check on `DackkaPlugin` tasks- as that could mask potential edge cases that we'd otherwise like to pay attention to.

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
@@ -92,8 +92,9 @@ tasks above). While we do not currently offer any configuration for the Dackka
 plugin, this could change in the future as needed. Currently, the DackkaPlugin
 provides sensible defaults to output directories, package lists, and so forth.
 
-The DackkaPlugin also provides three extra tasks:
+The DackkaPlugin also provides four extra tasks:
 [cleanDackkaDocumentation][registerCleanDackkaDocumentation],
+[separateJavadocAndKotlinDoc][registerSeparateJavadocAndKotlinDoc]
 [copyJavaDocToCommonDirectory][registerCopyJavaDocToCommonDirectoryTask] and
 [copyKotlinDocToCommonDirectory][registerCopyKotlinDocToCommonDirectoryTask].
 
@@ -101,6 +102,9 @@ _cleanDackkaDocumentation_ is exactly what it sounds like, a task to clean up (d
 the output of Dackka. This is useful when testing Dackka outputs itself- and
 shouldn't be apart of the normal flow. The reasoning is that it would otherwise
 invalidate the gradle cache.
+
+_separateJavadocAndKotlinDoc_ copies the Javadoc and Kotlindoc directories from Dackka into
+their own subdirectories- for easy and consistent differentiation.
 
 _copyJavaDocToCommonDirectory_ copies the JavaDoc variant of the Dackka output for each sdk,
 and pastes it in a common directory under the root project's build directory. This makes it easier
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
@@ -298,7 +298,7 @@ abstract class DackkaPlugin : Plugin<Project> {
         val transformJavadoc = project.tasks.register<FiresiteTransformTask>("firesiteTransformJavadoc") {
             dependsOnAndMustRunAfter("separateJavadoc")
 
-            referenceHeadTagsPath.set("google3/third_party/devsite/firebase/en/docs/reference/android")
+            referenceHeadTagsPath.set("en/docs/reference/android")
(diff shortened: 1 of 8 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -232,20 +249,72 @@ abstract class DackkaPlugin : Plugin<Project> {
...
+
+    private fun registerFiresiteTransformTask(
+        project: Project,
+        separatedFilesDirectory: Provider<File>,
+        targetDirectory: Provider<File>
+    ): TaskProvider<Task> {
+        val transformJavadoc = project.tasks.register<FiresiteTransformTask>("firesiteTransformJavadoc") {
+            dependsOnAndMustRunAfter("separateJavadoc")
+
+            referenceHeadTagsPath.set("google3/third_party/devsite/firebase/en/docs/reference/android")
```
vkryachko, line 301:
```suggestion
            referenceHeadTagsPath.set("en/docs/reference/android")
```
hm, I don't think we need the absolute path, but rather relative to the tenant(i.e. firebase)

buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaPlugin.kt
```diff
@@ -232,20 +249,72 @@ abstract class DackkaPlugin : Plugin<Project> {
...
+
+    private fun registerFiresiteTransformTask(
+        project: Project,
+        separatedFilesDirectory: Provider<File>,
+        targetDirectory: Provider<File>
+    ): TaskProvider<Task> {
+        val transformJavadoc = project.tasks.register<FiresiteTransformTask>("firesiteTransformJavadoc") {
+            dependsOnAndMustRunAfter("separateJavadoc")
+
+            referenceHeadTagsPath.set("en/docs/reference/android")
```
vkryachko, line 301: hm, on second thought I don't think even `en/` is needed. Actually can you confirm with @rachelsaunders ?
```suggestion
            referenceHeadTagsPath.set("docs/reference/android")
```

daymxn: Also left her a chat message about this

rachelsaunders: You should be able to use a relative reference and go all the way down to `/android/_reference-head-tags.html` and `/kotlin/_reference-head-tags.html`.

vkryachko: Hm, doesn't `/android` imply "relative to the root", whatever the root is. So if we for example have a file `docs/reference/kotlin/com/google/firebase/Blah.html` with such an include what will it resolve to?

daymxn: Solved. The `root` is `google3/third_party/devsite/firebase/en/`, which makes sense as that's where `_book.yaml` and `_shared` live. I've corrected the path link as such.

vkryachko:
I am wondering what is up with build-src-check tests?

daymxn:
> I am wondering what is up with build-src-check tests?

`publishWithJavadoc` is enabled by default, which works in prod for our sdks- but when tests aren't testing javadocs- it can be an issue. Primarily because our logic expects there to be sources to publish if your `publishJavadoc` is enabled. Otherwise, it throws an error (as it should).

I just pushed a commit that disables it by default for our tests, and exposes a similar configuration in the constructor params to change it.

vkryachko (approved):
👍 

While we are talking about tests, we would create a task to write some minimal smoke tests for kotlindoc and javadoc generation as part of `build-src-check`

daymxn:
> 👍
> 
> While we are talking about tests, we should create a task to write some minimal smoke tests for kotlindoc and javadoc generation as part of `build-src-check`

Agreed and done: [b/249825782](https://b.corp.google.com/issues/249825782)

PR #4150, opened 2022-09-28
Add menu to test app for selecting feedback trigger

lfkellogg:
Also:
- Move the regular feedback button trigger to an action bar menu
- Clean up some of the UX code (removing unnecessary type params, making padding uniform)
- Turn ShakeForFeedback from a singleton class into an object

### DIFF
--- a/firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
+++ b/firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
@@ -5,6 +5,8 @@ import android.app.Application
 class AppDistroTestApplication : Application() {
     override fun onCreate() {
         super.onCreate()
-        ShakeForFeedback.enable(this)
+
+        // Default feedback triggers can also be initialized here
+//        ShakeForFeedback.enable(this)
     }
 }
--- a/firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
+++ b/firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
@@ -17,8 +17,6 @@ import android.widget.ProgressBar
 import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.AppCompatButton
-import androidx.core.content.ContentProviderCompat.requireContext
-import androidx.core.widget.addTextChangedListener
 import androidx.core.widget.doOnTextChanged
 import com.google.android.gms.tasks.Task
 import com.google.android.material.textfield.TextInputLayout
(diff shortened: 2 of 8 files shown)


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
```diff
@@ -5,6 +5,8 @@ import android.app.Application
 class AppDistroTestApplication : Application() {
     override fun onCreate() {
         super.onCreate()
-        ShakeForFeedback.enable(this)
+
+        // Default feedback triggers can also be initialized here
+//        ShakeForFeedback.enable(this)
```
kaibolay, line 10: Remove this line? Or add commented-out code for the `ScreenshotDetectionFeedbackTrigger` as well?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -264,4 +307,12 @@ class MainActivity : AppCompatActivity() {
         updateAppButton.visibility = View.GONE
         updateAppButtonBackground.visibility = View.GONE
     }
+
+    companion object {
+        const val TAG = "MainActivity"
+        const val TRIGGER_NONE = "None"
+        const val TRIGGER_SHAKE = "Shake the device"
+        const val TRIGGER_SCREENSHOT = "Take a screenshot"
+        const val TRIGGER_NOTIFICATION = "Click the notification"
```
kaibolay, line 316: I think the four triggers should probably be an enum

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ShakeForFeedback.kt
```diff
@@ -9,9 +9,46 @@ import com.google.firebase.appdistribution.ktx.appDistribution
...
 
-class ShakeForFeedback private constructor() : ShakeDetector.Listener,
-    Application.ActivityLifecycleCallbacks {
+object ShakeForFeedback : ShakeDetector.Listener, Application.ActivityLifecycleCallbacks {
+    const val TAG: String = "ShakeForFeedback"
+
     private val shakeDetector = ShakeDetector(this)
+    private var isEnabled = false
+
+    fun enable(application: Application) = enable(application, null)
```
kaibolay, line 18: I think you don't need this overload by using [default arguments](https://kotlinlang.org/docs/functions.html#default-arguments) - i.e. `null` for `currentActivity` in the function below

kaibolay (approved):

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
lfkellogg: Stay tuned for the next PR :)

Right now the ScreenshotDetectionFeedbackTrigger cannot be enabled at the application level, but I will make it so in the next PR.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
lfkellogg: I thought that added a little too much boilerplate for my liking: having to declare a String field in the enum class, and then access that field everywhere we want to use it in code above. But I gave it a shot and it's actually not bad. Defining the field is super clean. So I like it after all, thanks for the push!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ShakeForFeedback.kt
lfkellogg: Kotlin FTW! Thanks

PR #4152, opened 2022-09-28
Reimplement screenshot trigger in Kotlin

lfkellogg:


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+  private var currentActivity: Activity? = null
+  private var currentUri: Uri? = null
+  private var isEnabled = false
+  private var hasRequestedPermission = false
+
+  override fun onChange(selfChange: Boolean, uri: Uri?) {
+    if (uri == null || !isExternalContent(uri) || seenImages.contains(uri)) {
+      return
+    }
+    currentActivity?.let { activity ->
```
kaibolay, line 41: I think http://go/kotlin/practices/scope-functions#avoid-using-let-where-smart-casts-suffice applies here.

Maybe you can just add `|| currentActivity == null` to the `if` above and Kotlin's "smart cast" can figure out that `currentActivity` is not null below?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+        val path = cursor.getString(cursor.getColumnIndexOrThrow(MediaStore.Images.Media.DATA))
+        info("Path: $path")
+        if (path.lowercase(Locale.getDefault()).contains("screenshot")) {
+          FirebaseAppDistribution.getInstance().startFeedback(infoTextResourceId, uri)
+        }
+      }
+    } catch (e: Exception) {
+      error("Could not determine if media change was due to taking a screenshot", e)
+    } finally {
+      cursor?.close()
```
kaibolay, line 129: it looks like `Cursor` implements `Closeable`, so you should be able to use try-with-resources. 
Oops, this is not Java - of course I mean `use` ([doc](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html))

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+    }
+  }
+
+  private fun listenForScreenshots() {
+    currentActivity?.let { activity ->
```
kaibolay, line 134: ditto: `if (currentActivity != null)` and take advantage of "smart cast"

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+      activity.contentResolver.registerContentObserver(
+        MediaStore.Images.Media.EXTERNAL_CONTENT_URI,
+        /* notifyForDescendants= */ true,
+        this
+      )
+    }
+  }
+
+  private fun stopListeningForScreenshots() {
+    currentActivity?.let { activity -> activity.contentResolver.unregisterContentObserver(this) }
```
kaibolay, line 144: ditto

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+      }
+    }
+
+    /**
+     * Start listening for screenshots, and start feedback when a new screenshot is detected.
+     *
+     * @throws IllegalStateException if [initialize] has not been called yet
+     */
+    open fun enable() {
+      instance?.also {
```
kaibolay, line 192: Curious: Why `also` instead of `let`?
Also (sic): what about just `if (instance == null) throw` as a "guard clause"?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+      )
+    }
+
+    /**
+     * Stop listening for screenshots.
+     *
+     * @throws IllegalStateException if [initialize] has not been called yet
+     */
+    open fun disable() {
+      instance?.also {
```
kaibolay, line 206: ditto

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -0,0 +1,226 @@
...
+      )
+    }
+
+    private fun isExternalContent(uri: Uri) =
+      uri.toString().matches("${MediaStore.Images.Media.EXTERNAL_CONTENT_URI}/\\d+".toRegex())
+
+    private fun isPermissionGranted(activity: Activity) =
+      ContextCompat.checkSelfPermission(activity, permissionToRequest) == PERMISSION_GRANTED
+
+    private fun info(msg: String) = Log.i(TAG, msg)
```
kaibolay, line 220: not sure if these pull their weight, but if you want to keep them, be my guest

kaibolay (approved):

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: Unfortunately I couldn't use smart cast here, because the `currentActivity` member could in theory have been set to null by some other thread in between the null check and the usage.

BUT I just realized another way to get around this is using functions that take a nullable argument. That argument becomes a local reference that I can then use with a smart cast.

We could talk about the possible issues from different threads updating and using currentActivity but I don't think that poses any major problems. There's a slim chance we try to use a reference to an activity that's already been hidden or destroyed. There won't be any NPEs or anything.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: Oh nice!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: Done

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: Done

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: Similar to the smart cast issue I mentioned, we can't use smart cast here because this member may have been reassigned in between the null check and usage.

We could use `if (instance == null) throw` but we'd still have to use a `!!` to assert that it's non-null.

Here I use `?.also` instead of `?.let` because `?.also` is safer when you want to handle the `else` case. It evaluates to the value of the nullable input value, instead of `let` which evaluates to the return value of the block. For example, this code would actually throw an exception regardless of whether `myNullableVar` is null or not:

```
myNullableVar?.let {
  println("Hi")
  null
} ?: throw IllegalStateException("This will always be thrown")
```

I'm just going to go ahead and get rid of all uses of both, though. I don't think a couple saved lines is worth the ding to readability.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: See other comment

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
lfkellogg: The log lines were formatting into these big ugly chunks like this:

```
Log.i(
  TAG,
  "Big long log line with a lot of text that will be logged out in one long line of logs"
)
```

But actually I'm already violating the linter by one-lining these shortened methods so maybe I should just one-line the Log.x calls...

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -188,39 +203,38 @@ private constructor(private val infoTextResourceId: Int, handler: Handler) :
...
+      requireInstance { it.disable() }
+    }
+
+    private fun requireInstance(func: (ScreenshotDetectionFeedbackTrigger) -> Unit) {
+      if (instance == null) {
+        throw IllegalStateException(
+          "You must call initialize() in your Application.onCreate() before enabling screenshot detection"
+        )
+      }
+      func(instance!!)
```
kaibolay, line 225: fancy!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
kaibolay: Thanks for explaining!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
kaibolay: Thanks for explaining!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
kaibolay: You could statically import `i()` and `e()` and shorten `TAG` to `T`. I'm not sure if I'm actually serious about this 🤣

kaibolay (approved):

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
thatfiredev: 🤣 
```kotlin
i(T, m)
```

Code Rating:
✅ Gets the job done
✅ Good SLOC
✅ No verbose
❌ Readability

lfkellogg:
> ```
> [I] fireci.gradle: > Task :firebase-appdistribution:testDebugUnitTest
> [I] fireci.gradle: 
> [I] fireci.gradle: com.google.firebase.appdistribution.impl.ApkUpdaterTest > updateApk_whenInstallFailed_setsError FAILED
> [I] fireci.gradle:     com.google.common.truth.AssertionErrorWithFacts at ApkUpdaterTest.java:236
> [I] fireci.gradle: 
> [I] fireci.gradle: 151 tests completed, 1 failed
> ```

Hmm, looks like an unrelated flake. Will address in a followup.

PR #4153, opened 2022-09-29
Fix ApkUpdaterTest flake

lfkellogg:
The test was asserting that the task is not complete, and then awaiting completion. However it's totally possible that it would have already completed by that assertion, which is probably why it failed randomly during a presubmit test on https://github.com/firebase/firebase-android-sdk/pull/4152.


kaibolay (approved):
Thanks!

PR #4155, opened 2022-09-29
update bom

VinayGuthal:


rlazo (commented):
You need to add them to the IGNORE list. And also, AFAIK you need to remove `-core`

davidmotson (approved):

rlazo (approved):

PR #4159, opened 2022-09-30
Implement notification trigger

kaibolay:


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+            } else {
+                requestPermission(activity)
+                hasRequestedPermission = true
+            }
+        }
+    }
+
+    private fun requestPermission(activity: Activity) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+            return // no need
```
lfkellogg, line 108: Before Tiramisu, would `ContextCompat.checkSelfPermission()` always return `PERMISSION_GRANTED`? If so then we would never reach this condition.

Even if we could reach this condition we'd just want to go ahead and show the notification, right? As is, we'd do nothing and set `hasRequestedPermission` to `true`.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+            cancelNotification(activity)
+        }
+    }
+
+    private fun showNotification(activity: Activity) {
+        if (ContextCompat.checkSelfPermission(activity, POST_NOTIFICATIONS) == PERMISSION_GRANTED) {
+            val intent = Intent(activity, TakeScreenshotAndTriggerFeedbackActivity::class.java)
+            intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY)
+            val pendingIntent =
+                PendingIntent.getActivity(activity, 0, intent, PendingIntent.FLAG_IMMUTABLE)
```
lfkellogg, line 82: Inline parameter comment for the `0` here?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+            val builder = NotificationCompat.Builder(activity, FEEBACK_NOTIFICATION_CHANNEL_ID)
+                .setSmallIcon(R.drawable.test_app_icon)
```
lfkellogg, line 84: Maybe leave a TODO here to pick a better icon? I believe this is an icon in the SDK resources (not the app resources) that is used for tests. Why that's available at runtime to the SDK is another thing we should fix at some point.

If you want to try one from the test app maybe `R.mipmap.ic_launcher`?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+                .setContentTitle(activity.getText(R.string.feedback_notification_title))
+                .setContentText(activity.getText(R.string.feedback_notification_text))
+                .setPriority(NotificationCompat.PRIORITY_HIGH)
```
lfkellogg, line 87: Should this be low priority?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+
+    private fun requestPermission(activity: Activity) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+            return // no need
+        }
+        if (activity.shouldShowRequestPermissionRationale(POST_NOTIFICATIONS)) {
+            Log.i(TAG, "Showing customer rationale for requesting permission.")
+            AlertDialog.Builder(activity)
+                .setMessage(
+                    "Using a notification to initiate feedback to the developer. To enable this feature, allow the app post notifications."
```
lfkellogg, line 114: "allow the app **to** post notifications"

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+        }
+    }
+
+    // Other lifecycle methods
+    override fun onActivityStarted(activity: Activity) {}
+    override fun onActivityStopped(activity: Activity) {}
+    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) {}
+
+    fun takeScreenshot() {
+        val activity = currentActivity
```
lfkellogg, line 180: So it looks like you are able to take a screenshot of a paused activity then, very cool. Out of curiosity, did you try taking a screenshot of the `TakeScreenshotAndTriggerFeedbackActivity` itself, and if so what happened?

firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -25,8 +25,8 @@ android {
         applicationId "com.googletest.firebase.appdistribution.testapp"
         minSdkVersion 23
         targetSdkVersion 33
-        versionName "1.0"
-        versionCode 1
+        versionName "2.0"
+        versionCode 2
```
lfkellogg, line 29: Just checking: do you want to keep this in?

firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -60,6 +60,7 @@ dependencies {
     println("Building with HEAD version ':firebase-appdistribution' of SDK")
 
     // All variants use the API
+    implementation project(':firebase-common')
```
lfkellogg, line 63: We're already adding `:firebase-common:ktx` below. Do we need this too? If so please put it down next to the other common one.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+    private var requestPermissionLauncher: ActivityResultLauncher<String>? = null
+
+    fun initialize(application: Application) {
+        // Create the NotificationChannel, but only on API 26+ because
+        // the NotificationChannel class is new and not in the support library
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val channel = NotificationChannel(
+                FEEBACK_NOTIFICATION_CHANNEL_ID,
+                application.getString(R.string.feedback_notification_channel_name),
+                NotificationManager.IMPORTANCE_HIGH
```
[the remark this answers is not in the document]
kaibolay: Yes, but unfortunately low importance messages sometimes just don't show up at all 😬  So I decided to err on the side of visibility - making sure the notification is seen. Since this is sample code selecting the appropriate importance is ultimately up to the developers integrating code based on this example into their app.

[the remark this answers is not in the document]
lfkellogg: Got it. I think it could be worth trying to make sure there's a level that makes sense though. If all of the levels either 1) make a sound or pop down over the current view, or 2) show up inconsistently, then it might not make sense to even supply sample code for a notification trigger.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: This was intended as defense-in-depth, but I can see how it was confusing. So I've changed it.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: Done

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: See above. I just want to make sure it's shown. I'll be up to the developer(s) adopting this code to select an appropriate priority.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: Done.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: It just took a screenshot of the empty activity.

firebase-appdistribution/test-app/test-app.gradle
kaibolay: Yes, I think we should get into the habit of increasing those numbers.
Adding the feedback functionality seems to enough for a major version bump.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,215 @@
...
+                    } else {
+                        Log.i(TAG, "Permission not granted")
+                        // TODO: Ideally we would show a message indicating the impact of not
+                        //   enabling the permission, but there's no way to know if they've
+                        //   permanently denied the permission, and we don't want to show them a
+                        //   message after each time we try to post a notification.
+                    }
+                }
+        } else {
+            Log.w(TAG, "Not listening for screenshots because this activity can't register for permission request results: $activity")
```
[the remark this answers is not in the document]
kaibolay: For example, I don't understand why registering the launcher happens in `onCreateActivity()` instead of in `requestPermission()` itself.

firebase-appdistribution/test-app/test-app.gradle
kaibolay: I don't need it. This was probably an artifact of a bad merge.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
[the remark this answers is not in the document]
kaibolay: I guess it's up to the developer even if it makes sounds. Some developers might really want to get feedback 😄

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: Done.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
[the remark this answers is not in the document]
lfkellogg: Sounds good let's have a look. I'd love to put it in some shared component but that goes against the idea of having an entire trigger in one nice copyable file.

But regardless unfortunately you have to register the launcher when the activity is initialized or else vague bad things happen... https://developer.android.com/reference/androidx/activity/result/ActivityResultCaller#registerForActivityResult(androidx.activity.result.contract.ActivityResultContract%3CI,O%3E,androidx.activity.result.ActivityResultCallback%3CO%3E)

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -0,0 +1,223 @@
...
+                Log.e(TAG, "Can't write $SCREENSHOT_FILE_NAME", e)
+            }
+        } else {
+            Log.e(TAG, "Can't take screenshot because current activity is unknown")
+            return
+        }
+    }
+}
+
+class TakeScreenshotAndTriggerFeedbackActivity : AppCompatActivity() {
```
lfkellogg, line 212: Just a note: if we're serious about supporting this use case we could add this activity to the SDK (along with tracking the previous activity so we could take the screenshot). Then they could just launch the intent for the activity and we could handle everything else. That might call for an API review though.

lfkellogg (approved):

PR #4170, opened 2022-10-05
Fix crash when requesting permissions after changing activities

lfkellogg:


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -35,7 +35,9 @@ object NotificationFeedbackTrigger : Application.ActivityLifecycleCallbacks {
   private var isEnabled = false
   private var hasRequestedPermission = false
   private var currentActivity: Activity? = null
-  private var requestPermissionLauncher: ActivityResultLauncher<String>? = null
+  // TODO(lkellogg): this is getting too complex - simplify it by, for example, only enabling this
+  //   on a per-activity basis instead of app-wide
+  private var requestPermissionLaunchers: MutableMap<Activity, ActivityResultLauncher<String>?> = mutableMapOf()
```
kaibolay, line 40: should this be a `WeakHashMap`? I don't think we want to prevent garbage collection of `Activities`.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -31,7 +31,9 @@ private constructor(private val infoTextResourceId: Int, handler: Handler) :
   ContentObserver(handler), Application.ActivityLifecycleCallbacks {
 
   private val seenImages = HashSet<Uri>()
-  private var requestPermissionLauncher: ActivityResultLauncher<String>? = null
+  // TODO(lkellogg): this is getting too complex - simplify it by, for example, only enabling this
+  //   on a per-activity basis instead of app-wide
```
kaibolay, line 35: I think permissions are always app-wide. But I agree that this needs to be simplified. I think requesting the permission in the `MainActivity` (maybe when the trigger is selected) should do the trick. Let's decide who should implement this simplification.

lfkellogg: I meant enabling the trigger app-wide. I'll clarify.

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/ScreenshotDetectionFeedbackTrigger.kt
```diff
@@ -62,10 +64,17 @@ private constructor(private val infoTextResourceId: Int, handler: Handler) :
           }
         }
     } else {
-      Log.w(TAG, "Not listening for screenshots because this activity can't register for permission request results: $activity")
+      Log.w(
+        TAG,
+        "Not listening for screenshots because this activity can't register for permission request results: $activity"
```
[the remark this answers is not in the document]
lfkellogg: Because my linter runs automatically now and this is the correct formatting apparently :)

kaibolay (approved):

PR #4177, opened 2022-10-06
Parse NO_CHANGE fetch response for template version

qdpham13:
Parse NO_CHANGE fetch response for template version. As part of [go/realtime-rc-forward-compatibility](http://goto.google.com/realtime-rc-forward-compatibility)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -575,11 +575,12 @@ public static FetchResponse forBackendUpdatesFetched(
...
     }
 
-    public static FetchResponse forBackendHasNoUpdates(Date fetchTime) {
+    public static FetchResponse forBackendHasNoUpdates(
+        Date fetchTime, ConfigContainer fetchedConfigs) {
       return new FetchResponse(
           fetchTime,
           Status.BACKEND_HAS_NO_UPDATES,
-          /*fetchedConfigs=*/ null,
+          /*fetchedConfigs=*/ fetchedConfigs,
```
danasilver, line 583: I might add a comment that we're including the `fetchConfigs` because they include the template version (even though they don't include the config entries or metadata)

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClientTest.java
```diff
@@ -167,7 +167,7 @@ public void fetch_noChange_responseNotSet() throws Exception {
     FetchResponse response = fetch(SECOND_ETAG);
 
     assertThat(response.getLastFetchETag()).isNull();
-    assertThat(response.getFetchedConfigs()).isNull();
+    assertThat(response.getFetchedConfigs()).isNotNull();
```
danasilver, line 170: Maybe assert it only contains template version?

danasilver (approved):

PR #4179, opened 2022-10-06
Notifications: Ask permission upfront, show only one persistent notification

kaibolay:


firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -177,7 +177,7 @@ public interface FirebaseAppDistribution {
    * @param screenshot URI to a bitmap containing a screenshot that will be included with the
    *     report, or null to not include a screenshot
    */
-  void startFeedback(@NonNull CharSequence infoText, @Nullable Uri screenshotUri);
+  void startFeedback(@NonNull CharSequence infoText, @Nullable Uri screenshot);
```
lfkellogg, line 180: Nice :)

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -59,80 +60,103 @@ object NotificationFeedbackTrigger : Application.ActivityLifecycleCallbacks {
...
-      showNotification(currentActivity)
+  /**
+   * Requests permission to show notifications for this application.
+   *
+   * This must be called during [Activity.onCreate].
+   * [enable] should then be called when you want to actually show the notification.
+   *
+   * @param activity the [Activity] object
+   */
+  @RequiresApi(Build.VERSION_CODES.TIRAMISU)
```
lfkellogg, line 71: Nice!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -13,33 +13,34 @@ import android.os.Build
...
-  const val SCREENSHOT_FILE_NAME = "com.googletest.firebase.appdistribution.testapp.screenshot.png"
 
   private var isEnabled = false
   private var hasRequestedPermission = false
-  private var currentActivity: Activity? = null
-  // TODO(lkellogg): this is getting too complex - simplify it by, for example, only enabling this
-  //   trigger on a per-activity basis instead of app-wide
-  private var requestPermissionLaunchers: WeakHashMap<Activity, ActivityResultLauncher<String>?> = WeakHashMap()
 
+  internal var currentActivity: Activity? = null // Activity to be used for screenshot
```
lfkellogg, line 34: Should we call this something like `activityToScreenshot`? Could then also remove the comment below that this "points to the previous activity" since it wouldn't be confusing that it's called "current".

lfkellogg (approved):
This looks great and works well for our test app so I think we should submit this as is. That said I just want to call out that I think when we move this to the SDK we could combine `initialize`, `requestPermission`, and `enable` into one call, and maybe even remove `disable`:

- `initialize()` needed to be called in the `Application.onCreate()` so we'd be able to hook into the `Activity.onCreate()` using lifecycle callbacks. But that's not necessary anymore now that we're telling them to call `requestPermission()` in the `Activity.onCreate()`. We can just combine the two.
- `enable()` can probably also be combined into the same method. Our test app wants the ability to toggle this on and off but I don't see that being super useful in a real app. They probably just want to either enable it or not when the app starts. Also if they're requesting the permission they probably just want to go ahead and enable it.
- For the same reason, I think we can consider removing `disable()`

kaibolay:
> This looks great and works well for our test app so I think we should submit this as is. That said I just want to call out that I think when we move this to the SDK we could combine `initialize`, `requestPermission`, and `enable` into one call, and maybe even remove `disable` [...]

Makes sense!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
kaibolay: Done.

PR #4183, opened 2022-10-07
Notification trigger API

lfkellogg:
This adds a new API allowing developers to show a notification that, when tapped, will take a screenshot of the running application and open the feedback form. This way the developers don't need to add any new UI elements to their app to trigger feedback.

Other changes include:

- The lifecycle notifier now tracks the previous activity in addition to the current one, so that we can ignore our own empty activity when taking a screenshot after the notification is tapped.
- Adds the SDK-based notification trigger to the test app

### DIFF
--- a/firebase-appdistribution-api/api.txt
+++ b/firebase-appdistribution-api/api.txt
@@ -16,7 +16,10 @@ package com.google.firebase.appdistribution {
   public interface FirebaseAppDistribution {
     method @NonNull public com.google.android.gms.tasks.Task<com.google.firebase.appdistribution.AppDistributionRelease> checkForNewRelease();
     method @NonNull public static com.google.firebase.appdistribution.FirebaseAppDistribution getInstance();
+    method public void cancelFeedbackNotification();
     method public boolean isTesterSignedIn();
+    method public void showFeedbackNotification(@NonNull int, int);
+    method public void showFeedbackNotification(@NonNull CharSequence, int);
     method @NonNull public com.google.android.gms.tasks.Task<java.lang.Void> signInTester();
     method public void signOutTester();
     method public void startFeedback(int);
--- a/firebase-appdistribution-api/api.txt
+++ b/firebase-appdistribution-api/api.txt
@@ -18,13 +18,13 @@ package com.google.firebase.appdistribution {
     method @NonNull public com.google.android.gms.tasks.Task<com.google.firebase.appdistribution.AppDistributionRelease> checkForNewRelease();
     method @NonNull public static com.google.firebase.appdistribution.FirebaseAppDistribution getInstance();
     method public boolean isTesterSignedIn();
-    method public void showFeedbackNotification(int, @NonNull com.google.firebase.appdistribution.InterruptionLevel);
+    method public void showFeedbackNotification(@StringRes int, @NonNull com.google.firebase.appdistribution.InterruptionLevel);
     method public void showFeedbackNotification(@NonNull CharSequence, @NonNull com.google.firebase.appdistribution.InterruptionLevel);
     method @NonNull public com.google.android.gms.tasks.Task<java.lang.Void> signInTester();
     method public void signOutTester();
-    method public void startFeedback(int);
+    method public void startFeedback(@StringRes int);
     method public void startFeedback(@NonNull CharSequence);
-    method public void startFeedback(int, @Nullable android.net.Uri);
+    method public void startFeedback(@StringRes int, @Nullable android.net.Uri);
     method public void startFeedback(@NonNull CharSequence, @Nullable android.net.Uri);
     method @NonNull public com.google.firebase.appdistribution.UpdateTask updateApp();
     method @NonNull public com.google.firebase.appdistribution.UpdateTask updateIfNewReleaseAvailable();
--- a/firebase-appdistribution-api/api.txt
+++ b/firebase-appdistribution-api/api.txt
@@ -14,17 +14,17 @@ package com.google.firebase.appdistribution {
   }
 
   public interface FirebaseAppDistribution {
+    method public void cancelFeedbackNotification();
     method @NonNull public com.google.android.gms.tasks.Task<com.google.firebase.appdistribution.AppDistributionRelease> checkForNewRelease();
     method @NonNull public static com.google.firebase.appdistribution.FirebaseAppDistribution getInstance();
-    method public void cancelFeedbackNotification();
     method public boolean isTesterSignedIn();
-    method public void showFeedbackNotification(@NonNull int, int);
+    method public void showFeedbackNotification(int, int);
     method public void showFeedbackNotification(@NonNull CharSequence, int);
     method @NonNull public com.google.android.gms.tasks.Task<java.lang.Void> signInTester();
     method public void signOutTester();
     method public void startFeedback(int);
     method public void startFeedback(@NonNull CharSequence);
-    method public void startFeedback(@NonNull int, @Nullable android.net.Uri);
+    method public void startFeedback(int, @Nullable android.net.Uri);
     method public void startFeedback(@NonNull CharSequence, @Nullable android.net.Uri);
     method @NonNull public com.google.firebase.appdistribution.UpdateTask updateApp();
     method @NonNull public com.google.firebase.appdistribution.UpdateTask updateIfNewReleaseAvailable();
--- a/firebase-appdistribution-api/api.txt
+++ b/firebase-appdistribution-api/api.txt
@@ -18,8 +18,8 @@ package com.google.firebase.appdistribution {
     method @NonNull public com.google.android.gms.tasks.Task<com.google.firebase.appdistribution.AppDistributionRelease> checkForNewRelease();
     method @NonNull public static com.google.firebase.appdistribution.FirebaseAppDistribution getInstance();
     method public boolean isTesterSignedIn();
-    method public void showFeedbackNotification(int, int);
-    method public void showFeedbackNotification(@NonNull CharSequence, int);
+    method public void showFeedbackNotification(int, @NonNull com.google.firebase.appdistribution.InterruptionLevel);
+    method public void showFeedbackNotification(@NonNull CharSequence, @NonNull com.google.firebase.appdistribution.InterruptionLevel);
     method @NonNull public com.google.android.gms.tasks.Task<java.lang.Void> signInTester();
     method public void signOutTester();
     method public void startFeedback(int);
@@ -47,6 +47,14 @@ package com.google.firebase.appdistribution {
     enum_constant public static final com.google.firebase.appdistribution.FirebaseAppDistributionException.Status NOT_IMPLEMENTED;
     enum_constant public static final com.google.firebase.appdistribution.FirebaseAppDistributionException.Status UNKNOWN;
     enum_constant public static final com.google.firebase.appdistribution.FirebaseAppDistributionException.Status UPDATE_NOT_AVAILABLE;
+  }
+
+  public enum InterruptionLevel {
+    enum_constant public static final com.google.firebase.appdistribution.InterruptionLevel DEFAULT;
+    enum_constant public static final com.google.firebase.appdistribution.InterruptionLevel HIGH;
+    enum_constant public static final com.google.firebase.appdistribution.InterruptionLevel LOW;
+    enum_constant public static final com.google.firebase.appdistribution.InterruptionLevel MAX;
+    enum_constant public static final com.google.firebase.appdistribution.InterruptionLevel MIN;
   }
 
   public interface OnProgressListener {
--- a/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -205,10 +205,10 @@ public interface FirebaseAppDistribution {
    *
    * @param infoTextResourceId string resource ID of text to display to the tester before collecting
    *     feedback data (e.g. Terms and Conditions)
-   * @param importance the amount the user should be interrupted by notifications from the feedback
-   *     notification channel. Once the channel's importance is set it cannot be changed except by
-   *     the user. See {@link NotificationChannel#setImportance}. On platforms below Android 8, the
-   *     importance will be translated into a comparable notification priority (see {@link
+   * @param importance the level of interruption for the feedback notification channel. Once the
+   *     channel's importance is set it cannot be changed except by the user. See {@link
+   *     NotificationChannel#setImportance}. On platforms below Android 8, the importance will be
+   *     translated into a comparable notification priority (see {@link
    *     NotificationCompat.Builder#setPriority}).
    */
   void showFeedbackNotification(int infoTextResourceId, int importance);
@@ -236,10 +236,10 @@ public interface FirebaseAppDistribution {
    *
    * @param infoText text to display to the tester before collecting feedback data (e.g. Terms and
    *     Conditions)
-   * @param importance the amount the user should be interrupted by notifications from the feedback
-   *     notification channel. Once the channel's importance is set it cannot be changed except by
-   *     the user. See {@link NotificationChannel#setImportance}. On platforms below Android 8, the
-   *     importance will be translated into a comparable notification priority (see {@link
+   * @param importance the level of interruption for the feedback notification channel. Once the
+   *     channel's importance is set it cannot be changed except by the user. See {@link
+   *     NotificationChannel#setImportance}. On platforms below Android 8, the importance will be
+   *     translated into a comparable notification priority (see {@link
    *     NotificationCompat.Builder#setPriority}).
    */
   void showFeedbackNotification(@NonNull CharSequence infoText, int importance);
--- a/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
+++ b/firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
@@ -15,11 +15,9 @@
 package com.google.firebase.appdistribution;
 
 import android.app.Activity;
-import android.app.NotificationChannel;
 import android.net.Uri;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import androidx.core.app.NotificationCompat;
 import com.google.android.gms.tasks.Task;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.appdistribution.internal.FirebaseAppDistributionProxy;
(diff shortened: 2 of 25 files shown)


firebase-appdistribution-api/api.txt
```diff
@@ -16,7 +16,10 @@ package com.google.firebase.appdistribution {
   public interface FirebaseAppDistribution {
     method @NonNull public com.google.android.gms.tasks.Task<com.google.firebase.appdistribution.AppDistributionRelease> checkForNewRelease();
     method @NonNull public static com.google.firebase.appdistribution.FirebaseAppDistribution getInstance();
+    method public void cancelFeedbackNotification();
     method public boolean isTesterSignedIn();
+    method public void showFeedbackNotification(@NonNull int, int);
```
kaibolay, line 21: Any idea why the `int` is annotated with `@NonNull` here (and also on line 27)? It's weird, but probably harmless.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -179,6 +182,76 @@ public interface FirebaseAppDistribution {
    */
   void startFeedback(@NonNull CharSequence infoText, @Nullable Uri screenshot);
 
+  /**
+   * Displays a notification that, when tapped, will take a screenshot of the current activity, then
+   * start a new activity to collect and submit feedback from the tester along with the screenshot.
+   *
+   * <p>On Android 13 and above, this method requires the <a
+   * href="https://developer.android.com/develop/ui/views/notifications/notification-permission">runtime
```
kaibolay, line 190: Does a `*` in the middle of the HTML tag work for JavaDoc?

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -179,6 +182,76 @@ public interface FirebaseAppDistribution {
...
+   *   <li>If the app is open, take a screenshot of the current activity
+   *   <li>If tester is not signed in, presents the tester with a Google Sign-in UI
+   *   <li>Starts a full screen activity for the tester to compose and submit the feedback
+   * </ol>
+   *
+   * <p>On Android 8 and above, the notification will be created in its own notification channel.
+   *
+   * @param infoTextResourceId string resource ID of text to display to the tester before collecting
+   *     feedback data (e.g. Terms and Conditions)
+   * @param importance the amount the user should be interrupted by notifications from the feedback
```
kaibolay, line 208: "amount" seems a bit odd here. Maybe something like "the level of interruption for this notification channel" (this is inspired by the JavaDoc of `NotificationChannel.setImportance()`)

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -268,8 +334,13 @@ public void onActivitySaveInstanceState(@NonNull Activity activity, @NonNull Bun
   @Override
   public void onActivityDestroyed(@NonNull Activity activity) {
     synchronized (lock) {
-      if (this.currentActivity == activity) {
-        this.currentActivity = null;
+      if (currentActivity == activity) {
+        updateCurrentActivity(null);
+      }
+
+      // If an activity is destroyed, delete all references to it, including the previous activity
```
kaibolay, line 341: nit: move this comment a few lines earlier?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -15,25 +15,53 @@
...
 
   @VisibleForTesting
-  static final String NOTIFICATION_TAG = "com.google.firebase.appdistribution.tag";
+  static final String CHANNEL_GROUP_ID =
+      "com.google.firebase.appdistribution.notification_channel_group_id";
+
+  @VisibleForTesting
+  enum Notification {
+    APP_UPDATE(
+        "com.google.firebase.appdistribution.notification_channel_id",
```
kaibolay, line 43: maybe remove to common prefix of all Strings (i.e. "com.google.firebase.appdistribution.") and add in in the constructor implementation

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionNotificationsManager.java
```diff
@@ -15,25 +15,53 @@
...
+        "com.google.firebase.appdistribution.app_update_notification_tag",
+        0),
+    FEEDBACK(
+        "com.google.firebase.appdistribution.feedback_notification_channel_id",
+        "com.google.firebase.appdistribution.feedback_notification_tag",
+        1);
+
+    final String channelId;
+    final String tag;
+    final int id;
```
kaibolay, line 53: I think you could just use the `ordinal` of the `enum` as the `id`.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TakeScreenshotAndStartFeedbackActivity.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private CharSequence infoText;
+
+  @Override
+  public void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    infoText = getIntent().getCharSequenceExtra(INFO_TEXT_EXTRA_KEY);
+    LogWrapper.getInstance().i(TAG, "Capturing screenshot and starting feedback");
+    FirebaseAppDistribution.getInstance().startFeedback(infoText);
+    finish();
```
kaibolay, line 36: I assume this takes care of getting rid of this activity even if `startFeedback()` fails for some reason. Nice!

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
```diff
@@ -7,8 +7,8 @@ class AppDistroTestApplication : Application() {
         super.onCreate()
 
         // Perform any required trigger initialization here
-        ScreenshotDetectionFeedbackTrigger.initialize(this, R.string.terms_and_conditions);
-        NotificationFeedbackTrigger.initialize(this);
+        ScreenshotDetectionFeedbackTrigger.initialize(this, R.string.terms_and_conditions)
+//        NotificationFeedbackTrigger.initialize(this);
```
kaibolay, line 11: Remove entirely? But maybe not - I think we should keep this trigger in the test app... (see below)

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -66,17 +67,15 @@ class MainActivity : AppCompatActivity() {
...
+        firebaseAppDistribution.showFeedbackNotification(
+            R.string.terms_and_conditions, NotificationManagerCompat.IMPORTANCE_HIGH);
 
         // Set up feedback trigger menu
         feedbackTriggerMenu = findViewById(R.id.feedbackTriggerMenu)
         val items = listOf(
             FeedbackTrigger.NONE.label,
             FeedbackTrigger.SHAKE.label,
             FeedbackTrigger.SCREENSHOT.label,
-            FeedbackTrigger.NOTIFICATION.label,
```
kaibolay, line 79: Why not keep this and call into the SDK for the functionality?
Or even more fancy: Have two selections for notifications: `NOTIFICATIONS_SDK` and `NOTIFICATIONS_APP` which stand for the notification provided by the SDK and (alternatively) a notification trigger implemented in the app (which can be more customized - so it's valuable example code)

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
```diff
@@ -221,9 +221,9 @@ class TakeScreenshotAndTriggerFeedbackActivity : Activity() {
       activity.openFileOutput(SCREENSHOT_FILE_NAME, Context.MODE_PRIVATE).use { outputStream ->
         bitmap.compress(Bitmap.CompressFormat.PNG, /* quality = */ 100, outputStream)
       }
```
kaibolay, line 223: Silly GitHub doesn't let me comment on line 213... Anyhow, should there be a `finish()` at the end of `onResume()`?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionNotificationsManager.java
lfkellogg: Yeah I thought about that but wondered if it would be better for search-ability to have the entire string defined here...

But thinking about it again the part after the last `.` is unique enough that that should be quite searchable. Done.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TakeScreenshotAndStartFeedbackActivity.java
lfkellogg: Yep exactly :)

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
lfkellogg: Hmm, yeah I like that idea. Done.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
lfkellogg: According to the rendered view in Android Studio it works 👍

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/NotificationFeedbackTrigger.kt
lfkellogg: True. Out of curiosity - any reason you split the behavior between onCreate and onResume? Or could it all go in the onCreate like it is in the SDK?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/AppDistroTestApplication.kt
lfkellogg: I put it back based on your feedback below.

firebase-appdistribution-api/api.txt
```diff
@@ -14,14 +14,17 @@ package com.google.firebase.appdistribution {
...
     method @NonNull public static com.google.firebase.appdistribution.FirebaseAppDistribution getInstance();
     method public boolean isTesterSignedIn();
+    method public void showFeedbackNotification(int, int);
+    method public void showFeedbackNotification(@NonNull CharSequence, int);
     method @NonNull public com.google.android.gms.tasks.Task<java.lang.Void> signInTester();
     method public void signOutTester();
     method public void startFeedback(int);
     method public void startFeedback(@NonNull CharSequence);
-    method public void startFeedback(@NonNull int, @Nullable android.net.Uri);
+    method public void startFeedback(int, @Nullable android.net.Uri);
```
lfkellogg, line 27: Removed the unnecessary `@NonNull` annotation.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -179,6 +182,76 @@ public interface FirebaseAppDistribution {
...
+   *
+   * <ol>
+   *   <li>If the app is open, take a screenshot of the current activity
+   *   <li>If tester is not signed in, presents the tester with a Google Sign-in UI
+   *   <li>Starts a full screen activity for the tester to compose and submit the feedback
+   * </ol>
+   *
+   * <p>On Android 8 and above, the notification will be created in its own notification channel.
+   *
+   * @param infoTextResourceId string resource ID of text to display to the tester before collecting
```
vkryachko, line 206: Does the underlying api require a resource id? if not, do we want to restrict to only allow resource ids as opposed to allowing passing in custom strings like constants?


If resourceId is preferred, please make sure to annotate this param with `@StringRes`.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -179,6 +182,76 @@ public interface FirebaseAppDistribution {
...
+   *     feedback data (e.g. Terms and Conditions)
+   * @param importance the level of interruption for the feedback notification channel. Once the
+   *     channel's importance is set it cannot be changed except by the user. See {@link
+   *     NotificationChannel#setImportance}. On platforms below Android 8, the importance will be
+   *     translated into a comparable notification priority (see {@link
```
vkryachko, line 211: If we are doing some translation, would it make sense to create an enum as opposed to relying on the platform constants being passed in?

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
```diff
@@ -179,6 +182,76 @@ public interface FirebaseAppDistribution {
...
+   *     NotificationCompat.Builder#setPriority}).
+   */
+  void showFeedbackNotification(int infoTextResourceId, int importance);
```
vkryachko, line 214: For my understanding, what will happen if called from background? do Android background restrictions apply to it?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -181,6 +225,28 @@ public void onResumed(Activity activity) {
...
+      return activity;
+    }
+  }
+
+  private void updateCurrentActivity(@Nullable Activity activity) {
+    synchronized (lock) {
+      if (currentActivity != activity) {
+        if (currentActivity != null) {
+          // Store a reference to the previous activity in case the current activity is ignored
+          previousActivity = currentActivity;
```
vkryachko, line 243: If the current activity is an ignored one, do you even need to store it?

kaibolay (approved):

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
lfkellogg: Good question - I just tested it and it works, even after navigating away from the app. This makes sense - notifications are actually the recommended way for apps to present UI to the user when in the background: https://developer.android.com/guide/components/activities/background-starts#display-notification

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
lfkellogg: I should make this comment clearer - what I mean is that the currentActivity _may be ignored later_ when `applyToNullableForegroundActivity(classToIgnore, ...)` is called.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
lfkellogg: Here's what that would look like: https://github.com/firebase/firebase-android-sdk/pull/4205/files

Let me know if you think we should go through an API review for this.

kaibolay: I like it (just left one comment for simplification). I don't think we need an API review now. But after EAP we should go through an API review (again) for everything. Many small details changed - and might keep changing - e.g. whether to accept `CharSequence` or `@StringRes int` or both.

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
kaibolay: My preference is in the other direction. If we eliminate one, I'd vote for the overload with `CharSequence` and only keep `@StringRes int`. The latter is a best practice. Hard-coded strings are smelly. Or we keep both, but add JavaDoc which explains that `@StringRes int` is preferred.

lfkellogg: That's a good point, Kai. I actually just noticed that if you use `TextView.setText(CharSequence)` you get a compiler warning: "String literal in setText can not be translated. Use Android resources instead."

So I'm tempted to remove the CharSequence option, but that would prevent the dev from using format strings: https://developer.android.com/guide/topics/resources/string-resource#formatting-strings

I feel like we should support that.

vkryachko: Would it make sense to add an overload that would mirror `getResource(int, Object...)`? i.e. something like

```java
void showFeedbackNotification(
      @NonNull InterruptionLevel interruptionLevel, @StringRes int infoTextResourceId, Object... formatArgs);
```

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -181,6 +225,29 @@ public void onResumed(Activity activity) {
...
+    }
+  }
+
+  private void updateCurrentActivity(@Nullable Activity activity) {
+    synchronized (lock) {
+      if (currentActivity != activity) {
+        if (currentActivity != null) {
+          // Store a reference to the previous activity in case the current activity is ignored
+          // later in call to applyToNullableForegroundActivity()
+          previousActivity = currentActivity;
```
vkryachko, line 244: I'm not sure if it's actually a problem in practice in you use case, but in general holding a reference to `currentActivity` should be safe, while it's not necessarily the case with `previousActivity` given that that activity may get destroyed while `currentActivity` is active. Is that a possible scenario here? What would happen if you try to use that activity?

lfkellogg: That's not a possible scenario because if `previousActivity` gets destroyed, we remove the reference to it on line 343 (in `onActivityDestroyed`).

firebase-appdistribution-api/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistribution.java
lfkellogg: We could, but if they want to use a format string AND html elements like `<a>` tags they would have to do some extra steps and then final pass us a styled CharSequence: https://developer.android.com/guide/topics/resources/string-resource#StylingWithHTML

This might seem like an edge case, but because this text will likely include links to terms and conditions I expect anyone who wants to use a format string will have to do this.

vkryachko (approved):

Issue #4146, opened 2022-09-27
App crash: NullPointerException


OlehHaidaienko:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: Android Studio Dolphin | 2021.3.1
  * Firebase Component: Performance
  * Component version: 20.1.1

#### Steps to reproduce:

After updating Firebase BoM from `29.2.0`  to `30.4.1` we started receiving a lot of crashes on Android 7, 8, and 9. Unfortunately, we can't reproduce this issue. 

Here is the crash log
```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int java.util.ArrayList.size()' on a null object reference
       at android.view.View.findFrameMetricsObserver(View.java:6539)
       at android.view.View.removeFrameMetricsListener(View.java:6510)
       at android.view.Window.removeOnFrameMetricsAvailableListener(Window.java:886)
       at androidx.core.app.FrameMetricsAggregator$FrameMetricsApi24Impl.remove(FrameMetricsAggregator.java:433)
       at androidx.core.app.FrameMetricsAggregator.remove(FrameMetricsAggregator.java:243)
       at com.google.firebase.perf.application.FrameMetricsRecorder.stop(FrameMetricsRecorder.java:110)
       at com.google.firebase.perf.application.AppStateMonitor.sendScreenTrace(AppStateMonitor.java:355)
       at com.google.firebase.perf.application.AppStateMonitor.onActivityStopped(AppStateMonitor.java:210)
       at android.app.Application.dispatchActivityStopped(Application.java:257)
       at android.app.Activity.onStop(Activity.java:1914)
       at androidx.fragment.app.FragmentActivity.onStop(FragmentActivity.java:496)
       at androidx.appcompat.app.AppCompatActivity.onStop(AppCompatActivity.java:252)
       at com.fusionmedia.investing.ui.activities.base.BaseSplashActivity.onStop(BaseSplashActivity.java:922)
       at android.app.Instrumentation.callActivityOnStop(Instrumentation.java:1433)
       at android.app.Activity.performStop(Activity.java:7367)
       at android.app.ActivityThread.callActivityOnStop(ActivityThread.java:4146)
       at android.app.ActivityThread.performStopActivityInner(ActivityThread.java:4124)
       at android.app.ActivityThread.handleStopActivity(ActivityThread.java:4199)
[...]

PR #4184, opened 2022-10-07
Catch NPE when removing activity from frameMetricsAggregator

jeremyjiang-dev:
Potential fix of #4146 

I was not able to reproduce the issue locally. I tried with Android version 7.1 API 25 and turned off hardware acceleration, but could not reproduce the issue.

It could be a bug from older Android versions.

Reference:
- A similar issue: https://bytemeta.vip/repo/getsentry/sentry-java/issues/1798
- Android's fix on Fix NPE in findFrameMetricsObserver: https://android.googlesource.com/platform/frameworks/base/+/140ff5ea8e2d99edc3fbe63a43239e459334c76b

## Update by @leotianlizhan
Let's take a look at [Android's source code, in android.view.View](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=7381;drc=master?q=addFrameMetrics&ss=android%2Fplatform%2Fsuperproject).
The exception was thrown from `findFrameMetricsObserver`, inside `removeFrameMetricsListener`
image
image
Since this null check didn't exists, this triggers NPE when `mFrameMetricsObserver` is uninitialized. As Jeremy listed in his reference, this was fixed in later Android versions. Why then, is `mFrameMetricsObserver` uninitialized? It's because `addFrameMetricsListener` fails silently!
image

## Conclusion
So overall, this is the **same issue caused by the view not being hardware-accelerated**. The correct error message is hidden away because `addFrameMetricsListener` fails silently, that's why we should catch them both and use the same error message. The developer's provided crash versions also lines up with when Android fixed it: the above reference fix was made in 2019, and Android 7, 8, 9 are all before 2019, with Android 10 releasing in 2019 thus fixing the issue. There is no crash before Android 7 because as Rosario pointed out, FrameMetrics was added in Android 7 (API 24).

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
@@ -108,7 +108,7 @@ public class FrameMetricsRecorder {
     Optional<PerfFrameMetrics> data = this.snapshot();
     try {
       frameMetricsAggregator.remove(activity);
-    } catch (IllegalArgumentException err) {
+    } catch (IllegalArgumentException | NullPointerException err) {
       logger.warn(
           "View not hardware accelerated. Unable to collect FrameMetrics. %s", err.toString());
       return Optional.absent();
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
@@ -108,16 +108,15 @@ public class FrameMetricsRecorder {
     Optional<PerfFrameMetrics> data = this.snapshot();
     try {
       frameMetricsAggregator.remove(activity);
-    } catch (IllegalArgumentException err) {
+    } catch (IllegalArgumentException | NullPointerException ex) {
+      // Both of these exceptions result from android.view.View.addFrameMetricsListener silently
+      // failing when the view is not hardware-accelerated. Successful addFrameMetricsListener
+      // stores an observer in a list, and initializes the list if it was uninitialized. Invoking
+      // View.removeFrameMetricsListener(listener) throws IAE if it doesn't exist in the list, or
+      // throws NPE if the list itself was never initialized.
       logger.warn(
-          "View not hardware accelerated. Unable to collect FrameMetrics. %s", err.toString());
-      frameMetricsAggregator.reset();
-      return Optional.absent();
-    } catch (NullPointerException err) {
-      frameMetricsAggregator.reset();
-      logger.warn(
-          "NullPointerException. %s", err.toString());
-      return Optional.absent();
+          "View not hardware accelerated. Unable to collect FrameMetrics. %s", ex.toString());
+      data = Optional.absent();
     }
     frameMetricsAggregator.reset();
     isRecording = false;
--- a/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
@@ -15,6 +15,7 @@
 package com.google.firebase.perf.application;
 
 import android.app.Activity;
+import android.os.Build;
 import android.util.SparseIntArray;
 import androidx.core.app.FrameMetricsAggregator;
 import androidx.fragment.app.Fragment;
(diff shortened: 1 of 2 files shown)


visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -111,6 +111,12 @@ public Optional<PerfFrameMetrics> stop() {
     } catch (IllegalArgumentException err) {
       logger.warn(
           "View not hardware accelerated. Unable to collect FrameMetrics. %s", err.toString());
+      frameMetricsAggregator.reset();
+      return Optional.absent();
+    } catch (NullPointerException err) {
+      frameMetricsAggregator.reset();
+      logger.warn(
+          "NullPointerException. %s", err.toString());
```
raymondlam, line 119: nit: Does the err.toString() have enough context to understand where the issue came from? Otherwise you can be more clear and say "NullPointerException when calling FrameMetricsAggregator.remove()"

raymondlam (approved):

thatfiredev:
@jeremyjiang-dev the underlying issue seems to be the fact that the `FrameMetrics` API was introduced in API 24, as documented [here](https://developer.android.com/reference/kotlin/androidx/core/app/FrameMetricsAggregator).

The pre-24 implementation of `FrameMetricsAggregator` returns null values, which causes this crash: https://github.com/androidx/androidx/blob/0bc4d894a0b7d46abcbc05aedb21ed72a8e71db5/core/core/src/main/java/androidx/core/app/FrameMetricsAggregator.java#L305-L330

Instead of catching this Exception, should we maybe add that condition to `FrameMetricsRecorder.isFrameMetricsRecordingSupported()` so that we don't start automatic screen tracing in devices with API < 24 ?

```kotlin
static boolean isFrameMetricsRecordingSupported() {
    try {
      Class<?> initializerClass = Class.forName(FRAME_METRICS_AGGREGATOR_CLASSNAME);
      return Build.VERSION.SDK_INT >= 24;
    } catch (ClassNotFoundException e) {
      return false;
    }
  }
```

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
```diff
@@ -111,6 +111,12 @@ public Optional<PerfFrameMetrics> stop() {
     } catch (IllegalArgumentException err) {
       logger.warn(
           "View not hardware accelerated. Unable to collect FrameMetrics. %s", err.toString());
+      frameMetricsAggregator.reset();
```
leotianlizhan, line 114: Do we need `frameMetricsAggregator.reset()` here when it's called after the `catch` blocks?

raymondlam:
Does this mean we are not capture screen rendering when a device has API < 24?

jeremyjiang-dev:
> **Update**: I took a second look at the crash log provided in the issue and it seems like the error is happening with `FrameMetricsApi24Impl`:
> 
> ```
> at androidx.core.app.FrameMetricsAggregator$FrameMetricsApi24Impl.remove(FrameMetricsAggregator.java:433)
> ```
> 
> So my previous comment (bellow) is not entirely true.
> 
> @jeremyjiang-dev the underlying issue seems to be the fact that the `FrameMetrics` API was introduced in API 24, as documented [here](https://developer.android.com/reference/kotlin/androidx/core/app/FrameMetricsAggregator).
> 
> The pre-24 implementation of `FrameMetricsAggregator` returns null values, which causes this crash: https://github.com/androidx/androidx/blob/0bc4d894a0b7d46abcbc05aedb21ed72a8e71db5/core/core/src/main/java/androidx/core/app/FrameMetricsAggregator.java#L305-L330
> 
> Instead of catching this Exception, should we maybe add that condition to `FrameMetricsRecorder.isFrameMetricsRecordingSupported()` so that we don't start automatic screen tracing in devices with API < 24 ?
> 
> ```kotlin
> static boolean isFrameMetricsRecordingSupported() {
>     try {
>       Class<?> initializerClass = Class.forName(FRAME_METRICS_AGGREGATOR_CLASSNAME);
>       return Build.VERSION.SDK_INT >= 24;
>     } catch (ClassNotFoundException e) {
>       return false;
>     }
>   }
> ```

Thanks Rosário for the investigation! I thought about API 24 could be the issue, but the issue also happens to Android 7, 8, and 9 according to the developer. Android 7, 8, and 9 are mapped to API level 24+, so I am not sure if API 24 was the cause. 
Ref: https://source.android.com/docs/setup/about/build-numbers

leotianlizhan:
@visumickey @raymondlam I updated this PR to categorize this NPE into the same error message "View no hardware accelerated", and a lengthy comment to explain why. I also updated the [PR description](https://github.com/firebase/firebase-android-sdk/pull/4184#issue-1401370924) with more detail explaining with Android Source Code why this is the case.

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java
leotianlizhan: Recategorized the error, please see updated description

thatfiredev:
@leotianlizhan amazing job on the investigation!! Thanks for sharing the results here (publicly instead of a internal source) :)

leotianlizhan:
PR is moved to https://github.com/firebase/firebase-android-sdk/pull/4284 due to CLA issues (PR author leaving Google) blocking merge on this PR.

PR #4186, opened 2022-10-07
[Fireperf][Fixit] Change `AppStartTrace` from singleton to normal object

leotianlizhan:
De-singleton `AppStartTrace`. This is the easiest picking since it only has one single dependee.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
@@ -68,7 +68,6 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
   private static final int CORE_POOL_SIZE = 0;
   private static final int MAX_POOL_SIZE = 1; // Only need single thread
 
-  private static volatile AppStartTrace instance;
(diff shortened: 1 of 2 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -129,29 +128,17 @@ public static void setLauncherActivityOnResumeTime(String activity) {
...
-                      CORE_POOL_SIZE,
-                      MAX_POOL_SIZE,
-                      /* keepAliveTime= */ MAX_LATENCY_BEFORE_UI_INIT + 10,
-                      TimeUnit.SECONDS,
-                      new LinkedBlockingQueue<>()));
-        }
-      }
-    }
-    return instance;
+  public AppStartTrace() {
```
jeremyjiang-dev, line 131: I am not sure if "De-singletoning this hides AppStartTrace.getInstance() from customers. " Customers can still call new AppStartTrace() I think.

visumickey (commented):
Are there any changes to the unit tests that need to accompany this change? We seem to be changing the way we allow to create/access AppStartTrace.

PR #4191, opened 2022-10-10
Change sendUnsentReports return type to async

nky89:
Change `FirebaseCrashlytics#sendUnsentReports` return type to `Task<Void>`

With  a `Task` type result returned, we can hold and wait the task to be finished if we want.


thatfiredev:
@acntwww I'm curious about your use-case. Why would you want to await on the result of this?

acntwww:
@thatfiredev I init the firebase sdk all by hand with other SDKs  , saying sdk-a. 

```java
private void init() {
  initFirebase();
  initSdkA(); <--- this is where the crash happens
}
```
This crash **may** not be uploaded. So I try to upload the crash info in  `initFirebase` function, like:

```java
private void initFirebase() {
  ...
  Task<Void> unsentReports = firebaseCrashlytics.sendUnsentReports();
  Utils.awaitEvenIfOnMainThread(unsentReports);
  ...
}
```

PR #4207, opened 2022-10-15
Fix Firestore failing to return empty results from the local cache

milaGGL:
Fix a bug where Firestore fails to return results from the local cache if the result set was empty.

This bug is due to the following logic in `QueryListener.java`:
https://github.com/firebase/firebase-android-sdk/blob/ae3bdb895e47b3d9a688dfbfe5ef3e94f2918c59/firebase-firestore/src/main/java/com/google/firebase/firestore/core/QueryListener.java#L143

which assumes that if the result set from the local cache is empty that there is no cached result; however, if the result set of the query is indeed empty then it should be returning that empty result set from the cache.

This fix improves the logic to use the presence of a hasCachedResults flag to indicate that the empty result set is cached data and should be raised to the client.

This bug fix is ported from https://github.com/firebase/firebase-js-sdk/pull/6624, which fixed [#5873](https://github.com/firebase/firebase-js-sdk/issues/5873)

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+Fix Firestore failing to raise initial snapshot from empty local cache result.
 
 # 24.4.0
 * [unchanged] Updated to accommodate the release of the updated
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-Fix Firestore failing to raise initial snapshot from empty local cache result.
+- [fixed] Fixed Firestore failing to raise initial snapshot from empty local cache result.
 
 # 24.4.0
 * [unchanged] Updated to accommodate the release of the updated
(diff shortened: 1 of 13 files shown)


dconeybe:
Please update the description of this PR. Feel free to (mostly) copy the description from https://github.com/firebase/firebase-js-sdk/pull/6624. Also, mention that this is a port of https://github.com/firebase/firebase-js-sdk/pull/6624. This really helps enabling cross linking between PRs.

dconeybe (changes requested):
Looks great! Just two very minor comments, plus changes to the PR's description.

dconeybe:
Oh, and please add a change log entry

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+Fix Firestore failing to raise initial snapshot from empty local cache result.
```
dconeybe, line 2: Prefix this line with `[fixed]`. See line 48 for an example.

dconeybe (changes requested):
Just one tiny fix to the change log. Otherwise, LGTM.

dconeybe (approved):
Approved!

PR #4211, opened 2022-10-17
Implementation: Exception message masking API

mkeeda:
Implementation this issue https://github.com/firebase/firebase-android-sdk/issues/4264


Tolriq:
@mrober by default you initialize in a content provider no one can install an handler before.

Maybe just allowing us to register a callback that can alter / drop the crashes / ANR before next step ?

mrober:
@Tolriq when you install your own uncaught exception handler, you'd want to do it after Crashlytics. Then you can do whatever you want, and propagate that up to Crashlytics.

Tolriq:
@mrober  Hum yes sorry I used the wrong example :( 

I'm manually enabling via `Firebase.crashlytics.setCrashlyticsCollectionEnabled(it)`

It does not seems that there's anything exposed publicly to wait for the handler to actually be set?

PR #4218, opened 2022-10-18
Update layout of Feedback activity.

kaibolay:
image

### DIFF
--- a/firebase-appdistribution/firebase-appdistribution.gradle
+++ b/firebase-appdistribution/firebase-appdistribution.gradle
@@ -64,6 +64,8 @@ dependencies {
     implementation "androidx.browser:browser:1.3.0"
     implementation "androidx.constraintlayout:constraintlayout:2.1.4"
 
+    implementation "com.google.android.material:material:1.6.1"
+
     androidTestImplementation project(':integ-testing')
     androidTestImplementation 'androidx.test.ext:junit:1.1.1'
     androidTestImplementation 'androidx.test:runner:1.2.0'
--- a/firebase-appdistribution/firebase-appdistribution.gradle
+++ b/firebase-appdistribution/firebase-appdistribution.gradle
@@ -64,7 +64,7 @@ dependencies {
     implementation "androidx.browser:browser:1.3.0"
     implementation "androidx.constraintlayout:constraintlayout:2.1.4"
 
-    implementation "com.google.android.material:material:1.6.1"
+    implementation "com.google.android.material:material:1.4.0"
 
     androidTestImplementation project(':integ-testing')
     androidTestImplementation 'androidx.test.ext:junit:1.1.1'
--- a/firebase-appdistribution/firebase-appdistribution.gradle
+++ b/firebase-appdistribution/firebase-appdistribution.gradle
@@ -29,6 +29,9 @@ android {
         multiDexEnabled true
         versionName version
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        vectorDrawables {
+            useSupportLibrary = true
+        }
     }
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
--- a/firebase-appdistribution/firebase-appdistribution.gradle
+++ b/firebase-appdistribution/firebase-appdistribution.gradle
@@ -67,8 +67,6 @@ dependencies {
     implementation "androidx.browser:browser:1.3.0"
     implementation "androidx.constraintlayout:constraintlayout:2.1.4"
 
-    implementation "com.google.android.material:material:1.4.0"
-
     androidTestImplementation project(':integ-testing')
     androidTestImplementation 'androidx.test.ext:junit:1.1.1'
     androidTestImplementation 'androidx.test:runner:1.2.0'
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
@@ -64,11 +64,8 @@ public class FeedbackActivity extends AppCompatActivity {
   }
 
   private void setupView() {
-    setTheme(R.style.Theme_Feedback);
+    setTheme(R.style.FeedbackTheme);
     setContentView(R.layout.activity_feedback);
-
-    TextView senderText = findViewById(R.id.senderText);
-    senderText.setText(getString(R.string.sender, "tester@e.mail"));
 
     TextView infoTextView = this.findViewById(R.id.infoText);
     infoTextView.setText(infoText);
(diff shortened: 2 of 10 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -30,10 +35,9 @@
 
 /** Activity for tester to compose and submit feedback. */
 public class FeedbackActivity extends AppCompatActivity {
-
   private static final String TAG = "FeedbackActivity";
-  private static final int THUMBNAIL_WIDTH = 200;
-  private static final int THUMBNAIL_HEIGHT = 200;
+  private static final int THUMBNAIL_WIDTH = 600;
```
jladieu, line 39: Nit: include units in the name?  I can't tell if this is pixels or percent or something.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -30,10 +35,9 @@
...
+  private static final int THUMBNAIL_HEIGHT = -1; // scale proportionally
```
jladieu, line 40: I don't understand this, even with the comment.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -60,22 +64,35 @@ protected void onCreate(Bundle savedInstanceState) {
...
     if (thumbnail != null) {
-      ImageView screenshotImageView = this.findViewById(R.id.thumbnail);
+      ImageView screenshotImageView = this.findViewById(R.id.screenshotImageView);
       screenshotImageView.setImageBitmap(thumbnail);
+      CheckBox checkBox = findViewById(R.id.screenshotCheckBox);
+      checkBox.setChecked(true);
+      checkBox.setOnClickListener(
+          v -> screenshotImageView.setVisibility(checkBox.isChecked() ? VISIBLE : GONE));
     } else {
       LogWrapper.getInstance().e(TAG, "No screenshot available");
```
jladieu, line 91: Will this text get translated?

jladieu (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
kaibolay: DONE

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
kaibolay: No. But none of the log messages are. They are not shown to the end user.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
kaibolay: It means that the height depends on the width and is adjusted proportionally when scaling. E.g. the original image was 1800px wide and 3000px high. So with a target width of 600px and the proportional target height is determined to be 1000px to keep the image in proportion. A negative target means to keep the proportion. Any advice on how to explain that concisely?

firebase-appdistribution/firebase-appdistribution.gradle
```diff
@@ -64,6 +67,8 @@ dependencies {
     implementation "androidx.browser:browser:1.3.0"
     implementation "androidx.constraintlayout:constraintlayout:2.1.4"
 
+    implementation "com.google.android.material:material:1.4.0"
```
lfkellogg, line 70: Is this just so we can use `Theme.MaterialComponents.Light.NoActionBar`? If so, could we use `Theme.AppCompat.Light.NoActionBar` instead since we're already depending on appcompat?

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
```diff
@@ -1,86 +1,139 @@
...
+        android:id="@+id/divider2"
+        android:layout_width="match_parent"
+        android:layout_height="1dp"
+        android:background="?android:attr/listDivider" />
+
+    <TextView
+        android:id="@+id/infoText"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text=""
```
lfkellogg, line 135: Is this necessary or could we remove it?

firebase-appdistribution/src/main/res/values/strings.xml
```diff
@@ -34,4 +33,12 @@
...
-</resources>
\ No newline at end of file
+  <string name="back_button_description">back</string>
+  <string name="send_feedback">Send feedback</string>
+  <string name="send_button_description">send</string>
+  <string name="sender">From: %1$s</string>
+  <string name="feedback_text_hint">Have feedback? Let us know how we can make this app better.</string>
+  <string name="screenshot">Screenshot</string>
+  <string name="no_screenshot">Screenshot (not available)</string>
+  <string name="screenshot_image_description">screenshot image</string>
```
lfkellogg, line 43: I think the convention for this would be to capitalize it: "Screenshot image"

firebase-appdistribution/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,4 @@
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <style name="Theme.Feedback" parent="Theme.MaterialComponents.Light.NoActionBar">
```
lfkellogg, line 2: Does "Theme.Feedback" get translated into "Theme_Feedback" like the value we set in the activity?

I think maybe we should just go with camelcase ("FeedbackTheme"). That lines up with the docs: https://developer.android.com/develop/ui/views/theming/themes#CustomizeTheme

lfkellogg (commented):
Nice! Could you add a screenshot (or more if you want to capture additional states) in the description? It's nice to be able to compare against the layout code.

firebase-appdistribution/firebase-appdistribution.gradle
kaibolay: Done

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
kaibolay: Done

firebase-appdistribution/src/main/res/values/themes.xml
kaibolay: Done

PR #4221, opened 2022-10-18
Performing IN expansion

ehsannas:


firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteQueryEngineTest.java
```diff
@@ -277,4 +277,94 @@ public void orQueryWithArrayMembershipUsingIndexes() throws Exception {
...
+        expectOptimizedCollectionScan(() -> runQuery(query2, SnapshotVersion.NONE));
+    assertEquals(docSet(query2.comparator()), result2);
+
+    // a IN [0,3] || a IN [0,2] should union them (similar to: a IN [0,2,3]).
+    Query query3 =
+        query("coll")
+            .filter(
+                orFilters(
+                    filter("a", "in", Arrays.asList(0, 3)),
+                    filter("a", "in", Arrays.asList(0, 2))));
```
wu-hui, line 326: Can we add another case with a more complex composite filter, like the ones in `LogicUtilTest`? Same below.

ehsannas: Done

ehsannas:
@wu-hui I decided to add the other tests to this same PR.

I have added tests that demonstrate that the following works:
query with multiple IN operations with and without CSI
query combining IN and array-contains operations with and without CSI
query combining IN and array-contains-any operations with and without CSI
query performing orderBy equality field

ehsannas:
I've posted #4231 to actually remove the restrictions and added integration tests there (those integration tests are expected to fail at this time).

PR #4231, opened 2022-10-21
Remove some query restrictions and add integration tests

ehsannas:
This PR actually removes the restrictions and adds integration tests (which are expected to fail at this time).


ehsannas:
As expected, the newly added tests currently fail against prod

ehsannas:
>Can we have some tests to cover indexed queries as well?

Do you mean the QueryTest integration test sets up an CSI similar to what's done in [IndexingTest](https://cs.opensource.google/firebase-sdk/firebase-android-sdk/+/master:firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java)?

FWIW, I've added the equivalent unit tests for CSI side [here](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteQueryEngineTest.java#L282-L540)

wu-hui (approved):
LGTM!

ehsannas:
Applying the "do not merge" label as relaxing these restrictions should not happen yet.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1134,4 +1134,167 @@ public void testOrQueriesWithArrayMembership() {
         "doc4",
         "doc6");
   }
+
+  @Test
+  public void multipleInOps() {
```
cherylEnkidu, line 1139: Previous functions inside this file use `test` prefix in the name, would it be better to keep this style?

ehsannas: Good idea. Done.

cherylEnkidu (commented):
Last test in `queriesInAndArrayContainsAnyArrayRules` needs to be removed

ehsannas:
@wu-hui would you please review the last commit?

wu-hui (approved):
LGTM!

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ValidationTest.java
```diff
@@ -864,6 +876,54 @@ public void testInvalidQueryFilters() {
...
+        reason);
+  }
+
+  @Test
+  public void testOrderByEqualityWithDisjunctions() {
+    CollectionReference collection = testCollection();
+    String reason =
+        "Performing an equality ('in' and '==') and an inequality (notEqualTo, notIn, "
+            + "lessThan, lessThanOrEqualTo, greaterThan, or greaterThanOrEqualTo) on the same field "
+            + "is not yet allowed in disjunctions.";
```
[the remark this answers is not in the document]
ehsannas: yes, it's considered a blocker for general availability.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -507,16 +507,6 @@ public Target toTarget() {
     return this.memoizedTarget;
   }
 
-  /** Returns true if the query contains any composite filters (AND/OR). Returns false otherwise. */
-  public boolean containsCompositeFilters() {
```
MarkDuckworth, line 511: Was this not used?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1178,4 +1178,171 @@ public void testOrQueriesWithArrayMembership() {
...
+        collection.where(
+            Filter.and(
+                Filter.inArray("a", asList(1, 3)),
+                Filter.or(
+                    Filter.inArray("a", asList(0, 2)),
+                    Filter.and(
+                        Filter.greaterThanOrEqualTo("b", 1), Filter.inArray("a", asList(1, 3))))));
+    checkOnlineAndOfflineResultsMatch(query6, "doc3", "doc4");
+
+    // Nested composite filter on different fields.
```
MarkDuckworth, line 1239: This seems very similar to "Nested composite filter on the same field." But, it's not a big deal.

MarkDuckworth (approved):
LGTM

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
ehsannas: There was a PR that removed the one usage of this function, but that PR failed to recognize that that was the only usage and forgot to remove this function along with it.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
ehsannas: the inner-most IN filter is on field `a` in both cases. The parent IN Filters in the former is also on `a`, but on `b` in the latter. Both should be valid queries.

PR #4232, opened 2022-10-22
Firestore: Add test that verifies count query error message when missing index

dconeybe:
This verifies that the descriptive error message that occurs when the server does not have the required index has the URL that customers can follow to create the index.


milaGGL (approved):
LGTM.

PR #4242, opened 2022-10-25
Migrate fireperf content provider to component

emilypgoogle:
Some tests are still flaky, some are not yet written, currently a prototype.


firebase-common/src/main/java/com/google/firebase/StartupTime.java
```diff
@@ -0,0 +1,50 @@
...
+public abstract class StartupTime {
+
+  /** @return The epoch time that Firebase began initializing, in milliseconds */
+  public abstract long getEpochMillis();
+
+  /**
+   * @return The number of nanoseconds from the start of the program to when Firebase began
+   *     initializing, measured by the JVM
+   */
+  public abstract long getStartupNanos();
```
leotianlizhan, line 36: Can we use [SystemClock.elapsedRealtime()](https://developer.android.com/reference/android/os/SystemClock)? This is because Fireperf is going to use [Process.getStartElapsedRealtime()](https://developer.android.com/reference/android/os/Process#getStartElapsedRealtime()) for our API 24+ app-start-trace, and that system API is in the `elapsedRealtime` timebase. Generally I find that `elapsedRealtime` is more widely supported in Android and that's also what 1P Google teams uses, such as PRIMES and TikTok.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -163,9 +166,16 @@ static AppStartTrace getInstance(TransportManager transportManager, Clock clock)
     this.clock = clock;
     this.configResolver = configResolver;
     this.executorService = executorService;
+
+    StartupTime startupTime = FirebaseApp.getInstance().get(StartupTime.class);
+    if (startupTime == null) {
+      firebaseStartupTime = new Timer();
+    }
+    firebaseStartupTime =
+        Timer.ofElapsedRealtime(TimeUnit.NANOSECONDS.toMillis(startupTime.getStartupNanos()));
```
leotianlizhan, line 175: `SystemClock.elapsedRealtime()` and `System.nanoTime()` aren't in the same timebase, so assigning one to the other would not work (iirc nanoTime is when VM starts, elapsedRealtime is the last time the device booted). For context on why Fireperf chose `elapsedRealtime` see the other earlier comment.

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -426,15 +426,21 @@ protected FirebaseApp(Context applicationContext, String name, FirebaseOptions o
...
             .addComponent(Component.of(applicationContext, Context.class))
             .addComponent(Component.of(this, FirebaseApp.class))
             .addComponent(Component.of(options, FirebaseOptions.class))
-            .setProcessor(new ComponentMonitor())
-            .build();
+            .setProcessor(new ComponentMonitor());
+
+    // Don't provide StartupTime in direct boot mode
+    if (UserManagerCompat.isUserUnlocked(applicationContext)) {
+      builder.addComponent(Component.of(StartupTime.now(), StartupTime.class));
```
vkryachko, line 440: I think calling `StartupTime.now()` here is a bit too late, as a bunch of things already happened that take non-negligible amount of time, ideally we should record the value as early as FirebaseInitProvider is loaded(i.e. in a static field.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfEarly.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>Responsible for initializing the AppStartTrace, and early initialization of ConfigResolver
+ */
+public class FirebasePerfEarly {
+  private final Handler mainHandler = new Handler(Looper.getMainLooper());
+
+  @NonNull
+  public static FirebasePerfEarly getInstance() {
+    return FirebaseApp.getInstance().get(FirebasePerfEarly.class);
+  }
```
vkryachko, line 39: why do we need this method?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -54,6 +55,13 @@ public List<Component<?>> getComponents() {
             .add(Dependency.requiredProvider(TransportFactory.class))
             .factory(FirebasePerfRegistrar::providesFirebasePerformance)
             .build(),
+        Component.builder(FirebasePerfEarly.class)
+            .name(LIBRARY_NAME)
+            .add(Dependency.required(FirebaseApp.class))
+            .add(Dependency.optionalProvider(StartupTime.class))
+            .alwaysEager()
```
vkryachko, line 62: eagerInDefaultApp instead? iiuc fireperf does not support non-default apps

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
```diff
@@ -93,8 +94,12 @@ private RemoteConfigManager() {
         firebaseRemoteConfig == null
             ? new ConcurrentHashMap<>()
             : new ConcurrentHashMap<>(firebaseRemoteConfig.getAll());
-    this.appStartTimeInMs =
-        TimeUnit.MICROSECONDS.toMillis(FirebasePerfProvider.getAppStartTime().getMicros());
+    StartupTime startupTime = FirebaseApp.getInstance().get(StartupTime.class);
```
vkryachko, line 97: can this be provided as a dependency instead of looking it up from firebaseApp?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfEarly.java
emilypgoogle: We don't, I've seen the pattern elsewhere so I mirrored it, removed it

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -416,6 +416,7 @@ protected FirebaseApp(Context applicationContext, String name, FirebaseOptions o
     this.applicationContext = Preconditions.checkNotNull(applicationContext);
     this.name = Preconditions.checkNotEmpty(name);
     this.options = Preconditions.checkNotNull(options);
+    StartupTime startupTime = StartupTime.now();
```
vkryachko, line 419: Not sure if I made this comment before but I think it's too late to record the start up time at this point. Can we instead capture it when FirebaseInitProvider initializes?

Additionally we want to make sure we only capture this value if we are "auto-initializing" i.e. if the developer manually initializes FirebaseApp we would not treat it as app starting up.

emilypgoogle: Sure, I initially misinterpreted and thought you wanted it earlier in the constructor.

firebase-perf/src/main/java/com/google/firebase/perf/config/RemoteConfigManager.java
emilypgoogle: I've attempted to resolve this but it's more just shuffling the method around to be cleaner, `RemoteConfigManager` initializes itself during class init and I believe it'd be more of a refactor to change this behavior if passing from truly outside of the class is the goal.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -54,6 +55,13 @@ public List<Component<?>> getComponents() {
             .add(Dependency.requiredProvider(TransportFactory.class))
             .factory(FirebasePerfRegistrar::providesFirebasePerformance)
             .build(),
+        Component.builder(FirebasePerfEarly.class)
```
vkryachko, line 58: should FirebasePerformance component take a dependency on FirebasePerfEarly to make sure it gets initialized before FirebasePerformance?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -54,6 +55,13 @@ public List<Component<?>> getComponents() {
...
+            .name(LIBRARY_NAME)
```
vkryachko, line 59: are we ok to have the same name for this component? i.e. it could be confusing when looking at startup traces. Given that FirebasePerformance itself is not eager, might make sense to make it unnamed, wdyt?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
emilypgoogle: Being eager I'm not sure it's strictly necessary, but it's good practice and trivial, so added.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
emilypgoogle: As a shim only giving the name to the early component makes sense, but it was easy enough to just give them unique names.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
vkryachko: Note that in order to actually cause initialization you need to call `container.get(FirebasePerfEarly.class` inside providesFirebasePerformance, declaring the dep will not do it automatically as it's only used to detect dependency cycles

vkryachko:
Looks like there is a merge conflict?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfEarly.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * <p>Responsible for initializing the AppStartTrace, and early initialization of ConfigResolver
+ */
+public class FirebasePerfEarly {
+  private final Handler mainHandler = new Handler(Looper.getMainLooper());
+
+  public FirebasePerfEarly(FirebaseApp app) {
+    Context context = app.getApplicationContext();
+
+    // Initialize ConfigResolver early for accessing device caching layer.
```
leotianlizhan, line 38: @vkryachko I wonder if we could merge the 2-part initialization (`FirebasePerfEarly` + `FirebasePerfRegistrar`) to simplify it further in the future. The reason we needed `FirebasePerfProvider` was to capture app-start time, but now that's no longer needed, and if Fireperf is completely started by component framework, if we get rid of `FirebasePerfEarly` then we also don't need `ConfigResolver` anymore. That's because `ConfigResolver` and its related classes were created when the old `FirebasePerfProvider` early initializes part of Fireperf that needed FRC, so we had to build our own disk cache for RC. If Fireperf starts after FRC as it should, we can just use FRC. However I'm not sure if this will be okay for the SDK-Observability-project, would it be fine for Fireperf's dependencies to start first?

leotianlizhan (commented):
It seems some unit tests are failing, otherwise the rest LGTM

firebase-common/src/main/java/com/google/firebase/provider/FirebaseInitProvider.java
```diff
@@ -27,11 +27,16 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /** Initializes Firebase APIs at app startup time. */
 public class FirebaseInitProvider extends ContentProvider {
-
   private static final String TAG = "FirebaseInitProvider";
+  @Nullable
+  public static StartupTime startupTime = StartupTime.now();
+  @NonNull
+  public static AtomicBoolean currentlyInitializing = new AtomicBoolean(false);
```
vkryachko, line 39: I think this could benefit from some encapsulation not to allow direct modification, might make sense to expose this as say `public static boolean isCurrentlyInitializing()` and making the atomic private, wdyt?

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfEarly.java
vkryachko: I don't have full context on why the 2-part logic was in place, but my understanding that part of it was to speed up the SDK and delay a lot of work as much as possible(i.e. see #2518).

But I trust your judgement to evolve this in a way that makes sense, after this initial migration. Happy to discuss it more.

firebase-common/src/main/java/com/google/firebase/StartupTime.java
```diff
@@ -0,0 +1,58 @@
...
+import androidx.annotation.NonNull;
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+/**
+ * Represents the time at which Firebase began initialization, both in unix time/epoch milliseconds
+ * and in milliseconds since boot. The absence of a StartupTime indicates an unreliable or
+ * misleading time, such as a launch in direct boot mode. Because of this, StartupTime cannot be
+ * guaranteed to be present, and instead should be optionally depended on, and its absence handled.
+ */
```
vkryachko, line 27: Nit: please write a concise summary fragment: go/java-practices/javadoc#summary_fragment

firebase-common/src/main/java/com/google/firebase/StartupTime.java
```diff
@@ -0,0 +1,58 @@
...
+@AutoValue
+public abstract class StartupTime {
+
+  public static @Nullable StartupTime getInstance() {
+    return FirebaseApp.getInstance().get(StartupTime.class);
+  }
```
vkryachko, line 33: I'd avoid coupling this value class to FirebaseApp, as this method is not safe to call when the default app is not initialized. i.e. it will throw IllegalStateException. Ideally SDKs should just inject StartupTime as a component, and given that it's not easy in the case of fireperf, let's not make it very convenient to work around components by providing such static methods.

I suggest introducing such a static method inside the perf sdk, or even just call `FirebaseApp.getInstance().get(StartupTime.class)` directly in places where startup time is needed.

vkryachko: Yeah, here's an example of how to silence that error:
https://github.com/firebase/firebase-android-sdk/blob/94ae2ac6fbd7971cbde2ba41824f24248e2ed5e7/firebase-messaging/src/main/java/com/google/firebase/messaging/MessagingAnalytics.java#L198-L199

fyi this is a custom check under https://github.com/firebase/firebase-android-sdk/blob/94ae2ac6fbd7971cbde2ba41824f24248e2ed5e7/tools/errorprone/src/main/java/com/google/firebase/errorprone/ComponentsAppGetCheck.java#L51-L57

firebase-common/src/main/java/com/google/firebase/StartupTime.java
vkryachko: and please add `@hide` as the last line of the javadoc, this will suppress public javadocs for this type.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -77,4 +89,10 @@ private static FirebasePerformance providesFirebasePerformance(ComponentContaine
 
     return component.getFirebasePerformance();
   }
+
+  @SuppressWarnings("FirebaseUseExplicitDependencies")
+  private static FirebasePerfEarly providesFirebasePerformanceEarly(ComponentContainer container) {
+    return new FirebasePerfEarly(
+        container.get(FirebaseApp.class), container.get(FirebaseApp.class).get(StartupTime.class));
```
vkryachko, line 96: It's not recommended to get things from FirebaseApp, the error you are seeing wrt "undeclared dep" is due to the fact that you're declaring an `optionalProvider` dep but are then calling container.get() instead of `container.getProvider()`, so you should either switch to use `optional() + c.get()` or `optionalProvider() + c.getProvider()`

firebase-common/src/main/java/com/google/firebase/provider/FirebaseInitProvider.java
```diff
@@ -27,11 +27,23 @@
...
+  @NonNull private static AtomicBoolean currentlyInitializing = new AtomicBoolean(false);
+
+  /** @hide */
+  public static @Nullable StartupTime getStartupTime() {
+    return startupTime;
+  }
+
+  public static boolean isCurrentlyInitializing() {
+    return currentlyInitializing.get();
+  }
```
vkryachko, line 46: can we make this `@hide` as well?

vkryachko (approved):

leotianlizhan (approved):

PR #4248, opened 2022-10-26
Merge fad/next into fad/iaf

lfkellogg:


firebase-appdistribution-api/CHANGELOG.md
```diff
@@ -1 +1,56 @@
 # Unreleased
```
kaibolay, line 1: How does this file work?

firebase-appdistribution/CHANGELOG.md
```diff
@@ -1,17 +1,57 @@
 # Unreleased
```
kaibolay, line 1: How does this file work?

kaibolay (approved):

firebase-appdistribution-api/CHANGELOG.md
lfkellogg: We will need to update this file as we merge changes into the `master` branch (see go/firebase-android/docs/workflow#release-notes-aka-changelog). To avoid having to retroactively write these we should probably start adding to this as we merge into our feature branches as well, even if the presubmit check doesn't run on those PRs.

PR #4251, opened 2022-10-27
fix crash

kaibolay:


lfkellogg (approved):
LGTM but how does this fix a crash?

kaibolay:
> LGTM but how does this fix a crash?

The `Button` can end up being an `AppCompatButton` and for some reason that gave me a runtime error even though `AppCompatButton` extends `Button`...

kaibolay:
> LGTM but how does this fix a crash?

The `Button` can end up being an `AppCompatButton`

PR #4253, opened 2022-10-27
Clean up shake detection trigger example

lfkellogg:
Renames it to be more aligned with the other triggers, adds javadocs, and reformats the file.


kaibolay:
Shouldn't this be merged info `fad/in-app-feedback`?

kaibolay (approved):

PR #4255, opened 2022-10-28
Simplify custom notification trigger example and move to activity

lfkellogg:
image

### DIFF
--- a/firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/CustomNotificationFeedbackTrigger.kt
+++ b/firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/CustomNotificationFeedbackTrigger.kt
@@ -6,16 +6,12 @@ import android.app.*
 import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager.PERMISSION_DENIED
-import android.content.pm.PackageManager.PERMISSION_GRANTED
 import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.util.Log
-import androidx.activity.result.ActivityResultCaller
-import androidx.activity.result.contract.ActivityResultContracts
-import androidx.annotation.RequiresApi
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.content.ContextCompat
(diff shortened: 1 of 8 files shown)


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/CustomNotificationFeedbackTrigger.kt
```diff
@@ -6,204 +6,113 @@ import android.app.*
...
-      Log.i(TAG, "Already has permission.")
-      return
-    }
-
-    if (hasRequestedPermission) {
-      Log.i(TAG, "Already request permission; Not trying again.")
-      return
-    }
+  fun showNotification(activity: Activity) {
+    synchronized(this) {
```
kaibolay, line 44: why?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/CustomNotificationFeedbackTrigger.kt
```diff
@@ -6,204 +6,113 @@ import android.app.*
...
-    notificationManager.cancel(FEEDBACK_NOTIFICATION_ID)
-  }
-
-  override fun onActivityResumed(activity: Activity) {
-    if (isEnabled) {
-      if (activity !is TakeScreenshotAndTriggerFeedbackActivity) {
-        Log.d(TAG, "setting current activity")
-        activityToScreenshot = activity
+  fun cancelNotification() {
+    synchronized(this) {
```
kaibolay, line 96: why?

firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/CustomNotificationFeedbackTrigger.kt
```diff
@@ -6,204 +6,113 @@ import android.app.*
...
 
-class TakeScreenshotAndTriggerFeedbackActivity : Activity() {
+class CustomNotificationTakeScreenshotActivity : Activity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    val activity = CustomNotificationFeedbackTrigger.activityToScreenshot
-    if (activity == null) {
-      Log.e(TAG, "Can't take screenshot because activity is unknown")
-      return
+    synchronized(CustomNotificationFeedbackTrigger) {
```
kaibolay, line 108: Why? Also, I think this synchronizes on the class - not the instance like the other two `synchronize` blocks.

Maybe this is an attempt to somehow protect modifications to `activityToScreenshot`, but I think read/writes to reference variables are [atomic](https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html). 

The assignment to the local variable `activity` on line 109 makes sure that the value doesn't change between the `null` check on line 110 and line 114. Inlining `CustomNotificationFeedbackTrigger.activityToScreenshot` without synchronization would indeed be a problem.

So I think we should add a comment to the assignment on line 109 talking about this issue.

kaibolay (approved):

PR #4257, opened 2022-10-28
Fix FAILED_PRECONDITION when writing to a deleted document in a transaction

milaGGL:
This fix is ported from Web SDK https://github.com/firebase/firebase-js-sdk/pull/6550, and its follow-up PR https://github.com/firebase/firebase-js-sdk/pull/6729.

### DIFF
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -13,6 +13,7 @@
 The Kotlin extensions library transitively includes the updated
   `firebase-firestore` library. The Kotlin extensions library has the following
   additional updates:
+
 * [feature] Firebase now supports Kotlin coroutines.
   With this release, we added
   [`kotlinx-coroutines-play-services`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-play-services/){: .external}
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
-- [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction. Ported from web SDK.
-  See the [issue report in GitHub](https://github.com/firebase/firebase-js-sdk/issues/5871).
+- [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction.
+  (#5871).
+
 - [fixed] Fixed Firestore failing to raise initial snapshot from empty local cache result.
 
 # 24.4.0
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction. Ported from web SDK  (#5871).JMASW N
 - [fixed] Fixed Firestore failing to raise initial snapshot from empty local cache result.
 
 # 24.4.0
@@ -10,7 +11,7 @@
 The Kotlin extensions library transitively includes the updated
   `firebase-firestore` library. The Kotlin extensions library has the following
   additional updates:
-
+SDEW
 * [feature] Firebase now supports Kotlin coroutines.
   With this release, we added
   [`kotlinx-coroutines-play-services`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-play-services/){: .external}
--- a/firebase-firestore/CHANGELOG.md
+++ b/firebase-firestore/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
-- [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction. Ported from web SDK  (#5871).JMASW N
+- [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction. Ported from web SDK.
+  See the [issue report in GitHub](https://github.com/firebase/firebase-js-sdk/issues/5871).
 - [fixed] Fixed Firestore failing to raise initial snapshot from empty local cache result.
 
 # 24.4.0
@@ -11,7 +12,6 @@
 The Kotlin extensions library transitively includes the updated
   `firebase-firestore` library. The Kotlin extensions library has the following
   additional updates:
-SDEW
 * [feature] Firebase now supports Kotlin coroutines.
   With this release, we added
   [`kotlinx-coroutines-play-services`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-play-services/){: .external}
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
@@ -82,8 +82,8 @@ public class TransactionTest {
    * Used for testing that all possible combinations of executing transactions result in the desired
    * document value or error.
    *
-   * <p>`run()`, `withExistingDoc()`,`withNonexistentDoc()` and `withDeletedDoc()` don't actually do
-   * anything except assign variables into the TransactionTester.
+   * <p>`run()`, `withExistingDoc()`, `withNonexistentDoc()` and `withDeletedDoc()` don't actually
+   * do anything except assign variables into the TransactionTester.
    *
    * <p>`expectDoc()`, `expectNoDoc()`, and `expectError()` will trigger the transaction to run and
    * assert that the end result matches the input.
@@ -188,7 +188,7 @@ public class TransactionTest {
           waitFor(docRef.delete());
           break;
         default:
-          throw new Error("invalid fromDocumentType: " + fromDocumentType);
+          throw new RuntimeException("invalid fromDocumentType: " + fromDocumentType);
       }
     }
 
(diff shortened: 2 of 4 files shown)


firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Unreleased
+- [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction. Ported from web SDK.
```
dconeybe, line 2: nit: Remove "Ported from web SDK". That is useful in the PR description but not in the change log entry.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
...
+  See the [issue report in GitHub](https://github.com/firebase/firebase-js-sdk/issues/5871).
```
dconeybe, line 3: Replace

```
See the [issue report in GitHub](https://github.com/firebase/firebase-js-sdk/issues/5871).
```

with simply

```
(#5871).
```

(see line 78 below for an example).

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -69,20 +69,29 @@ void runStage(Transaction transaction, DocumentReference docRef)
...
+    DELETED,
+  }
+
   /**
    * Used for testing that all possible combinations of executing transactions result in the desired
    * document value or error.
    *
-   * <p>`run()`, `withExistingDoc()`, and `withNonexistentDoc()` don't actually do anything except
-   * assign variables into the TransactionTester.
+   * <p>`run()`, `withExistingDoc()`,`withNonexistentDoc()` and `withDeletedDoc()` don't actually do
```
dconeybe, line 85: nit: Add a space after this comma:

```
,`withNonexistentDoc()`
```

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -160,8 +175,20 @@ private void expectError(Code expected) {
...
+          break;
+        case NON_EXISTENT:
+          // Nothing to do; document does not exist.
+          break;
+        case DELETED:
+          waitFor(docRef.set(map("foo", "bar0")));
+          waitFor(docRef.delete());
+          break;
+        default:
+          throw new Error("invalid fromDocumentType: " + fromDocumentType);
```
dconeybe, line 191: `Error` in Java is rarely, if ever used directly. Use `RuntimeException` instead.

`Error` typically indicates an irrecoverable internal error inside the JVM, such as out of memory or invalid opcode.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -637,6 +705,29 @@ public void testDoesNotRetryOnPermanentError() {
...
+    DocumentReference doc = firestore.collection("foo").document();
+    CountDownLatch latch = new CountDownLatch(2);
+    waitFor(
+        firestore.runTransaction(
+            transaction -> {
+              latch.countDown();
+              transaction.get(doc);
+              // Do a write outside of the transaction.
+              waitFor(doc.set(map("count", 1)));
+              // Now try to set the doc within the transaction.This should fail once
```
dconeybe, line 720: nit: Add a space after this period:

```
.This
```

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -637,6 +705,29 @@ public void testDoesNotRetryOnPermanentError() {
     assertEquals(1, count.get());
   }
 
+  @Test
+  public void testRetryOnAlreadyExistsError() {
+    final FirebaseFirestore firestore = testFirestore();
+    DocumentReference doc = firestore.collection("foo").document();
+    CountDownLatch latch = new CountDownLatch(2);
```
dconeybe, line 712: Use `AtomicInteger` instead of `CountDownLatch`.

`CountDownLatch` is used to coordinate the order of execution of two different threads. But you are just using as a counter, so `AtomicInteger` is what you want.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -637,6 +705,29 @@ public void testDoesNotRetryOnPermanentError() {
...
+    waitFor(
+        firestore.runTransaction(
+            transaction -> {
+              latch.countDown();
+              transaction.get(doc);
+              // Do a write outside of the transaction.
+              waitFor(doc.set(map("count", 1)));
```
dconeybe, line 719: Shouldn't the test skip writing to the document on the 2nd invocation? Otherwise, wouldn't the 2nd invocation? Otherwise, wouldn't the 2nd transaction commit also fail and get retried?

dconeybe (changes requested):

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
milaGGL: hmm 🤔 , correct me if I am wrong. 
In the first attempt, transaction.get() gets the reference along with "doc doesn't exist"status. We create the doc outside of the transaction, and then try transaction.set(). Here, transaction still thinks "doc doesn't  exist", and it will be rejected the with the "already-exists" error due to the conflict.
In the retry, transaction.get() gets the reference to the doc, which exists this time, we update the doc outside the transaction, and then transaction tries transaction.set(). As there are no conflicts of existence this time, it is going to succeed.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -709,23 +709,23 @@ public void testDoesNotRetryOnPermanentError() {
   public void testRetryOnAlreadyExistsError() {
     final FirebaseFirestore firestore = testFirestore();
     DocumentReference doc = firestore.collection("foo").document();
-    CountDownLatch latch = new CountDownLatch(2);
+    AtomicInteger retryCount = new AtomicInteger(0);
```
dconeybe, line 712: As a rule, I always recommend to avoid naming a variable with a number of "retries", `retryCount` in this case. The reason is that it is not obvious whether or not it includes the first attempt, or if it's only counting the subsequent "retries". Please choose another name for this variable, such as `transactionCallbackCount` or `numTransactionCallbackInvocations` or something better.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -709,23 +709,23 @@ public void testDoesNotRetryOnPermanentError() {
...
     waitFor(
         firestore.runTransaction(
             transaction -> {
-              latch.countDown();
+              retryCount.getAndIncrement();
```
dconeybe, line 716: Save the result of `retryCount.getAndIncrement()` to a local variable, and use that later on, rather than invoking more methods on `retryCount`, such as `retryCount.get()`.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -709,23 +709,23 @@ public void testDoesNotRetryOnPermanentError() {
...
               transaction.get(doc);
               // Do a write outside of the transaction.
-              waitFor(doc.set(map("count", 1)));
-              // Now try to set the doc within the transaction.This should fail once
+              if (retryCount.get() == 1) waitFor(doc.set(map("count", retryCount.get())));
+              // Now try to set the doc within the transaction. This should fail once
               // with ALREADY_EXISTS error.
-              transaction.set(doc, map("count", 2));
+              transaction.set(doc, map("count", 22));
```
dconeybe, line 722: Instead of setting `map("count", 22)` here, can you set a different, completely unrelated field? Otherwise, it incorrectly implies that it is somehow related. But, in fact, it doesn't matter. e.g. `map("foo", "bar")`.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
```diff
@@ -637,6 +705,29 @@ public void testDoesNotRetryOnPermanentError() {
...
+    final FirebaseFirestore firestore = testFirestore();
+    DocumentReference doc = firestore.collection("foo").document();
+    AtomicInteger retryCount = new AtomicInteger(0);
+    waitFor(
+        firestore.runTransaction(
+            transaction -> {
+              retryCount.getAndIncrement();
+              transaction.get(doc);
+              // Do a write outside of the transaction.
+              if (retryCount.get() == 1) waitFor(doc.set(map("count", retryCount.get())));
```
dconeybe, line 719: Please use a "dummy" key/value pair here instead of `map("count", retryCount.get())`, which seems to imply that the value of `retryCount` is relevant.

e.g. `map("foo1", "bar1")`

Then maybe use `map("foo2", "bar2")` to `transaction.set()` also. What I'm looking for is a way for these values to clearly suggest that the key/value pair itself is completely irrelevant, and that the values set by `doc.set()` is completely independent from those set by `transaction.set()`.

dconeybe (changes requested):
Just a few more minor suggestions. Looks great!

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/TransactionTest.java
dconeybe: In case anyone is following this conversation, the resolution was that the subsequent call to `doc.set(map("count", 1))` was elided by the backend and did not affect the document's `updateTime`. As a result, the test actually passed. If the value `1` was changed to something dynamic, like `retryCount.get()`, then the test would fail.

dconeybe:
Oh also, please remove `Fixes #5871` from the PR description, since it does not, in fact, fix an issue in the other repository. And add a link to the PRs (there are two of them) that are being ported in this PR.

dconeybe (approved):
Ship it!

PR #4259, opened 2022-10-28
Fix flakiness in MessagingAnalyticsRoboTest.java

willxu-google:
Properly reset SharedPreferences test state in between each test.


gsakakihara (approved):
The coverage test failure appears to be in an unrelated project and I don't see how this change could be causing it, but you should check to make sure.

> Did you forget to add a changelog entry? (Add the 'no-changelog' label to the PR to silence this warning.)

This probably should be fixed, just adding the 'no-changelog' label since this is a test-only change that shouldn't be included in the release notes.

PR #4272, opened 2022-11-02
Early return in getDocumentsMatchingTarget

cherylEnkidu:
#no-changelog


ehsannas (commented):
getFieldIndex is actually a bit of an expensive function, and I'd rather not call it twice. We can either not change this code at all (it's a very minor difference from the iOS implementation). Or, you can do what you did in the iOS sdk: use `List<Pair<Target,FieldIndex>>`

ehsannas (approved):

PR #4277, opened 2022-11-03
Upgrade GRPC tp 1.50.2

tom-andersen:


firebase-firestore/CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
 # Unreleased
 - [fixed] Fix FAILED_PRECONDITION when writing to a deleted document in a transaction.
-  (#5871).
-
-- [fixed] Fixed Firestore failing to raise initial snapshot from empty local cache result.
-- [fixed] Removed invalid suggestions to use `GenericTypeIndicator` from error messages.
+  ([Issue #5871](//github.com/firebase/firebase-android-sdk/issues/5871))
```
dconeybe, line 3: You can just leave the issue/PR links in this format: `(#5871)`. No need to make them hyperlinks. At least, that's what's done elsewhere in this file. When you create the CL with the release notes that get published on the web, those should be clickable hyperlinks.

dconeybe:
nit: There is a typo in the PR title: "tp" should be "to".

firebase-firestore/CHANGELOG.md
tom-andersen: There are examples in changelog where issues have a complete hyperlink. That is where I picked up style by copy/paste. So given the inconsistency, what is best? Making the link takes a little bit more effort, but not much. I would be happy to raise the bar to this level.

When I make the CL, I do so by copying the changelog directly. So having them be the same is easiest in my opinion.

dconeybe: I like it. Copypasteability is great so I support "raising the bar". Plus, it makes the links clickable when you look at the rendered markdown in the GitHub Web UI.

wu-hui (approved):

PR #4283, opened 2022-11-04
[Fireperf][AASA] Additional experiments for `_experiment_as_ttid`

leotianlizhan:
These experiments helps single out certain possibilities from the outliers we see in `_experiment_as_ttid`:
1. `SystemClock.uptimeMillis()` is another monotonic timebase, but excluding deep sleep. This helps single out if the app was entering deep sleep thus causing massive duration
2. `OnPreDraw` is what PRIMES uses for all of 1P Google, thus with lots of success. Just looking at onDraw there were many bugs and hoops the code has to deal with, and `onPreDraw` just looks more stable, so perhaps it can be a bug or something else there as well.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
@@ -383,7 +383,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
 
   @Override
   public void onActivityPaused(Activity activity) {
-    if (this.firstDrawDone != null && this.preDraw != null) {
+    if (isExperimentTraceDone()) {
       return;
     }
     Timer onPauseTime = clock.getTime();
@@ -397,7 +397,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
 
   @Override
   public void onActivityStopped(Activity activity) {
-    if (this.firstDrawDone != null && this.preDraw != null) {
+    if (isExperimentTraceDone()) {
       return;
     }
     Timer onStopTime = clock.getTime();
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
@@ -99,7 +99,6 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
   private Timer onResumeTime = null;
   private Timer firstDrawDone = null;
   private Timer preDraw = null;
-  private List<TraceMetric> preDrawSubtraces = new ArrayList<>();
 
   private PerfSession startSession;
   private boolean isStartedFromBackground = false;
@@ -236,7 +235,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
     this.experimentTtid.addPerfSessions(this.startSession.build());
 
     if (subtraceCount > 0) {
-      executorService.execute(() -> this.logColdStart(this.experimentTtid));
+      executorService.execute(() -> this.logExperimentTtid(this.experimentTtid));
 
       if (isRegisteredForLifecycleCallbacks) {
         // After AppStart trace is queued to be logged, we can unregister this callback.
@@ -267,7 +266,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
     this.experimentTtid.addSubtraces(subtrace.build());
 
     if (subtraceCount > 0) {
-      executorService.execute(() -> this.logColdStart(this.experimentTtid));
+      executorService.execute(() -> this.logExperimentTtid(this.experimentTtid));
 
       if (isRegisteredForLifecycleCallbacks) {
         // After AppStart trace is queued to be logged, we can unregister this callback.
@@ -342,7 +341,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks {
     }
   }
 
-  private void logColdStart(TraceMetric.Builder metric) {
+  private void logExperimentTtid(TraceMetric.Builder metric) {
     transportManager.log(metric.build(), ApplicationProcessState.FOREGROUND_BACKGROUND);
   }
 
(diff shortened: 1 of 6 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/util/PreDrawListener.java
```diff
@@ -0,0 +1,51 @@
...
+// limitations under the License.
+package com.google.firebase.perf.util;
+
+import android.os.Handler;
+import android.os.Looper;
+import android.view.View;
+import android.view.ViewTreeObserver;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class PreDrawListener implements ViewTreeObserver.OnPreDrawListener {
```
raymondlam, line 22: nit: can you add a javadoc for the class?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -205,13 +210,68 @@ private void recordFirstDrawDone() {
...
+
+    subtrace = TraceMetric.newBuilder();
+    subtrace
+        .setName("_experiment_preDraw_uptimeMillis")
+        .setClientStartTimeUs(start.getMicros())
+        .setDurationUs(start.getDurationUptimeMicros(this.preDraw));
+    this.experimentTtid.addSubtraces(subtrace.build());
+
+    if (subtraceCount > 0) {
+      executorService.execute(() -> this.logExperimentTtid(this.experimentTtid));
```
raymondlam, line 269: What does it mean to log this.experimentTtid twice?

Once in `recordFirstDrawDone` with 2 subtraces and then `recordFirstDrawDonePreDraw` with 2 more (total of 4)?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -205,13 +210,68 @@ private void recordFirstDrawDone() {
...
+    subtrace = TraceMetric.newBuilder();
+    subtrace
+        .setName("_experiment_uptimeMillis")
+        .setClientStartTimeUs(start.getMicros())
+        .setDurationUs(start.getDurationUptimeMicros(this.firstDrawDone));
+    this.experimentTtid.addSubtraces(subtrace.build());
+
+    this.experimentTtid.addPerfSessions(this.startSession.build());
+
+    if (subtraceCount > 0) {
```
raymondlam, line 237: Is this ever false? (similarly for recordFirstDrawDonePreDraw)

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -205,13 +210,68 @@ private void recordFirstDrawDone() {
...
+      executorService.execute(() -> this.logExperimentTtid(this.experimentTtid));
+
+      if (isRegisteredForLifecycleCallbacks) {
```
raymondlam, line 240: Should this be outside the `if (subtraceCount > 0) {` (if it's still necessarily)

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: This if block of code is because I am not sure if the order with which the "`postAtFrontOfQueue` in `onDraw`" and the "`post` in `onPreDraw`" is deterministic. `subtraceCount` is taken at the beginning of the method, so if the other method already was called, it would not be zero, so we log the trace. However, if the other method was not called, then the current method is the first one to be invoked, it shouldn't log the trace because it still needs to wait for the other method.

raymondlam: I see. It would be more clear to use either a class variable like hasLoggedExperiment and a comment to make it more obvious what is being done here.

firebase-perf/src/main/java/com/google/firebase/perf/util/PreDrawListener.java
leotianlizhan: Done

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: Addressed in comment above

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: I don't think so because we want to unregister callbacks when we finish the `_experiment` trace, which is when everything is recorded and ready.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -205,16 +210,73 @@ private void recordFirstDrawDone() {
...
+        // After AppStart trace is queued to be logged, we can unregister this callback.
+        unregisterActivityLifecycleCallbacks();
+      }
+    }
+  }
+
+  private boolean isExperimentTraceDone() {
+    return this.preDraw != null && this.firstDrawDone != null;
```
raymondlam, line 277: Just to confirm, we are waiting for both preDraw and firstDrawDone to be captured before we send the experiment trace.

If they are on different threads, I think you'll need to synchronize this check and set the preDraw/firstDrawDone after the subtraces have been added.

raymondlam (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: They are all on the main thread. The 4 places this is being called is `mainThreadHandler.post(...)`, `mainThreadHandler.postAtFrontOfQueue(...)`, `onActivityPaused` and `onActivityStopped`, all of which happen in the same main thread sequentially.

raymondlam: Thanks for confirming.

PR #4287, opened 2022-11-05
Initial version of the build release artifacts workflow.

rlazo:
This version of the workflow hardcodes the list of sdks to release to include Firestore only. Also, generated javadoc is not yet equivalent to what's generated by the regular release process.

### DIFF
--- a/.github/workflows/build-release-artifacts.yml
+++ b/.github/workflows/build-release-artifacts.yml
@@ -7,7 +7,7 @@ on:
       - 'releases/**'
 
 jobs:
-  create-branches:
+  build-artifacts:
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build-release-artifacts.yml
+++ b/.github/workflows/build-release-artifacts.yml
@@ -1,0 +1,25 @@
+name: Build Release Artifacts
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - 'releases/**'
+
+jobs:
+  create-branches:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-release-artifacts.yml
```diff
@@ -0,0 +1,25 @@
...
+  create-branches:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Perform gradle build
+        run: |
+          ./gradlew  firebasePublish -PprojectsToPublish=firebase-firestore -PpublishMode=RELEASE -PincludeFireEscapeArtifacts=true
```
yifanyang, line 19: I think hardcoding is good in this initial version. Eventually we want to make the value of `-PprojectsToPublish` coming from `release.cfg`?

What I'm thinking -- Maybe we can make a [custom action](https://docs.github.com/en/actions/creating-actions/about-custom-actions) around `release.cfg` (for its creation, parsing, etc). That way it is possible to be used in multiple workflows and its input/output format can be defined based on its intended usage.

Although this is just one way to design it.

.github/workflows/build-release-artifacts.yml
```diff
@@ -0,0 +1,25 @@
+name: Build Release Artifacts
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - 'releases/**'
```
yifanyang, line 7: I thought the release branches are named `*.release` in [create_releases.yml](https://github.com/firebase/firebase-android-sdk/blob/master/.github/workflows/create_releases.yml)?

.github/workflows/build-release-artifacts.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  create-branches:
```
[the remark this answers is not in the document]
rlazo: Done.

yifanyang (approved):

PR #4288, opened 2022-11-07
Register executors as components.

vkryachko:
The intent for those is to be used by all Firebase SDKs and forbid creating their own at will.


firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
```diff
@@ -118,12 +113,12 @@ private DefaultHeartBeatController(
...
       Set<HeartBeatConsumer> consumers,
-      Provider<UserAgentPublisher> userAgentProvider) {
+      Provider<UserAgentPublisher> userAgentProvider,
+      Executor backgroundExecutor) {
     this(
         () -> new HeartBeatInfoStorage(context, persistenceKey),
         consumers,
-        new ThreadPoolExecutor(
-            0, 1, 30, TimeUnit.SECONDS, new LinkedBlockingQueue<>(), THREAD_FACTORY),
+        backgroundExecutor,
```
vkryachko, line 121: @VinayGuthal is it safe to switch to a multi-threaded executor, looks like this used to rely on a single-threaded one.

firebase-common/src/main/java/com/google/firebase/concurrent/CustomThreadFactory.java
```diff
@@ -0,0 +1,41 @@
...
+package com.google.firebase.concurrent;
+
+import java.util.Locale;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicLong;
+
+class CustomThreadFactory implements ThreadFactory {
+  private static final ThreadFactory DEFAULT = Executors.defaultThreadFactory();
+  private final AtomicLong threadCount = new AtomicLong();
```
rlazo, line 24: why the counter?

firebase-common/src/test/AndroidManifest.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 Google LLC -->
```
rlazo, line 2:
```suggestion
<!-- Copyright 2022 Google LLC -->
```

tools/lint/src/main/kotlin/CheckRegistry.kt
```diff
@@ -29,6 +29,8 @@ class CheckRegistry : IssueRegistry() {
                 NonAndroidxNullabilityDetector.NON_ANDROIDX_NULLABILITY,
                 DeferredApiDetector.INVALID_DEFERRED_API_USE,
                 ProviderAssignmentDetector.INVALID_PROVIDER_ASSIGNMENT
+                // TODO(vkryachko): enable the check after migration.
```
rlazo, line 32: You mean, after adding the @suppressLint in current usages?

firebase-common/src/main/java/com/google/firebase/concurrent/CustomThreadFactory.java
vkryachko: This is common practice to give different names to threads, i.e. `executor-foo-1`, `executor-foo-2` etc.

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/DefaultHeartBeatController.java
vkryachko: Discussed offline, it should be safe

firebase-common/src/main/java/com/google/firebase/concurrent/CustomThreadFactory.java
rlazo: Didn't see it in the name, thanks!

rlazo (approved):

PR #4293, opened 2022-11-08
Avoid throwing in ForcedSender.sendBlocking

mrober:
Make ForcedSender.sendBlocking do nothing, instead of throw, when given a wrong Transport type.

This will make it easier to mock Transport in tests. This method is best-effort anyway, so doing nothing instead of throwing is fine.

### DIFF
--- a/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ForcedSender.java
+++ b/transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ForcedSender.java
@@ -14,21 +14,28 @@
 
 package com.google.android.datatransport.runtime;
 
+import android.annotation.SuppressLint;
 import androidx.annotation.Discouraged;
 import androidx.annotation.WorkerThread;
 import com.google.android.datatransport.Priority;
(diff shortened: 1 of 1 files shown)


transport/transport-runtime/src/main/java/com/google/android/datatransport/runtime/ForcedSender.java
```diff
@@ -26,16 +25,11 @@
...
-  }
-
-  private static TransportContext getTransportContextOrThrow(Transport<?> transport) {
     if (transport instanceof TransportImpl) {
-      return ((TransportImpl<?>) transport).getTransportContext();
+      TransportContext context =
+          ((TransportImpl<?>) transport).getTransportContext().withPriority(priority);
+      TransportRuntime.getInstance().getUploader().logAndUpdateState(context, 1);
     }
-    throw new IllegalArgumentException("Expected instance of TransportImpl.");
```
vkryachko, line 38: Let's add a warning log statement here, there's logging util available in this sdk under
`com.google.android.datatransport.runtime.logging.Logging`

vkryachko (approved):

PR #4297, opened 2022-11-09
Enable thread pool linter check.

vkryachko:
All violations are now suppressed, bugs filed to fix each product.


tools/lint/src/main/kotlin/ThreadPoolDetector.kt
```diff
@@ -19,7 +19,15 @@ class ThreadPoolDetector : Detector(), SourceCodeScanner {
         "newScheduledThreadPool",
         "newSingleThreadExecutor",
         "newSingleThreadScheduledExecutor",
-        "newWorkStealingPool"
+        "newWorkStealingPool",
+        "<init>"
```
emilypgoogle, line 23: Is having `<init>` as a method target here actually linting `<init>` invocations? It looks like you're handling constructors with a different method.

emilypgoogle (approved):

PR #4301, opened 2022-11-10
Add action to show a message when merging to the main branch

rlazo:


rlazo:
This action will be disabled after merged until it's useful. I haven't found a way of creating a disabled action.

rlazo:
> Would it make sense to have this check require adding a tag to the pr and fail otherwise? this might make it easier to spot the action required to merge the pr

Done

PR #4308, opened 2022-11-10
Persist Realtime RC backoffs

qdpham13:
Change exponential backoff logic to use base RC's backoff logic and persist backoff state to disk.


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -351,6 +397,14 @@ public synchronized void beginRealtimeHttpStream() {
       return;
     }
 
+    ConfigMetadataClient.RealtimeBackoffMetadata backoffMetadata =
+        metadataClient.getRealtimeBackoffMetadata();
+    Date currentTime = new Date(clock.currentTimeMillis());
+    if (currentTime.before(backoffMetadata.getBackoffEndTime())) {
```
danasilver, line 403: Wouldn't we want to wait to retry if the backoff end time is in the future? Maybe I'm misreading this?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -381,6 +436,10 @@ public synchronized void beginRealtimeHttpStream() {
       if (responseCode == null
           || responseCode == HttpURLConnection.HTTP_OK
           || isStatusCodeRetryable(responseCode)) {
+        if (responseCode == null || isStatusCodeRetryable(responseCode)) {
```
danasilver, line 439: Unnesting this conditional would be easier to read since it partially overlaps with the parent condition

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -334,6 +336,8 @@ public void onError(@NonNull FirebaseRemoteConfigException error) {
             listeners,
             mockRetryListener,
             scheduledExecutorService);
+    realtimeMetadataClient =
```
danasilver, line 339: Are there other tests we can add to assert the backoff behavior?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigMetadataClient.java
```diff
@@ -67,16 +67,22 @@ public class ConfigMetadataClient {
   private static final String BACKOFF_END_TIME_IN_MILLIS_KEY = "backoff_end_time_in_millis";
   private static final String NUM_FAILED_FETCHES_KEY = "num_failed_fetches";
   private static final String LAST_TEMPLATE_VERSION = "last_template_version";
+  private static final String NUM_FAILED_REALTIME_STREAMS_KEY = "num_failed_realtime_stream";
```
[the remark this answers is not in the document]
qdpham13: Updated

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigMetadataClient.java
```diff
@@ -67,16 +67,22 @@ public class ConfigMetadataClient {
...
+  private static final String REALTIME_BACKOFF_END_TIME_IN_MILLIS_KEY =
+      "realtime_backoff_end_time_in_millis";
+  static final int NO_FAILED_REALTIME_STREAMS = 0;
```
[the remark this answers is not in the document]
qdpham13: I grouped them together and added a descriptive comment

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -54,6 +59,14 @@
 import org.json.JSONObject;
 
 public class ConfigRealtimeHttpClient {
+  /**
+   * The exponential backoff intervals, up to ~4 hours.
+   *
+   * <p>Every value must be even.
+   */
+  @VisibleForTesting
+  static final int[] BACKOFF_TIME_DURATIONS_IN_MINUTES = {2, 4, 8, 16, 32, 64, 128, 256};
```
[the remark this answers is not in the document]
qdpham13: Yeah you're right, updated to 8

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -205,24 +223,50 @@ private void setRequestParams(HttpURLConnection httpURLConnection) throws IOExce
...
+   * Returns a random backoff duration from the range {@code timeoutDuration} +/- 50% of {@code
+   * timeoutDuration}, where {@code timeoutDuration = }{@link
+   * #BACKOFF_TIME_DURATIONS_IN_MINUTES}{@code [numFailedStreams-1]}.
+   */
+  private long getRandomizedBackoffDurationInMillis(int numFailedStreams) {
+    // The backoff duration length after numFailedFetches.
+    long timeOutDurationInMillis =
+        MINUTES.toMillis(
+            BACKOFF_TIME_DURATIONS_IN_MINUTES[
+                Math.min(numFailedStreams, BACKOFF_TIME_DURATIONS_IN_MINUTES.length) - 1]);
```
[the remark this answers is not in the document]
qdpham13: Added some comments and simplified it

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: backoffMetadata,getBackoffEndTime() is the backoff time for the Http stream so if the currentTime is before that then we should retry the connection later in the future

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: Simplified it to remove the nesting

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
qdpham13: Added some more tests here that go through the path of Realtime

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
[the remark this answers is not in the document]
danasilver: Maybe note they should be the same (or set one based on the other)

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -205,24 +223,60 @@ private void setRequestParams(HttpURLConnection httpURLConnection) throws IOExce
...
+
+  /**
+   * Returns a random backoff duration from the range {@code timeoutDuration} +/- 50% of {@code
+   * timeoutDuration}, where {@code timeoutDuration = }{@link
+   * #BACKOFF_TIME_DURATIONS_IN_MINUTES}{@code [numFailedStreams-1]}.
+   */
+  private long getRandomizedBackoffDurationInMillis(int numFailedStreams) {
+    // Index of backoff duration. Use the minimum of numFailedFetch or the max length of backoff
+    // duration array.
+    int idx = Math.min(numFailedStreams, BACKOFF_TIME_DURATIONS_IN_MINUTES.length);
```
danasilver, line 270: I think this would be easier to read as a conditional. Thoughts?

```
int maxBackoffAttempts = BACKOFF_TIME_DURATIONS_IN_MINUTES.length;
long timeoutDuration = MINUTES.toMillis(BACKOFF_TIME_DURATIONS_IN_MINUTES[maxBackoffAttempts];

// Shorten the duration if there are remaining backoff attempts.
if (numFailedStreams < maxBackoffAttempts) {
  timeoutDuration = MINUTES.toMillis(BACKOFF_TIME_DURATIONS_IN_MINUTES[numFailedStreams])
}

return timeoutDuration / 2 + random.nextInt((int) timoutDuration);
```

qdpham13: Yeah I like the conditional idea, I updated it to have a variation of this

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -269,14 +326,16 @@ public void startHttpConnection() {
   /** Retries HTTP stream connection asyncly in random time intervals. */
   @SuppressLint("VisibleForTests")
   public synchronized void retryHttpConnection() {
```
qdpham13, line 328: Rename method

qdpham13: Done

PR #4310, opened 2022-11-11
Set ndkVersion for Crashlytics NDK SDK.

mrober:
Set ndkVersion to 25.1.8937393 for Crashlytics NDK SDK.

We used to build with 21... but that is no longer required.


kmandrika (approved):

vkryachko:
fyi @yifanyang , we can probably remove our custom ndk download from all GHA workflows.

PR #4328, opened 2022-11-15
Realtime Fetch ID

qdpham13:
Implementation of Realtime Fetch ID: [go/realtime-rc-fetch-id](http://goto.google.com/realtime-rc-fetch-id)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -195,7 +195,8 @@ public void run() {
 
   @VisibleForTesting
   public synchronized void fetchLatestConfig(int remainingAttempts, long targetVersion) {
-    Task<ConfigFetchHandler.FetchResponse> fetchTask = configFetchHandler.fetch(0L);
+    Task<ConfigFetchHandler.FetchResponse> fetchTask =
+        configFetchHandler.realtimeFetch(0L, FETCH_RETRY - remainingAttempts + 1);
```
danasilver, line 199: This could use a comment since the math is a bit cryptic on its own. Maybe even worth just keeping a separate attempts counter?

Slightly related, `FETCH_RETRIES_ALLOWED` or `INITIAL_FETCH_RETRIES` would be more descriptive as a variable name.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -156,13 +163,67 @@ public Task<FetchResponse> fetch() {
...
+   *   <li>The backoff duration from a previous throttled exception has not expired,
+   *   <li>The backend responded with a throttled error, or
+   *   <li>The backend responded with unavailable errors for the last two fetch requests.
+   * </ul>
+   *
+   * @return A {@link Task} representing a Realtime fetch call that returns a {@link FetchResponse}
+   *     with the configs fetched from the backend. If the backend was not called or the backend had
+   *     no updates, the {@link FetchResponse}'s configs will be {@code null}.
+   */
+  public Task<FetchResponse> realtimeFetch(
```
danasilver, line 210: It feels a bit off to me that half the header comes from calling a particular method (`realtimeFetch`) and half from a param passed the method (`fetchAttemptNumber`). It also breaks the encapsulation of fetch within realtime to have realtime implementation details here.

Since the type and attempt are fetch details and realtime is just making use of them, passing both in (i.e. `fetchWithMetadata(long minimumFetchIntervalInSeconds, FetchType fetchType, int fetchAttempt)`) could make sense. A `FetchType` enum could make sure the value passed in makes sense.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandlerTest.java
```diff
@@ -337,6 +337,22 @@ public void fetchWithExpiration_cacheHasNotExpired_doesNotFetchFromBackend() thr
     verifyBackendIsNeverCalled();
   }
 
+  @Test
```
danasilver, line 340: Nice realtime tests! 😄  Mind splitting the ones that aren't for the fetch id into their own PR to make this a bit easier to review? Also, I'm not sure I see a test for the header being set?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
qdpham13: Updated.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
qdpham13: I guess this makes sense. I updated it to create a new method that isn't so closely tied to Realtime.

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandlerTest.java
qdpham13: Removed and will move to a separate PR

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
```diff
@@ -626,4 +691,14 @@ public ConfigContainer getFetchedConfigs() {
       int LOCAL_STORAGE_USED = 2;
     }
   }
+
+  public enum FetchType {
+    BASE,
```
danasilver, line 696: The enum value can be stored in an instance field, which is more clear to read and makes it independent of the enum itself, i.e.

```
BASE("Base"),
...

private final int value;

FetchType(value) {
  this.value = value;
}

String getValue() {
  return value;
}
```

danasilver (approved):

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
qdpham13: Updated.

PR #4330, opened 2022-11-15
Some improvements to firebase-appdistribution/test-app

lfkellogg:
Mostly changes how the SDK is used to better match the setup guide:
- Use prod/beta build flavors to control whether the full SDK is used
- Only show feedback menu item for beta variant
- Always initialize the default notification-based trigger
- Improve error handling in MainActivity


firebase-appdistribution/test-app/src/main/java/com/googletest/firebase/appdistribution/testapp/MainActivity.kt
```diff
@@ -106,6 +99,9 @@ class MainActivity : AppCompatActivity() {
                 }
             }
         }
+
+        firebaseAppDistribution.showFeedbackNotification(
```
kaibolay, line 103: Why always show this? I liked the choice...

kaibolay (approved):

PR #4333, opened 2022-11-16
add generateReleaseConfig job

davidmotson:


buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+        val ktxModuleName = "$moduleName:ktx"
+        listOf(moduleName, ktxModuleName).filter { availableModules.contains(it) }
+    }
+
+    private fun parseSubProjects(rootDir: File) = File(rootDir, "subprojects.cfg").readLines().filter { !it.startsWith("#") }.toSet()
+
+    @Throws(IOException::class)
+    private fun initRepo(repoPath: File): Git {
+        if (!repoPath.exists()) {
+            throw RuntimeException("Repo path doesn't exist")
```
daymxn, line 43: FWIW, `Git.open` would throw the `IOException` if the file couldn't be found. I'm not sure what the error message would be though, so this may be preferable.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+    @Throws(GitAPIException::class)
+    private fun getObjectRefForBranchName(repo: Git, branchName: String): ObjectId {
+        return repo.branchList().setListMode(ListBranchCommand.ListMode.REMOTE).call()
+                .filter { ref: Ref -> ref.name == "refs/remotes/origin/$branchName" }
+                .map { obj: Ref -> obj.objectId }.firstOrNull()
+                ?: throw RuntimeException("Could not find branch named $branchName")
+    }
+
+    private fun getChangedLibraries(repo: Git, startRef: ObjectId, endRef: ObjectId, directories: List<String>): List<String> {
+        return FIREBASE_LIBRARIES.filter { libraryPath: String ->
```
daymxn, line 57: I think you meant for this to be `directories.filter`

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+            repo.log()
+                    .addPath("$libraryPath/*.kt")
```
daymxn, line 59: Doesn't look like we can get away with this, unless I'm mistaken:
```
Note that regex expressions or wildcards are not supported.
```
From the docs for `addPath`

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    @Throws(GitAPIException::class)
+    private fun getObjectRefForBranchName(repo: Git, branchName: String): ObjectId {
+        return repo.branchList().setListMode(ListBranchCommand.ListMode.REMOTE).call()
+                .filter { ref: Ref -> ref.name == "refs/remotes/origin/$branchName" }
+                .map { obj: Ref -> obj.objectId }.firstOrNull()
+                ?: throw RuntimeException("Could not find branch named $branchName")
+    }
+
+    private fun getChangedLibraries(repo: Git, startRef: ObjectId, endRef: ObjectId, directories: List<String>): List<String> {
```
daymxn, line 56: FWIW, we should probably reduce this to only care about the `CHANGELOG.md` file. It should be the job of people or CI tooling during PRs to catch changed kotlin and java files for releasable artifacts. Otherwise, this could throw false positives- as it may target artifacts in the repo that shouldn't be released.

daymxn, line 56: Changing these param names to something like `previousReleaseRef` and `currentReleaseRef` would make it easier to understand whats going on here.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    private fun writeReleaseConfig(configPath: File, libraries: List<String>, releaseName: String) {
+        File(configPath, "release.cfg").writeText(
+                """
+                    [release]
+                    name = $releaseName
+                    mode = RELEASE
+                    
+                    [modules]
+                    ${libraries.joinToString("\n                    ")}
```
daymxn, line 78: Theres a couple ways you could avoid needing that long string for padding that would be more type safe:

```kotlin
${libraries.joinToString { "$it\n".padEnd(20, ' ') }}
```
```kotlin
${libraries.joinToString("\n".padEnd(20, ' '))}
```
```kotlin
${libraries.joinToString("\n", postfix = " ".repeat(20))}
```

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+                """.trimIndent()
+        )
+    }
+
+    companion object {
+        private val FIREBASE_LIBRARIES = listOf(
```
daymxn, line 84: Whats the difference between this and `subprojects.cfg`? Is this like a reduced list of who releases?

rlazo:
I found a bug running the task

1. Checkout the `release_generator_job` branch.
2. Use `ep/fireperf-startup-component` as the past release branch like this

```
./gradlew generateReleaseConfig -PcurrentRelease=m130 -PpastRelease=ep/fireperf-startup-component
```

The generated configuration is

```
[release]
name = m130
mode = RELEASE

[modules]
firebase-crashlytics
firebase-crashlytics:ktx
firebase-crashlytics-ndk
firebase-firestore
firebase-firestore:ktx
firebase-functions
firebase-functions:ktx
firebase-ml-modeldownloader
firebase-ml-modeldownloader:ktx
firebase-perf
firebase-perf:ktx
firebase-storage
firebase-storage:ktx
```

But it you diff the branches, you'll see that new changes to `AppCheck` have been introduced since `ep/fireperf-startup-component` was created, **but it's not part of generated the configuration**.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
rlazo: +1. AFAIK all projects in `subprojects.cfg` do release.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,118 @@
...
+                ?: throw RuntimeException("Could not find branch named $branchName")
+    }
+
+    private fun getChangedLibraries(repo: Git, startRef: ObjectId, endRef: ObjectId, directories: List<String>): List<String> {
+        return FIREBASE_LIBRARIES.filter { libraryPath: String ->
+            repo.log()
+                    .addPath("$libraryPath/*.kt")
+                    .addPath("$libraryPath/*.java")
+                    .addPath("$libraryPath/CHANGELOG.md")
+                    .addPath("$libraryPath/*.gradle")
```
rlazo, line 62: what about changes to AndroidManifest.xml or resources

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
davidmotson: I hadn't noticed that, I was following the more generic git documentation. 

My gut is that despite what the documentation claims, this form of wildcard (similar to the git log command) does work.

because interpreted literally, there is no diff that could include this.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
davidmotson: Re-thinking this, this error is highly *highly* unlikely, given its pulled from project.rootDir, this line was needed when I was working on functionality

I'll remove this check in favor of the built in IOException

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
davidmotson: I had originally done this as a way to distinguish that -ktx libraries release alongside core libraries even if there are changes only in one of them. 

but I think i've come up with a better way to do this.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
davidmotson: i've changed this to just the directory to resolve a bug noticed by rlazo below. I'll try to narrow it in from here.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
davidmotson: see comment above about widening this to just the path, will try to narrow it from here, changelog.md is a possible narrowing target, but its worth discussing.

davidmotson:
I've fixed the above bug, although the stated branch can no longer be used to repro because master got merged into it. Instead i used  -PpastRelease=dependabot/gradle/com.google.protobuf-protobuf-gradle-plugin-0.9.1

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -1,16 +1,16 @@
...
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.api.ListBranchCommand
 import org.eclipse.jgit.api.errors.GitAPIException
 import org.eclipse.jgit.lib.Constants
 import org.eclipse.jgit.lib.ObjectId
-import org.eclipse.jgit.lib.Ref
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.TaskAction
 
+data class FirebaseLibrary(val moduleNames: List<String>, val directories: List<String>)
```
daymxn, line 12: Lemme find out David hip to FP practices 👀 

OT: Good job here, cleans things up quite a bit.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -19,53 +19,50 @@ open class ReleaseGenerator : DefaultTask() {
...
 
-    private fun findChangedModules(changedDirs: List<String>, availableModules: Set<String>): List<String> = changedDirs.flatMap { directory ->
-        val moduleName = directory.replace("/", ":")
-        val ktxModuleName = "$moduleName:ktx"
-        listOf(moduleName, ktxModuleName).filter { availableModules.contains(it) }
-    }
+    private fun extractLibraries(availableModules: Set<String>): List<FirebaseLibrary> {
+        val nonKtxModules = availableModules.filter { !it.endsWith("ktx") }.toSet()
+        return nonKtxModules
+                .filter { possibleModule -> !nonKtxModules.any { it.startsWith("$possibleModule:") } }
```
daymxn, line 35: Not everyday you get to see a good use of `any`. Great job!

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -19,53 +19,50 @@ open class ReleaseGenerator : DefaultTask() {
...
-    private fun initRepo(repoPath: File): Git {
-        if (!repoPath.exists()) {
-            throw RuntimeException("Repo path doesn't exist")
-        }
-        return Git.open(repoPath)
+                    FirebaseLibrary(moduleNames, directories)
+                }
     }
 
+    private fun parseSubProjects(rootDir: File) = File(rootDir, "subprojects.cfg").readLines().filter { !it.startsWith("#") }.filter { it.isNotEmpty() }.toSet()
```
daymxn, line 46: Very small nit:

```kotlin
File(project.rootDir, "subprojects.cfg").readLines().filterNot {
      it.startsWith("#") || it.isEmpty()
}.toSet()
```

Might be preferable, depends on taste. But since both filters are checking for the inverse existence of something- it might make more sense.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,79 @@
...
+data class FirebaseLibrary(val moduleNames: List<String>, val directories: List<String>)
+
+open class ReleaseGenerator : DefaultTask() {
+    @TaskAction
+    @Throws(Exception::class)
+    fun generateReleaseConfig() {
+        val currentRelease = project.property("currentRelease").toString()
+        val pastRelease = project.property("pastRelease").toString()
+        val rootDir = project.rootDir
+        val availableModules = parseSubProjects(rootDir)
```
rlazo, line 21: Getting the candidates by parsing the `subprojects.cfg` file itself can introduce subtle bugs. Not all subprojects listed there are actually "firebase libraries". For example `firebase-config:bandwagoner` which is only used for testing. You can filter the projects themselves based on whether they have the firebase library extension like this

```
        val firebaseLibraries = project.subprojects
                .mapNotNull { it.extensions.findByType(FirebaseLibraryExtension::class.java) }
                .filterNot { it.artifactId.get().contains("ktx") }
                .map { e ->
                    val librariesPathToRelease = e.librariesToRelease.map { it.project.path.drop(1) }
                    FirebaseLibrary(
                            librariesPathToRelease,
                            librariesPathToRelease.map { it.replace(":", "/") })
                }
```

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,79 @@
...
+        val firebaseLibraries = extractLibraries(availableModules)
+
+        val repo = Git.open(rootDir)
+        val headRef = repo.repository.resolve(Constants.HEAD)
+        val branchRef = getObjectRefForBranchName(repo, pastRelease)
+
+        val changedDirs = getChangedLibraries(repo, branchRef, headRef, firebaseLibraries)
+        writeReleaseConfig(rootDir, changedDirs, currentRelease)
```
rlazo, line 29: Could you add a param to print the list of libraries as a comma-separated string too? That output could eventually be passed directly to the `firebasePublish` command as a property, i.e. `-PprojectsToPublish=` removing the need for the config file itself. Just to future proof ourselves.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
davidmotson: ha, thanks! unfortunately i had to get rid of it. feelsbadman.

buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
rlazo: `integ-testing` lives in `subprojects.cfg` and has a `gradle.properties` file but doesn't release.

rlazo:
friendly ping, what's the status of this change?

PR #4348, opened 2022-11-21
Update android executors

VinayGuthal:


firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloader.java
```diff
@@ -55,15 +55,18 @@ public class FirebaseModelDownloader {
   // TODO(b/258424267): Migrate to go/firebase-android-executors
   @SuppressLint("ThreadPoolCreation")
   FirebaseModelDownloader(
```
vkryachko, line 57: nit: remove the todo and suppressLint, here and throughout.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloaderRegistrar.java
```diff
@@ -45,15 +50,24 @@ public class FirebaseModelDownloaderRegistrar implements ComponentRegistrar {
   @NonNull
   @RequiresApi(api = VERSION_CODES.KITKAT)
   public List<Component<?>> getComponents() {
+    Qualified<Executor> bgExecutor = Qualified.qualified(Background.class, Executor.class);
+    Qualified<ExecutorService> bgExecutorService =
```
vkryachko, line 54: is `ExecutorService` required here, can mlkit just use the plain `Executor` instead?

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/internal/CustomModelDownloadService.java
```diff
@@ -97,13 +96,15 @@ public class CustomModelDownloadService {
   // TODO(b/258424267): Migrate to go/firebase-android-executors
   @SuppressLint("ThreadPoolCreation")
   public CustomModelDownloadService(
-      FirebaseApp firebaseApp, FirebaseInstallationsApi installationsApi) {
+      FirebaseApp firebaseApp,
+      FirebaseInstallationsApi installationsApi,
+      ExecutorService executorService) {
```
vkryachko, line 101: Looks like network io is performed on this executor, so bgExecutor is not suitable for this.

Relatedly, down below(lint 164) it looks like a urlconnection is being opened in the caller thread and it should be moved into the continuation of the installationAuthTokenTask. pls move it there.

firebase-ml-modeldownloader/src/main/java/com/google/firebase/ml/modeldownloader/FirebaseModelDownloaderRegistrar.java
```diff
@@ -45,15 +50,24 @@ public class FirebaseModelDownloaderRegistrar implements ComponentRegistrar {
...
             .add(Dependency.required(FirebaseInstallationsApi.class))
+            .add(Dependency.required(bgExecutor))
+            .add(Dependency.required(bgExecutorService))
             .factory(
                 c ->
                     new FirebaseModelDownloader(
-                        c.get(FirebaseApp.class), c.get(FirebaseInstallationsApi.class)))
+                        c.get(FirebaseApp.class),
+                        c.get(FirebaseInstallationsApi.class),
+                        FirebaseExecutors.newSequentialExecutor(c.get(bgExecutor)),
```
vkryachko, line 69: cc @rlazo do you know if a sequential executor is required here, or can a multi-threaded executor be used here instead?

rlazo: From reading the code IMO it should be OK to use a multi-threaded executor here.

vkryachko:
are integ test failures legit?

VinayGuthal:
@vkryachko seems that way.. Will have a look

PR #4350, opened 2022-11-21
Migrate some executors in fad/in-app-feedback

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -65,7 +65,7 @@ class FirebaseAppDistributionImpl implements FirebaseAppDistribution {
   private final SignInStorage signInStorage;
   private final ReleaseIdentifier releaseIdentifier;
   private final ScreenshotTaker screenshotTaker;
-  private final Executor taskExecutor;
+  private final Executor blockingExecutor;
```
lfkellogg, line 68: Requires blocking executor here because some of the uses may wait on a lock via synchronized blocks.

lfkellogg: @vkryachko Some of the uses in this file do not block at all. Would you recommend that we also inject the Lightweight executor, or is it up to us if we're OK using the Blocking executor for lighter weight tasks?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ScreenshotTaker.java
```diff
@@ -40,21 +39,16 @@ class ScreenshotTaker {
...
   private final FirebaseAppDistributionLifecycleNotifier lifecycleNotifier;
-  private final Executor taskExecutor;
-
-  ScreenshotTaker(
-      FirebaseApp firebaseApp, FirebaseAppDistributionLifecycleNotifier lifecycleNotifier) {
-    this(firebaseApp, lifecycleNotifier, Executors.newSingleThreadExecutor());
-  }
+  private final Executor backgroundExecutor;
 
   @VisibleForTesting
```
kaibolay, line 44: remove this annotation - this constructor is now called from `FirebaseAppDistributionRegistrar`

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ScreenshotTaker.java
lfkellogg: Ah good catch

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
vkryachko: >  may wait on a lock via synchronized blocks

Do any of these blocks do network or disk IO or call `task.get()`? or are they all in memory operations?

vkryachko: (optional)

Additionally, speaking of synchronized blocks in the context of tasks and executors, consider refactoring to something like the following [go/android-concurrent/whichexecutor#sequentialexecutor](go/android-concurrent/whichexecutor#sequentialexecutor), where code that needs synchronization could run on the sequential "decorator", and code that does not can run on the executor directly.

The caveat is that we can't use ListenableFutures as described in the doc above and have to use Tasks instead. see [TaskCompletionSource](https://developers.google.com/android/reference/com/google/android/gms/tasks/TaskCompletionSource) on how to do it.

vkryachko: I've also noticed that for some continuations in this file you are not passing in the executor, i.e. `onSuccessTask`, `continueWithTask` etc. This results in the continuations to run on the main thread, is that intentional?

lfkellogg: > Do any of these blocks do network or disk IO synchronously or call task.get()? or are they all in memory operations? 

Actually looking closer `.signInTester()` accesses SharedPreferences, which I understand could in theory touch disk.

Besides that the other potential blocking calls should all be in memory. They involve synchronized blocks (like [this one](https://github.com/firebase/firebase-android-sdk/blob/fad/in-app-feedback/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterSignInManager.java#L123)), which in theory shouldn't block for long if at all, but could block briefly depending on various race conditions.

I assume logging using `android.util.Log` does not block but maybe that's wishful thinking...

> Additionally, speaking of synchronized blocks in the context of tasks and executors, consider refactoring to something like the following [go/android-concurrent/whichexecutor#sequentialexecutor](https://github.com/firebase/firebase-android-sdk/pull/go/android-concurrent/whichexecutor#sequentialexecutor)

That looks great, we'll look into that. It doesn't feel great having synchronized blocks all over the place right now.

> I've also noticed that for some continuations in this file you are not passing in the executor, i.e. onSuccessTask, continueWithTask etc. This results in the continuations to run on the main thread, is that intentional?

I see a couple of them are accidental, actually. I'll fix those. Others though are intentional, if not ideal. We run a lot of Task handlers on the main thread if they should execute quickly. But that's mostly for convenience, rather than create an executor to run something that should return immediately. Maybe in this new world of explicit shared executors we should use `@Lightweight` for these.

lfkellogg: Added the couple that were missing from the `startFeedback*` code paths. I'm going to revisit the other missing ones in a separate PR since they don't concern our in-app feedback WIP.

vkryachko: > > Do any of these blocks do network or disk IO synchronously or call task.get()? or are they all in memory operations?
> 
> Actually looking closer `.signInTester()` accesses SharedPreferences, which I understand could in theory touch disk.

Disk it touched only when the SharedPreferences object is constructed, and for any synchronous writes like commit, so those operations should use the `@Background` executor as using an unbounded pool(i.e. `@Blocking`) for disk io puts a lot of pressure on the file system.

> 
> Besides that the other potential blocking calls should all be in memory. They involve synchronized blocks (like [this one](https://github.com/firebase/firebase-android-sdk/blob/fad/in-app-feedback/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterSignInManager.java#L123)), which in theory shouldn't block for long if at all, but could block briefly depending on various race conditions.

In this case it's best to use a blocking executor since it's unbounded in the number of threads which avoids dead locks, but it's better to refactor to a sequential as mentioned below.
> 
> I assume logging using `android.util.Log` does not block but maybe that's wishful thinking...

It's safe to log and it does not block.

> 
> > Additionally, speaking of synchronized blocks in the context of tasks and executors, consider refactoring to something like the following [go/android-concurrent/whichexecutor#sequentialexecutor](https://github.com/firebase/firebase-android-sdk/pull/go/android-concurrent/whichexecutor#sequentialexecutor)
> 
> That looks great, we'll look into that. It doesn't feel great having synchronized blocks all over the place right now.
> 
> > I've also noticed that for some continuations in this file you are not passing in the executor, i.e. onSuccessTask, continueWithTask etc. This results in the continuations to run on the main thread, is that intentional?
> 
> I see a couple of them are accidental, actually. I'll fix those. Others though are intentional, if not ideal. We run a lot of Task handlers on the main thread if they should execute quickly. But that's mostly for convenience, rather than create an executor to run something that should return immediately. Maybe in this new world of explicit shared executors we should use `@Lightweight` for these.

In my opinion this is a bad default that Tasks provide(even though at the time it was a reasonable thing to do), the general android guidance is to run as little code on the UI thread as possible, i.e. only things that `must` run there, like UI related code. Given this default, I agree that using executors explicitly in all continuations is a better approach and we should even add a linter warning for calls without an executor(maybe?).

lfkellogg: OK thanks Vlad, in summary we will:

- Move construction of SharedPreferences or synchronous writes to a `@Background` executor
- Avoid using synchronized blocks in favor of sequential executors or other approaches (already removed some in https://github.com/firebase/firebase-android-sdk/pull/4353)
- Use explicit executors for Task handlers throughout instead of defaulting to UI thread

vkryachko: Thanks Lee,

> Move construction of SharedPreferences or synchronous writes to a @Background executor

One thing I'll add: it would be awesome if you could implement this as a self-contained class that just does the right thing, that way we would be able to put it in say firebase-common, register it as a component, and have all teams benefit from it. wdyt?

vkryachko: As for the UI thread, looks like all sdks use it **a lot** :( #4363 

So it could be a follow up effort once the executor migration is done

lfkellogg: Sounds good. I'm going to make each of these changes in followup PRs.

PR #4352, opened 2022-11-22
Add action to create bom

emilypgoogle:
Creates the bom and uploads it as an artifact. Original job performed testing, but @rlazo confirmed that testing should be executed somewhere else, so this job does not.

### DIFF
--- a/.github/workflows/make-bom.yml
+++ b/.github/workflows/make-bom.yml
@@ -1,0 +1,45 @@
+name: Make BoM
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


yifanyang (approved):

.github/workflows/make-bom.yml
```diff
@@ -0,0 +1,45 @@
+name: Make BoM
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
```
rlazo, line 8: ubuntu-latest maybe?

rlazo (approved):

PR #4353, opened 2022-11-22
Remove some synchronized blocks from FirebaseAppDistributionImpl

lfkellogg:
Uses a new `TaskCache` instead of relying on locks.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskCache.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.android.gms.tasks.Task;
+
+/**
+ * A cache for Tasks, for use in cases where we only ever want one active task at a time for a
+ * particular operation.
+ */
+class TaskCache<T extends Task> {
+
+  /** A functional interface for a producer of a new Task. */
+  interface TaskProducer<T extends Task> {
```
kaibolay, line 26: Could you just use the generic `java.util.function.Supplier` here?

kaibolay, line 26: If you keep this, add `@FunctionalInterface`

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskCache.java
lfkellogg: Done

PR #4363, opened 2022-11-24
Add lint check that detects UI thread continuations.

vkryachko:


tools/lint/src/main/kotlin/TasksMainThreadDetector.kt
```diff
@@ -0,0 +1,73 @@
...
+    /** Calling methods on the wrong thread */
+    @JvmField
+    val TASK_MAIN_THREAD =
+      Issue.create(
+        id = "TaskMainThread",
+        briefDescription = "Use an explicit Executor for Task continuations",
+        explanation =
+          """
+                    Not providing an executor results in continuations to execute on the main thread \
+                    which is not a good default. Please pass in an executor explicitly.
```
rlazo, line 65:
```suggestion
                    Not providing an executor results in continuations being executed on the main thread, \
                    which in most cases is not intended. Please pass in an executor explicitly.
```

tools/lint/src/main/kotlin/TasksMainThreadDetector.kt
```diff
@@ -0,0 +1,73 @@
+package com.google.firebase.lint.checks
```
rlazo, line 1: qq, do we need the copyright header? I guess yes

tools/lint/src/main/kotlin/TasksMainThreadDetector.kt
rlazo: Deep down we know is not a good default :-D, but let's try spin it differently.

rlazo (approved):

vkryachko:
Oops, I prematurely merged, will send a follow up PR with suggested changed

PR #4366, opened 2022-11-24
Use dagger in functions.

vkryachko:
This simplifies injection of firebase components so they don't have to be passed all the way from the registrar into classes that actually use components.

This change adds `+4.22 kB` to the size of the sdk, but is a small enough increase compared to the advantages it provides

### DIFF
--- a/firebase-functions/src/main/java/com/google/firebase/functions/FunctionsComponent.java
+++ b/firebase-functions/src/main/java/com/google/firebase/functions/FunctionsComponent.java
@@ -1,0 +1,78 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


firebase-functions/src/main/java/com/google/firebase/functions/FunctionsComponent.java
```diff
@@ -0,0 +1,78 @@
+// Copyright 2018 Google LLC
```
rlazo, line 1:
```suggestion
// Copyright 2022 Google LLC
```

rlazo (approved):

firebase-functions/src/main/java/com/google/firebase/functions/FunctionsComponent.java
vkryachko: oops, looks like auto merge does not take these comments into account

PR #4376, opened 2022-11-29
Update crashpad submodule and fixed the build

mrober:
Tested on crow emulator x86 and x86 64, with multiple android api levels.
Tested on physical arm devices, 32 bit and 64 bit.


firebase-crashlytics-ndk/src/main/jni/crashpad/crashpad_util/include/execvpe.h
```diff
@@ -0,0 +1,9 @@
+#ifndef CRASHPAD_UTIL_EXECVPE_H
+#define CRASHPAD_UTIL_EXECVPE_H
+
+#if __ANDROID_API__ < 21
+extern int
+execvpe(const char *__file, char *const *__argv, char *const *__envp);
```
kmandrika, line 6: this should have an extern "C" around it

mrober: Done

kmandrika (approved):

PR #4379, opened 2022-11-29
Add workflow triggers for startup time test

yifanyang:


rlazo (approved):

vkryachko:
Can we make sure we only measure changed modules as opposed to all modules?

PR #4389, opened 2022-12-01
Fix fcm tests.

vkryachko:


vkryachko:
@gsakakihara I am not actually sure it's fixing anything, I got an error message saying:

```
Main looper has queued unexecuted runnables. This might be the cause of the test failure. You might need a shadowOf(Looper.getMainLooper()).idle() call.
```

Wdyt?

PR #4392, opened 2022-12-02
Enable for Kotlin Gradle dsl.

vkryachko:
Additionally migrate to using a version catalog to simplify dependency management.


firebase-common/ktx/ktx.gradle.kts
```diff
@@ -0,0 +1,60 @@
+// Copyright 2019 Google LLC
```
rlazo, line 1: 2022

gradle/libs.versions.toml
```diff
@@ -0,0 +1,20 @@
+[versions]
```
rlazo, line 1: nit, mixed single a double quotes

rlazo: also, maybe sort alphabetically :-)

PR #4393, opened 2022-12-02
Migrate RC to common executors.

vkryachko:


firebase-config/src/main/java/com/google/firebase/remoteconfig/RemoteConfigComponent.java
```diff
@@ -88,17 +86,16 @@ public class RemoteConfigComponent {
   private Map<String, String> customHeaders = new HashMap<>();
 
   /** Firebase Remote Config Component constructor. */
-  // TODO(b/258275481): Migrate to go/firebase-android-executors
-  @SuppressLint("ThreadPoolCreation")
   RemoteConfigComponent(
       Context context,
+      @Background Executor executor,
```
danasilver, line 91: Should this be `@Blocking`, too?

danasilver (approved):

PR #4394, opened 2022-12-02
Allow task subclasses to implement overloads.

vkryachko:


tools/lint/src/test/kotlin/TasksMainThreadDetectorTests.kt
```diff
@@ -0,0 +1,139 @@
...
+    .trimIndent()
+}
+
+class TasksMainThreadDetectorTests : LintDetectorTest() {
+  override fun getDetector(): Detector = TasksMainThreadDetector()
+
+  override fun getIssues(): MutableList<Issue> =
+    mutableListOf(TasksMainThreadDetector.TASK_MAIN_THREAD)
+
+  fun test_continueWith_withoutExecutor_shouldFail() {
```
rlazo, line 41: does it work with spaces? 

fun `long name with spaces`

rlazo (approved):

tools/lint/src/test/kotlin/TasksMainThreadDetectorTests.kt
vkryachko: It does not, might be a limitation of `LintDetectorTest`

PR #4395, opened 2022-12-02
Fix RC Executor annotation

vkryachko:


vkryachko:
Sending follow up PR, since the previous got merged via auto merge

danasilver (approved):
Thanks!

PR #4406, opened 2022-12-05
Submit startup time request asynchronously via task queue

yifanyang:


vkryachko (commented):
task queue vs synchronous post seems like an implementation detail, any reason it has to be known on the client? Can it be done in the service instead? i.e. could the service's POST handler add the task to the queue, additionally it might remove the need for Task Queue ACLs, wdyt?

yifanyang:
> task queue vs synchronous post seems like an implementation detail, any reason it has to be known on the client? Can it be done in the service instead? i.e. could the service's POST handler add the task to the queue, additionally it might remove the need for Task Queue ACLs, wdyt?

I agree. Can you also take a look at [tg/1668088](http://tg/1668088), which let task queue to post back to itself? That way only one query param `async=true` is required on the client side.

Concerns that I have about the second approach:
- when the service is running locally, task queue cannot post to it
- the service needs to know its public url before it is deployed

yifanyang:
@vkryachko @rlazo 

This is ready for review. Changed to this other approach [tg/1668088](http://tg/1668088). The latest test run [log](https://github.com/firebase/firebase-android-sdk/actions/runs/3625032747/jobs/6112622053#step:9:5240) in this pull request shows that the request is returned immediately and the service is now processing the request asynchronously.

vkryachko (approved):

PR #4407, opened 2022-12-05
Refine startup time test app

yifanyang:
- Forece initialize all components
- Include fire-perf-early in measurements
- Test only changed products in pull requests


yifanyang:
@vkryachko @rlazo 

This is ready for review. This pull request:
- forces initializing all components in the main activity
- includes `fire-perf-early` in the measurements
- tests only changed products in pull requests

Per discussion offline, the approach to test only changed products is to:
- still build the test apk with all products included
- filter out traces collected for products that are not affected in the pull request

firebase-common/src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -472,7 +472,7 @@ public boolean isDefaultApp() {
   /** @hide */
   @VisibleForTesting
   @RestrictTo(Scope.TESTS)
-  void initializeAllComponents() {
+  public void initializeAllComponents() {
```
vkryachko, line 475: Instead of making it public, I suggest creating a function in the benchmark app under `com.google.firebase` like:

```kotlin

internal fun initializeAllComponentsForBenchmark(app : Firebase) {
  app.initializeAllComponents()
}
```

This will overcome the visibility restriction and also prevent `app. initializeAllComponents()` from showing up in autocomplete for 3p developers.

yifanyang: Done

PR #4426, opened 2022-12-06
Add Diff Javadoc workflow

emilypgoogle:
Adds a workflow to diff javadoc in PRs. The workflow will add one collapsed comment and edit it on new commits, showing a git styled diff in github markdown for each file changed.

### DIFF
--- a/.github/workflows/diff-javadoc.yml
+++ b/.github/workflows/diff-javadoc.yml
@@ -23,8 +23,14 @@ jobs:
           distribution: temurin
           cache: gradle
 
+      - name: Changed Modules
+        id: changed-modules
+        run: |
+          git diff --name-only HEAD~1 | xargs printf -- '--changed-git-paths %s\n' | xargs ./gradlew writeChangedProjects --output-file-path=modules.json
+          echo "run=$(cat modules.json | sed "s/[]\"[]//g" | sed "s/,/\n/g" | xargs printf -- "%s:kotlinDoc "))" >> $GITHUB_OUTPUT
+
       - name: Build
-        run: ./gradlew kotlindoc
+        run: ./gradlew ${{ steps.changed-modules.outputs.run }}
 
       - name: Move original docs
         run: mv build ~/diff/original
@@ -32,7 +38,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Build
-        run: ./gradlew kotlindoc
+        run: ./gradlew ${{ steps.changed-modules.outputs.run }}
 
       - name: Move modified docs
         run: mv build ~/diff/modified
--- a/.github/workflows/diff-javadoc.yml
+++ b/.github/workflows/diff-javadoc.yml
@@ -13,8 +13,6 @@ jobs:
         run: mkdir ~/diff
 
       - uses: actions/checkout@v3
-        with:
-          ref: ${{ github.base_ref }}
 
       - name: Set up JDK 11
         uses: actions/setup-java@v2
@@ -33,15 +31,17 @@ jobs:
         run: ./gradlew ${{ steps.changed-modules.outputs.run }}
 
       - name: Move original docs
-        run: mv build ~/diff/original
+        run: mv build ~/diff/modified
 
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
 
       - name: Build
         run: ./gradlew ${{ steps.changed-modules.outputs.run }}
 
       - name: Move modified docs
-        run: mv build ~/diff/modified
+        run: mv build ~/diff/original
 
       - name: Diff docs
         run: >
--- a/.github/workflows/diff-javadoc.yml
+++ b/.github/workflows/diff-javadoc.yml
@@ -12,16 +12,16 @@ jobs:
       - name: Make Dir
         run: mkdir ~/diff
 
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
+
       - name: Set up JDK 11
         uses: actions/setup-java@v2
         with:
           java-version: 11
           distribution: temurin
           cache: gradle
-
-      - uses: actions/checkout@v3
-        with:
-          ref: ${{ github.base_ref }}
 
       - name: Build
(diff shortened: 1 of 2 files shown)


.github/workflows/diff-javadoc.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: temurin
+          cache: gradle
+
+      - name: Build
+        run: ./gradlew kotlindoc
```
vkryachko, line 27: I think it may make sense to generate docs only for products that changed to avoid flakes and speed up this job. there is an example on how it's done in another workflow https://github.com/firebase/firebase-android-sdk/blob/00ea7ee68c9660e3d56c502b3f3df83026fda8df/.github/workflows/health-metrics.yml#L126-L129

And https://github.com/firebase/firebase-android-sdk/blob/00ea7ee68c9660e3d56c502b3f3df83026fda8df/.github/workflows/ci_tests.yml#L33-L34

wdyt?

vkryachko: Health metrics does not use a matrix and instead uses a modules.json directly in the same action. So I think something along those lines can be done here as well, i.e.

1. checkout PR
2. run writeChangedProjects
3. parse the modules into something like `CHANGED=":firebase-common,firebase-components"` and then run `./gradlew {CHANGED}:kotlindoc` on both branches

.github/workflows/diff-javadoc.yml
```diff
@@ -0,0 +1,58 @@
+name: Diff Javadoc
+
+on:
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
```
rlazo, line 8: latest

.github/workflows/diff-javadoc.yml
```diff
@@ -0,0 +1,58 @@
...
+      pull-requests: write
+    steps:
+      - name: Make Dir
+        run: mkdir ~/diff
+
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
+
+      - name: Set up JDK 11
```
rlazo, line 19: i think setup would be better done first.

.github/workflows/diff-javadoc.yml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Make Dir
+        run: mkdir ~/diff
+
+      - uses: actions/checkout@v3
```
rlazo, line 15: I *think* you can use

```
  with:
    path: main
```

to save you creating a diff directory

.github/workflows/diff-javadoc.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Build
+        run: ./gradlew kotlindoc
+
+      - name: Move modified docs
+        run: mv build ~/diff/modified
+
+      - name: Diff docs
+        run: >
+          `# Recursively diff directories, including new files, git style, with 3 lines of context`
```
rlazo, line 42: Kudos for making it clear what happens here!

.github/workflows/diff-javadoc.yml
emilypgoogle: I've run into an issue where a lot of our SDK projects don't support the `kotlinDoc` task and are bricking the entire build when I try to run them, and I don't have a clean way to filter them out to only build the tasks that do, thinking it might make sense to revisit this optimization in the future as it's taking more time than I expected, thoughts?

.github/workflows/diff-javadoc.yml
emilypgoogle: Setup needs gradle files to be present, so they need to be checked out

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/ChangedModulesTask.kt
```diff
@@ -43,14 +48,21 @@ abstract class ChangedModulesTask : DefaultTask() {
     val projects =
       AffectedProjectFinder(project, changedGitPaths.toSet(), listOf())
         .find()
+        .filter {
+          val ext = it.extensions.findByType(FirebaseLibraryExtension::class.java)
```
vkryachko, line 52: nit
```suggestion
          val ext = it.extensions.findByType<FirebaseLibraryExtension>()
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/ChangedModulesTask.kt
```diff
@@ -32,6 +33,10 @@ abstract class ChangedModulesTask : DefaultTask() {
   @set:Option(option = "output-file-path", description = "Hello")
   abstract var outputFilePath: String
 
+  @get:Input
+  @set:Option(option = "only-firebase-sdks", description = "Hello")
```
vkryachko, line 37: I know the descriptions are not good above either, do you mind updating them?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ci/ChangedModulesTask.kt
```diff
@@ -43,14 +48,21 @@ abstract class ChangedModulesTask : DefaultTask() {
     val projects =
       AffectedProjectFinder(project, changedGitPaths.toSet(), listOf())
         .find()
+        .filter {
+          val ext = it.extensions.findByType(FirebaseLibraryExtension::class.java)
+          !onlyFirebaseSDKs || (ext != null && ext.publishJavadoc)
```
vkryachko, line 53: I would remove the `publishJavadoc` check to make this task general purpose and not only useful for this workflow. iiuc it should not break anything since all firebase libraries have the `kotlindoc` task registered even if publishJavadoc is false

vkryachko (approved):

PR #4430, opened 2022-12-07
Use sequential executor in place of synchronized blocks for cached new release

lfkellogg:
Continuing to remove `synchronized` blocks based on conversation on https://github.com/firebase/firebase-android-sdk/pull/4350.

This also uses an executor for more Task continuations.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -211,42 +214,39 @@ public Task<Void> signInTester() {
...
-        () -> {
-          if (!isTesterSignedIn()) {
-            return Tasks.forException(
-                new FirebaseAppDistributionException(
-                    "Tester is not signed in", AUTHENTICATION_FAILURE));
-          }
+    if (!isTesterSignedIn()) {
+      return Tasks.forException(
+          new FirebaseAppDistributionException("Tester is not signed in", AUTHENTICATION_FAILURE));
+    }
```
lfkellogg, line 227: I also pulled this out of the `getOrCreateTask` because it does not actually depend on whether there is a cached task.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SequentialReference.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.firebase.appdistribution.UpdateTask;
+import com.google.firebase.concurrent.FirebaseExecutors;
+import java.util.concurrent.Executor;
+
+/**
+ * A reference to an object that uses a sequential executor to ensure non-concurrent reads and
+ * writes while preventing lock contention.
+ */
+class SequentialReference<T> {
```
kaibolay, line 28: I have to admit I don't really grok this. I really hope @vkryachko does.

Or maybe there is a different paradigm that can be used? This looks like low-level infrastructure which seems to be useful in general - so I'd assume there is some established pattern to do things like this.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionRegistrar.java
```diff
@@ -47,13 +48,19 @@ public class FirebaseAppDistributionRegistrar implements ComponentRegistrar {
...
             .name(LIBRARY_NAME)
             .add(Dependency.required(FirebaseApp.class))
             .add(Dependency.requiredProvider(FirebaseInstallationsApi.class))
             .add(Dependency.required(blockingExecutor))
-            .factory(c -> buildFirebaseAppDistribution(c, c.get(blockingExecutor)))
+            .add(Dependency.required(lightweightExecutor))
+            .factory(
+                c ->
+                    buildFirebaseAppDistribution(
+                        c, c.get(blockingExecutor), c.get(lightweightExecutor)))
```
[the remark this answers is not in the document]
lfkellogg: Can't move the creation of the `Qualified<Executor>` because they are needed to declare the dependencies on lines 58-59.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SequentialReference.java
lfkellogg: I've updated this to be simpler based on our conversation today.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskImpl.java
```diff
@@ -63,6 +63,13 @@ void updateProgress(@NonNull UpdateProgress updateProgress) {
     }
   }
 
+  void shadow(Executor executor, UpdateTask updateTask) {
```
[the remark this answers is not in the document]
lfkellogg: Done.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
```diff
@@ -442,14 +458,17 @@ public void signInTester_whenDialogDismissed_taskFails() {
...
+  public void updateIfNewReleaseAvailable_whenSignInDialogCanceled_taskFails()
+      throws InterruptedException {
     when(mockSignInStorage.getSignInStatus()).thenReturn(false);
     Task signInTask = firebaseAppDistribution.updateIfNewReleaseAvailable();
 
     AlertDialog dialog = assertAlertDialogShown();
     dialog.onBackPressed(); // cancel dialog
-    shadowOf(getMainLooper()).idle();
+    TestUtils.awaitAsyncOperations(testExecutor);
+    TestUtils.awaitAsyncOperations(testExecutor);
```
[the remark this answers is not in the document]
lfkellogg: I've cleaned this up a bit to only await what we need.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
```diff
@@ -25,10 +26,8 @@
 import static com.google.firebase.appdistribution.impl.ErrorMessages.NETWORK_ERROR;
 import static com.google.firebase.appdistribution.impl.ErrorMessages.RELEASE_NOT_FOUND_ERROR;
 import static com.google.firebase.appdistribution.impl.ErrorMessages.UPDATE_CANCELED;
-import static com.google.firebase.appdistribution.impl.TestUtils.assertTaskFailure;
```
kaibolay, line 29: nit: I think static imports for all the `TaskUtil` methods would make sense

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SequentialReference.java
```diff
@@ -0,0 +1,82 @@
...
+        });
+    return taskCompletionSource.getTask();
+  }
+
+  /**
+   * Gets a {@link Task} that will complete with the value.
+   *
+   * <p>Unless a direct executer is passed to {@link Task#addOnSuccessListener(Executor,
+   * OnSuccessListener)} or similar methods, be aware that the value may have changed by the time it
+   * is used.
```
vkryachko, line 70: I wonder if it's actually a problem, i.e. when directExecutor is used here's the sequence of events:

0. `state` = `0`
1. `get()` called(returns `0`)
2. a concurrent `set(1)` is queued but waits for `1` to complete
3. `get()` continuation executes and returns the transformed task using `0`
4. task from 3 is returned to the caller
5. 2 updates the value to `1`
6. the caller observes results based on `0` when it is actually now equal `1`

Now let's what happens when a "real" executor is used:

0. `state` = `0`
1. `get()` called(returns `0`)
2. a concurrent `set(1)` and updates the state to `1`
3. `get()` continuation executes and returns the transformed task using `0`
4. task from 3 is returned to the caller
6. the caller observes results based on `0` when it is actually now equal `1`

So regardless of the sequence of events, the outcome is the same, the value may have changed by the time the caller observes the result, unless direct executor is used throughout the whole continuation change(which generally is not a good idea).

Point being: consider using lightweight throughout unless there is some internal invariant that I am missing and will get violated by this.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
```diff
@@ -106,13 +110,16 @@ public class FirebaseAppDistributionServiceImplTest {
...
       AppDistributionReleaseInternal.builder()
           .setBuildVersion("3")
           .setDisplayVersion("3.0")
           .setReleaseNotes("Newer version.")
           .setBinaryType(BinaryType.APK)
-          .setDownloadUrl(TEST_URL);
+          .setDownloadUrl(TEST_URL)
+          .build();
+
+  private final ExecutorService testExecutor = Executors.newSingleThreadExecutor();
```
vkryachko, line 122: It's actually preferable to use the actual executor from components as it will detect violations, i.e. disk reads, network io, etc and crash the tests. Otherwise there is no guarantee that the code under test uses the right executor which will slow down all of firebase and cause logcat warnings in developers' apps

vkryachko: you can use this utility class to get a hold of the right executor:
https://github.com/firebase/firebase-android-sdk/blob/master/integ-testing/src/main/java/com/google/firebase/concurrent/TestOnlyExecutors.java

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SequentialReference.java
lfkellogg: Yes, good call.

I think there's another reason not to necessarily recommend the direct executor. I was able to reproduce some slightly different behavior: if I introduce an artificial delay in the execution of the initial `get()`, then only a direct executor ensures the entire chain off of the `get()` will execute with the correct value before the first `set(1)` happens. This is because any callbacks get called on the sequential executor's thread immediately after the read or write happens. I was able to repro this.

But this was an artificial test, and it gets complicated in the case where the `get()` succeeds immediately. When the task is already complete before the handlers are added, the callbacks end up being called immediately on the **calling** thread, not the sequential executor's thread. And in that case there's no practical difference between using a direct executor or a "real" executor.

So a direct executor really doesn't guarantee an order of execution. More reason not to recommend it IMO. I think we just need to understand when data might be old and whether that's a problem.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -82,14 +78,17 @@ class FirebaseAppDistributionImpl implements FirebaseAppDistribution {
       @NonNull ApkUpdater apkUpdater,
       @NonNull AabUpdater aabUpdater,
       @NonNull SignInStorage signInStorage,
-      @NonNull FirebaseAppDistributionLifecycleNotifier lifecycleNotifier) {
+      @NonNull FirebaseAppDistributionLifecycleNotifier lifecycleNotifier,
+      @NonNull Executor lightweightExecutor) {
```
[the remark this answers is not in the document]
lfkellogg: Done

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -211,42 +215,41 @@ public Task<Void> signInTester() {
...
-                    }
-                  });
-        });
+    return checkForNewReleaseTaskCache.getOrCreateTask(
+        () ->
+            newReleaseFetcher
+                .checkForNewRelease()
+                .onSuccessTask(lightweightExecutor, release -> cachedNewRelease.set(release))
+                .onSuccessTask(
+                    FirebaseExecutors.directExecutor(),
```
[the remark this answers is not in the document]
lfkellogg: As discussed, direct executor doesn't actually improve correctness in all cases. And correctness isn't even strictly necessary here. I'll remove it.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SequentialReference.java
```diff
@@ -0,0 +1,82 @@
...
+ * A reference to an object that uses a sequential executor to ensure non-concurrent reads and
+ * writes while preventing lock contention.
+ */
+class SequentialReference<T> {
+
+  private final Executor sequentialExecutor;
+  private T value;
+
+  /** Get a {@link SequentialReference} that controls access uses the given sequential executor. */
+  static SequentialReference withSequentialExecutor(Executor sequentialExecutor) {
```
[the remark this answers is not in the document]
lfkellogg: Ah, I actually wasn't considering synchronization in terms of needing to perform reads of multiple values at once, but makes sense that you would need to do that. Agreed, in that case we should just have all of the state be combined into a single reference. I'll remove this.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SequentialReference.java
```diff
@@ -0,0 +1,82 @@
...
+  Task<T> get() {
+    TaskCompletionSource<T> taskCompletionSource = new TaskCompletionSource<>();
+    sequentialExecutor.execute(() -> taskCompletionSource.setResult(value));
+    return taskCompletionSource.getTask();
+  }
+
+  @VisibleForTesting
+  T getSnapshot() {
+    return value;
+  }
```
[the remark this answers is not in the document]
lfkellogg: Yeah I think that's a good move (technically we'll have to use `Tasks.await()`).

I was assuming the tests would do their own awaiting on the executor but might as well enforce it to be safe. Updated.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
lfkellogg: Oh, that's awesome. Done.

firebase-appdistribution/firebase-appdistribution.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     implementation project(':firebase-installations-interop')
     implementation project(':firebase-common')
     testImplementation project(path: ':firebase-appdistribution')
+    testImplementation project(':integ-testing')
```
kaibolay, line 52: what is this? is it modern?

lfkellogg: This gives us access to `TestOnlyExecutors` so we can use the real Firebase executors in our tests.

kaibolay (approved):

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
```diff
@@ -236,7 +237,7 @@ public void checkForNewRelease_authenticationFailure_signOutTester() throws Inte
     Task<AppDistributionRelease> task = firebaseAppDistribution.checkForNewRelease();
 
     awaitTaskFailure(task, AUTHENTICATION_FAILURE, "Test");
-    awaitTermination(testExecutor);
+    awaitTermination(lightweightExecutor);
```
vkryachko, line 240: hm, these executors cannot be shutdown so I am not sure `awaitTermination` would work

vkryachko: but looks like tests have passed, so use your judgement

lfkellogg: Hmm, I'm realizing that all of our `awaitTermination()` calls are really just sleeping for 100 ms because we're never shutting them down first. We need a better way to wait for executor-run tasks throughout our tests. Since these are passing right now though I'm inclined to tackle that in a followup.

vkryachko (approved):

PR #4434, opened 2022-12-08
Port app.get() check from errorprone to lint

vkryachko:
Motivation: Make sure the check works for kotlin, not only java.


tools/lint/src/main/kotlin/FirebaseAppGetDetector.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.UMethod
+import org.jetbrains.uast.getParentOfType
+
+@Suppress("DetectorIsMissingAnnotations")
+class FirebaseAppGetDetector : Detector(), SourceCodeScanner {
+  override fun getApplicableMethodNames(): List<String> = listOf("get")
+
+  override fun visitMethodCall(context: JavaContext, call: UCallExpression, method: PsiMethod) {
+    if (!isFirebaseAppGet(method)) {
```
rlazo, line 37: probably as readable as a single conditional (with the one below)

tools/lint/src/main/kotlin/FirebaseAppGetDetector.kt
```diff
@@ -0,0 +1,95 @@
...
+      return
+    }
+
+    if (withinGetInstance(call)) {
+      return
+    }
+    call.getParentOfType<UMethod>() ?: return
```
rlazo, line 44: if we only want to check if it exists, IMO it would be easier to read inside an if statement

rlazo:
Besides a couple of nits, LGTM

rlazo (approved):

PR #4442, opened 2022-12-12
Fix overlay bug leas to patch mutation optimization not being applied

wu-hui:


firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
```diff
@@ -35,10 +40,17 @@
...
+    Patch,
+    Set,
+    Delete
+  }
+
   private final QueryEngine queryEngine;
 
   private final int[] overlaysReadByCollection = new int[] {0};
   private final int[] overlaysReadByKey = new int[] {0};
+  private final Map<DocumentKey, OverlayType> overlayTypes = new HashMap();
```
ehsannas, line 53: would it make more sense to store the overlay itself (`Map<DocumentKey, Mutation>`) and have the test check the overlay? If we do this, we don't need to define `OverlayType`, and we check not only the type but also the overlay content.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
```diff
@@ -160,12 +181,19 @@ private DocumentOverlayCache wrapOverlayCache(DocumentOverlayCache subject) {
       @Override
       public Overlay getOverlay(DocumentKey key) {
         ++overlaysReadByKey[0];
-        return subject.getOverlay(key);
+        Overlay overlay = subject.getOverlay(key);
+        overlayTypes.put(key, getOverlayType(overlay));
```
ehsannas, line 185: for example we'd do: `overlays.put(key, overlay)`

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
```diff
@@ -1689,4 +1697,21 @@ public void testMultipleFieldPatchesOnLocalDocs() {
...
+    applyRemoteEvent(updateRemoteEvent(doc("foo/baz", 10, map("a", 1)), asList(2), emptyList()));
+    applyRemoteEvent(updateRemoteEvent(doc("foo/bar", 20, map()), asList(2), emptyList()));
+    writeMutation(patchMutation("foo/baz", map("b", 2)));
+
+    resetPersistenceStats();
+
+    localStore.executeQuery(query, /* usePreviousResults= */ true);
+    assertRemoteDocumentsRead(/* byKey= */ 0, /* byCollection= */ 2);
+    assertOverlaysRead(/* byKey= */ 0, /* byCollection= */ 1);
+    assertOverlayTypes(keyMap("foo/baz", CountingQueryEngine.OverlayType.Patch));
```
ehsannas, line 1715: for example we'd check:
```
assertOverlays(keyMap("foo/baz", patchMutation("foo/baz", map("b",2))));
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/OverlayedDocument.java
```diff
@@ -22,21 +22,27 @@
...
+   * The fields that are locally mutated by patch mutations.
+   *
+   * <p>If hasOverlay() is true and the overlayed document is from set or delete mutations, this
+   * returns null. If hasOverlay() is false, this returns FieldMask.EMPTY.
    */
+  // TODO(b/262245989): This screams for a proper sum type (Tagged Union) which does not exist in
+  // Java (yet).
   public @Nullable FieldMask getMutatedFields() {
-    return mutatedFields;
+    return hasOverlay ? mutatedFields : FieldMask.EMPTY;
```
ehsannas, line 46: Since `mutatedFields` is Nullable, I'm thinking whether it should be the responsibility of the caller (who is creating the OverlayedDocument object) to pass `null` for `mutatedFields` when we have a set/delete mutation, and pass `EMPTY` when there's no mutation. This way, this class won't need to store the `hasOverlay` boolean nor does it have to have this logic.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/OverlayedDocument.java
```diff
@@ -22,21 +22,27 @@
 public class OverlayedDocument {
   private Document overlayedDocument;
   private FieldMask mutatedFields;
```
ehsannas, line 24: should we make this `@Nullable private FieldMask mutatedFields;` ?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalDocumentsView.java
```diff
@@ -185,7 +190,10 @@ private Map<DocumentKey, OverlayedDocument> computeViews(
     for (Map.Entry<DocumentKey, MutableDocument> entry : docs.entrySet()) {
       result.put(
           entry.getKey(),
-          new OverlayedDocument(entry.getValue(), mutatedFields.get(entry.getKey())));
+          new OverlayedDocument(
+              entry.getValue(),
+              keysWithOverlay.contains(entry.getKey()),
```
ehsannas, line 195: the only reason we need `keysWithOverlay` is because of the `recalculateAndSaveOverlays` above, right? Otherwise we could just use `mutatedFields.contains(entry.getKey())`?

It makes me wonder about the discrepancy between `mutatedFields` and `keysWithOverlays`. Let's discuss it offline.

wu-hui: I never thought of it actually, I think it is reasonable. FWIW, short of a SUM type solution, I am pretty indifferent to exactly how we will address this. It's all a hack.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/OverlayedDocument.java
wu-hui: Done.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/OverlayedDocument.java
wu-hui: Done.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
wu-hui: It would if the mutation used as overlay is the same (or can be easily reconstructed manually). They often have different Preconditions (overlay mutation tend to have no preconditions at all), and does not work with ServerTimestamp for example.

Potentially we could try to construct overlays carefully in each call site, but IMO this is exposing too much implementation details in the tests. The contract here is applying Overlay gets the same documents as applying mutations, not Overlay == Mutations.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
wu-hui: See other comment.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalStoreTestCase.java
wu-hui: This would need to be `assertOverlays(..., patchMutation("foo/baz", map(...), Precondition.None))` to pass.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/CountingQueryEngine.java
ehsannas: fair enough.

ehsannas (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/OverlayedDocument.java
```diff
@@ -33,9 +33,14 @@ public Document getDocument() {
...
-   * The fields that are locally mutated by patch mutations. If the overlayed document is from set
-   * or delete mutations, this returns null.
+   * The fields that are locally mutated by patch mutations.
+   *
+   * <p>If the overlayed document is from set or delete mutations, returns null.
+   *
+   * <p>If there is no overlay (mutation) for the document, returns FieldMask.EMPTY.
    */
+  // TODO(b/262245989): This screams for a proper sum type (Tagged Union) which does not exist in
+  // Java (yet).
```
ehsannas, line 43: Is this TODO/comment still useful here?

wu-hui: Nope, deleted.

PR #4444, opened 2022-12-12
Adds support to read the build ids

tyrelrussell:
Reads the resource file containing the list of build ids and their associated library name and architecture.  Writes the payload into the report.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CommonUtils.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CommonUtils.java
@@ -66,6 +66,12 @@ public class CommonUtils {
   static final String MAPPING_FILE_ID_RESOURCE_NAME =
       "com.google.firebase.crashlytics.mapping_file_id";
   static final String LEGACY_MAPPING_FILE_ID_RESOURCE_NAME = "com.crashlytics.android.build_id";
+  static final String BUILD_IDS_LIB_NAMES_RESOURCE_NAME =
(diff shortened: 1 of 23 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CommonUtils.java
```diff
@@ -597,6 +603,42 @@ public static String getMappingFileId(Context context) {
...
+    int archId = CommonUtils.getResourcesIdentifier(context, BUILD_IDS_ARCH_RESOURCE_NAME, "array");
+    int buildId =
+        CommonUtils.getResourcesIdentifier(context, BUILD_IDS_BUILD_ID_RESOURCE_NAME, "array");
+
+    if (libId != 0 && archId != 0 && buildId != 0) {
+      libNames = context.getResources().getStringArray(libId);
+      arch = context.getResources().getStringArray(archId);
+      buildIds = context.getResources().getStringArray(buildId);
+
+      if (libNames.length != buildIds.length || arch.length != buildIds.length) {
```
kmandrika, line 623: What is the purpose of this check?

tyrelrussell: Due to the limitations of the xml that Android resources must conform to, I have to store the arrays separately whic h in theory, e.g. if some enterprising user edited them, they could end up being different lengths and be incoherent.  This simply checks that the lists are all the same length at least.

kmandrika (approved):

mrober (approved):

PR #4457, opened 2022-12-14
[Fireperf][AASA] Experiment time-to-initial-display 3

leotianlizhan:
This PR is the 3rd iteration of experiments for Android App Start Alignment. 
The changes from the 2nd iteration are the following:
1. removed `uptimeMillis`
2. reprioritized trace/subtraces (default trace is now `preDraw + post`)
3. (fixed) real background detection
4. added `preDraw + postAtFrontOfQueue` subtrace
5. added `procStart-to-classLoad` subtrace
6. added `firstForeground` subtrace
7. added `firstBackground` subtrace

### Reasoning
`procStart-to-classLoad` and `firstForegound` can be used to experiment thresholds for filtering out non-detectable backgrounding and interruptions.
`firstBackground` can be used to filter detected backgrounding for all 3 different methodologies (preDraw, preDrawFoQ, onDrawFoQ)

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
@@ -123,7 +123,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
   // TODO: remove after experiment
   private int onDrawCount = 0;
   private final DrawCounter onDrawCounterListener = new DrawCounter();
-  private boolean systemBackgroundCheck = false;
+  private boolean systemForegroundCheck = false;
 
   /**
    * Called from onCreate() method of an activity by instrumented byte code.
@@ -213,7 +213,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
     Context appContext = context.getApplicationContext();
     if (appContext instanceof Application) {
       ((Application) appContext).registerActivityLifecycleCallbacks(this);
-      systemBackgroundCheck = systemBackgroundCheck || isAnyAppProcessInForeground(appContext);
+      systemForegroundCheck = systemForegroundCheck || isAnyAppProcessInForeground(appContext);
       isRegisteredForLifecycleCallbacks = true;
       this.appContext = appContext;
     }
@@ -308,7 +308,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
             .setDurationUs(getStartTimerCompat().getDurationMicros(getClassLoadTimeCompat()))
             .build());
     this.experimentTtid.putCustomAttributes(
-        "systemDeterminedBackground", systemBackgroundCheck ? "true" : "false");
+        "systemDeterminedForeground", systemForegroundCheck ? "true" : "false");
     this.experimentTtid.putCounters("onDrawCount", onDrawCount);
     this.experimentTtid.addPerfSessions(this.startSession.build());
     logExperimentTrace(this.experimentTtid);
@@ -321,7 +321,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
       return;
     }
 
-    systemBackgroundCheck = systemBackgroundCheck || isAnyAppProcessInForeground(appContext);
+    systemForegroundCheck = systemForegroundCheck || isAnyAppProcessInForeground(appContext);
     launchActivity = new WeakReference<Activity>(activity);
     onCreateTime = clock.getTime();
 
--- a/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
@@ -16,13 +16,16 @@ package com.google.firebase.perf.metrics;
 
 import android.annotation.SuppressLint;
 import android.app.Activity;
+import android.app.ActivityManager;
 import android.app.Application;
 import android.app.Application.ActivityLifecycleCallbacks;
 import android.content.Context;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.PowerManager;
 import android.os.Process;
 import android.view.View;
+import android.view.ViewTreeObserver;
 import androidx.annotation.Keep;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -117,6 +120,11 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
   private PerfSession startSession;
   private boolean isStartedFromBackground = false;
 
+  // TODO: remove after experiment
+  private int onDrawCount = 0;
+  private final DrawCounter onDrawCounterListener = new DrawCounter();
+  private boolean systemBackgroundCheck = false;
+
   /**
    * Called from onCreate() method of an activity by instrumented byte code.
    *
@@ -205,6 +213,7 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
     Context appContext = context.getApplicationContext();
     if (appContext instanceof Application) {
       ((Application) appContext).registerActivityLifecycleCallbacks(this);
+      systemBackgroundCheck = systemBackgroundCheck || isAnyAppProcessInForeground(appContext);
       isRegisteredForLifecycleCallbacks = true;
       this.appContext = appContext;
     }
@@ -298,7 +307,9 @@ public class AppStartTrace implements ActivityLifecycleCallbacks, LifecycleObser
             .setClientStartTimeUs(getStartTimerCompat().getMicros())
             .setDurationUs(getStartTimerCompat().getDurationMicros(getClassLoadTimeCompat()))
             .build());
-
+    this.experimentTtid.putCustomAttributes(
+        "systemDeterminedBackground", systemBackgroundCheck ? "true" : "false");
+    this.experimentTtid.putCounters("onDrawCount", onDrawCount);
     this.experimentTtid.addPerfSessions(this.startSession.build());
     logExperimentTrace(this.experimentTtid);
   }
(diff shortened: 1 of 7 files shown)


firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -173,15 +183,14 @@ static AppStartTrace getInstance(TransportManager transportManager, Clock clock)
...
-          Timer.ofElapsedRealtime(startupTime.getElapsedRealtime(), startupTime.getUptimeMillis());
-    }
     this.experimentTtid = TraceMetric.newBuilder().setName("_experiment_app_start_ttid");
+    this.processStartTime =
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.N
+            ? Timer.ofElapsedRealtime(Process.getStartElapsedRealtime())
+            : null;
+    StartupTime firebaseStart = FirebaseApp.getInstance().get(StartupTime.class);
+    this.firebaseClassLoadTime =
+        firebaseStart != null ? Timer.ofElapsedRealtime(firebaseStart.getElapsedRealtime()) : null;
```
visumickey, line 193: Can we leave some comments here on what this section is trying to do?

It might be helpful to qualify this section with a TODO that some of these can be cleaned up after the experiment analysis is done.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -203,92 +213,90 @@ public synchronized void unregisterActivityLifecycleCallbacks() {
...
-        .setDurationUs(start.getDurationUptimeMicros(this.firstDrawDone));
-    this.experimentTtid.addSubtraces(subtrace.build());
-
-    this.experimentTtid.addPerfSessions(this.startSession.build());
-
-    if (isExperimentTraceDone()) {
-      executorService.execute(() -> this.logExperimentTtid(this.experimentTtid));
+        .setClientStartTimeUs(getStartTimer().getMicros())
+        .setDurationUs(getStartTimer().getDurationMicros(this.preDrawPostTime));
+    logExperimentTrace(this.experimentTtid);
```
visumickey, line 262: We are logging logExperimentTrace from two places - `recordPreDraw` and `recordPreDrawFrontOfQueue` (line 299). Are we logging events twice?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -337,42 +346,50 @@ public synchronized void onActivityResumed(Activity activity) {
...
   }
 
-  private void logExperimentTtid(TraceMetric.Builder metric) {
-    transportManager.log(metric.build(), ApplicationProcessState.FOREGROUND_BACKGROUND);
+  /** Helper for logging all experiments in one trace. */
+  private void logExperimentTrace(TraceMetric.Builder metric) {
+    if (this.preDrawPostTime == null
+        || this.preDrawPostAtFrontOfQueueTime == null
+        || this.onDrawPostAtFrontOfQueueTime == null) {
+      return;
```
visumickey, line 372: I'm worried if we would drop traces because of these checks? Are we sure every app start trace would contains these timers? I would rather recommend capturing them and explore in the analysis to see how common is it to have these values as NULL.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -395,31 +412,44 @@ private void logAppStartTrace() {
...
+  public void onAppEnteredForeground() {
+    if (isStartedFromBackground || isTooLateToInitUI || firstForegroundTime != null) {
       return;
     }
-    Timer onPauseTime = clock.getTime();
-    TraceMetric.Builder subtrace =
+    // firstForeground is equivalent to the first Activity onStart. This marks the beginning of
+    // observable backgrounding. Prior to this point, backgrounding cannot be observed.
+    firstForegroundTime = clock.getTime();
+    this.experimentTtid.addSubtraces(
```
visumickey, line 430: What happens if there are multiple foregrounding/backgrounding during app starts. Would we have a multiple subtraces with the same name as "_experiment_firstforegrounding"/"backgrounding"?

visumickey: If we want to capture all of foregrounds/backgrounds, I would remove the term "first" from the naming. Else, would recommend capturing it just once.

firebase-perf/src/main/java/com/google/firebase/perf/util/Timer.java
```diff
@@ -43,23 +43,17 @@ public class Timer implements Parcelable {
...
 
   /**
    * Returns a new Timer object as if it was stamped at the given elapsedRealtime. Uses current
    * wall-clock as a reference to extrapolate the wall-clock at the given elapsedRealtime.
    *
    * @param elapsedRealtimeMillis timestamp in the {@link SystemClock#elapsedRealtime()} timebase
    */
-  public static Timer ofElapsedRealtime(final long elapsedRealtimeMillis, final long uptimeMillis) {
-    long uptimeMicros = MILLISECONDS.toMicros(uptimeMillis);
+  public static Timer ofElapsedRealtime(final long elapsedRealtimeMillis) {
```
visumickey, line 53: NIT (not for now) I got confused with the class name "Timer" here (with [this](https://developer.android.com/reference/java/util/Timer)).

Should we rename this as "TraceTime"/"TraceTimestamp" just to avoid ambiguity?

visumickey (commented):
For the most part, the changes look good to me. Thanks for a quick iteration on this.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: `logExperimentTrace` should start with a check that won't send until all 3 traces are captured. The reason being that there's no guarantee which one will finish first for draw vs predraw, I can explain better in person.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: That's due to limitation that we only want to have 1 experiment trace, and add the rest as subtraces. If I send each trace by themselves then we would have multiple traces, which I don't have a problem with but it's just that it doesn't match our ariane so I wonder if it's okay.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: This should only capture the first one, anything after would not be added to the trace.

firebase-perf/src/main/java/com/google/firebase/perf/util/Timer.java
leotianlizhan: Yea personally I also did not like the name `Timer`. This class is more like a Timestamp than a Timer, but I just kept it the same.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: Done

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -203,92 +215,92 @@ public synchronized void unregisterActivityLifecycleCallbacks() {
...
-      return Timer.ofElapsedRealtime(
-          Process.getStartElapsedRealtime(), Process.getStartUptimeMillis());
+  private @NonNull Timer getClassLoadTimeCompat() {
+    // Prefered: static-initializer time of the 1st Firebase class during init
+    if (firebaseClassLoadTime != null) {
+      return firebaseClassLoadTime;
     }
-    return firebaseStartupTime;
+    // Fallback: static-initializer time of the current class
+    return PERF_CLASS_LOAD_TIME;
```
raymondlam, line 253: No action needed. Just to confirm, we aren't concerned with which class load time to use right? Potentially because the values should still be really close.
It doesn't look like there is a mechanism to determine which one was used.

firebaseClassLoadTime vs PERF_CLASS_LOAD_TIME

raymondlam (approved):

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -173,15 +191,16 @@ static AppStartTrace getInstance(TransportManager transportManager, Clock clock)
...
     this.experimentTtid = TraceMetric.newBuilder().setName("_experiment_app_start_ttid");
+    // Set the timestamp for process-start (beginning of BIND_APPLICATION), if available
+    this.processStartTime =
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.N
+            ? Timer.ofElapsedRealtime(Process.getStartElapsedRealtime())
+            : null;
+    // Set the timestamp for Firebase's first class's class-loading (approx.), if available
+    StartupTime firebaseStart = FirebaseApp.getInstance().get(StartupTime.class);
+    this.firebaseClassLoadTime =
+        firebaseStart != null ? Timer.ofElapsedRealtime(firebaseStart.getElapsedRealtime()) : null;
```
visumickey, line 203: NIT: Would firebaseStart ever be NULL? Since it is classLoad, it should not be NULL I presume? Is this check for completeness?

visumickey (approved):
The changes looks good. Can you ensure that you test these changes on the minimum supported version by Fireperf before merging in the changes?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
leotianlizhan: According to Acore, it could be null when developers disable the ContentProvider (not documented but it's possible in Android). This should be extremely rare but we still need to cover the null case.

PR #4466, opened 2022-12-19
get ready for translation

kaibolay:
- removed unused strings
- extracted hard-coded strings
- added `translation_description`


kaibolay:
oops, wrong base

PR #4480, opened 2022-12-20
Move SharedPreferences usage to background thread

lfkellogg:
Followup from https://github.com/firebase/firebase-android-sdk/pull/4350

Creating a SharedPreferences object can touch storage, so we should do so only on a background thread.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
@@ -124,18 +124,7 @@ class FirebaseAppDistributionImpl implements FirebaseAppDistribution {
           remakeUpdateConfirmationDialog = false;
           dialogHostActivity = null;
 
-          signInStorage
-              .getSignInStatus()
-              .onSuccessTask(
-                  lightweightExecutor,
-                  signedIn -> {
-                    if (signedIn) {
-                      return Tasks.forResult(null);
-                    }
-                    return lifecycleNotifier
-                        .applyToForegroundActivityTask(this::showSignInConfirmationDialog)
-                        .onSuccessTask(lightweightExecutor, unused -> signInTester());
-                  })
+          signInWithConfirmationIfNeeded()
               .onSuccessTask(lightweightExecutor, unused -> checkForNewRelease())
               .continueWithTask(
                   lightweightExecutor,
@@ -181,6 +170,22 @@ class FirebaseAppDistributionImpl implements FirebaseAppDistribution {
 
           return updateTask;
         });
+  }
+
+  @NonNull
+  private Task<Void> signInWithConfirmationIfNeeded() {
+    return signInStorage
+        .getSignInStatus()
+        .onSuccessTask(
+            lightweightExecutor,
+            signedIn -> {
+              if (signedIn) {
+                return Tasks.forResult(null);
+              }
+              return lifecycleNotifier
+                  .applyToForegroundActivityTask(this::showSignInConfirmationDialog)
+                  .onSuccessTask(lightweightExecutor, unused -> signInTester());
+            });
   }
 
   @NonNull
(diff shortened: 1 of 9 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -126,9 +124,18 @@ public UpdateTask updateIfNewReleaseAvailable() {
...
+              .onSuccessTask(
+                  lightweightExecutor,
+                  signedIn -> {
+                    if (signedIn) {
+                      return Tasks.forResult(null);
+                    }
+                    return lifecycleNotifier
+                        .applyToForegroundActivityTask(this::showSignInConfirmationDialog)
+                        .onSuccessTask(lightweightExecutor, unused -> signInTester());
+                  })
```
kaibolay, line 138: maybe extract a method named something like `signInIfNeeded()` or `ensureSignedIn()` which encapsulates this

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -285,35 +296,39 @@ public UpdateTask updateApp() {
...
-                new FirebaseAppDistributionException(
-                    ErrorMessages.DOWNLOAD_URL_NOT_FOUND,
-                    FirebaseAppDistributionException.Status.DOWNLOAD_FAILURE));
-          }
-
-          if (release.getBinaryType() == BinaryType.AAB) {
-            return aabUpdater.updateAab(release);
-          } else {
-            return apkUpdater.updateApk(release, showDownloadInNotificationManager);
+                    "Tester is not signed in", AUTHENTICATION_FAILURE));
```
kaibolay, line 306: Can this use `assertTesterIsSignedIn()`?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
```diff
@@ -16,32 +16,55 @@
...
+  private Task<SharedPreferences> getSharedPreferences() {
+    TaskCompletionSource<SharedPreferences> taskCompletionSource = new TaskCompletionSource<>();
+    backgroundExecutor.execute(
+        () -> taskCompletionSource.setResult(getSharedPreferencesBlocking()));
+    return taskCompletionSource.getTask();
   }
 
-  boolean getSignInStatus() {
-    return signInSharedPreferences.get().getBoolean(SIGNIN_TAG, false);
+  private SharedPreferences getSharedPreferencesBlocking() {
```
kaibolay, line 67: add comment explaining that this might involve I/O

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
kaibolay: actually, this seems like `TesterSignInManager.signInTester()`  - could that be called from here?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TestUtils.java
```diff
@@ -95,6 +101,26 @@ static void awaitAsyncOperations(ExecutorService executorService) throws Interru
...
+        () -> {
+          while (progressEvents.size() < count) {
+            try {
+              Thread.sleep(100);
+            } catch (InterruptedException e) {
+              throw new RuntimeException("Interrupted while waiting for progress events", e);
+            }
+          }
+        });
+    executor.awaitTermination(100, TimeUnit.MILLISECONDS);
```
kaibolay, line 119: Isn't 100 ms too short? The thread sleeps 100 ms for every iteration.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
lfkellogg: `TesterSignInManager.signInTester()` opens the actual sign in screen in the browser.

Agreed that it would be nice to extract this though, if only to make this method more readable. I called it `signInWithConfirmationIfNeeded()`.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
lfkellogg: Oh actually yes! It's a little tricky because I have to use `TaskUtils.onSuccessUpdateTask()` but it works.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
lfkellogg: Done

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TestUtils.java
lfkellogg: Ah, true. Adjusted to 50ms sleeps and a 500ms timeout.

kaibolay (approved):

PR #4484, opened 2022-12-22
Release Generator Rebuilt PR

davidmotson:


buildSrc/src/main/java/com/google/firebase/gradle/ReleaseGenerator.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.eclipse.jgit.api.errors.GitAPIException
+import org.eclipse.jgit.lib.Constants
+import org.eclipse.jgit.lib.ObjectId
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.TaskAction
+
+data class FirebaseLibrary(val moduleNames: List<String>, val directories: List<String>)
+
+open class ReleaseGenerator : DefaultTask() {
+  @TaskAction
```
rlazo, line 15: I *think* the properties below can be read using this pattern 

https://github.com/firebase/firebase-android-sdk/blob/master/buildSrc/src/main/java/com/google/firebase/gradle/plugins/DackkaGenerationTask.kt#L70

davidmotson: I tried to set these locally, however this format doesn't seem to work for command line options, but rather gradle config options.

rlazo (approved):

PR #4486, opened 2022-12-22
Remove synchronized blocks from ApkInstaller

lfkellogg:
Uses a `TaskCompletionSourceCache`, which allows caching tasks which can then be resolved later in a thread-safe manner.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkInstaller.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkInstaller.java
@@ -14,46 +14,35 @@
 
 package com.google.firebase.appdistribution.impl;
 
-import static com.google.firebase.appdistribution.impl.TaskUtils.safeSetTaskException;
-
 import android.app.Activity;
 import android.content.Intent;
-import androidx.annotation.GuardedBy;
 import androidx.annotation.Nullable;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
+import com.google.firebase.annotations.concurrent.Lightweight;
 import com.google.firebase.appdistribution.FirebaseAppDistribution;
 import com.google.firebase.appdistribution.FirebaseAppDistributionException;
+import java.util.concurrent.Executor;
 
 /** Class that handles installing APKs in {@link FirebaseAppDistribution}. */
 class ApkInstaller {
   private static final String TAG = "ApkInstaller";
 
   private final FirebaseAppDistributionLifecycleNotifier lifeCycleNotifier;
+  private final TaskCompletionSourceCache<Void> installTaskCompletionSourceCache;
+  private final @Lightweight Executor lightweightExecutor;
 
-  @GuardedBy("installTaskLock")
-  private TaskCompletionSource<Void> installTaskCompletionSource;
-
-  private final Object installTaskLock = new Object();
-
-  ApkInstaller(FirebaseAppDistributionLifecycleNotifier lifeCycleNotifier) {
+  ApkInstaller(
+      FirebaseAppDistributionLifecycleNotifier lifeCycleNotifier,
+      @Lightweight Executor lightweightExecutor) {
     this.lifeCycleNotifier = lifeCycleNotifier;
-    lifeCycleNotifier.addOnActivityStartedListener(this::onActivityStarted);
+    this.installTaskCompletionSourceCache = new TaskCompletionSourceCache<>(lightweightExecutor);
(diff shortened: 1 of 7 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkInstaller.java
```diff
@@ -84,12 +69,9 @@ private void startInstallActivity(String path, Activity currentActivity) {
   }
 
   void trySetInstallTaskError() {
```
kaibolay, line 71: does this still need to be called "try"?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkInstaller.java
```diff
@@ -14,46 +14,35 @@
...
-
-  void onActivityStarted(@Nullable Activity activity) {
-    synchronized (installTaskLock) {
-      if (installTaskCompletionSource == null
-          || installTaskCompletionSource.getTask().isComplete()
-          || activity == null) {
-        return;
-      }
-    }
+  ApkInstaller(@Lightweight Executor lightweightExecutor) {
```
kaibolay, line 44: Might make sense to inline it, i.e. let the caller(s) call `FirebaseAppDistributionLifecycleNotifier.getInstance()` themselves?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TaskCompletionSourceCache.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private TaskCompletionSource<T> cachedTaskCompletionSource;
+  private final Executor sequentialExecutor;
+
+  /**
+   * Constructor for a {@link TaskCompletionSourceCache} that controls access using its own
+   * sequential executor backed by the given base executor.
+   *
+   * @param baseExecutor Executor (typically {@link Lightweight}) to back the sequential executor.
```
kaibolay, line 46: Are there any cases where this should not (or is not) the "lightweight" executor? 

If so, I think it should be documented which circumstances require a different executor. Does it depend on what  `TaskCompletionSourceProducer.produce()` does?

If not, maybe change "typically" to "always" and annotate the consturctor parameter accordingly.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TaskUtils.java
```diff
@@ -144,5 +145,12 @@ static <T> UpdateTask onSuccessUpdateTask(
     return updateTask;
   }
 
+  /** Set a {@link TaskCompletionSource} to be resolved with the result of another {@link Task}. */
+  static <T> void shadowTask(TaskCompletionSource<T> taskCompletionSource, Task<T> task) {
```
kaibolay, line 149: Can/should this be a method on `TaskCompletionSource` itself?
If so, can/should it use the `TaskCompletionsSource`'s executor instead of the "direct" one?

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/ApkInstaller.java
```diff
@@ -14,46 +14,35 @@
...
+  private final TaskCompletionSourceCache<Void> installTaskCompletionSourceCache;
+  private final @Lightweight Executor lightweightExecutor;
 
-  @GuardedBy("installTaskLock")
-  private TaskCompletionSource<Void> installTaskCompletionSource;
-
-  private final Object installTaskLock = new Object();
-
-  ApkInstaller(FirebaseAppDistributionLifecycleNotifier lifeCycleNotifier) {
+  ApkInstaller(
```
kaibolay, line 35: should this constructor be annotated with `@VisibleForTesting`?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TaskUtils.java
lfkellogg: TaskCompletionSource is provided by the play-services library, so we can't add it there (although we could file a feature request).

Also just a note: `TaskCompletionSource` doesn't actually have its own executor. When you set a result on it, it passes that through to the underlying `TaskImpl`, which executes any handlers on the executors that were specified when they were added. Using direct executor here ensures that any handlers that were themselves added using a direct executor will behave as expected: they'll be executed on the thread that set the result. (I'll add a note to that effect).

PR #4487, opened 2022-12-22
Only enqueue SharedPreferences creation to background

lfkellogg:
Previously all reads and writes were enqueued to the background executor, but that is unnecessary. We want reads/writes to happen ASAP because the sign in related operations on the API (`isTesterSignedIn`, `signOutTester`) return immediately.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
@@ -244,9 +244,8 @@ class FirebaseAppDistributionImpl implements FirebaseAppDistribution {
 
   @Override
   public void signOutTester() {
-    cachedNewRelease
-        .set(null)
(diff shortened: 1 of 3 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -244,9 +244,8 @@ public Task<Void> signInTester() {
 
   @Override
   public void signOutTester() {
-    cachedNewRelease
-        .set(null)
-        .addOnSuccessListener(lightweightExecutor, unused -> signInStorage.setSignInStatus(false));
+    cachedNewRelease.set(null);
+    signInStorage.setSignInStatus(false);
```
lfkellogg, line 248: I also realized there is no point in waiting for one of these changes to be done before executing the other. They should both happen ASAP.

lfkellogg:
Closed in favor of https://github.com/firebase/firebase-android-sdk/pull/4491

PR #4490, opened 2022-12-22
Do not wait to store state when signing out

lfkellogg:
Neither task, either to set the cached release to null or to set the signed it status to false, can ever fail. And the order does not matter. So just kick them off so they can occur as quickly as possible. It's especially important for these changes to occur ASAP since `signOutTester()` returns immediately, and ideally we want the state to reflect that call immediately so the developer sees consistent behavior from the API.

Also adds a new test.

### DIFF
--- a/firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
+++ b/firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
@@ -522,16 +522,10 @@ public class FirebaseAppDistributionServiceImplTest {
       throws InterruptedException, FirebaseAppDistributionException, ExecutionException {
     Task<AppDistributionReleaseInternal> setCachedNewReleaseTask =
         firebaseAppDistribution.getCachedNewRelease().set(TEST_RELEASE_NEWER_AAB_INTERNAL);
-
-    // Confirm that the cached new release is initially set
     awaitTask(setCachedNewReleaseTask);
-    assertThat(setCachedNewReleaseTask.getResult()).isEqualTo(TEST_RELEASE_NEWER_AAB_INTERNAL);
 
-    // Sign out the tester
     firebaseAppDistribution.signOutTester();
-    awaitAsyncOperations(lightweightExecutor);
 
-    // Confirm that the cached new release is now null
     Task<AppDistributionReleaseInternal> cachedNewReleaseTask =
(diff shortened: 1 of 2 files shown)


firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
```diff
@@ -506,12 +511,33 @@ private AlertDialog assertAlertDialogShown() {
...
     verify(mockSignInStorage).setSignInStatus(false);
   }
 
+  @Test
+  public void signOutTester_unsetsCachedNewRelease()
+      throws InterruptedException, FirebaseAppDistributionException, ExecutionException {
+    Task<AppDistributionReleaseInternal> setCachedNewReleaseTask =
+        firebaseAppDistribution.getCachedNewRelease().set(TEST_RELEASE_NEWER_AAB_INTERNAL);
+
+    // Confirm that the cached new release is initially set
```
kaibolay, line 526: I'm not a big fan of this pattern - I think, in general, we should trust (but not verify) test setup.
In this case I think you should be able to drop this comment, the blank line above it, and the assert on line 528

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
```diff
@@ -506,12 +511,33 @@ private AlertDialog assertAlertDialogShown() {
...
+    awaitTask(setCachedNewReleaseTask);
+    assertThat(setCachedNewReleaseTask.getResult()).isEqualTo(TEST_RELEASE_NEWER_AAB_INTERNAL);
+
+    // Sign out the tester
+    firebaseAppDistribution.signOutTester();
+    awaitAsyncOperations(lightweightExecutor);
```
kaibolay, line 532: If I understand this change correctly, this should no be needed

lfkellogg: Ah you're right, great catch!

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionServiceImplTest.java
lfkellogg: Sounds good. Removed.

kaibolay (approved):

PR #4491, opened 2022-12-22
Clean up sign in state management

lfkellogg:
- Speed up typical reads and writes by only using the background executor in SignInStorage to initialize the SharedPreferences object, not protect all references
- Do not complete the Task returned by signInTester until the tester sign in status is updated
- Use an actual SignInStorage instance in TesterSignInManagerTest

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
@@ -24,12 +24,16 @@ import java.util.concurrent.Executor;
 
 /** Class that handles storage for App Distribution SignIn persistence. */
 class SignInStorage {
-
   private static final String SIGNIN_PREFERENCES_NAME = "FirebaseAppDistributionSignInStorage";
   private static final String SIGNIN_TAG = "firebase_app_distribution_signin";
 
   private final Context applicationContext;
   @Background private final Executor backgroundExecutor;
+  private SharedPreferences sharedPreferences;
+
+  private interface SharedPreferencesFunction<T> {
+    T apply(SharedPreferences sharedPreferences);
+  }
 
   SignInStorage(Context applicationContext, @Background Executor backgroundExecutor) {
     this.applicationContext = applicationContext;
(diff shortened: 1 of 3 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
```diff
@@ -24,48 +24,58 @@
...
-  }
-
   private SharedPreferences getSharedPreferencesBlocking() {
     // This may construct a new SharedPreferences object, which requires storage I/O
     return applicationContext.getSharedPreferences(SIGNIN_PREFERENCES_NAME, Context.MODE_PRIVATE);
   }
+
+  private <T> Task<T> applyToSharedPreferences(SharedPreferencesFunction<T> func) {
+    // Check nullness of sharedPreferences directly even though multiple threads could be calling
+    // this function at once. This isn't a problem because: 1) once it is set it will never be set,
```
kaibolay, line 67: nit: is this overlong line something `:googleJavaFormat` would fix?
extra-nit: "set back to null" -> "reset back to null"?

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TesterSignInManagerTest.java
```diff
@@ -228,18 +236,37 @@ public void signInTester_whenSignInCalledMultipleTimes_secondCallHasNoEffect()
...
+    when(signInStorage.setSignInStatus(anyBoolean()))
+        .thenReturn(Tasks.forException(expectedException));
+    Task signInTask = testerSignInManager.signInTester();
+    awaitAsyncOperations(backgroundExecutor);
+    awaitAsyncOperations(blockingExecutor);
+
+    // Simulate re-entering app after successful sign in, via SignInResultActivity
+    testerSignInManager.onActivityCreated(mockSignInResultActivity);
+
+    awaitTaskFailure(signInTask, UNKNOWN, "Error storing tester sign in state", expectedException);
```
kaibolay, line 263: nit: long line

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
lfkellogg: Huh, it doesn't appear overlong to me (100 characters). `:googleJavaFormat` thinks its fine too.

I like "reset back to null", changed.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TesterSignInManagerTest.java
lfkellogg: `:googleJavaFormat` says this is fine too. Looks like 100 chars is the max.

PR #4493, opened 2022-12-22
Make isTesterSignedIn more reliable

lfkellogg:
There was an edge case where, if the first thing you did when opening the app was to call `signOutTester`, calling `isTesterSignedIn` immediately afterwords would return `true`. This makes a few tweaks which should prevent this from happening in any reasonable scenario:

- Sets the sign in status in storage immediately when `signOutTester()` is called
- Stores SharedPreferences even when `getSignInStatusBlocking()` is called

Also adds some more tests to cover the failure mode, and some general test cleanup.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
@@ -16,6 +16,7 @@ package com.google.firebase.appdistribution.impl;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.android.gms.tasks.Tasks;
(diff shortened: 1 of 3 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
```diff
@@ -54,12 +57,14 @@ Task<Boolean> getSignInStatus() {
   }
 
   boolean getSignInStatusBlocking() {
```
kaibolay, line 59: Should this method be renamed?

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/SignInStorage.java
lfkellogg: No I don't think so. It still returns the sign in status, and it is still potentially blocking.

I changed `getSharedPreferencesBlocking()` to `getAndCacheSharedPreferences()` because if I included `Blocking` it would be super long and I didn't think it was necessary for the private method name to communicate that it blocks.

PR #4508, opened 2023-01-03
Add DackkaPlugin tests

daymxn:
Per [b/249825782](https://b.corp.google.com/issues/249825782),

This adds tests to `build-src-check` for the `DackkaPlugin`. These tests come with a predefined output that is expected from Dackka, and will rerun the project used to generate said output- and compare it with the expected output. You can read more about it in the documentation for `DackkaPluginTests`.

This also adds a `testPlugins` that can be used instead of passing `enablePluginTests` around the command line. Although support remains for `enablePluginTests`.

Example:
```
./gradlew -b buildSrc/build.gradle.kts testPlugins
```


buildSrc/build.gradle.kts
```diff
@@ -98,17 +98,30 @@ gradlePlugin {
     }
 }
```
rlazo, line 100: Why these changes enable unconditionally the test? The old code left it to the caller to enable it or not, and now the `-PenablePluginTests` wouldn't work, right?

daymxn: If you look at line 117-120, it still retains the old functionality of only activating if the property is set. These new tasks I'm registering do *not* get called automatically when tests are ran. They have to be explicitly called. They exist so you can just call the task itself if you wanna run plugin tests, versus calling test with the `enablePluginTests` parameter. Although both methods work totally fine still.

It's worth noting that `enablePluginTest` is basically the same thing as `enableTests` (this is true of the current impl and the previous impl). We just disable all buildSrc tests unless the param is specified. Here, I offer an alternative task you can call instead- which allows your to use gradle tasks windows and such instead of needing to find the param for enabling tests.

buildSrc/src/test/kotlin/com/google/firebase/gradle/plugins/DackkaPluginTests.kt
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * We do this by running the [DackkaPlugin] against a small project fixture and comparing the output
+ * to a pre-compiled output that represents what we expect Dackka to generate.
+ *
+ * ## Resources
+ *
+ * The resources for the tests are stored under `src/test/resources/dackka-plugin-tests` with two
+ * sub directories that will be explained below.
+ *
+ * ### project Directory containing a small gradle project, with various sub-projects. These exist
```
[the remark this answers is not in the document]
daymxn: You are right!

buildSrc/build.gradle.kts
rlazo: Got it, thanks for the clarification!

rlazo:
Ping me when the javadoc changes are final. I'd like to take a quick look at them before approving the PR. Thanks!

PR #4511, opened 2023-01-04
Separate thread work from handler and add a wrapper around the Http work

qdpham13:
* Refactors Realtime
    * Moves all thread work into ConfigRealtimeHttpClient
    * Adds ConfigRealtimeHttpStream class that is a wrapper around the HttpUrlConnection
* Fixes bug/261868622


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpStream.java
```diff
@@ -0,0 +1,341 @@
+// Copyright 2022 Google LLC
```
danasilver, line 1: I'm a bit confused about the separation between `ConfigRealtimeHttpClient` and `ConfigRealtimeHttpStream` since `ConfigRealtimeHttpClient` still maintains a reference to the `HttpUrlConnection` and the business logic is split between the two classes. I think it'd be easier to read all the HTTP connection logic in one place. Thoughts?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHandler.java
```diff
@@ -55,63 +51,28 @@ public ConfigRealtimeHandler(
...
+            namespace,
+            listeners,
+            scheduledExecutorService);
   }
 
-  private synchronized boolean canCreateRealtimeHttpClientTask() {
-    return !listeners.isEmpty() && realtimeHttpClientTask == null;
-  }
-
-  private synchronized Runnable createRealtimeHttpClientTask(
```
danasilver, line 73: Nice - makes sense to move this out of the handler

danasilver (changes requested):
Can the other tests be added to this PR? I'd keep the changed code and tests together if possible to make reviewing and potentially reverting easier.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpStream.java
qdpham13: In ConfigRealtimeHttpClient there's actually no reference to HttpUrlConnection; it's a reference to the FutureTask that is returned after submitting the ConfigRealtimeHttpStream thread to the executor so that we can know if we have a thread running or not. The reasons why I want to keep the Http work separated into it's own class are so it's centralized in a place where there's no other type of work going on and to make it more thread safe since that background thread will be only confined to the ConfigRealtimeHttpStream class and it won't touch any other classes/variables

qdpham13:
> Can the other tests be added to this PR? I'd keep the changed code and tests together if possible to make reviewing and potentially reverting easier.

Added back the tests

PR #4514, opened 2023-01-04
Move Firebase Installations to the new, injected executors

davidmotson:


firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
```diff
@@ -140,7 +141,10 @@ public class FirebaseInstallationsTest {
   public void setUp() {
     MockitoAnnotations.initMocks(this);
     FirebaseApp.clearInstancesForTest();
-    executor = new ThreadPoolExecutor(0, 1, 30L, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
+    backgroundExecutor =
```
rlazo, line 144: Can you use `com.google.firebase.concurrent.TestOnlyExecutors` ?

davidmotson: Tried to use a few of the executors available there, but they all seem to cause tests to fail from too many invocations. I spent a bit of time trying to figure out why, but ended up rolling back the change.

rlazo: Looking into why this is the case more careful, I think tests are failing correctly. Using the `TestOnlyExecutors` in the tests do cause the backend to be called multiple times because it's being called concurrently multiple times.

The source of the issue is that previously FIS was using a single-thread executor for all tasks, which meant that they were not actually being executed concurrently. 

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-installations/src/main/java/com/google/firebase/installations/FirebaseInstallations.java#L93

Moving FIS to common executors is changing this, thus making the test to fail.

We could either see if making FIS truly concurrent is viable, or migrate them to use sequential executor

https://firebase.github.io/firebase-android-sdk/components/executors/#sequential-executor

davidmotson: After looking through the synchronization model here it seems that the network executor is the problem

The concurrent network requests just shouldn't happen, so I've replaced the network executor with a sequential one, i've also replicated this setup in the test, and this appears to have fixed it.

firebase-installations/firebase-installations.gradle
```diff
@@ -44,6 +44,8 @@ dependencies {
 
     implementation 'androidx.multidex:multidex:2.0.1'
     implementation 'com.google.android.gms:play-services-tasks:18.0.1'
+    implementation project(path: ':firebase-annotations')
+    testImplementation project(path: ':integ-testing')
```
rlazo, line 48: move it together with the rest of `testImplementation` declarations.

firebase-installations/src/test/java/com/google/firebase/installations/FirebaseInstallationsTest.java
```diff
@@ -140,7 +144,10 @@ public class FirebaseInstallationsTest {
   public void setUp() {
     MockitoAnnotations.initMocks(this);
     FirebaseApp.clearInstancesForTest();
-    executor = new ThreadPoolExecutor(0, 1, 30L, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
+    backgroundExecutor = TestOnlyExecutors.background();
+    networkExecutor =
+        FirebaseExecutors.newSequentialExecutor(
+            new ThreadPoolExecutor(0, 5, 30L, TimeUnit.SECONDS, new LinkedBlockingQueue<>()));
```
rlazo, line 150: Use the blocking executor from TestOnlyExecutors here too. The interface they implement is https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html which implements both `Executor` and `ExecutorService`.

rlazo (approved):

daymxn:
@davidmotson Should this invoke a release? If so, could we add something to the changelog about it in a separate commit?

PR #4518, opened 2023-01-04
Improve readability of a test in ApkUpdaterTest

lfkellogg:
- The latch's name made it seem like it was a mock
- Inlining more of the setup makes it more readable

### DIFF
--- a/firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/ApkUpdaterTest.java
+++ b/firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/ApkUpdaterTest.java
@@ -42,6 +42,7 @@ import com.google.firebase.appdistribution.FirebaseAppDistributionException.Stat
 import com.google.firebase.appdistribution.UpdateProgress;
 import com.google.firebase.appdistribution.UpdateStatus;
 import com.google.firebase.appdistribution.UpdateTask;
+import com.google.firebase.concurrent.FirebaseExecutors;
 import com.google.firebase.concurrent.TestOnlyExecutors;
 import java.io.ByteArrayInputStream;
 import java.io.File;
@@ -186,12 +187,12 @@ public class ApkUpdaterTest {
         .getInputStream();
     // Block thread actually making the request on a latch, which gives us time to add listeners to
     // the returned UpdateTask in time to get all the progress updates
-    CountDownLatch mockConnectionLatch = new CountDownLatch(1);
+    CountDownLatch countDownLatch = new CountDownLatch(1);
     when(mockHttpsUrlConnectionFactory.openConnection(TEST_URL))
         .thenAnswer(
             invocation -> {
               try {
-                mockConnectionLatch.await();
+                countDownLatch.await();
               } catch (InterruptedException e) {
                 throw new AssertionError("Interrupted while waiting in mock");
               }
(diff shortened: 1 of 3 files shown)


firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/ApkUpdaterTest.java
```diff
@@ -202,7 +202,7 @@ public void updateApk_whenSuccessfullyUpdated_notificationsSetCorrectly()
 
     UpdateTask updateTask = apkUpdater.updateApk(TEST_RELEASE, true);
     List<UpdateProgress> events =
-        awaitProgressEvents(updateTask, 3, mockConnectionLatch::countDown);
+        awaitProgressEvents(updateTask, 3, countDownLatch::countDown);
```
kaibolay, line 205: I think this is difficult to understand. It might make sense to split up the `awaitProgressEvents()` method to make this more readable. E.g.:

- declare the `progressEvents` lists here (inline)
- add the listener here (inline)
- count down the latch here (inline)
- call the truncated helper function: `awaitProgressEvents(progressEvents, 3)`

lfkellogg: Yes I like that a lot better, thanks!

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TestUtils.java
```diff
@@ -104,24 +105,15 @@ static void awaitAsyncOperations(ExecutorService executorService) throws Interru
   /** Await a specified number of progress events on an {@link UpdateTask}. */
   static List<UpdateProgress> awaitProgressEvents(UpdateTask updateTask, int count)
```
kaibolay, line 106: Aren't all callers of this prone to the problem that the task might have completed before the listener is added?
I think it might make sense to inline this everywhere (and employ a latch to prevent the race)

lfkellogg: Yeah I think you're right. There's only one other usage and maybe we're just getting lucky that it hasn't been failing. Updated.

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TestUtils.java
```diff
@@ -101,27 +98,9 @@ static void awaitAsyncOperations(ExecutorService executorService) throws Interru
     shadowOf(getMainLooper()).idle();
   }
 
-  /** Await a specified number of progress events on an {@link UpdateTask}. */
-  static List<UpdateProgress> awaitProgressEvents(UpdateTask updateTask, int count)
+  /** Await a specified number of progress events being added to the given collection. */
+  static void awaitProgressEvents(Collection<UpdateProgress> progressEvents, int count)
```
kaibolay, line 102: Sorry, but I keep having ideas for this:

I think generalizing this would actually make it more readable. It doesn't need to be full-featured like http://google3/java/com/google/testing/util/WaitForCondition.java

How about:

```
  static void waitForCondition(BooleanSupplier condition) {
      ExecutorService executor = TestOnlyExecutors.blocking();
    executor.execute(
        () -> {
          while (!condition.getAsBoolean()) {
            try {
              Thread.sleep(50);
              throw new RuntimeException("Interrupted while waiting for progress events", e);
            }
          }
        });
    executor.awaitTermination(500, TimeUnit.MILLISECONDS);
    assertThat(condition.getAsBoolean());
  }
```

That way the caller becomes more expressive:

```
  waitForCondition(() -> progressEvents.size() == 3);
```

kaibolay (approved):
Sorry 'bout all the bikeshedding

firebase-appdistribution/src/test/java/com/google/firebase/appdistribution/impl/TestUtils.java
lfkellogg: No I think this makes more sense now that most of the progress event specific logic is gone. I also removed the Executor since I realized that wasn't actually adding value. It was also inadvertently causing us to wait a minimum of 500ms each time.

PR #4520, opened 2023-01-04
Remove synchronized blocks from TesterSignInManager

lfkellogg:
Also moves to `@LightweightExecutor` now that there are no more blocking operations.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterSignInManager.java
```diff
@@ -120,27 +109,25 @@ void onActivityResumed(Activity activity) {
...
+    if (hasBeenSentToBrowserForCurrentTask) {
+      signInTaskCompletionSourceCache
+          .setException(
               new FirebaseAppDistributionException(
-                  ErrorMessages.AUTHENTICATION_CANCELED, AUTHENTICATION_CANCELED));
-        }
-      }
+                  ErrorMessages.AUTHENTICATION_CANCELED, AUTHENTICATION_CANCELED))
+          .addOnSuccessListener(
+              unused -> LogWrapper.e(TAG, "App Resumed without sign in flow completing."));
```
kaibolay, line 121: nit: "resumed" should be lower case.

kaibolay (approved):

PR #4523, opened 2023-01-04
Support multiple progress listeners on an UpdateTask

lfkellogg:
This is a bug that we should probably fix, even though the typical customer use case is to add only one listener. [The API ](https://firebase.google.com/docs/reference/android/com/google/firebase/appdistribution/UpdateTask#addOnProgressListener(com.google.firebase.appdistribution.OnProgressListener)) definitely seems like it's meant to support multiple listeners.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskImpl.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskImpl.java
@@ -55,7 +55,7 @@ class UpdateTaskImpl extends UpdateTask {
 
   UpdateTaskImpl() {
     synchronized (taskCompletionLock) {
-      this.taskCompletionSource = new TaskCompletionSource<>();
+      taskCompletionSource = new TaskCompletionSource<>();
     }
   }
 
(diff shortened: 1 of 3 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskImpl.java
```diff
@@ -94,9 +97,9 @@ public UpdateTask addOnProgressListener(
       @Nullable Executor executor, @NonNull OnProgressListener listener) {
     ManagedListener managedListener = new ManagedListener(executor, listener);
     synchronized (lock) {
-      this.listener = managedListener;
+      this.listeners.add(managedListener);
```
kaibolay, line 100: nit: remove `this.` here and in all other places where it's not explicitly needed

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/UpdateTaskImpl.java
lfkellogg: Done!

PR #4524, opened 2023-01-05
Implement BloomFilter class

milaGGL:
- [x] Implement Bloomfilter class
- [x] write validation test cases for Bloomfilter
- [x] add golden test


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,146 @@
...
+    BigInteger bigInteger1 = new BigInteger(Long.toUnsignedString(num1));
+    BigInteger bigInteger2 = new BigInteger(Long.toUnsignedString(num2));
+
+    // Calculate hashed value h(i) = h1 + (i * h2).
+    BigInteger hashValue = bigInteger1.add(bigInteger2.multiply(BigInteger.valueOf(index)));
+
+    // Wrap if hash value overflow 64bit.
+    if (hashValue.compareTo(this.MAX_64_BIT_UNSIGNED_INTEGER) == 1) {
+      hashValue = new BigInteger(Long.toUnsignedString(hashValue.longValue()));
+    }
```
milaGGL, line 122: TODO: looking for a better way to keep only the lowest 64bit when overflow

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Arrays;
+
+public class BloomFilter {
+  private static final String TAG = "BloomFilter";
+
+  private final int size;
+  private final byte[] bitmap;
+  private final int hashCount;
+
+  public BloomFilter(@NonNull byte[] bitmap, @NonNull int padding, @NonNull int hashCount) {
```
dconeybe, line 30: nit: Remove `@NonNull` for primitive types (i.e. int, long, boolean, byte, float, double, char, short) because they are not nullable.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+        throw new IllegalArgumentException("Invalid padding when bitmap length is 0: " + padding);
+      }
+    }
+    this.bitmap = bitmap;
+    this.hashCount = hashCount;
+    this.size = bitmap.length * 8 - padding;
+  }
+
+  @NonNull
+  public int getSize() {
```
dconeybe, line 56: Add javadocs for getSize() and isEmpty(). The word "size" in this context is unclear: is it referring to the number of bytes in the bitmap or the number of bits of the bitmap that are part of the bloom filter? If possible, remove these methods altogether (or make them private) since I don't foresee them being useful outside of this class. If I'm wrong and they _are_ useful, then feel free to leave them as-is. If they are only used for unit testing, them give them default visibility (by removing the `public` visibilitly modifier) and add the `@VisibleForTesting` annotation.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public static byte[] MD5Hash(String value) {
+    try {
+      MessageDigest digest = MessageDigest.getInstance("MD5");
+      digest.update(value.getBytes());
+      return digest.digest();
+    } catch (NoSuchAlgorithmException e) {
+      Logger.warn(TAG, "Could not create hashing algorithm: MD5.", e);
+      return null;
```
dconeybe, line 94: Rather than returning null, re-throw the exception in a RuntimeException. `MessageDigest.getInstance("MD5")` should _never_ throw because Android always supports "MD5": https://developer.android.com/reference/java/security/MessageDigest. You should just be able to write `throw new RuntimeException(e);`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+    for (int i = 0; i < this.hashCount; i++) {
+      int index = this.getBitIndex(hash1, hash2, i);
+      if (!this.isBitSet(index)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  public static byte[] MD5Hash(String value) {
```
dconeybe, line 87: nit: Add `@NonNull` to `MD5Hash()`. Also, rename it to `md5Hash` because it is unconventional to start Android method names with uppercase characters.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+      if (padding != 0) {
+        throw new IllegalArgumentException("Invalid padding when bitmap length is 0: " + padding);
+      }
+    }
+    this.bitmap = bitmap;
+    this.hashCount = hashCount;
+    this.size = bitmap.length * 8 - padding;
+  }
+
+  @NonNull
```
dconeybe, line 55: Remove `@NonNull` from `getSize()` since its return value is a primitive integer.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  public static byte[] MD5Hash(String value) {
+    try {
+      MessageDigest digest = MessageDigest.getInstance("MD5");
+      digest.update(value.getBytes());
```
dconeybe, line 90: In the spirit of minimizing the amount of code in a try/catch block, move everything out of the try/catch block except the call that potentially throws the NoSuchAlgorithmException (i.e. the call to `MessageDigest.getInstance("MD5")`).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public boolean mightContain(@NonNull String value) {
+    // Empty bitmap or empty value should always return false on membership check.
+    if (this.isEmpty() || value.isEmpty()) {
+      return false;
+    }
+
+    byte[] md5HashedValue = this.MD5Hash(value);
+    if (md5HashedValue == null || md5HashedValue.length != 16) {
```
dconeybe, line 71: If `md5HashedValue.length != 16` then that is a critical, irrecoverable error. Throw a RuntimeException in this case with a message like `"invalid md5HashedValue.length: " + md5HashedValue.length + " (expected 16)"`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
dconeybe, line 90: I think you can just return digest.digest(value.getBytes()) and skip the call to update().

https://developer.android.com/reference/java/security/MessageDigest?hl=en#digest(byte[])

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+    long result = 0;
+    for (int i = 0; i < 8 && i < bytes.length; i++) {
+      result |= (bytes[offset + i] & 0xFFL) << (i * 8);
+    }
+    return result;
+  }
+
+  // Calculate the ith hash value based on the hashed 64bit integers,
+  // and calculate its corresponding bit index in the bitmap to be checked.
+  private int getBitIndex(long num1, long num2, int index) {
```
dconeybe, line 109: nit: Rename `num1` and `num2` to `hash1` and `hash2`, respectively.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+    // Calculate hashed value h(i) = h1 + (i * h2).
+    Long hashValue2 = num1 + num2 * index;
```
dconeybe, line 111: nit: Use `long` instead of `Long` here. There is no need to create a boxed Long object.

dconeybe, line 111: nit: Add parentheses around `num2 * index` to be clear about the order of operations (if the linter lets you do this).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+    return (byteAtIndex & (0x01 << offset)) != 0;
+  }
+
+  @Override
+  public String toString() {
+    return "BloomFilter{"
+        + "bitmap="
+        + Arrays.toString(bitmap)
+        + ", hashCount="
+        + hashCount
```
dconeybe, line 129: Maybe put bitmap at the _end_ of the String returned from toString() so that the hashCount and size are easy to find in the string (i.e. you don't have to scroll to the end of a potentially-long bitmap string).

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,134 @@
...
+    byte byteAtIndex = this.bitmap[(index / 8)];
+    int offset = index % 8;
+    return (byteAtIndex & (0x01 << offset)) != 0;
+  }
+
+  @Override
+  public String toString() {
+    return "BloomFilter{"
+        + "bitmap="
+        + Arrays.toString(bitmap)
```
dconeybe, line 127: Consider using Base64 to produce a more compact string representation that Arrays.toString().

https://developer.android.com/reference/android/util/Base64

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,169 @@
...
+   *
+   * <p>Full document path is generated by concatenating documentPrefix and number n, eg,
+   * projects/project-1/databases/database-1/documents/coll/doc12.
+   *
+   * <p>The test result is generated by checking the membership of documents from documentPrefix+0
+   * to documentPrefix+2n. The membership results from 0 to n is expected to be true, and the
+   * membership results from n to 2n is expected to be false with some false positive results.
+   */
+  @Test
+  @SuppressWarnings("DefaultCharset")
```
dconeybe, line 118: Please find a way to remove this `@SuppressWarnings("DefaultCharset")` annotation. It should be pretty easy to fix by simply specifying "utf8" when decoding strings from bytes.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,169 @@
...
+  public void testBloomFilterGoldenTest() throws Exception {
+    String documentPrefix = "projects/project-1/databases/database-1/documents/coll/doc";
+
+    // Import the golden test files for bloom filter
+    HashMap<String, JSONObject> parsedSpecFiles = new HashMap<>();
+    File jsonDir = new File("src/test/resources/bloom_filter_golden_test_data");
```
dconeybe, line 124: The golden files should be loaded from Android "resources", like how the spec test json files are.

The spec test json files are located in https://github.com/firebase/firebase-android-sdk/tree/master/firebase-firestore/src/test/resources/json. The directory name "json" is entirely arbitrary, so choose a different directory name for your golden test files.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,169 @@
...
+    File[] jsonFiles = jsonDir.listFiles();
+    for (File f : jsonFiles) {
```
dconeybe, line 126: nit: Use the full name "file" for the variable name, instead of "f". Non-standard abbreviations tend to detract from readability. In this case especially, saving 3 characters is not justified IMO.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,169 @@
...
+      // Find corresponding membership test result.
+      JSONObject resultJSON =
+          parsedSpecFiles.get(fileName.replace("bloom_filter_proto", "membership_test_result"));
+      String membershipTestResults = resultJSON.getString("membershipTestResults");
+
+      // Run and compare mightContain result with the expectation.
+      for (int i = 0; i < membershipTestResults.length(); i++) {
+        boolean expectedMembershipResult = membershipTestResults.charAt(i) == '1';
+        boolean mightContain = bloomFilter.mightContain(documentPrefix + i);
+        assertEquals(mightContain, expectedMembershipResult);
```
dconeybe, line 165: Use Truth here instead of the base assertEquals. Using Truth I believe there is a way to attach some contextual information to the failure message. It would be really helpful if the failure message included the name of the golden file and the value of `i` that failed.

dconeybe (changes requested):

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
milaGGL: The test cases are under Android "resources", under a directory named "bloom_filter_golden_test_data".

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
milaGGL: AssertEquals can take messages as well, and java keeps suggesting me simplify assertTrue with assertEquals.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,131 @@
...
+package com.google.firebase.firestore.remote;
+
+import android.util.Base64;
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class BloomFilter {
+  private final int size;
```
milaGGL, line 24: could the size go over 2^32 bit? It happens when the bitmap length is over 536,870,912.

dconeybe: No. The size of the bitmap for 1000 entries is about 8000 bytes, which equals 64000 bits. The bitmap will never be large enough to overflow an int.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
dconeybe: Okay I see that this is how the spec tests work too. Please ignore my comment. It was in error.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,149 @@
...
+      if (padding != 0) {
+        throw new IllegalArgumentException("Invalid padding when bitmap length is 0: " + padding);
+      }
+    }
+    this.bitmap = bitmap;
+    this.hashCount = hashCount;
+    this.size = bitmap.length * 8 - padding;
+  }
+
+  /** Return if a bloom filter is empty. */
```
dconeybe, line 53: You can remove this javadoc comment for isEmpty(). Its name is self-explanatory. If you add back getSize() then that one should have a javadoc comment since "size" is ambiguous (number of bytes or number of bits).

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest = Config.NONE)
+public class BloomFilterTest {
+
+  @Test
+  public void testEmptyBloomFilter() {
```
dconeybe, line 40: Names of test methods in Java don't need to begin with "test" since they are annotated with `@Test`. Instead, please rename the methods to something like "mightContainShouldReturnFalseOnEmptyBloomFilter". Please rename the other test methods to follow this convention.

dconeybe, line 40: For an empty bloom filter, instead of checking isEmpty() (a private method) just call mightContain() with a bunch of different values, including the empty string, and expect it to return false every time.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,178 @@
...
+    BloomFilter bloomFilter = new BloomFilter(new byte[0], 0, 0);
+    assertTrue(bloomFilter.isEmpty());
+  }
+
+  @Test
+  public void testEmptyBloomFilterThrowException() {
```
dconeybe, line 46: An another example, a better name for "testEmptyBloomFilterThrowException" is "constructorShouldThrowIAEOnNegativeHashValue". Also, this test case should be split up so that each test case tests only one thing. The tests for different invalid arguments should be in different test methods.

go/unit-testing-practices#behavior-testing-examples

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,178 @@
...
+   * <p>Full document path is generated by concatenating documentPrefix and number n, eg,
+   * projects/project-1/databases/database-1/documents/coll/doc12.
+   *
+   * <p>The test result is generated by checking the membership of documents from documentPrefix+0
+   * to documentPrefix+2n. The membership results from 0 to n is expected to be true, and the
+   * membership results from n to 2n is expected to be false with some false positive results.
+   */
+  @Test
+  public void testBloomFilterGoldenTest() throws Exception {
+    String documentPrefix = "projects/project-1/databases/database-1/documents/coll/doc";
```
dconeybe, line 119: Make `documentPrefix` a private static final constant in this class, named something like `GOLDEN_DOCUMENT_PREFIX`.

dconeybe, line 119: I'd like to create one test method per golden test. Otherwise, the first golden test failure will hide any other failures from any other tests. I think the test way to do this would be to create a helper method like "runGoldenTest(String name)" and then a bunch of `@Test` methods that just call the runGoldenTest() method with the name of the test, which will be the file name.

dconeybe (changes requested):

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
milaGGL: Strange, this is already removed in local and pushed up.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -115,27 +115,27 @@ public void testBloomFilterMightContainOnEmptyStringAlwaysReturnFalse() {
...
+    assert jsonFiles != null;
+    for (File file : jsonFiles) {
+      if (!file.toString().endsWith(".json")) {
         continue;
       }
 
       // Read the files into a map.
       StringBuilder builder = new StringBuilder();
-      BufferedReader reader = new BufferedReader(new FileReader(f));
+      BufferedReader reader = new BufferedReader(new FileReader(file));
```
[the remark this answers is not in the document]
milaGGL: using InputStreamReader instead:
Screenshot 2023-01-09 at 3 19 12 PM

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
milaGGL: This mightContain on empty BloomFilter is tested separately below.  
Related test case: non empty string on empty bloom filter, empty string on empty bloom filter, empty string on non empty bloom filter. All should return false.

dconeybe (changes requested):
I'm not quite done my review, but here's some feedback that I've accumulated so far.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return "BloomFilter{"
+        + "hashCount="
+        + hashCount
+        + ", size="
+        + size
+        + ", bitmap=\""
```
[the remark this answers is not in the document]
milaGGL: Either way works for me.  But without the "", calling toString  will return this:
`BloomFilter{hashCount=1, size=7, bitmap=AQ==}`
In contrast:`BloomFilter{hashCount=1, size=7, bitmap="AQ=="}` looked prettier to me, especially with the "=" trailing at the end.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    return this.bitCount;
+  }
+
+  /**
+   * Check whether the given string is a possible member of the bloom filter. It might return false
+   * positive result, ie, the given string is not a member of the bloom filter, but the method
+   * returned true.
+   *
+   * @param value the string to be tested for membership.
+   * @return true if the given string might be contained in the bloom filter.
```
dconeybe, line 70: Add something like ", or false if the given string is definitely not contained in the bloom filter" after "true if the given string might be contained in the bloom filter".

I generally recommend to explicitly document the situations under which "true" and "false" are returned from a method, even if it's "obvious". It may not be so "obvious" to someone else.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /** Hash a string using md5 hashing algorithm, and return an array of 16 bytes. */
+  @NonNull
+  private byte[] md5HashDigest(@NonNull String value) {
+    return md5HashMessageDigest.digest(value.getBytes(StandardCharsets.UTF_8));
+  }
+
+  @NonNull
+  private MessageDigest createMd5HashMessageDigest() {
```
dconeybe, line 103: nit: add the `static` modifier to `createMd5HashMessageDigest()`.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    try {
+      return  MessageDigest.getInstance("MD5");
```
dconeybe, line 105: nit: Collapse the two spaces before `MessageDigest` to a single space.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,168 @@
...
+      result |= (bytes[offset + i] & 0xFFL) << (i * 8);
+    }
+    return result;
+  }
+
+  /**
+   * Calculate the ith hash value based on the hashed 64 bit unsigned integers, and calculate its
+   * corresponding bit index in the bitmap to be checked.
+   */
+  private int getBitIndex(long hash1, long hash2, int index) {
```
dconeybe, line 124: Consider renaming the `index` argument to something like `hashIndex` or `hashNum`. There are two "indexes" being discussed in the documentation and IMO it's helpful to be clear which index is being referred to.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,168 @@
...
+  private static long unsignedRemainder(long dividend, long divisor) {
+    long quotient = ((dividend >>> 1) / divisor) << 1;
+    long remainder = dividend - quotient * divisor;
+    return remainder - (remainder >= divisor ? divisor : 0);
+  }
+
+  /** Return whether the bit at the given index in the bitmap is set to 1. */
+  private boolean isBitSet(int index) {
+    // To retrieve bit n, calculate: (bitmap[n / 8] & (0x01 << (n % 8))).
+    byte byteAtIndex = this.bitmap[(index / 8)];
```
dconeybe, line 152: nit: Remove the parentheses around `(index / 8)` since they are superfluous.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+
+  @Test
+  public void instantiateEmptyBloomFilter() {
+    BloomFilter bloomFilter = new BloomFilter(new byte[0], 0, 0);
+    assertEquals(bloomFilter.getBitCount(), 0);
+  }
+
+  @Test
+  public void instantiateNonEmptyBloomFilter() {
+    BloomFilter bloomFilter1 = new BloomFilter(new byte[] {1}, 0, 1);
```
dconeybe, line 51: For tests like `instantiateNonEmptyBloomFilter()`, which effectively test more than one thing, separate each test into its own logic scope using `{` and `}`. This provides several benefits, including

1. It logically groups the two tests together, improving readability.
2. It prevents accidental usage of variables from one test in another, improving robustness.
3. Variable names can be re-used, increasing the symmetry between the tests, making it easier for a human to see what's the same and what's different between the cases.

For example, the body of `instantiateNonEmptyBloomFilter()` would become this:

```java
{
  BloomFilter bloomFilter = new BloomFilter(new byte[] {1}, 0, 1);
  assertEquals(bloomFilter.getBitCount(), 8);
}
{
  BloomFilter bloomFilter = new BloomFilter(new byte[] {1}, 7, 1);
  assertEquals(bloomFilter.getBitCount(), 1);
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/BloomFilter.java
```diff
@@ -0,0 +1,168 @@
...
+        throw new IllegalArgumentException("Invalid hash count: " + hashCount);
+      }
+    } else {
+      if (hashCount < 0) {
+        throw new IllegalArgumentException("Invalid hash count: " + hashCount);
+      }
+
+      // Empty bloom filter should have 0 padding.
+      if (padding != 0) {
+        throw new IllegalArgumentException("Invalid padding when bitmap length is 0: " + padding);
```
dconeybe, line 50: Suggested re-wording of this exception's message:

```
"Invalid padding when bitmap length is 0: " + padding
```

to

```
"Expected padding of 0 when bitmap length is 0, but got " + padding
```

The re-wording is a bit better because it also conveys the _expected_ value, not just the _invalid_ value.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+    IllegalArgumentException emptyBloomFilterException =
+        assertThrows(IllegalArgumentException.class, () -> new BloomFilter(new byte[0], 0, -1));
+    assertThat(emptyBloomFilterException).hasMessageThat().contains("Invalid hash count: -1");
+    IllegalArgumentException nonEmptyBloomFilterException =
+        assertThrows(IllegalArgumentException.class, () -> new BloomFilter(new byte[] {1}, 1, -1));
+    assertThat(nonEmptyBloomFilterException).hasMessageThat().contains("Invalid hash count: -1");
+  }
+
+  @Test
+  public void constructorShouldThrowIAEOnOverflowPadding() {
```
dconeybe, line 102: nit: Suggest renaming `constructorShouldThrowIAEOnOverflowPadding()` to `constructorShouldThrowIAEIfPaddingIsTooLarge()`. To me, the term "overflow" (incorrectly) implies that the value was once valid but grew to size that overflowed the valid range.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+   *
+   * <p>Full document path is generated by concatenating documentPrefix and number n, eg,
+   * projects/project-1/databases/database-1/documents/coll/doc12.
+   *
+   * <p>The test result is generated by checking the membership of documents from documentPrefix+0
+   * to documentPrefix+2n. The membership results from 0 to n is expected to be true, and the
+   * membership results from n to 2n is expected to be false with some false positive results.
+   */
+  private void runGoldenTest(String testFile) throws Exception {
+    String resultFile = testFile.replace("bloom_filter_proto", "membership_test_result");
```
dconeybe, line 151: Verify that `resultFile.equals(testFile) == false`. If the `testFile` does not contain "bloom_filter_proto" then both `testFile` and `resultFile` will be equal, which is not right. Throw IllegalArgumentException in this case.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+
+    String membershipTestResults = resultJSON.getString("membershipTestResults");
+
+    // Run and compare mightContain result with the expectation.
+    for (int i = 0; i < membershipTestResults.length(); i++) {
+      boolean expectedMembershipResult = membershipTestResults.charAt(i) == '1';
+      boolean mightContain = bloomFilter.mightContain(GOLDEN_DOCUMENT_PREFIX + i);
+      assertEquals(
+          "mightContain() result doesn't match the expectation. File: "
+              + testFile
```
dconeybe, line 171: You can remove the `testFile` from the failure message, since the test file can now be inferred from the test case that failed.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+        new BloomFilter(Base64.getDecoder().decode(bitmap), padding, hashCount);
+
+    String membershipTestResults = resultJSON.getString("membershipTestResults");
+
+    // Run and compare mightContain result with the expectation.
+    for (int i = 0; i < membershipTestResults.length(); i++) {
+      boolean expectedMembershipResult = membershipTestResults.charAt(i) == '1';
+      boolean mightContain = bloomFilter.mightContain(GOLDEN_DOCUMENT_PREFIX + i);
+      assertEquals(
+          "mightContain() result doesn't match the expectation. File: "
```
dconeybe, line 170: Consider this re-wording of the failure message: `mightContain(...) returned X, but expected Y`.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+              + ". Document: "
+              + GOLDEN_DOCUMENT_PREFIX
+              + i,
+          mightContain,
+          expectedMembershipResult);
+    }
+  }
+
+  private JSONObject readJsonFile(String fileName) throws Exception {
+    // Read the file into JSON object.
```
dconeybe, line 181: This comment basically repeats the method's name. Just remove it.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
```diff
@@ -0,0 +1,252 @@
...
+   * bloom filter.
+   *
+   * <p>Full document path is generated by concatenating documentPrefix and number n, eg,
+   * projects/project-1/databases/database-1/documents/coll/doc12.
+   *
+   * <p>The test result is generated by checking the membership of documents from documentPrefix+0
+   * to documentPrefix+2n. The membership results from 0 to n is expected to be true, and the
+   * membership results from n to 2n is expected to be false with some false positive results.
+   */
+  private void runGoldenTest(String testFile) throws Exception {
```
dconeybe, line 150: Mark `runGoldenTest()` and `readJsonFile()` as `static`, if possible.

dconeybe (changes requested):
All of my remaining comments are minor nits. Thanks!

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/BloomFilterTest.java
dconeybe: Please make this change in `constructorShouldThrowNPEOnNullBitmap()`, `constructorShouldThrowIAEOnNegativePadding()`, `constructorShouldThrowIAEOnNegativeHashCount()`, `mightContainWithEmptyStringMightReturnFalsePositiveResult()`, and `bloomFilterToString()` too.

dconeybe (changes requested):
Just one very minor nit, then ready to merge! Feel free to merge without waiting for my final approval since I may be OOO when you address that comment.

dconeybe (approved):

PR #4526, opened 2023-01-05
Focus some synchronized blocks in lifecycle notifier

lfkellogg:
We were including some code inside `synchronized` blocks that did not need the lock.

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
@@ -235,22 +235,27 @@ class FirebaseAppDistributionLifecycleNotifier implements Application.ActivityLi
         return Tasks.forResult(
             getActivityWithIgnoredClass(currentActivity, previousActivity, classToIgnore));
       }
+    }
 
-      TaskCompletionSource<Activity> task = new TaskCompletionSource<>();
+    TaskCompletionSource<Activity> task = new TaskCompletionSource<>();
 
-      addOnActivityResumedListener(
-          new OnActivityResumedListener() {
-            @Override
-            public void onResumed(Activity activity) {
-              synchronized (lock) {
-                task.setResult(
-                    getActivityWithIgnoredClass(activity, previousActivity, classToIgnore));
-              }
-              removeOnActivityResumedListener(this);
-            }
-          });
+    addOnActivityResumedListener(
+        new OnActivityResumedListener() {
+          @Override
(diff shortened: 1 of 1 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -235,22 +235,27 @@ <A extends Activity> Task<Activity> getForegroundActivity(@Nullable Class<A> cla
         return Tasks.forResult(
             getActivityWithIgnoredClass(currentActivity, previousActivity, classToIgnore));
       }
+    }
```
lfkellogg, line 238: `synchronized (lock)` above now ends here, instead of surrounding the entire method.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
```diff
@@ -235,22 +235,27 @@ <A extends Activity> Task<Activity> getForegroundActivity(@Nullable Class<A> cla
...
-              }
-              removeOnActivityResumedListener(this);
-            }
-          });
+    addOnActivityResumedListener(
+        new OnActivityResumedListener() {
+          @Override
+          public void onResumed(Activity activity) {
+            task.setResult(
+                getActivityWithIgnoredClass(activity, getPreviousActivity(), classToIgnore));
```
lfkellogg, line 247: This `setResult` call no longer happens inside a `synchronized (lock)`. Only the access on `previousActivity` does, below.

lfkellogg:
Flaky failing tests will be fixed by: https://github.com/firebase/firebase-android-sdk/pull/4529

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionLifecycleNotifier.java
kaibolay: But why? I don't see why the access to `previousActivity` needs to be protected by the `lock`. It's either `null` or not `null` - with or without the synchronization. So either `null` or not `null` gets passed as the second parameter to `getActivityWithIgnoredClass()` and that's fine, isn't it?

lfkellogg: You're right, but we need to wrap any access of `previousActivity` in a synchronized block because it's annotated by a `@GuardedBy`.

PR #4530, opened 2023-01-05
Resolve javadoc diff failures

emilypgoogle:
This PR resolves two issues relating to the javadoc diff action failing. Both scenarios involved no diff existing as far as I can tell, but it's much cleaner to succeed and no-op. Part one is just making empty build folders to move over if the changed SDKs don't output anything to build when generating javadoc (I didn't realize this was possible originally). The second change however involves letting the final step of posting the comment fail without issue. As of yet I've only seen this fail when provided with an empty file, however this might also mean we miss a breakage (though, the comment consistently disappearing is probably also a reasonable tip off). My original approach involved checking the length of the comment and skipping the step if it was too small, but it seemed hacky and prone to breakage, so I went with this solution instead.

### DIFF
--- a/.github/workflows/diff-javadoc.yml
+++ b/.github/workflows/diff-javadoc.yml
@@ -31,7 +31,7 @@ jobs:
           echo "run=$(cat modules.json | sed "s/[]\"[]//g" | sed "s/,/\n/g" | xargs printf -- "%s:kotlinDoc ")" >> $GITHUB_OUTPUT
 
       - name: Build
-        run: ./gradlew ${{ steps.changed-modules.outputs.run }}
+        run: mkdir build && ./gradlew ${{ steps.changed-modules.outputs.run }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/diff-javadoc.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           echo "run=$(cat modules.json | sed "s/[]\"[]//g" | sed "s/,/\n/g" | xargs printf -- "%s:kotlinDoc ")" >> $GITHUB_OUTPUT
 
       - name: Build
-        run: ./gradlew ${{ steps.changed-modules.outputs.run }}
+        run: mkdir build && ./gradlew ${{ steps.changed-modules.outputs.run }}
```
rlazo, line 34: Add a comment here on why we need to do this

rlazo (approved):

PR #4531, opened 2023-01-05
Always call foreground activity callbacks on the UI thread

lfkellogg:
This will make it easier to reason about what runs on what thread.


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionRegistrar.java
```diff
@@ -60,18 +62,22 @@ public class FirebaseAppDistributionRegistrar implements ComponentRegistrar {
...
+                        lightweightExecutor,
+                        uiThreadExecutor))
             // construct FirebaseAppDistribution instance on startup so we can register for
             // activity lifecycle callbacks before the API is called
             .alwaysEager()
             .build(),
         Component.builder(FeedbackSender.class)
             .add(Dependency.required(FirebaseApp.class))
             .add(Dependency.requiredProvider(FirebaseInstallationsApi.class))
-            .add(Dependency.required(backgroundExecutor))
```
lfkellogg, line 74: I realized this wasn't used in FeedbackSender.

kaibolay (approved):

Issue #4532, opened 2023-01-06
Firestore data is slow to load on app restart


bswhite1:
### [READ] Step 1: Are you in the right place?

Yes

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 2021.3.1 Patch 1
  * Firebase Component: Firestore
  * Component version: 24.4.1 / Master as of 12/22/2022

### [REQUIRED] Step 3: Describe the problem

Firestore data is slow to load on app restart

#### Steps to reproduce:

Do a fresh app install.
Start app and time how long it takes for snapshot listeners to retrieve data.
Since this is a fresh load, it will have to get all data from server, so this should be the longest time measured in this test.

Stop app, restart, and time how long it takes for snapshot listeners to retrieve data.
Since the app should have the data stored in cache, this should be significantly faster.
What I see is that this is as long as if not worse than a fresh load time.

All stops / restarts done repetitively will give roughly the same poor times.

However, if you stop the listeners, then restart them, the following restarts will be faster.

All of the restarts after this will be significantly faster, like 1/4 of the fresh load times.

#### Debugging steps taken:

App restarts seem to be using executeFullCollectionScan rather than performQueryUsingRemoteKeys which hinders performance

Do fresh app install.
start app for first time.
Create query and document listeners
firestore.core.SyncEngine.listen() calls localStore.allocateTarget
target does not exist in cache so it is allocated in the SQLiteTargetCache with a lastLimboFreeSnapshotVersion = (NONE, NONE) then placed into the queryDataByTarget array.
queryDataByTarget.lastLimboFreeSnapshotVersion = (NONE, NONE)

---

eventually we receive the data from the server and firestore.core.SyncEngine.handleRemoteEvent() is called
This will call localStore.applyRemoteEvent() followed by firestore.core.SyncEngine.emitNewSnapsAndNotifyLocalStore() 

localStore.applyRemoteEvent() will set oldTargetData from the queryDataByTarget array.
oldTargetData.lastLimboFreeSnapshotVersion = (NONE, NONE) from above.

newTargetData is from oldTargetData.withResumeToken
newTargetData.lastLimboFreeSnapshotVersion = (NONE, NONE), snapshotVersion = (valid seconds, nanoseconds)

queryDataByTarget is updated but still has lastLimboFreeSnapshotVersion = (NONE, NONE), snapshotVersion = (valid seconds, nanoseconds)
[...]

PR #4533, opened 2023-01-06
updating target cache when there is a local change

bswhite1:
Potential fix for issue 4532

https://github.com/firebase/firebase-android-sdk/issues/4532


firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -606,6 +606,8 @@ public void notifyLocalViewChanges(List<LocalViewChanges> viewChanges) {
               TargetData updatedTargetData =
                   targetData.withLastLimboFreeSnapshotVersion(lastLimboFreeSnapshotVersion);
               queryDataByTarget.put(targetId, updatedTargetData);
+
+              targetCache.updateTargetData(updatedTargetData);
```
wu-hui, line 610: Only if you feel like it, could you try to call `shouldPersistTargetData(targetData, updatedTargetData, ..)` here too. And then add similar logic handling version update, but apply to lastLimboFreeVersion, in `shouldPersistTargetData`?

This way we only persist target data when `lastLimboFreeVersion` advances over a big enough interval, otherwise it is persisted pretty regularly, defeating the purpose of having this optimization in the first place.

Please let me know if you don't want to bother with this, I can take it from here.

wu-hui (commented):
This is an impressive investigation, thanks for the fix!

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
bswhite1: Yeah, I was thinking along the same lines. My original reason to not do that was that shouldPersistTargetData takes in changes of type firestore/remote/TargetChange and the notifyLocalViewChanges has changes of type firestore/local/LocalViewChanges

I can make a new function, but also wasn't sure if we care about the other checks shouldPersistTargetData does, like: resume token, snapshot version times, and changes? Or are all of those ok to still be handled by the original shouldPersistTargetData inside of applyRemoteEvent?

ie Do we only need to do a check on lastLimboFreeSnapshotVersion delta within notifyLocalViewChanges?

bswhite1: @wu-hui any preference or suggestion?

wu-hui: Yes, we only need to check for lastLimboFreeSnapshotVersion within notifyLocalViewChanges and leave everything else unchanged.

You could do this in shouldPersisteTargetData, by adding an extra check for lastLimboFreeSnapshotVersion in that method. And call it from notifyLocalViewChanges.

Alternatively, you could explicitly check for lastLimboFreeSnapshotVersion from within notifyLocalViewChanges, but I think keeping the logic in one place makes slightly more sense here.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
```diff
@@ -606,6 +606,20 @@ public void notifyLocalViewChanges(List<LocalViewChanges> viewChanges) {
               TargetData updatedTargetData =
                   targetData.withLastLimboFreeSnapshotVersion(lastLimboFreeSnapshotVersion);
               queryDataByTarget.put(targetId, updatedTargetData);
+
```
bswhite1, line 609: @wu-hui It seemed weird to create a fake TargetChange to pass into shouldPersistTargetData, so I just did the check here. It does work as expected. If you want to still make the changes to use shouldPersistTargetData, feel free.

bswhite1:
Anyone available to review this?

bswhite1:
@argzdev @wu-hui What do we need to do to get this merged?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/LocalStore.java
wu-hui: Can you pass a null as TargetChange? Then in shouldPersistTargetData, you can do an early return for the limboFreeSnapshotVersion check, then check for null on TargetChange, together with the changes > 0 check.

I'll merge your PR first, then do a followup change myself if you prefer. Thank you!

bswhite1: Hadn't thought of passing null, that could work.
I'll let you merge this and make the follow up changes, I am swamped with other work.
Thanks!

wu-hui (approved):
LGTM! Thank you for your work!!

PR #4534, opened 2023-01-06
Extract APK hash on the background thread

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/NewReleaseFetcher.java
```diff
@@ -62,31 +62,42 @@ public Task<AppDistributionReleaseInternal> checkForNewRelease() {
...
-      return false;
+      return Tasks.forResult(true);
     }
+
+    return isSameAsInstalledRelease(retrievedNewRelease)
+        .onSuccessTask(
+            lightweightExecutor,
+            isSame -> {
+              if (isSame) {
+                LogWrapper.v(TAG, "New release is older or is currently installed");
```
kaibolay, line 96: Is "older" always correct? Could be "newer" as well, right?

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/NewReleaseFetcher.java
lfkellogg: Ah sorry I meant to clarify this because I knew it was misleading. It can be "older" in that if we can return `false` here only if the release has the same app version name, and the same APK hash, and has either an **equal or older build version**. If the build version was newer it would have returned `true` in the condition above.

I've added a comment and renamed some methods to make it clearer what's going on here.

PR #4552, opened 2023-01-12
Extend publishing plugin to support release configFiles.

rlazo:
This enables the plugin to read the list of projects to be released from a configuration file, besides getting them from the CLI.

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
@@ -101,7 +101,7 @@ public class PublishingPlugin implements Plugin<Project> {
             gradle -> {
               List<String> projectsNames;
               if (!publishConfigFilePath.isEmpty()) {
-                projectsNames = readReleaseConfigFile(publishConfigFilePath, project);
+                projectsNames = readReleaseConfigFile(publishConfigFilePath);
               } else {
                 projectsNames =
                     Arrays.asList(projectNamesToPublish.split(projectsToPublishSeparator, -1));
@@ -218,15 +218,15 @@ public class PublishingPlugin implements Plugin<Project> {
             });
   }
 
-  private List<String> readReleaseConfigFile(String publishConfigurationFilePath, Project project) {
+  private List<String> readReleaseConfigFile(String publishConfigurationFilePath) {
     try (Stream<String> stream = Files.lines(Path.of(publishConfigurationFilePath))) {
       return stream
           .dropWhile((line) -> !line.equals("[modules]"))
           .skip(1)
           .collect(Collectors.toList());
     } catch (IOException e) {
-      project.getLogger().error("Error reading publish configuration file. ", e);
-      return List.of();
+      throw new IllegalArgumentException(
+          "Error reading configuration file " + publishConfigurationFilePath, e);
     }
   }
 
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
```diff
@@ -194,6 +218,18 @@ public void apply(Project project) {
             });
   }
 
+  private List<String> readReleaseConfigFile(String publishConfigurationFilePath) {
+    try (Stream<String> stream = Files.lines(Path.of(publishConfigurationFilePath))) {
+      return stream
+          .dropWhile((line) -> !line.equals("[modules]"))
+          .skip(1)
```
emilypgoogle, line 225: What are we skipping here?

rlazo (commented):
Add comment to the stream

emilypgoogle (approved):

buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
```diff
@@ -83,8 +99,16 @@ public void apply(Project project) {
         .getGradle()
         .projectsEvaluated(
             gradle -> {
+              List<String> projectsNames;
+              if (!publishConfigFilePath.isEmpty()) {
+                projectsNames = readReleaseConfigFile(publishConfigFilePath);
+              } else {
+                projectsNames =
+                    Arrays.asList(projectNamesToPublish.split(projectsToPublishSeparator, -1));
```
daymxn, line 107: nit, but you can also do
```kotlin
List.of(projectNamesToPublish.split(projectsToPublishSeparator, -1));
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
```diff
@@ -194,6 +218,19 @@ public void apply(Project project) {
             });
   }
 
+  private List<String> readReleaseConfigFile(String publishConfigurationFilePath) {
+    try (Stream<String> stream = Files.lines(Path.of(publishConfigurationFilePath))) {
+      return stream
+          .dropWhile((line) -> !line.equals("[modules]"))
```
daymxn, line 224: This is a smart way to do it, just drop all elements until you hit modules, skip over the signifier, and return the remaining content as a list.

Good job!

buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
```diff
@@ -41,12 +46,22 @@
...
+ * </pre>
+ *
+ * <pre>
  * ./gradlew -PprojectsToPublish="firebase-inappmessaging,firebase-inappmessaging-display"\
  *           -PpublishMode=(RELEASE|SNAPSHOT) \
  *           firebasePublish
  * </pre>
  *
  * <ul>
+ *   <li>{@code publishConfigFilePath} is the path to the configuration file from which to read the
```
daymxn, line 61: I think it may be worth while adding a little snippet here showcasing the expected format of the config file. I know you already explain it here, but an example could help clarify it. Like so:
```
/**
 *      {@code publishConfigFilePath} is the path to the configuration file from which to read the
 *       list of projects to release. The file format should be consistent with Python's
 *       configparser, and the list of projects should be in a section called "modules". If both
 *       this, and {@code projectsToPublish} are specified, this property takes precedence.
 *       <br>
 *       <br>
 *       Example config file content:
 *       <pre>
 *         [modules]
 *         firebase-database
 *         firebase-common
 *         firebase-firestore
 *       </pre>
 *
 */


```

PR #4553, opened 2023-01-12
Use file-based release configuration for the build release action.

rlazo:
The action was using a dummy configuration that only created release artifacts for Firestore.


rlazo:
Depends on #4552

emilypgoogle (approved):

PR #4554, opened 2023-01-12
Migrate FireEscapeArtifactPlugin to Kotlin

emilypgoogle:
Originally I was going to submit multiple migrations at once, but I've been running into issues with other plugins, so I thought I might as well get at least one up.


buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.publish.PublishingExtension
+import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.api.publish.maven.plugins.MavenPublishPlugin
+import org.gradle.api.tasks.bundling.Jar
+import org.gradle.api.tasks.bundling.Zip
+
+class FireEscapeArtifactPlugin : Plugin<Project> {
+  var project: Project? = null
```
daymxn, line 28: This didn't make much sense to begin with in the Java class tbf. A shared property like this is useful when you lack an init function, or if your init function isn't the caller of your methods. In our case, the init method (apply) handles calling all the methods. So we shouldn't need to define this variable at all, and should instead explicitly pass around the project as method properties where needed. If doing so becomes too verbose- then it signifies a design issue and should be addressed instead.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+
+  override fun apply(target: Project) {
+    project = target
+
+    project!!.afterEvaluate {
+      val firebaseLibrary = extensions.findByType(FirebaseLibraryExtension::class.java)
```
daymxn, line 34: `findByType` returns null if not found, and below I see you circumvent that by making a null check. This would throw an NPE- which is generally not desired. Instead, you should use `getByType`, which will throw an `UnknownDomainObjectException` on the spot- which would be better for debugging and more null safe.

Additionally, *almost* every time a method requires a class argument, if the method was rewritten in Kotlin, it's likely to be `reified`. This means you can pass the type directly instead of as a parameter, like so:
```kotlin
val firebaseLibrary = extensions.getByType<FirebaseLibraryExtension>()
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.gradle.api.tasks.bundling.Jar
+import org.gradle.api.tasks.bundling.Zip
+
+class FireEscapeArtifactPlugin : Plugin<Project> {
+  var project: Project? = null
+
+  override fun apply(target: Project) {
+    project = target
+
+    project!!.afterEvaluate {
```
daymxn, line 33: While I think the `project` variable should be discarded entirely, a little note that `!!` is generally a code smell. Instead, you could've accessed the non null variant of project:
```kotlin
target.afterEvaluate {}
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+      val firebaseLibrary = extensions.findByType(FirebaseLibraryExtension::class.java)
+      val fireEscapeTask =
+        tasks.create("mavenAarFireEscapeArtifact", Zip::class.java) { classifier = "fireescape" }
```
daymxn, line 36: Same as above, `create` is `reified`, and can be called like such:
```kotlin
val fireEscapeTask = tasks.create<Zip>("maven") {
 // ...
}
```

Additionally, `classifier` is deprecated on `Zip`, and you should instead use `archiveClassifier` like so:

```kotlin
archiveClassifier.set("fireescape")
```

So the finished product would be something like:
```kotlin
val fireEscapeTask = tasks.create<Zip>("maven") {
  archiveClassifier.set("fireescape")
}
```

emilypgoogle: Attempting to use the reified syntax you described for create seems to make the compiler think I'm using the `name, actionTask` version of this function, and tells me no type arguments were expected, does something more need to be done?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+      plugins.withType(MavenPublishPlugin::class.java) {
+        extensions.configure(PublishingExtension::class.java) {
+          publications.withType(MavenPublication::class.java) {
+            if ("mavenAar" == name) {
```
daymxn, line 40: Since `mavenAar` will always exist, you can instead shorten this via `getByName`:
```kotlin
publications.getByName<MavenPublication>("mavenAar") {
            
}
```

So with the power of reification:

```kotlin
plugins.withType<MavenPublishPlugin> {
  extensions.configure<PublishingExtension> {
    publications.getByName<MavenPublication>("mavenAar") {
      configurePublication(this, fireEscapeTask, firebaseLibrary.type)
    }
  }
}
```

Or even:

```kotlin
plugins.withType<MavenPublishPlugin> {
  extensions.configure<PublishingExtension> {
    configurePublication(
      publications.getByName<MavenPublication>("mavenAar"),
      fireEscapeTask,
      firebaseLibrary.type
    )
  }
}
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
daymxn: Double check that you're:
A.) Using the correct `Zip` class (as there seems to be two). The correct one is: `org.gradle.api.tasks.bundling.Zip`
B.) Have the DSL variant imported, which happens automatically on my machine but yours may differ. The package is `org.gradle.kotlin.dsl.create`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+
+  private fun proguardMappingFileTask(): Task? {
+    return project?.tasks?.create("fireEscapeProguardMapping") {
+      val task = this
+      project.tasks.all {
+        if (name == "assembleRelease") {
+          task.dependsOn(this)
+        }
+      }
+      outputs.file(File(project.buildDir, "outputs/mapping/release/mapping.txt"))
```
daymxn, line 70: You can use the extension method `fileFromBuildDir` I have defined in `GradleUtils` here:
```kotlin
outputs.file(project.fileFromBuildDir("tmp/javadoc/api.txt"))
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  private fun apiTxtFileTask(): Task? {
+    return project?.tasks?.create("fireEscapeApiText") {
```
daymxn, line 75: Ideally you would `register` the task instead of creating it, as this would differ the creation until it's needed, and create a `TaskProvider` instead of a Task. Additionally, appending such to the method name is the common procedure:
```kotlin
private fun registerApiTxtFileTask(project: Project) =
     project.tasks.register("fireEscapeApiText") {
      dependsOn(JAVADOC_TASK_NAME)
      outputs.file(project.fileFromBuildDir("tmp/javadoc/api.txt"))
    }
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+      outputs.file(File(project.buildDir, "tmp/javadoc/api.txt"))
+    }
+  }
+
+  private fun javadocTask(): Task? {
+    return project?.tasks?.create("fireescapeJavadocJar", Jar::class.java) {
+      dependsOn(JAVADOC_TASK_NAME)
+      from(File(project.buildDir, "/docs/javadoc/reference"))
+      include("**/*")
+      archiveName = "fireescape-javadoc.jar"
```
daymxn, line 86: `archiveName` is also deprecated, and should be replaced by `archiveFileName`:
```kotlin
archiveFileName.set("fireescape-javadoc.jar")
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private fun apiTxtFileTask(): Task? {
+    return project?.tasks?.create("fireEscapeApiText") {
+      dependsOn(JAVADOC_TASK_NAME)
+      outputs.file(File(project.buildDir, "tmp/javadoc/api.txt"))
+    }
+  }
+
+  private fun javadocTask(): Task? {
```
daymxn, line 81: Similar to above, refactoring this to take into account reification and lazy configuration, you could do something like such:
```kotlin
  private fun registerJavadocTask(project: Project) =
    project.tasks.register<Jar>("fireescapeJavadocJar") {
      dependsOn(JAVADOC_TASK_NAME)

      from(project.fileFromBuildDir("/docs/javadoc/reference"))
      include("**/*")

      archiveFileName.set("fireescape-javadoc.jar")
      archiveClassifier.set("javadoc")
    }
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private fun configurePublication(
+    publication: MavenPublication,
+    artifactTask: Zip,
+    libraryType: LibraryType
+  ) {
+    publication.artifact(artifactTask)
+    artifactTask.from(apiTxtFileTask())
+    if (libraryType == LibraryType.ANDROID) {
```
daymxn, line 56: We actually have a util method for this now on `Project`:
```kotlin
if (project.isAndroid()) {

}
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+            if ("mavenAar" == name) {
+              configurePublication(this, fireEscapeTask, firebaseLibrary!!.type)
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private fun configurePublication(
```
daymxn, line 49: While I'm the biggest supporter of functional practices, I feel this method is not really needed. Especially under the context of registering tasks as individuals in a modular way. You could even abstract it all the way to:
```kotlin
override fun apply(target: Project) {
    target.afterEvaluate {
      val apiTxtFile = registerApiTxtFileTask(project)
      val proguardMappingFile = registerProguardMappingFileTask(project)
      val javadoc = registerJavadocTask(project)

      val zippedArtifact = project.tasks.register<Zip>("maven") {
        from(apiTxtFile)
        if (project.isAndroid()) {
          from(proguardMappingFile)
        }
      }

      plugins.withType<MavenPublishPlugin> {
        extensions.configure<PublishingExtension> {
          publications.getByName<MavenPublication>("mavenAar") {
            artifact(zippedArtifact)
            artifact(javadoc)
          }
        }
      }
    }
  }
```

But I also wouldn't be against this either:

```kotlin
  override fun apply(target: Project) {
    target.afterEvaluate {
      if (supportsMavenPublishing(target)) {
        val apiTxtFile = registerApiTxtFileTask(project)
        val proguardMappingFile = registerProguardMappingFileTask(project)
        val javadoc = registerJavadocTask(project)
        val zippedArtifact = registerArtifactTask(project, apiTxtFile, proguardMappingFile)

        configurePublication(project, zippedArtifact, javadoc)
      }
    }
  }

  private fun supportsMavenPublishing(project: Project) =
    project.plugins.hasPlugin(MavenPublishPlugin::class.java)

  private fun configurePublication(
    project: Project,
    zippedArtifact: TaskProvider<Zip>,
    javadoc: TaskProvider<Jar>
  ) {
    project.extensions.configure<PublishingExtension> {
      publications.getByName<MavenPublication>("mavenAar") {
        artifact(zippedArtifact)
        artifact(javadoc)
      }
    }
  }
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,90 @@
...
+    if (libraryType == LibraryType.ANDROID) {
+      artifactTask.from(proguardMappingFileTask())
+    }
+    publication.artifact(javadocTask())
+  }
+
+  private fun proguardMappingFileTask(): Task? {
+    return project?.tasks?.create("fireEscapeProguardMapping") {
+      val task = this
+      project.tasks.all {
```
daymxn, line 65: I would differ this to `apply`, and let the method focus on itself. So that may look like:
```kotlin
  override fun apply(target: Project) {
    target.afterEvaluate {
      if (supportsMavenPublishing(target)) {
        val apiTxtFile = registerApiTxtFileTask(project)
        val proguardMappingFile = registerProguardMappingFileTask(project)
        val javadoc = registerJavadocTask(project)
        val zippedArtifact = registerArtifactTask(project, apiTxtFile, proguardMappingFile)

        project.tasks.findByName("assembleRelease")?.dependsOn(proguardMappingFile)

        configurePublication(project, zippedArtifact, javadoc)
      }
    }
  }

  private fun registerProguardMappingFileTask(project: Project) =
    project.tasks.register("fireEscapeProguardMapping") {
      outputs.file(project.fileFromBuildDir("outputs/mapping/release/mapping.txt"))
    }
```

daymxn (changes requested):
Left some comments going over idiomatic improvements from Java to Kotlin- and just general design stuff. Kotlin is great in that it can be exactly like Java. But it's also great in that it provides much much more in the realm of creating idiomatic and functional systems. But with that, comes a different way of thinking- and it's a learning process. But you're doing great so far, so keep it up!

Also, regarding `build-src-check` failing: the error it's spitting out is a generic "we failed to spin a docker image". We've been having this problem across the board- and (afaik) it has nothing to do with your PR specifically. We believe it's an issue with GH themselves, but it's something worth investigating further. @rlazo has an open bug regarding it as well: [b/264048287 ](https://b.corp.google.com/issues/264048287)

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.gradle.api.tasks.bundling.Zip
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.getByName
+import org.gradle.kotlin.dsl.register
+
+class FireEscapeArtifactPlugin : Plugin<Project> {
+
+  override fun apply(target: Project) {
+    target.afterEvaluate {
+      if (supportsMavenPublishing(target)) {
```
rlazo, line 34: early return to reduce nesting

daymxn: since the function doesn't return anything you'll have to return at the label btw. so something like this:
```kotlin
if (supportsMavenPublishing(target)) return@AfterEvalutate
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FireEscapeArtifactPlugin.kt
```diff
@@ -0,0 +1,82 @@
...
+        extensions.configure<PublishingExtension> {
+          publications.getByName<MavenPublication>("mavenAar") {
+            artifact(zippedArtifact)
+            artifact(javadoc)
+          }
+        }
+      }
+    }
+  }
+  private fun supportsMavenPublishing(project: Project): Boolean {
```
rlazo, line 56: Since these only have a single return statement, you can use `=` instead of `{}`

emilypgoogle: I get compiler warnings about not specifying return type and end up having a more verbose lambda signature if I switch to this syntax, what's the cost/benefit going on with = functions or is it just an aesthetic choice?

rlazo: It's mostly to reduce boilerplate. What warnings are you getting?

daymxn: Curious about this too, because I don't get any warnings. Only if the return type is too ambiguous. Additionally, you can do this as well:
```kotlin
private fun supportsMavenPublishing(project: Project): Boolean =
  project.plugins.hasPlugin(MavenPublishPlugin::class.java)
```

emilypgoogle: I tried again and got no problems, sometimes it's that easy I suppose

rlazo (approved):

daymxn (approved):
LGTM!

PR #4555, opened 2023-01-12
Report updated experiments if either experiment id or variant id changes.

danasilver:
Considering experiments changed if either `experiment id` or `variant id` changes aligns with how the [A/B testing SDK for iOS behaves](https://github.com/firebase/firebase-ios-sdk/blob/eb30b6fdd809cbfe3a9eb7f3ccf1dd46326bef2d/FirebaseABTesting/Sources/ABTConditionalUserPropertyController.m#L261).

### DIFF
--- a/firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
+++ b/firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
@@ -246,7 +246,7 @@ public class FirebaseABTesting {
 
     // Remove experiments no longer assigned.
     List<ConditionalUserProperty> experimentsToRemove =
-            getExperimentsToRemove(experimentsInAnalytics, replacementExperiments);
+        getExperimentsToRemove(experimentsInAnalytics, replacementExperiments);
     removeExperiments(experimentsToRemove);
 
     // Add newly assigned or updated (changed variant id).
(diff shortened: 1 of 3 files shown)


firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
```diff
@@ -204,12 +204,8 @@ public void reportActiveExperiment(AbtExperimentInfo activeExperiment) throws Ab
   public void validateRunningExperiments(List<AbtExperimentInfo> runningExperiments)
       throws AbtException {
     throwAbtExceptionIfAnalyticsIsNull();
-    Set<String> runningExperimentIds = new HashSet<>();
-    for (AbtExperimentInfo runningExperiment : runningExperiments) {
-      runningExperimentIds.add(runningExperiment.getExperimentId());
-    }
     List<ConditionalUserProperty> experimentsToRemove =
-        getExperimentsToRemove(getAllExperimentsInAnalytics(), runningExperimentIds);
+        getExperimentsToRemove(getAllExperiments(), runningExperiments);
```
karenyz, line 208: not added by you, but `getAllExperiments` doesn't really indicate that it's from Analytics. Maybe just add a comment here to reduce confusion for our future selves

karenyz (approved):

firebase-abt/src/main/java/com/google/firebase/abt/FirebaseABTesting.java
danasilver: yeah, the naming is a bit confusing. added some comments for now and maybe will revisit renaming the public method in another change.

karenyz (approved):

PR #4556, opened 2023-01-12
Wrap target backend property in double quotes

rlazo:
Without it, it's generated as a symbol, causing the build to break.


ehsannas (approved):
Oops. Thanks for the fi.

PR #4562, opened 2023-01-13
Add Route header classes

daymxn:
Per [b/265455809](https://b.corp.google.com/issues/265455809),

This adds data class headers for the `FirebaseAuthExchange` backend API routes. Each route follows the proto definition as defined by the API, with nearly identical naming conventions, and proper documentation across the board.

Additionally, this PR refactors the groovy build file to kotlin- to be more align with future plans. `firebase-authexchange-interop`'s  build file is still in groovy, and will be addressed down the line as needed.


firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
...
+ */
+@Serializable internal sealed interface Request
+
+/** Defines what an HTTP response may look like when submitting a [Request]. */
+@Serializable internal sealed interface Response
+
+/**
+ * OIDC Credential.
+ *
+ * Useful for apps with external login already set-up, otherwise known as a headless OIDC flow.
```
rlazo, line 17: If this is an OIDC credential, would it make sense to call the class that? Not sure if implicit credentials is the right naming.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Auth Token that can be used to access certain Firebase Services.
+ *
+ * This is merely the proto type for the network layer, and is completely separate from the user
+ * facing [AuthExchangeToken][com.google.firebase.authexchange.AuthExchangeToken].
+ *
+ * @property accessToken signed JWT containing claims that identify a user
+ * @property ttl the duration from the time this token is minted until its expiration
+ */
+@Serializable internal data class AuthExchangeTokenP(val accessToken: String, val ttl: String)
```
rlazo, line 35: What's the TTL format as a string?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
...
+ * @property idToken JWT encoded OIDC token returned from a third party provider
+ */
+@Serializable internal data class ImplicitCredentialsP(val idToken: String)
+
+/**
+ * Auth Token that can be used to access certain Firebase Services.
+ *
+ * This is merely the proto type for the network layer, and is completely separate from the user
+ * facing [AuthExchangeToken][com.google.firebase.authexchange.AuthExchangeToken].
+ *
```
rlazo, line 31: what do you mean by a proto type in this case?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
...
+ * @property accessToken signed JWT containing claims that identify a user
+ * @property ttl the duration from the time this token is minted until its expiration
+ */
+@Serializable internal data class AuthExchangeTokenP(val accessToken: String, val ttl: String)
+
+/**
+ * Request header for the `/ExchangeInstallationAuthToken` endpoint.
```
rlazo, line 38: Is this a header or the actual request?

here and below

firebase-authexchange/firebase-authexchange.gradle.kts
```diff
@@ -0,0 +1,53 @@
+plugins {
```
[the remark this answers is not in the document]
daymxn: Good catch! I had a script I wrote that would automatically add it to any missing files, but it doesn't auto-run. We should probably have some sort of check regarding that. Not anything that's enforced, since there are way too many exceptions, but maybe like a git hook or gha that looks at files added and lets you know if any of interest are missing copyright notices (`.kt`,`.java`,etc.,).

firebase-authexchange/firebase-authexchange.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+    }
+    getByName("test") {
+      java.srcDirs("src/test/kotlin")
+    }
+  }
+
+  testOptions.unitTests.isIncludeAndroidResources = true
+}
+
+dependencies {
```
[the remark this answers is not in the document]
daymxn: I never realized vlad added this, that's perfect.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
+package com.google.firebase.authexchange.network
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Defines what an HTTP request may look like when interacting with the backend API for
```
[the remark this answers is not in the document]
daymxn: > You can be more direct in the doc:
> 
> `A request to the backend`
> 
> and then, below
> 
> `A response from the backend`

Honestly, I would pitch otherwise. While I think the goal (especially on the first line) should be to be as direct as possible- being too brief and assuming a common understanding often-times leads to a misunderstanding. I've personally ran into it before, where there's been some documentation that assumed a common understanding that I lacked- because my process of learning more was starting at the code and working up :')
Curious to hear your thoughts though

> I'm curious why are you defining them as sealed interfaces.

Because they can all be declared at compile time (people won't be extending them outside this file). Additionally, it comes with the benefit of additional compiler optimizations, and exhausted casting. So we can do stuff like this:

```kotlin
when(request) {
 ExchangeInstallationAuthTokenRequestP -> {}
 ExchangeCustomTokenRequestP -> {}
 ExchangeOidcTokenRequestP -> {}
}
```

And the compiler recognizes that this is all the possible combinations, so there's not need for an `else` statement that just throws an error. There's a couple other benefits that've gotta do with type casting- but those are what come to mind initially.

> Also, why internal.

Because you can't hide packages like in java with `package-info`. You would have to manually exclude each file. But more importantly, it lets folks know if they should ever try to accidentally use one of these classes- that it's intended for internal use only. That is, it's much easier for developers to see that they should use `com.google.firebase.authexchange.AuthExchangeToken` instead of `com.google.firebase.authexchange.network.routes.AuthExchangeToken`, because one is explicitly marked as `internal` wheres the other one is just public.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
...
+ * OIDC Credential.
+ *
+ * Useful for apps with external login already set-up, otherwise known as a headless OIDC flow.
+ *
+ * While this isn't the only OIDC flow that the API supports, it's the only flow that the mobile SDK
+ * supports- so there is no base class for credentials.
+ *
+ * @property idToken JWT encoded OIDC token returned from a third party provider
+ */
+@Serializable internal data class ImplicitCredentialsP(val idToken: String)
```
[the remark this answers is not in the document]
daymxn: Protobuf, implying that these data classes are representations of gRPC related data (that is, they're raw classes that shouldn't be used directly, and are intended to only be used by the network layer). I'm not quite sure where I picked this up from, but I've used this naming convention by habit in all of my personal gRPC projects. Helps more easily difference between the network and application layer.

Although if you're against it, I'm fine with removing it:)

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,103 @@
...
+/** Defines what an HTTP response may look like when submitting a [Request]. */
+@Serializable internal sealed interface Response
+
+/**
+ * OIDC Credential.
+ *
+ * Useful for apps with external login already set-up, otherwise known as a headless OIDC flow.
+ *
+ * While this isn't the only OIDC flow that the API supports, it's the only flow that the mobile SDK
+ * supports- so there is no base class for credentials.
```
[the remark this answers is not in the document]
daymxn: SGTM

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: I would be comfortable with that, although the property name would still be `implicitCredentials` in the API call, because that's what the API expects.

I can't link it here since it's an internal link- but if you look at the bug linked in this PR (referenced in the OP), and go to the first link in the comments of that bug, you can see all the proto defs for the API. These definitions and their names are derived from that- since these are all just data classes for the API- intended to be encoded to JSON via `@Serializable`

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: A long with an "s" appended to the end of it.

It's a `Duration` from the (our) protobuf package, defined [here](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/duration.proto).

Since this is translated directly from the JSON, I kept it as a string- as that's how protobuf exports it. You can read the header on `Duration` to learn more about that.

Technically I could write a custom deserializer for translating it to a `Duration` class, or something of the like, but I opted not to just cause that's how proto exports it, and all we do is translate it into a long for the application layer. Although I'm not against it. Thoughts?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: As in it's a data class used for communicating via protobuf APIs (the backend for auth echange). Although I do think the phrasing gives off the wrong impression. Thoughts on this instead:
```
 * Not to be confused with [AuthExchangeToken][com.google.firebase.authexchange.AuthExchangeToken],
 * this is intended only for network requests.
 ```
 Or maybe just removing it entirely?

File: firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt (hunk 4 above)
Comment (daymxn) [reply firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt:38]: So nothing here actually implements the requests themselves. These are all just headers, or type defs, of what the API expects. Skeletons you could say. Actual implementation and utilization of these classes will be elsewhere in the SDK.

The API is currently defined in protobuf in G3, and all these classes provide a type-safe kotlin variant of the API as it is defined. Additionally, we can use kotlin's serialization, and immediately encode classes to JSON requests for the backend, and vise versa for JSON responses from the backend. 

Just a little quick example (psuedo code):
```kotlin
data class GetPersonRequestP(val name: String)
data class GetPersonResponseP(val person: Person?, val error: String?)

data class Person(val name: String, age: Int, phoneNumber: String) 

fun getPerson() {
  val myQuery = GetPersonRequestP("Rodrigo")

  val response: GetPersonResponseP = googleServer:GetRequest<GetPersonRequestP>(myQuery)

  response?.person?.let {
    println("Rodrigo is ${it.age} years old!")
  }
}
```

All subclasses of `Request` are definitions for what individual API endpoints expect in their request call.

All subclasses of `Response` are definitions for what API endpoints return in their response to request calls.

Anything else is just extra data types that either need to be arguments in a call to a certain API, or are apart of a response.

Does that answer your question? Or did I misunderstand what you were asking?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: Actually, I forgot we can specify a `@SerialName` for encoding and decoding, so we would call it something else but it gets translated to `implicitCredentials` when it's sent. But seeing as this is ONLY for the API, and shouldn't be used beyond that- I think it's totally fine keeping it as is. Thoughts?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rlazo: All below are either `Request` or `Response` subclasses, and they are named as such, but the doc says "headers" which is different. A header is metadata about the request, but these classes contain the actual content of the requests/responses.

firebase-authexchange/firebase-authexchange.gradle.kts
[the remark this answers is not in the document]
daymxn: LOL, well I'll be paying more attention to that one for sure

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: > All below are either `Request` or `Response` subclasses, and they are named as such, but the doc says "headers" which is different. A header is metadata about the request, but these classes contain the actual content of the requests/responses.

They contain the content sure, but only in the sense that it's a data class and doesn't have any content. I think C headers- where data types and function signatures are declared in header files, and the source file implements said methods.

Although, what do you think would be a better name? We could also always just split it up into two separate files; `request.kt` and `response.kt`.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rlazo: Thanks for the detailed explanation. No need to do any custom deserializer unless absolutely needed.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rosalyntan: +1 to "request header" implying network request headers, for example the `x-firebase-appcheck` header.

Though I'm not sure I understand what you mean when you say the actual implementation of these classes will exist elsewhere -- from my understanding of a Kotlin data class, it is essentially like a struct, which is fully instantiable (i.e. it's not like a Java interface where you need to implement the interface before it can be used). Also, are these data classes not implementations of the `Request` interface already?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,105 @@
+package com.google.firebase.authexchange.network
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Defines what an HTTP request may look like when interacting with the backend API for
+ * [FirebaseAuthExchange][com.google.firebase.authexchange.FirebaseAuthExchange].
+ */
+@Serializable internal sealed interface Request
```
rosalyntan, line 10: This is probably just my lack of Kotlin knowledge -- but what is the purpose of having an empty interface here?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @property token relative resource name of the audience project and location
+ * @property providerId the display name or id of the OIDC provider
+ * @property implicitCredentials JWT token from the OIDC provider, provided by the developer
+ */
+@Serializable
+internal data class ExchangeOidcTokenRequestP(
+  val token: String,
+  val providerId: String,
+  val implicitCredentials: ImplicitCredentialsP
```
rosalyntan, line 78: The `ExchangeOidcTokenRequest` can take either implicit credentials or auth code credentials (see the [API  proto](http://google3/google/firebase/authexchange/v1alpha/token_exchange.proto;l=142;rcl=500590766)). Is there a way to represent this as a `oneof`?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
```diff
@@ -0,0 +1,105 @@
...
+ * Request header for the `/ExchangeInstallationAuthToken` endpoint.
+ *
+ * @see ExchangeTokenResponseP
+ *
+ * @property token relative resource name of the audience project and location
+ * @property installationAuthToken the installation token issued to the app
+ */
+@Serializable
+internal data class ExchangeInstallationAuthTokenRequestP(
+  val token: String,
```
rosalyntan, line 49: The `token` field is unnecessary as part of the request object (here and in all the requests). In the backend, the `token` field is extracted from the request URL.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: Just to provide a parent class, and the `sealed` type. Because it's a sealed type, I could theoretically do something like such:
```Kotlin
<T: Request> fun handleRequest(request: T) =
  when (request) {
    is ExchangeOidcTokenRequest -> // ...
    is ExchangeCustomTokenRequest -> // ...
    is ExchangeInstallationTokenRequest -> // ...
  } // the compiler understands that these are all the possible subclasses of Request
```
Although it's generally just good functional design to have a parent class for common types.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: For sure, although do we use it on our end? I initially had the following comment:
```kotlin
 * While this isn't the only OIDC flow that the API supports, it's the only flow that the mobile SDK
 * supports- so there is no base class for credentials.
 ```
Because I was under the impression we don't utilize `AuthCodeCredentials` on the SDK?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
daymxn: Ahh I see. Is it the `key` param, or is it actually called `token`?

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rosalyntan: Gotcha, thanks for the explanation!

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rosalyntan: I see -- in terms of naming clarity, I think you can just remove the term "header" from the documentation and leave the rest of the comment the same.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rosalyntan: What do you mean by `key` param? For additional context -- here is the [request path](http://google3/google/firebase/authexchange/v1alpha/token_exchange_service.proto;l=35;rcl=492327771). Note the `token=` portion of the request path -- that is where the `token` field is populated from.

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rosalyntan: The auth code OIDC flow is the "headful" flow (detailed [here](https://docs.google.com/document/d/1qpekuA5yyImI8MCAsNmt3HIMa6uHzlGuqmEtLgViAVI/edit?resourcekey=0-8g6zKHNr6Ii_RsxQ3j6GTw#)).

firebase-authexchange/src/main/kotlin/com/google/firebase/authexchange/network/Routes.kt
rlazo: +1

rlazo:
friendly ping

PR #4565, opened 2023-01-17
Update make_release_notes to generate based on the release config

rlazo:
Beyond that change, we are also making some extra format changes to better suit our usecase.

### DIFF
--- a/docs/make_release_notes.py
+++ b/docs/make_release_notes.py
@@ -244,6 +244,8 @@ def main():
         products = releasing_products(args.releasecfg)
 
     for product in products:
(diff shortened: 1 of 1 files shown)


emilypgoogle (approved):
I'm definitely not a python expert but the idea here makes sense. I did notice near the end we repeat the same regex `^#{1,6} ` 3 times, might be a good idea to move it out to a constant? That was just part of a formatting change by you though, and not an actual logic change.

rlazo:
> 

I saw that, but they are all part of a larger regexp, so I don't think it makes sense to leave it separate for now.

PR #4574, opened 2023-01-19
Add expected count to target

milaGGL:
Ported from https://github.com/firebase/firebase-js-sdk/pull/6854


firebase-firestore/src/main/java/com/google/firebase/firestore/local/TargetData.java
```diff
@@ -96,7 +105,21 @@ public TargetData withResumeToken(ByteString resumeToken, SnapshotVersion snapsh
         purpose,
         snapshotVersion,
         lastLimboFreeSnapshotVersion,
-        resumeToken);
+        resumeToken,
+        expectedCount);
```
dconeybe, line 109: `withResumeToken(...)` should specify a null `expectedCount` in the newly-created `TargetData` object.

https://github.com/firebase/firebase-js-sdk/blob/dd6683583b7c6b07586c2551f4d362b48fb83f7c/packages/firestore/src/local/target_data.ts#L108

firebase-firestore/src/main/java/com/google/firebase/firestore/local/TargetData.java
```diff
@@ -158,7 +187,8 @@ public boolean equals(Object o) {
         && purpose.equals(targetData.purpose)
         && snapshotVersion.equals(targetData.snapshotVersion)
         && lastLimboFreeSnapshotVersion.equals(targetData.lastLimboFreeSnapshotVersion)
-        && resumeToken.equals(targetData.resumeToken);
+        && resumeToken.equals(targetData.resumeToken)
+        && expectedCount == targetData.expectedCount;
```
dconeybe, line 191: This `==` comparison here is not correct. If both `expectedCount` and `targetData.expectedCount` are distinct `Integer` objects but have the same underlying integer value then they will compare unequal using the `==` operator. We want objects to compare equal in this case.

Instead, use `Objects.equals(...)` to compare the objects for equality. It will correctly handle null values and use `.equals()` if they are both non-null.

https://developer.android.com/reference/java/util/Objects?hl=en#equals(java.lang.Object,%20java.lang.Object)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/TargetData.java
```diff
@@ -170,6 +200,7 @@ public int hashCode() {
     result = 31 * result + snapshotVersion.hashCode();
     result = 31 * result + lastLimboFreeSnapshotVersion.hashCode();
     result = 31 * result + resumeToken.hashCode();
+    result = 31 * result + (expectedCount != null ? expectedCount.hashCode() : 0);
```
dconeybe, line 203: To correctly handle `null` when calculating the hash code, you can use `Objects.hashCode()`. Please do that here because it is a bit more readable.

https://developer.android.com/reference/java/util/Objects?hl=en#hashCode(java.lang.Object)

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java
```diff
@@ -499,6 +499,10 @@ public Target encodeTarget(TargetData targetData) {
       builder.setResumeToken(targetData.getResumeToken());
     }
 
+    if (targetData.getExpectedCount() != null) {
+      builder.setExpectedCount(Int32Value.newBuilder().setValue(targetData.getExpectedCount()));
```
dconeybe, line 503: This code block is inconsistent with the web sdk. In the web sdk, it only sets expectedCount if a resume token or snapshot version is specified. Here, it is setting expectedCount regardless of whether or not a resume token or snapshot version is specified.

We generally aim to keep the sdks in sync as much as possible because it's confusing when things are done differently in one sdk compared to another.

Please make the two SDKs do the same thing in this scenario.

I think the way that the web sdk does it is best because it makes it very clear that expectedCount is only meaningful if a resume token or snapshot version is used.

https://github.com/firebase/firebase-js-sdk/blob/dd6683583b7c6b07586c2551f4d362b48fb83f7c/packages/firestore/src/remote/serializer.ts#L1013-L1025

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java
```diff
@@ -369,7 +369,14 @@ public void listen(TargetData targetData) {
 
   private void sendWatchRequest(TargetData targetData) {
     watchChangeAggregator.recordPendingTargetRequest(targetData.getTargetId());
-    watchStream.watchQuery(targetData);
+    if (!targetData.getResumeToken().isEmpty()
+        || targetData.getSnapshotVersion().compareTo(SnapshotVersion.NONE) > 0) {
+      int expectedCount = this.getRemoteKeysForTarget(targetData.getTargetId()).size();
+      TargetData newTargetData = targetData.withExpectedCount(expectedCount);
```
dconeybe, line 375: Please make this consistent with the web sdk as well. Namely, replace the `targetData` local variable rather than creating the `newTargetData` local variable. Also, move the call to `watchStream.watchQuery(targetData)` to _outside_ of the "if" block.

https://github.com/firebase/firebase-js-sdk/blob/dd6683583b7c6b07586c2551f4d362b48fb83f7c/packages/firestore/src/remote/remote_store.ts#L338-L346

firebase-firestore/src/test/resources/json/listen_spec_test.json
```diff
@@ -99,6 +99,7 @@
         "watchEntity": {
           "docs": [
             {
+              "createTime": 0,
```
dconeybe, line 102: I know this is annoying, but can you remove all of the `createTime` lines from this file, since they are not part of the bloom filter tests?

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -1006,6 +1006,9 @@ private void validateExpectedState(@Nullable JSONObject expectedState) throws JS
                   targetData.withResumeToken(
                       ByteString.EMPTY, version(queryDataJson.getInt("readTime")));
             }
+            if (queryDataJson.has("expectedCount")) {
```
dconeybe, line 1009: Wow, is this the only change you had to make to the spec test runner? Can you just confirm that if you modify the spec tests so that they should fail that they do actually fail? I'm sure you tried this, but I'm just surprised that updating the spec tests was so easy!

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java
milaGGL: yeah, this is the part that I am a bit hesitated. In web the if statments is like: 
```
if (targetData.resumeToken.approximateByteSize() > 0) {
  } else if (targetData.snapshotVersion.compareTo(SnapshotVersion.min()) > 0) {
```
while in Android it is 
```
if (targetData.getResumeToken().isEmpty()
        && targetData.getSnapshotVersion().compareTo(SnapshotVersion.NONE) > 0) {
} else{
}
```

milaGGL: updated the if statement to indicate resume token or snapshotVersion is required to add expectedCount: 
```
   if (targetData.getExpectedCount() != null
        && (!targetData.getResumeToken().isEmpty()
            || targetData.getSnapshotVersion().compareTo(SnapshotVersion.NONE) > 0)) {
      builder.setExpectedCount(Int32Value.newBuilder().setValue(targetData.getExpectedCount()));
    }
```

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
milaGGL: Yes, in WEB we updated the spec_test_builder and spec test file as well. But in Android, we just need to wait for WEB to "build spec test"->"serialize into json file" -> "import into android" -> "deserialize and run spec test".

so only need to update spec test runner.

milaGGL: this Spec is only to add the expectedCount. In the next PR, for applying bloom filter, we need to do much more than this.

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteSerializer.java
dconeybe: Even though the "if" statement has the extra check for `targetData.getResumeToken().isEmpty()`, we should still mirror the structure from the web sdk. We only want to set expected count if either the "if" or the "else" branch above is entered, so we may as well put the logic into those branches.

firebase-firestore/src/test/resources/json/listen_spec_test.json
dconeybe: I'm not sure what happened, but there are still a bunch of `"createTime": 0` lines in this file. Could you remove them too?

dconeybe (changes requested):

dconeybe (approved):
LGTM!

PR #4579, opened 2023-01-19
Replace Android version with SDK version in the Realtime stream request.

danasilver:
Fixes b/257069716. Populate using `BuildConfig` as we do in the [Fetch request](https://github.com/firebase/firebase-android-sdk/blob/994c4b60e60f00b21a2357f7c488d512b1d506ae/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHttpClient.java#L344).


qdpham13 (approved):
LGTM

PR #4582, opened 2023-01-20
Port spect test changes from web SDK

milaGGL:
Port spec test changes from Web SDK master branch. 
Spec test changes is made in :https://github.com/firebase/firebase-js-sdk/pull/6781


ehsannas:
As discussed offline, let's decouple this from the `createTime` work.

PR #4598, opened 2023-01-26
Style gallery button for uploading custom screenshots

tagboola:
fad-sdk-upload-custom-screenshot

### DIFF
--- a/firebase-appdistribution/src/main/res/drawable/btn_disabled_stroked.xml
+++ b/firebase-appdistribution/src/main/res/drawable/btn_disabled_stroked.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
+    <solid android:color="#FFFFFF" />
+    <corners android:radius="8dp"/>
+    <stroke android:width="1dp" android:color="#0F000000" />
+</shape>
--- a/firebase-appdistribution/src/main/res/drawable/btn_disabled_stroked.xml
+++ b/firebase-appdistribution/src/main/res/drawable/btn_disabled_stroked.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
-    <solid android:color="#FFFFFF" />
+    <solid android:color="#0D000000" />
     <corners android:radius="8dp"/>
     <stroke android:width="1dp" android:color="#0F000000" />
 </shape>
--- a/firebase-appdistribution/src/main/res/drawable/btn_stroked.xml
+++ b/firebase-appdistribution/src/main/res/drawable/btn_stroked.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
-    <solid android:color="#FFFFFF"/>
+    <solid android:color="#00000000"/>
     <corners android:radius="8dp"/>
     <stroke android:width="1dp" android:color="#1A000000" />
 </shape>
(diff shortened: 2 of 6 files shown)


firebase-appdistribution/src/main/res/layout/activity_feedback.xml
```diff
@@ -100,25 +100,35 @@
                 android:background="@drawable/card_background"
                 android:orientation="vertical">
 
-                <CheckBox
-                    android:id="@+id/screenshotCheckBox"
-                    android:layout_width="wrap_content"
+                <RelativeLayout
```
kaibolay, line 103: Since there are only two child elements and their relative positioning is pretty straightforward, wouldn't it make sense to use a simpler option for this container, e.g. `LinearLayout` (horizontal)?

tagboola: I actually tried a `LinearLayout` before, but it was clear to me how I could get the checkbox to align to the left and the button to align to the right. With `RelativeLayout` you get the nice `android:layout_*` attributes which makes it easier.

lfkellogg: `RelativeLayout` looks good to me. Alternatively if you wanted to make it work with `LinearLayout` you might be able to add an empty `View` with a weight of 1 in between the two, as a spacer: https://developer.android.com/develop/ui/views/layout/linear#Weight

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
```diff
@@ -100,25 +100,35 @@
...
+                    android:layout_width="match_parent"
                     android:layout_height="wrap_content"
-                    android:text="@string/screenshot" />
+                    android:layout_marginTop="5dp"
+                    android:layout_marginBottom="5dp">
```
lfkellogg, line 107: Curious what Tars thinks, but this looks a little cramped to me. I'd probably increase the bottom margin a bit here, maybe to `10dp`?

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
tagboola: `LinearLayout` with a spacer seems jankier, so I'd lean towards keeping it as a `RelativeLayout`.

firebase-appdistribution/src/main/res/drawable/btn_disabled_stroked.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
+    <solid android:color="#FFFFFF" />
```
lfkellogg, line 3: Is it possible to make the background transparent? It looks a little weird since the parent container background isn't the same white, and the mocks show it as the same color as its parent.

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
kaibolay: Or maybe you can use `gravity` instead of an empty `View`: 

> A [LinearLayout](https://developer.android.com/reference/android/widget/LinearLayout) respects [...] the gravity (right, center, or left alignment) of each child.

firebase-appdistribution/src/main/res/drawable/btn_stroked.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle">
+    <solid android:color="#FFFFFF"/>
```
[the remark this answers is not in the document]
tagboola: Updated-- Made this black with 5% opacity for the disabled state.

firebase-appdistribution/src/main/res/drawable/btn_disabled_stroked.xml
tagboola: Done.

lfkellogg (approved):
Looks awesome!

firebase-appdistribution/src/main/res/layout/activity_feedback.xml
tagboola: Tried this, and it wasn't working. I'm probably not using `gravity` correctly here, but regardless I think the `RelativeLayout` is sufficient here and I like the explicitness of the `android:layout_*` attributes.
Screenshot 2023-01-26 at 2 04 57 PM

lfkellogg:
By the way, you can ignore the Smoke Tests failure. I think our last rebase from `master` pulled over some change that breaks them. This happens sometimes.

PR #4602, opened 2023-01-27
Realtime RC Bug Patch: Fixes deadlock issue that caused app to freeze when transitioning from background to foreground

qdpham13:
**Problem:**  Bug that occurs when app goes from foreground -> background -> foreground causing the app to freezes due to a deadlock. This happens b/c beginRealtimeHttpStream is synchronized and blocking, when the app goes from background -> foreground it tries to restart the stream and calls makeRealtimeHttpConnection. Since beginRealtimeHttpStream is blocking, the call to makeRealtimeHttpConnection is blocked until the connection is finished after 5 min 
due to the synchronization from the first method.

**Solution:** Make beginRealtimeHttpStream not synchronized by moving the "GuardedBy" variable to other, synchronized methods.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -66,7 +66,7 @@ public class ConfigRealtimeHttpClient {
   private final Set<ConfigUpdateListener> listeners;
 
   @GuardedBy("this")
-  private HttpURLConnection httpURLConnection;
+  private boolean isHttpConnectionRunning;
 
   @GuardedBy("this")
   private int httpRetriesRemaining;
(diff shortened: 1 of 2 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -277,7 +277,11 @@ public synchronized void retryHttpConnection() {
   }
 
   private synchronized void makeRealtimeHttpConnection(long retrySeconds) {
-    if (canMakeHttpStreamConnection() && httpRetriesRemaining > 0) {
+    if (!canMakeHttpStreamConnection()) {
+      return;
```
danasilver, line 281: Seems like we should return an error if there's a call to make the connection and it can't be made?

danasilver (approved):
Could you make the PR title a bit more descriptive for future reference

PR #4605, opened 2023-01-28
Make a deep copy of the other config before modifying it to compute changes.

danasilver:
`ConfigContainer.copyOf` does not make a true "copy". The `other` `ConfigContainer` should not be modified between fetching and activating, since this produces an inaccurate activated config.


qdpham13 (approved):
LGTM

PR #4610, opened 2023-01-31
Debug smoke tests

yifanyang:


daymxn (approved):
LGTM- you're the goat for looking into this, great job! The `firebase-storage` unit test failure is a flake, which I was investigating today- but needs further inspection. I haven't looked at the implementation tests though, so double check that they're just flakes since we're bumping the gradle version (I doubt it'll cause any issues, but just to be safe).

yifanyang:
> LGTM- you're the goat for looking into this, great job! The `firebase-storage` unit test failure is a flake, which I was investigating today- but needs further inspection. I haven't looked at the implementation tests though, so double check that they're just flakes since we're bumping the gradle version (I doubt it'll cause any issues, but just to be safe).

@daymxn Thanks! The unit test for `firebase-storage` passed after a few reruns. The instrumentation tests for `firebase-firestore` and `firebase-storage` are still broken. The failures look like  real issues to me, not just flakes. They started to emerge in #4424, probably not related to this pull request.

PR #4622, opened 2023-02-03
Revert "Migrate vendor plugin to new transform api. (#4424)"

yifanyang:
This reverts commit ffc5d33e693909e3ff3b45479a180c32db17c063.


daymxn (approved):
Any idea what specifically caused it? Because I'd like to keep us on AGP `7.4.X` so long as it doesn't cause any issues.

yifanyang:
> Any idea what specifically caused it? Because I'd like to keep us on AGP `7.4.X` so long as it doesn't cause any issues.

No idea. Will roll back for now and investigate later ([b/268091395](http://b/268091395)).

PR #4623, opened 2023-02-03
Implement the GitSubmodulePlugin

daymxn:
Per [b/267779128](https://b.corp.google.com/issues/267779128),

This adds a new Plugin called `GitSubmodulePlugin`. It's a light weight plugin that facilitates interaction with Git Submodules from Gradle. 

Our current release action is broken ([b/267188389](https://b.corp.google.com/issues/267188389)), because we have SDKs with submodules- but never actually pull the submodules as apart of our build process. We _could_ get away with just a quick little hacky task to fix this, but I opted for a Plugin instead. Primarily to take advantage of the following benefits:
 - Better build times through native Gradle plugin improvements
 - More modular implementation, should any other SDKs start using Git Submodules
 - Further establishment of plugin design principles in Gradle/Kotlin
 - Easier to maintain and document

Documentation is provided as annotations in KDoc form for everything added, including the tasks themselves. Namely, the plugin adds three tasks:

`initializeGitSubmodules`
`updateGitSubmodules`
`removeGitSubmodules`

Each task is annotated with background information on what it does.

_Note that this plugin also binds the updating of submodules to the build process, so this immediately fixes our release pipeline without further action._

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/bomgenerator/tagging/ShellExecutor.java
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/bomgenerator/tagging/ShellExecutor.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.function.Consumer;
 import org.gradle.api.GradleException;
 
+// TODO(b/267668143): With modernization efforts
 public class ShellExecutor {
   private final Runtime runtime;
   private final File cwd;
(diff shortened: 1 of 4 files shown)


vkryachko:
For the release job, isn't it easier to just add the following to our workflow config?

```yaml
with:
  submodules: true
```

emilypgoogle (approved):

davidmotson (approved):

daymxn:
> For the release job, isn't it easier to just add the following to our workflow config?
> 
> ```yaml
> with:
>   submodules: true
> ```

Good question. Yes and no. While it would quickly and easily fix the release job, that's only in the context of the workflow. The underlying task itself would still be broken. So if you were to be debugging this locally, it wouldn't align with what you would get from the workflow. Additionally, this may come up in several places in the future regarding submodules not being initialized for a build. I believe it'd be much easier long-term, and more robust- to instead fix it at the build level versus the workflow level. WDYT?

yifanyang (commented):
Do we vision that going forward product teams shall manipulate submodules only through gradle?

- If so, do we need to support other submodule commands? Not sure if all existing submodule use scenarios are covered.
- If not, can product teams run standalone git submodule commands? Will they be interfered with implicit invocations from gradle?

PR #4628, opened 2023-02-06
Revert "Make FetchType header capitalized to match backend"

qdpham13:
Reverts firebase/firebase-android-sdk#4620

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigFetchHandler.java
@@ -693,8 +693,8 @@ public class ConfigFetchHandler {
   }
 
   public enum FetchType {
-    BASE("BASE"),
-    REALTIME("REALTIME");
+    BASE("Base"),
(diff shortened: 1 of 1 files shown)


danasilver:
What's the "correct" state for the header? We can migrate/fix it in the SDKs while making the backend adaptable to both

qdpham13:
> What's the "correct" state for the header? We can migrate/fix it in the SDKs while making the backend adaptable to both

The backend has the header values as all caps but the change that I'm going to make in the backend will make the header case-insensitive. So it's up to us if we want to update the header value in the SDKs to be all caps as well or just leave it as is, it won't affect the behavior once the backend change is submitted

danasilver:
Got it. It makes sense to me the backend and frontend align for simplicity, even if the backend is resilient to any casing. IIUC that means the client sends it uppercased

qdpham13:
> Got it. It makes sense to me the backend and frontend align for simplicity, even if the backend is resilient to any casing. IIUC that means the client sends it uppercased

Yeah that makes sense. Going to go ahead and close this PR that reverts the capitalization change

PR #4631, opened 2023-02-07
Pass feedback trigger type to create feedback call

tagboola:
Pipe through feedback trigger to API client calls and pass it as a header in call to create feedback

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
@@ -78,7 +78,7 @@ public class FeedbackActivity extends AppCompatActivity {
     if (savedInstanceState != null) {
       releaseName = savedInstanceState.getString(RELEASE_NAME_KEY);
       infoText = savedInstanceState.getCharSequence(INFO_TEXT_KEY);
-      feedbackTrigger = FeedbackTrigger.valueOf(savedInstanceState.getString(FEEDBACK_TRIGGER_KEY));
+      feedbackTrigger = FeedbackTrigger.fromString(savedInstanceState.getString(FEEDBACK_TRIGGER_KEY));
       String screenshotUriString = savedInstanceState.getString(SCREENSHOT_URI_KEY);
       if (screenshotUriString != null) {
         screenshotUri = Uri.parse(screenshotUriString);
@@ -86,6 +86,7 @@ public class FeedbackActivity extends AppCompatActivity {
     } else {
       releaseName = getIntent().getStringExtra(RELEASE_NAME_KEY);
       infoText = getIntent().getCharSequenceExtra(INFO_TEXT_KEY);
+      feedbackTrigger = FeedbackTrigger.fromString(getIntent().getStringExtra(FEEDBACK_TRIGGER_KEY));
       if (getIntent().hasExtra(SCREENSHOT_URI_KEY)) {
         screenshotUri = Uri.parse(getIntent().getStringExtra(SCREENSHOT_URI_KEY));
       }
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
@@ -78,7 +78,8 @@ public class FeedbackActivity extends AppCompatActivity {
     if (savedInstanceState != null) {
       releaseName = savedInstanceState.getString(RELEASE_NAME_KEY);
       infoText = savedInstanceState.getCharSequence(INFO_TEXT_KEY);
-      feedbackTrigger = FeedbackTrigger.fromString(savedInstanceState.getString(FEEDBACK_TRIGGER_KEY));
+      feedbackTrigger =
+          FeedbackTrigger.fromString(savedInstanceState.getString(FEEDBACK_TRIGGER_KEY));
       String screenshotUriString = savedInstanceState.getString(SCREENSHOT_URI_KEY);
       if (screenshotUriString != null) {
         screenshotUri = Uri.parse(screenshotUriString);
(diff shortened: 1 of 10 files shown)


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TakeScreenshotAndStartFeedbackActivity.java
```diff
@@ -38,7 +38,13 @@ public void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     infoText = getIntent().getCharSequenceExtra(INFO_TEXT_EXTRA_KEY);
     LogWrapper.i(TAG, "Capturing screenshot and starting feedback");
-    firebaseAppDistribution.startFeedback(infoText);
+    if (firebaseAppDistribution instanceof FirebaseAppDistributionImpl) {
+      ((FirebaseAppDistributionImpl) firebaseAppDistribution)
```
tagboola, line 42: This is gross, but I don't want to introduce a public method in the interface that takes in the feedback trigger. Any other suggestions?

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackActivity.java
```diff
@@ -75,6 +78,7 @@ protected void onCreate(Bundle savedInstanceState) {
     if (savedInstanceState != null) {
       releaseName = savedInstanceState.getString(RELEASE_NAME_KEY);
       infoText = savedInstanceState.getCharSequence(INFO_TEXT_KEY);
+      feedbackTrigger = FeedbackTrigger.valueOf(savedInstanceState.getString(FEEDBACK_TRIGGER_KEY));
```
kaibolay, line 81: I think this might not work as intended. I think the default `valueOf()` implementation is based on the `name()` of the `Enum`, not the customized `toString()` that you implemented. So you could implement a `fromString()` and use that instead of `valueOf()` here - or use `name()` instead of `toString()` when serializing the value for the `Bundle`.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TakeScreenshotAndStartFeedbackActivity.java
kaibolay: You could inject `FirebaseAppDistributionImpl` instead of `FirebaseAppDistribution`

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
```diff
@@ -437,14 +450,18 @@ private void doStartFeedback(CharSequence infoText, @Nullable Uri screenshotUri)
...
+      @Nullable Uri screenshotUri,
+      FeedbackTrigger trigger) {
     return lifecycleNotifier.consumeForegroundActivity(
         activity -> {
           LogWrapper.i(TAG, "Launching feedback activity");
           Intent intent = new Intent(activity, FeedbackActivity.class);
           // in development-mode the releaseName might be null
           intent.putExtra(FeedbackActivity.RELEASE_NAME_KEY, releaseName);
           intent.putExtra(FeedbackActivity.INFO_TEXT_KEY, infoText);
+          intent.putExtra(FeedbackActivity.FEEDBACK_TRIGGER_KEY, trigger.toString());
```
kaibolay, line 464: You probably should use `name()` instead of `toString()` here if you keep `valueOf()` for deserialization.

Also, maybe there's a simple way to test that this parameter is passed (i.e. serialized and deserialized) correctly.

kaibolay (changes requested):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
tagboola: Created a `fromString()` method for deserialization. It would be nice to test this, but we don't have any tests for the Activity where the deserialization is performed. I did test this locally and it seems to work. I'm gonna time-box an investigation into adding an test for FeedbackActivity.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TakeScreenshotAndStartFeedbackActivity.java
kaibolay: No, I don't think so: The code here is the (real) impl. The API (with its default no-op impl) stands on its own.

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TakeScreenshotAndStartFeedbackActivity.java
tagboola: Done.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FirebaseAppDistributionImpl.java
tagboola: Activity testing is hard. Going to punt on this for now but may revisit it this week.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/TesterApiHttpClient.java
```diff
@@ -104,6 +105,26 @@ JSONObject makeGetRequest(String tag, String path, String token)
...
+  }
+
+  /**
+   * Make a POST request to the tester API at the given path using a FIS token for auth with
+   * feedback trigger type.
+   *
+   * @return the response body
+   */
+  public JSONObject makePostRequest(
+      String tag, String path, String token, String requestBody, FeedbackTrigger trigger)
```
lfkellogg, line 118: Instead of adding this method, I'd prefer to make `makePostRequest(String tag, String path, String token, String requestBody, Map<String, String> extraHeaders)` below package private, and then call that directly from `FirebaseAppDistributionTesterApiClient#createFeedback()`. That way this this class stays agnostic to feedback-specific logic.

lfkellogg (approved):

kaibolay (approved):

PR #4632, opened 2023-02-07
Pass trigger header on commit feedback call

tagboola:
Note to self: confirm which endpoint is actually recording the metric before changing the client 🤦🏿‍♂️

### DIFF
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
@@ -47,7 +47,7 @@ class FeedbackSender {
             lightweightExecutor, feedbackName -> attachScreenshot(feedbackName, screenshotUri))
         .onSuccessTask(
             lightweightExecutor,
-            (feedbackName) -> testerApiClient.commitFeedback(feedbackName, trigger));
+            feedbackName -> testerApiClient.commitFeedback(feedbackName, trigger));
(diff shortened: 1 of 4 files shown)


lfkellogg (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
```diff
@@ -42,10 +42,12 @@ Task<Void> sendFeedback(
...
       FeedbackTrigger trigger) {
     return testerApiClient
-        .createFeedback(releaseName, feedbackText, trigger)
+        .createFeedback(releaseName, feedbackText)
         .onSuccessTask(
             lightweightExecutor, feedbackName -> attachScreenshot(feedbackName, screenshotUri))
-        .onSuccessTask(lightweightExecutor, testerApiClient::commitFeedback);
+        .onSuccessTask(
+            lightweightExecutor,
+            (feedbackName) -> testerApiClient.commitFeedback(feedbackName, trigger));
```
kaibolay, line 50: nit: (no) (need) (for) (parenthesis)

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
tagboola: Done.

PR #4634, opened 2023-02-07
Fix firebase-storage UploadTest flakes

daymxn:
Per b/268217935,

This fixes the failing `UploadTest` test in `firebase-storage`. It seems the issue originated from a couple different places:
1.) `ControllableSchedulerHelper` wasn't correctly synchronizing when validating threads
2.) Previous test cases that have no use now were present
3.) Since `addOnFailureListener` calls don't originate from the same thread, it seems the `Assert` calls were throwing hidden exceptions

I've removed the message check, since it's sort of redundant. But if we chose to add it again, it should be added outside the scope of the failure listener.

This bug was particularly difficult to debug, as it only seemed to fail on CI. My assumption is that it has to do with how resources and threads are handled in the image we have on CI- but I couldn't find anything concrete.


yifanyang (approved):
Thanks for looking into it!

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
```diff
@@ -128,56 +128,38 @@ public void fileUploadWith500() throws Exception {
...
-  //      Assert.fail();
-  //    } catch (IllegalStateException ignore) {
-  //      // Task is not yet done.
-  //    }
-  //
-  //    Assert.assertNull(task.getException());
-  //
-  //    task.addOnFailureListener(
-  //        (exception) -> {
-  //          Assert.assertEquals(
```
maneesht, line 154: I think it's worth adding back the message check, as the exception thrown could be something else entirely, and this test would erroneously pass

maneesht (changes requested):

firebase-storage/src/test/java/com/google/firebase/storage/UploadTest.java
daymxn: Ahh I was under the impression this was the only likely exception to be thrown. SGTM, will add an alternative check at the end- that's slightly less strict to be safe.

maneesht (approved):
LGTM, thanks!

PR #4685, opened 2023-02-15
Reduce default timeout on main thread by 1 second to avoid ANRs.

mrober:


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/Utils.java
```diff
@@ -72,31 +72,25 @@ public static <T> Task<T> race(Executor executor, Task<T> t1, Task<T> t2) {
 
   /** Similar to Tasks.call, but takes a Callable that returns a Task. */
   public static <T> Task<T> callTask(Executor executor, Callable<Task<T>> callable) {
-    final TaskCompletionSource<T> tcs = new TaskCompletionSource<T>();
+    final TaskCompletionSource<T> tcs = new TaskCompletionSource<>();
```
themiswang, line 75: New to java, why we don't need type template here anymore?

mrober: Java can do type inference of generics like this since Java 7. The way it was written before is probably left over from Java 6. I cleaned it up real quick because it was close to my other changes.

Issue #4214, opened 2022-10-17
 Attempt to invoke virtual method 'com.google.firebase.inappmessaging.model.MessageType kb.i.c()'


gonojuarez:
### [READ] Step 1: Are you in the right place?

Issues filed here should be about bugs in __the code in this repository__. 
If you have a general question, need help debugging, or fall into some
other category use one of these other channels:

  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 2021.2.1 Patch 2
  * Firebase Component: inAppMessaging
  * Component version: 20.2.0

### [REQUIRED] Step 3: Describe the problem
Firebase crash.

#### Steps to reproduce:

message.getMessageType is null

#### Relevant Code:
```
// branch code
private List<Action> extractActions(InAppMessage message) {
 List<Action> actions = new ArrayList<>();
 switch (message.getMessageType()) {
  case BANNER:
   actions.add(((BannerMessage) message).getAction());
  break;
  case CARD:
    actions.add(((CardMessage) message).getPrimaryAction());
    actions.add(((CardMessage) message).getSecondaryAction());
   break;
  case IMAGE_ONLY:
    actions.add(((ImageOnlyMessage) message).getAction());
  break;
  case MODAL: 
    actions.add(((ModalMessage) message).getAction());
  break;
  default:
   // An empty action is treated like a dismiss
     actions.add(Action.builder().build());
  }
   return actions;
}

// Log
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'com.google.firebase.inappmessaging.model.MessageType kb.i.c()' on a null object reference
       at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.extractActions(FirebaseInAppMessagingDisplay.java:437)
       at com.google.firebase.inappmessaging.display.FirebaseInAppMessagingDisplay.inflateBinding(FirebaseInAppMessagingDisplay.java:308)
[...]

PR #4691, opened 2023-02-16
Fiam crash fix

eldhosembabu:
Crash fix: [Attempt to invoke virtual method 'com.google.firebase.inappmessaging.model.MessageType kb.i.c()'](https://github.com/firebase/firebase-android-sdk/issues/4214#top)

https://github.com/firebase/firebase-android-sdk/issues/4214

- Added a null check to avoid the crash.

### DIFF
--- a/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
+++ b/firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
@@ -292,8 +292,9 @@ public class FirebaseInAppMessagingDisplay extends FirebaseInAppMessagingDisplay
   // it is safe to ignore this warning
   @SuppressLint("ClickableViewAccessibility")
   private void inflateBinding(final Activity activity, final BindingWrapper bindingWrapper) {
-    if(inAppMessage==null){
-      // inAppMessage could get null because of raise conditions from prior inapp message timers. So adding a safety null check. See b/269599986
+    if (inAppMessage == null) {
+      // inAppMessage could get null because of raise conditions from prior inapp message timers. So
+      // adding a safety null check. See b/269599986
       return;
     }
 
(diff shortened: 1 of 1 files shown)


firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -292,6 +292,12 @@ public void run() {
   // it is safe to ignore this warning
   @SuppressLint("ClickableViewAccessibility")
   private void inflateBinding(final Activity activity, final BindingWrapper bindingWrapper) {
+    if (inAppMessage == null) {
+      // inAppMessage could get null because of raise conditions from prior inapp message timers. So
```
rlazo, line 296:
```suggestion
      // inAppMessage could get null because of race conditions from prior inapp message timers. So
```

firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -292,6 +292,12 @@ public void run() {
   // it is safe to ignore this warning
   @SuppressLint("ClickableViewAccessibility")
   private void inflateBinding(final Activity activity, final BindingWrapper bindingWrapper) {
+    if (inAppMessage == null) {
```
rlazo, line 295: Just want to be sure that we are not masking any deeper bug with this check. Is it OK to do nothing in these cases? what would be the user experience in this case? e.g. a message could be lost, several messages, etc.

eldhosembabu: I've already have my analysis mentioned here : https://github.com/firebase/firebase-android-sdk/issues/4214#issuecomment-1433779590

So if the race condition happens, we will not be presenting the inappmessage during that event and will be shown when then next same event happens. If we get significant reports on issues related with this, we will look towards a long term solution.

rlazo (approved):

PR #4692, opened 2023-02-16
Make OR APIs public

ehsannas:
And re-enable tests.

The tests that require a composite index were separated out and marked such that they'll only be executed against the emulator. In a couple of other cases, I made minor tweaks such that they would still serve the purpose of the test, but wouldn't require a composite index.


MarkDuckworth (approved):
LGTM

PR #4694, opened 2023-02-17
Update heartbeat to have the most recent useragent for a given day

VinayGuthal:


firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
```diff
@@ -148,7 +157,18 @@ synchronized void storeHeartBeat(long millis, String userAgentString) {
...
+      String storedUserAgentString = getStoredUserAgentString(dateString);
+      if (storedUserAgentString == null) {
+        // Heartbeat already sent for today.
+        return;
+      }
+      if (storedUserAgentString.equals(userAgentString)) {
+        // UserAgent not updated.
+        return;
+      } else {
+        updateStoredUserAgent(userAgentString, dateString);
```
rlazo, line 169: how often can this happen? I'm thinking about the additional cost of reading (line 160), and writing(169), the user agent string.

VinayGuthal: i think the updateStoredUserAgent doesn't happen as much. But the other case when the user agent remains the same might happen a lot which is why i think line 160 is necessary

chkuang-g: My understanding is this is called when `initializeApp()` is called and when the app is resumed from the background.
Does not seem to be too frequent.

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorageTest.java
```diff
@@ -75,6 +75,30 @@ public void storeOneHeartbeat_storesProperly() {
...
+            .isEqualTo(new ArrayList<String>(Collections.singleton("1970-01-01")));
+    heartBeatInfoStorage.deleteAllHeartBeats();
+    heartBeatInfoStorage.storeHeartBeat(100, "test-agent-2");
+    heartBeatInfoStorage.storeHeartBeat(1000, "test-agent-1");
+    assertThat(heartBeatInfoStorage.getHeartBeatCount()).isEqualTo(0);
+    results = (ArrayList<HeartBeatResult>) heartBeatInfoStorage.getAllHeartBeats();
+    assertThat(results.size()).isEqualTo(0);
+    assertThat(heartBeatInfoStorage.getHeartBeatCount()).isEqualTo(0);
+    results = (ArrayList<HeartBeatResult>) heartBeatInfoStorage.getAllHeartBeats();
+    assertThat(results.size()).isEqualTo(0);
```
rlazo, line 99: Aren't these 3 lines the same as the previous 3?

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorageTest.java
```diff
@@ -75,6 +75,30 @@ public void storeOneHeartbeat_storesProperly() {
     assertThat(results.size()).isEqualTo(0);
   }
 
+  @Config(sdk = 29)
+  @Test
+  public void storeOneHeartbeat_updatesProperly() {
```
rlazo, line 80: can we tests with several dates? new user agent in a different date doesn't override the old user agent

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorageTest.java
```diff
@@ -75,6 +75,30 @@ public void storeOneHeartbeat_storesProperly() {
...
+    assertThat(heartBeatInfoStorage.getHeartBeatCount()).isEqualTo(1);
+    heartBeatInfoStorage.storeHeartBeat(10, "test-agent-1");
+    ArrayList<HeartBeatResult> results =
+            (ArrayList<HeartBeatResult>) heartBeatInfoStorage.getAllHeartBeats();
+    assertThat(results.size()).isEqualTo(1);
+    assertThat(results.get(0).getUserAgent()).isEqualTo("test-agent-1");
+    assertThat(results.get(0).getUsedDates())
+            .isEqualTo(new ArrayList<String>(Collections.singleton("1970-01-01")));
+    heartBeatInfoStorage.deleteAllHeartBeats();
+    heartBeatInfoStorage.storeHeartBeat(100, "test-agent-2");
```
rlazo, line 92: Add a comment to explain why we expect this to be empty

firebase-common/src/main/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorage.java
rlazo: Reading `SharedPreferences` is `synchronized` so it can get expense, but maybe since we are already in a `synchronized` block is not too much worse? what impact do you think this change can have?

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorageTest.java
VinayGuthal: Yes sure.. I can do that

firebase-common/src/test/java/com/google/firebase/heartbeatinfo/HeartBeatInfoStorageTest.java
VinayGuthal: yes seems copy paste issue.

chkuang-g (commented):
LGTM, logical-wise.

I will leave this to @rlazo for final approval

rlazo (approved):

PR #4695, opened 2023-02-17
Return Server produced error if Http status code is 403

qdpham13:
https://buganizer.corp.google.com/issues/268398841
Pass back server error if it is the 403 API enablement response.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -492,7 +492,7 @@ public class ConfigRealtimeHttpClient {
       } else {
         String errorMessage =
             String.format(
-                "Unable to connect to the server. Try again in a few minutes. HTTP Status code: %d",
+                "Unable to connect to the server. Try again in a few minutes. HTTP status code: %d",
                 responseCode);
         if (responseCode == 403) {
           errorMessage = parseErrorResponseMessage(httpURLConnection.getErrorStream());
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -420,7 +420,6 @@ public class ConfigRealtimeHttpClient {
       String message = "";
       while ((message = bufferedReader.readLine()) != null) {
         response.append(message);
-        Log.i(TAG, message);
       }
     } catch (IOException ex) {
       // Unable to parse error message.
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -412,7 +412,7 @@ public class ConfigRealtimeHttpClient {
         || statusCode == HttpURLConnection.HTTP_GATEWAY_TIMEOUT;
   }
 
-  private String parseErrorResponseMessage(InputStream inputStream) {
+  private String parseForbiddenErrorResponseMessage(InputStream inputStream) {
     StringBuilder response = new StringBuilder();
 
     try {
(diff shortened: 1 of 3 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -470,12 +489,17 @@ public void beginRealtimeHttpStream() {
       if (connectionFailed || responseCode == HttpURLConnection.HTTP_OK) {
         retryHttpConnectionWhenBackoffEnds();
       } else {
+        String errorMessage =
+            String.format(
+                "Unable to connect to the server. Try again in a few minutes. Http Status code: %d",
+                responseCode);
+        if (responseCode == 403) {
```
danasilver, line 496: What if we get a 403 without a message we expect? We should still return some error message in the exception.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -470,12 +489,17 @@ public void beginRealtimeHttpStream() {
       if (connectionFailed || responseCode == HttpURLConnection.HTTP_OK) {
         retryHttpConnectionWhenBackoffEnds();
       } else {
+        String errorMessage =
+            String.format(
+                "Unable to connect to the server. Try again in a few minutes. Http Status code: %d",
```
danasilver, line 494: Nit: NTTP is capitalized

danasilver:
Any tests we can add for this?

firebase-config/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigTest.java
```diff
@@ -1332,6 +1332,23 @@ public void realtime_successfulStatusCode_doesNotIncreaseConfigMetadataBackoffDa
...
+        .when(configRealtimeHttpClientSpy)
+        .closeRealtimeHttpStream(any(HttpURLConnection.class));
+    when(mockHttpURLConnection.getErrorStream())
+        .thenReturn(new ByteArrayInputStream("{ \"code\": 403 }".getBytes(StandardCharsets.UTF_8)));
+    when(mockHttpURLConnection.getResponseCode()).thenReturn(403);
+    configRealtimeHttpClientSpy.beginRealtimeHttpStream();
+
+    verify(configRealtimeHttpClientSpy, never()).startAutoFetch(any());
+    verify(configRealtimeHttpClientSpy, never()).retryHttpConnectionWhenBackoffEnds();
+    verify(mockStreamErrorEventListener).onError(any(FirebaseRemoteConfigServerException.class));
```
danasilver, line 1349: Maybe add the message we expect from the backend and verify it's parsed correctly?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -423,6 +423,7 @@ private String parseErrorResponseMessage(InputStream inputStream) {
       }
     } catch (IOException ex) {
       // Unable to parse error message.
+      return "Unable to connect to the server, access is forbidden. Http Status code: 403";
```
danasilver, line 426: Same "HTTP" capitalization here. Would also call this method `parseForbiddenResponseMessage` if it's 403-specific and add a comment why we're trying to parse it for future reference

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -412,18 +412,21 @@ private boolean isStatusCodeRetryable(int statusCode) {
...
-  private String parseErrorResponseMessage(InputStream inputStream) {
+  private String parseForbiddenErrorResponseMessage(InputStream inputStream) {
     StringBuilder response = new StringBuilder();
 
     try {
       BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(inputStream));
       String message = "";
       while ((message = bufferedReader.readLine()) != null) {
         response.append(message);
+        Log.i(TAG, message);
```
danasilver, line 423: Extra log?

danasilver (approved):

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: Good catch, thanks!

PR #4696, opened 2023-02-17
Add integration test for bloom filter

milaGGL:
ported from https://github.com/firebase/firebase-js-sdk/pull/7045

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -1035,7 +1035,6 @@ public class QueryTest {
   @Test
   public void resumingQueryShouldRemoveDeletedDocumentsIndicatedByExistenceFilter()
       throws InterruptedException {
-    // TODO(Mila):Remove this condition once the bug is resolved.
     assumeFalse(
         "Skip this test when running against the Firestore emulator as there is a bug related to "
             + "sending existence filter in response: b/270731363.",
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -14,6 +14,7 @@
 
 package com.google.firebase.firestore;
 
+import static com.google.firebase.firestore.testutil.IntegrationTestUtil.isRunningAgainstEmulator;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.nullList;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.querySnapshotToIds;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.querySnapshotToValues;
@@ -21,16 +22,15 @@ import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCol
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollectionWithDocs;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
-import static com.google.firebase.firestore.testutil.IntegrationTestUtil.writeAllDocs;
 import static com.google.firebase.firestore.testutil.TestUtil.expectError;
 import static com.google.firebase.firestore.testutil.TestUtil.map;
-import static com.google.firebase.firestore.util.Util.autoId;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeFalse;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1029,6 +1031,46 @@ public void testMultipleUpdatesWhileOffline() {
...
+  @Test
+  public void resumingQueryShouldRemoveDeletedDocumentsIndicatedByExistenceFilter() {
+    Map<String, Map<String, Object>> testDocs = new LinkedHashMap<>();
+    for (int i = 1; i <= 100; i++) {
+      testDocs.put("doc" + i, map("key", i));
+    }
+
+    // Setup firestore with disabled persistence and populate a collection with testDocs.
+    FirebaseFirestore firestore = testFirestore();
+    firestore.setFirestoreSettings(
```
dconeybe, line 1043: I don't think that disabling persistence is necessary. It doesn't matter if the cache is populated with the documents that were written by `writeAllDocs()`. Or am I missing something?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1029,6 +1031,46 @@ public void testMultipleUpdatesWhileOffline() {
...
+        new FirebaseFirestoreSettings.Builder().setPersistenceEnabled(false).build());
+    CollectionReference collection = firestore.collection(autoId());
+    writeAllDocs(collection, testDocs);
+
+    QuerySnapshot snapshot1 = waitFor(collection.get());
```
dconeybe, line 1048: Add a comment above this line that the query below will populate the cache and save the resume token.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1029,6 +1031,46 @@ public void testMultipleUpdatesWhileOffline() {
...
+              for (int i = 1; i <= 50; i++) {
+                DocumentReference docRef = collection.document("doc" + i);
+                transaction.delete(docRef);
+              }
+              return null;
+            }));
+
+    // Wait 10 seconds, during which Watch will stop tracking the query
+    // and will send an existence filter rather than "delete" events.
+    try {
```
dconeybe, line 1064: Remove the try/catch and just add "throws Exception" to this method's signature.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1029,6 +1031,46 @@ public void testMultipleUpdatesWhileOffline() {
...
+
+    QuerySnapshot snapshot1 = waitFor(collection.get());
+    assertEquals(snapshot1.size(), 100);
+
+    // Delete 50 docs in transaction so that it doesn't affect local cache.
+    waitFor(
+        firestore.runTransaction(
+            transaction -> {
+              for (int i = 1; i <= 50; i++) {
+                DocumentReference docRef = collection.document("doc" + i);
```
dconeybe, line 1056: Get the names of the documents to delete from `testDocs` rather than recalculating them here.

This (a) helps with readability because it shows that you're deleting documents that you just created and (b) avoids code duplication of generating the document names twice.

dconeybe (changes requested):

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
milaGGL: Without disabling persistence, it is always returning 100 docs after deleting the 50 docs. 
And in web SDK, it is the same result and we skipped the test if persistence is enabled as well.

dconeybe: Something doesn't seem right then. If persistence is disabled then the resume token will be garbage collected as soon as the first call to `collection.get()` completes. Then, later on, when it calls `collection.get()` the 2nd time, the client will not specify a resume token in the request since there isn't one in persistence. For this test to be valid, we _need_ persistence enabled in order to have a resume token specified in the 2nd request, which triggers the bloom filter logic.

I took a closer look at the corresponding PR in the web sdk, https://github.com/firebase/firebase-js-sdk/pull/7045, and now I remember that I forgot to ask you about this back when I reviewed it. I believe the test in the web sdk is invalid as well: that integration test _only_ makes sense when persistence is _enabled_.

That's really weird that the 2nd call to `collection.get()` would return the original 100 documents. This needs to be investigated. Or maybe my brain isn't functioning properly and you can let me know!

dconeybe (changes requested):

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
milaGGL: Discussed offline.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1029,6 +1032,47 @@ public void testMultipleUpdatesWhileOffline() {
     assertEquals(asList(map("foo", "zzyzx", "bar", "2")), querySnapshotToValues(snapshot2));
   }
 
+  @Test
+  public void resumingQueryShouldRemoveDeletedDocumentsIndicatedByExistenceFilter()
+      throws InterruptedException {
+    // TODO(Mila):Remove this condition once the bug is resolved.
```
dconeybe, line 1038: You can remove this comment: "TODO(Mila):Remove this condition once the bug is resolved" because it is implied by the assumeFalse() line below.

dconeybe (changes requested):
Just one very minor change requested. Otherwise, LGTM.

PR #4710, opened 2023-02-22
Remove redundant testUtil sourceset from `firebase-database`

jmwski:
Right now a bunch of tests in `firebase-database` generate unbound reference warnings in intellij even though those same tests compile and run successfully. All the unbound references are for symbols in `firebase-database/src/testUtils/*.java`. I noticed that we had two gradle `sourceSets` that point to this location. Removing the `test` sourceset clears the intellij warnings.

### DIFF
--- a/firebase-database/CHANGELOG.md
+++ b/firebase-database/CHANGELOG.md
@@ -3,6 +3,7 @@
 # 20.1.0
 * [unchanged] Updated to accommodate the release of the updated
 [database] Kotlin extensions library.
+* [fixed] Remove redundant gradle sourceSet to clear spurious Android Studio unbound reference errors.
 
 
 ## Kotlin


firebase-database/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # 20.1.0
 * [unchanged] Updated to accommodate the release of the updated
 [database] Kotlin extensions library.
+* [fixed] Remove redundant gradle sourceSet to clear spurious Android Studio unbound reference errors.
```
maneesht, line 6: Not sure if we need a changelog here, as this isn't a user-facing change.

rlazo: Yes, no need of changelog entries for non-user visible changes.

rlazo:
CI unit tests are broken

https://github.com/firebase/firebase-android-sdk/actions/runs/4247517279/jobs/7385688642

PR #4713, opened 2023-02-23
Initial commit for Firebase Sessions SDK

mrober:
Initial commit for Firebase Sessions SDK

- Setup subproject, including test structure with a place holder api
- Source code and gradle build files are all in Kotlin
- Set min sdk level to 16, which is what Crashlytics and Fireperf are

### DIFF
--- a/firebase-sessions/firebase-sessions.gradle.kts
+++ b/firebase-sessions/firebase-sessions.gradle.kts
@@ -40,6 +40,7 @@ android {
 }
 
 dependencies {
+  implementation(project(":firebase-common"))
   implementation(project(":firebase-common:ktx"))
   implementation(project(":firebase-components"))
   implementation(libs.androidx.annotation)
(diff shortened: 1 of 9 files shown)


firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+  }
+  kotlinOptions { jvmTarget = "1.8" }
+  testOptions.unitTests.isIncludeAndroidResources = true
+}
+
+dependencies {
+  implementation(project(":firebase-common"))
+  implementation(project(":firebase-common:ktx"))
```
davidmotson, line 44: I think you only need a dependency on firebase-common:ktx here. The other ones appear to get pulled in transitively.

mrober: Removed

davidmotson (approved):

PR #4714, opened 2023-02-23
Update DackkaPlugin to use 1.2.0

daymxn:
Per [b/262894853](https://b.corp.google.com/issues/262894853),

This upgrades the our Dackka version from `1.0.3` to `1.2.0` and changes our `DackkaPlugin` to accommodate such. With this, comes additional configuration that Dackka has gracefully exposed for us (alongside a plethora of bug fixes):

- Changing reference paths between Kotlin and Java ([b/243674305](https://b.corp.google.com/issues/243674305))
- Changing reference head tag paths between Kotlin and Java ([b/248316730](https://b.corp.google.com/issues/248316730))
- Removing the google group ID ([b/257293594](https://b.corp.google.com/issues/257293594)) (although we aren't able to take advantage of this, as it's currently a global setting- see [b/270593375](https://b.corp.google.com/issues/270593375))

As such, this PR does the following:

- Removes our transform for reference paths
- Removes our transform for reference head tags paths
- Updates the `FiresiteTransform` documentation
- Removes the `seperateDackkaFiles` task, as it is replaced by Dackka exposing configuration to set output paths between Kotlin and Java
- Removes the `copyJavaDocToCommonDirectory` and `copyKotlinDocToCommonDirectory` tasks, and combines them into a single `copyDocsToCommonDirectory` (thanks to the removal of the `seperateDackkaFiles` task)
- Builds Javadocs in Dackka (thanks to [b/243324828](https://b.corp.google.com/issues/243324828))
- Removes the needing of doc stubs
- Avoids building Javadocs for `-ktx` libraries
- Cleans up the `DackkaPlugin` code entirely to be more accessible to folks unfamiliar with Kotlin
- Updates the `DackkaPlugin` documentation
- Provides additional documentation for DackkaPlugin methods
- Removes the need for `DEVSITE_TENANT` in our Dackka generation (and is set via the new `fqPluginName` Dackka configuration instead)
- Updates `DackkaGenerationTask` to accommodate the changes from `1.0.3` to `1.2.0`
- Updates the `DackkaGenerationTask` documentation
- Cleans up the `DackkaGenerationTask` code entirely to be more accessible to folks unfamiliar with Kotlin
- Adds a note to `FirebaseLibraryExtension` about refactoring it into a modernized version that respects newer gradle conventions
- Refactors some util methods to be more Kotlin idiomatic
- Adds additional documentation to some util methods
- Adds a util method to create `Attribute` types, and set their value; in a more Kotlin friendly way
- Adds a util method to submit WorkQueues in a more Kotlin friendly way
- Adds extension properties to `Project` types that map to `FirebaseLibraryExtension`, and identify if the module is a `-ktx` variant
- Removes the old `firesite_transform.sh` script
- Fixes spelling errors in various Javadocs
- Hides artifacts from doc generation that should've previously been hidden
- Fixes various Javadocs to conform to existing Dackka bugs
- Reduces some verbose links to their short-hand variant
- Adds a `package-list` file for the `AndroidX` package

This upgrade was particularly large, because it involved (finally) fully migrating our Javadoc generation out of Metalava and into Dackka. Any future upgrades to Dackka versions should be very small in comparison.

> Note that the changes that occur from doc generation in Dackka instead of Metalava are already approved and merged into Firesite.


firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateUpdateHandler.java
```diff
@@ -31,6 +31,7 @@ public abstract class AppStateUpdateHandler implements AppStateCallback {
   // It must be a weak reference because unregisterForAppState() is called typically from
   // Trace.stop() and user may forget to call Trace.stop(), if it was a strong reference,
   // the registration in AppStateMonitor will prevent Trace to be deallocated.
+  /** @hide */
```
rlazo, line 34: Why is a `private final` var need the hide annotation? I think we can document it.

firebase-annotations/src/main/java/com/google/firebase/annotations/concurrent/Background.java
```diff
@@ -24,6 +24,8 @@
  *
  * <p>For operations that can block for long periods of time, like network requests, use the {@link
  * Blocking} executor.
+ *
+ * @hide
```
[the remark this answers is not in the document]
daymxn: This one was actually a mistake, fixed.

firebase-perf/src/main/java/com/google/firebase/perf/application/AppStateUpdateHandler.java
daymxn: Dackka currently leaks `private final` variables with public getters. And if you look at the public getter for this- it's hidden as well. This isn't supposed to be in the public docs, it's only `@VisibleForTesting`.

See [b/241259955](https://b.corp.google.com/issues/241259955) for context.

daymxn: If you feel strongly about it, I'm down for it. But I do weigh more towards not doing it.

If you take a look at [this commit](https://github.com/firebase/firebase-android-sdk/pull/4714/commits/f4ea4f130ecaa8ef2420e812972549cdee28c651), there's a considerable amount of changes that originated from a single bug ([go/dokka-upstream-bug/2665](http://go/dokka-upstream-bug/2665)).

Would you also suggest we add a `TODO()` at the site of each one? I think leaving as is should be fine- as it doesn't cause any issues. Should someone come across it and change it, if we haven't updated Dackka to a version that fixes it then they'll see the issue as soon as they try to submit a PR and it notes the Javadoc changes. Otherwise, it won't cause any harm.

Thoughts?

rlazo: FWIW the changes in that commit, while necessary to make dackka work, do not jump as weird the same way as annotating a private field `@hide`. That being said, I don't feel strongly about it, so feel free to leave it as is.

rlazo (approved):

PR #4722, opened 2023-02-28
Add task for Version bumping

daymxn:
Per [b/270991015](https://b.corp.google.com/issues/270991015),

In the first of many efforts to modernize and clean-up our post release tasks, this adds the `VersionBumpTask`, which can be utilized to bump the [locally] current version of the given project. It also exposes configuration inputs as such that the task can be isolated from our post-release clean-up tasks, should we ever desire such. And of course, it provides sensible defaults that align with our intentions for it.

Additionally, this adds the `ModuleVersion` data class- a type safe variant of standard SemVer versions. This gets utilized directly in the `VersionBumpTask`, and will likely be used in future tasks as well (as needed).

And finally, this adds a bare-bones `PostReleasePlugin` that registers the `VersionBumpTask`, and provides background information on why the task exists.


buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -0,0 +1,77 @@
...
+     * ModuleVersion.fromString("a.b.c") // null
+     * ```
+     *
+     * @param str a [String] that matches the `MAJOR.MINOR.PATCH` format.
+     */
+    fun fromStringOrNull(str: String): ModuleVersion? = runCatching { fromString(str) }.getOrNull()
+  }
+
+  /** Returns a copy of this [ModuleVersion], with the [patch] increased by one. */
+  fun bump() = copy(patch = patch + 1)
```
rlazo, line 76: `bump` is too generic if it always do patch version bumps

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PostReleasePlugin.kt
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Registers the `versionBump` task for the provided [Project].
+   *
+   * Each SDK has a `gradle.properties` file at its root. This file has a `version` variable, that
+   * is set to the current version of said module. At the end of a release, we bump the versions at
+   * master of SDKs that went out that release to be one patch higher than the currently released
+   * version. We do this to differentiate between the current project, and the latest released
+   * version. If we didn’t bump the version, gradle could end up using a cached version of the
+   * released artifact instead of your local copy during development. As such, the version in every
+   * SDK at master is marked as one patch higher than the actual latest released version.
```
rlazo, line 40:
```suggestion
   * Each SDK has a `gradle.properties` file at its root. This file has a `version` variable, set
   * to its current version. After a release, this `version` should be bumped up to differentiate
   * between code at HEAD, and the latest released version.
```

rlazo: While the extra context seems useful, I think it's enough with less detail. wdyt?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
davidmotson: Maybe taking an enum of the kind of bump, and the default no-param version is a patch?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -0,0 +1,77 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.gradle.plugins
+
+/**
+ * Type-Safe representation of your standard [SemVer](https://semver.org/) versioning scheme.
+ *
+ * Additional labels for pre-release builds are not supported at this time. All versions should fall
```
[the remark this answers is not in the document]
daymxn: fwiw, I'm pretty sure appcheck is out of beta, no? but true, future SDKs will probably need it as well. sgtm!

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -0,0 +1,77 @@
...
+ */
+data class ModuleVersion(val major: Int, val minor: Int, val patch: Int) {
+
+  /** Formatted as `MAJOR.MINOR.PATCH` */
+  override fun toString() = "$major.$minor.$patch"
+
+  companion object {
+
+    /**
+     * Uses Regex to extrapolate the version variables from a provided [String], and turns them into
```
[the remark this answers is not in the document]
daymxn: ope, lemme fix that. It previously did, but I didn't like the usage of regex- so I refactored it.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @see fromStringOrNull
+     */
+    fun fromString(str: String): ModuleVersion =
+      runCatching {
+          val (major, minor, patch) = str.split(".").map { it.toInt() }
+
+          return ModuleVersion(major, minor, patch)
+        }
+        .getOrElse { throw IllegalArgumentException("Invalid format for provided version.", it) }
```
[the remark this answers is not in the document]
daymxn: This more closely aligns with Kotlin stdlib methods; where the base method throws, and an extension of `orNull` is provided to catch it. I don't feel particularly strong about this though- so if you'd much rather we provide an exception free implementation only- I'm down for it.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
daymxn: sgtm:)

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PostReleasePlugin.kt
daymxn: SGTM!

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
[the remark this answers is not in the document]
rlazo: Yeah, I'm more in favor of a exception free experience.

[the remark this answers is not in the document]
daymxn: sgtm:)

rlazo (approved):

PR #4723, opened 2023-02-28
Implement SessionInitiator and hook it up to lifecycle events

mrober:
Implement SessionInitiator and hook it up to lifecycle events.

This is responsible for calling the initiateSessionStart callback whenever a session starts.

Used Longs for time, not go/goodtime, to keep the SDK light on older android api levels. Put the unit on every time variable.

Added unit tests. Used fakes, not mocks or stubs frameworks, similar to the iOS SDK.


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -14,16 +14,33 @@
...
 import com.google.firebase.ktx.app
 
-class FirebaseSessions internal constructor() {
+class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
+  init {
+    val sessionInitiator = SessionInitiator(System::currentTimeMillis, this::initiateSessionStart)
+    val context = firebaseApp.applicationContext.applicationContext
+    if (context is Application) {
+      context.registerActivityLifecycleCallbacks(sessionInitiator.activityLifecycleCallbacks)
+    }
```
samedson, line 30: Should we warn or error log if the context isn't an Application?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -14,16 +14,33 @@
...
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.ktx.app
 
-class FirebaseSessions internal constructor() {
+class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
+  init {
+    val sessionInitiator = SessionInitiator(System::currentTimeMillis, this::initiateSessionStart)
+    val context = firebaseApp.applicationContext.applicationContext
+    if (context is Application) {
+      context.registerActivityLifecycleCallbacks(sessionInitiator.activityLifecycleCallbacks)
```
samedson, line 29: Suggestion: Make this part of the constructor for SessionInitiator instead of exposing the `activityLifecycleCallbacks` property

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
```diff
@@ -0,0 +1,140 @@
...
+    assertThat(sessionStartCounter.count).isEqualTo(1)
+
+    assertThat(sessionStartCounter.count).isEqualTo(1)
+  }
+
+  companion object {
+    private const val SMALL_INTERVAL_MS = 3 * 1000L // not enough time to initiate a new session
+    private const val LARGE_INTERVAL_MS = 90 * 60 * 1000L // enough to initiate another session
+  }
+}
```
samedson, line 140: These tests are great 👍

samedson (approved):
Couple suggestions, but otherwise looks good!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: I did it this way so the SessionInitiator outer class wouldn't have any Context or Activity in the ctor to make the tests simpler.

davidmotson (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
```diff
@@ -0,0 +1,68 @@
...
+  init {
+    initiateSessionStart()
+  }
+
+  fun appBackgrounded() {
+    backgroundTimeMs = currentTimeMs()
+  }
+
+  fun appForegrounded() {
+    val intervalMs = currentTimeMs() - backgroundTimeMs
```
visumickey, line 46: Duration APIs have simple ways to compare/[find difference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-duration/minus.html) time. If there are platform APIs - I would recommend using them over manual calculations unless there is a need for using custom manual calculations.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+
+  internal val activityLifecycleCallbacks =
+    object : ActivityLifecycleCallbacks {
+      override fun onActivityResumed(activity: Activity) = appForegrounded()
+
+      override fun onActivityPaused(activity: Activity) = appBackgrounded()
+
+      override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) = Unit
```
visumickey, line 58: Curious if this will be the place to kick start a session event?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
```diff
@@ -0,0 +1,140 @@
...
+
+class SessionInitiatorTest {
+  class FakeTime {
+    var currentTimeMs = 0L
+      private set
+
+    fun addTimeMs(intervalMs: Long) {
+      currentTimeMs += intervalMs
+    }
+  }
```
visumickey, line 30: Do you see a need for reusing this? May be we can create set of Mocks/Fakes that could be reused? Not for now - but for later.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
```diff
@@ -0,0 +1,140 @@
...
+
+  class SessionStartCounter {
+    var count = 0
+      private set
```
visumickey, line 34: Is this a way to denote `count` to have a setter?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
```diff
@@ -0,0 +1,140 @@
...
+    sessionInitiator.appForegrounded()
+
+    assertThat(sessionStartCounter.count).isEqualTo(1)
+
+    assertThat(sessionStartCounter.count).isEqualTo(1)
+  }
+
+  companion object {
+    private const val SMALL_INTERVAL_MS = 3 * 1000L // not enough time to initiate a new session
+    private const val LARGE_INTERVAL_MS = 90 * 60 * 1000L // enough to initiate another session
```
visumickey, line 138: Similar to earlier - We should consider using Duration APIs.

visumickey (commented):
Thanks for making these changes. Since these are merged now - we can make these changes (if any) in a separate PR.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -14,16 +14,35 @@
...
```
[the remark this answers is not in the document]
mrober: Sam suggested moving this to the SessionInitiator ctor. If we do that, then that instance can only get registered once. Would that be enough to ensure this?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -14,16 +14,35 @@
...
+    } else {
+      Log.w(TAG, "Failed to register lifecycle callbacks, unexpected context ${context.javaClass}.")
+    }
+  }
+
   @Discouraged(message = "This will be replaced with a real API.")
   fun greeting(): String = "Matt says hi!"
```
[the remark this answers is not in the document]
mrober: This greeting is just a place holder for now. I used it in the android test to verify this gets initialized properly. I also use it in my test app to prove the sdk is loaded, it's just for fun and I will remove it when we have a real internal api.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessionsRegistrar.kt
```diff
@@ -30,12 +32,14 @@ internal class FirebaseSessionsRegistrar : ComponentRegistrar {
     listOf(
       Component.builder(FirebaseSessions::class.java)
         .name(LIBRARY_NAME)
-        .factory { FirebaseSessions() }
+        .add(Dependency.required(FirebaseApp::class.java))
+        .factory { container -> FirebaseSessions(container.get(FirebaseApp::class.java)) }
+        .eagerInDefaultApp()
```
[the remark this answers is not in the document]
mrober: I set eager because I want the first session to start at app launch. Would lazy still be good enough for that?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * The [SessionInitiator] is responsible for calling the [initiateSessionStart] callback whenever a
+ * session starts. This will happen at a cold start of the app, and when the app has been in the
+ * background for a period of time (default 30 min) and then comes back to the foreground.
+ *
+ * @hide
+ */
+internal class SessionInitiator(
+  private val currentTimeMs: () -> Long,
+  private val initiateSessionStart: () -> Unit
```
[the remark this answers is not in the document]
mrober: Unit in Kotlin corresponds to `void` in Java.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
mrober: Do you mean onActivityCreated? We call initiateSessionStart() when the class is constructed for the cold boot case.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
mrober: Yeah if this is used somewhere else, I'll pull it out into a top level class and maybe make a package like `testing`.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
mrober: Yeah this is some Kotlin syntax sugar. var means it has a getter and setter of the same visibility by default, public here. Then I overrode the setter to be private.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
```diff
@@ -0,0 +1,68 @@
...
+) {
+  private var backgroundTimeMs = currentTimeMs()
+  private val sessionTimeoutMs = 30 * 60 * 1000L // TODO(mrober): Get session timeout from settings
```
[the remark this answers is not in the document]
mrober: Yup! I had though the `java.time` classes would not work with min sdk 16 (requires 26), but it turns out `kotlin.time.Duration` is fine. I sent a new PR #4733 with this refactored.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
mrober: Done in #4733.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
mrober: Done in #4733.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessionsRegistrar.kt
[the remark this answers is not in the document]
visumickey: I think lazy should be fine. With eager my concern is that we might impact the startup time of the application. Especially since we would be dependent on libraries like Installations and Firelog.

The only caveat here would be the start time of a session. I think the start time of a session and the application start time are different from each other. So, I think lazy should be fine.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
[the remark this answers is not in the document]
visumickey: Got it. Thanks for clarifying!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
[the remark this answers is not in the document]
visumickey: Thanks for introducing Duration APIs there. It looks much cleaner there.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
visumickey: Thank you!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
visumickey: Yes, that was my question. I get that part now. Will the code to start a new session be called in a later PR?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
visumickey: Sounds good, thanks.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
visumickey: Thanks for the clear explanation. Seems to a very useful construct. Should we use this construct in other places of our code?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
visumickey: Thank you!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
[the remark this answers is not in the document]
visumickey: It depends on if the SessionInitator constructor will be called just once or more than once. Also, would the loss of a SessionInitiator object cause a deRegister to be called automatically? If so, this refactor should be fine.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
mrober: The `init { ... }` block gets called as part of the ctor, and that calls `initiateSessionStart()` already. So we start a new session when this object is constructed, representing the cold boot. And we also call it on foreground after the session timeout elapsed.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessionsRegistrar.kt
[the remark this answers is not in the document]
mrober: If I make it lazy, it never initiates on my test app if I never access the SDK. Are we sure that's what we want? When Crashlytics and Fireperf depend on it, will they always access it on every launch?

Issue #4726, opened 2023-03-01
Deep links don't work when clicking an In-App Message


bartek-wesolowski:
### Environment

  * Android Studio version: Android Studio Electric Eel | 2022.1.1 Patch 1
  * Firebase Component: In-App Messaging
  * Component version: 20.3.0

#### Steps to reproduce:

Create a Firebase In-App Messaging campaign with some action. The action link should be a deep link handled by the app. When you then click the message shown in the app, the link is opened in a Chrome Custom Tab instead of being handled by the app.

I suspect that this fragment in `FirebaseInAppMessagingDisplay.java` is the culprit:
```kotlin
  private void launchUriIntent(Activity activity, Uri uri) {
    if (ishttpOrHttpsUri(uri) && supportsCustomTabs(activity)) {
      // If we can launch a chrome view, try that.
      CustomTabsIntent customTabsIntent = new CustomTabsIntent.Builder().build();
      Intent intent = customTabsIntent.intent;
      intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
      intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
      customTabsIntent.launchUrl(activity, uri);
    } else {
      // If we can't launch a chrome view try to launch anything that can handle a URL.
      Intent browserIntent = new Intent(Intent.ACTION_VIEW, uri);
      ResolveInfo info = activity.getPackageManager().resolveActivity(browserIntent, 0);
      browserIntent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
      browserIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
      if (info != null) {
        activity.startActivity(browserIntent);
      } else {
        // If the device can't resolve a url then log, but don't crash.
        Logging.loge("Device cannot resolve intent for: " + Intent.ACTION_VIEW);
      }
    }
  }
```

There should be an option to not use Custom Tabs or, even better, to handle click actions in a custom way.

bartek-wesolowski:
I implemented two alternative solutions in https://github.com/firebase/firebase-android-sdk/pull/4728 and https://github.com/firebase/firebase-android-sdk/pull/4729. I prefer the one with a custom action handler.

bartek-wesolowski:
The second solution would also fix https://github.com/firebase/firebase-android-sdk/issues/3056

argzdev:
[...]

PR #4728, opened 2023-03-01
Allow disabling Custom Tabs when handling In-App Message click actions

bartek-wesolowski:
By default, In-App Message links are handled by opening a Custom Tab. This behavior might be undesirable because deep links don't open in the app. It would be great if there was a way of disabling Custom Tabs.

Related issue https://github.com/firebase/firebase-android-sdk/issues/4726

I implemented an alternative (arguably better) solution here https://github.com/firebase/firebase-android-sdk/pull/4729


firebase-inappmessaging-display/src/main/java/com/google/firebase/inappmessaging/display/FirebaseInAppMessagingDisplay.java
```diff
@@ -82,6 +82,8 @@ public class FirebaseInAppMessagingDisplay extends FirebaseInAppMessagingDisplay
       20 * 1000; // auto dismiss after 20 seconds for banner
   static final long INTERVAL_MILLIS = 1000;
 
+  public boolean useCustomTabs = true;
```
thatfiredev, line 85: We should probably make this variable private and have public getters/setters to be consistent with other Firebase APIs.

PR #4733, opened 2023-03-01
Use kotlin.time.Duration instead of Long for time intervals

mrober:
Use `kotlin.time.Duration` instead of `Long` for time intervals.

`kotlin.time.Duration` is safe to use with min api level 16. It is `java.time.Duration` that requres api level 26. We still can't use `java.time.Instant` because of our min api level, but I avoided it by making everything `Duration`s.

Also switched to using `SystemClock.elapsedRealtime()` which is *"the time since the system was booted, and include deep sleep. This clock is guaranteed to be monotonic, and continues to tick even when the CPU is in power saving modes, so is the recommend basis for general purpose interval timing."*

Fixed the copyright format on Kotlin files so it shows collapsed by default in AS.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
@@ -19,6 +19,8 @@ package com.google.firebase.sessions
 import android.app.Activity
 import android.app.Application.ActivityLifecycleCallbacks
 import android.os.Bundle
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
 
 /**
  * The [SessionInitiator] is responsible for calling the [initiateSessionStart] callback whenever a
@@ -28,23 +30,23 @@ import android.os.Bundle
  * @hide
  */
 internal class SessionInitiator(
-  private val currentTimeMs: () -> Long,
+  private val elapsedRealtime: () -> Duration,
   private val initiateSessionStart: () -> Unit
 ) {
-  private var backgroundTimeMs = currentTimeMs()
-  private val sessionTimeoutMs = 30 * 60 * 1000L // TODO(mrober): Get session timeout from settings
+  private var backgroundTime = elapsedRealtime()
+  private val sessionTimeout = 30.minutes // TODO(mrober): Get session timeout from settings
 
   init {
     initiateSessionStart()
   }
 
   fun appBackgrounded() {
-    backgroundTimeMs = currentTimeMs()
+    backgroundTime = elapsedRealtime()
   }
 
   fun appForegrounded() {
-    val intervalMs = currentTimeMs() - backgroundTimeMs
-    if (intervalMs > sessionTimeoutMs) {
+    val interval = elapsedRealtime() - backgroundTime
+    if (interval > sessionTimeout) {
       initiateSessionStart()
     }
   }
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/WallClock.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/WallClock.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/WallClock.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlin.time.Duration.Companion.nanoseconds
+
+/** Util object for "wall clock" time functions. */
+internal object WallClock {
+  /** Gets the [Duration] elapsed in "wall clock" time since device boot. */
+  fun elapsedRealtime(): Duration =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+      SystemClock.elapsedRealtimeNanos().nanoseconds
+    } else {
+      SystemClock.elapsedRealtime().milliseconds
```
samedson, line 32: Does one of these need to be multiplied, or is that handled as they are `Duration`s?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
```diff
@@ -28,23 +30,23 @@ import android.os.Bundle
  * @hide
  */
 internal class SessionInitiator(
-  private val currentTimeMs: () -> Long,
+  private val elapsedRealtime: () -> Duration,
```
samedson, line 33: Could you add a comment to explain why this is elapsed time instead of current time (eg. what the difference between this and the iOS SDK)?

samedson: It looks like this is just time since boot instead of unix timestamt for ios

samedson (approved):
Couple suggestions. Will leave to Visu to review too

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/WallClock.kt
mrober: The `Duration` extensions handle the units. They are correct here, I also tested on both an old and new api level to get both code paths.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt
mrober: When playing with the clocks for this PR, I found that unix time is not a good way to measure intervals like session length. Unix time can change when the customer changes the time on their device, daylight savings, or other funny timezone things even leap seconds, etc. Maybe iOS should change to a different clock, something like this that is a monotonic uniform clock.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
```diff
@@ -134,7 +137,7 @@ class SessionInitiatorTest {
   }
 
   companion object {
-    private const val SMALL_INTERVAL_MS = 3 * 1000L // not enough time to initiate a new session
-    private const val LARGE_INTERVAL_MS = 90 * 60 * 1000L // enough to initiate another session
+    private val SMALL_INTERVAL = 3.seconds // not enough time to initiate a new session
+    private val LARGE_INTERVAL = 90.minutes // enough to initiate another session
```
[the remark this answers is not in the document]
mrober: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/WallClock.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.nanoseconds
+
+/** Util object for "wall clock" time functions. */
+internal object WallClock {
+  /** Gets the [Duration] elapsed in "wall clock" time since device boot. */
+  fun elapsedRealtime(): Duration =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+      SystemClock.elapsedRealtimeNanos().nanoseconds
```
visumickey, line 30: For easier understanding, can we convert the nanoSeconds here to milliseconds? That way, it becomes clear in the code that both are represented as duration of time with reference of milliseconds level precision captured from the SystemClock?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/WallClock.kt
mrober: Done. And since we will always be millis precision, we can just use the older api for all apl levels.

PR #4734, opened 2023-03-01
Remove irrelevant exceptions and update tests

qdpham13:
- Remove exceptions that users can't do anything about and that don't break Realtime
- Remote duplicate test and update test to not rely on exception

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
@@ -105,6 +105,7 @@ public class ConfigAutoFetch {
         handleNotifications(inputStream);
         inputStream.close();
       } catch (IOException ex) {
+        // Stream was interrupted due to a transient issue and the system will retry the connection.
         Log.d(TAG, "Stream was cancelled due to an exception: " + ex.toString());
       } finally {
         httpURLConnection.disconnect();
@@ -151,6 +152,8 @@ public class ConfigAutoFetch {
               }
             }
           } catch (JSONException ex) {
+            // Message was mangled up and so it was unable to be parsed. User is notified of this
+            // because it there could be a new configuration that needs to be fetched.
             propagateErrors(
                 new FirebaseRemoteConfigClientException(
                     "Unable to parse config update message.",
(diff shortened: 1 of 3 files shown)


danasilver:
Could you add some notes or code comments where the exceptions are removed/added indicating what the issue is/why we're catching and swallowing the exception?

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -105,11 +105,8 @@ public void listenForNotifications() {
         handleNotifications(inputStream);
         inputStream.close();
       } catch (IOException ex) {
-        propagateErrors(
-            new FirebaseRemoteConfigClientException(
-                "Unable to parse config update message.",
-                ex.getCause(),
-                FirebaseRemoteConfigException.Code.CONFIG_UPDATE_MESSAGE_INVALID));
+        // Stream was interrupted due to a transient issue and the system will retry the connection.
+        Log.d(TAG, "Stream was cancelled due to an exception: " + ex.toString());
```
danasilver, line 109: Would include "Retrying" somewhere in this message.

qdpham13: added

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigAutoFetch.java
```diff
@@ -105,11 +105,12 @@ public void listenForNotifications() {
...
-        propagateErrors(
-            new FirebaseRemoteConfigClientException(
-                "Unable to parse config update message.",
-                ex.getCause(),
-                FirebaseRemoteConfigException.Code.CONFIG_UPDATE_MESSAGE_INVALID));
+        // Stream was interrupted due to a transient issue and the system will retry the connection.
+        Log.d(
+            TAG,
+            String.format(
+                "Stream was cancelled due to an exception: %s. Retrying the connection...",
```
danasilver, line 112: Sorry last nit 😅  Was thinking about the exception in the middle of the log and remembered `Log.d` can take an exception as a parameter and print it, so we can use those for the logs here: https://developer.android.com/reference/android/util/Log.html#d(java.lang.String,%20java.lang.String,%20java.lang.Throwable)

qdpham13: Added it to the logs

danasilver (approved):

PR #4743, opened 2023-03-03
Tool that calculate workflow/job failure rate and collect failure logs

sunmou99:
A tool that calculate workflow/job failure rate and a tool that collect ci_test failure logs

see: https://github.com/firebase/firebase-android-sdk/tree/sunmou/ci_status/ci/fireci/workflow_summary


rlazo:
The script is incide the `fireci` directory, but it doesn't integrate with fireci.

@yifanyang could you take a look at that part?

ci/fireci/workflow_summary/github.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2023 Google LLC
```
rlazo, line 1: Seems like `PyGithub` does provide some of these utilities. For example 

```
/repos/{owner}/{repo}/actions/workflows/{workflow_id}/runs
GET: [github.Workflow.Workflow.get_runs()](https://pygithub.readthedocs.io/en/latest/github_objects/Workflow.html#github.Workflow.Workflow.get_runs)
```
https://pygithub.readthedocs.io/en/latest/apis.html

I could be missing some context here, why we preferred this option rather than use the api lib?

ci/fireci/workflow_summary/collect_ci_test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+
+    if job['failure_rate'] > 0:
+      failure_rate = job['failure_rate']  
+      total_count = job['total_count'] 
+      success_count = job['success_count'] 
+      failure_count = job['failure_count']      
+
+      log_file_path = os.path.join(file_folder, '%s.log'%job_name)
+      file_log = open(log_file_path, 'w')
+      file_log.write('\n{}:\nFailure rate:{:.2%} \nTotal count: {} (success: {}, failure: {})\nFailed jobs:'.format(job_name, failure_rate, total_count, success_count, failure_count))
```
rlazo, line 61: Prefer f-strings over ''.format()

ci/fireci/workflow_summary/github.py
sunmou99: Last time (~ 1 year ago) I checked PyGithub, it didn't have great support for Github workflows/actions. Plus, using Python3's `requests` module for REST can avoid an additional dependency.

sunmou99: I just checked the PyGithub APIs, it still seems have no workflow jobs/logs related APIs.

yifanyang (changes requested):
> The script is incide the `fireci` directory, but it doesn't integrate with fireci.
> 
> @yifanyang could you take a look at that part?

If the script is supposed to be standalone, maybe just place it inside `ci` but not `ci/fireci`. If you do want to integrate with `fireci`, use the decorator `@ci_command`, for example:

https://github.com/firebase/firebase-android-sdk/blob/46e7a69781b6651582be2ea9413cd6c30913b829/ci/fireci/fireciplugins/fireperf.py#L27-L40

Another thing: the script depends on a couple of external libraries, such as `argparse`, `requests`. Do you want to add a setup instruction for your script?

rlazo:
> > The script is incide the `fireci` directory, but it doesn't integrate with fireci.
> > @yifanyang could you take a look at that part?
> 
> If the script is supposed to be standalone, maybe just place it inside `ci` but not `ci/fireci`. If you do want to integrate with `fireci`, use the decorator `@ci_command`, for example:
> 
> https://github.com/firebase/firebase-android-sdk/blob/46e7a69781b6651582be2ea9413cd6c30913b829/ci/fireci/fireciplugins/fireperf.py#L27-L40
> 
> Another thing: the script depends on a couple of external libraries, such as `argparse`, `requests`. Do you want to add a setup instruction for your script?

That's one of the benefits of integrating with fireci, since it already has it setup, right?

yifanyang:
> > > The script is incide the `fireci` directory, but it doesn't integrate with fireci.
> > > @yifanyang could you take a look at that part?
> > 
> > 
> > If the script is supposed to be standalone, maybe just place it inside `ci` but not `ci/fireci`. If you do want to integrate with `fireci`, use the decorator `@ci_command`, for example:
> > https://github.com/firebase/firebase-android-sdk/blob/46e7a69781b6651582be2ea9413cd6c30913b829/ci/fireci/fireciplugins/fireperf.py#L27-L40
> > 
> > Another thing: the script depends on a couple of external libraries, such as `argparse`, `requests`. Do you want to add a setup instruction for your script?
> 
> That's one of the benefits of integrating with fireci, since it already has it setup, right?

Yes. Here is the example:

https://github.com/firebase/firebase-android-sdk/blob/8d99ce005ac8412e12b396e8a8df199c40970236/ci/fireci/setup.cfg#L5-L17

sunmou99:
> > The script is incide the `fireci` directory, but it doesn't integrate with fireci.
> > @yifanyang could you take a look at that part?
> 
> If the script is supposed to be standalone, maybe just place it inside `ci` but not `ci/fireci`. If you do want to integrate with `fireci`, use the decorator `@ci_command`, for example:
> 
> https://github.com/firebase/firebase-android-sdk/blob/46e7a69781b6651582be2ea9413cd6c30913b829/ci/fireci/fireciplugins/fireperf.py#L27-L40
> 
> Another thing: the script depends on a couple of external libraries, such as `argparse`, `requests`. Do you want to add a setup instruction for your script?

Standalone for now, thus I moved it to `ci`.

ci/workflow_summary/workflow_information.py
```diff
@@ -0,0 +1,250 @@
...
+
+  parser.add_argument('-f', '--folder', help='Workflow and job information will be store here, default is current datatime')
+
+  args = parser.parse_args()
+  return args
+
+
+if __name__ == '__main__':
+  main()
+
```
yifanyang, line 250: The last line is not empty?

yifanyang (changes requested):
I see that you moved them to a new folder. But still, the script is using external libraries. We cannot assume it is available anywhere. We will need to clarify which version of those dependencies to use and how to install them. The current practice (such as the setup for `fireci`) is to use Python virtual environment to isolate installations for different purposes.

Another nit: can you make the commit title more concise?

ci/workflow_summary/github.py
```diff
@@ -0,0 +1,75 @@
...
+    return False
+
+
+def list_workflows(token, workflow_id, params):
+  """https://docs.github.com/en/rest/actions/workflow-runs#list-workflow-runs-for-a-workflow"""
+  url = f'{GITHUB_API_URL}/actions/workflows/{workflow_id}/runs'
+  headers = {'Accept': 'application/vnd.github+json', 'Authorization': f'token {token}'}
+  with requests.get(url, headers=headers, params=params,
+                    stream=True, timeout=TIMEOUT_LONG) as response:
+    logging.info("list_workflows: %s, params: %s, response: %s", url, params, response)
```
rlazo, line 50: Use single-quotes to be consistent with the rest of the file.

ci/workflow_summary/github.py
```diff
@@ -0,0 +1,75 @@
...
+BASE_URL = 'https://api.github.com'
+GITHUB_API_URL = '%s/repos/%s/%s' % (BASE_URL, OWNER, REPO)
+
+
+def set_api_url(owner, repo): 
+  if owner and repo:
+    global OWNER, REPO, GITHUB_API_URL
+    OWNER = owner
+    REPO = repo
+    GITHUB_API_URL = '%s/repos/%s/%s' % (BASE_URL, OWNER, REPO)
```
rlazo, line 37: Modifying global state can cause subtle and hard to debug issues. You could potentially create a class to wrap these functions and stores `owner` and `repo`.

ci/workflow_summary/github.py
```diff
@@ -0,0 +1,75 @@
...
+    return response.json()
+
+def job_logs(token, job_id):
+  """https://docs.github.com/rest/reference/actions#download-job-logs-for-a-workflow-run"""
+  url = f'{GITHUB_API_URL}/actions/jobs/{job_id}/logs'
+  headers = {'Accept': 'application/vnd.github+json', 'Authorization': f'token {token}'}
+  with requests.get(url, headers=headers, allow_redirects=False,
+                    stream=True, timeout=TIMEOUT_LONG) as response:
+    logging.info("job_logs: %s response: %s", url, response)
+    if response.status_code == 302:
```
rlazo, line 69: Why `job_logs` supports redirects while the others don't?

ci/workflow_summary/workflow_information.py
```diff
@@ -0,0 +1,249 @@
...
+  if args.folder:
+    file_folder = os.path.normpath(args.folder)
+  else:
+    file_folder = os.path.normpath(datetime.datetime.utcnow().strftime('%Y-%m-%d+%H:%M:%S'))
+  if not os.path.exists(file_folder):
+    os.makedirs(file_folder)
+
+  workflow_summary = get_workflow_summary(args)
+  workflow_summary_file_path = os.path.join(file_folder, 'workflow_summary.json')
+  json.dump(workflow_summary, open(workflow_summary_file_path,'w'))
```
rlazo, line 49: An `open` without a close leaks resources, prefer 

```
with open(workflow_summary_file_path, 'w') as fd:
  json.dump(workflow_summary, fd)
```

ci/workflow_summary/workflow_information.py
```diff
@@ -0,0 +1,249 @@
...
+  list_workflows_params = {'status': 'completed', 'created': created, 'page': workflow_page, 'per_page': per_page}
+  if args.event:
+    list_workflows_params['event'] = args.event
+  if args.actor:
+    list_workflows_params['actor'] = args.actor
+  if args.branch:
+    list_workflows_params['branch'] = args.branch
+
+  request_workflow_list = True
+  while request_workflow_list:
```
rlazo, line 92: `for page in itertools.count()` can replace the while with +=1 and your `else:` body in line 112 becomes a `break`

ci/fireci/workflow_summary/github.py
rlazo: I Didn't realize that the were no logs related API available. Thanks for the clarification!

ci/workflow_summary/github.py
sunmou99: `job_logs` doesn't support redirects. otherwise, we are not able to get the link to download the logs.

ci/workflow_summary/workflow_information.py
```diff
@@ -0,0 +1,250 @@
...
+    list_workflows_params['actor'] = args.actor
+  if args.branch:
+    list_workflows_params['branch'] = args.branch
+
+  while True:
+    workflow_page += 1
+    list_workflows_params['page'] = workflow_page
+    workflows = gh.list_workflows(token, workflow_name, list_workflows_params)
+
+    if 'workflow_runs' in workflows and workflows['workflow_runs']:
```
rlazo, line 99: It will improve readability if you invert the conditional:

```
while True:
  ...
  if 'workflow_runs' not in workflows or not workflows['workflow_runs']:
    break
  ...
```
This will reduce nesting

ci/workflow_summary/collect_ci_test_logs.py
```diff
@@ -0,0 +1,91 @@
...
+
+  file_folder = args.folder
+  if not os.path.exists(file_folder):
+    logging.error(f'{file_folder} doesn\'t exist')
+    exit(1)
+
+  job_summary = json.load(open(os.path.join(file_folder, 'job_summary.json')))
+
+  for job_name in job_summary:
+    if job_name in ['Determine changed modules','Unit Tests (matrix)','Publish Tests Results','Unit Test Results','Instrumentation Tests','Unit Tests']:
```
rlazo, line 48: Use a file level constant for this list to make it easier to read.

ci/workflow_summary/github.py
```diff
@@ -0,0 +1,62 @@
...
+    headers = {'Accept': 'application/vnd.github+json', 'Authorization': f'token {token}'}
+    with requests.get(url, headers=headers, allow_redirects=False,
+                      stream=True, timeout=TIMEOUT_LONG) as response:
+      logging.info('job_logs: %s response: %s', url, response)
+      if response.status_code == 302:
+        with requests.get(response.headers['Location'], headers=headers, allow_redirects=False,
+                          stream=True, timeout=TIMEOUT_LONG) as get_log_response:
+          return get_log_response.content.decode('utf-8')
+      else:
+        print('no log avaliable')
```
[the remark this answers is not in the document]
sunmou99: Fixed. This is the old debugging code. I kept it but forgot to modify it.

sunmou99:
> I see that you moved them to a new folder. But still, the script is using external libraries. We cannot assume it is available anywhere. We will need to clarify which version of those dependencies to use and how to install them. The current practice (such as the setup for `fireci`) is to use Python virtual environment to isolate installations for different purposes.
> 
> Another nit: can you make the commit title more concise?

I think it's fine that not specifying the python library version.
1. We are using the very basic functions from the external libraries, thus it should work in most versions.
2. Install library with multiple versions can be a headache. This tool is not a Python program like fireci, thus unable to leverage the config file
3. This tool doesn't have a lot of customers
4. It take time to find out and test out the minimum support version.

Once we have plan to advertise the tool, let's take the library version in to consideration.

yifanyang:
> 1. We are using the very basic functions from the external libraries, thus it should work in most versions.

The script may not work with future versions of these libraries, if there is a backward incompatible change. The purpose of specifying a version is to make sure the script always resolve to the version we have verified here to be working.

> 2. Install library with multiple versions can be a headache. This tool is not a Python program like fireci, thus unable to leverage the config file

I recommend Python [virtual environments](https://docs.python.org/3/tutorial/venv.html) to isolate installations of different libraries.

> 3. This tool doesn't have a lot of customers

Maybe it's just my opinion, I'd prefer whatever we deliver to be with a good quality, regardless of the customer base.

> 4. It take time to find out and test out the minimum support version.

See the first reply above. We only need a version that we are sure everything is working right?

> Once we have plan to advertise the tool, let's take the library version in to consideration.

Nevertheless, it is up to you whether you want to incorporate these feedback.

rlazo (approved):

PR #4747, opened 2023-03-06
Add SessionGenerator to generate the Session ID

samedson:
Adds a SessionGenerator that rotates the Session ID and index.

#no-changelog

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -25,7 +25,7 @@ import com.google.firebase.ktx.app
 
 class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
 
-  private val sessionGenerator = SessionGenerator(true)
+  private val sessionGenerator = SessionGenerator(collectEvents = true)
 
   init {
     val sessionInitiator = SessionInitiator(WallClock::elapsedRealtime, this::initiateSessionStart)
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -24,6 +24,9 @@ import com.google.firebase.ktx.Firebase
 import com.google.firebase.ktx.app
 
 class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
+
+  private val sessionGenerator = SessionGenerator(true)
+
   init {
     val sessionInitiator = SessionInitiator(WallClock::elapsedRealtime, this::initiateSessionStart)
     val context = firebaseApp.applicationContext.applicationContext
@@ -40,6 +43,8 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
   private fun initiateSessionStart() {
     // TODO(mrober): Generate a session
     Log.i(TAG, "Initiate session start")
+
+    sessionGenerator.generateNewSession()
   }
 
   companion object {
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
@@ -37,7 +37,13 @@ internal data class SessionInfo(
  * @hide
  */
 internal class SessionGenerator(collectEvents: Boolean) {
-  private var thisSession: SessionInfo = SessionInfo("", "", collectEvents, -1)
+  private var thisSession: SessionInfo =
+    SessionInfo(
+      sessionId = "",
+      firstSessionId = "",
+      collectEvents = collectEvents,
+      sessionIndex = -1
+    )
 
   private var firstSessionId = ""
   private var sessionIndex: Int = -1
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
@@ -47,7 +47,6 @@ internal class SessionGenerator(collectEvents: Boolean) {
   // Generates a new Session ID. If there was already a generated Session ID
   // from the last session during the app's lifecycle, it will also set the last Session ID
   fun generateNewSession() {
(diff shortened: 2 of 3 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -24,6 +24,9 @@ import com.google.firebase.ktx.Firebase
 import com.google.firebase.ktx.app
 
 class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
+
+  private val sessionGenerator = SessionGenerator(true)
```
mrober, line 28: Can you write it with a named param like `SessionGenerator(collectEvents = true)`? It will explain the meaning of the boolean literal.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+  fun generateNewSession() {
+    val newSessionId = UUID.randomUUID().toString().replace("-", "").lowercase()
+
+    // If firstSessionId is set, use it. Otherwise set it to the
+    // first generated Session ID
+    firstSessionId = if (firstSessionId.isEmpty()) newSessionId else firstSessionId
+
+    sessionIndex += 1
+
+    thisSession = SessionInfo(newSessionId,
```
mrober, line 58: Is this formatted using ktfmt google internal style?

samedson: Did that now 👍

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
samedson: Yes I'll do this everywhere - didn't realize it was possible!

mrober (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -40,6 +43,8 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
   private fun initiateSessionStart() {
     // TODO(mrober): Generate a session
```
visumickey, line 44: @samedson Should we remove this TODO?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+internal class SessionGenerator(collectEvents: Boolean) {
+  private var firstSessionId = ""
+  private var sessionIndex: Int = -1
+  private var collectEvents = collectEvents
+
+  private var thisSession: SessionInfo =
+    SessionInfo(
+      sessionId = "",
+      firstSessionId = "",
+      collectEvents = collectEvents,
```
visumickey, line 48: Does SessionInfo need the `collectEvents` state?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+package com.google.firebase.sessions
+
+import java.util.UUID
+
+/**
+ * [SessionInfo] is a data class responsible for storing information about the current Session.
+ *
+ * @hide
+ */
+internal data class SessionInfo(
```
visumickey, line 26: Is this fine to have a separate class defined inside another class? I see a need for this `sessionInfo` to be accessible outside of this context - should we move this to something more open for other classes to use?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionGeneratorTest.kt
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package com.google.firebase.sessions
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+class SessionGeneratorTest {
+  fun isValidSessionId(sessionId: String): Boolean {
+    if (sessionId.length != 32) {
```
visumickey, line 24: Is this a valid assumption? Why a specific need to be exactly 32 characters? I would rather recommend validating alpha numeric value of the sessionID.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * The [SessionGenerator] is responsible for generating the Session ID, and keeping the
+ * [SessionInfo] up to date with the latest values.
+ *
+ * @hide
+ */
+internal class SessionGenerator(collectEvents: Boolean) {
+  private var firstSessionId = ""
+  private var sessionIndex: Int = -1
+  private var collectEvents = collectEvents
```
visumickey, line 42: Do we plan to have a dynamic toggle of `collectEvents` state? Eg: SessionGenerator starts with the collectEvents state to be `false` and somewhere between turns to be `true`. Is this a usecase we want to support?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
samedson: Yeah we'll use this later in the SDK. This class is almost identical to the iOS one.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionGeneratorTest.kt
samedson: I believe this is a valid assumption of UUIDs - we have this in the iOS SDK too

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
samedson: Not anymore - we decided that the collectEvents state is consistent for the run of the app

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
samedson: Done

PR #4753, opened 2023-03-07
Update Realtime RC javadocs to align with iOS docs.

danasilver:
Took a pass at revising the reference docs based on https://github.com/firebase/firebase-ios-sdk/pull/10060.


qdpham13 (approved):
LGTM

PR #4762, opened 2023-03-09
Check if the trace times are not NULL before creating the AppStartTrace.

### LINKED ISSUE #4730: com.google.firebase.perf.util.Timer.getDurationMicros NullPointerException
### environment

  * Android Studio version: 2022.3.1
  * Firebase Component: firebase-perf
  * Component version:20.2.0

### problem
com.google.firebase.perf.util.Timer.getDurationMicros function has a NullPointerException
For details, please see the log below
```  
Fatal Exception: java.lang.NullPointerException: Attempt to read from field 'long  
com.google.firebase.perf.util.Timer.elapsedRealtimeMicros' on a null object reference in method 'long  
com.google.firebase.perf.util.Timer.getDurationMicros(com.google.firebase.perf.util.Timer)'
       at com.google.firebase.perf.util.Timer.getDurationMicros(Timer.java:144)
       at com.google.firebase.perf.metrics.AppStartTrace.logAppStartTrace(AppStartTrace.java:320)
       at com.google.firebase.perf.metrics.AppStartTrace.lambda$t99Qe_LxxEJSDI0PgQVKJpUGMEc()
       at com.google.firebase.perf.metrics.-$$Lambda$AppStartTrace$t99Qe_LxxEJSDI0PgQVKJpUGMEc.run(:2)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
       at java.lang.Thread.run(Thread.java:1012)
```

visumickey:
Check if the trace times are not NULL before creating the AppStartTrace. Fixes #4730


firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -404,26 +404,31 @@ private void logAppStartTrace() {
...
+      traceMetricBuilder = TraceMetric.newBuilder();
+      traceMetricBuilder
+          .setName(Constants.TraceNames.ON_START_TRACE_NAME.toString())
+          .setClientStartTimeUs(onCreateTime.getMicros())
+          .setDurationUs(onCreateTime.getDurationMicros(onStartTime));
+      subtraces.add(traceMetricBuilder.build());
+    }
+
+    // onResumeTime is not captured in all situations, so checking for valid value before using it.
+    if (onResumeTime != null) {
```
raymondlam, line 424: For `onResumeTime`, this should always be populated since `logAppStartTrace` is only called after assigning `onResumeTime` in `onActivityResumed`

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -404,26 +404,31 @@ private void logAppStartTrace() {
...
+      traceMetricBuilder = TraceMetric.newBuilder();
+      traceMetricBuilder
+          .setName(Constants.TraceNames.ON_RESUME_TRACE_NAME.toString())
+          .setClientStartTimeUs(onStartTime.getMicros())
+          .setDurationUs(onStartTime.getDurationMicros(onResumeTime));
+      subtraces.add(traceMetricBuilder.build());
+    }
```
raymondlam, line 431: Move this under the `onStartTime != null` if statement since this depends on `onStartTime` not being null.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -404,26 +404,31 @@ private void logAppStartTrace() {
...
-    traceMetricBuilder = TraceMetric.newBuilder();
-    traceMetricBuilder
-        .setName(Constants.TraceNames.ON_RESUME_TRACE_NAME.toString())
-        .setClientStartTimeUs(onStartTime.getMicros())
-        .setDurationUs(onStartTime.getDurationMicros(onResumeTime));
-    subtraces.add(traceMetricBuilder.build());
+    TraceMetric.Builder traceMetricBuilder = TraceMetric.newBuilder()
+              .setName(Constants.TraceNames.ON_CREATE_TRACE_NAME.toString())
+              .setClientStartTimeUs(getClassLoadTimeCompat().getMicros())
+              .setDurationUs(getClassLoadTimeCompat().getDurationMicros(onCreateTime));
```
raymondlam, line 410: Is it possible for `onCreateTime` to be null? Eg. Could `StartFromBackgroundRunnable` be triggered before onCreate even gets called? The comment around it suggests so. If so, we will need more if to cover `ON_CREATE_TRACE_NAME` and `ON_START_TRACE_NAME`

raymondlam (commented):
It would be nice if there were some tests for these edge cases. I'm not sure how easy it is for that.
Seeing that there is some trickiness to these cases, I would recommend removing the subtraces rather than keeping them.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
visumickey: That is a good point. Makes sense.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
visumickey: Done.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
visumickey: In case of app starting from background using `StartFromBackgroundRunnable`, logAppStartTrace would not even be called since we do not dispatch app start trace in those situations. So, this code path would not even executed in that case. Essentially, `OnActivityResumed` will not be even called. If `OnActivityResumed` is called, it is fair to assume that `OnActivityCreate` would be called.

visumickey:
> It would be nice if there were some tests for these edge cases. I'm not sure how easy it is for that. Seeing that there is some trickiness to these cases, I would recommend removing the subtraces rather than keeping them.

I thought about this. But felt the changes would be a lot more than the scope of this PR. So, will create a separate PR to remove all these experiments.

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
visumickey: Marking it as resolved. Please feel free to reopen if you disagree with my response.

rlazo (approved):

PR #4763, opened 2023-03-09
Add SessionEvent data class

mrober:
Add SessionEvent data class and start populating some of the fields.

Renamed the internal class to SessionState to keep it separated from the data classes to be serialized. 

#no-changelog


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -41,10 +41,19 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
   fun greeting(): String = "Matt says hi!"
 
   private fun initiateSessionStart() {
-    // TODO(mrober): Generate a session
-    Log.i(TAG, "Initiate session start")
+    val sessionState = sessionGenerator.generateNewSession()
+    val sessionEvent =
+      SessionEvent(
```
samedson, line 46: Could we encapsulate this in a class or function that generates the event based on the sessionState input? Eventually this will expand a lot once we have different classes to populate the session.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -41,10 +41,19 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
...
+        eventType = SESSION_START,
+        sessionData =
```
samedson, line 48: Then along the lines of the above comment, could we add a test that makes sure we populate the session correctly given a sessionState? This will be trivial for now but more important later on.

samedson: We may want to mock SessionState for this test

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
samedson: This constructor may be a constructor on `SessionEvent` itself if that's possible for data classes

mrober: The only for the data class ctor to take it, is to make it a property which seems ugly. So I make a function instead.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -41,10 +40,10 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
   fun greeting(): String = "Matt says hi!"
 
   private fun initiateSessionStart() {
-    // TODO(mrober): Generate a session
-    Log.i(TAG, "Initiate session start")
+    val sessionState = sessionGenerator.generateNewSession()
+    val sessionEvent = SessionEvent.sessionStart(sessionState)
```
visumickey, line 44: I think this is very close to the iOS implementation. Just to confirm, here is how I think the current piece of code would evolve for event dispatch. Please let me know if you think otherwise.

1. Session Generator is stateful and it returns sessions related information for the current session event. (ID, Index, previous ID, parent ID and likes)
2.SessionEvent class takes those details to fill in more details
3. Firebase Sessions later embellishes the event with details like data collection state.
4. Firebase Sessions will use Firelog services to hand over the event.

Curious as to where would the config layer fit in here and how would sampling be done to either capture or skip an event?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
```diff
@@ -0,0 +1,69 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.sessions
+
+/**
+ * Contains the relevant information around an App Quality Session.
```
visumickey, line 20: Lets not call it "App Quality Session" yet. Let's reference it as "Firebase Session Start Event".

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * See go/app-quality-unified-session-definition for more details. Keep in sync with
+ * https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/ProtoSupport/Protos/sessions.proto
+ */
+// TODO(mrober): Add and populate all fields from sessions.proto
+internal data class SessionEvent(
+  /** The type of event being reported. */
+  val eventType: EventType,
```
visumickey, line 28: Do we need to remember this state? It is already a part of SessionStartEvent, can we use it from there? I'm trying to avoid multiple source of truth here.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
```diff
@@ -0,0 +1,69 @@
...
+package com.google.firebase.sessions
+
+/**
+ * Contains the relevant information around an App Quality Session.
+ *
+ * See go/app-quality-unified-session-definition for more details. Keep in sync with
+ * https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/ProtoSupport/Protos/sessions.proto
+ */
+// TODO(mrober): Add and populate all fields from sessions.proto
+internal data class SessionEvent(
```
visumickey, line 26: Just confirming that this class will eventually do more things than what it is today: Add more App related fields to the session event?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -19,11 +19,11 @@ package com.google.firebase.sessions
 import java.util.UUID
 
 /**
- * [SessionInfo] is a data class responsible for storing information about the current Session.
+ * [SessionState] is a data class responsible for storing information about the current Session.
```
visumickey, line 22: Nitpick comment here. I prefer `SessionInfo` or `SessionDetails` over `SessionState`. Session state makes me feel this this talks about the state of the session where in state would mean - Created, embellished, cancelled, dropped, dispatched. But in reality, we are saving the details of the current session.

Just a preference here. If you feel too strong about `SessionState`, I can be convinced.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -32,43 +32,39 @@ internal data class SessionInfo(
 
 /**
  * The [SessionGenerator] is responsible for generating the Session ID, and keeping the
- * [SessionInfo] up to date with the latest values.
+ * [SessionState] up to date with the latest values.
  *
  * @hide
  */
-internal class SessionGenerator(collectEvents: Boolean) {
+internal class SessionGenerator(private var collectEvents: Boolean) {
```
visumickey, line 39: wow! We can call parameters with access protectors? `private var` is so cool! TIL!

visumickey (approved):
Let's add no change log hash tag to the description.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
mrober: This is there to match the [proto](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/ProtoSupport/Protos/sessions.proto) 1:1, this data class will be serialized and sent to firelog.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
mrober: Yup this will eventually have all the fields from the proto.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
mrober: Renamed to `SessionDetails`.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
mrober: Done. Called it "Firebase Session Event"

samedson (approved):
LGTM

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: Yup, and for sampling we'll handle that when we do Settings. We will pass in something like `collectEvents = shouldCollectEvents(...)` to the SessionGenerator.

PR #4770, opened 2023-03-10
Validate query purpose in spec tests

milaGGL:
#no-changelog
Port from https://github.com/firebase/firebase-js-sdk/pull/7113

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
@@ -990,9 +990,6 @@ public abstract class SpecTestCase implements RemoteStoreCallback {
           expectedActiveTargets.put(targetId, new ArrayList<>());
           for (int i = 0; i < queryArrayJson.length(); i++) {
             Query query = parseQuery(queryArrayJson.getJSONObject(i));
-            // TODO: populate the purpose of the target once it's possible to encode that in the
-            // spec tests. For now, hard-code that it's a listen despite the fact that it's not
-            // always the right value.
 
             QueryPurpose purpose = QueryPurpose.LISTEN;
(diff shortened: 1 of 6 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -1134,9 +1136,11 @@ private void validateActiveTargets() {
       TargetData expectedTarget = expectedQueries.get(0);
       TargetData actualTarget = actualTargets.get(expected.getKey());
 
-      // TODO: validate the purpose of the target once it's possible to encode that in the
-      // spec tests. For now, only validate properties that can be validated.
+      // TODO: Simplify the TargetData validation when the sequenceNumber can be consistent as well.
```
dconeybe, line 1139: I think this comment should be re-worded a bit. I don't fully understand what it means. Does something like this make sense:

TODO: Replace the assertEquals() checks on the individual properties of TargetData below with the single assertEquals on the TargetData objects themselves if the sequenceNumber is ever made to be consistent.

firebase-firestore/src/test/java/com/google/firebase/firestore/spec/SpecTestCase.java
```diff
@@ -1134,9 +1136,11 @@ private void validateActiveTargets() {
...
       // assertEquals(expectedTarget, actualTarget);
+
+      assertEquals(expectedTarget.getPurpose(), actualTarget.getPurpose());
+
```
dconeybe, line 1143: ultranit: Can these blank line be removed, to put the check for getPurpose() right beside the other assertEquals() lines?

dconeybe (approved):
LGTM! Thanks for making this improvement to the spec tests. I just left two minor comments.

PR #4775, opened 2023-03-13
Add Firebase Sessions Test App

mrober:
Add Firebase Sessions Test App. This app will be used for local development, and eventually for CI automated tests.


firebase-sessions/test-app/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
visumickey, line 1: If we are planning to not have proguarding enabled, do we even need to have this file?

firebase-sessions/test-app/src/main/kotlin/com/google/firebase/testing/sessions/MainActivity.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import com.google.firebase.sessions.FirebaseSessions
+
+class MainActivity : AppCompatActivity() {
+  @SuppressLint("DiscouragedApi")
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_main)
+
+    findViewById<TextView>(R.id.greeting_text).text = FirebaseSessions.instance.greeting()
```
visumickey, line 31: Can we have some other random text here? Is the intent to refer to FIrebaseSessions to verify if the integration is working?

My concern with this is that some auto initializing FirebaseSessions functionality could not be verified.

firebase-sessions/test-app/src/test/kotlin/com/google/firebase/testing/sessions/ExampleUnitTest.kt
```diff
@@ -0,0 +1,33 @@
...
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+class ExampleUnitTest {
+  @Test
+  fun addition_isCorrect() {
+    assertEquals(4, 2 + 2)
+  }
+}
```
visumickey, line 33: Do we need these boiler plates?

firebase-sessions/test-app/test-app.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  compileOptions {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+  }
+  kotlinOptions { jvmTarget = "1.8" }
+}
+
+dependencies {
+  implementation(project(":firebase-sessions"))
```
visumickey, line 48: Let's add TODO to say that we need to remove this thicker dependency once we have the test app depending on Crashlytics and/or Fireperf.

Something we did in iOS: We have the same app with different flavors - App with Crashlytics and no Fireperf, App with Fireperf and no Crashlytics, App with both. If we have to do something similar, can we do that here as well?

firebase-sessions/test-app/test-app.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+    minSdk = 16
+    targetSdk = 33
+    versionCode = 1
+    versionName = "1.0"
+    multiDexEnabled = true
+  }
+  buildTypes {
+    release {
+      isMinifyEnabled = false
+      proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
```
visumickey, line 37: Do we need this for our test app?

visumickey (approved):
Thanks for the test app. This is a great way to validate our integrations.

Should we merge in the WebP files? Even if it is a boiler plate, can we skip them?

Curious question: I want to learn how the SDK behaves in a Jetpack Compose App. How can we achieve that? Should we need a separate app or just make a new app target which uses Jetpack compose?

firebase-sessions/test-app/proguard-rules.pro
mrober: Removed.

firebase-sessions/test-app/src/main/kotlin/com/google/firebase/testing/sessions/MainActivity.kt
mrober: Yeah this was for local testing. Not calling it explicitly makes sense for when this is a CI test, I removed it. I have verified the auto initializing locally by seeing the message in logcat.

firebase-sessions/test-app/test-app.gradle.kts
mrober: Done.

PR #4778, opened 2023-03-13
Add JSON encoder for SessionEvents

mrober:
Add JSON encoder for `SessionEvent`s.

Also moved the logic to construct a Session Start event out of the data class, this is to make the data class cleaner and easier to read.

### DIFF
--- a/firebase-sessions/firebase-sessions.gradle.kts
+++ b/firebase-sessions/firebase-sessions.gradle.kts
@@ -42,11 +42,10 @@ android {
 }
 
 dependencies {
-  implementation(project(":encoders:firebase-encoders"))
+  implementation("com.google.firebase:firebase-encoders:17.0.0")
   implementation(project(":encoders:firebase-encoders-json"))
-  implementation(project(":firebase-common"))
-  implementation(project(":firebase-common:ktx"))
-  implementation(project(":firebase-components"))
+  implementation("com.google.firebase:firebase-common-ktx:20.3.1")
+  implementation("com.google.firebase:firebase-components:17.1.0")
   implementation(libs.androidx.annotation)
 
   testImplementation(libs.androidx.test.junit)
--- a/firebase-sessions/firebase-sessions.gradle.kts
+++ b/firebase-sessions/firebase-sessions.gradle.kts
@@ -42,12 +42,17 @@ android {
 }
 
 dependencies {
+  implementation(project(":encoders:firebase-encoders"))
+  implementation(project(":encoders:firebase-encoders-json"))
   implementation(project(":firebase-common"))
   implementation(project(":firebase-common:ktx"))
   implementation(project(":firebase-components"))
   implementation(libs.androidx.annotation)
 
+  testImplementation(libs.androidx.test.junit)
+  testImplementation(libs.androidx.test.runner)
   testImplementation(libs.junit)
+  testImplementation(libs.robolectric)
   testImplementation(libs.truth)
 
   androidTestImplementation(libs.androidx.test.junit)
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -40,8 +40,8 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
   fun greeting(): String = "Matt says hi!"
 
   private fun initiateSessionStart() {
-    val sessionState = sessionGenerator.generateNewSession()
-    val sessionEvent = SessionEvents.sessionStart(sessionState)
+    val sessionDetails = sessionGenerator.generateNewSession()
+    val sessionEvent = SessionEvents.startSession(sessionDetails)
 
     Log.i(TAG, "Initiate session start: $sessionEvent")
   }
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -41,7 +41,7 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
 
   private fun initiateSessionStart() {
     val sessionState = sessionGenerator.generateNewSession()
-    val sessionEvent = SessionEvent.sessionStart(sessionState)
+    val sessionEvent = SessionEvents.sessionStart(sessionState)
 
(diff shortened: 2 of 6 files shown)


firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -42,12 +42,17 @@ android {
 }
 
 dependencies {
+  implementation(project(":encoders:firebase-encoders"))
+  implementation(project(":encoders:firebase-encoders-json"))
   implementation(project(":firebase-common"))
```
davidmotson, line 47: Since we merged pinned deps on master, it'd be worthwhile to convert these to pinned deps.

davidmotson (approved):

firebase-sessions/firebase-sessions.gradle.kts
mrober: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -41,7 +41,7 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
 
   private fun initiateSessionStart() {
     val sessionState = sessionGenerator.generateNewSession()
-    val sessionEvent = SessionEvent.sessionStart(sessionState)
+    val sessionEvent = SessionEvents.sessionStart(sessionState)
```
visumickey, line 44: A couple of minor nitpicks (outside the scope of this PR)

1. Let's rename "sessionState" with "sessionDetails"
2. Can we rename the API `sessionStart` to `startSession`? - Just finding it more readable.

Very minor nitpicks - Feel free to do this in another PR (if that makes it clean for you).

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -41,7 +41,7 @@ class FirebaseSessions internal constructor(firebaseApp: FirebaseApp) {
...
 
     Log.i(TAG, "Initiate session start: $sessionEvent")
```
visumickey, line 46: Curious on Kotlin capability - Just printing `$sessionEvent` would parse through the whole structure? If so, this is just cool!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
```diff
@@ -29,23 +31,10 @@ internal data class SessionEvent(
...
-            sessionDetails.sessionIndex,
-          ),
-      )
-  }
-}
+)
 
 /** Enum denoting all possible session event types. */
-internal enum class EventType(val number: Int) {
+internal enum class EventType(override val number: Int) : NumberedEnum {
```
visumickey, line 37: Why `override`?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
```diff
@@ -0,0 +1,67 @@
...
+ * Contains functions for [SessionEvent]s.
+ *
+ * @hide
+ */
+internal object SessionEvents {
+  /** JSON [DataEncoder] for [SessionEvent]s. */
+  // TODO(mrober): Replace with firebase-encoders-processor when it can encode Kotlin data classes.
+  internal val SESSION_EVENT_ENCODER: DataEncoder =
+    JsonDataEncoderBuilder()
+      .configureWith {
```
visumickey, line 34: How does it work for hierarchical fields? Are these encoders just for the KotlinData classes? As we encode the fields, they use these encoders to package the fields?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionEventEncoderTest.kt
```diff
@@ -0,0 +1,73 @@
...
+package com.google.firebase.sessions
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.google.common.truth.Truth.assertThat
+import com.google.firebase.sessions.SessionEvents.SESSION_EVENT_ENCODER
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class SessionEventEncoderTest {
```
visumickey, line 26: Should we have tests for empty values? How would the encoder behave incase of `NULL` values versus `0` value (or) `empty` values for fields?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: Yup! It does the toString for us, including calling toString on all the nested fields.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
mrober: This is some pretty nice Kotlin syntax. Imagine it like the interface were Java and had a method `int getNumber();` then this enum implemented it and overrode the getter method and returned the member variable.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
mrober: There are encoders for every type we need to encode. This will be replaced by an annotation processor that will generate this for us after Acore updates firebase-encoders-processor.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionEventEncoderTest.kt
mrober: So far none of the fields are nullable, but I added test cases for empty string and 0.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: 1. I missed this in my last PR, fixed!
2. Done. Do you see it like a verb, "do start the session" ? I think I saw it as a noun "this is the session start".

visumickey: yeah! I go with the "Verb on the Noun" making it read like a sentence.

visumickey (approved):

PR #4782, opened 2023-03-14
Remove common dependency in AppCheck Interop

emilypgoogle:
AppCheck Interop had a common dependency and was using it to return `FirebaseException`s in `AppCheckTokenResult`. Going forward with third party SDK integration with Maps or similar, binary size is a large concern, and this change will allow SDK consumers to implement AppCheck integration with minimal included binary size moving forward. It is currently the plan to keep AppCheck Interop as a non-public SDK, only being exposed through third party SDKs.

AppCheck Interop is currently a non-public facing SDK, and all uses of the method being changed are in tests, so the breaking change does not require an API bump or a major release bump.

There has been the suggestion that if this interface is going to be used by Maps, it potentially should have a name other than `InternalAppCheckTokenProvider`, but a change like that would require sweeping changes, so it hasn't been included here.


rosalyntan (commented):
Looks like there are a few failing tests -- can you take a look at those?

Also re: "AppCheck Interop is currently a non-public facing SDK, and all uses of the method being changed are in tests, so the breaking change does not require an API bump or a major release bump."

Just to check, have you verified that Firebase SDK usages of the App Check Interop (i.e. Storage, Firestore, etc.) also do not rely on `FirebaseException` being returned from the interop?

emilypgoogle:
@rosalyntan 
> Just to check, have you verified that Firebase SDK usages of the App Check Interop (i.e. Storage, Firestore, etc.) also do not rely on `FirebaseException` being returned from the interop?

The method's uses are all null checks or rethrows, so nothing relies on the backing type past implementing `Throwable`, however the signature will have changed.

In regards to the version bump situation, however, after a handful of conversations it seems like the best option is to bump appcheck and interop to a new major, and move the 3 dependent SDKs (functions, firestore, storage) up a minor to depend on the new version. Current plan is to do this in M131 for IO. I've got the version bumps ready in the PR as I understand them but would appreciate a sanity check.

This does however give us the possibility to rename `InternalAppCheckTokenProvider` to something else since we're already breaking, I believe Victor in the past suggested simply dropping internal for `AppCheckTokenProvider`. The benefit is this type is going to be visible in the Maps API surface and would result in a slightly less peculiar name, though ultimately the user is expected to be passing AppCheck as a whole regardless and documentation will reflect this, so it's not an absolutely necessary change.

What do you think about the versioning and name change?

appcheck/firebase-appcheck-debug-testing/gradle.properties
```diff
@@ -1,2 +1,2 @@
-version=16.1.3
+version=17.0.0
```
rosalyntan, line 1: Should the SafetyNet and PlayIntegrity App Check SDKs also be bumped? I think we were generally trying to keep the version numbers aligned.

firebase-firestore/firebase-firestore.gradle
```diff
@@ -126,7 +126,7 @@ dependencies {
     implementation project(':protolite-well-known-types')
     implementation 'com.google.firebase:firebase-database-collection:18.0.1'
     implementation 'com.google.firebase:firebase-components:17.1.0'
-    implementation 'com.google.firebase:firebase-appcheck-interop:16.1.1'
+    implementation project(':appcheck:firebase-appcheck-interop')
```
rosalyntan, line 129: RTDB also has a dependency on the App Check interop, so will need to be bumped as well.

In addition, App Check integration was also just added to Firebase Auth, so we'll need to add this change to Auth as well once the M130 release goes out. cc @prameshj

rosalyntan:
> In regards to the version bump situation, however, after a handful of conversations it seems like the best option is to bump appcheck and interop to a new major, and move the 3 dependent SDKs (functions, firestore, storage) up a minor to depend on the new version. Current plan is to do this in M131 for IO. I've got the version bumps ready in the PR as I understand them but would appreciate a sanity check.

Version bumps sound good, though I left a couple comments in the PR. 

> This does however give us the possibility to rename InternalAppCheckTokenProvider to something else since we're already breaking, I believe Victor in the past suggested simply dropping internal for AppCheckTokenProvider. The benefit is this type is going to be visible in the Maps API surface and would result in a slightly less peculiar name, though ultimately the user is expected to be passing AppCheck as a whole regardless and documentation will reflect this, so it's not an absolutely necessary change.

I think re-naming is okay, though I think there is some benefit in making it clear that the interop library is not intended for public consumption. Perhaps instead of just AppCheckTokenProvider, it can be InteropAppCheckTokenProvider? Also can you clarify what you mean by this type being visible in the Maps API surface? Will developers using the Maps API have to interact with the interop types directly?

Also @malcolmdeck for thoughts on this, since I'm not sure who the new owner of the Android App Check SDK will be.

appcheck/firebase-appcheck-debug-testing/gradle.properties
emilypgoogle: They seemed behind the last minor so I wasn't sure, bumped them as well

emilypgoogle:
I agree on `InteropAppCheckTokenProvider`
> Will developers using the Maps API have to interact with the interop types directly?
They won't need to perform any calls on type types, I believe the extent of the interaction is passing a real instance of AppCheck to a method something like `Maps.registerAppCheckTokenProvider(InteropAppCheckTokenProvider provider)`

appcheck/firebase-appcheck-interop/src/main/java/com/google/firebase/appcheck/AppCheckTokenResult.java
```diff
@@ -16,13 +16,12 @@
...
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import com.google.firebase.FirebaseException;
 
 /**
  * Class to hold the result emitted by a Firebase App Check service application verification
  * attempt. The result will always contain a token, which will either be a valid raw JWT attesting
- * application identity, or a dummy value. The result may optionally contain a {@link
- * FirebaseException} if application verification does not succeed.
+ * application identity, or a dummy value. The result may optionally contain a {@link Exception} if
```
malcolmdeck, line 23: Nit: "... contain an {@link Exception} if"

malcolmdeck:
I think that we may want to run the Interop class naming by the API Council (even if it's just an email rather than formal review). It's not the hugest deal since developers won't actually need to interact with that Interface, but if we're making a new symbol public it's probably worth making sure that we get it right, especially since whatever we put in Maps is likely to be the name for all future external integrators (of which we will expect there to be many).

firebase-firestore/firebase-firestore.gradle
emilypgoogle: RTDB included and I have a bug tracking Auth for M131

emilypgoogle:
> I think that we may want to run the Interop class naming by the API Council (even if it's just an email rather than formal review).

Got an LGTM and no dissenting opinions over a couple days, looks like we're in the clear for this, so I went through with the name change. Currently have 1 integration test failing that I'm looking into, not sure if it's just flaky though.

appcheck/firebase-appcheck-safetynet/gradle.properties
```diff
@@ -1,2 +1,2 @@
-version=16.1.3
+version=17.0.0
```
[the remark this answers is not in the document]
emilypgoogle: After speaking to @rlazo I've reverted this version bump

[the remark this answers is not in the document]
rlazo: +1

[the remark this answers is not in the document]
rlazo: Technically, since it's deprecated, it shouldn't belong to any library group :-)

PR #4783, opened 2023-03-14
Support JPGs (in addition to PNGs) as feedback screenshots and set correct content-type/filename

lfkellogg:


firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
```diff
@@ -54,6 +72,85 @@ private Task<String> attachScreenshot(String feedbackName, @Nullable Uri screens
...
+            /* selection= */ null,
+            /* projectionArgs= */ null,
+            /* sortOrder= */ null)) {
+      if (returnCursor == null) {
+        LogWrapper.w(
+            TAG,
+            String.format(
+                "Unable to get filename from URI '%s', using default '%s'",
+                contentUri, DEFAULT_FILENAME));
+        return DEFAULT_FILENAME;
```
kaibolay, line 129: This should probably be based on the content type (if we were able to figure that out), so have a default filename for each (supported) content type.

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
```diff
@@ -54,6 +72,85 @@ private Task<String> attachScreenshot(String feedbackName, @Nullable Uri screens
...
+      String name = returnCursor.getString(nameIndex);
+      return name;
+    } catch (Exception e) {
+      LogWrapper.e(
+          TAG,
+          String.format(
+              "Error getting filename from URI '%s', using default '%s'",
+              contentUri, DEFAULT_FILENAME),
+          e);
+      return DEFAULT_FILENAME;
```
kaibolay, line 142: ditto

lfkellogg: Done

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
```diff
@@ -54,6 +71,95 @@ private Task<String> attachScreenshot(String feedbackName, @Nullable Uri screens
...
+  }
+
+  private String getContentType(Uri screenshotUri) {
+    String contentType = null;
+    if (screenshotUri.getScheme().equals("file")) {
+      contentType = getContentTypeForFilename(screenshotUri.getLastPathSegment());
+    } else if (screenshotUri.getScheme().equals("content")) {
+      contentType = contentResolver.getType(screenshotUri);
+    }
+    if (contentType == null) {
```
kaibolay, line 97: nit: I'd flip the condition:

```
  if (contentType != null} {
    return contentType;
  }
  LogWrapper...
  return CONTENT_TYPE_PNG
```

kaibolay (approved):

firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/FeedbackSender.java
lfkellogg: Done.

PR #4785, opened 2023-03-15
Use the private setter pattern on SessionGenerator.currentSession

mrober:
This is more Kotlin idiomatic. Also made some properties val instead of var, I think this is cleaner.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
@@ -36,35 +36,25 @@ internal data class SessionDetails(
  *
  * @hide
  */
-internal class SessionGenerator(private var collectEvents: Boolean) {
-  private var firstSessionId = ""
-  private var sessionIndex: Int = -1
+internal class SessionGenerator(private val collectEvents: Boolean) {
+  private val firstSessionId = generateSessionId()
+  private var sessionIndex = 0
 
-  private var thisSession: SessionDetails =
-    SessionDetails(
-      sessionId = "",
-      firstSessionId = "",
-      collectEvents,
-      sessionIndex,
-    )
+  var currentSession =
+    SessionDetails(sessionId = "", firstSessionId = "", collectEvents, sessionIndex = -1)
+    private set
 
-  // Generates a new Session ID. If there was already a generated Session ID
-  // from the last session during the app's lifecycle, it will also set the last Session ID
+  /** Generates a new session. The first session's sessionId will match firstSessionId. */
   fun generateNewSession(): SessionDetails {
-    val newSessionId = UUID.randomUUID().toString().replace("-", "").lowercase()
-
-    // If firstSessionId is set, use it. Otherwise set it to the
-    // first generated Session ID
-    firstSessionId = firstSessionId.ifEmpty { newSessionId }
-
-    sessionIndex += 1
-
-    thisSession =
-      SessionDetails(sessionId = newSessionId, firstSessionId, collectEvents, sessionIndex)
-
-    return thisSession
+    currentSession =
+      SessionDetails(
+        sessionId = if (sessionIndex == 0) firstSessionId else generateSessionId(),
+        firstSessionId,
+        collectEvents,
+        sessionIndex++,
+      )
+    return currentSession
   }
 
-  val currentSession: SessionDetails
-    get() = thisSession
+  private fun generateSessionId() = UUID.randomUUID().toString().replace("-", "").lowercase()
 }
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
@@ -36,25 +36,29 @@ internal data class SessionDetails(
  *
  * @hide
  */
-internal class SessionGenerator(private val collectEvents: Boolean) {
+internal class SessionGenerator(
+  private val collectEvents: Boolean,
+  private val uuidGenerator: () -> UUID = UUID::randomUUID
+) {
   private val firstSessionId = generateSessionId()
-  private var sessionIndex = 0
+  private var sessionIndex = -1
 
-  var currentSession =
-    SessionDetails(sessionId = "", firstSessionId = "", collectEvents, sessionIndex = -1)
+  /** The current generated session, must not be accessed before calling [generateNewSession]. */
+  lateinit var currentSession: SessionDetails
     private set
 
   /** Generates a new session. The first session's sessionId will match firstSessionId. */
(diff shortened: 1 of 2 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -36,35 +36,25 @@ internal data class SessionDetails(
...
 
-  private var thisSession: SessionDetails =
-    SessionDetails(
-      sessionId = "",
-      firstSessionId = "",
-      collectEvents,
-      sessionIndex,
-    )
+  var currentSession =
+    SessionDetails(sessionId = "", firstSessionId = "", collectEvents, sessionIndex = -1)
```
mrober, line 44: Do we still need this default instance? Would it make sense to make currentSession nullable or lateinit?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -36,35 +36,25 @@ internal data class SessionDetails(
...
+        firstSessionId,
+        collectEvents,
+        sessionIndex++,
+      )
+    return currentSession
   }
 
-  val currentSession: SessionDetails
-    get() = thisSession
+  private fun generateSessionId() = UUID.randomUUID().toString().replace("-", "").lowercase()
```
mrober, line 59: It might not be worth it, but if we made this an interface like `SessionIdGenerator` and passed in, we could make the unit tests deterministic. Then have a seperate test to validate the generated session ids. What do you think?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
samedson: It mostly depends on preference. Do you want to deal with nulls in the classes that use this, or assume that what you get is valid? 

I think we've designed this class to always return a valid currentSession, so I lean towards keeping it not nullable.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
samedson: I like that idea. 2 suggestions:
 - You can pass in a function to the constructor instead of making a SessionIdGenerator class with a generate() method. Only reason is a function might be lighter weight.
  - This class / function that we pass in should only return the `UUID.randomUUID()` part of this statement. This `SessionGenerator` should be responsible for toString and `.replace`ing it so we can assert that the dashes are removed in tests.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -36,35 +36,25 @@ internal data class SessionDetails(
...
-    thisSession =
-      SessionDetails(sessionId = newSessionId, firstSessionId, collectEvents, sessionIndex)
-
-    return thisSession
+    currentSession =
+      SessionDetails(
+        sessionId = if (sessionIndex == 0) firstSessionId else generateSessionId(),
+        firstSessionId,
+        collectEvents,
+        sessionIndex++,
```
samedson, line 54: I think modifying a value while it's being passed in is a little too sneaky for me. I would rather we put this on a separate line. For me, it makes it hard to reason about what `sessionIndex` currently is. Eg. on line 51 we check if sessionIndex is 0.  I know that happens before this increments, but if we reorder the arguments we could break this.

I would suggest we increment this before we construct SessionDetails, and initialize it to `-1`. WDYT?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -36,35 +36,25 @@ internal data class SessionDetails(
...
+      )
```
samedson, line 55: Should we name these arguments with `argument = ` or is that too pedantic because the variable names match?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
visumickey: I don't think we need a default instance. That does not serve anything. SessionGenerator should generate session only when there is a need.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -36,35 +36,25 @@ internal data class SessionDetails(
...
-  private var thisSession: SessionDetails =
-    SessionDetails(
-      sessionId = "",
-      firstSessionId = "",
-      collectEvents,
-      sessionIndex,
-    )
+  var currentSession =
+    SessionDetails(sessionId = "", firstSessionId = "", collectEvents, sessionIndex = -1)
+    private set
```
visumickey, line 45: Is this not associated with any var/val? Felt like a hanging `private set`.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
visumickey: A separate interface just for sessionId sounds like an overkill. Do you think this structure is not testable? I think with the SessionDetails structure and generateNewSession, we should be able to cover all scenarios for testing.

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
mrober: Done.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
mrober: I like the `argument = ` syntax when it's not clear what the value means, like `SessionGenerator(collectEvents = false)` is so much more readable than just `SessionGenerator(false)`. If we need to say like `Session(sessionId = sessionId)` that would just be silly.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
mrober: I was concerned about the randomness in the tests, and asserts on individual fields of the SessionDetails data class to avoid a big assert on the entire object since it contained randomness.

I played around with this. I made it take a function, not an interface, in the ctor. I left one test with the random UUIDs to verify it generates valid session ids.

samedson (approved):
LGTM!

Issue #4766, opened 2023-03-10
Crash in Binary XML file line #31 layout/modal_portrait_inner: Error inflating class android.widget.ImageView.


vitaliygerasymchuk:
### [REQUIRED] Step 2: Describe your environment

        * Android Studio version: 2022.1.1 Patch 2
        * Firebase Component: _____ firebase-bom:31.2.3, crashlytics, analytics, messaging, firebase-inappmessaging-display
        * Component version: 31.2.3

### [REQUIRED] Step 3: Describe the problem
The Crashlytics contains multiple crashes of this type.
![image](https://user-images.githubusercontent.com/31439136/224250303-2c535f2b-8c57-4391-a0f8-5353fecc2ab5.png)

#### Steps to reproduce:
Caused by android.content.res.Resources$NotFoundException: Drawable (missing name) with resource ID #0x7f0700b1
Caused by android.content.res.Resources$NotFoundException: Unable to find resource ID #0x7f0700b1

#### Relevant Code:
This crash event happened in the first 3 seconds of the user's session. The crash event for this issue happened 3 times per user on average. Internal issue. No code.

argzdev:
Hi, @vitaliygerasymchuk. Thanks for reaching out. Looking at the stacktrace, I don't see any indication of Firebase. Can you tell us why you think Firebase is causing the issue here?

vitaliygerasymchuk:
Hi @argzdev. I probably missed to put additional logs. So the crash is caused here:
com.google.firebase.inappmessaging.display.internal.bindingwrappers.ModalBindingWrapper line 31 InAppMessageLayoutConfig. 

Here's the title of the crashs: ModalBindingWrapper.inflate
android.content.res.Resources$NotFoundException - Unable to find resource ID #<address>

Original StackTrace:

`Fatal Exception: android.view.InflateException: Binary XML file line #13 in com.example.test:layout/modal: Binary XML file line #31 in com.example.test:layout/modal_portrait_inner: Error inflating class android.widget.ImageView
`
`Caused by java.lang.reflect.InvocationTargetException:
       at java.lang.reflect.Constructor.newInstance0(Constructor.java)
       at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
       at android.view.LayoutInflater.createView(LayoutInflater.java:852)
       at android.view.LayoutInflater.createView(LayoutInflater.java:774)
[...]

PR #4797, opened 2023-03-17
Adding the resources to main drawable folder instead of the debug/drawable

argzdev:
Proposed fix for #4766.

The drawable `placeholder_550_350.png` being referenced in `@+id/image_view` of `modal_portrait_inner.xml` is in the `debug/res/drawable-mdpi/` folder. This should be placed in the drawable folder of the main directory not just the debug folder, since this drawable is being referenced in the release build.

P.S. I'm wondering if we should just remove the debug/res/drawable folder, so all resources are referenced in the main directory(?)


eldhosembabu:
Thanks for the PR..the resources moved are referenced using "tools" namespace and so not sure whether this could be the right fix.

So i think its better to reproduce the issue first and verify with this fix before merging.

PR #4799, opened 2023-03-18
Firestore: QueryTest.java: add a test that resumes a query with existence filter

dconeybe:
This is a port of https://github.com/firebase/firebase-js-sdk/pull/7134


milaGGL (approved):
LGTM

PR #4801, opened 2023-03-20
Populate firebaseInstallationId field in session

mrober:
Populate `firebaseInstallationId` field in session data.
Add `SessionCoordinator` class and plumb in a `CoroutineDispatcher`.
Setup coroutines including testing.

### DIFF
--- a/firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionCoordinatorTest.kt
+++ b/firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionCoordinatorTest.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionCoordinatorTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(AndroidJUnit4::class)
+class SessionCoordinatorTest {
+  @Test
+  fun attemptLoggingSessionEvent_populatesFid() = runTest {
+    val sessionCoordinator =
+      SessionCoordinator(
+        firebaseInstallations = FakeFirebaseInstallations("FaKeFiD"),
+        backgroundDispatcher = StandardTestDispatcher(testScheduler),
```
mrober, line 37: `:integ-testing` doesn't have any test `CoroutineDispatcher`s to work well with `runTesting`.

samedson (approved):
This looks good!

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionCoordinatorTest.kt
mrober: `:integ-testing` is to provide executors for testing that mimic the executors provided by `firebase-common` with policies e.g. don't do blocking network requests on Background. See https://firebase.github.io/firebase-android-sdk/components/executors/#testing

So these tests don't mimic the Background policy. I tried to use `TestOnlyExecutors.background().asCoroutineDispatcher()` when I first wrote this test, but then I couldn't call `runCurrent()` to wait for the suspend functions to finish before doing the assert.

davidmotson (approved):

PR #4806, opened 2023-03-21
feat(appcheck): create FirebaseAppCheck.tokenChanges Flow

thatfiredev:
Adding the Flow approved in [go/kotlin-flows-firebase-android-addendum](http://goto.google.com/kotlin-flows-firebase-android-addendum).

Note that I've used `.buffer(capacity = Channel.CONFLATED)` in this one to handle back-pressure. This will drop older values, which I assume are not relevant because developers might only be interested in the newest `AppCheckToken` emitted by the `Flow`.


rlazo (approved):

rlazo (changes requested):
If you plan to merge it now, it has to be done also in non-ktx version of the file

PR #4808, opened 2023-03-21
Improve real-time reference docs.

danasilver:
Internal b/274648602.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/ConfigUpdateListener.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/ConfigUpdateListener.java
@@ -18,8 +18,9 @@ import androidx.annotation.NonNull;
 import javax.annotation.Nonnull;
 
 /**
- * Event listener interface for real-time RC updates. Implement {@code ConfigUpdateListner} to call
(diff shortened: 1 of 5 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/ConfigUpdateListener.java
```diff
@@ -32,7 +33,7 @@ public interface ConfigUpdateListener {
   void onUpdate(@NonNull ConfigUpdate configUpdate);
 
   /**
-   * Callback for when an error occurs while listening or fetching a config update.
+   * Callback for when an error occurs while listening for or fetching a config update.
```
qdpham13, line 36: Maybe `Callback for when an error occurs when listening for updates of fetching the latest config.`

qdpham13 (approved):
Had just one suggestion but LGTM otherwise

firebase-config/src/main/java/com/google/firebase/remoteconfig/ConfigUpdateListener.java
danasilver: Good suggestion! Added "version" too.

PR #4812, opened 2023-03-22
Remove a call to UAST inline function

rlazo:


jsjeon (approved):
Thanks for the upstream change!

PR #4813, opened 2023-03-23
Firestore: QueryTest.java: improve the test that resumes a query with existence filter to actually validate the existence filter.

dconeybe:
This PR improves the test resumingAQueryShouldUseExistenceFilterToDetectDeletes() in QueryTest.java, which was added by the previous PR https://github.com/firebase/firebase-android-sdk/pull/4799, to _also_ validate that the existence filter sent from Watch was correct. This was achieved by adding some test-only "hooks" into the SDK's internals that enable the test to "listen" for existence filter mismatches. This is a port of https://github.com/firebase/firebase-js-sdk/pull/7149.


firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregatorTestingHooks.java
```diff
@@ -0,0 +1,110 @@
...
+  interface ExistenceFilterMismatchListener {
+    @AnyThread
+    void onExistenceFilterMismatch(ExistenceFilterMismatchInfo info);
+  }
+
+  @AutoValue
+  abstract static class ExistenceFilterMismatchInfo {
+
+    static ExistenceFilterMismatchInfo create(int localCacheCount, int existenceFilterCount) {
+      return new AutoValue_WatchChangeAggregatorTestingHooks_ExistenceFilterMismatchInfo(
```
milaGGL, line 93: is `AutoValue_WatchChangeAggregatorTestingHooks_ExistenceFilterMismatchInfo` a class? Is it generated by the AutoValue library? why we wan to use it here???

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregatorTestingHooks.java
```diff
@@ -0,0 +1,110 @@
...
+import androidx.annotation.VisibleForTesting;
+import com.google.auto.value.AutoValue;
+import com.google.firebase.firestore.ListenerRegistration;
+import com.google.firebase.firestore.util.Executors;
+import java.util.HashMap;
+import java.util.Map;
+
+final class WatchChangeAggregatorTestingHooks {
+
+  private WatchChangeAggregatorTestingHooks() {}
```
milaGGL, line 30: in web, the TestingHooks class used singleton design. Here, I assume, it is done by making the constructor private and accessing the existenceFilterMismatchListeners via WatchChangeAggregatorTestingHooksAccessor. Why we want to use different designs in 2 SDKs?

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregatorTestingHooks.java
dconeybe: Yes, `AutoValue_WatchChangeAggregatorTestingHooks_ExistenceFilterMismatchInfo` is indeed a class that is generated by the "AutoValue" library: https://github.com/google/auto/blob/main/value/userguide/index.md

AutoValue is an idomatic way to create "data classes" in Java at Google. It saves writing the boilerplate of setters, getters, builders, etc. It does indeed look like overkill here since the class only contains two integers; however, in the mila/BloomFilter branch it makes a little more sense because it also contains the bloom filter parameters: https://github.com/firebase/firebase-android-sdk/blob/971632689bc363743b75c6cf6baf25bdd0317397/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregatorTestingHooks.java#L128-L164

We also use AutoValue in a few places in the Firestore Android SDK, such as here: https://github.com/firebase/firebase-android-sdk/blob/687e079d401db1a75d2559038b879726a476ca3e/firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndex.java#L36-L37 and https://github.com/firebase/firebase-android-sdk/blob/687e079d401db1a75d2559038b879726a476ca3e/firebase-firestore/src/main/java/com/google/firebase/firestore/model/FieldIndex.java#L68-L69

firebase-firestore/src/main/java/com/google/firebase/firestore/remote/WatchChangeAggregatorTestingHooks.java
dconeybe: We definitely _could_ use a singleton "Testing Hooks" class here in Android as well. Here in Android I chose to use `static` methods rather than expose a singleton instance, which are effectively the same thing. I felt like using `static` methods in Java was more idiomatic. I didn't think that consistency of the design of testing hooks across the SDKs was terribly important because each language has its own way of doing that sort of thing.

All that being said, I'll see what I can do to make it more consistent with the Web SDK and see what you think. Maybe it will be better? I honestly didn't try.

milaGGL (approved):
LGTM

PR #4816, opened 2023-03-23
Migrate Database to KTS + VC

daymxn:
Per [b/274927706](https://b.corp.google.com/issues/274927706),

This migrates all the `database` build scripts from Groovy to Kotlin, and from the old ext based common dep versioning to the new Version Catalog based versioning.

More specifically, this migrates the following SDKs:

- `firebase-database`
- `firebase-database-ktx`
- `firebase-database-collection`

This also fixes a minor bug with `IntegrationTestValues` that causes the IDE to freak out regarding resources populated at runtime not being present.


firebase-database/firebase-database.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+dependencies {
+  implementation("com.google.firebase:firebase-common:20.3.1")
+  implementation("com.google.firebase:firebase-components:17.1.0")
+  implementation("com.google.firebase:firebase-appcheck-interop:16.1.1")
+  implementation("com.google.firebase:firebase-database-collection:18.0.1")
+
+  implementation("androidx.annotation:annotation:1.1.0")
+  implementation("com.google.android.gms:play-services-base:18.0.1")
+  implementation("com.google.android.gms:play-services-tasks:18.0.1")
+  implementation("com.google.android.gms:play-services-basement:18.1.0")
```
rlazo, line 61: These also exist in the version catalog

rlazo: Since it depends on common, and common depends on 

```
    implementation(libs.playservices.basement)
    implementation(libs.playservices.tasks)
```

those are already in the latest version.

So the only question would be base.

daymxn: Hmm, I'll run some tests tomorrow and peep if upgrading everything breaks anything (or find out what _we can_ upgrade). If not, it'll just be a minor bump anyways; but I'll ping db folk to make sure they're alright with it. Thoughts?

rlazo: I think it's OK to just bump those, and then wait until database has a meaningful change to do the release. 

The last release of DB is [20.1.0](https://maven.google.com/web/index.html?q=firebase-co#com.google.firebase:firebase-database:20.1.0) and it depends in `firebase-common:20.2.0`,  which is outdated anyway.

rlazo: In case you want to take a look, here's the list of deps from a kotlin app that only depends on db

https://gist.github.com/rlazo/55e8fc9f331dfb4083e6ad0b2bb9f809

daymxn: SGTM :)

firebase-database/firebase-database.gradle.kts
```diff
@@ -50,18 +50,14 @@ android {
...
+  implementation("com.google.firebase:firebase-common:20.3.2")
   implementation("com.google.firebase:firebase-components:17.1.0")
+  implementation("com.google.firebase:firebase-auth-interop:20.0.0")
   implementation("com.google.firebase:firebase-appcheck-interop:16.1.1")
   implementation("com.google.firebase:firebase-database-collection:18.0.1")
 
-  implementation("androidx.annotation:annotation:1.1.0")
-  implementation("com.google.android.gms:play-services-base:18.0.1")
-  implementation("com.google.android.gms:play-services-tasks:18.0.1")
-  implementation("com.google.android.gms:play-services-basement:18.1.0")
```
rlazo: Even though we can get this deps transitively from common, I think is better to leave them here. But using the version catalog

gradle/libs.versions.toml
```diff
@@ -58,4 +58,5 @@ kotest-property = { module = "io.kotest:kotest-property-jvm", version.ref = "kot
 quickcheck = { module = "net.java:quickcheck", version.ref = "quickcheck" }
 
 [bundles]
-kotest = ["kotest-runner", "kotest-assertions", "kotest-property"]
\ No newline at end of file
+kotest = ["kotest-runner", "kotest-assertions", "kotest-property"]
+playservices = ["playservices-base", "playservices-basement", "playservices-tasks"]
```
rlazo, line 62: Given that common does depend on base and tasks, but not basements, I think there's a point to be made to not group them all together.

firebase-database/firebase-database.gradle.kts
daymxn: They actually are here still, I just moved them to a `bundle`:
```toml
playservices = ["playservices-base", "playservices-basement", "playservices-tasks"]
```

gradle/libs.versions.toml
daymxn: Folks can still pull dependencies on individual libs, for example:
```toml
implementation(libs.bundles.playservices)
implementation(libs.playservices.base)
```

Additionally, `functions` could take advantage of this bundle- as they depend on all three as well.

rlazo (approved):

PR #4828, opened 2023-03-28
Improve bloom filter application test coverage

milaGGL:
#no-changelog

### DIFF
--- a/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalSerializerTest.java
+++ b/firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalSerializerTest.java
@@ -422,9 +422,8 @@ public final class LocalSerializerTest {
             snapshotVersion,
             limboFreeVersion,
             resumeToken,
-            42);
+            /* expectedCount= */ 1234);
 
-    // Let the RPC serializer test various permutations of query serialization.
     com.google.firestore.v1.Target.QueryTarget queryTarget =
         remoteSerializer.encodeQueryTarget(query.toTarget());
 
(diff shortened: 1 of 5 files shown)


firebase-firestore/src/test/java/com/google/firebase/firestore/local/LocalSerializerTest.java
```diff
@@ -404,6 +404,51 @@ public void testEncodesTargetData() {
     assertEquals(targetData, decoded);
   }
 
+  @Test
+  public void localSerializerDropsExpectedCountInTargetData() {
```
dconeybe, line 408: nit: Please rename this test method slightly to "localSerializerShouldDropExpectedCountInTargetData". This more clearly articulates what the _expected_ behavior is.

firebase-firestore/src/test/java/com/google/firebase/firestore/remote/RemoteEventTest.java
```diff
@@ -456,6 +458,120 @@ public void testExistenceFilterMismatchClearsTarget() {
...
+    TargetChange mapping1 = targetChange(resumeToken, true, null, asList(doc1, doc2), null);
+    assertEquals(mapping1, event.getTargetChanges().get(1));
+
+    TargetChange mapping2 = targetChange(resumeToken, false, null, null, null);
+    assertEquals(mapping2, event.getTargetChanges().get(2));
+
+    // This BloomFilter will return true on both MightContain(doc1) and MightContain(doc2).
+    BitSequence.Builder bitSequence = BitSequence.newBuilder();
+    bitSequence.setPadding(7);
+    bitSequence.setBitmap(ByteString.copyFrom(new byte[] {0x42, (byte) 0xFE}));
```
dconeybe, line 557: How about using 0xFF as the bloom filter bytes, which will "contain" everything? Then you don't need to worry about hand-picking a special bloom filter. If you do this, please update the comment on line 554 to state that it will return true for _all_ documents.

dconeybe (changes requested):

dconeybe (approved):

PR #4841, opened 2023-04-03
Fix workflows for Firebase Sessions

mrober:
Fix workflows for Firebase Sessions

- Use the `googleServices.gradle` file instead of the Google Services plugin directly
- Remove a package level function because it was generating a public class called `ApplicationInfoKt` that was showing up in the api surface
- Add `AndroidManifest.xml` file with `versionName` set to fix android tests
- Update pinned deps


mrober:
The Javadoc failure should be fixed when #4847 lands.

firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -42,11 +42,11 @@ android {
 }
 
 dependencies {
+  implementation("com.google.firebase:firebase-common-ktx:20.3.2")
+  implementation("com.google.firebase:firebase-components:17.1.0")
+  implementation("com.google.firebase:firebase-encoders-json:18.0.1")
```
visumickey, line 47: Changes look good to me here. But curious why we choose pinned versions over a HEAD dependency! Would be a good learning for all of for future.

firebase-sessions/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  package="com.google.firebase.sessions.test"
+  android:versionCode="1"
```
visumickey, line 19: Can we do a dynamic version code here? For eg: number of days since January 2023? I'm trying to see if we can get some dynamicity in the data that is generated especially for the ApplicationInfo proto message.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
```diff
@@ -74,4 +74,18 @@ internal object SessionEvents {
         ),
       applicationInfo = getApplicationInfo(firebaseApp)
     )
+
+  fun getApplicationInfo(firebaseApp: FirebaseApp): ApplicationInfo {
+    val packageName = firebaseApp.applicationContext.packageName
+    val packageInfo = firebaseApp.applicationContext.packageManager.getPackageInfo(packageName, 0)
```
visumickey, line 80: To avoid such a long include, can we import the package to make it look leaner?

firebase-sessions/test-app/test-app.gradle.kts
```diff
@@ -47,3 +46,6 @@ dependencies {
   implementation("androidx.core:core-ktx:1.9.0")
   implementation("com.google.android.material:material:1.8.0")
 }
+
+extra["packageName"] = "com.google.firebase.testing.sessions"
```
visumickey, line 50: Interesting. Why specify the package name here and not in the manifest?

visumickey (approved):

firebase-sessions/firebase-sessions.gradle.kts
mrober: There was an effort across the whole repo to pin dep versions see [go/firebase-android-cross-deps](http://go/firebase-android-cross-deps)

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
mrober: These aren't package imports, these are getters. Like `firebaseApp.getApplicationContext().getPackageManager()`. I can still make it look leaner by adding another val.

PR #4842, opened 2023-04-03
Fix notification on API level 19

lfkellogg:
Adds .png icons for use on platforms that don't support SVG (.xml).


firebase-appdistribution/src/main/res/drawable-anydpi-v24/ic_rate_review.xml
lfkellogg, line 1: This looks different from the original SVG version because I regenerated it as an image and it spit out the 3 PNGs and this XML. I'm using a newer version of Android Studio so I'm guessing this is the newer style.

PR #4847, opened 2023-04-04
Include Kotlin directories in the Dackka plugin

mrober:
Include Kotlin directories in the Dackka plugin.

The `generateDackkaDocumentation` task was not outputting anything on `firebase-sessions` because it was ignoring the `main/kotlin/` src dir, it only looked at `main/java/` before. This was causing some workflows to fail on every PR since sessions is entirely written in Kotlin.


daymxn (approved):
Good catch! We haven't had any `kotlin` directories yet, so this never came up. Our other kotlin files lived under the `java` directory :')

PR #4848, opened 2023-04-04
Make ktx headers h4 instead of h3

rlazo:


daymxn (approved):
Do we care to update previous iterations as well or nah?

PR #4849, opened 2023-04-04
Add Kotlin flow for real-time config updates.

danasilver:


firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
```diff
@@ -45,6 +53,29 @@ fun remoteConfigSettings(
...
+/**
+ * Starts listening for config updates from the Remote Config backend and emits [ConfigUpdate]s via
+ * a [Flow]. See [FirebaseRemoteConfig.addOnConfigUpdateListener] for more information.
+ *
+ * - When the returned flow starts being collected, an [ConfigUpdateListener] will be attached.
+ * - When the flow completes, the listener will be removed. If there are no attached listeners,
+ *   the connection to the Remote Config backend will be closed.
+ */
+fun FirebaseRemoteConfig.configUpdates(): Flow<ConfigUpdate> {
+  return callbackFlow {
```
thatfiredev, line 65: Since `configUpdates` doesn't have any arguments, it could probably be a property instead of a function:

```suggestion
val FirebaseRemoteConfig.configUpdates
  get() = callbackFlow<ConfigUpdate> {
```

firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
```diff
@@ -45,6 +53,29 @@ fun remoteConfigSettings(
...
+    val registration = Firebase.remoteConfig.addOnConfigUpdateListener(object: ConfigUpdateListener {
```
thatfiredev, line 66: Since this is an extension for the `FirebaseRemoteConfig` class, it is already in scope and we don't need to invoke `Firebase.remoteConfig` again.
```suggestion
    val registration = addOnConfigUpdateListener(object: ConfigUpdateListener {
```

firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
```diff
@@ -45,6 +53,29 @@ fun remoteConfigSettings(
...
+      override fun onUpdate(configUpdate: ConfigUpdate) {
+        trySendBlocking(configUpdate)
```
thatfiredev, line 68: Ideally we'd call `trySendBlocking` in a background thread to avoid blocking the main thread. Does the Remote Config SDK have a background executor that can be used here?

For instance, in RTDB we leveraged the existing scheduler: https://github.com/firebase/firebase-android-sdk/blob/c062de8e689b037beed7f5f3a0faafe2bf9fea90/firebase-database/ktx/src/main/kotlin/com/google/firebase/database/ktx/Database.kt#L85

thatfiredev (commented):
@danasilver happy to see a RC Kotlin Flow :)

I know this is WIP, but I wanted to provide some early feedback.

firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
danasilver: Nice, thanks

firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
danasilver: Good call

firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
danasilver: Ah, thanks. I wasn't sure if `trySendBlocking` would be called in the same thread as the callback itself (which is called with a background executor. I updated `FirebaseRemoteConfig` to expose its executor and use that here. LMK if that's what you had in mind.

danasilver:
Appreciate the review @thatfiredev! Super helpful. I updated the PR - let me know what you think.

firebase-config/ktx/src/main/kotlin/com/google/firebase/remoteconfig/ktx/RemoteConfig.kt
thatfiredev: I think that should work :)

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -582,6 +582,15 @@ void startLoadingConfigsFromDisk() {
     fetchedConfigsCache.get();
   }
 
+  /**
+   * Execute a runnable in Remote Config's background thread pool.
+   *
+   * @hide
+   */
+  public void schedule(Runnable runnable) {
```
thatfiredev, line 590: Have you tried making this method private so that we can remove the `@hide`?
I assume it works since the KTX function is under the same package name. But if it doesn't we can keep it as is.

thatfiredev (commented):
Changes LGTM :)
Just added one minor comment

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
danasilver: Yeah, I was hoping that, too but no luck. I think the `.ktx` sub-package means it needs to be public here.

maciel310 (approved):

PR #4851, opened 2023-04-04
Integrate firebase-sessions with GoogleDataTransport

samedson:
- Adds a dependency on data-transport
 - Adds TestSessionEventData.kt to store Session Event objects used for testing

#no-changelog

### DIFF
--- a/firebase-sessions/firebase-sessions.gradle.kts
+++ b/firebase-sessions/firebase-sessions.gradle.kts
@@ -47,9 +47,11 @@ dependencies {
   implementation("com.google.firebase:firebase-encoders-json:18.0.1")
   implementation("com.google.firebase:firebase-encoders:17.0.0")
   implementation("com.google.firebase:firebase-installations-interop:17.1.0")
+  implementation("com.google.android.datatransport:transport-api:3.0.0")
   implementation(libs.androidx.annotation)
 
   runtimeOnly("com.google.firebase:firebase-installations:17.1.3")
+  runtimeOnly(project(":firebase-datatransport"))
 
   testImplementation(libs.androidx.test.junit)
   testImplementation(libs.androidx.test.runner)
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.sessions
+
+import com.google.android.datatransport.*
+import com.google.android.datatransport.TransportFactory
+import com.google.firebase.inject.Provider
+
+/**
+ * The [EventGDTLoggerInterface] is for testing purposes so that we can mock EventGDTLogger in other
+ * classes that depend on it.
+ *
+ * @hide
+ */
+internal interface EventGDTLoggerInterface {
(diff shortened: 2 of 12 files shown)


firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -51,6 +51,8 @@ dependencies {
 
   runtimeOnly("com.google.firebase:firebase-installations:17.1.3")
 
+  implementation("com.google.android.datatransport:transport-api:3.0.0")
+
```
samedson, line 55: I also see the following in other SDKs. Do I need them?

```

   implementation("com.google.android.datatransport:transport-runtime:3.1.8")
   implementation("com.google.android.datatransport:transport-backend-cct:3.1.8")
```

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
```diff
@@ -0,0 +1,61 @@
...
+      .send(Event.ofData(sessionEvent))
+  }
+
+  private fun encode(value: SessionEvent): ByteArray {
+    return SessionEvents.SESSION_EVENT_ENCODER.encode(value).toByteArray()
+  }
+
+  companion object {
+    // TODO What do we put for the AQS Log Source
+    private const val AQS_LOG_SOURCE = "AQS_LOG_SOURCE"
```
samedson, line 59: What do we put for the log source? It seems to be a number on iOS, but a string on Android

firebase-sessions/firebase-sessions.gradle.kts
visumickey: I think you don't need `com.google.android.datatransport:transport-backend-cct`. Basically, there is an API component to use the APIs, runtime component to do the actual operation and backend-cct is for clearcut compatibility. We don't need CCT compatibility here.

Here is how Fireperf depends on Firelog. https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/firebase-perf.gradle#L104-L105

So either a runtime dependency on their runtime dependency or a pointed dependency to a production version - both are fine.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
visumickey: I think it needs both LogSourceName and LogSourceNumber. Firelog reference https://github.com/firebase/firebase-android-sdk/blob/master/transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/LogRequest.java#L38-L42.

This is how we did it Fireperf. https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigurationConstants.java#L581-L590

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionCoordinator.kt
```diff
@@ -47,6 +48,12 @@ internal class SessionCoordinator(
         }
 
       Log.i(TAG, "Initiate session start: $sessionEvent")
```
mrober, line 50: Can you move this log to after the `.log` in the try below? Then we either get a log of success or failure on the session start being sent.

mrober (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionCoordinator.kt
samedson: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
samedson: I see a lot of other SDKs just using the log source name, so I'm trying that out. Gonna send some events in the test app hopefully

firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -47,9 +47,11 @@ dependencies {
   implementation("com.google.firebase:firebase-encoders-json:18.0.1")
   implementation("com.google.firebase:firebase-encoders:17.0.0")
   implementation("com.google.firebase:firebase-installations-interop:17.1.0")
+  implementation("com.google.android.datatransport:transport-api:3.0.0")
   implementation(libs.androidx.annotation)
 
   runtimeOnly("com.google.firebase:firebase-installations:17.1.3")
+  runtimeOnly(project(":firebase-datatransport"))
```
samedson, line 54: This was required to make FireLog work

PR #4852, opened 2023-04-04
Add comparability to ModuleVersion and use it to fix a behavior bug i…

davidmotson:
…n PublishingPlugin

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroupRegistrar.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroupRegistrar.kt
@@ -25,8 +25,8 @@ class LibraryGroupRegistrar {
     val maxVersion =
       librariesByGroup
         .get(libraryGroup)
-        .map { ModuleVersion.fromStringOrNull(it.project.version.toString()) }
-        .filterNotNull()
+        .map {it.project.version.toString()}
+        .mapNotNull { ModuleVersion.fromStringOrNull(it) }
         .maxOrNull()
     if (maxVersion != null) {
       librariesByGroup
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroupRegistrar.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroupRegistrar.kt
@@ -25,11 +25,15 @@ class LibraryGroupRegistrar {
     val maxVersion =
       librariesByGroup
         .get(libraryGroup)
-        .map { it.project.version.toString() }
-        .filter { it.first().isDigit() } // to filter out any flag values
+        .map { ModuleVersion.fromStringOrNull(it.project.version.toString()) }
+        .filterNotNull()
         .maxOrNull()
-        ?: "unspecified"
-    librariesByGroup.get(libraryGroup).forEach { it.project.version = maxVersion }
+    if (maxVersion != null) {
+      librariesByGroup
+        .get(libraryGroup)
+        .filter { ModuleVersion.fromStringOrNull(it.project.version.toString()) == null }
+        .forEach { it.project.version = maxVersion.toString() }
+    }
   }
 
   fun getLibrariesForGroup(name: String) = librariesByGroup.get(name)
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -57,17 +57,17 @@ enum class PreReleaseVersionType {
...
- *
  * ```
  * "12.13.1-beta" // 12.13.1-beta01
  * ```
  *
- * @see fromStringsOrNull
- *
  * @property type an enum of [PreReleaseVersionType] that identifies the pre-release identifier
  * @property build an [Int] that specifies the build number; defaults to one
+ * @see fromStringsOrNull
```
daymxn, line 66: This breaks the popup rendering (when you hover over the class), but ironically doesn't break the inline rendering. The fix is having it above the `@property` annotations :')

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -135,10 +135,21 @@ data class ModuleVersion(
...
+
+  override fun compareTo(other: ModuleVersion) =
+    compareValuesBy(
+      this,
+      other,
+      { it.major },
+      { it.minor },
+      { it.patch },
+      { it.pre == null }, // False (no extra) sorts above true (has extra)
+      { it.pre } // gradle uses lexicographic ordering
```
daymxn, line 151: this comment isn't needed, as it technically doesn't really apply here.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
```diff
@@ -135,10 +135,21 @@ data class ModuleVersion(
...
+  override fun toString() = "$major.$minor.$patch${pre?.let { "-${it.toString()}" } ?: ""}"
+
+  override fun compareTo(other: ModuleVersion) =
+    compareValuesBy(
+      this,
+      other,
+      { it.major },
+      { it.minor },
+      { it.patch },
+      { it.pre == null }, // False (no extra) sorts above true (has extra)
```
daymxn, line 150:
```suggestion
      { it.pre == null }, // a version with no prerelease version takes precedence 
```

Thoughts?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroupRegistrar.kt
```diff
@@ -25,11 +25,15 @@ class LibraryGroupRegistrar {
     val maxVersion =
       librariesByGroup
         .get(libraryGroup)
-        .map { it.project.version.toString() }
-        .filter { it.first().isDigit() } // to filter out any flag values
+        .map { ModuleVersion.fromStringOrNull(it.project.version.toString()) }
+        .filterNotNull()
```
daymxn, line 29: You can use `mapNotNull` instead :)

```kotlin
val maxVersion =
      librariesByGroup
        .get(libraryGroup)
        .map { it.project.version.toString() }
        .mapNotNull { ModuleVersion.fromStringOrNull(it.project.version.toString()) }
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroupRegistrar.kt
```diff
@@ -25,11 +25,15 @@ class LibraryGroupRegistrar {
     val maxVersion =
       librariesByGroup
         .get(libraryGroup)
```
daymxn, line 27: The more Kotlin iodimatic way would be:
```kotlin
librariesByGroup[libraryGroup]
```

But I'm not against this since it's a sequence of method calls

daymxn (approved):
Left some nits, but it's good to go otherwise :)

buildSrc/src/main/java/com/google/firebase/gradle/plugins/ModuleVersion.kt
davidmotson: Ah, unsure what moved these.

PR #4863, opened 2023-04-06
Improve CheckHeadDependencies error reporting.

rlazo:
Error now includes the sdk that's pulling the dependency.


rlazo:
@davidmotson please confirm that ignoring `proto-well-known-types` is OK. Thanks!

VinayGuthal (approved):

PR #4870, opened 2023-04-06
Fix encoding for Session Events

samedson:
- Also updates logging to log the Encoded Session Event instead of the data object (only in debug mode) so that we can catch encoding issues

TODO (in a follow up)
 - Make sure we're setting the time correctly
 - Fill in Device Name

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
@@ -16,6 +16,7 @@
 
 package com.google.firebase.sessions
 
+import android.util.Log
 import com.google.android.datatransport.*
 import com.google.android.datatransport.TransportFactory
 import com.google.firebase.inject.Provider
@@ -53,11 +54,11 @@ internal class EventGDTLogger(private val transportFactoryProvider: Provider<Tra
   }
 
   private fun encode(value: SessionEvent): ByteArray {
+    Log.w("[SAM]", SessionEvents.SESSION_EVENT_ENCODER.encode(value))
     return SessionEvents.SESSION_EVENT_ENCODER.encode(value).toByteArray()
   }
 
   companion object {
-    // TODO What do we put for the AQS Log Source
     private const val AQS_LOG_SOURCE = "FIREBASE_APPQUALITY_SESSION"
   }
 }
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
@@ -32,7 +32,7 @@ internal interface EventGDTLoggerInterface {
 }
 
 /**
- * The [EventGDTLogger] is responsible for encoding and logging events to the Google Data Transport
+ * The [EventGDTLogger] is responsible for encoding        and logging events to the Google Data Transport
  * library.
  *
  * @hide
@@ -54,11 +54,14 @@ internal class EventGDTLogger(private val transportFactoryProvider: Provider<Tra
   }
 
   private fun encode(value: SessionEvent): ByteArray {
-    Log.w("[SAM]", SessionEvents.SESSION_EVENT_ENCODER.encode(value))
-    return SessionEvents.SESSION_EVENT_ENCODER.encode(value).toByteArray()
+    val jsonEvent = SessionEvents.SESSION_EVENT_ENCODER.encode(value)
+    Log.d(TAG, "Encoded Session Start event $jsonEvent")
+    return jsonEvent.toByteArray()
   }
 
   companion object {
+    private const val TAG = "EventGDTLogger"
+
     private const val AQS_LOG_SOURCE = "FIREBASE_APPQUALITY_SESSION"
   }
 }
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
@@ -32,7 +32,7 @@ internal interface EventGDTLoggerInterface {
 }
 
 /**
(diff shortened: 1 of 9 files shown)


visumickey:
> 

Are these TODOs still to be done or are they verified?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
```diff
@@ -59,6 +59,41 @@ internal data class SessionInfo(
...
+
   /** Identifies a unique device+app installation: go/firebase-installations */
   var firebaseInstallationId: String = "",
 )
+
+/** Contains the data collection state for all dependent SDKs and sampling info */
+internal data class DataCollectionStatus(
+  val performance: DataCollectionState = DataCollectionState.COLLECTION_ENABLED,
+  val crashlytics: DataCollectionState = DataCollectionState.COLLECTION_ENABLED,
+  val sessionSamplingRate: Double = 1.0,
```
visumickey, line 76: Interesting to have sessionSamplingRate inside the `DataCollectionStatus`. I was thinking DataCollectionStatus is for external dependencies and used sessionSamplingRate as a remote config. But this looks a bit different from that intention. Curious to hear your rationale to have session sampling rate inside DataCollectionStatus.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
```diff
@@ -59,6 +59,41 @@ internal data class SessionInfo(
...
+  var dataCollectionStatus: DataCollectionStatus = DataCollectionStatus(),
+
   /** Identifies a unique device+app installation: go/firebase-installations */
   var firebaseInstallationId: String = "",
 )
+
+/** Contains the data collection state for all dependent SDKs and sampling info */
+internal data class DataCollectionStatus(
+  val performance: DataCollectionState = DataCollectionState.COLLECTION_ENABLED,
+  val crashlytics: DataCollectionState = DataCollectionState.COLLECTION_ENABLED,
```
visumickey, line 75: Is the default `COLLECTION_ENABLED` (for now?)?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionEventTest.kt
```diff
@@ -32,7 +32,8 @@ class SessionEventTest {
     val sessionEvent =
       SessionEvents.startSession(
         FakeFirebaseApp.fakeFirebaseApp(),
-        TestSessionEventData.TEST_SESSION_DETAILS
+        TestSessionEventData.TEST_SESSION_DETAILS,
+        TestSessionEventData.TEST_SESSION_TIMESTAMP_US,
```
visumickey, line 36: Any reason why the event timestamp is configurable? Can we leave it as a constant?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
```diff
@@ -53,11 +54,14 @@ internal class EventGDTLogger(private val transportFactoryProvider: Provider<Tra
   }
 
   private fun encode(value: SessionEvent): ByteArray {
-    return SessionEvents.SESSION_EVENT_ENCODER.encode(value).toByteArray()
+    val jsonEvent = SessionEvents.SESSION_EVENT_ENCODER.encode(value)
+    Log.d(TAG, "Encoded Session Start event $jsonEvent")
```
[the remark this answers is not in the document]
samedson: I'm down for Verbose

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/EventGDTLogger.kt
```diff
@@ -53,11 +54,14 @@ internal class EventGDTLogger(private val transportFactoryProvider: Provider<Tra
...
+    return jsonEvent.toByteArray()
   }
 
   companion object {
-    // TODO What do we put for the AQS Log Source
+    private const val TAG = "EventGDTLogger"
```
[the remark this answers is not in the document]
samedson: It looks like: `V/EventGDTLogger: Encoded Session Start event {"event_type":1...`

We've been doing it like this so I'll update this file and others

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
samedson: Yeah I'm putting it like this so it looks right in our data but we'll need to change this when we integrate with the product SDKs

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionEventTest.kt
samedson: `startSession` is also used by the SDK. When it leaves off the parameter it uses the current time. Since we can't have the time be different for every test I made it optional to just set the time

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvent.kt
samedson: This is how it is in our proto - 
 I wanted to be consistent with how it's defined: https://source.corp.google.com/piper///depot/google3/logs/proto/firebase/appquality/sessions/session_event.proto;l=67?q=DataCollectionStatus

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionEventTest.kt
mrober: I think this is fine for now, but we need a proper time solution at some point. We can't just use [go/javatime](http://go/javatime) because it requires Android API level 26+. Maybe we can add another function in `WallClock` and try to abstract this away?

mrober (approved):

PR #4874, opened 2023-04-07
Changing Fireperf's SDK Floats to Doubles

jposuna:
RC Flags are stored as Double and our SDK loses precision when converting back to Float. This causes sampling rates to not behave as expected.

### DIFF
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
@@ -230,6 +230,8 @@ public class DeviceCacheManager {
     try {
       // Default value should never be used because key existence is checked above.
       long defaultValue = Double.doubleToLongBits(0.0);
+      // SharedPreferences does not allow storing a Double directly. We store the double's bits as a
+      // long so here we convert that back to a double.
       return Optional.of(Double.longBitsToDouble(sharedPref.getLong(key, defaultValue)));
     } catch (ClassCastException e) {
       logger.debug(
@@ -257,6 +259,9 @@ public class DeviceCacheManager {
         return false;
       }
     }
+    // SharedPreferences does not allow storing a Double directly. The main way to store it without
+    // losing precision is to store the double's bits as a long so it can then be converted back to
+    // a double.
     sharedPref.edit().putLong(key, Double.doubleToRawLongBits(value)).apply();
     return true;
   }
--- a/firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
+++ b/firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
@@ -209,11 +209,12 @@ public class DeviceCacheManager {
    * Retrieves value stored in caching layer for {@code key}, if caching layer is not initialized,
    * or value with the desired type doesn't exist in caching layer, return {@code Optional.empty()}.
    */
-  public Optional<Float> getFloat(String key) {
+  public Optional<Double> getDouble(String key) {
     if (key == null) {
-      logger.debug("Key is null when getting float value on device cache.");
+      logger.debug("Key is null when getting double value on device cache.");
       return Optional.absent();
     }
+    key = newKeyForDoubleMigration(key);
 
     if (sharedPref == null) {
       setContext(getFirebaseApplicationContext());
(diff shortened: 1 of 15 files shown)


raymondlam (approved):

firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
```diff
@@ -228,10 +229,11 @@ public Optional<Float> getFloat(String key) {
 
     try {
       // Default value should never be used because key existence is checked above.
-      return Optional.of(sharedPref.getFloat(key, 0.0f));
+      long defaultValue = Double.doubleToLongBits(0.0);
```
visumickey, line 232: Can you leave some comments for the reason why we are typecasting from Double to Long and viceversa below? It looks a bit weird at first, but I think this is done for a good reason!

firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
```diff
@@ -242,21 +244,33 @@ public Optional<Float> getFloat(String key) {
...
     return true;
   }
 
+  /**
+   * This method creates a new key to disambiguate from an old one.
+   *
+   * <p>It should be used when we are changing the type of a key, so that we don't try to read the
+   * key with the old type and cause a crash.
+   */
+  private static String newKeyForDoubleMigration(String key) {
```
visumickey, line 270: Do we really need a new key? If the old value is recorded as float and the new value type we are looking up is Double:

1. Can we typecast to Double since Double should be able to handle the float precision?
2. To mitigate the potential exception thrown, can we bring in an exception handler to handle this in place?

If none of the above cases are true, then a new flag is fine. But I would want to avoid a flag transition phase if possible. Also, the new flag name does not look that friendly as a name :) (Huge thanks for abstracting :))

visumickey (approved):
The changes look good to me.

There is one scenario I'm curious about for testing - Is there a test we can write for one flag that is already cached as Float and how with this new code change we are backward compatible?

firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
jposuna: Done.

firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
jposuna: I have changed the code to not use the new key. I think it's a bit more complicated now, but overall it should be fine.

firebase-perf/src/main/java/com/google/firebase/perf/util/ImmutableBundle.java
```diff
@@ -68,21 +68,26 @@ public Optional<Boolean> getBoolean(String key) {
...
-      return Optional.fromNullable((Float) o);
-    } catch (ClassCastException e) {
-      logger.debug("Metadata key %s contains type other than float: %s", key, e.getMessage());
+    Object o = bundle.get(key);
+    if (o == null) {
+      return Optional.absent();
+    }
+    if (o instanceof Float) {
+      return Optional.of(((Float) o).doubleValue());
+    }
```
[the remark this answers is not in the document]
jposuna: I've removed the separate key, but for the record, this is a different class and a different storage. This is checking the manifest, not the SharedPreferences (which is where I was using the new key) so the new key did not affect this.

firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
```diff
@@ -242,18 +250,22 @@ public Optional<Float> getFloat(String key) {
    *
    * @return whether provided value has been saved to device caching layer.
    */
-  public boolean setValue(String key, float value) {
+  public boolean setValue(String key, double value) {
```
visumickey, line 253: Now that we are using the same key, do we know the behavior of replacing a float value with a Double value in shared preferences? I don't think that should cause an issue, but still checking for completeness.

firebase-perf/src/test/java/com/google/firebase/perf/config/DeviceCacheManagerTest.java
```diff
@@ -171,64 +174,77 @@ public void setValueString_keyIsNull_returnsFalse() {
...
-    deviceCacheManager.setValue("some_key", 1.2f);
+    deviceCacheManager.setValue("some_key", 1.2);
 
-    assertThat(deviceCacheManager.getFloat("some_key").get()).isEqualTo(1.2f);
+    assertThat(deviceCacheManager.getDouble("some_key").get()).isEqualTo(1.2);
   }
 
   @Test
-  public void getFloat_firebaseAppNotExist_returnsEmpty() {
+  public void getDouble_valueIsSetAsFloat_returnsSetValue() {
```
visumickey, line 203: Thanks for adding this test. Feeling confident now.

visumickey (approved):

firebase-perf/src/main/java/com/google/firebase/perf/config/DeviceCacheManager.java
jposuna: Good point. I did check but I'm adding a test to make sure it's "documented".

PR #4877, opened 2023-04-07
Implement local manifest override configs

visumickey:
Implement local overrides for configs from the application's manifest file.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SessionsSettings.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SessionsSettings.kt
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-package com.google.firebase.sessions
+package com.google.firebase.sessions.settings
 
+import android.content.Context
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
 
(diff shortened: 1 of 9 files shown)


samedson (approved):
Looks good!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SessionsSettings.kt
```diff
@@ -24,26 +25,52 @@ import kotlin.time.Duration.Companion.minutes
...
+  // Order of preference for all the configs below:
+  // 1. Honor local overrides
+  // 2. If no local overrides, use remote config
+  // 3. If no remote config, fall back to SDK defaults.
+
   // Setting to qualify if sessions service is enabled.
   val sessionsEnabled: Boolean
     get() {
+      if (localOverrideSettings.sessionEnabled != null) {
+        return localOverrideSettings.sessionEnabled!!
```
mrober, line 41: Instead of this pattern and using `!!`, Kotlin has nice syntax for dealing with null like:

Elvis if it's just 1 or the other:
`get() =  localOverrideSettings.sessionEnabled :? true`

There is also this for when we have multiple cases:
```
localOverrideSettings.sessionEnabled?.let { return it }

someOtherCases?.let { return it }

return true
```

PR #4887, opened 2023-04-12
Fixes fireperf e2e tests

VinayGuthal:
Perf plugin depends on asm which is not being pulled in transitively in snapshot build which is why we need a direct dependency on asm

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -45,7 +45,7 @@ dependencies {
     // resolution works, otherwise it breaks Fireperf Test Apps.
     // See https://github.com/gradle/gradle/issues/12286
     implementation("com.google.firebase:perf-plugin:$perfPluginVersion")
-
+    implementation("org.ow2.asm", "asm-tree", "9.5")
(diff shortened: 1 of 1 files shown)


buildSrc/build.gradle.kts
```diff
@@ -45,14 +45,14 @@ dependencies {
...
     implementation("com.google.firebase:perf-plugin:$perfPluginVersion")
-
     implementation("com.google.auto.value:auto-value-annotations:1.8.1")
     annotationProcessor("com.google.auto.value:auto-value:1.6.5")
     implementation(kotlin("gradle-plugin", "1.7.10"))
     implementation("org.json:json:20210307")
 
     implementation("org.eclipse.aether:aether-api:1.0.0.v20140518")
     implementation("org.eclipse.aether:aether-util:1.0.0.v20140518")
+    implementation("org.ow2.asm:asm-tree:9.5 ")
```
rlazo, line 55:
```suggestion
    implementation("org.ow2.asm:asm-tree:9.5")
```

rlazo (approved):

PR #4888, opened 2023-04-13
replace sherter plugin with spotless

thatfiredev:
Opening the project with Android Studio Flamingo (Java 17) shows the following error:

```
Execution failed for task ':buildSrc:verifyGoogleJavaFormat'.
> Problems: syntax errors
```

It seems to be due to the fact that the sherter plugin has issues with Java 17: https://github.com/sherter/google-java-format-gradle-plugin/issues/72

Also, sherter is no longer maintained: https://github.com/sherter/google-java-format-gradle-plugin/issues/57#issuecomment-782886280

 This PR migrates away from sherter to [spotless](https://github.com/diffplug/spotless)

#### List of changes:

- Update the GH Actions workflows to use JDK 17
- Replace the sherter plugin with the spotless plugin - now being used to format both Java (`google-java-format`) and Kotlin (`ktfmt` with googleStyle)
- Update the README with new formatting instructions
- Run `./gradlew :spotlessApply` to fix formatting


rlazo:
Never heard of spotless, looks nice! Seems like spotless also supports kotlin, could it be used across for both languages so we don't need to remember 2 different commands as we do now?

thatfiredev:
@rlazo Great idea! I'll update the PR

rlazo:
Yeah, updating to the latest AGP and gradle is going to take time. That PR you linked was an early exploration, and it's very outdated at this point.

So, as it stands, we would need to upgrade everything together (gradle, agp, and spotless)?

thatfiredev:
That's what it seems like.

But I noticed that the first commit in this PR (a484b0ec22d87147b80575bf57e30ab0080bf8fe) built successfully (except for some failures from flaky tests), I will try to isolate that commit in a separate PR. Hopefully it will fix the Android Studio build errors for product teams before we upgrade to AGP 8.

svenjacobs:
Hello, thanks for working on this. This is currently a blocker for external contributions to the Firebase Android SDK because the default JDK of Android Studio is now `17`, so the Gradle sync process fails with the mentioned error.

thatfiredev:
@svenjacobs Thanks for the nudge, I totally forgot about this PR. Let me try to get it into a mergeable state.

imaiya:
Is this pr not planning to merge yet? This problem has plagued my team

thatfiredev:
Closing this PR, as it has already been done in #6100

PR #4892, opened 2023-04-13
Fill in device_model field for AQS

samedson:
I assume this needs to be consistent with the Crashlytics Device Model: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsReportDataCapture.java#L220

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
@@ -16,6 +16,7 @@
 
 package com.google.firebase.sessions
 
+import android.os.Build
 import com.google.firebase.FirebaseApp
 import com.google.firebase.encoders.DataEncoder
 import com.google.firebase.encoders.FieldDescriptor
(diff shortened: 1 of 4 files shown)


firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/TestSessionEventData.kt
```diff
@@ -45,7 +45,7 @@ internal object TestSessionEventData {
       applicationInfo =
         ApplicationInfo(
           appId = FakeFirebaseApp.MOCK_APP_ID,
-          deviceModel = "",
```
mrober, line 48: Do the tests have access to `Build.MODEL` too?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
mrober, line 1: Did some funny formatting happen to this file? Also I think Google style is to avoid the .* imports, and I don't see what class you needed from the import.

samedson: Yeah the formatter did a weird one on this

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/TestSessionEventData.kt
samedson: Seems they do - updated

PR #4893, opened 2023-04-14
Fix warning about mismatch between java/kotlin target version

rlazo:


rlazo:
For reference, this is only necessary until we jump to gradle 8

daymxn (approved):
:OOO

PR #4897, opened 2023-04-14
add github actions for semver check

VinayGuthal:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/publish/PublishingPlugin.java
```diff
@@ -143,6 +143,15 @@ public void apply(Project project) {
                   .set("projectsToPublish", projectsToPublish);
 
               Publisher publisher = new Publisher(publishMode, projectsToPublish);
+              project
+                  .getTasks()
+                  .create(
```
rlazo, line 148: prefer `register`

rlazo (approved):

PR #4904, opened 2023-04-17
Prefer immutable set for test initialization

rlazo:


daymxn (approved):
This gives me an idea...

PR #4905, opened 2023-04-17
Add TimeProvider interface to abstract away the clock

mrober:
Add TimeProvider interface to abstract away the "wall time" clock. This is to avoid passing around a Long of the time in microseconds, which is a bad practice. Also made a nicer FakeTimeProvider.

We can't use `java.time.Instant` because our min SDK level is too low.


firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/FakeTimeProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package com.google.firebase.sessions.testing
+
+import com.google.firebase.sessions.TimeProvider
+import com.google.firebase.sessions.testing.TestSessionEventData.TEST_SESSION_TIMESTAMP_US
+import kotlin.time.Duration
+import kotlin.time.DurationUnit
+
+/** Fake [TimeProvider] that allows programmatically elapsing time forward. */
+class FakeTimeProvider(private val initialTimeUs: Long = TEST_SESSION_TIMESTAMP_US) : TimeProvider {
```
visumickey, line 25: Totally love this idea. Let's add a comment that by default elapsed time would be Zero unless the time is moved using `addInterval` API.

visumickey (approved):
Nice and a sleek refactor. Thanks for doing this Matt.

samedson (approved):

PR #4917, opened 2023-04-20
Introduce remote config cache for storing configs fetched remotely.

visumickey:
Pending: Implement remote config fetching capability and updating the caches with the fetched value.

### DIFF
--- a/firebase-sessions/firebase-sessions.gradle.kts
+++ b/firebase-sessions/firebase-sessions.gradle.kts
@@ -26,7 +26,7 @@ firebaseLibrary {
 
 android {
   val targetSdkVersion: Int by rootProject
-  compileSdk = targetSdkVersion
+  compileSdk = 30
   defaultConfig {
     minSdk = 16
     targetSdk = targetSdkVersion
@@ -48,6 +48,7 @@ dependencies {
   implementation("com.google.firebase:firebase-encoders:17.0.0")
   implementation("com.google.firebase:firebase-installations-interop:17.1.0")
   implementation("com.google.android.datatransport:transport-api:3.0.0")
+  implementation ("androidx.datastore:datastore-preferences:1.0.0")
   implementation(libs.androidx.annotation)
 
   runtimeOnly("com.google.firebase:firebase-installations:17.1.3")
--- a/firebase-sessions/firebase-sessions.gradle.kts
+++ b/firebase-sessions/firebase-sessions.gradle.kts
@@ -26,7 +26,7 @@ firebaseLibrary {
 
 android {
   val targetSdkVersion: Int by rootProject
-  compileSdk = 30
+  compileSdk = 33
   defaultConfig {
     minSdk = 16
     targetSdk = targetSdkVersion
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
@@ -24,7 +24,6 @@ import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
 internal class RemoteSettings(val context: Context) : SettingsProvider {
-  private val SESSION_CONFIGS_NAME = "firebase_session_settings"
   private val Context.dataStore by preferencesDataStore(name = SESSION_CONFIGS_NAME)
   private val settingsCache = SettingsCache(context.dataStore)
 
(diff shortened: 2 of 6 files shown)


firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -26,7 +26,7 @@ firebaseLibrary {
 
 android {
   val targetSdkVersion: Int by rootProject
-  compileSdk = targetSdkVersion
+  compileSdk = 30
```
mrober, line 29: I noticed Crashlytics uses 33. I am not sure if it matters, but maybe we could use 33 here too so there are only 2 different versions, not 3, when building with other Firebase SDKs.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -0,0 +1,89 @@
...
+
+import android.content.Context
+import android.net.Uri
+import androidx.datastore.preferences.preferencesDataStore
+import java.net.URL
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+internal class RemoteSettings(val context: Context) : SettingsProvider {
+  private val SESSION_CONFIGS_NAME = "firebase_session_settings"
```
mrober, line 27: Nit: Constants like this should be in a companion object and have the `const` modifier.  That will be like a `static final` in Java.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -0,0 +1,89 @@
...
+  private val Context.dataStore by preferencesDataStore(name = SESSION_CONFIGS_NAME)
```
mrober, line 28: Is this adding an extension property to the Context class? Is the scope limited to this instance of this class?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -0,0 +1,89 @@
...
+    return settingsCache.hasCacheExpired()
+  }
+
+  companion object SettingsFetcher {
+    private val FIREBASE_SESSIONS_BASE_URL_STRING = "https://firebase-settings.crashlytics.com"
+    private val FIREBASE_PLATFORM = "android"
+    private val fetchInProgress = false
+    private val settingsUrl: URL = run {
+      var uri = Uri.Builder()
+      uri.scheme("https")
```
mrober, line 64: I know this part is incomplete, but for the builder pattern is it better to say like `val uri = Uri.Builder().scheme("https").authority(...).build()` instead?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SessionsSettings.kt
```diff
@@ -28,6 +28,7 @@ import kotlin.time.Duration.Companion.minutes
 internal class SessionsSettings(val context: Context) {
 
   var localOverrideSettings = LocalOverrideSettings(context)
+  var remoteSettings = RemoteSettings(context)
```
mrober, line 31: Should these be `private val` instead?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SettingsCache.kt
```diff
@@ -0,0 +1,162 @@
...
+  val sessionSamplingRate: Double?,
+  val cacheDuration: Int?,
+  val sessionRestartTimeout: Int?,
+  val cacheUpdatedTime: Long?
+)
+
+internal class SettingsCache(private val store: DataStore<Preferences>) {
+  private var sessionConfigs =
+    SessionConfigs(
+      sessionEnabled = null,
```
mrober, line 39: If you put the default ` = null` on each property when you define the data class, this default instance can be just one line. That lets you write out each property name one fewer times.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SettingsCache.kt
```diff
@@ -0,0 +1,162 @@
...
+  internal fun hasCacheExpired(): Boolean {
+    if (sessionConfigs.cacheUpdatedTime != null) {
+      val currentTimestamp = System.currentTimeMillis()
+      val timeDifferenceSeconds = (currentTimestamp - sessionConfigs.cacheUpdatedTime!!) / 1000
+      if (timeDifferenceSeconds < sessionConfigs.cacheDuration!!) return false
+    }
+    return true
+  }
+
+  fun sessionsEnabled(): Boolean? {
```
mrober, line 87: Each of these one line functions can be replaced with the notation `fun sessionsEnabled(): Boolean? = sessionConfigs.sessionEnabled` it will make this class much shorter.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
visumickey: Sounds good. Done!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
visumickey: Yes to both the questions. The key reason is that we should not create more than once instance of the preference data store and the the recommendation is to add it as a extension to the Context class. But I want to limit the exposure of this dataStore to just this class.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
visumickey: Yes! Sorry - Swift bias here. Modified it.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SessionsSettings.kt
visumickey: Yes, made it private now. But eventually I think I will change it var since both of these would be a part of `SessionSettings` constructor.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SettingsCache.kt
visumickey: Like the succinctness. Applied it.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SettingsCache.kt
visumickey: Good suggestion. Done.

mrober (approved):

PR #4922, opened 2023-04-21
Firestore: Internal method AggregateQuerySnapshot.createWithCount() added for CPP SDK

dconeybe:
See https://github.com/firebase/firebase-cpp-sdk/pull/1283


cherylEnkidu (approved):
Thank you for the fix!

PR #4928, opened 2023-04-24
Firestore: Port spec test changes from #7021 that bundle the readTime and resumeToken into a data structure

dconeybe:
Port spec test changes from https://github.com/firebase/firebase-js-sdk/pull/7021 (Bundle the readTime and resumeToken into a data structure)


tom-andersen (approved):
LGTM

PR #4929, opened 2023-04-24
Firestore Spec Tests: Port JS PR 7229 (optimized query resumption using bloom filter)

dconeybe:
Port spec test changes from https://github.com/firebase/firebase-js-sdk/pull/7229 (Optimize local cache sync when resuming a query that had docs deleted).

Note that the integer value of `QueryPurpose.LIMBO_RESOLUTION` changed from `2` to `3` in the spec test JSON files as a result of this port and, therefore, some custom logic was added into `SpecTestCase.java` to handle this correctly.


tom-andersen (approved):
LGTM

PR #4940, opened 2023-04-26
Refactor PublishingPlugin

daymxn:
Per [b/279790953](https://b.corp.google.com/issues/279790953),

This fixes an issue that was preventing our release config generation from functioning. Refer to #4923 for additional context on the original change.

More specifically, this changes the output directory of our release report to be the root directory, as our CI expects it to be.

This also fixes the following bugs:

[b/280323342](https://b.corp.google.com/issues/280323342) - Add a test to avoid this issue in the future
[b/280322962](https://b.corp.google.com/issues/280322962) - Fix IDEA sync breaking Gradle tasks

_Note: this shouldn't be merged until after the current release (M131), as to avoid issues._


daymxn:
> Did you test this running the `create release` action?

[Yup, no issues.](https://github.com/firebase/firebase-android-sdk/actions/runs/4853419243)

PR #4967, opened 2023-05-02
Implement remote config fetcher to fetch and cache configs.

visumickey:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -54,38 +64,102 @@ internal class RemoteSettings(val context: Context) : SettingsProvider {
...
-          .appendPath("settings")
-          .appendQueryParameter("build_version", "")
-          .appendQueryParameter("display_version", "")
-
-      URL(uri.build().toString())
+  internal fun clearCachedSettings() {
+    runBlocking { settingsCache.removeConfigs() }
+  }
+
+  private fun fetchConfigs() {
```
mrober, line 71: Since this function has a few calls to runBlocking and an await, maybe this whole function should be a suspend function?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -54,38 +64,102 @@ internal class RemoteSettings(val context: Context) : SettingsProvider {
...
+          // Logger.logError("[Settings] Fetching newest settings failed with error: \(error)")
+          fetchInProgress = false
+        }
+      )
     }
   }
+
+  companion object {
+    private const val SESSION_CONFIGS_NAME = "firebase_session_settings"
+    private var fetchInProgress = false
```
mrober, line 163: Should `fetchInProgress` be defined in the class? I think we will only make one instance of this class, but it might get weird in the future if this var is in the companion object.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -54,38 +64,102 @@ internal class RemoteSettings(val context: Context) : SettingsProvider {
...
-          .appendPath("gmp")
-          // TODO(visum) Replace below with the GMP APPId
-          .appendPath("GMP_APP_ID")
-          .appendPath("settings")
-          .appendQueryParameter("build_version", "")
-          .appendQueryParameter("display_version", "")
-
-      URL(uri.build().toString())
+  internal fun clearCachedSettings() {
+    runBlocking { settingsCache.removeConfigs() }
```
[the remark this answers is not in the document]
visumickey: This was running in main thread. But modified the code to run the asynchronous tasks on a non-main thread/using default dispatcher threads.

Reference: https://reflectoring.io/understanding-kotlin-coroutines-tutorial/#:~:text=runBlocking%7B%7D%20also%20starts%20a,function%20body%20complete%20their%20execution.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
visumickey: Good idea. Done.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
visumickey: Done.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/RemoteSettingsTest.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    val remoteSettings =
+      RemoteSettings(
+        context,
+        firebaseInstallations,
+        SessionEvents.getApplicationInfo(firebaseApp),
+        fakeFetcher,
+        SESSION_TEST_CONFIGS_NAME
+      )
+    Truth.assertThat(remoteSettings.sessionEnabled).isNull()
```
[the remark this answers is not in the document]
visumickey: Done.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/FakeRemoteConfigFetcher.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package com.google.firebase.sessions.testing
+
+import com.google.firebase.sessions.settings.CrashlyticsSettingsFetcher
+import org.json.JSONObject
+
+class FakeRemoteConfigFetcher : CrashlyticsSettingsFetcher {
+  var responseJSONObject: JSONObject = JSONObject()
```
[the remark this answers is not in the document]
visumickey: Like this idea, done.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -47,45 +61,100 @@ internal class RemoteSettings(val context: Context) : SettingsProvider {
...
+    if (!settingsCache.hasCacheExpired()) {
+      return
+    }
+
+    fetchInProgress = true
+
+    // Get the installations ID before making a remote config fetch
+    var installationId = firebaseInstallationsApi.id.await()
+    if (installationId == null) {
+      fetchInProgress = false
```
samedson, line 91: Should we return here and remove the else indentation?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -47,45 +61,100 @@ internal class RemoteSettings(val context: Context) : SettingsProvider {
...
-          .appendQueryParameter("display_version", "")
-
-      URL(uri.build().toString())
+  internal fun clearCachedSettings() {
+    runBlocking { launch(Dispatchers.Default) { settingsCache.removeConfigs() } }
+  }
+
+  suspend private fun fetchConfigs() {
+    // Check if a fetch is in progress. If yes, return
+    if (fetchInProgress) {
```
samedson, line 77: Does fetchInProgress need to be synchronized?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/RemoteSettingsTest.kt
```diff
@@ -0,0 +1,185 @@
...
+    remoteSettings.updateSettings()
+
+    assertThat(remoteSettings.sessionEnabled).isFalse()
+    assertThat(remoteSettings.samplingRate).isEqualTo(0.75)
+    assertThat(remoteSettings.sessionRestartTimeout).isEqualTo(40.minutes)
+
+    fetchedResponse.remove("app_quality")
+
+    // Sleep for a second before updating configs
+    Thread.sleep(2000)
```
samedson, line 163: Do the sleeps slow down tests?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/FakeRemoteConfigFetcher.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.json.JSONObject
+
+internal class FakeRemoteConfigFetcher(var responseJSONObject: JSONObject = JSONObject()) :
+  CrashlyticsSettingsFetcher {
+  override suspend fun doConfigFetch(
+    headerOptions: Map<String, String>,
+    onSuccess: suspend (JSONObject) -> Unit,
+    onFailure: suspend () -> Unit
+  ) {
+    onSuccess(responseJSONObject)
```
samedson, line 29: Should we test out the `onFailure` case too?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
visumickey: Thanks for flagging it. Made it an `AtomicBoolean` to protect it.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/RemoteSettingsTest.kt
samedson: Ah I see - we should be able to avoid sleeps. Sleeps are bad cause they slow down the tests and are flakey. We get the time here it seems: https://github.com/firebase/firebase-android-sdk/blob/8c12472a74bf95b35df3fbf29928420a106ba6f9/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/SettingsCache.kt#L73

We can pass this in as an function parameter to RemoteSettings. That way we can fast forward time in the test. We do that in our Session Initiator Tests: https://github.com/firebase/firebase-android-sdk/blob/8c12472a74bf95b35df3fbf29928420a106ba6f9/firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt#L136

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
```diff
@@ -47,45 +62,102 @@ internal class RemoteSettings(val context: Context) : SettingsProvider {
     }
 
   override fun updateSettings() {
-    fetchConfigs()
+    runBlocking { launch(Dispatchers.Default) { fetchConfigs() } }
```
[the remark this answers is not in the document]
visumickey: I see. Any reason why we need to use the dispatchers provided by Firebase Common and not the ones provided by the platform?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/FakeRemoteConfigFetcher.kt
visumickey: Totally right. In the RemoteSettings class, we do reset the fetchInProgress to false in case of any failures.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettings.kt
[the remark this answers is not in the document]
visumickey: I will address this in a followup PR.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/RemoteSettingsTest.kt
visumickey: Tried making the refactor to accomodate the TimeProvider, but that seem to take a lot more refactor that what it is today. Example: Time class internally does not provide time comparison with another time value. Until that is done, it is hard to refactor to use that class. So, will leave it as is for now since the time internal to sleep here is just 2 seconds.

PR #4979, opened 2023-05-04
Add FAC getLimitedUseToken to internal interop interface

sam-gc:


appcheck/firebase-appcheck/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # Unreleased
-
+* [unchanged] Update to internal interop API
```
rosalyntan, line 2: Maybe `[changed] Internal updates to allow Firebase SDKs to obtain limited-use tokens.` or something liked that?

firebase-functions/src/androidTest/java/com/google/firebase/functions/TestInteropAppCheckTokenProvider.java
```diff
@@ -66,6 +66,12 @@ public Task<AppCheckTokenResult> getToken(boolean forceRefresh) {
     return Tasks.forResult(testToken);
   }
 
+  @NonNull
```
rosalyntan, line 69: Is the logic to have the functions SDK invoke the new interop method in a separate PR? If so, maybe move this change there as well?

rosalyntan (approved):

firebase-functions/src/androidTest/java/com/google/firebase/functions/TestInteropAppCheckTokenProvider.java
sam-gc: Separate PR, but this change needs to be here to allow this to build (since it implements the interface directly)

PR #4983, opened 2023-05-05
Add thread safe FirebaseSessionsDependencies

mrober:
Add thread safe FirebaseSessionsDependencies. All the thread safe code will live in this class.

FirebaseSessions will use this to get all the dependencies after they have been registered by calling `getSubscribers()` in a coroutine, and it will block until all dependencies are registered. This is kind of like how on iOS it uses promises and waits for them to be fulfilled, except it uses a Mutex that gets added as locked and not unlocked until the dependency resisters itself.


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+object FirebaseSessionsDependencies {
+  private val dependencies = synchronizedMap(mutableMapOf<SessionSubscriber.Name, Dependency>())
+
+  /**
+   * Add a subscriber as a dependency to the Sessions SDK. Every dependency must register itself, or
+   * the Sessions SDK will never generate a session.
+   */
+  fun addDependency(subscriberName: SessionSubscriber.Name) {
+    if (dependencies.containsKey(subscriberName)) {
+      throw IllegalArgumentException("Dependency $subscriberName already added.")
```
samedson, line 40: This looks great - my only question is around the throws. Is the expectation that the callers are going to always try catch these functions?

mrober: No these are runtime exceptions to prevent the caller from using us wrong. We will do the Crashlytics and Performance integrations correctly, so these exceptions will never happen. But if another product integrated with us in the future, but did something funny like add themself as a dependency twice, this will stop them when they test it.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // The dependency is locked until the subscriber registers itself.
+    dependencies[subscriberName] = Dependency(Mutex(locked = true))
+  }
+
+  /** Register and unlock the subscriber. This must be called before [getSubscribers] can return. */
+  internal fun register(subscriber: SessionSubscriber) {
```
samedson, line 48: I'm not sure if it's important, but this might not be thread safe. I think we can set the subscriber variable twice and unlock twice.

mrober: Do you mean the first check if the subscriber is already registered? That is just for a nicer exception message, it's not safe. If a race makes that check fail to notice, we will overwrite the subscriber, but then the mutex unlock will throw an exception for sure.

visumickey (commented):
I'm slightly confused with the sequence of operations here:

1. Who would call addDependency? Is it the dependency like Crashlytics/Fireperf?
2. Who would call register? I presume it is the dependency. Correct me if I'm wrong
3. Why is there a mutex lock on dependency between `addDependency` and `Register`?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/SessionSubscriber.kt
```diff
@@ -0,0 +1,35 @@
...
+  /** [SessionSubscriber.Name]s are used for identifying subscribers. */
+  enum class Name {
+    CRASHLYTICS,
+    PERFORMANCE,
+  }
+
+  /** [SessionDetails] contains session data passed to subscribers whenever the session changes */
+  data class SessionDetails(val sessionId: String)
+
+  fun onSessionChanged(sessionDetails: SessionDetails)
```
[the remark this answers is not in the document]
mrober: This is a different SessionDetails that currently only contains the sessionId. This is the same as what iOS does. We can change it to just a string of a sessionId, but I like the idea of it being a container like this because we can add more fields in the future without breaking the api.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+    val subscriberName = subscriber.sessionSubscriberName
+    val dependency = getDependency(subscriberName)
+
+    dependency.subscriber?.run {
+      throw IllegalArgumentException("Subscriber $subscriberName already registered.")
```
visumickey, line 53: I'm thinking that re-register does not have to throw an exception either. Again a similar rationale - It is not a breaking situation for the app/developers.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+    dependency.subscriber = subscriber
+
+    // Unlock to show the subscriber has been registered, it is possible to get it now.
+    dependency.mutex.unlock()
+  }
+
+  /** Gets the subscribers safely, blocks until all the subscribers are registered. */
+  internal suspend fun getSubscribers(): Map<SessionSubscriber.Name, SessionSubscriber> {
```
visumickey, line 62: Why `suspend` - Is this to ensure that all the subscribers who are in the process of registering are complete before returning? If so, I would rather not worry about this too much and actually return the current state rather than wait for all the register/dependencies are done with the subscription?

visumickey: Do you want to call it `getRegisteredSubscribers`?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+  /** Gets the subscriber regardless of being registered. This is exposed for testing. */
+  @VisibleForTesting
+  internal operator fun get(subscriberName: SessionSubscriber.Name): SessionSubscriber {
+    return getDependency(subscriberName).subscriber
+      ?: throw IllegalStateException("Subscriber $subscriberName has not been registered.")
+  }
+
+  /** Resets the registered subscribers for testing purposes. */
+  @VisibleForTesting
+  internal fun reset() {
```
visumickey, line 78: Should we clean up the dependency list here as well? `reset` makes me feel that all the dependencies/registrations will be cleaned up.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+    dependencies.keys.forEach { dependencies[it] = Dependency(Mutex(locked = true)) }
+  }
+
+  private fun getDependency(subscriberName: SessionSubscriber.Name): Dependency {
```
visumickey, line 82: This API is so close to the `get(subscriberName)` that I don't know what is the difference between both and when would I use which API.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/SessionSubscriber.kt
```diff
@@ -0,0 +1,35 @@
...
+
+  val isDataCollectionEnabled: Boolean
+
+  val sessionSubscriberName: Name
```
visumickey, line 34: Should this be `SessionSubscriber.Name` for clarity?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependenciesTest.kt
```diff
@@ -0,0 +1,127 @@
...
+
+  @Test(expected = TimeoutCancellationException::class)
+  fun getSubscribers_neverRegister_waitsForever() = runTest {
+    // The register never happens, wait until the timeout.
+    withTimeout(2.seconds) { FirebaseSessionsDependencies.getSubscribers() }
+  }
+
+  companion object {
+    init {
+      // Add only Crashlytics as a dependency, not Performance.
```
visumickey, line 119: Why not Performance?

visumickey (commented):
Changes look good to me. Just waiting for the naming of the functions to change to approving it.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: Done.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/SessionSubscriber.kt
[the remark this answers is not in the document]
mrober: Removed exception anyway.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: I made both `addDependency` and `register` not throw, they will log a debug message and return.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: This needs to be a suspend function because `mutex.withLock` is.

It explicitly does wait until all dependencies are registered to keep the logic in the call site simpler. We will be able to just say `val subscribers = FirebaseSessionsDependencies.getSubscribers()` and then on the next line know that `subscribers` is populated with all the dependent subscribers, they are registered, and know if they have data collection enabled.

If a subscriber enables data collection later, then sessions after that will count as data collection enabled.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: I changed this to clear all depedencies, and updated the tests. This is cleaner.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: I removed `get` and have explicit `getSubscriber` and `getDependency` functions now. The `get` was nice to use the `[]` syntax, but this is more readable.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/SessionSubscriber.kt
mrober: Done.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependenciesTest.kt
mrober: This is no longer needed now the new `reset` method clears depedencies, it is easier to create more isolated tests.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
visumickey: That makes sense. Sounds good to have it as a `suspend` method. Since we are calling to get all the successfully registered subscribers, do you want to call it `getRegisteredSubscribers`? This give a good comprehension of what this method does.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,93 @@
...
+    return getDependency(subscriberName).subscriber
+      ?: throw IllegalStateException("Subscriber $subscriberName has not been registered.")
+  }
+
+  /** Resets all the dependencies for testing purposes. */
+  @VisibleForTesting internal fun reset() = dependencies.clear()
+
+  private fun getDependency(subscriberName: SessionSubscriber.Name): Dependency {
+    return dependencies.getOrElse(subscriberName) {
+      throw IllegalStateException(
```
visumickey, line 86: Sorry, a late nit pick: Should this throw an exception? Can we get away with a Log message? Basically when this exception occurs for a developers, there is nothing he would be able to fix it. So, thinking if this should be converted to a log message.

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: I tried making this nullable, but it is pretty ugly to deal with nulls through this class. If the dependency hasn't been added then `getSubscriber` will end up throwing anyway, just with a less useful message. I intend this exception to help the 1P developer integrate a Firebase product with Sessions in the future.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: Done.

PR #4987, opened 2023-05-08
Add option to allow SDK create cache indexes automatically to improve query execution locally

cherylEnkidu:
Preparing work for index auto creation feature.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
@@ -15,6 +15,8 @@
 package com.google.firebase.firestore;
 
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
@@ -100,5 +102,22 @@ public class IndexingTest {
                     + "  ],\n"
                     + "  \"fieldOverrides\": []\n"
                     + "}"));
+  }
+
+  @Test
+  public void testAutomaticIndexingSetSuccessfully()
+      throws ExecutionException, InterruptedException {
+    // Use persistent disk cache (default)
+    FirebaseFirestore db = testFirestore();
+    FirebaseFirestoreSettings settings =
+        new FirebaseFirestoreSettings.Builder(db.getFirestoreSettings())
+            // Use persistent disk cache (default)
+            .setLocalCacheSettings(PersistentCacheSettings.newBuilder().build())
+            .build();
+    db.setFirestoreSettings(settings);
+    Tasks.await(db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
+    assertTrue(db.getClient().getLocalStore().getQueryEngine().getAutomaticIndexingEnabled());
+    Tasks.await(db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
+    assertFalse(db.getClient().getLocalStore().getQueryEngine().getAutomaticIndexingEnabled());
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
@@ -15,8 +15,7 @@
 package com.google.firebase.firestore;
 
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
@@ -105,19 +104,33 @@ public class IndexingTest {
   }
 
   @Test
-  public void testAutomaticIndexingSetSuccessfully()
-      throws ExecutionException, InterruptedException {
+  public void testAutomaticIndexingSetSuccessfully() {
     // Use persistent disk cache (default)
     FirebaseFirestore db = testFirestore();
     FirebaseFirestoreSettings settings =
         new FirebaseFirestoreSettings.Builder(db.getFirestoreSettings())
-            // Use persistent disk cache (default)
             .setLocalCacheSettings(PersistentCacheSettings.newBuilder().build())
             .build();
     db.setFirestoreSettings(settings);
-    Tasks.await(db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
-    assertTrue(db.getClient().getLocalStore().getQueryEngine().getAutomaticIndexingEnabled());
-    Tasks.await(db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
-    assertFalse(db.getClient().getLocalStore().getQueryEngine().getAutomaticIndexingEnabled());
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
+    assertDoesNotThrow(
+        () -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
+  }
+
+  @Test
+  public void testAutomaticIndexingSetSuccessfullyUseDefault() {
+    // Use persistent disk cache (default)
+    FirebaseFirestore db = testFirestore();
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
+    assertDoesNotThrow(
+        () -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
+  }
+
+  public void assertDoesNotThrow(Runnable block) {
+    try {
+      block.run();
+    } catch (Exception e) {
+      fail("Should not have thrown " + e);
+    }
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
@@ -18,6 +18,7 @@ import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCol
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
 import static com.google.firebase.firestore.testutil.TestUtil.assertDoesNotThrow;
+import static com.google.firebase.firestore.testutil.TestUtil.expectError;
 import static com.google.firebase.firestore.testutil.TestUtil.map;
 import static org.junit.Assert.assertEquals;
 
@@ -108,7 +109,7 @@ public class IndexingTest {
   }
 
   @Test
-  public void testAutomaticIndexingSetSuccessfully() {
+  public void testAutoIndexCreationSetSuccessfully() {
     // Use persistent disk cache (default)
     FirebaseFirestore db = testFirestore();
     FirebaseFirestoreSettings settings =
@@ -138,7 +139,7 @@ public class IndexingTest {
   }
 
   @Test
-  public void testAutomaticIndexingSetSuccessfullyUseDefault() {
+  public void testAutoIndexCreationSetSuccessfullyUsingDefault() {
     // Use persistent disk cache (default)
     FirebaseFirestore db = testFirestore();
 
@@ -160,5 +161,19 @@ public class IndexingTest {
 
     results = waitFor(collection.whereEqualTo("match", true).get());
     assertEquals(1, results.size());
+  }
+
+  @Test
+  public void testAutoIndexCreationAfterFailsTermination() {
+    FirebaseFirestore db = testFirestore();
+    waitFor(db.terminate());
+
+    expectError(
+        () -> db.getPersistentCacheIndexManager().enableIndexAutoCreation(),
+        "The client has already been terminated");
+
+    expectError(
+        () -> db.getPersistentCacheIndexManager().disableIndexAutoCreation(),
+        "The client has already been terminated");
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
@@ -15,7 +15,7 @@
 package com.google.firebase.firestore;
 
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
-import static org.junit.Assert.fail;
+import static com.google.firebase.firestore.testutil.TestUtil.assertDoesNotThrow;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
@@ -124,13 +124,5 @@ public class IndexingTest {
     assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
     assertDoesNotThrow(
         () -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
-  }
-
-  public void assertDoesNotThrow(Runnable block) {
-    try {
-      block.run();
-    } catch (Exception e) {
-      fail("Should not have thrown " + e);
-    }
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
@@ -14,8 +14,12 @@
 
 package com.google.firebase.firestore;
 
+import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testCollectionWithDocs;
 import static com.google.firebase.firestore.testutil.IntegrationTestUtil.testFirestore;
+import static com.google.firebase.firestore.testutil.IntegrationTestUtil.waitFor;
 import static com.google.firebase.firestore.testutil.TestUtil.assertDoesNotThrow;
+import static com.google.firebase.firestore.testutil.TestUtil.map;
+import static org.junit.Assert.assertEquals;
 
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 import com.google.android.gms.tasks.Task;
@@ -112,15 +116,49 @@ public class IndexingTest {
             .setLocalCacheSettings(PersistentCacheSettings.newBuilder().build())
             .build();
     db.setFirestoreSettings(settings);
+
+    CollectionReference collection =
+        testCollectionWithDocs(
+            map(
+                "a", map("match", true),
+                "b", map("match", false),
+                "c", map("match", false)));
+    QuerySnapshot results = waitFor(collection.whereEqualTo("match", true).get());
+    assertEquals(1, results.size());
+
     assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().enableIndexAutoCreation());
+
+    results = waitFor(collection.whereEqualTo("match", true).get());
+    assertEquals(1, results.size());
+
     assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().disableIndexAutoCreation());
+
+    results = waitFor(collection.whereEqualTo("match", true).get());
+    assertEquals(1, results.size());
   }
 
   @Test
   public void testAutomaticIndexingSetSuccessfullyUseDefault() {
     // Use persistent disk cache (default)
     FirebaseFirestore db = testFirestore();
+
+    CollectionReference collection =
+        testCollectionWithDocs(
+            map(
+                "a", map("match", true),
+                "b", map("match", false),
+                "c", map("match", false)));
+    QuerySnapshot results = waitFor(collection.whereEqualTo("match", true).get());
+    assertEquals(1, results.size());
+
     assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().enableIndexAutoCreation());
+
+    results = waitFor(collection.whereEqualTo("match", true).get());
+    assertEquals(1, results.size());
+
     assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().disableIndexAutoCreation());
+
+    results = waitFor(collection.whereEqualTo("match", true).get());
+    assertEquals(1, results.size());
   }
 }
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
@@ -112,17 +112,15 @@ public class IndexingTest {
             .setLocalCacheSettings(PersistentCacheSettings.newBuilder().build())
             .build();
     db.setFirestoreSettings(settings);
-    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
-    assertDoesNotThrow(
-        () -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().enableIndexAutoCreation());
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().disableIndexAutoCreation());
   }
 
   @Test
   public void testAutomaticIndexingSetSuccessfullyUseDefault() {
     // Use persistent disk cache (default)
     FirebaseFirestore db = testFirestore();
-    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
-    assertDoesNotThrow(
-        () -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(false));
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().enableIndexAutoCreation());
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().disableIndexAutoCreation());
   }
 }
--- a/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
+++ b/firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
@@ -406,6 +406,8 @@ public class FirebaseFirestore {
     return client.configureFieldIndexes(parsedIndexes);
   }
 
+  // TODO(csi): Remove the `hide` and scope annotations.
+  /** @hide */
   @RestrictTo(RestrictTo.Scope.LIBRARY)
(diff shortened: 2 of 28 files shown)


firebase-firestore/src/main/java/com/google/firebase/firestore/local/AutoIndexing.java
```diff
@@ -0,0 +1,7 @@
+package com.google.firebase.firestore.local;
+
+public class AutoIndexing {
```
wu-hui, line 3: `QueryContext` might be a better name.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -82,12 +82,13 @@ public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
       return result;
     }
 
-    result = performQueryUsingRemoteKeys(query, remoteKeys, lastLimboFreeSnapshotVersion);
+    AutoIndexing counter = new AutoIndexing();
+    result = performQueryUsingRemoteKeys(query, remoteKeys, lastLimboFreeSnapshotVersion, counter);
```
wu-hui, line 86: Here `counter` can be `@Nullable` to indicate index-free queries do no collect information.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteRemoteDocumentCache.java
```diff
@@ -239,6 +257,7 @@ private void processRowInBackground(
         () -> {
           MutableDocument document =
               decodeMaybeDocument(rawDocument, readTimeSeconds, readTimeNanos);
+          counter.fullScanCount++;
```
wu-hui, line 260: Potential race condition, this can go in the `synchronized` block below.

wu-hui (commented):
Some comments, I think you get the overall idea however. 

While you are at it, maybe write a simple count-based heuristics: compare the doc reads from full collection scan, and indexed-doc reads from a imaginary perfect indexed query, and make a decision to turn on indexing?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
cherylEnkidu: doesn't `index-free queries` do a full collection scan, which collects infomation?

wu-hui: Index free does not do full collection scan. It takes the result set from backend, and amend it with the local mutations and documents with newer version than the result set. It does not accurately reflect how much a full collection scan will cost.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+public class QueryContext {
+  QueryContext() {}
+
+  int fullScanCount = 0;
```
wu-hui, line 20: Nit: it's probably better the do the increment using a method.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -233,6 +239,21 @@ public void deleteFieldIndex(FieldIndex index) {
     }
   }
 
+  @Override
+  public void createTargetIndices(Target target) {
+    hardAssert(started, "IndexManager not started");
+
+    if (!autoClientIndexingEnabled) return;
```
wu-hui, line 246: I am not sure this is useful..nobody would call createTargetIndices if this is configured false?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -233,6 +239,21 @@ public void deleteFieldIndex(FieldIndex index) {
...
+
+    for (Target subTarget : getSubTargets(target)) {
+      IndexType type = getIndexType(subTarget);
+      if (type == IndexType.NONE) {
```
wu-hui, line 250: This can be None or Partial? When we decide to build index, we can always make sure it's a full index.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
```diff
@@ -186,6 +186,34 @@ public boolean servedByIndex(FieldIndex index) {
     return true;
   }
 
+  public FieldIndex BuildTargetIndex() {
```
wu-hui, line 189: Even with a prototype, this code should be tested. CSI is complicated enough to warrant some tests.

Specifically, I want to see:

A list of query/targets feed to the matcher and get `FieldIndex` from buildTargetIndex().

All these generated indexes should all have `servedByIndex()` return true..they should all be a full match too.

wu-hui (commented):
Looks good for a prototype.

I want to see some tests for the change in TargetIndexMatcher, then we should be able to move forward.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
```diff
@@ -186,6 +188,48 @@ public boolean servedByIndex(FieldIndex index) {
...
+        uniqueFields.add(filter.getField());
+        segments.add(
+            FieldIndex.Segment.create(filter.getField(), FieldIndex.Segment.Kind.ASCENDING));
+      }
+    }
+
+    for (OrderBy orderBy : orderBys) {
+      if (uniqueFields.contains(orderBy.getField())) {
+        continue;
+      }
```
wu-hui, line 218: I think we can add this between line 218 and 219:

``` Java
      // Stop adding more segments if we see a order-by on key. Typically this is the default
      // implicit order-by which is covered in the index_entry table as a separate column.
      // If it is not the default order-by, the generated index will be missing some segments
      // optimized for order-bys, which is probably fine.
      if (orderBy.getField().isKeyField()) {
        break;
      }
```

wu-hui (changes requested):
Please add more tests to `SQLiteLocalStoreTest`. 

This is a place where you can:

* setup a collection
* turn on auto-indexing
* run a query that should have added a field index
* verify field indexes are added
* call backfillIndexes()
* verify query result and document read count is as expected as an indexed query.

You should be able to find enough examples in that file.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
```diff
@@ -101,4 +102,35 @@ public void testBadIndexDoesNotCrashClient() {
...
+  @Test
+  public void testAutomaticIndexingSetSuccessfully() {
+    // Use persistent disk cache (default)
+    FirebaseFirestore db = testFirestore();
+    FirebaseFirestoreSettings settings =
+        new FirebaseFirestoreSettings.Builder(db.getFirestoreSettings())
+            .setLocalCacheSettings(PersistentCacheSettings.newBuilder().build())
+            .build();
+    db.setFirestoreSettings(settings);
+    assertDoesNotThrow(() -> db.getPersistentCacheIndexManager().setAutomaticIndexingEnabled(true));
```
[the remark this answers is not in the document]
cherylEnkidu: IMO the above scenario is not helpful in integration test. Because even after turn it on, the auto indexing cannot have immediate effect.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+public class QueryContext {
+  public QueryContext() {}
+
+  private int documentCount = 0;
```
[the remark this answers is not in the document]
cherylEnkidu: I changed to `documentReadCount` :)

firebase-firestore/src/main/java/com/google/firebase/firestore/local/SQLiteIndexManager.java
```diff
@@ -233,6 +233,19 @@ public void deleteFieldIndex(FieldIndex index) {
     }
   }
 
+  @Override
+  public void createTargetIndices(Target target) {
```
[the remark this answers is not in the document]
cherylEnkidu: The problem for Or queries is they can only be treated as the whole. So either all `subTaregt` use indexes. Or neither of them.

For example, in `A || B` , only A has a full matched index. The SDK will run both `A` and `B` using full collection scan. After that, if the SDK finds it is more efficient to use indexes, it will try to create indexes for both `A` and `B`. So the next time this or query will be served using indexes.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
```diff
@@ -0,0 +1,31 @@
...
+  public QueryContext() {}
+
+  /** Counts the number of documents passed through during local query execution. */
+  private int documentReadCount = 0;
+
+  public int getDocumentReadCount() {
+    return documentReadCount;
+  }
+
+  public void increaseDocumentCount() {
```
dconeybe, line 28: Suggested rename: "increaseDocumentCount" -> "incrementDocumentReadCount"

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -87,7 +97,39 @@ public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
       return result;
     }
 
-    return executeFullCollectionScan(query);
+    QueryContext context = new QueryContext();
+    result = executeFullCollectionScan(query, context);
+    if (result != null && automaticIndexingEnabled) {
+      CreateCacheIndexes(query, context, result.size());
```
dconeybe, line 103: nit: The method name `CreateCacheIndexes` should start with a lowercase letter, conventionally.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -87,7 +97,39 @@ public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
...
+
+  /**
+   * Decides whether SDK should create a full matched field index for this query based on query
+   * context and query result size.
+   */
+  // TODO(csi): Auto experiment data.
+  private void CreateCacheIndexes(Query query, QueryContext context, int resultSize) {
+    String decisionStr = "";
+    // If evaluation is updated, please update tests in SQLiteLocalStoreTest.java
+    if (context.getDocumentReadCount() > 2 * resultSize) {
```
dconeybe, line 116: I really think we should hardcode some absolute minimum document read count that must be exceeded before creating an index. What do you think?

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.firestore.local;
+
+/** A tracker to keep a record of important details during database local query execution. */
+public class QueryContext {
```
dconeybe, line 18: IMO the class name "QueryContext" is not great. To me, its name implies that it is used by the query execution as "input" or a way to acquire resources needed to perform a query. Since this is an internal class I'd name it exactly what it does, something like "DocumentReadCounter". The generality of naming it "QueryContext" is confusing to me.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
```diff
@@ -192,6 +192,59 @@ public boolean servedByIndex(FieldIndex index) {
     return true;
   }
 
+  /** Returns a full matched field index for this target. */
+  public FieldIndex BuildTargetIndex() {
```
dconeybe, line 196: nit: The first character of `BuildTargetIndex` should be lowercase, conventionally.

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
cherylEnkidu: Could you be more specific what `minimum document read count` refers to?

If it is minimum document reads in full collection scan. I believe there is beneficial of reading 1 result from 3 documents collection. And this difference can be solved by heuristic.

if it is minimum result. Is it also useful when result size is 0. If we using index, then we don't need to go through all the documents to find out there is nothing matched.

cherylEnkidu: Discussed in the meeting, we are going to set the threshold to 100.

firebase-firestore/src/main/java/com/google/firebase/firestore/PersistentCacheIndexManager.java
```diff
@@ -0,0 +1,55 @@
...
+ * <p>To use, calling {@link FirebaseFirestore#getPersistentCacheIndexManager()} to get an instance.
+ */
+// TODO(csi): Remove the `hide` and scope annotations.
+/** @hide */
+@RestrictTo(RestrictTo.Scope.LIBRARY)
+public class PersistentCacheIndexManager {
+  @NonNull private FirestoreClient client;
+
+  @RestrictTo(RestrictTo.Scope.LIBRARY)
+  public PersistentCacheIndexManager(FirestoreClient client) {
```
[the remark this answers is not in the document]
cherylEnkidu: The reason for adding visibility modifier is because this will prevent developer from creating it directly. All PersistentCacheIndexManager object should comes from `getPersistentCacheIndexManager()`

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
cherylEnkidu: This class will be shared with other features, for example streaming counts. That's why it has a generic name.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -366,4 +366,201 @@ public void testDeeplyNestedServerTimestamps() {
     }
     assertThat(error.get()).isNull();
   }
+
+  @Test
+  public void testIndexAutoCreationWorks() {
```
wu-hui, line 371: testCanAutoCreateIndexes

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -366,4 +366,201 @@ public void testDeeplyNestedServerTimestamps() {
...
+
+    applyRemoteEvent(addedRemoteEvent(doc("coll/f", 20, map("matches", true)), targetId));
+
+    executeQuery(query);
+    assertRemoteDocumentsRead(/* byKey= */ 2, /* byCollection= */ 1);
+    assertQueryReturned("coll/a", "coll/e", "coll/f");
+  }
+
+  @Test
+  public void testIndexAutoCreationDoesNotWorkWhenCollectionSizeIsTooSmall() {
```
wu-hui, line 401: testDoNotAutoCreateIndexesForSmallCollections

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -366,4 +366,201 @@ public void testDeeplyNestedServerTimestamps() {
...
+
+    applyRemoteEvent(addedRemoteEvent(doc("foo/a", 10, map("matches", true)), targetId2));
+    applyRemoteEvent(addedRemoteEvent(doc("foo/b", 10, map("matches", false)), targetId2));
+    applyRemoteEvent(addedRemoteEvent(doc("foo/c", 10, map("matches", false)), targetId2));
+    applyRemoteEvent(addedRemoteEvent(doc("foo/d", 10, map("matches", false)), targetId2));
+    applyRemoteEvent(addedRemoteEvent(doc("foo/e", 10, map("matches", true)), targetId2));
+
+    executeQuery(query2);
+    assertRemoteDocumentsRead(/* byKey= */ 0, /* byCollection= */ 2);
+
```
wu-hui, line 531: add backfillIndexes here.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -366,4 +366,201 @@ public void testDeeplyNestedServerTimestamps() {
...
+  @Test
+  public void testIndexAutoCreationWorks() {
+    Query query = query("coll").filter(filter("matches", "==", true));
+    int targetId = allocateQuery(query);
+
+    enableIndexAutoCreation();
+    setMinCollectionSizeToAutoCreateIndex(0);
+
+    applyRemoteEvent(addedRemoteEvent(doc("coll/a", 10, map("matches", true)), targetId));
```
wu-hui, line 378: Can we setup the testing docs differently for each test you added:

we should try to cover different types and filters, like strings, numbers, arrays and inequalities/array queries.

Does this make sense?

wu-hui (commented):
Some improvements are still needed, but I think we are very close now.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
dconeybe: Discussed offline. I still feel that we should name this class according to its immediate purpose to improve readability (e.g. a name like "DocumentReadCounter" instead of "QueryContext"). If some future change wants to generalize its purpose then it can be easily refactored and renamed accordingly. But this won't block my approval, and would like to hear from @wu-hui .

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -87,7 +102,48 @@ public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
...
+      indexManager.createTargetIndices(query.toTarget());
+    } else {
+      decisionStr = " not";
+    }
+
+    if (Logger.isDebugEnabled()) {
+      Logger.debug(
+          LOG_TAG,
+          "Query ran locally using a full collection scan, walking through %s documents in total "
```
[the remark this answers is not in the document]
cherylEnkidu: I think there is benefit to logging something when SDK decides not creating any indexes. Since developer may wonder why auto creation has no effect after they turn it on.

wu-hui (approved):
LGTM with some nits.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
```diff
@@ -0,0 +1,29 @@
...
```
dconeybe, line 18: Will `QueryContext` objects be passed between threads anywhere? If yes, make `documentReadCount` an `AtomicInteger` instead of a plain-old `int`.

dconeybe (changes requested):
Thank you for addressing my previous comments. After re-reading, I have some more suggestions, mostly relating to readability rather than to functionality of the code.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -61,16 +62,41 @@
...
 
+  private static final int MIN_COLLECTION_SIZE_TO_AUTO_CREATE_INDEX = 100;
+
+  // TODO(csi): update the temp cost.
+  /**
+   * This cost represents the evaluation result of (([index, docKey] + [docKey, docContent]) per
+   * document in the result set) / ([docKey, docContent] per documents in full collection scan)
+   * coming from experiment https://github.com/firebase/firebase-android-sdk/pull/5064.
+   */
+  private static final int RELATIVE_INDEX_READ_COST = 3;
```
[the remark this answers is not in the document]
cherylEnkidu: I would like to rename it to `RELATIVE_INDEX_READ_COST_PER_DOCUMENT`, the reason for keeping relative is the number is only meaningful when comparing to per document cost in full collection scan.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
```diff
@@ -418,14 +422,15 @@ public Task<Void> setIndexConfiguration(@NonNull String json) {
...
-    if (persistentCacheIndexManager != null) {
-      return persistentCacheIndexManager;
-    }
+    Assert.hardAssertNonNull(
+        persistentCacheIndexManager, "persistentCacheIndexManager has not been initialized.");
+
     if (settings.isPersistenceEnabled()
         || settings.getCacheSettings() instanceof PersistentCacheSettings) {
-      persistentCacheIndexManager = new PersistentCacheIndexManager(client);
+      persistentCacheIndexManager.compareAndSet(null, new PersistentCacheIndexManager(client));
```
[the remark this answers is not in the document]
cherylEnkidu: Thanks for your suggestion! I am going to do
```
if (persistentCacheIndexManager.get() == null) {
  persistentCacheIndexManager.set(new PersistentCacheIndexManager(client));
}
```

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -110,10 +126,26 @@ public ImmutableSortedMap<DocumentKey, Document> getDocumentsMatchingQuery(
...
+    if (context.getDocumentReadCount() < minCollectionSizeToAutoCreateIndex) {
+      Logger.debug(
+          LOG_TAG,
+          "SDK will only creates cache indexes for collection contains more than or equal to "
+              + "%s documents.",
+          minCollectionSizeToAutoCreateIndex);
+      return;
+    }
+
+    if (Logger.isDebugEnabled()) {
```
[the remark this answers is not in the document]
cherylEnkidu: I check the code for `Logger.java` and what you said make sense to me. I will get rid of `if` here.

firebase-firestore/src/main/java/com/google/firebase/firestore/FirebaseFirestore.java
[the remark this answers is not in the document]
dconeybe: That is a race condition, because another thread could have set it to non-null after you check if it's null. You need to do the compareAndSet instead of a blind set.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -61,16 +62,41 @@
 public class QueryEngine {
   private static final String LOG_TAG = "QueryEngine";
 
+  private static final int MIN_COLLECTION_SIZE_TO_AUTO_CREATE_INDEX = 100;
```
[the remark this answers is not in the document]
cherylEnkidu: I took your advise other than `autoIndexCreationRelativeReadCost`, `INDEX_AUTO_CREATION_DEFAULT_RELATIVE_INDEX_READ_COST_PER_DOCUMENT` and `setAutoIndexCreationRelativeIndexReadCost`. Since after changes, the name becomes `indexAutoCreationRelativeIndexReadCostPerDocument`. It is way too long and this cost can potentially used in other place (e.g. decide whether index look up worth it) other than auto creation.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java
```diff
@@ -61,29 +61,35 @@
  */
 public class QueryEngine {
   private static final String LOG_TAG = "QueryEngine";
-  private static final int MIN_COLLECTION_SIZE_TO_AUTO_CREATE_INDEX = 100;
+
+  private static final int INDEX_AUTO_CREATION_DEFAULT_MIN_COLLECTION_SIZE = 100;
```
dconeybe, line 65: nit: Move "DEFAULT" to the front of `INDEX_AUTO_CREATION_DEFAULT_MIN_COLLECTION_SIZE`, to be consistent with the naming of `DEFAULT_RELATIVE_INDEX_READ_COST_PER_DOCUMENT` (to improve readability).

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
dconeybe: Fair enough. QueryContext it is.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
dconeybe: Cheryl, can you confirm whether `QueryContext` instances will be shared between threads? I think the answer is "no", but I'd like you to confirm, because if it _is_ shared between threads then the `int` member variables needs some protection to ensure proper synchronization between threads.

dconeybe (changes requested):
Just a few final comments, but LGTM for the most part.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryContext.java
cherylEnkidu: `documentReadCount` will not be shared among multi threads. The only place calling it is in the main thread. @wu-hui please correct me if I am wrong.
```
db.query(sql.toString())
        .binding(bindVars)
        .forEach(
            row -> {
              processRowInBackground(backgroundQueue, results, row, filter);
              if (context != null) {
                context.incrementDocumentReadCount();
              }
            });
            ```

dconeybe: As long as the lambda specified to `forEach` is called by the same thread as `db.query` (which I imagine it is).

wu-hui: Yeah, they are from the same thread.

dconeybe: Ok thanks for confirming. I'm now satisifed that `QueryContext` can safely use `int` rather than `AtomicInteger`.

dconeybe (approved):

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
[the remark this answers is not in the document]
cherylEnkidu: update: add tests

PR #4989, opened 2023-05-08
Add register api to FirebaseSessions so products can integrate

mrober:
Add register api to FirebaseSessions so products can integrate with Sessions.


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionCoordinator.kt
```diff
@@ -31,30 +28,26 @@ import kotlinx.coroutines.tasks.await
...
-        eventGDTLogger.log(sessionEvent)
-
-        Log.i(TAG, "Successfully logged Session Start event: ${sessionEvent.sessionData.sessionId}")
-      } catch (e: RuntimeException) {
-        Log.e(TAG, "Error logging Session Start event to DataTransport: ", e)
+        firebaseInstallations.id.await()
+      } catch (ex: Exception) {
+        Log.e(TAG, "Error getting Firebase Installation ID: ${ex}. Using an empty ID")
+        // Use an empty fid if there is any failure.
+        ""
```
visumickey, line 40: Do we want to allow dispatching events with empty `fid`s? Can we stop from such events to go through? Empty FID could lead to invalid data ingested.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -0,0 +1,91 @@
...
+import kotlinx.coroutines.sync.withLock
+
+/**
+ * [FirebaseSessionsDependencies] determines when a dependent SDK is installed in the app. The
+ * Sessions SDK uses this to figure out which dependencies to wait for to getting the data
+ * collection state. This is thread safe.
+ *
+ * This is important because the Sessions SDK starts up before dependent SDKs.
+ */
+object FirebaseSessionsDependencies {
```
visumickey, line 31: Looks like we might have to rebase this branch to the thread-safe branch to ensure that this contains only changes relevant this PR.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionCoordinatorTest.kt
```diff
@@ -43,8 +43,7 @@ class SessionCoordinatorTest {
     val firebaseInstallations = FakeFirebaseInstallations("FaKeFiD")
     val sessionCoordinator =
       SessionCoordinator(
-        firebaseInstallations,
-        context = TestOnlyExecutors.background().asCoroutineDispatcher() + coroutineContext,
+        firebaseInstallations = FakeFirebaseInstallations("FaKeFiD"),
```
visumickey, line 46: Could it be `firebaseInstallations = firebaseInstallations`? Why do we recreate a new instance here?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -66,8 +69,16 @@ internal constructor(
     }
   }
 
-  @Discouraged(message = "This will be replaced with a real API.")
-  fun greeting(): String = "Matt says hi!"
+  /** Register the [subscriber]. This must be called for every dependency. */
+  fun register(subscriber: SessionSubscriber) {
+    FirebaseSessionsDependencies.register(subscriber)
+
+    Log.d(
```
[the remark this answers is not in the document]
mrober: I copied this log exactly from iOS https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseSessions/Sources/FirebaseSessions.swift#L240

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -77,9 +88,34 @@ internal constructor(
...
-    )
+    sessionStartScope.launch {
+      val subscribers = FirebaseSessionsDependencies.getSubscribers()
+
+      if (subscribers.isEmpty()) {
+        Log.e(
+          TAG,
+          "Sessions SDK did not have any dependent SDKs register as dependencies. Events will not be sent."
+        )
+        return@launch
```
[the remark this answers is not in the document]
mrober: Yeah it's a bit ugly. I plan to refactor this so it can be in a suspend function but I will need to update a few places so didn't do it in this PR.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -77,9 +88,34 @@ internal constructor(
...
     }
 
-    sessionCoordinator.attemptLoggingSessionEvent(
-      SessionEvents.startSession(firebaseApp, sessionDetails, sessionSettings, timeProvider)
-    )
+    sessionStartScope.launch {
+      val subscribers = FirebaseSessionsDependencies.getSubscribers()
+
+      if (subscribers.isEmpty()) {
+        Log.e(
```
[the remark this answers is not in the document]
mrober: Fixed.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionCoordinator.kt
mrober: I am not sure. @samedson and I talked and thought this was better, we even updated iOS to do the same https://github.com/firebase/firebase-ios-sdk/pull/11161 I am ok with dropping the event instead but we'll need to revert that iOS change to keep it consistent.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionCoordinatorTest.kt
mrober: Fixed!

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -77,9 +88,34 @@ internal constructor(
       return
```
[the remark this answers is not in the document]
mrober: Done.

firebase-sessions/api.txt
```diff
@@ -17,3 +17,51 @@ package com.google.firebase.sessions {
 
 }
 
+package com.google.firebase.sessions.api {
```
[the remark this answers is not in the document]
mrober: This was auto-generated by `./gradlew :firebase-sessions:generateApiTxtFile` but it seems fine if I delete all these internal apis.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: Done.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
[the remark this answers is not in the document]
visumickey: The text looks too elaborate for an external developer. But never mind, lets worry about this messaging later.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/SessionSubscriber.kt
```diff
@@ -16,12 +16,15 @@
...
 
+import androidx.annotation.Discouraged
+
 /** [SessionSubscriber] is an interface that dependent SDKs must implement. */
 interface SessionSubscriber {
   /** [SessionSubscriber.Name]s are used for identifying subscribers. */
   enum class Name {
     CRASHLYTICS,
     PERFORMANCE,
+    @Discouraged(message = "This is for testing purposes only.") MATT_SAYS_HI,
```
visumickey, line 27: Why this?

firebase-sessions/api.txt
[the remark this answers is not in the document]
visumickey: What are the legit APIs that should find a place here? Secondly, how do we control what does inside this file? We can do this as a separate validation task before we go out for launch.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionCoordinator.kt
visumickey: Do you remember the rationale as to why it is fine to dispatch an event with empty FID? Because, backend is going to drop it anyways since this event would be accounted as invalid.

visumickey (commented):
All changes look good to me except for the empty FID scenario. Once that is clarified, this is good to get merged.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/SessionSubscriber.kt
mrober: This is so we can make integration tests that don't rely on Crashlytics or Perf being initialized, but won't break if we do add in Crashlytics or Perf. We use it in the test app. I was worried when we do the gradle configs to include Crashlytics, include Perf, that there would be configurations that wouldn't be valid.

firebase-sessions/api.txt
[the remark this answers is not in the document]
mrober: Ideally I would like this to not expose any public API since it's for internal products only. But we need at least this to be public for Firebase to work with it.

visumickey (approved):

samedson (approved):
This looks great 👍

PR #4993, opened 2023-05-09
Add Sessions to Crashlytics

mrober:
Add Sessions to Crashlytics and populate an internal custom key. Crashlytics still needs to be updated to add a new field to the Report for this id.


build.gradle
```diff
@@ -39,6 +39,7 @@ buildscript {
         classpath 'com.google.gms:google-services:4.3.15'
         classpath "com.ncorti.ktfmt.gradle:plugin:0.11.0"
         classpath 'com.google.firebase:firebase-appdistribution-gradle:3.0.2'
+        classpath 'com.google.firebase:firebase-crashlytics-gradle:2.9.5'
```
visumickey, line 42: Why is this change in here? Is this relevant to this integration?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/CrashlyticsRegistrar.java
```diff
@@ -24,20 +24,29 @@
...
+  }
+
   @Override
   public List<Component<?>> getComponents() {
     return Arrays.asList(
         Component.builder(FirebaseCrashlytics.class)
             .name(LIBRARY_NAME)
             .add(Dependency.required(FirebaseApp.class))
             .add(Dependency.required(FirebaseInstallationsApi.class))
+            .add(Dependency.required(FirebaseSessions.class))
```
visumickey, line 49: Just a curiosity question here: Is the order important here? If so, why are we above the NativeComponent?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -176,6 +179,26 @@ public Void call() throws Exception {
           }
         });
 
+    // TODO(mrober): Replace with a real session implementation.
+    firebaseSessions.register(new SessionSubscriber() {
+      @Override
+      public void onSessionChanged(@NonNull SessionDetails sessionDetails) {
+        Logger.getLogger().d("onSessionChanged: " + sessionDetails);
+        core.setInternalKey("sessionId", sessionDetails.getSessionId());
```
visumickey, line 187: What does internalKey mean for Crashlytics? Is the plan to use this internal key to fill in the proto? If so, any implications in retaining this as a internalKey?

visumickey (approved):

build.gradle
mrober: Yes, otherwise any test app in this repo can't have Crashlytics.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
mrober: This is intended to be replaced with the new field in the report. I added a TODO for it.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/CrashlyticsRegistrar.java
mrober: I don't think the order is important. I just grouped it with other Firebase* deps. Let me know if the order does matter and I can move it. But looking at the other deps, I don't see any real order.

visumickey (approved):

PR #4996, opened 2023-05-10
Bump testing deps to address jsonwebtoken issues

rlazo:
See

https://deps.dev/maven/com.google.firebase%3Afirebase-firestore/24.6.0

Project side card, then vulnerabilities


smoke-tests/src/androidTest/backend/functions/functions/package.json
```diff
@@ -10,12 +10,12 @@
     "logs": "firebase functions:log"
   },
   "engines": {
-    "node": "12"
+    "node": "19"
```
daymxn, line 13: I'm assuming this is fine since it's a test?

daymxn (approved):

daymxn:
This prob deserves a changelog entry btw

smoke-tests/src/androidTest/backend/functions/functions/package.json
rlazo: I think the same

rlazo:
> This prob deserves a changelog entry btw

Since this only affects tests, it doesn't affect devs life

PR #5019, opened 2023-05-19
Javadoc fixes

rosalyntan:
Javadoc fixes in App Check and Functions.


rosalyntan:
@rlazo does this need a changelog? (i.e. should these Javadoc fixes trigger the SDK to be released if there are no other changes that go in?)

rlazo:
> @rlazo does this need a changelog? (i.e. should these Javadoc fixes trigger the SDK to be released if there are no other changes that go in?)

Not really. We could potentially re-generate the javadocs available if we really want and update them without triggering a release. Alternatively, we could wait until the next time the SDK is released and have the updates published then.

rlazo (approved):

PR #5020, opened 2023-05-20
Integrate Firebase Performance with Firebase Sessions.

visumickey:


firebase-perf/src/test/java/com/google/firebase/perf/FirebasePerformanceTest.java
```diff
@@ -573,6 +574,7 @@ private FirebasePerformance initializeFirebasePerformancePreferences(
         mock(FirebaseInstallationsApi.class),
         transportFactoryProvider,
         spyRemoteConfigManager,
+        mock(FirebaseSessions.class),
```
mrober, line 577: How well does this mock? I thought classes in Kotlin were `final` by default and couldn't be mocked?

visumickey: Firebase Performance depends on [Mockito-Inline](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-perf/firebase-perf.gradle#L124) which allows for Kotlin final classes to be mocked. [Reference](https://stackoverflow.com/a/63233687)

mrober (approved):

PR #5028, opened 2023-05-23
Clean up how we launch suspend functions

mrober:
Make `initiateSessionStart` a suspend function to clean up how we launch other suspend functions.

This will make it very easy to have suspend functions deeper in our code. Anything `initiateSessionStart` calls can easily be a suspend function now. And we can use `withContext` to change context, e.g. to blocking for settings.

Added a `SessionInitiateListener` interface so we don't have to pass around an awkward type for the suspend function.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -55,16 +55,24 @@ internal constructor(
 
   init {
     sessionSettings.updateSettings()
+
+    val sessionInitiateListener =
+      object : SessionInitiateListener {
+        // Avoid making a public function in FirebaseSessions for onInitiateSession.
+        override suspend fun onInitiateSession(sessionDetails: SessionDetails) {
+          initiateSessionStart(sessionDetails)
+        }
+      }
+
     val sessionInitiator =
       SessionInitiator(
         timeProvider,
         backgroundDispatcher,
-        object : SessionStartListener {
-          // Avoid making a public function in FirebaseSessions for onSessionStart.
-          override suspend fun onSessionStart() = initiateSessionStart()
-        },
+        sessionInitiateListener,
         sessionSettings,
+        sessionGenerator,
       )
+
     val appContext = firebaseApp.applicationContext.applicationContext
     if (appContext is Application) {
       appContext.registerActivityLifecycleCallbacks(sessionInitiator.activityLifecycleCallbacks)
@@ -87,9 +95,7 @@ internal constructor(
     )
   }
 
-  private suspend fun initiateSessionStart() {
-    val sessionDetails = sessionGenerator.generateNewSession()
-
+  private suspend fun initiateSessionStart(sessionDetails: SessionDetails) {
     val subscribers = FirebaseSessionsDependencies.getRegisteredSubscribers()
 
     if (subscribers.isEmpty()) {
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -29,6 +29,7 @@ import com.google.firebase.sessions.api.SessionSubscriber
 import com.google.firebase.sessions.settings.SessionsSettings
 import kotlinx.coroutines.CoroutineDispatcher
 
+/** The [FirebaseSessions] API provides methods to register a [SessionSubscriber]. */
 class FirebaseSessions
 internal constructor(
   private val firebaseApp: FirebaseApp,
@@ -46,10 +47,11 @@ internal constructor(
       firebaseInstallations,
       applicationInfo,
     )
-  private val sessionGenerator = SessionGenerator(collectEvents = shouldCollectEvents())
+  private val timeProvider: TimeProvider = Time()
+  private val sessionGenerator =
+    SessionGenerator(collectEvents = shouldCollectEvents(), timeProvider)
   private val eventGDTLogger = EventGDTLogger(transportFactoryProvider)
   private val sessionCoordinator = SessionCoordinator(firebaseInstallations, eventGDTLogger)
-  private val timeProvider: TimeProvider = Time()
 
   init {
     sessionSettings.updateSettings()
@@ -116,7 +118,7 @@ internal constructor(
     }
 
     sessionCoordinator.attemptLoggingSessionEvent(
-      SessionEvents.startSession(firebaseApp, sessionDetails, sessionSettings, timeProvider)
+      SessionEvents.startSession(firebaseApp, sessionDetails, sessionSettings)
     )
   }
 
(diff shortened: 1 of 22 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -46,18 +44,24 @@ internal constructor(
...
-  private val sessionStartScope = CoroutineScope(backgroundDispatcher)
 
   init {
     sessionSettings.updateSettings()
     val sessionInitiator =
-      SessionInitiator(timeProvider, this::initiateSessionStart, sessionSettings)
+      SessionInitiator(
+        timeProvider,
+        backgroundDispatcher,
+        object : SessionStartListener {
```
mrober, line 60: This is so we don't have to make `onSessionStart` public.

samedson: Can we add a comment for that?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -80,42 +84,40 @@ internal constructor(
...
-          subscriber.onSessionChanged(SessionSubscriber.SessionDetails(sessionDetails.sessionId))
-        }
-      }
+    Log.d(TAG, "Data Collection is enabled for at least one Subscriber")
 
-      if (!sessionGenerator.collectEvents) {
-        Log.d(TAG, "Sessions SDK has sampled this session")
-        return@launch
+    subscribers.values.forEach { subscriber ->
+      if (subscriber.isDataCollectionEnabled) {
```
samedson, line 108: We should still notify subscribers when their data collection is turned off, and when all subscriber data collection is turned off. 

We can move this up to line 99 and remove this if statement.

It might also be good to put a comment here stating that we should notify subscribers regardless of sampling and data collection state

mrober: Done

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionInitiatorTest.kt
```diff
@@ -140,19 +170,30 @@ class SessionInitiatorTest {
...
     val sessionInitiator =
-      SessionInitiator(fakeTimeProvider, sessionStartCounter::initiateSessionStart, settings)
+      SessionInitiator(
+        fakeTimeProvider,
+        TestOnlyExecutors.background().asCoroutineDispatcher() + coroutineContext,
+        sessionStartListener = sessionStartCounter,
+        settings,
+      )
+
+    runCurrent()
```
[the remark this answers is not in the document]
mrober: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -80,42 +85,39 @@ internal constructor(
     )
   }
 
-  private fun initiateSessionStart() {
+  private suspend fun initiateSessionStart() {
```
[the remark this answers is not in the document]
mrober: I refactored this so the `SessionDetails` has the timestamp. Then the `SessionInitiator` will generate the session on the main thread so the timestamp is as current as possible, and pass it in to the `SessionInitiateListener.onInitiateSession` suspend function.

samedson (approved):
🚢

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -88,42 +103,37 @@ internal constructor(
...
-      if (!sessionGenerator.collectEvents) {
-        Log.d(TAG, "Sessions SDK has sampled this session")
-        return@launch
-      }
+    Log.d(TAG, "Data Collection is enabled for at least one Subscriber")
 
-      sessionCoordinator.attemptLoggingSessionEvent(
-        SessionEvents.startSession(firebaseApp, sessionDetails, sessionSettings, timeProvider)
-      )
+    if (!sessionGenerator.collectEvents) {
```
visumickey, line 129: Just validating the functionality here: Even if the session data collection is sampled, we will be generating sessions, notifying the dependencies of the sessionInfo, but just that the session will not be dispatched?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -20,23 +20,21 @@ import java.util.UUID
...
 /**
  * [SessionDetails] is a data class responsible for storing information about the current Session.
- *
- * @hide
  */
 internal data class SessionDetails(
   val sessionId: String,
   val firstSessionId: String,
   val sessionIndex: Int,
+  val sessionTimestampUs: Long,
```
visumickey, line 28: Let's call it `sessionStartTimestampUs` for clarity sake.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/EventGDTLoggerTest.kt
```diff
@@ -26,16 +26,17 @@ import com.google.firebase.sessions.settings.SessionsSettings
...
 import com.google.firebase.sessions.testing.TestSessionEventData
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.test.runTest
 import org.junit.After
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 
+@OptIn(ExperimentalCoroutinesApi::class)
```
visumickey, line 39: What is this for?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -46,18 +45,34 @@ internal constructor(
       blockingDispatcher,
       backgroundDispatcher,
       firebaseInstallations,
-      applicationInfo
+      applicationInfo,
```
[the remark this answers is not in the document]
mrober: That's a style thing. I like the trailing `,` because when we add another line, we don't have to also edit the previous last line. Kotlin style encourages it https://kotlinlang.org/docs/coding-conventions.html#trailing-commas

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: That's right.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/EventGDTLoggerTest.kt
mrober: The test I edited calls `runTest`, which is marked as `@ExperimentalCoroutinesApi`. Since I touched the file, I also opted in to suppress the warning.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
mrober: Done.

PR #5031, opened 2023-05-24
Fix behaviour with multiple subscribers

mrober:
Immediately call the callback if Sessions generated a session before the subscriber subscribed. iOS does this too. This will fix the `FirebaseSessionsTest` when having multiple subscribers in the test app.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
@@ -46,6 +46,10 @@ internal class SessionGenerator(
   lateinit var currentSession: SessionDetails
     private set
 
+  /** Returns if a session has been generated. */
+  val hasGenerateSession: Boolean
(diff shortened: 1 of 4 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
```diff
@@ -46,6 +46,10 @@ internal class SessionGenerator(
   lateinit var currentSession: SessionDetails
     private set
 
+  /** Returns if a session has been generated. */
+  val hasGenerateSession: Boolean
```
visumickey, line 50: Where do we use this value?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt
mrober: This is used as the condition to send the current session when a subscriber subscribes in `FirebaseSessions`.

PR #5034, opened 2023-05-25
Fix equality and inequality on the same field

wu-hui:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
```diff
@@ -165,13 +167,17 @@ public boolean servedByIndex(FieldIndex index) {
...
-      if (!matchesFilter(inequalityFilter, segment) || !matchesOrderBy(orderBys.next(), segment)) {
-        return false;
+      // If there is an inequality filter and the field was not in one of the equality filters
+      // above, the next segment must match both the filter and the first orderBy clause.
+      if (!equalitySegments.contains(inequalityFilter.getField().canonicalString())) {
+        FieldIndex.Segment segment = segments.get(segmentIndex);
+        if (!matchesFilter(inequalityFilter, segment)
+            || !matchesOrderBy(orderBys.next(), segment)) {
+          return false;
+        }
```
cherylEnkidu, line 178: This code block seems to allow the case like
```java
query("collId")
            .filter(filter("a", ">=", 5))
            .filter(filter("a", "==", 0))
            .orderBy(orderBy("b"))
```
But shouldn't it only allow case like
```java
query("collId")
            .filter(filter("a", ">=", 5))
            .filter(filter("a", "==", 0))
            .orderBy(orderBy("a"))
            .orderBy(orderBy("b"))
```

cherylEnkidu (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
wu-hui: First,

```
query("collId")
            .filter(filter("a", ">=", 5))
            .filter(filter("a", "==", 0))
            .orderBy(orderBy("b"))
```
is an illegal firestore query, the first orderby needs to be the same as the inequality.

Second, even in the world where this is a legal query, the index can serve the query still, because the existence of the equality here. An equality and an inequality on the same field is essentially an equality at the end of the day.

cherylEnkidu: Thank you for your explanation!

cherylEnkidu (approved):

PR #5045, opened 2023-05-31
Add and populate appQualitySessionId in Crashlytics reports

mrober:
Add and populate appQualitySessionId in Crashlytics reports. This works by saving the latest AQS session id to disk on every event, which is slightly different than iOS which saves it on every session change. The JVM can save to disk in the exception handler, so it's ok to do it this way for Android. Then when a report is finalized, take the most recent AQS session id for that Crashlytics session and include it in the report.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
@@ -142,7 +142,9 @@ public class CrashlyticsReportPersistence {
       writeTextFile(fileStore.getSessionFile(sessionId, fileName), json);
 
       String appQualitySessionId = sessionsSubscriber.getAppQualitySessionId();
-      if (appQualitySessionId != null) {
+      if (appQualitySessionId == null) {
+        Logger.getLogger().w("Missing AQS session id for Crashlytics session " + sessionId);
+      } else {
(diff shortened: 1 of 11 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
```diff
@@ -132,8 +140,14 @@ public void persistEvent(
...
     try {
       writeTextFile(fileStore.getSessionFile(sessionId, fileName), json);
-    } catch (IOException e) {
-      Logger.getLogger().w("Could not persist event for session " + sessionId, e);
+
+      String appQualitySessionId = sessionsSubscriber.getAppQualitySessionId();
+      if (appQualitySessionId != null) {
+        writeTextFile(
+            fileStore.getSessionFile(sessionId, AQS_SESSION_ID_FILENAME), appQualitySessionId);
+      }
```
samedson, line 148: Should we log a warning when it's null?

samedson (approved):
This looks good - just one suggestion

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistence.java
mrober: Done.

PR #5050, opened 2023-06-01
NO_RELEASE_CHANGE pin protolite version for transitive dep

davidmotson:


protolite-well-known-types/gradle.properties
```diff
@@ -1,2 +1,3 @@
-version=18.0.1
+# This is done to prevent issues with transitive dependencies
```
rlazo, line 1: Create an internal tracking bug to reference here. Then you can make it look more like a TODO (suggested wording)

# IMPORTANT (b/1111111): Keep version and latestReleasedVersion in sync unless you are releasing a new version of the library to prevent issues with transitive dependencies.

daymxn (approved):

PR #5056, opened 2023-06-06
Pin perf dependency on firebase-datatransport

rlazo:


visumickey:
The change looks good to me. Can you explain why this is necessary?

visumickey (approved):

PR #5059, opened 2023-06-07
Add different default sampling rates for when RC fetch failed

mrober:


firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigResolver.java
```diff
@@ -331,7 +338,12 @@ public double getNetworkRequestSamplingRate() {
       return deviceCacheValue.get();
     }
 
-    // 3. Returns default value if there is no valid value from above approaches.
+    // 3. Check if RC fetch failed.
+    if (remoteConfigManager.isLastFetchFailed()) {
```
visumickey, line 342: What is the meaning of lastFetchFailed? Are we accounting for any kind of a fetch failure to fallback to these defaults?

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigurationConstants.java
```diff
@@ -218,6 +230,11 @@ protected Long getDefault() {
       return 100L;
     }
 
+    @Override
+    protected Long getDefaultOnRcFetchFail() {
+      return getDefault() * 3;
```
visumickey, line 235: Let's add a comment to say that we are increasing the default 3 folds. Eg: 100ms to 300ms.

firebase-perf/src/test/java/com/google/firebase/perf/config/ConfigurationConstantsTest.java
```diff
@@ -91,6 +91,7 @@ public void getInstance_SessionsCpuCaptureFrequencyForegroundMs_validateConstant
         SessionsCpuCaptureFrequencyForegroundMs.getInstance();
 
     assertThat(configFlag.getDefault()).isEqualTo(100L);
+    assertThat(configFlag.getDefaultOnRcFetchFail()).isEqualTo(300L);
```
visumickey, line 94: Would we have such asserts for the sampling rates?

visumickey (approved):
Thank you for making those changes.

firebase-perf/src/main/java/com/google/firebase/perf/config/ConfigurationConstants.java
mrober: Done

firebase-perf/src/test/java/com/google/firebase/perf/config/ConfigurationConstantsTest.java
mrober: Added.

PR #5060, opened 2023-06-07
Use AQS provided sessionId for Fireperf events.

visumickey:


firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
```diff
@@ -194,6 +195,7 @@ public static FirebasePerformance getInstance() {
     this.configResolver = configResolver;
     this.configResolver.setMetadataBundle(mMetadataBundle);
     this.configResolver.setApplicationContext(appContext);
+    // TODO (Should this be done after the register call to firebase sessions?)
     sessionManager.setApplicationContext(appContext);
```
samedson, line 199: I guess what is the impact of calling this?

mrober (approved):

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerformance.java
visumickey: The actual impact here is that the sessionID relevant to app start happens when the applicationContext is set. With SessionsRegistration done later after setting applicationContext, it would happen that app start trace might not have a sessionID associated with it.

samedson: I see so in the case of a crash or something interrupting it

samedson (approved):

samedson (approved):
LGTM

mrober (approved):

PR #5061, opened 2023-06-07
Add check if Sessions is disabled

mrober:


samedson (approved):
🚢

visumickey (commented):
Do we have tests to validate this change? Basically I'm surprised that we did not have tests where we validate that we do not sent events when sessions SDK is disabled. We should add tests to ensure that this scenario is tested.

PR #5064, opened 2023-06-07
Auto index Creation Experiment

cherylEnkidu:
This experiment provides the number used in `RELATIVE_INDEX_READ_COST_PER_DOCUMENT ` for #4987  . 
It runs query locally on collection contains 10 ~ 100 documents, compares the time spent between full collection scan and index look up. And the result relative number is used in Android SDK to decide whether index will help improve local query performance.


wu-hui:
Experiment looks good!

PR #5068, opened 2023-06-08
Fix firebase-crashlytics-ktx instrumented tests

mrober:
This is needed because we upped the target sdk and firebase-crashlytics-ktx had been using some deprecated apis.

### DIFF
--- a/firebase-crashlytics/ktx/ktx.gradle
+++ b/firebase-crashlytics/ktx/ktx.gradle
@@ -53,14 +53,14 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.1.0'
 
     testImplementation "org.robolectric:robolectric:$robolectricVersion"
-    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.13.2'
     testImplementation "com.google.truth:truth:$googleTruthVersion"
-    testImplementation 'org.mockito:mockito-core:2.25.0'
+    testImplementation 'org.mockito:mockito-core:3.4.3'
 
-    androidTestImplementation 'junit:junit:4.12'
+    androidTestImplementation 'junit:junit:4.13.2'
     androidTestImplementation "com.google.truth:truth:$googleTruthVersion"
-    androidTestImplementation 'androidx.test:runner:1.2.0'
-    androidTestImplementation 'androidx.test:core:1.2.0'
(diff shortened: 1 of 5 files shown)


firebase-crashlytics/ktx/ktx.gradle
```diff
@@ -13,59 +13,43 @@
...
-        }
-        androidTest.java.srcDirs += 'src/androidTest/kotlin'
-    }
-    testOptions.unitTests.includeAndroidResources = true
+  compileSdk 33
+  defaultConfig {
+    minSdk 16
+    multiDexEnabled true
+    targetSdk 33
+    testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
```
visumickey, line 33: Did we miss `versionName` here?

firebase-crashlytics/ktx/ktx.gradle
```diff
@@ -13,59 +13,43 @@
...
-    testImplementation "org.robolectric:robolectric:$robolectricVersion"
-    testImplementation 'junit:junit:4.12'
-    testImplementation "com.google.truth:truth:$googleTruthVersion"
-    testImplementation 'org.mockito:mockito-core:2.25.0'
-
-    androidTestImplementation 'junit:junit:4.12'
-    androidTestImplementation "com.google.truth:truth:$googleTruthVersion"
-    androidTestImplementation 'androidx.test:runner:1.2.0'
-    androidTestImplementation 'androidx.test:core:1.2.0'
+  implementation "com.google.firebase:firebase-common-ktx:20.3.2"
```
visumickey, line 46: Seems like we have pruned a lot of dependencies. Can we ensure that there are no functionality breakage here? I'm guessing many of these changes need to be directly done in the `master` branch?

visumickey (approved):

firebase-crashlytics/ktx/ktx.gradle
mrober: Yeah I think this build file was originally copied from the perf ktx build file, and a lot of the deps that were copied were never used. For example, all the testImplementation deps are not needed since there are no unit tests, only android tests.

firebase-crashlytics/ktx/ktx.gradle
mrober: Fixed

PR #5078, opened 2023-06-15
Do not require sign in to use developer mode

lfkellogg:
The docs say that dev mode allows you to test in-app feedback without actually distributing your app. However that's not true because you still have to sign into the SDK.

This expands dev mode to allow testing the SDK in a signed in state without actually signing in:
- `signInTester()` skips opening the browser and immediately marks the tester as signed in
- Sign in state is managed using a separate key in app storage

Also, this expands dev mode to cover the in-app update functionality as well. It simply disables the check for new release and prints a log message instead.

### DIFF
--- a/firebase-appdistribution/CHANGELOG.md
+++ b/firebase-appdistribution/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+* [feature] Improved development mode to allow all API calls to be made without having to sign in.
 
 # 16.0.0-beta08
 * [fixed] Fixed an issue where a crash happened whenever a feedback
--- a/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/DevModeDetector.java
+++ b/firebase-appdistribution/src/main/java/com/google/firebase/appdistribution/impl/DevModeDetector.java
@@ -1,0 +1,50 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.appdistribution.impl;
+
+import androidx.annotation.Nullable;
+import java.lang.reflect.Method;
+import javax.inject.Inject;
+
+/**
+ * A detector to recognize when the SDK is in development mode, which disables any functionality
+ * that requires the tester to have signed in.
+ */
+class DevModeDetector {
+
+  @Inject
+  DevModeDetector() {}
+
+  boolean isDevModeEnabled() {
+    return Boolean.valueOf(getSystemProperty("debug.firebase.appdistro.devmode"));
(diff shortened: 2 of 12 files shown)


firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -45,15 +45,26 @@ android {
         prod {
             dimension "environment"
             versionNameSuffix "-prod"
+
+            firebaseAppDistribution {
+                artifactType "AAB"
+                releaseNotes "Prod variant of Android SDK test app"
+                // testers "your email here"
+//                testers "edmuregreyjoy.190269@gmail.com, robbbaelish.960619@gmail.com, yarastone.537949@gmail.com"
+                testers "lkdogfooding@gmail.com"
```
kaibolay, line 54: you probably don't want this change...

lfkellogg: Eek!

firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -73,6 +84,7 @@ dependencies {
 
     // All variants use the API
     implementation project(':firebase-appdistribution-api:ktx')
+//    implementation 'com.google.firebase:firebase-appdistribution-api-ktx:16.0.0-beta08'
```
kaibolay, line 87: delete outright instead of adding comment?

firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -81,6 +93,7 @@ dependencies {
 
     // Beta flavor uses the full implementation
     betaImplementation project(':firebase-appdistribution')
+//    betaImplementation 'com.google.firebase:firebase-appdistribution:16.0.0-beta08'
```
kaibolay, line 96: delete outright instead of adding comment?

firebase-appdistribution/test-app/test-app.gradle
```diff
@@ -45,15 +45,26 @@ android {
...
         beta {
             dimension "environment"
             versionNameSuffix "-beta"
 
             firebaseAppDistribution {
+                artifactType "AAB"
                 releaseNotes "Beta variant of Android SDK test app"
                 // testers "your email here"
+//                testers "edmuregreyjoy.190269@gmail.com, robbbaelish.960619@gmail.com, yarastone.537949@gmail.com"
+                testers "lkdogfooding@gmail.com"
```
kaibolay, line 67: ditto

subprojects.cfg
```diff
@@ -81,4 +81,4 @@ transport:transport-runtime-testing
 
 #firebase-storage:test-app
 #appcheck:firebase-appcheck:test-app
-#firebase-appdistribution:test-app
+firebase-appdistribution:test-app
```
kaibolay, line 84: I *think* this change should not be committed because of reasons

firebase-appdistribution/test-app/test-app.gradle
lfkellogg: Did not mean to commit any of this, sorry

subprojects.cfg
lfkellogg: Sorry yes please refresh

firebase-appdistribution/test-app/test-app.gradle
lfkellogg: Sorry yes please refresh

firebase-appdistribution/test-app/test-app.gradle
lfkellogg: Sorry yes please refresh

kaibolay (approved):

PR #5093, opened 2023-06-21
Refactor real-time RC to use tasks for auth token requests

qdpham13:
Update real-time http methods to use Tasks for auth tokens and FID (future).

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -480,10 +480,12 @@ public class ConfigRealtimeHttpClient {
               setIsHttpConnectionRunning(true);
 
               try {
+                // If HTTP connection task failed throw exception to move to the catch block.
                 if (!httpURLConnectionTask.isSuccessful()) {
                   throw new IOException(httpURLConnectionTask.getException());
                 }
 
+                // Get HTTP connection and check response code.
                 httpURLConnection = httpURLConnectionTask.getResult();
                 responseCode = httpURLConnection.getResponseCode();
 
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -172,7 +172,10 @@ public class ConfigRealtimeHttpClient {
     }
   }
 
-  private void setCommonRequestHeaders(HttpURLConnection httpURLConnection) {
+  private void setCommonRequestHeaders(HttpURLConnection httpURLConnection, String authToken) {
+    // Auth token
+    httpURLConnection.setRequestProperty(INSTALLATIONS_AUTH_TOKEN_HEADER, authToken);
+
     // API Key
     httpURLConnection.setRequestProperty(API_KEY_HEADER, this.firebaseApp.getOptions().getApiKey());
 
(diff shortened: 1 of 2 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -315,13 +304,37 @@ private URL getUrl() {
...
+    Task<InstallationTokenResult> installationAuthTokenTask = firebaseInstallations.getToken(false);
+    // When the auth token task has finished, set up HTTP connection with headers and params.
+    return Tasks.whenAllComplete(installationAuthTokenTask)
+        .continueWithTask(
+            this.scheduledExecutorService,
+            (completedInstallationTask) -> {
+              if (!installationAuthTokenTask.isSuccessful()) {
+                return Tasks.forException(
+                    new FirebaseRemoteConfigClientException(
+                        "Firebase Installations failed to get installation ID for real-time.",
```
danasilver, line 318: I think the exception should be `installation auth token` here

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -455,59 +468,83 @@ public void beginRealtimeHttpStream() {
...
+            this.scheduledExecutorService,
+            (completedHttpUrlConnectionTask) -> {
+              Integer responseCode = null;
+              HttpURLConnection httpURLConnection = null;
+              setIsHttpConnectionRunning(true);
+
+              try {
+                // If HTTP connection task failed throw exception to move to the catch block.
+                if (!httpURLConnectionTask.isSuccessful()) {
+                  throw new IOException(httpURLConnectionTask.getException());
```
danasilver, line 485: Thinking: More refactoring would propagate the failed task, but we're limiting the scope of this change. It's a bit confusing to throw the `IOException` here just to handle it a few lines below.

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -455,59 +468,83 @@ public void beginRealtimeHttpStream() {
...
+    // Make async call to create and setup HTTP connection.
+    Task<HttpURLConnection> httpURLConnectionTask = createRealtimeConnection();
+    // When the connection task has finished, begin real-time actions.
+    Tasks.whenAllComplete(httpURLConnectionTask)
+        .continueWith(
+            this.scheduledExecutorService,
+            (completedHttpUrlConnectionTask) -> {
+              Integer responseCode = null;
+              HttpURLConnection httpURLConnection = null;
+              setIsHttpConnectionRunning(true);
```
danasilver, line 480: Should this call happen after checking that the task completed successfully?

danasilver (approved):
Couple thoughts and one comment/question.

danasilver (approved):

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: Yeah I think more refactoring can be done to handle the failed task better, I threw the `IOException` here b/c I wanted to keep the same try/catch structure that we had here and to reduce the scope of these changes

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
qdpham13: updated

Issue #5091, opened 2023-06-21
Fatal Exception: java.util.ConcurrentModificationException


muriloventuroso:
While running the Android app, the following error was reported in the crash log.

```
Exception java.util.ConcurrentModificationException:
  at java.util.ArrayList$Itr.next (ArrayList.java:860)
  at com.google.firebase.firestore.core.Query.matchesOrderBy (Query.java:401)
  at com.google.firebase.firestore.core.Query.matches (Query.java:425)
  at com.google.firebase.firestore.local.SQLiteRemoteDocumentCache.lambda$getDocumentsMatchingQuery$3 (SQLiteRemoteDocumentCache.java:257)
  at com.google.firebase.firestore.local.SQLiteRemoteDocumentCache.lambda$processRowInBackground$2 (SQLiteRemoteDocumentCache.java:242)
  at com.google.firebase.firestore.util.BackgroundQueue.lambda$execute$0 (BackgroundQueue.java:39)
  at com.google.firebase.firestore.util.ThrottledForwardingExecutor.lambda$execute$0 (ThrottledForwardingExecutor.java:54)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:919)
```

### Steps to reproduce

I still haven't been able to reproduce the error on my Android devices or the emulator. But I get this error report everyday from my app users.

## Additional context

Occurs on Android 10 (SDK 29)
Firebase Android SDK 32.0.0

muriloventuroso:
Another trace:

```
Exception java.lang.RuntimeException:
  at com.google.firebase.firestore.util.AsyncQueue.lambda$panic$3 (AsyncQueue.java:545)
  at android.os.Handler.handleCallback (Handler.java:883)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at android.os.Looper.loop (Looper.java:214)
  at android.app.ActivityThread.main (ActivityThread.java:7386)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:492)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:980)
Caused by java.lang.RuntimeException: java.util.ConcurrentModificationException
[...]

PR #5099, opened 2023-06-22
Making methods in core.Query that memoize results thread safe.

MarkDuckworth:


firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -325,11 +325,16 @@ public List<OrderBy> getExplicitOrderBy() {
   }
 
   /**
-   * Returns the full list of ordering constraints on the query.
+   * Returns the full list of ordering constraints on the query. This might include additional
+   * sort orders added implicitly to match the backend behavior.
    *
-   * <p>This might include additional sort orders added implicitly to match the backend behavior.
+   * This method is marked as synchronized to prevent the memoizedOrderBy from being
```
wu-hui, line 331: I think in comments we should avoid talking about implementation details too much, some remarks like "This is synchronized because it actually modifies internal state in some cases" should be good enough.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -325,11 +325,16 @@ public List<OrderBy> getExplicitOrderBy() {
...
+   * created in multiple threads.
+   *
+   * The returned list is unmodifiable, to prevent ConcurrentModificationExceptions,
```
wu-hui, line 334: It might be worthwhile to changed the return type, and delete this statement.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -458,8 +464,13 @@ public int compare(Document doc1, Document doc2) {
     }
   }
 
-  /** @return A {@code Target} instance this query will be mapped to in backend and local store. */
-  public Target toTarget() {
+  /**
+   * This method is marked as synchronized to prevent the memoizedTarget from being
```
wu-hui, line 468: Similarly to above comment, we can just state that this might change internal state.

firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
```diff
@@ -752,6 +763,119 @@ public void testOrQuery() {
         /* not match */ Arrays.asList(doc1, doc2, doc4, doc5));
   }
 
+  @Test
+  public void testSynchronousMatchesOrderBy() {
```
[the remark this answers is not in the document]
MarkDuckworth: I was unable to reproduce the issue in https://github.com/firebase/firebase-android-sdk/issues/5091, but I did observe a different race condition when debugging this test. However, I believe that this change will fix 5091 or lead us to find the root cause of the ConcurrentModificationException.

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
MarkDuckworth: Updated

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
MarkDuckworth: updated

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
MarkDuckworth: We can't change the return type since Collection.unmodifiableList(...) returns a List<T>. But I did add one unit test to assert the returned list is unmodifiable.

wu-hui (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java
```diff
@@ -362,7 +367,7 @@ public List<OrderBy> getOrderBy() {
                   : Direction.ASCENDING;
           res.add(lastDirection.equals(Direction.ASCENDING) ? KEY_ORDERING_ASC : KEY_ORDERING_DESC);
         }
-        memoizedOrderBy = res;
+        memoizedOrderBy = Collections.unmodifiableList(res);
```
Georgezhang714, line 370: #5099

firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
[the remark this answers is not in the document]
Georgezhang714: > I was unable to reproduce the issue in https://github.com/firebase/firebase-android-sdk/issues/5091, but I did observe a different race condition when debugging this test. However, I believe that this change will fix 5091 or lead us to find the root cause of the ConcurrentModificationException.

Georgezhang714 (changes requested):

PR #5107, opened 2023-06-26
Fetch remote settings on session, not on launch

mrober:
Fetch remote settings on session, not on launch. This will make the sdk feel lighter by preventing network activity from affecting app launch. This will also prevent fetching the fid until a session is being sent, meaning data collection was enabled for at least one sdk.


firebase-sessions/src/test/kotlin/com/google/firebase/sessions/settings/RemoteSettingsTest.kt
```diff
@@ -225,6 +230,66 @@ class RemoteSettingsTest {
       remoteSettings.clearCachedSettings()
     }
 
+  @Test
+  fun remoteSettings_fetchWhileFetchInProgress() =
```
samedson, line 234: This is a great test 👍

samedson (approved):
Just one suggestion

PR #5117, opened 2023-06-27
specify the bom dir as an output property of the BomGeneratorTask

davidmotson:
[b/285900964](https://b.corp.google.com/issues/285900964)


daymxn (approved):
Looks great! As long as you've manually tested it and it works fine, LGTM:)

PR #5133, opened 2023-07-05
Update semver tests and the pom validator tests to handle new sdk

VinayGuthal:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/PomValidator.kt
```diff
@@ -43,10 +43,15 @@ abstract class PomValidator : DefaultTask() {
 
   @TaskAction
   fun run() {
-    var diff = diffWithPomFromURL(getLatestReleasePomUrl())
+    try {
+      var diff = diffWithPomFromURL(getLatestReleasePomUrl())
```
daymxn, line 47: I realize I accidentally left this as `var`- this should probably be changed to `val`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.kt
```diff
@@ -109,11 +109,13 @@ class FirebaseLibraryPlugin : BaseFirebaseLibraryPlugin() {
       from(project.zipTree("build/outputs/aar/${releaseAar}"))
       into(project.file("semver/current-version"))
     }
+
     project.tasks.register<Copy>("extractPreviousClasses") {
       dependsOn("copyPreviousArtifacts")
-
-      from(project.zipTree("semver/previous.aar"))
-      into(project.file("semver/previous-version"))
+      if (project.file("semver/previous.aar").exists()) {
```
daymxn, line 115: Keep in mind that if this file is added by any task- this will fail. This check will occur at configuration time, not execution time.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/semver/GmavenCopier.kt
```diff
@@ -41,6 +41,11 @@ abstract class GmavenCopier : DefaultTask() {
         mavenHelper.getLatestReleasedVersion(),
         !aarAndroidFile.get()
       )
-    URL(gMavenPath).openStream().use { Files.copy(it, Paths.get(filePath.get())) }
+    try {
+      URL(gMavenPath).openStream().use { Files.copy(it, Paths.get(filePath.get())) }
+    } catch (_: java.io.FileNotFoundException) {
+      // Gmaven Artifact doesn't exist.
```
daymxn, line 47: You should be fine to replace these empty returns with `StopActionException` for more information.
```kotlin
throw StopActionException("GMaven artifact doesn't exist")
```

daymxn (approved):
left some comments- they aren't major; so if you need to get this out asap go ahead:)

PR #5135, opened 2023-07-05
Firestore: add test bloomFilterShouldCorrectlyEncodeComplexUnicodeCharacters() to QueryTest.java

dconeybe:
Add test `bloomFilterShouldCorrectlyEncodeComplexUnicodeCharacters()` to `QueryTest.java`. This test verifies that the bloom filter correctly encodes document paths that contain special Unicode characters, like combining characters and characters above 0xFFFF (i.e. characters outside of the BMP, basic multilingual plane).

Googlers see b/289234093 for details.

This is a port of https://github.com/firebase/firebase-js-sdk/pull/7412

PR #5145, opened 2023-07-07
Firestore: Update test bloomFilterShouldCorrectlyEncodeComplexUnicodeCharacters() in QueryTest.java

dconeybe:
Modify the test `bloomFilterShouldCorrectlyEncodeComplexUnicodeCharacters()` in `QueryTest.java`, that was added by the recent PR https://github.com/firebase/firebase-android-sdk/pull/5135, to directly check the bloom filter for containing the document paths with complex Unicode characters. Previously, it was indirectly checking this via the other properties of the existence filter. As a result, the "retry" logic was removed, since it doesn't matter if there is a false positive.

To add the ability to directly interrogate the bloom filter, the testing hooks needed to be augmented to expose the  `BloomFilter`, projectId, and databaseId.

This is a port of https://github.com/firebase/firebase-js-sdk/pull/7413

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -1220,109 +1220,104 @@ public class QueryTest {
       testDocs.put(docId, map("foo", 42));
     }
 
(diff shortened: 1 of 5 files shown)


milaGGL (approved):
LGTM

PR #5162, opened 2023-07-14
Support deleteAllIndexes in PersistentCacheIndexManager

cherylEnkidu:


dconeybe (commented):
LGTM; however, I'll wait for someone more familiar with this code to give the formal LGTM.

firebase-firestore/src/main/java/com/google/firebase/firestore/PersistentCacheIndexManager.java
```diff
@@ -52,4 +52,12 @@ public void enableIndexAutoCreation() {
   public void disableIndexAutoCreation() {
     client.disableIndexAutoCreation();
   }
+
+  /**
+   * Removes all persistent cache indexes. Please note this function will also deletes indexes
+   * generated by {@link FirebaseFirestore#setIndexConfiguration(String)}.
```
wu-hui, line 58: We are removing this method, so no need to mention here.

wu-hui (approved):
LGTM with minor nit.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
dconeybe, line 1: Is there a test that calls deleteAllIndexes() after terminate()? If not, could you add one?

dconeybe (changes requested):

firebase-firestore/src/main/java/com/google/firebase/firestore/PersistentCacheIndexManager.java
cherylEnkidu: however when the feature releases, the deprecated setIndexConfiguration is still available to use. I would like to add notes mentioning the function is deprecated. Please let me know if you feel strong I should remove this comment.

PR #5165, opened 2023-07-14
Connect PostReleasePlugin with PublishingPlugin

daymxn:
Per [b/274794264](https://b.corp.google.com/issues/274794264),

This connects the `PostReleasePlugin` plugin with releasing libraries, such that tasks like `VersionBumpTask` can be executed from the root directory- while only applying to releasing libraries.

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.kt
@@ -54,7 +54,6 @@ class FirebaseLibraryPlugin : BaseFirebaseLibraryPlugin() {
 
     project.apply<DackkaPlugin>()
     project.apply<GitSubmodulePlugin>()
-    project.apply<PostReleasePlugin>()
     project.tasks.getByName("preBuild").dependsOn("updateGitSubmodules")
   }
 
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/GradleUtils.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/GradleUtils.kt
@@ -58,6 +58,38 @@ fun Provider<File>.childFile(path: String) = map { File("${it.path}/$path") }
 fun File.childFile(childPath: String) = File("$path/$childPath")
 
 /**
+ * Rewrites the lines of a file.
+ *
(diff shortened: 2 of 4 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/GradleUtils.kt
```diff
@@ -57,6 +57,38 @@ fun Provider<File>.childFile(path: String) = map { File("${it.path}/$path") }
...
+ * file.rewriteLines { it.replace(" ", "\t") }
+ *
+ * // Rewrite the lines of the file, capitalizing the first letter of each word.
+ * file.rewriteLines { it.capitalizeWords() }
+ * ```
+ *
+ * @see [readLines]
+ * @see [writeText]
+ */
+fun File.rewriteLines(terminateWithNewline: Boolean = true, block: (String) -> String) {
```
davidmotson, line 87: the "terminateWithNewLine" param here feels like a YANGI smell.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.kt
```diff
@@ -54,7 +54,6 @@ class FirebaseLibraryPlugin : BaseFirebaseLibraryPlugin() {
 
     project.apply<DackkaPlugin>()
     project.apply<GitSubmodulePlugin>()
-    project.apply<PostReleasePlugin>()
```
davidmotson, line 57: I'm unsure I understand why this deletion is ok. 

Is the plugin just being applied differently? Is this offset by the change in PublishingPlugin.kt:98?

davidmotson (approved):

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GradleUtils.kt
daymxn: The fact that it exists, or its naming convention?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/FirebaseLibraryPlugin.kt
daymxn: > Is this offset by the change in PublishingPlugin.kt:98?

Exactly. Perviously, `PostReleasePlugin` was applied to every firebase lib. Now, it only gets applied to releasing libs.

PR #5167, opened 2023-07-14
Firestore: QueryTest.java: Use a WriteBatch to delete documents rather than a transaction

dconeybe:
This is a port of https://github.com/firebase/firebase-js-sdk/pull/7415


milaGGL (approved):
LGTM

PR #5170, opened 2023-07-17
Create ant.yml

Watemlifts:
- name: Setup Go environment uses: actions/setup-go@v4.0.1 with: # The Go version to download (if necessary) and use. Supports semver spec and ranges. go-version: # optional # Path to the go.mod or go.work file. go-version-file: # optional # Set this option to true if you want the action to always check for the latest available version that satisfies the version spec check-latest: # optional # Used to pull node distributions from go-versions. Since there's a default, this is typically not supplied by the user. When running this action on github.com, the default value is sufficient. When running on GHES, you can pass a personal access token for github.com if you are experiencing rate limiting. token: # optional, default is ${{ github.server_url == 'https://github.com' && github.token || '' }} # Used to specify whether caching is needed. Set to true, if you'd like to enable caching. cache: # optional, default is true # Used to specify the path to a dependency file - go.sum cache-dependency-path: # optional # Target architecture for Go to use. Examples: x86, x64. Will use system architecture by default. architecture: # optional

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -1,0 +1,42 @@
+# This workflow will build a Java project with Ant
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-ant
+
+name: Java CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


Watemlifts:
Push automatic

PR #5184, opened 2023-07-21
Firestore: QueryTest.java: use a different Firestore instance instead of a WriteBatch

dconeybe:
Use a different Firestore instance instead of a `WriteBatch` to avoid affecting the local cache in `QueryTest.java`.

I had originally thought that `WriteBatch` sidestepped the local cache, just like a transaction did; however, that turned out to be incorrect and using a `WriteBatch` nullified the test cases that were modified to use it. This is a follow-up to https://github.com/firebase/firebase-android-sdk/pull/5167.

This PR was ported from https://github.com/firebase/firebase-js-sdk/pull/7486.

### DIFF
--- a/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
+++ b/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
@@ -1077,17 +1077,18 @@ public class QueryTest {
       }
       assertWithMessage("createdDocuments").that(createdDocuments).hasSize(100);
 
-      // Delete 50 of the 100 documents. Use a WriteBatch, rather than DocumentReference.delete(),
-      // to avoid affecting the local cache.
+      // Delete 50 of the 100 documents. Use a different Firestore instance to avoid affecting the
+      // local cache.
       HashSet<String> deletedDocumentIds = new HashSet<>();
       {
-        WriteBatch writeBatchForDocumentDeletes = collection.getFirestore().batch();
+        FirebaseFirestore db2 = testFirestore();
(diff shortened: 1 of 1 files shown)


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1077,17 +1077,18 @@ public void resumingAQueryShouldUseBloomFilterToAvoidFullRequery() throws Except
...
       assertWithMessage("createdDocuments").that(createdDocuments).hasSize(100);
 
-      // Delete 50 of the 100 documents. Use a WriteBatch, rather than DocumentReference.delete(),
-      // to avoid affecting the local cache.
+      // Delete 50 of the 100 documents. Use a different Firestore instance to avoid affecting the
+      // local cache.
       HashSet<String> deletedDocumentIds = new HashSet<>();
       {
-        WriteBatch writeBatchForDocumentDeletes = collection.getFirestore().batch();
+        FirebaseFirestore db2 = testFirestore();
```
milaGGL, line 1084: Does this testFirestore() do the same? getting different instances to same firestore?

dconeybe: Yes! If curious about the details, see https://github.com/firebase/firebase-android-sdk/blob/d91342e4b919a89084f5369c47d047bbcc85b385/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java.

milaGGL (approved):
LGTM

PR #5192, opened 2023-07-24
Store rollouts state in cache

themiswang:
Create a `RolloutAssignment` model for storing the instance at UserMetadata in memory.
Implemented serialization and  de-serialization from json to data model and vice-versa (this is used for file read and write).

`RolloutAssignment` will not become a part of the proto in CrashlyticsReport because we might later port it to Interop SDK so both RC and Crash can leverage on.

#no-changelog

### DIFF
--- a/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStoreTest.java
+++ b/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStoreTest.java
@@ -14,15 +14,19 @@
 
 package com.google.firebase.crashlytics.internal.metadata;
 
+import com.google.common.truth.Truth;
 import com.google.firebase.crashlytics.internal.CrashlyticsTestCase;
 import com.google.firebase.crashlytics.internal.common.CrashlyticsBackgroundWorker;
 import com.google.firebase.crashlytics.internal.persistence.FileStore;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import org.junit.Test;
 
 @SuppressWarnings("ResultOfMethodCallIgnored") // Convenient use of files.
 public class MetaDataStoreTest extends CrashlyticsTestCase {
@@ -42,6 +46,14 @@ public class MetaDataStoreTest extends CrashlyticsTestCase {
   private static final String UNICODE = "あいうえおかきくけ";
 
   private static final String ESCAPED = "\ttest\nvalue";
+
+  private static final List<RolloutAssignment> ROLLOUTS_STATE = new ArrayList<>();
+
+  static {
+    RolloutAssignment assignment =
+        RolloutAssignment.create("rollout_1", "my_feature", "false", "control", 1);
+    ROLLOUTS_STATE.add(assignment);
+  }
 
   private FileStore fileStore;
   private final CrashlyticsBackgroundWorker worker = new CrashlyticsBackgroundWorker(Runnable::run);
@@ -280,6 +292,14 @@ public class MetaDataStoreTest extends CrashlyticsTestCase {
   public void testReadKeys_noStoredData() {
     final Map<String, String> readKeys = storeUnderTest.readKeyData(SESSION_ID_1);
     assertEquals(0, readKeys.size());
+  }
+
+  @Test
+  public void testWriteReadRolloutState() throws Exception {
+    storeUnderTest.writeRolloutState(SESSION_ID_1, ROLLOUTS_STATE);
+    List<RolloutAssignment> readRolloutsState = storeUnderTest.readRolloutsState(SESSION_ID_1);
+
+    Truth.assertThat(readRolloutsState).isEqualTo(ROLLOUTS_STATE);
   }
 
   public static void assertEqualMaps(Map<String, String> expected, Map<String, String> actual) {
--- a/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignmentListTest.java
+++ b/firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignmentListTest.java
@@ -1,0 +1,137 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.firebase.crashlytics.internal.metadata;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.firebase.crashlytics.internal.CrashlyticsTestCase;
+import com.google.firebase.encoders.DataEncoder;
+import com.google.firebase.encoders.json.JsonDataEncoderBuilder;
+import java.util.ArrayList;
+import java.util.List;
+import org.json.JSONException;
+import org.junit.Test;
+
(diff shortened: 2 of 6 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignment.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.firebase.encoders.annotations.Encodable;
+import java.util.Objects;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * model used in user metadata context which make rollout assignment serialize, deserialize easier
+ */
+@Encodable
+public class RolloutAssignment {
```
danasilver, line 26: Is the thinking that this will live in the Interop SDK (once we check that in)?

And wonder if it could use [AutoValue](https://github.com/google/auto/blob/main/value/userguide/index.md#in-your-value-class) to take care of the boilerplate?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStore.java
```diff
@@ -135,6 +140,46 @@ Map<String, String> readKeyData(String sessionId, boolean isInternal) {
     return Collections.emptyMap();
   }
 
+  public List<RolloutAssignment> readRolloutsState(String sessionId) {
```
samedson, line 143: Seems this is only used for testing? Can we add a comment about that?

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStoreTest.java
```diff
@@ -14,15 +14,19 @@
 
 package com.google.firebase.crashlytics.internal.metadata;
 
+import com.google.common.truth.Truth;
```
samedson, line 17: Should we be doing these changes in Kotlin? I think if we switch it should be easy to modify because there is a convert to kotlin tool in Android Studio. Understandable if we want to keep Java for consistency

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignmentList.java
```diff
@@ -0,0 +1,55 @@
...
+
+package com.google.firebase.crashlytics.internal.metadata;
+
+import com.google.firebase.crashlytics.internal.Logger;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/** Handles RolloutsState for metadata */
+public class RolloutAssignmentList {
```
samedson, line 23: I only see this being used in tests - is this being used yet?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignment.java
themiswang: Yup, this can live in interop and can be used by both RC and Crash, added AutoValue annotation.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignmentList.java
themiswang: Nope, just prepare for integration, this class will eventually add to UserMetaData.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStoreTest.java
themiswang: @mrober needs to finish some build config change in order to add Kotlin changes to Crashlytics. And from the conversation with Matt, would be easier to start the conversion from data model first.

themiswang: I purpose let's keep everything in java for now, and once we are ready to work in Kotlin I can re-write some of the class, what do you think?

samedson: Works for me 👍

samedson (approved):
🚢 🚢 🚢  first Rollouts PR! 🚢 🚢 🚢

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStoreTest.java
```diff
@@ -43,6 +47,8 @@ public class MetaDataStoreTest extends CrashlyticsTestCase {
 
   private static final String ESCAPED = "\ttest\nvalue";
 
+  private static final List<RolloutAssignment> ROLLOUTS_STATE = new ArrayList<>();
```
mrober, line 50: Does this need to be static? I am worried some unexpected behaviour could happen when running multiple tests, and state from one test affecting another. If this is changed to a member variable, then the setup will be more consistent.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignmentListTest.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @Test
+  public void testRollAssignmentInit_json() throws JSONException {
+    RolloutAssignment expected =
+        RolloutAssignment.create("rollout_1", "my_feature", "false", "control", 1);
+    RolloutAssignment rolloutAssignmentInitFromJson =
+        RolloutAssignment.create(ROLLOUT_ASSIGNMENT_JSON);
+
+    Truth.assertThat(rolloutAssignmentInitFromJson).isEqualTo(expected);
```
mrober, line 93: Nit: If you static import `Truth.assertThat` these will look better. It's just a style thing.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/MetaDataStoreTest.java
themiswang: Hmmm I can change them to non-static, but I do use it as a static variable. After ROLLOUTS_STATE is configured in setup() and I do not expect ant test case change the value. It is used as expected result comparison.

firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignmentListTest.java
```diff
@@ -0,0 +1,137 @@
...
+              }
+            })
+        .start();
+
+    new Thread(
+            new Runnable() {
+              @Override
+              public void run() {
+                List<RolloutAssignment> list = rolloutAssignmentList.getKeysMapList();
+                if (changed.isEmpty()) {
```
mrober, line 128: What is this testing? Is there a race between the two threads? Sometimes changed is empty and sometimes it's not?

themiswang: Yup want to check on the race condition. If you have one thread writing and one thread reading. I think the result is un-predictable. But want to make sure if writing before reading, reading will have the updated value, otherwise reads the previous value

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/metadata/RolloutAssignment.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * model used in user metadata context which make rollout assignment serialize, deserialize easier
+ */
+@Encodable
+@AutoValue
+public abstract class RolloutAssignment {
+
+  @Encodable.Field(name = "rolloutId")
+  public abstract String rolloutId();
```
mrober, line 32: Why do you need to set the field name manually, isn't it able to get the name automatically from the method name? Do you need to name them like `getRolloutId()`?

themiswang: Ah I see, let me change to getRolloutId(), I was following some example within the codebase.

mrober (approved):

PR #5194, opened 2023-07-25
Mila/multiple inequality support

milaGGL:


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1599,10 +1562,565 @@ public void testOrderByEquality() {
...
-    Query query2 = collection.where(Filter.inArray("a", asList(2, 3))).orderBy("a");
+    Query query2 = collection.where(inArray("a", asList(2, 3))).orderBy("a");
     checkOnlineAndOfflineResultsMatch(query2, "doc6", "doc3");
   }
+
+  /** Multiple Inequality */
+  @Test
+  public void testMultipleInequalityOnDifferentFields() {
+    // TODO(MIEQ): Enable this test against production when possible.
+    assumeTrue(
```
ehsannas, line 1576: I think we don't need these `assumeTrue`s. We can discuss offline.

ehsannas (approved):

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
```diff
@@ -1599,10 +1562,565 @@ public void testOrderByEquality() {
...
+                .whereGreaterThanOrEqualTo("key", "a")
+                .whereLessThanOrEqualTo("sort", 2)
+                .orderBy("v", Direction.DESCENDING)
+                .limitToLast(2)
+                .get());
+    assertEquals(asList("doc4", "doc1"), querySnapshotToIds(snapshot7));
+  }
+
+  @Test
+  public void testMultipleInequalityOnUnaryValues() {
```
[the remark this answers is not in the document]
milaGGL: my mistake. Should call it sentinel values.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/QueryTest.java
ehsannas: discussed offline

firebase-firestore/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 * [fixed] Implement equals method on Filter class. [#5210](//github.com/firebase/firebase-android-sdk/issues/5210)
+* [feature] Added support for multiple inequality in compound queries.
```
ehsannas, line 3: Added support for having multiple inequality fields in compound queries

firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java
```diff
@@ -615,6 +615,94 @@ public void testImplicitOrderBy() {
         baseQuery.orderBy(orderBy("foo", "desc")).orderBy(orderBy("bar", "asc")).getOrderBy());
   }
 
+  @Test
+  public void testImplicitOrderByInMultipleInequality() {
```
xwkuang5, line 619: Can we also add the following tests?

1. A mix of explicit order by and implicit order by fields from MIEQ
2. Exercise the normalization logic that uses the last explicit order by direction, i.e., `SELECT * FROM k WHERE a > 1 ORDER BY b DESC` normalizes to `SELECT * FROM k WHERE a > 1 ORDER BY b DESC, a DESC, __key__ DESC`

xwkuang5 (commented):
The order by normalization logic looks good to me.

firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexManager.java
```diff
@@ -73,7 +73,7 @@ enum IndexType {
    * <p>Values for this index are persisted asynchronously. The index will only be used for query
    * execution once values are persisted.
    */
-  void addFieldIndex(FieldIndex index);
+  void addFieldIndex(@Nullable FieldIndex index);
```
dconeybe, line 76: Drive-by comment: IMO the `FieldIndex` parameter of `addFieldIndex()` should NOT be nullable. Since a null field index is nonsensical it should throw an exception in this case. Otherwise, the code could be silently ignoring bugs where null is being specified incorrectly. Instead, the caller should assume the responsibility to _not_ call the function if its field index is null.

wu-hui: Agree, please address @cherylEnkidu

firebase-firestore/src/test/java/com/google/firebase/firestore/model/TargetIndexMatcherTest.java
```diff
@@ -802,20 +804,32 @@ public void testBuildTargetIndexWithInAndOrderBy() {
...
+  public void testBuildTargetIndexReturnsNullForMultipleInequality() {
+    Query query = query("collId").filter(filter("a", ">=", 1)).filter(filter("b", "<=", 10));
+    Target target = query.toTarget();
+    TargetIndexMatcher targetIndexMatcher = new TargetIndexMatcher(target);
+    assertTrue(targetIndexMatcher.hasMultipleInequality());
+    FieldIndex expectedIndex = targetIndexMatcher.buildTargetIndex();
+    assertNull(expectedIndex);
+  }
+
+  private void validateBuildTargetIndexCreateFullMatchIndexForSingleInequality(Query query) {
```
wu-hui, line 827: I think the rename is not an improvement for readability. Can we revert the name back?

wu-hui (approved):

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
```diff
@@ -86,7 +88,11 @@ public class TargetIndexMatcher {
   // The collection ID (or collection group) of the query target.
   private final String collectionId;
 
-  private @Nullable FieldFilter inequalityFilter;
+  // The inequality filters of the target (if it exists).
+  // Note: The sort on FieldFilters is not required. Using SortedSet here just to utilize
+  // the custom comparator.
+  private final SortedSet<FieldFilter> inequalityFilters;
```
dconeybe, line 94: nit: Use a HashSet if you can since ordering doesn't matter.

dconeybe (changes requested):
@cherylEnkidu Just one minor comment about using a SortedSet. Other than that, the indexing-related changes LGTM. I did _not_ review any other parts of this PR (I only reviewed the changes made by @cherylEnkidu)

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
milaGGL: I need to save the FieldFilters into a set solely based on the Field, so any inequality filters that has same Field but different operators needs to be deduplicated. SortedSet gives me the ability to do custom comparison based on the Field, but HashSet wouldn't, if i remember correctly.

dconeybe: HashSet and TreeSet are the same thing. But HashSet uses hashCode() and equals() to determine equality and perform lookups, where TreeSet uses the comparator and stores the elements ascending. So deduplication can be done using either; however, if TreeSet is more convenient because you have the comparator already available then that's fine. But you can definitely achieve the same thing with HashSet. And I'm glad you documented at least that the sort order is not relevant because that will make it much easier in the future if we ever want to swap it out.

firebase-firestore/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 # Unreleased
+* [feature] Added support for multiple inequality in compound queries.
```
ehsannas, line 2: please 
```
* [feature] Added support for multiple inequality in compound queries. [#5194](https://github.com/firebase/firebase-android-sdk/pull/5194)
```

firebase-firestore/src/main/java/com/google/firebase/firestore/model/TargetIndexMatcher.java
milaGGL: I am storing `FieldFilter`s only based on its `Field` property, so build in `hashCode()`,`equals()` doesn't meet my need here.
I do not wish to override these functions in the `FiledFilter` class to only consider the `Field` property as it is not a generic case for the class. 
So, giving a custom comparator in TreeSet works nicely here.

dconeybe (approved):

Issue #5195, opened 2023-07-25
NPE in sessions when packageInfo.versionName returns null


gerbiljames:
### [REQUIRED] Step 2: Describe your environment

  * Android Studio version: 2022.2.1 Patch 2
  * Firebase Component: Analytics, Dynamic Links, Remote Config, Cloud Messaging
  * Component version: BoM 32.2.0

#### Steps to reproduce:

When the Firebase SDK is initialised in an environment where `packageInfo.versionName` returns `null` an NPE is thrown. This is specifically happening for instrumented tests when the testing target is an Android library and not an application.

#### Relevant Code:

Captured stack trace:
```
java.lang.NullPointerException: packageInfo.versionName must not be null
	at com.google.firebase.sessions.SessionEvents.getApplicationInfo(SessionEvents.kt:77)
	at com.google.firebase.sessions.FirebaseSessions.<init>(FirebaseSessions.kt:41)
	at com.google.firebase.sessions.FirebaseSessionsRegistrar.getComponents$lambda-0(FirebaseSessionsRegistrar.kt:48)
	at com.google.firebase.sessions.FirebaseSessionsRegistrar.$r8$lambda$JITndpZCWeA0w9BDlkcI3l22oGY(Unknown Source:0)
	at com.google.firebase.sessions.FirebaseSessionsRegistrar$$ExternalSyntheticLambda0.create(Unknown Source:0)
	at com.google.firebase.tracing.ComponentMonitor.lambda$processRegistrar$0(ComponentMonitor.java:38)
	at com.google.firebase.tracing.ComponentMonitor$$ExternalSyntheticLambda0.create(Unknown Source:4)
	at com.google.firebase.components.ComponentRuntime.lambda$discoverComponents$0$com-google-firebase-components-ComponentRuntime(ComponentRuntime.java:140)
	at com.google.firebase.components.ComponentRuntime$$ExternalSyntheticLambda1.get(Unknown Source:4)
	at com.google.firebase.components.Lazy.get(Lazy.java:53)
	at com.google.firebase.components.ComponentContainer$-CC.$default$get(ComponentContainer.java:48)
	at com.google.firebase.components.ComponentRuntime.get(Unknown Source:0)
	at com.google.firebase.components.ComponentContainer$-CC.$default$get(ComponentContainer.java:24)
	at com.google.firebase.components.ComponentRuntime.get(Unknown Source:0)
[...]

PR #5198, opened 2023-07-26
Fix NPE when no version name is set

mrober:
Fix NPE when no version name is set. This will make it easier for apps to have Android tests while including the Sessions SDK. Also removed references to ApplicationInfoFlags to make it easier to minify apps with older api levels.

When the version name is unset, use the version code.

### DIFF
--- a/firebase-sessions/CHANGELOG.md
+++ b/firebase-sessions/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* [fixed] Fixed NPE when no version code is set.
+* [fixed] Fixed NPE when no version name is set.
 
 # 1.0.0
 
--- a/firebase-sessions/CHANGELOG.md
+++ b/firebase-sessions/CHANGELOG.md
@@ -1,3 +1,7 @@
 # Unreleased
 
+* [fixed] Fixed NPE when no version code is set.
+
+# 1.0.0
+
 * [feature] Initial Firebase sessions library.
--- a/firebase-sessions/CHANGELOG.md
+++ b/firebase-sessions/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-* [fixed] Fixed NPE when no version name is set.
+* [fixed] Fixed NPE when no version name is
+  set ([#5195](//github.com/firebase/firebase-android-sdk/issues/5195)).
 
 # 1.0.0
 
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
@@ -16,7 +16,6 @@
 
 package com.google.firebase.sessions
 
-import android.content.pm.PackageInfo
 import android.os.Build
 import com.google.firebase.FirebaseApp
 import com.google.firebase.encoders.DataEncoder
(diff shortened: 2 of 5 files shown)


mrober:
#5195

firebase-sessions/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 # Unreleased
 
+* [fixed] Fixed NPE when no version name is set.
```
visumickey, line 3: Can you add a reference to the Github issue here?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
```diff
@@ -74,7 +75,7 @@ internal object SessionEvents {
       androidAppInfo =
         AndroidApplicationInfo(
           packageName = packageName,
-          versionName = packageInfo.versionName,
+          versionName = packageInfo.versionName ?: buildVersion,
```
visumickey, line 78: Is build version guaranteed?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/LocalOverrideSettings.kt
```diff
@@ -18,32 +18,21 @@ package com.google.firebase.sessions.settings
...
 import android.content.Context
 import android.content.pm.PackageManager
-import android.os.Build
 import android.os.Bundle
 import kotlin.time.Duration
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
 
 internal class LocalOverrideSettings(context: Context) : SettingsProvider {
+  @Suppress("DEPRECATION") // TODO(mrober): Use ApplicationInfoFlags when target sdk set to 33
```
visumickey, line 27: With the version based checks removed, do we know if the new version APIs are backward compatible?

visumickey (approved):
Generally fine with the changes. But one thing to confirm before landing this PR. We need to make sure that the play store prevents submitting apps without `version_name`. As long as the impact is only with the tests scenario, this change looks good to me.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
mrober: The `buildVersion` is the `versionCode` as a String. When no versionCode is set, it returns 0. Play Store won't let apps upload two versions with the same versionCode. So it'll never be null, and it'll always be unique for apps published to the Play Store.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/LocalOverrideSettings.kt
mrober: They are, I tested it. And Firebase will be setting the target sdk to 33 soon anyway, then we can remove this check and use the new api only.

mrober:
> Generally fine with the changes. But one thing to confirm before landing this PR. We need to make sure that the play store prevents submitting apps without `version_name`. As long as the impact is only with the tests scenario, this change looks good to me.

Yes you have to set both versionCode and versionName for Play Store https://developer.android.com/studio/publish/versioning#versioningsettings and the versionCode cannot be reused.

PR #5201, opened 2023-07-26
Update checkout action to version 3.5.3

rlazo:
NO_RELEASE_CHANGE


emilypgoogle (approved):
Wonder if it'd be a good idea to have a repository of actions we used (env variable? local flie?) pinned to certain versions so we didn't need to update all of our actions like this. Worth considering for the future so we don't need too many PRs like this.

PR #5205, opened 2023-07-26
Populate DataCollectionStatus for Crashlytics and Perf in Sessions

mrober:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -139,7 +139,7 @@ internal constructor(
     }
 
     sessionCoordinator.attemptLoggingSessionEvent(
-      SessionEvents.startSession(firebaseApp, sessionDetails, sessionSettings)
+      SessionEvents.startSession(firebaseApp, sessionDetails, sessionSettings, subscribers)
```
visumickey, line 142: Is this pass by value of reference? Since we are passing this around for every event, we would be more recent. But still checking if there are any race conditions involved here.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
```diff
@@ -80,4 +86,13 @@ internal object SessionEvents {
         )
     )
   }
+
+  private fun toDataCollectionState(subscriber: SessionSubscriber?): DataCollectionState =
```
visumickey, line 90: Is the subscriber optional?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt
mrober: Yeah. If it's not present, then it means that sdk is not installed.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: The `subscribers` map is a "read only" map that gets copied every time we call `FirebaseSessionsDependencies.getRegisteredSubscribers()`. So it won't change between here and the `startSession` unless we do it on purpose, we would have to cast it to a `MutableMap`.

PR #5210, opened 2023-07-27
Implement Filter equals method.

tom-andersen:


firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
```diff
@@ -0,0 +1,188 @@
...
+public class FilterTest {
+
+  @Test
+  public void equalTo() {
+    Filter filter = Filter.equalTo("x", "v");
+    assertEquals(filter, filter);
+    assertEquals(filter, Filter.equalTo(FieldPath.of("x"), "v"));
+    assertNotEquals(filter, Filter.equalTo("x", "z"));
+    assertNotEquals(filter, Filter.equalTo("y", "v"));
+    assertNotEquals(filter,Filter.notEqualTo("y", "v"));
```
ehsannas, line 38: add assert where the fields match and the values match, but the operator is different.
```
assertNotEquals(filter,Filter.notEqualTo("x", "v"));
```

can be done for the other tests too

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
```diff
@@ -0,0 +1,188 @@
...
+    assertNotEquals(filter, Filter.notInArray("x", ImmutableList.of("v2", "v3")));
+    assertNotEquals(filter, Filter.notInArray("y", ImmutableList.of("v1", "v2")));
+    assertNotEquals(filter, Filter.inArray("y", ImmutableList.of("v1", "v2")));
+  }
+
+  @Test
+  public void or() {
+    Filter filter = Filter.or(
+        Filter.equalTo("x", ImmutableList.of("v1", "v2")),
+        Filter.equalTo("y", ImmutableList.of("v3", "v4"))
```
ehsannas, line 135: technically, it's not correct to pass a List as a value of the `equalTo` filter operator. I believe it'd error out when the filter gets parsed. It doesn't really make a different in the context of what is being unit tested here though. But to be precise, let's replace `equalTo` with `inArray`.

Applies elsewhere below.

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
```diff
@@ -0,0 +1,188 @@
...
+            Filter.equalTo("y", ImmutableList.of("v3", "v4"))
+        )
+    );
+    assertNotEquals(
+        filter,
+        Filter.and(
+            Filter.equalTo("y", ImmutableList.of("v3", "v4")),
+            Filter.equalTo("x", ImmutableList.of("v1", "v2"))
+        )
+    );
```
ehsannas, line 157: add a case where the operator remains unchanged, but the _filter_ order changes.

i.e.
```
assertNotEquals(
        filter,
        Filter.or(
            Filter.equalTo("y", ImmutableList.of("v3", "v4")),
            Filter.equalTo("x", ImmutableList.of("v1", "v2"))
        )
    );
```

add a case where the operator remains unchanged, and the filter order remains unchanged, but the list value order changes.

i.e.
```
assertNotEquals(
        filter,
        Filter.or(
          Filter.equalTo("x", ImmutableList.of("v2", "v1")),
          Filter.equalTo("y", ImmutableList.of("v4", "v3"))
        )
    );
```

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
```diff
@@ -0,0 +1,188 @@
...
+            Filter.equalTo("y", ImmutableList.of("v3", "v4"))
+        )
+    );
+    assertNotEquals(
+        filter,
+        Filter.or(
+            Filter.equalTo("y", ImmutableList.of("v3", "v4")),
+            Filter.equalTo("x", ImmutableList.of("v1", "v2"))
+        )
+    );
```
ehsannas, line 186: add the same two cases as `or()` test above.

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
tom-andersen: Right. That was the intent. I'll fix.

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
tom-andersen: Will do.

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
tom-andersen: Right. That was the intent of last assert, so I will fix it to match what you say.

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
tom-andersen: Already there. See last assert.

tom-andersen: Ignore above comment. Will do.

firebase-firestore/src/test/java/com/google/firebase/firestore/FilterTest.java
tom-andersen: Sorry, I don't see point of second test you want to add. That is already covered in the `inArray` tests, after I switched operator to be `inArray`.

tom-andersen: No wait... It should have been covered, but I actually changed the values, not just the order. I will add test up there. I think the And/Or operators need to be trusted to call equals methods on composed Filters.

ehsannas (approved):

PR #5213, opened 2023-07-31
Increase test coverage for persistent cache indexing

cherylEnkidu:


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
```diff
@@ -108,6 +108,9 @@ public void testBadIndexDoesNotCrashClient() {
                     + "}"));
   }
 
+  /**
+  * After Auto Index Creation is enabled, through public API there is no way to state of indexes sitting inside SDK. So this test only checks the API of auto index creation.
```
dconeybe, line 112: Please fix these comments to wrap at 100 characters.

dconeybe (changes requested):
Just one tiny nit about line length.

Note that in the web port of client-side index auto creation (https://github.com/firebase/firebase-js-sdk/pull/7542) I added "testing hooks" that provide a way for integration tests to verify that the indexes are (or are not) created, as expected. This is something we should consider porting to android and ios IMO. The "testing hooks" infrastructure exists in android and ios too (it was created for the bloom filter work) so we'd just need to add to it.

cherylEnkidu:
As discussed in the meeting, testing hook will be added in a separated PR. Tracking number: b/296100693

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/FirestoreTest.java
```diff
@@ -1419,4 +1419,77 @@ public void testCannotGetDocumentWithMemoryEagerGcEnabled() {
...
+    // Use persistent disk cache (explicit)
+    FirebaseFirestore db1 = testFirestore();
+    FirebaseFirestoreSettings settings1 =
+        new FirebaseFirestoreSettings.Builder(db1.getFirestoreSettings())
+            .setLocalCacheSettings(PersistentCacheSettings.newBuilder().build())
+            .build();
+    db1.setFirestoreSettings(settings1);
+    PersistentCacheIndexManager indexManager1 = db1.getPersistentCacheIndexManager();
+    PersistentCacheIndexManager indexManager2 = db1.getPersistentCacheIndexManager();
+    assertEquals(indexManager1, indexManager2);
```
dconeybe, line 1468: nit: Use `assertSame()` instead of `assertEquals()` in order to verify that they are exactly the same object, compared using the `==` operator.

This comment applies here and elsewhere in this test, including using `assertNotSame()` instead of `assertNotEquals()` when testing for object identity.

https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html#assertSame(java.lang.Object,%20java.lang.Object)

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/IndexingTest.java
```diff
@@ -108,9 +108,13 @@ public void testBadIndexDoesNotCrashClient() {
                     + "}"));
   }
 
+  /**
```
dconeybe, line 111: Add a `TODO(b/296100693) Add testing hooks to verify indexes are created as expected` comment here, and in the comment of the following function. This will help future you or somebody else correlate the buganizer ticket with the code in question, and vice versa.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteIndexManagerTest.java
```diff
@@ -1189,6 +1190,49 @@ public void testIndexTypeForOrQueries() throws Exception {
     validateIndexType(query11, IndexManager.IndexType.PARTIAL);
   }
 
+  @Test
+  public void TestCreateTargetIndexesCreatesFullIndexesForEachSubTargets() {
```
dconeybe, line 1194: nit: `TestCreateTargetIndexesCreatesFullIndexesForEachSubTargets` -> `TestCreateTargetIndexesCreatesFullIndexesForEachSubTarget` (that is, remove the trailing `s` on `Targets`).

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -398,19 +399,50 @@ public void testCanAutoCreateIndexes() {
...
+
+    executeQuery(query);
+    assertRemoteDocumentsRead(/* byKey= */ 2, /* byCollection= */ 1);
+    assertQueryReturned("coll/f", "coll/e", "coll/a");
+  }
+
   @Test
   public void testDoesNotAutoCreateIndexesForSmallCollections() {
-    Query query = query("coll").filter(filter("count", ">=", 3));
+    Query query = query("coll").filter(filter("foo", "==", 9)).filter(filter("count", ">=", 3));
```
dconeybe, line 435: Could you explain the reason for this change? It doesn't seem to be any different from before given the name of the test, "testDoesNotAutoCreateIndexesForSmallCollections". That is, the collection is "small" in both cases. I ask because this seems like you're trying to add some test coverage which may warrant its own test method rather than extending this test method to cover two cases.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -460,18 +492,22 @@ public void testDoesNotAutoCreateIndexesWhenIndexLookUpIsExpensive() {
 
   @Test
   public void testIndexAutoCreationWorksWhenBackfillerRunsHalfway() {
-    Query query = query("coll").filter(filter("matches", "==", "foo"));
+    Query query =
```
dconeybe, line 495: Can you explain the changes to `testIndexAutoCreationWorksWhenBackfillerRunsHalfway` also? Again, it appears that the code before and after _both_ test the scenario explained by the test method's name, but the "before" code is simpler.

dconeybe (changes requested):
A few minor nits, but mostly LGTM.

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
cherylEnkidu: `cover two cases` -> yes, this is what I try to do here.

The reason for that is I try to diversify the queries used in indexing tests. For other tests in this file, I tests queries on different types of fields. Then I released these tests are focus on one field, so I decided to change some of that to test on multiple fields.

dconeybe: Ok that makes sense. For unit tests it's usually more readable to separate tests for distinct cases into distinct methods. For example, this test, by its name, is testing that indexes are _not_ created when the collection is too small. So the query should be as simple as possible so as not to distract from what is actually being testing. By making the query more complex it detracts from the readability of the test case because there is a lot of "noise". For example, looking at the documents that are added it takes some time to determine which of those documents would actually match the query. If, instead, the query just filtered on "foo=42" then it would be immediately obvious which documents match.

So consider breaking up this (and some other) tests into two different test methods: one for testing that indexes are _not_ created when the collection is too small and one for testing two filters (an equality and an inequality). If you'd like to do this but not right now then you can create a taskflow task as a reminder.

http://go/tott/649

dconeybe (approved):

PR #5221, opened 2023-08-02
Auto transform

VinayGuthal:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/PackageTransform.kt
```diff
@@ -0,0 +1,321 @@
...
+  fun deprecateKTX(src: String, pkgName: String) {
+    File(src).walk().forEach {
+      if (it.absolutePath.endsWith(".kt")) {
+        val lines = File(it.absolutePath).readLines()
+        val output = mutableListOf<String>()
+        for (i in 0 until lines.size) {
+          output.add(lines[i])
+          if (lines[i].contains("*/")) {
+            var symbol = ""
+            var ctr = i + 1
```
rlazo, line 133: what `ctr` stands for

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PackageTransform.kt
```diff
@@ -0,0 +1,321 @@
...
+              !x.contains("LibraryVersionComponent.create(LIBRARY_NAME, BuildConfig.VERSION_NAME)")
+          }
+        File(it.absolutePath).delete()
+        val newFile = File(it.absolutePath)
+        newFile.writeText(lines.joinToString("\n"))
+      }
+    }
+  }
+
+  fun readDependencies(path: String): List<String> {
```
rlazo, line 186: you can read the whole file in a string, then find `dependencies {` and then start processing it.

firebase-common/ktx/src/main/kotlin/com/google/firebase/ktx/Logging.kt
```diff
@@ -0,0 +1,29 @@
+// Copyright 2023 Google LLC
```
rlazo, line 1: What about `legacyRegistrar`?

rlazo:
Just to keep record: we still want to keep most of the tests in the KTX artifact to make sure usage of the api works for people not migrating.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PackageTransform.kt
rlazo: probably we can come up with a better naming... not sure what though :-)

rlazo:
> Just to keep record: we still want to keep most of the tests in the KTX artifact to make sure usage of the api works for people not migrating.

This is WAI, please ignore.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PackageTransform.kt
rlazo: Feel free to ignore since it's not going to be merged to master.

PR #5223, opened 2023-08-03
refactor the LG check into its own task, minor rewrite for kotlin con…

davidmotson:
…vention

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
@@ -81,6 +81,8 @@ abstract class PublishingPlugin : Plugin<Project> {
         registerCheckHeadDependenciesTask(project, releasingFirebaseLibraries)
       val validateProjectsToPublish =
         registerValidateProjectsToPublishTask(project, releasingFirebaseLibraries)
+      val validateLibraryGroupsToPublish =
+        registerValidateLibraryGroupsToPublishTask(project, releasingFirebaseLibraries)
       val publishReleasingLibrariesToBuildDir =
         registerPublishReleasingLibrariesToBuildDirTask(project, releasingProjects)
       val generateKotlindocsForRelease =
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -287,13 +290,27 @@ abstract class PublishingPlugin : Plugin<Project> {
...
+  ) =
+    project.tasks.register(VALIDATE_LIBRARY_GROUPS_TO_PUBLISH_TASK) {
+      doLast {
+        val libraryGroupProjects = releasinglibraries.flatMap { it.librariesToRelease }
+        val missingProjects = libraryGroupProjects - releasinglibraries
+        if (missingProjects.isNotEmpty()) {
           throw GradleException(
-            "Some libraries in library groups are not in the release: " +
-              Sets.difference(libraryGroupProjects, releasingProjects).map { it.displayName }
+            "Some libraries in library groups are not in the release: ${missingProjects.map { it.mavenName }}"
```
daymxn, line 313: Without the `joinToString("\n")` this will likely print something like:
```
Some libraries in library groups are not in the release: ["com.google.firebase:firebase-annotations:16.2.0", ""com.google.firebase:firebase-components:17.1.0"]
```

Instead of:
```
Some libraries in library groups are not in the release: 
com.google.firebase:firebase-annotations:16.2.0
com.google.firebase:firebase-components:17.1.0
```
But it's not that big of a deal tbf.

rlazo (approved):

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -475,6 +492,7 @@ abstract class PublishingPlugin : Plugin<Project> {
 
     const val GENERATE_BOM_TASK = "generateBom"
     const val VALIDATE_PROJECTS_TO_PUBLISH_TASK = "validateProjectsToPublish"
+    const val VALIDATE_LIBRARY_GROUPS_TO_PUBLISH_TASK = "validateLGsToPublish"
```
daymxn, line 495:
```suggestion
    const val VALIDATE_LIBRARY_GROUPS_TO_PUBLISH_TASK = "validateLibraryGroupsToPublish"
```

We already have long names like:
```kotlin
    const val PUBLISH_RELEASING_LIBS_TO_BUILD_TASK = "publishReleasingLibrariesToBuildDir"
    const val PUBLISH_RELEASING_LIBS_TO_LOCAL_TASK = "publishReleasingLibrariesToMavenLocal"
```

daymxn (approved):

Issue #5176, opened 2023-07-18
`firebase-sessions:1.0.0` incompatible with `datastore-preferences:1.1.0-alpha04`


martinbonnin:
_follow up from https://github.com/joreilly/Confetti/pull/809#pullrequestreview-1533842937_

Having both `firebase-sessions:1.0.0` and `datastore-preferences:1.1.0-alpha04` in the dependencies:

```kotlin
dependencies {
  implementation("androidx.datastore:datastore-preferences:1.1.0-alpha04")
  implementation("com.google.firebase:firebase-sessions:1.0.0")
}
```

fails with:

```
$ ./gradlew assembleDebug
> Task :checkDebugAarMetadata FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':checkDebugAarMetadata'.
> Could not resolve all files for configuration ':debugRuntimeClasspath'.
   > Failed to transform datastore-preferences-1.1.0-alpha04.aar (androidx.datastore:datastore-preferences:1.1.0-alpha04) to match attributes {artifactType=android-aar-metadata, org.gradle.status=release}.
      > Could not find datastore-preferences-1.1.0-alpha04.aar (androidx.datastore:datastore-preferences:1.1.0-alpha04).
        Searched in the following locations:
            https://dl.google.com/dl/android/maven2/androidx/datastore/datastore-preferences/1.1.0-alpha04/datastore-preferences-1.1.0-alpha04.aar
```

My superficial understanding of this is that `datastore-preferences` was transformed to a KMP artifact and therefore `datastore-preference` doesn't contain a `.aar` anymore and thus the build is failing.

Looks like this is happening somewhere [here](https://github.com/firebase/firebase-android-sdk/blob/7b940f822555f554318cd7df64d6a67537bdc3e2/buildSrc/src/main/java/com/google/firebase/gradle/plugins/BaseFirebaseLibraryPlugin.kt#L219) where the pom is rewritten to detect which dependencies are android and which are not. Looks like `datastore-preferences` needs to be set as a non-android dependency now?

In all cases, a simple workaround for Android project is to depend on `datastore-preferences-android` instead:

```kotlin
dependencies {
  implementation("androidx.datastore:datastore-preferences-android:1.1.0-alpha04")
  implementation("com.google.firebase:firebase-sessions:1.0.0")
}
```

But this won't work for KMP projects.

Sample repo [here](https://github.com/martinbonnin/test-datastore)

[...]

PR #5225, opened 2023-08-03
Enable gradle metadata publishing

yigit:
This PR enables GradleMetadataPublishing, which is necessary to properly resolve Kotlin Multiplatform artifacts.

Fixes: #5176


daymxn (approved):
LGTM! Sorry for the delay, it's been a busy schedule lately. I did some local testing, and everything seems well. Seems the test failures are flakes:)

PR #5230, opened 2023-08-07
Add Firebase Remote Config Test App for Feature Rollouts

samedson:
This is copied from the Sessions test app. The single test file isn't filled in, so we can do that in a follow up


firebase-config/test-app/test-app.gradle.kts
```diff
@@ -0,0 +1,67 @@
...
+  }
+  compileOptions {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+  }
+  kotlinOptions { jvmTarget = "1.8" }
+}
+
+dependencies {
+  implementation(project(":firebase-crashlytics"))
```
themiswang, line 49: Wonder do we want to add ndk dependencies as well? Or it will be followed up with a following pr?

firebase-config/test-app/src/androidTest/kotlin/com/google/firebase/testing/sessions/FirebaseConfigTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+  @After
+  fun cleanUp() {
+    FirebaseApp.clearInstancesForTest()
+  }
+
+  @Test
+  fun initializeConfig_generatesSessionEvent() {
+
+    // // Force the Firebase Config SDK to initialize.
```
themiswang, line 45: Any reason we comment out all tests?

danasilver: I think just cause we don't have anything to test yet

danasilver:
Wonder if we should call this "rollouts-test-app" to distinguish it from the RC-specific test app in the adjacent directory (bandwagoner)? But OK merging this and refactoring if we want later

firebase-config/test-app/src/androidTest/kotlin/com/google/firebase/testing/sessions/FirebaseConfigTest.kt
```diff
@@ -0,0 +1,78 @@
+/*
```
qdpham13, line 1: I know on the RC side Kotlin is basically just a wrapper of Java but I was wondering if we would need one app in Java and another in Kotlin so that we test both code paths

qdpham13 (commented):
Is the purpose of this test app to be run in pre-checks when we're pushing changes?

firebase-config/test-app/src/androidTest/kotlin/com/google/firebase/testing/sessions/FirebaseConfigTest.kt
qdpham13: Sounds good

firebase-firestore/src/test/java/com/google/firebase/firestore/local/SQLiteLocalStoreTest.java
```diff
@@ -444,9 +444,7 @@ public void testDoesNotAutoCreateIndexesWhenIndexLookUpIsExpensive() {
     applyRemoteEvent(
         addedRemoteEvent(doc("coll/e", 10, map("array", Arrays.asList(2, 0, 8))), targetId));
 
-    // First time query runs without indexes.
-    // Based on current heuristic, collection document counts (5) > 2 * resultSize (2).
-    // Full matched index should be created.
+    // SDK will not create indexes since relative read cost is too large.
```
samedson, line 447: Are these changes accidentally included?

qdpham13:
This Test App has been updated to mention Remote Config instead of Sessions and can be run using the README.md

themiswang (approved):

danasilver (approved):

PR #5237, opened 2023-08-14
Expose the transport factory as a Qualified dependency

rlazo:
The datatransport factory was exposed as a single, unqualified component. 

Now it's exposed as both the unqualified component, to maintain backwards compatibility, and two qualified components with matching endpoints.

### DIFF
--- a/firebase-datatransport/gradle.properties
+++ b/firebase-datatransport/gradle.properties
@@ -1,3 +1,3 @@
-version=18.1.10
+version=18.1.9
 latestReleasedVersion=18.1.8
 android.enableUnitTestBinaryResources=true


firebase-datatransport/gradle.properties
```diff
@@ -1,4 +1,3 @@
-version=18.1.9
+version=18.1.10
```
emilypgoogle, line 1: Why is this incrementing past an unreleased patch version?

emilypgoogle (commented):
Code looks fine, curious about the versioning situation

firebase-datatransport/gradle.properties
rlazo: Great catch, thanks! Fixed

emilypgoogle (approved):

PR #5241, opened 2023-08-15
Fix semver check

VinayGuthal:
During the m134 release we were releasing firebase-sessions for the first time and due to which the semver test was broken. Hence the if conditions were added. These if conditions were not being evaluated correctly since it dependend on a malleable state of things. So now it's moved to checking in Gmaven itself if the project exists instead of depending on a malleable state which keeps being updated by the gradle tasks.


rlazo:
Please add a more descriptive title/description to know what was the error.

rlazo (approved):

PR #5254, opened 2023-08-17
Finish PostReleasePlugin

daymxn:
Per [b/292140435](https://b.corp.google.com/issues/292140435),

This implements a task to handle conversion of project level dependencies to pinned dependencies post release. Relevant documentation explaining everything is added as well.

The following is also fixed by this PR:

- [b/296419486](https://b.corp.google.com/issues/296419486) - Connect `updatePinnedDependencies` task with `PostReleasePlugin`
- [b/296419551](https://b.corp.google.com/issues/296419551) - Implement a `postReleaseCleanup` task
- [b/296419351](https://b.corp.google.com/issues/296419351) - Add tests for the `updatePinnedDependencies` task
- [b/296420102](https://b.corp.google.com/issues/296420102) - Implement a common test interface for dynamically created projects
- [b/296420881](https://b.corp.google.com/issues/296420881) - Migrate `PublishingPluginTests` to the new testing interface
- [b/296422065](https://b.corp.google.com/issues/296422065) - Migrate `PublishingPluginTests` to kotest assertions

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
@@ -60,23 +60,54 @@ abstract class UpdatePinnedDependenciesTask : DefaultTask() {
 
   @TaskAction
   fun build() {
-    val needsChanging = computeLibrariesThatNeedChanging()
+    val needsChanging = findProjectLevelDependenciesToChange()
+
+    if (needsChanging.isEmpty()) throw StopExecutionException("No libraries to change.")
 
     val lines = buildFile.get().readLines()
     val newLines = updateDependencyLines(lines, needsChanging)
 
-    if (newLines == lines) throw missingProjectLevelDependenciesError(needsChanging)
+    validateNewLines(needsChanging, lines, newLines)
 
     outputFile.get().writeText(newLines.joinToString("\n") + "\n")
   }
 
-  private fun computeLibrariesThatNeedChanging(): List<FirebaseLibraryExtension> {
+  private fun validateNewLines(
+    needsChanging: List<FirebaseLibraryExtension>,
+    lines: List<String>,
+    newLines: List<String>
+  ) {
+    if (newLines == lines) throw missingProjectLevelDependenciesError(needsChanging)
+
+    val diff = lines.diff(newLines)
+    val changedLines = diff.mapNotNull { it.first ?: it.second }
+
+    val (librariesCorrectlyChanged, linesChangedIncorrectly) =
+      needsChanging.partition { lib -> changedLines.any { it.contains(lib.path) } }
+
+    val librariesNotChanged = needsChanging - librariesCorrectlyChanged
+
+    if (linesChangedIncorrectly.isNotEmpty())
+      throw RuntimeException(
+        "The following lines were caught by our REGEX, but should not have been:\n ${linesChangedIncorrectly.joinToString("\n")}"
+      )
+
+    if (librariesNotChanged.isNotEmpty())
+      throw RuntimeException(
+        "The following libraries were not found, but should have been:\n ${librariesNotChanged.joinToString("\n") { it.mavenName }}"
+      )
+
+    if (librariesCorrectlyChanged.size > needsChanging.size)
+      throw RuntimeException(
+        "Too many libraries were caught by our change, possible REGEX false positive:\n ${changedLines.joinToString("\n")}"
+      )
+  }
+
+  private fun findProjectLevelDependenciesToChange(): List<FirebaseLibraryExtension> {
     val firebaseLibrary = project.firebaseLibrary
     val sisterProjects = firebaseLibrary.getLibrariesToRelease()
     val dependencies = projectLevelDependenciesForLibrary(firebaseLibrary)
     val needsChanging = dependencies - sisterProjects
-
-    if (needsChanging.isEmpty()) throw StopExecutionException("No libraries to change.")
 
     return needsChanging
   }
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
@@ -77,7 +77,11 @@ abstract class UpdatePinnedDependenciesTask : DefaultTask() {
     lines: List<String>,
     newLines: List<String>
   ) {
-    if (newLines == lines) throw missingProjectLevelDependenciesError(needsChanging)
+    if (newLines == lines)
+      throw RuntimeException(
+        "Expected the following project level dependencies, but found none: " +
+          "${needsChanging.joinToString("\n") { it.mavenName }}"
+      )
 
     val diff = lines.diff(newLines)
     val changedLines = diff.mapNotNull { it.first ?: it.second }
@@ -105,15 +109,12 @@ abstract class UpdatePinnedDependenciesTask : DefaultTask() {
 
   private fun findProjectLevelDependenciesToChange(): List<FirebaseLibraryExtension> {
     val firebaseLibrary = project.firebaseLibrary
-    val sisterProjects = firebaseLibrary.getLibrariesToRelease()
-    val dependencies = projectLevelDependenciesForLibrary(firebaseLibrary)
-    val needsChanging = dependencies - sisterProjects
 
-    return needsChanging
+    return firebaseLibrary.projectLevelDependencies - firebaseLibrary.librariesToRelease
   }
 
-  private fun projectLevelDependenciesForLibrary(library: FirebaseLibraryExtension) =
-    library.resolveProjectLevelDependencies().filterNot { it.path in DEPENDENCIES_TO_IGNORE }
+  private val FirebaseLibraryExtension.projectLevelDependencies: List<FirebaseLibraryExtension>
+    get() = resolveProjectLevelDependencies().filterNot { it.path in DEPENDENCIES_TO_IGNORE }
 
   private fun updateDependencyLines(
     lines: List<String>,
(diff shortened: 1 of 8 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+    val newLines = updateDependencyLines(lines, needsChanging)
+
+    if (newLines == lines) throw missingProjectLevelDependenciesError(needsChanging)
+
+    outputFile.get().writeText(newLines.joinToString("\n") + "\n")
+  }
+
+  private fun computeLibrariesThatNeedChanging(): List<FirebaseLibraryExtension> {
+    val firebaseLibrary = project.firebaseLibrary
+    val sisterProjects = firebaseLibrary.getLibrariesToRelease()
```
rlazo, line 75: what do you mean by `sisterProjects`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+    val dependencies = projectLevelDependenciesForLibrary(firebaseLibrary)
```
rlazo, line 76: `dependencies` are more generic than what a `projectLevelDependenciesForLibrary` method would return

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
rlazo: Also, you can use `firebaseLibrary.librariesToRelease`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
rlazo: Also, this looks like a good candidate for a property

```
  private val FirebaseLibraryExtension.projectLevelDependencies
    get() = resolveProjectLevelDependencies().filterNot { it.path in DEPENDENCIES_TO_IGNORE }
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+    val needsChanging = dependencies - sisterProjects
+
+    if (needsChanging.isEmpty()) throw StopExecutionException("No libraries to change.")
```
rlazo, line 79: Not sure the method should be the one throwing if empty...

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+    lines.replaceMatches(DEPENDENCY_REGEX) {
+      val projectName = it.firstCapturedValue
+      val projectToChange = libraries.find { it.path == projectName }
+      val latestVersion = projectToChange?.latestVersion
+
+      latestVersion?.let { "\"${projectToChange.mavenName}:$latestVersion\"" }
+    }
+
+  // This should never occur, but if it does- it's probably a regex error. Better safe than sorry?
+  private fun missingProjectLevelDependenciesError(
```
rlazo, line 100: I'd rather not wrap the exception throwing inside a function since it obscures where errors are found.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+    expectedLibraries: List<FirebaseLibraryExtension>
+  ) =
+    RuntimeException(
+      "Expected the following project level dependencies, but found none: " +
+        "${expectedLibraries.joinToString("\n") { it.mavenName }}"
+    )
+
+  companion object {
+    /** TODO() */
```
rlazo, line 109: todo

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+abstract class UpdatePinnedDependenciesTask : DefaultTask() {
+  @get:[InputFile]
+  abstract val buildFile: Property<File>
+
+  @get:[OutputFile]
+  abstract val outputFile: Property<File>
+
+  @TaskAction
+  fun build() {
+    val needsChanging = computeLibrariesThatNeedChanging()
```
rlazo, line 63: what you are computing is the "dependencies" that need to be updated, right?

If so, use a naming that closer reflects that, like "getOutdatedProjectLevelDependencies"

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    val lines = buildFile.get().readLines()
+    val newLines = updateDependencyLines(lines, needsChanging)
+
+    if (newLines == lines) throw missingProjectLevelDependenciesError(needsChanging)
```
rlazo, line 68: As implemented, you are relying on the equality of the text to know if this succeeded or not. Rather, we should know which, if any, deps couldn't be updated and notify the user. Then it will be easier to understand what failed.

daymxn: Good idea, I've added a verify method :)

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
daymxn: `sisterProjects` == `siblingProjects` == libraries in the same library group

Good catch on the Java method call, I'll change that rq

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
daymxn: SGTM

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
daymxn: SGTM. Mainly did it to abstract the long line and help with readability.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -60,29 +60,61 @@ abstract class UpdatePinnedDependenciesTask : DefaultTask() {
 
   @TaskAction
   fun build() {
```
rlazo, line 62: As it stands right now, without context it takes a bit to read through the function to fully understand what's doing. 

- `build` as the function name doesn't match what the action does
- `needsChanging` is not a noun, so when reading through the code is not clear what's stored in the val
- `lines`, `newlines` , and `validateNewLines`. While yes, at the end of the day what will end up happening is lines in a file changing, we can keep this function focus on the "what" should happen(update dependencies), rather that on "how" it should happen (change lines in a file). 

Question:

- Do we need to throw if there's nothing to change? Why is an early return not enough?

rlazo (commented):
When ready to review again, don't forget to re-request the review in the PR. I'm way too prone to lose track of the PRs to review and I don't want to block you. Thanks!

buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
rlazo: Could we use `onlyIf` for this scenario?

daymxn: Yes and no.

Yes, we could technically do it- but we shouldn't be able to. The only reason we **are** able to is because of the way our publishing plugin is structured. Additionally, it would require us to provide the project level dependencies as inputs- and would require the task that registers this (the post release task) to know about the existence of `DEPENDENCIES_TO_IGNORE` .

rlazo: curious on why we shouldn't be able to. It's a task that implements the check, and the check is ran at execution time.

daymxn: On second thought, we could actually make it work if we wrapped the `librariesToUpdate` or w/e in a `Provider`. The initial problem was that while `onlyIf` evaluates at execution time- the data would need to be used outside of the `onlyIf`- as a task input. And that would need to be configured at configuration time. But that's a non-issue if we wrap it in a `Provider`.

My only concern here now would be having `PostReleasePlugin` be aware of `DEPENDENCIES_TO_IGNORE`, although tbf It's not *that* big of a deal- since it *is* an implementation detail.

I'm content either way, wdywd?

rlazo: Thanks for the explanation. It's OK to go with our current approach then.

rlazo (approved):

PR #5233, opened 2023-08-09
Update Crashlytics Report Proto with new Rollouts fields

themiswang:
- Update CrashlyticsReport with new proto, followed the backend change: [cl/554467862](http://cl/554467862)
- Adding RolloutAssignmentList to UserMetadata
- Adding limits to parameter value for rollout assignment

#no-changelog

PR #5255, opened 2023-08-17
Initial commit of firebase-config-interop SDK.

danasilver:


firebase-config-interop/src/main/java/com/google/firebase/remoteconfig/interop/rollouts/RolloutAssignment.java
```diff
@@ -0,0 +1,89 @@
...
+  public abstract String getParameterValue();
+
+  /** The Remote Config template version containing this rollout. */
+  public abstract long getTemplateVersion();
+
+  public static final DataEncoder ROLLOUT_ASSIGNMENT_JSON_ENCODER =
+      new JsonDataEncoderBuilder().configureWith(AutoRolloutAssignmentEncoder.CONFIG).build();
+
+  /** Create a {@code RolloutAssignment} from its JSON-encoded representation. */
+  public static RolloutAssignment create(String jsonString) throws JSONException {
```
themiswang, line 60: Wonder for `RolloutAssignment`, is it possible to have a parameter value length validation function to restrict the max length pass to crashlytics? https://github.com/firebase/firebase-android-sdk/pull/5233/files#diff-442de867f9aec30da44b1c83269c068ef34ac6b6c8332359bba5f0740d28bc8fR72

danasilver: I think we should keep that validation/restriction in the Crashlytics SDK since other consumer SDKs might not have the same requirement

themiswang: Seems the implementation will become a bit redundant then, on Crashlytics side whenever we get the list of RolloutAssignment, we still need to iterate the list of object and re-create a new object for the ones which exceed the limit. If that's the case I don't see the reason we want to use the same object. Or is it possible to provide an api in `RolloutsState` to give other SDK the ability to return the rolloutsState with validation? for example:

```
getRolloutsStateWithValidatedParameterValueLength(int maxLenght): {}
```

firebase-config-interop/src/main/java/com/google/firebase/remoteconfig/interop/rollouts/RolloutsState.java
```diff
@@ -0,0 +1,33 @@
...
+import androidx.annotation.NonNull;
+import com.google.auto.value.AutoValue;
+import java.util.Set;
+
+/** Model representing the state of all rollouts assigned to an app instance at a point in time. */
+@AutoValue
+public abstract class RolloutsState {
+  /** Rollouts assigned at this point in time. May include zero or more assignments. */
+  @NonNull
+  public abstract Set<RolloutAssignment> getRolloutAssignments();
```
themiswang, line 28: Saw a warning in unit test, suggest using `ImmutableSet` https://github.com/firebase/firebase-android-sdk/actions/runs/5894721664/job/15989122819?pr=5255

danasilver: Good call, I see that's AutoValue's recommendation and likely what we'll want to pass from RC anyway. Updated.

danasilver: I noticed that none of the other Firebase SDKs depend on guava or use any of the `Immutable*` classes outside of tests. Just based on some quick digging around, it looks like there's a lot of concern about the size of the dependency. I'm going to keep `Set` here

firebase-config-interop/src/main/java/com/google/firebase/remoteconfig/interop/FirebaseRemoteConfigInterop.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.google.firebase.remoteconfig.interop;
+
+import androidx.annotation.NonNull;
+import com.google.firebase.remoteconfig.interop.rollouts.RolloutsStateSubscriber;
+
+public interface FirebaseRemoteConfigInterop {
+  void registerRolloutsStateSubscriber(
+      @NonNull String namespace, @NonNull RolloutsStateSubscriber subscriber);
```
samedson, line 24: Can we add a comment for what the namespace is? Also maybe this could be a good place to explain how other SDKs should use this

samedson (approved):
Just had one comment. Nice to see the interop in!

qdpham13 (approved):
LGTM

themiswang:
You probably need to update your branch as well, let me know if below the list of git command works for you!
```
git fetch
git checkout featureRollouts // checkout feature branch
git pull --rebase // update local feature branch
git checkout ds-rc-interop // checkout your pr branch
git rebase featureRollouts // rebase feature branch to your pr branch
git push --force // still need a force push...
```

themiswang (approved):

firebase-config-interop/src/main/java/com/google/firebase/remoteconfig/interop/FirebaseRemoteConfigInterop.java
danasilver: Yeah, good call.

firebase-config-interop/src/main/java/com/google/firebase/remoteconfig/interop/rollouts/RolloutAssignment.java
danasilver: Discussed offline: we'll merge the PR as-is and add the validation logic here later if it makes sense.

danasilver:
Nice, thanks for those commands @themiswang! 👌

PR #5257, opened 2023-08-17
Add GMaven Version test

emilypgoogle:
NO_RELEASE_CHANGE

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenHelper.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenHelper.kt
@@ -37,6 +37,24 @@ class GmavenHelper(val groupId: String, val artifactId: String) {
     return "${GMAVEN_ROOT}/${groupIdAsPath}/${artifactId}/${version}/${fileName}"
   }
 
+  fun hasReleasedVersion(version: String): Boolean {
+    try {
+      val groupIdAsPath = groupId.replace(".", "/")
+      val mavenMetadataUrl = "${GMAVEN_ROOT}/${groupIdAsPath}/${artifactId}/maven-metadata.xml"
+      val factory: DocumentBuilderFactory = DocumentBuilderFactory.newInstance()
+      val builder: DocumentBuilder = factory.newDocumentBuilder()
+      val doc: Document = builder.parse(URL(mavenMetadataUrl).openStream())
+      doc.documentElement.normalize()
+      val versions = doc.getElementsByTagName("version")
+      for (i in 0..versions.length - 1) {
+        if (versions.item(i).textContent == version) {
+          return true
+        }
+      }
+    } catch (e: FileNotFoundException) {}
+    return false
+  }
+
   fun getLatestReleasedVersion(): String {
     try {
       val groupIdAsPath = groupId.replace(".", "/")
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenHelper.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenHelper.kt
@@ -38,24 +38,24 @@ class GmavenHelper(val groupId: String, val artifactId: String) {
   }
 
   fun hasReleasedVersion(version: String): Boolean {
-    try {
-      val groupIdAsPath = groupId.replace(".", "/")
-      val mavenMetadataUrl = "${GMAVEN_ROOT}/${groupIdAsPath}/${artifactId}/maven-metadata.xml"
-      val factory: DocumentBuilderFactory = DocumentBuilderFactory.newInstance()
-      val builder: DocumentBuilder = factory.newDocumentBuilder()
-      val doc: Document = builder.parse(URL(mavenMetadataUrl).openStream())
-      doc.documentElement.normalize()
+    val doc: Document? = getDocument()
+    if (doc != null) {
       val versions = doc.getElementsByTagName("version")
       for (i in 0..versions.length - 1) {
         if (versions.item(i).textContent == version) {
           return true
         }
       }
-    } catch (e: FileNotFoundException) {}
+    }
     return false
   }
 
   fun getLatestReleasedVersion(): String {
+    val doc: Document? = getDocument()
+    return doc?.getElementsByTagName("latest")?.item(0)?.getTextContent() ?: ""
+  }
+
+  fun getDocument(): Document? {
     try {
       val groupIdAsPath = groupId.replace(".", "/")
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenHelper.kt
```diff
@@ -37,6 +37,24 @@ class GmavenHelper(val groupId: String, val artifactId: String) {
     return "${GMAVEN_ROOT}/${groupIdAsPath}/${artifactId}/${version}/${fileName}"
   }
 
+  fun hasReleasedVersion(version: String): Boolean {
```
rlazo, line 40: Can we refactor this and the function below? they seem to do a very similar job

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
```diff
@@ -0,0 +1,56 @@
+package com.google.firebase.gradle.plugins
```
rlazo, line 1: Missing copyright header.

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
```diff
@@ -0,0 +1,56 @@
...
+  fun run() {
+    val mavenHelper = GmavenHelper(groupId.get(), artifactId.get())
+    val latestMavenVersion = mavenHelper.getLatestReleasedVersion()
+    // Either the Maven metadata does not exist, or the library hasn't been released
+    if (latestMavenVersion.isEmpty()) {
+      return
+    }
+    if (version.get() == latestReleasedVersion.get()) {
+      throw GradleException(
+        "Current and latest released versions from gradle.properties are the same (${version.get()})" +
```
rlazo, line 29: Could you use the var names here? IMO it would make it easier to understand what we refer to by "current" and "latest".

rlazo: Here and below

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
```diff
@@ -0,0 +1,56 @@
...
+          "\nLatest release should be ${latestMavenVersion}"
+      )
+    }
+    if (latestMavenVersion == version.get()) {
+      throw GradleException(
+        "Current version from gradle.properties (${version.get()}) is already the latest release on GMaven"
```
rlazo, line 35: maybe add an instruction on how to fix the issue? here and elsewhere

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
```diff
@@ -0,0 +1,56 @@
...
+  @get:Input abstract val groupId: Property<String>
+
+  @get:Input abstract val artifactId: Property<String>
+
+  @get:Input abstract val latestReleasedVersion: Property<String>
+
+  @get:Input abstract val version: Property<String>
+
+  @TaskAction
+  fun run() {
```
rlazo, line 20: The check verifies if 2 different fields are correct

- `latestReleasedVersion` should match the `latest` value in gmaven
- `version` should be one above `latestReleasedVersion`

By only throwing after verifying both fields we can have a better error message that shows both error, if present.

protolite-well-known-types/gradle.properties
```diff
@@ -1,5 +1,5 @@
 # IMPORTANT (b/285892320) Keep version and latestReleasedVersion in sync 
 # unless you are releasing a new version of the library to prevent issues 
 # with transitive dependencies.
-version=18.0.0
+version=18.0.1
```
rlazo, line 4: This is WAI. See the comment above `version`

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
emilypgoogle: With the adjustment to printing an error then including instructions for desired state, I think it might be less confusing to only print a single error with instructions rather than print several related errors, so the release operator doesn't assume there are multiple problems to fix. WDYT?

protolite-well-known-types/gradle.properties
emilypgoogle: I'll pay better attention next time! Added a hardcoded exclusion to protolite-well-known-types

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenHelper.kt
```diff
@@ -38,34 +38,34 @@ class GmavenHelper(val groupId: String, val artifactId: String) {
...
+    }
     return false
   }
 
   fun getLatestReleasedVersion(): String {
+    val doc: Document? = getDocument()
+    return doc?.getElementsByTagName("latest")?.item(0)?.getTextContent() ?: ""
+  }
+
+  fun getDocument(): Document? {
```
rlazo, line 58: too generic

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
```diff
@@ -20,35 +34,43 @@ abstract class GmavenVersionChecker : DefaultTask() {
...
     val mavenHelper = GmavenHelper(groupId.get(), artifactId.get())
     val latestMavenVersion = mavenHelper.getLatestReleasedVersion()
+    val info =
+      "\n  latestReleasedVersion in gradle.properties should match the latest release on GMaven (${latestMavenVersion})" +
+        "\n  version in gradle.properties should be a version bump above this, following SemVer, and should not be released on GMaven"
     // Either the Maven metadata does not exist, or the library hasn't been released
     if (latestMavenVersion.isEmpty()) {
       return
     }
+    // b/285892320
```
rlazo, line 44:
```suggestion
    // TODO(b/285892320): Remove condition when bug fixed
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/GmavenVersionChecker.kt
rlazo: sgtm

rlazo (approved):

PR #5262, opened 2023-08-18
Add a 40 second sleep at the end of each Fireperf e2e test to ensure that events are sent before the app exits.

raymondlam:
This should fix the flaky TAP tests that expect traces to be sent.


visumickey (approved):
Changes look good to me. Let's do a 30 second delay instead of 15.

visumickey (approved):

PR #5264, opened 2023-08-18
Add RC test app e2e config file

qdpham13:
- Add config file to include the RC test app into the github actions CI workflow
- google-services.json has been added as an env var under `REMOTE_CONFIG_E2E_GOOGLE_SERVICES`


.github/workflows/config-e2e.yml
```diff
@@ -0,0 +1,42 @@
+name: Firebase Remote Config Rollouts E2E Tests
+
+on:
+  schedule:
+    - cron: 24 */4 * * *  # every 4 hours @ 24 minutes past the hour
```
danasilver, line 5: Any reason for 24 minutes?

.github/workflows/config-e2e.yml
```diff
@@ -0,0 +1,42 @@
+name: Firebase Remote Config Rollouts E2E Tests
```
danasilver, line 1: Nit: consider generalizing to just "Firebase Remote Config E2E Tests" -- though I think this is changeable later

danasilver (approved):
Couple questions, super minor

.github/workflows/config-e2e.yml
qdpham13: Sure. I included the Rollouts b/c I wasn't sure if we plan to use this in the future as an RC test app and the project has `Rollouts` in the name. But agreed we can change this later if we want.

.github/workflows/config-e2e.yml
qdpham13: Not really, just using the same frequency that Sessions used. This can be updated as we see fit in the future.

PR #5274, opened 2023-08-23
Persist the aqs session id on new aqs session

mrober:
Persist the aqs session id on new aqs session, instead of on event. This is needed for ANRs and native crashes that can't persist on event. 

This also populates the aqs session id and fid in the Report.
The new `CrashlyticsAppQualitySessionsStore` class is based on the existing `MetaDataStore` class but with the current session id not being final.

### DIFF
--- a/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
+++ b/firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
@@ -31,7 +31,6 @@ import com.google.firebase.crashlytics.internal.Logger;
 import com.google.firebase.crashlytics.internal.common.AppData;
 import com.google.firebase.crashlytics.internal.common.BuildIdInfo;
 import com.google.firebase.crashlytics.internal.common.CommonUtils;
-import com.google.firebase.crashlytics.internal.common.CrashlyticsAppQualitySessionsStore;
 import com.google.firebase.crashlytics.internal.common.CrashlyticsAppQualitySessionsSubscriber;
 import com.google.firebase.crashlytics.internal.common.CrashlyticsCore;
 import com.google.firebase.crashlytics.internal.common.DataCollectionArbiter;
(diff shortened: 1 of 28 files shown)


firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/model/CrashlyticsReport.java
```diff
@@ -185,7 +188,7 @@ public CrashlyticsReport withSessionEndFields(
   /** Augment an existing {@link CrashlyticsReport} with the given app quality session id. */
   @NonNull
   public CrashlyticsReport withAppQualitySessionId(@Nullable String appQualitySessionId) {
-    Builder builder = toBuilder();
+    Builder builder = toBuilder().setAppQualitySessionId(appQualitySessionId);
```
themiswang, line 191: Do we want to put down some notes around why we set AQS ID for two different place?

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
```diff
@@ -91,9 +92,12 @@ public class FirebaseCrashlytics {
     final ExecutorService crashHandlerExecutor =
         ExecutorUtils.buildSingleThreadExecutorService("Crashlytics Exception Handler");
 
-    CrashlyticsAppQualitySessionsSubscriber sessionsSubscriber =
-        new CrashlyticsAppQualitySessionsSubscriber(arbiter);
-    firebaseSessions.register(sessionsSubscriber);
+    // TODO(mrober): Can we determine at this point if the app includes the ndk?
+    // If we can set hasNdk to false for non-native apps, there will be less disk io.
+    CrashlyticsAppQualitySessionsStore appQualitySessionsStore =
+        new CrashlyticsAppQualitySessionsStore(fileStore, /* hasNdk= */ true);
```
themiswang, line 98: Where do we set this `hasNdk` to false for SDK app? realized this variable is declared as `final` so we ain't change the value during the app life cycle

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/internal/common/CrashlyticsAppQualitySessionsStore.java
```diff
@@ -0,0 +1,171 @@
...
+  private static void safeDeleteCorruptAqsFile(File aqsFile) {
+    if (aqsFile.exists() && aqsFile.delete()) {
+      Logger.getLogger().i("Deleted corrupt aqs file: " + aqsFile.getAbsolutePath());
+    }
+  }
+
+  private static void checkNotOnMainThread() {
+    // TODO(mrober): Remove this after testing and validation.
+    if (Looper.getMainLooper() == Looper.myLooper()) {
+      throw new IllegalStateException("Running on main thread when expected not to!");
```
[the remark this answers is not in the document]
mrober: I plan to remove this check before landing. This is for validation to ensure we don't accidentally call this on the main thread.

firebase-crashlytics/src/main/java/com/google/firebase/crashlytics/FirebaseCrashlytics.java
mrober: We removed the hasNdk boolean because we can never not persist on session, even if the app does not have the ndk, because it could still have anrs.

themiswang (approved):

kmandrika (approved):

PR #5280, opened 2023-08-25
Implement RolloutsStateSubscriptionHandler with stubbed RolloutsStateFactory.

danasilver:
TODO: Rebase `featureRollouts` after https://github.com/firebase/firebase-android-sdk/pull/5277 is merged.

PR 1 of 3 implementing the pub/sub pattern in RC.

PRs:

1. [Add dependency with stubbed implementation.](https://github.com/firebase/firebase-android-sdk/pull/5277)
2. **This PR: Implement publish and subscribe.**
3. Implement getActiveRolloutsState.


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/rollouts/RolloutsStateSubscriptionsHandler.java
```diff
@@ -16,19 +16,73 @@
...
+      @NonNull Executor executor) {
+    this.activatedConfigsCache = activatedConfigsCache;
+    this.rolloutsStateFactory = rolloutsStateFactory;
+    this.executor = executor;
+  }
 
-  public RolloutsStateSubscriptionsHandler() {}
+  // Thread-safe implementation for subscribers, using a ConcurrentHashMap as the underlying
+  // implementation with set-like accessors.
+  private Set<RolloutsStateSubscriber> subscribers =
```
qdpham13, line 49: Could we initialize this in the constructor?

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
```diff
@@ -610,6 +610,7 @@ private boolean processActivatePutTask(Task<ConfigContainer> putTask) {
       // values from the put task must be non-null.
       if (putTask.getResult() != null) {
         updateAbtWithActivatedExperiments(putTask.getResult().getAbtExperiments());
+        rolloutsStateSubscriptionsHandler.publishActiveRolloutsState(putTask.getResult());
```
qdpham13, line 613: What do you think about creating a new var from ` putTask.getResult()` called something like `activatedConfigs` above line 611 so we just need to call it once and then reference it in the rest of the `if` statement

qdpham13 (commented):
A couple comments

firebase-config/src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfig.java
danasilver: Oh yeah, that sounds good, will do that

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/rollouts/RolloutsStateSubscriptionsHandler.java
danasilver: We could - I think it's more readable to declare and initialize the field at the same time if it's a simple initialization, though, so would prefer to keep as-is. I moved it up above the constructor though

qdpham13: Cool

qdpham13 (approved):
LGTM

PR #5283, opened 2023-08-28
Add Rollouts as a field to diff in Realtime RC

qdpham13:
- Diff rolloutMetadata for the same key as part of Realtime RC
     - Keys can belong to different Rollouts so one key could have multiple Rollouts associated with it
             1. Create a map of maps: { `parameterKey` : { `rolloutId` : `variantId` }}
             2. Compare rolloutMetadata for fetched and activated keys in `getChangedParams`


firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigContainerTest.java
```diff
@@ -350,4 +464,17 @@ private static ImmutableMap<String, String> generateP13nMetadata(
     return ImmutableMap.of(
         PERSONALIZATION_ID, p13nId, ARM_INDEX, armIndex, CHOICE_ID, choiceId, GROUP, group);
   }
+
+  private static JSONArray generateRolloutMetadata(int numberOfMetadata) throws JSONException {
```
danasilver, line 468: not blocking - we could probably improve the pattern for creating metadata (both ABT and Rollouts) since it's not super clear what it means to call `generateRolloutMetadata(1)` in a test.

Maybe an improvement is calling this `generateNRolloutMetadataEntries(int numberOfMetadata)`?

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigContainerTest.java
```diff
@@ -336,6 +336,120 @@ public void configContainer_withRolloutsMetadata_returnsCorrectRolloutsMetadata(
...
+            .withRolloutMetadata(fetchedRolloutMetadata)
+            .build();
+
+    Set<String> changedParams = config.getChangedParams(other);
+
+    assertThat(changedParams).containsExactly("key_1");
+  }
+
+  @Test
+  public void getChangedParams_addNewRolloutMetadata_returnsUpdatedKey() throws Exception {
```
danasilver, line 408:
```suggestion
  public void getChangedParams_addNewKeyWithRolloutMetadata_returnsNewKey() throws Exception {
```

does that describe the test case better?

danasilver (approved):
Couple super minor comments

PR #5293, opened 2023-09-03
Fix Changelog Error

cherylEnkidu:
During auto rebase, the change log entry of auto creation cache index was put inside the previous release cycle by mistake.


dconeybe (approved):
This has happened to me on more than one occasion too!

PR #5294, opened 2023-09-05
Bloom filter failure alt strategy

bswhite1:
The issue is that the bloom filter seems to crash when we have too many things in the query. So it removes all docs from the synced document set and sets them all to "limbo", and we get way to may individual listeners. We also do a hard stop and re-listen on the stream, which adds this target to the event.getTargetMismatches.

So the fix is to set a reset flag, and copy all of the sync documents into a temporary list (resetPending()) Then the files won't be put into limbo if they are part of the pendingResetDocuments list. When we get the stop / re-listen data from the server and it adds the known documents back to the synced document set, we then remove those from the pendingResetDocuments list. Whatever documents now remain in the pendingResetDocuments are truly in "Limbo" and we can move those into the limboDocuments list and request a listen on those as before.

Has been in our custom firebase-android-sdk for over 2 months.


bswhite1:
Fixed with pull request: https://github.com/firebase/firebase-android-sdk/pull/5506

PR #5297, opened 2023-09-05
Restructure library group logic in buildSrc

rlazo:
Library groups are now compute at the plugin, rather than being a singleton class using registrars. This fixes the previous bug that caused duplicate entries to be generated for the release generator because the singleton class was kept alive by the gradle daemon.

Additionally, it simplifies the verification of missing libraries due to library groups. Now, the build **will fail** if the release.json file does not include all the libraries in the library group.


buildSrc/src/main/java/com/google/firebase/gradle/plugins/UpdatePinnedDependenciesTask.kt
```diff
@@ -107,10 +108,13 @@ abstract class UpdatePinnedDependenciesTask : DefaultTask() {
...
-  private fun findProjectLevelDependenciesToChange(): List<FirebaseLibraryExtension> {
+  private fun findProjectLevelDependenciesToChange(
+    libraryGroups: Map<String, List<FirebaseLibraryExtension>>
+  ): List<FirebaseLibraryExtension> {
     val firebaseLibrary = project.firebaseLibrary
 
-    return firebaseLibrary.projectLevelDependencies - firebaseLibrary.librariesToRelease
+    return firebaseLibrary.projectLevelDependencies -
+      libraryGroups.getOrDefault(firebaseLibrary.libraryGroupName, emptyList())
   }
```
daymxn, line 118: Could reduce this to:
```kotlin
  private fun findProjectLevelDependenciesToChange(
    libraryGroups: Map<String, List<FirebaseLibraryExtension>>
  ): List<FirebaseLibraryExtension> =
    with(project.firebaseLibrary) {
      projectLevelDependencies - libraryGroups.getOrDefault(libraryGroupName, emptyList())
    }
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/LibraryGroups.kt
```diff
@@ -0,0 +1,58 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.google.firebase.gradle.plugins
+
+import org.gradle.api.GradleException
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.findByType
+
+fun computeLibraryGroups(project: Project): Map<String, List<FirebaseLibraryExtension>> {
```
daymxn, line 20: Could you add some documentation to these functions?

daymxn (commented):
Left some comments on stuff that could be cleaned up- but nothing that should really _prevent_ merging.

tbh, I very much dislike this way of fetching library groups, and would rather there be a more straightforward way instead of passing a function with the root project. although, this is something I don't think we could really solve until we get to migrating `FirebaseLibraryExtension` to Kotlin. and furthermore, we have more pressing matters- so deff a backlog issue.

All in all though, looking good!!

rlazo:
> 



> Left some comments on stuff that could be cleaned up- but nothing that should really _prevent_ merging.
> 
> tbh, I very much dislike this way of fetching library groups, and would rather there be a more straightforward way instead of passing a function with the root project. although, this is something I don't think we could really solve until we get to migrating `FirebaseLibraryExtension` to Kotlin. and furthermore, we have more pressing matters- so deff a backlog issue.
> 
> All in all though, looking good!!

Hey Daymon, curious to hear more about your ideas. When redesigning where the global library groups information should be, I discarted the `FirebaseLibraryExtension` since it has the view of a single Library, which is not enough information to compute a complete view of the whole universe of libraries and groups. It didn't seem to fit it as an extension function either, since we wouldn't be providing additionally functionality, but additional information, which wasn't there at creation time. Not a fan of that either.

I could have go with encapsulating "LibraryGroups" as a class, but wasn't fully convinced.

daymxn:
> > 
> 
> > Left some comments on stuff that could be cleaned up- but nothing that should really _prevent_ merging.
> > tbh, I very much dislike this way of fetching library groups, and would rather there be a more straightforward way instead of passing a function with the root project. although, this is something I don't think we could really solve until we get to migrating `FirebaseLibraryExtension` to Kotlin. and furthermore, we have more pressing matters- so deff a backlog issue.
> > All in all though, looking good!!
> 
> Hey Daymon, curious to hear more about your ideas. When redesigning where the global library groups information should be, I discarted the `FirebaseLibraryExtension` since it has the view of a single Library, which is not enough information to compute a complete view of the whole universe of libraries and groups. It didn't seem to fit it as an extension function either, since we wouldn't be providing additionally functionality, but additional information, which wasn't there at creation time. Not a fan of that either.
> 
> I could have go with encapsulating "LibraryGroups" as a class, but wasn't fully convinced.

I agree. My main gripe here is the lack of global memoization (while also keeping things lazy)- something we could accomplish with a deferred property in `FirebaseLibraryExtension`- but it'd need to be written in Kotlin (so that it could be added directly to the class). You can't take advantage of delegates like that in extension properties unfortunately.

rlazo:
> > > 
> > 
> > 
> > > Left some comments on stuff that could be cleaned up- but nothing that should really _prevent_ merging.
> > > tbh, I very much dislike this way of fetching library groups, and would rather there be a more straightforward way instead of passing a function with the root project. although, this is something I don't think we could really solve until we get to migrating `FirebaseLibraryExtension` to Kotlin. and furthermore, we have more pressing matters- so deff a backlog issue.
> > > All in all though, looking good!!
> > 
> > 
> > Hey Daymon, curious to hear more about your ideas. When redesigning where the global library groups information should be, I discarted the `FirebaseLibraryExtension` since it has the view of a single Library, which is not enough information to compute a complete view of the whole universe of libraries and groups. It didn't seem to fit it as an extension function either, since we wouldn't be providing additionally functionality, but additional information, which wasn't there at creation time. Not a fan of that either.
> > I could have go with encapsulating "LibraryGroups" as a class, but wasn't fully convinced.
> 
> I agree. My main gripe here is the lack of global memoization (while also keeping things lazy)- something we could accomplish with a deferred property in `FirebaseLibraryExtension`- but it'd need to be written in Kotlin (so that it could be added directly to the class). You can't take advantage of delegates like that in extension properties unfortunately.

I still don't think that this info belongs to `FirebaseLibraryExtension`. I fail to see an scenario in which knowing about library groups is useful at a subproject level. That's why I think this should live at the plugin level

daymxn:
> I still don't think that this info belongs to `FirebaseLibraryExtension`. I fail to see an scenario in which knowing about library groups is useful at a subproject level. That's why I think this should live at the plugin level

Actually, about half of our LG usage happens at the subproject level. For example, both `UpdatePinnedDependenciesTask` and `ReleaseGenerator` get the sibling libraries on a per library basis.

rlazo:
> > I still don't think that this info belongs to `FirebaseLibraryExtension`. I fail to see an scenario in which knowing about library groups is useful at a subproject level. That's why I think this should live at the plugin level
> 
> Actually, about half of our LG usage happens at the subproject level. For example, both `UpdatePinnedDependenciesTask` and `ReleaseGenerator` get the sibling libraries on a per library basis.

I could be wrong, but I see tasks in two groups:

- _Subproject level tasks_, which are tasks that **do not** need to know anything about the world outside their suproject to work. Dackka is a good example, since generating the refdocs for one subproject doesn't depend on the others. What about referencing methos across SDKs? That's a dependency declared in the build file that could point to an artifact in maven.
- _Root project level tasks_, are tasks that require cross subproject knowledge to work. `UpdatePinnedDependenciesTask` needs to know whether two libraries connected by a project level dependency belong to the same library group, which is a global concept. Knowing your own library group tells you nothing about the other libs library group,  you have to go and check. As for `ReleaseGenerator`, we can know at a subproject level whether a particular sdk should be included in the release on its own merit, but that's not enough because we can include them for other reasons. Also, the release generator has to accumulate which sdks to release to create a single config.

daymxn:
> > > I still don't think that this info belongs to `FirebaseLibraryExtension`. I fail to see an scenario in which knowing about library groups is useful at a subproject level. That's why I think this should live at the plugin level
> > 
> > 
> > Actually, about half of our LG usage happens at the subproject level. For example, both `UpdatePinnedDependenciesTask` and `ReleaseGenerator` get the sibling libraries on a per library basis.
> 
> I could be wrong, but I see tasks in two groups:
> 
> * _Subproject level tasks_, which are tasks that **do not** need to know anything about the world outside their suproject to work. Dackka is a good example, since generating the refdocs for one subproject doesn't depend on the others. What about referencing methos across SDKs? That's a dependency declared in the build file that could point to an artifact in maven.
> * _Root project level tasks_, are tasks that require cross subproject knowledge to work. `UpdatePinnedDependenciesTask` needs to know whether two libraries connected by a project level dependency belong to the same library group, which is a global concept. Knowing your own library group tells you nothing about the other libs library group,  you have to go and check. As for `ReleaseGenerator`, we can know at a subproject level whether a particular sdk should be included in the release on its own merit, but that's not enough because we can include them for other reasons. Also, the release generator has to accumulate which sdks to release to create a single config.

Yeah that's a good point. Maybe just an easier way of fetching the data relevant to a given group then? Ideally, I think something like this would be the cleanest and most straightforward:
```kotlin
// straightforward, easy identification of point of truths; but a bit verbose
LibraryGroups.projectsForGroup(library.libraryGroup) 
```
or
```kotlin
library.findSiblings() // a list of libraries in the same group ("sibling" projects)
```

wdyt?

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -147,6 +154,19 @@ abstract class PublishingPlugin : Plugin<Project> {
     }
   }
 
+  protected fun registerLibraryGroupsTaks(
+    project: Project,
+    libraryGroups: Map<String, List<FirebaseLibraryExtension>>
+  ) {
+    project.tasks.register("libraryGroups") {
```
daymxn, line 161: Should make this a companion const, similar to the other tasks, so that if another needs to a dependency on it- we're passing out consts versus assuming the name of a task.

```kotlin
const val LIBRARY_GROUPS_TASK = "libraryGroups"
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -147,6 +154,19 @@ abstract class PublishingPlugin : Plugin<Project> {
     }
   }
 
+  protected fun registerLibraryGroupsTaks(
```
daymxn, line 157: Why `protected` and not `private`?

daymxn, line 157: Should add this to the list of tasks in the docs of `PublishingPlugin`

```
- [LIBRARY_GROUPS_TASK][registerLibraryGroupsTaks]
```

daymxn (approved):
Left some comments regarding alignment with other task designs and documentation, but otherwise looks good to me!

e: Also, make sure the smoke test failure isn't legit

davidmotson (approved):

PR #5300, opened 2023-09-06
Update component runtime to pick up any one of the couroutine dispatcher interface

VinayGuthal:


firebase-components/src/main/java/com/google/firebase/components/ComponentRuntime.java
```diff
@@ -107,6 +108,8 @@ private void discoverComponents(List<Component<?>> componentsToAdd) {
     // instead of executing such code in the synchronized block below, we store it in a list and
     // execute right after the synchronized section.
     List<Runnable> runAfterDiscovery = new ArrayList<>();
+    List<Component<?>> componentsAdding = new ArrayList<>();
```
rlazo, line 111: There's one small behavioural change which out of precaution I think we should restore. The current code works as follows

1. Call `discoverComponents` with the list of `componentsToAdd`
2. During execution, we process pending providers, **and modify** `componentsToAdd`
3. We ensure the list is ok using the `CycleDetector`
4. Do what needs to be done with the components

This change no longer updates `componentsToAdd` to include what is actually going to be processed.


. In the current version, we get as parameter `componentsToAdd` which is the list of things we have to add

firebase-components/src/main/java/com/google/firebase/components/ComponentRuntime.java
```diff
@@ -122,16 +125,41 @@ private void discoverComponents(List<Component<?>> componentsToAdd) {
           Log.w(ComponentDiscovery.TAG, "Invalid component registrar.", ex);
         }
       }
+      for (int i = 0; i < processedComponents.size(); i++) {
+        Object[] interfaces = processedComponents.get(i).getProvidedInterfaces().toArray();
+        for (Object anInterface : interfaces) {
+          if (anInterface.toString().contains("kotlinx.coroutines.CoroutineDispatcher")) {
```
rlazo, line 131: Since this code is very specific to `CoroutineDispatcher` I think you can add a comment explaining why this needs to happen and use var names specific too, like `existingCoroutineDispatcherInterfaces`. This makes it clear that this handles an special case

firebase-components/src/main/java/com/google/firebase/components/ComponentRuntime.java
```diff
@@ -122,16 +125,41 @@ private void discoverComponents(List<Component<?>> componentsToAdd) {
...
+            existingInterfaces.add(anInterface.toString());
+          }
+        }
+      }
+      for (int i = 0; i < componentsToAdd.size(); i++) {
```
rlazo, line 136: Do you have any idea how expensive these loops can be? Since this code runs during initialization and, AFAIR in the main thread, any extra work will impact start up for all apps.

An alternative would be to set a class field boolean that works as a flag of whether we have seen the dispatcher before. Then, we would only need to loop once, and we wouldn't need to create an additional list and copy over every component (`componentsAdding`) nor the list of interfaces already seen (`existingInterfaces`)

WDYT?

rlazo:
> ## Unit Test Results
>   24 files  +  18    24 suites  +18   16s [⏱️](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) -2s 115 tests +  97  115 [✔️](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) +  97  0 [💤](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) ±0  0 [❌](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) ±0  230 runs  +194  230 [✔️](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) +194  0 [💤](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) ±0  0 [❌](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) ±0 
> 
> Results for commit [e6481e2](https://github.com/firebase/firebase-android-sdk/commit/e6481e27c0043790652a7e0c9f062db6e0db4206). ± Comparison against base commit [fe84885](https://github.com/firebase/firebase-android-sdk/commit/fe84885ec87d7457926812ab4ecfa70bf8583581).
> 
> This pull request removes 18 and adds 115 tests. Note that renamed tests count towards both.
> ♻️ This comment has been updated with latest results.

@VinayGuthal the "This pull request removes 18 and adds 115 tests. Note that renamed tests count towards both." sounds  wrong, any ideas?

firebase-components/src/main/java/com/google/firebase/components/ComponentRuntime.java
```diff
@@ -123,6 +124,24 @@ private void discoverComponents(List<Component<?>> componentsToAdd) {
...
+      // kotlinx.coroutines.CoroutineDispatcher interface could be provided by both new version of
+      // firebase-common and old version of firebase-common-ktx. In this scenario take the first
+      // interface which was provided.
+
+      Iterator<Component<?>> it = componentsToAdd.iterator();
+      while (it.hasNext()) {
+        Component component = it.next();
+        for (Object anInterface : component.getProvidedInterfaces().toArray()) {
+          if (anInterface.toString().contains("kotlinx.coroutines.CoroutineDispatcher")) {
+            if (processedCoroutineDispatcherInterfaces) {
```
rlazo, line 136: You can merge this condition with the one above

rlazo:
Please double check this crash

https://firebase.corp.google.com/project/fireescape-c4819/testlab/histories/bh.150fab898fb191d1/matrices/5367404277897084480/executions/bs.beb53886721b2b11

rlazo (approved):

PR #5303, opened 2023-09-07
Add fetch and crash buttons to test app. Fix transitive deps.

danasilver:


firebase-config/test-app/src/main/kotlin/com/google/firebase/testing/config/MainActivity.kt
```diff
@@ -17,14 +17,28 @@
...
 class MainActivity : AppCompatActivity() {
+
+  private lateinit var remoteConfig: FirebaseRemoteConfig
+
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_main)
 
     findViewById<TextView>(R.id.greeting_text).text = getText(R.string.firebase_greetings)
+    findViewById<Button>(R.id.crash_button).setOnClickListener {
```
themiswang, line 35: Can we call this "JVM Crash"? Now I am thinking about adding more type of crashes and integrate with NDK later for this test app later.

danasilver: SGTM. Updated.

samedson (approved):

PR #5304, opened 2023-09-07
Check for null ConfigContainer before constructing RolloutsState.

danasilver:
It's possible the ConfigContainerTask returns null while the ConfigContainer is being instantiated from the disk cache (ex. at app startup when Crashlytics registers a handler.)


themiswang (approved):
Thanks for looking into it!

qdpham13 (approved):
I still wonder what the root cause of the problem is; why is registerRolloutsStateSubscriber being invoked without an explicit call

danasilver:
Yeah, it might be nice to improve how `ConfigContainer#get()` works to avoid having to check for null or make it more obvious.

Crashlytics calls `registerRolloutsStateSubscriber` when the app starts up here https://github.com/firebase/firebase-android-sdk/compare/master...featureRollouts#diff-447eefe42b216df1d7d22d64d8cd444756272587be28cdcc6ac9c58f92fc7571R38

Issue #5102, opened 2023-06-23
Getting StackOverflowError crash when recordException in invoked in crashlytics


aloysiusralph:
### Environment

  * Android Studio version: Android Studio Dolphin | 2021.3.1 Patch 1
  * Firebase Component: Crashlytics
  * Component version: 18.0.0 +( Firebase BOM 28.0.0+)

#### Steps to reproduce:

We are recording rxjava exceptions using the recordException method in FirebaseCrashlytics class. It is working fine with firebase BOM-27.1.0 (crashlytics 17.4.1).

Recently when we upgraded the firebase BOM to 32.1.0 (crashlytics 18.3.7), when the recordException is invoked, it is resulting in the StackOverflowError crash. With further investigations, we found that it is happening from BOM 28.0.0(crashlytics 18.0.0)

The issue is intermittent and has 8/10 reproducibility.

The crash is happening in TrimmedThrowableData class where it ends up in recursive loop because the exception sent here is having infinite number of causes.

Upon further debugging with the firebase crashlytics source code, we found that the exception has finite causes up until writeNonFatalException method in CrashlyticsController class. Inside the CrashlyticsBackgroundWorker runnable the same exception has infinite causes.

#### Crash Logs:

```
   E  FATAL EXCEPTION: Crashlytics Exception Handler1

                                                                                                   Process: com.mobile.android.dev, PID: 7371

                                                                                                    java.lang.StackOverflowError: stack size 1039KB

                                                                                                    at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:126)

                                                                                                    at java.lang.AbstractStringBuilder.appendNull(AbstractStringBuilder.java:494)

                                                                                                    at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:447)

                                                                                                    at java.lang.StringBuilder.append(StringBuilder.java:137)

                                                                                                    at com.google.firebase.crashlytics.internal.stacktrace.TrimmedThrowableData.<init>(TrimmedThrowableData.java:32)

                                                                                                    at com.google.firebase.crashlytics.internal.stacktrace.TrimmedThrowableData.<init>(TrimmedThrowableData.java:42)

                                                                                                    at com.google.firebase.crashlytics.internal.stacktrace.TrimmedThrowableData.<init>(TrimmedThrowableData.java:42)

[...]

PR #5310, opened 2023-09-08
Don't use recursion in TrimmedThrowableData to avoid stack overflow

mrober:
Don't use recursion in TrimmedThrowableData. This fixes an issue when exceptions with long caused by chains cause a stack overflow.

Pair with PolinaGo


mrober:
#5102

kmandrika (approved):

ralphaloysius:
@mrober Thanks for addressing this.  Any idea when this fix will be released in firebase bom?

aloysiusralph:
@mrober Thanks for addressing this. Any idea when this will be released?

PR #5312, opened 2023-09-11
firebase-perf: Add custom method to set final static method for test

utzcoz:
Robolectric's ReflectionHelpers removes supporting of setting fnial static method becasuse it not the reposibility of Robolectric, and it will not be allowed by the latest JDK. This CL just adds custom method to do the same things as a workaround. When the project updates the JDK someday, the maintainer needs to rewrite the test or provides some test helper method to avoid set final static method for test.


rlazo:
> @rlazo This is another preparation PR for Robolectric 4.10.3 as it removes the supporting of set final static field from `ReflectionHelpers` because the team thinks it is not the Robolectric's responsibility. This PR add a workaround to implement the same function by Firebase's test itself.
> 
> Also I found `logGaugeMetric_globalCustomAttributesAreNotAdded` is common flaky on CI although my changes don't effect it. Could you help take a look at it? Thanks.

Yes, I double check locally using your branch and it's passing. When ready, feel free to assign me as reviewer of this PR. Thanks!

rlazo:
@visumickey could you take a look at this change? Thanks!

rlazo:
Thanks a lot for your contribution @utzcoz !

PR #5317, opened 2023-09-11
Add ProcessDetails to sessions sdk

mrober:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -51,36 +51,26 @@ internal constructor(
...
 
     val appContext = firebaseApp.applicationContext.applicationContext
     if (appContext is Application) {
-      appContext.registerActivityLifecycleCallbacks(sessionInitiator.activityLifecycleCallbacks)
-
-      firebaseApp.addLifecycleEventListener { _, _ ->
-        Log.w(TAG, "FirebaseApp instance deleted. Sessions library will not collect session data.")
-        appContext.unregisterActivityLifecycleCallbacks(sessionInitiator.activityLifecycleCallbacks)
-      }
+      isAppDefaultProcess = ProcessUtils.isDefaultProcess(appContext)
```
samedson, line 70: Could we add a test by passing this method in and mocking it?

jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
```diff
@@ -0,0 +1,84 @@
+/*
```
jrothfeder, line 1: Consider adding a test.

jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+import android.app.ActivityManager.RunningAppProcessInfo
+import android.app.Application
+import android.content.Context
+import android.os.Build
+import android.os.Process
+
+/** Utils related to processes. */
+internal object ProcessUtils {
+  /** Returns whether the current process should generate a new session or not. */
+  fun shouldProcessGenerateNewSession(context: Context): Boolean =
```
visumickey, line 29: Should this function move out of this file? I feel that the decision maker for this should be in something like `SessionInitiator` rather than `ProcessUtils`.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+    isDefaultProcess(context) && isForegroundImportance()
+
+  /** Returns whether the current process is the app's default process or not. */
+  private fun isDefaultProcess(context: Context): Boolean =
+    getCurrentProcessName(context) == getDefaultProcessName(context)
+
+  /** Returns whether the current process or service is running in the foreground or not. */
+  private fun isForegroundImportance(): Boolean {
```
visumickey, line 37: I'm not sure what this function is supposed to return. Are we checking if the current process is in Foreground?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
```diff
@@ -0,0 +1,84 @@
...
+   * Returns if the current process is running as a foreground service.
+   *
+   * This generally indicates that the process is doing something the user actively cares about.
+   */
+  private fun isForegroundService(runningAppProcessInfo: RunningAppProcessInfo): Boolean =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+      runningAppProcessInfo.importance == RunningAppProcessInfo.IMPORTANCE_FOREGROUND_SERVICE
+    } else {
+      // Default to false if the api level is too low to check.
+      false
```
visumickey, line 82: What happens in these scenarios? We would not generate a session?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
jrothfeder: Disagree. This needs to be checked outside of `SessionInitiator`.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
jrothfeder: Correct. If  shouldProcessGenerateNewSession == false, then we do not generate a session. :)

jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
visumickey: I'm fine with not doing it in SessionInitiator. The challenge is to add this function in the ProcessUtils. Ideally ProcessUtils should serve information about the process - but the decision of should I start a session or not should be made of this utility class (ideally in the place where we determine if we should start a session).

One of the choice I proposed is to do it somewhere [here](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionInitiator.kt#L58-L65) in `SessionInitiator` class. If we have a better place, totally fine.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
visumickey: Interesting - in that case, any lower API versions, would we be not able to generate sessions? Should we record this as a limitation? Do we know the number of instances in this API version? We don't have to go deeper to this. But getting a sense of the limitation would be helpful to make a decision.

jrothfeder: Unless I'm missing something, the code checks if it's a foreground process OR if it's a foreground service. So, if it's unable to determine if it's a foreground service, it will still generate a session if it's a foreground process. Is this your take as well?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessUtils.kt
jrothfeder: Yeah, that class comes in to play after we've decided that the process is going to generate sessions. It won't be involved at all if the process should not generate sessions.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetails.kt
```diff
@@ -0,0 +1,74 @@
...
+
+  /** The name of the current process, or null if it couldn't be found. */
+  private val currentProcessName: String? =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+      Application.getProcessName()
+    } else {
+      findProcessName(context, Process.myPid())
+    }
+
+  override val isDefaultProcess: Boolean = currentProcessName == defaultProcessName
```
visumickey, line 53: Would this value be updated everytime when referenced? Was thinking if this variable needs a get() pattern with it. Something like below:

```
override val isDefaultProcess: Boolean = 
get() {
  return currentProcessName == defaultProcessName
}
```

mrober: I had it like that at first, but I realized the process name is stable so it'll never give a different answer.

PR #5318, opened 2023-09-11
Reference ConfigContainer metadata for Rollouts constants.

danasilver:
Fixes a NPE parsing Rollouts metadata.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
@@ -37,9 +37,9 @@ public class ConfigContainer {
   static final String PERSONALIZATION_METADATA_KEY = "personalization_metadata_key";
   static final String TEMPLATE_VERSION_NUMBER_KEY = "template_version_number_key";
   static final String ROLLOUT_METADATA_KEY = "rollout_metadata_key";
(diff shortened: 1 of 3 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
```diff
@@ -37,9 +37,9 @@ public class ConfigContainer {
   static final String PERSONALIZATION_METADATA_KEY = "personalization_metadata_key";
   static final String TEMPLATE_VERSION_NUMBER_KEY = "template_version_number_key";
   static final String ROLLOUT_METADATA_KEY = "rollout_metadata_key";
-  static final String ROLLOUT_METADATA_AFFECTED_KEYS = "affectedParameterKeys";
-  static final String ROLLOUT_METADATA_ID = "rolloutId";
-  static final String ROLLOUT_METADATA_VARIANT_ID = "variantId";
+  public static final String ROLLOUT_METADATA_AFFECTED_KEYS = "affectedParameterKeys";
```
danasilver, line 40: Doesn't block this PR, but wondering if these constants could be snake case like the others

qdpham13: No, they're the key values within RolloutMetadata. They might actually be able to be moved to `RemoteConfigConstants` in the future.

qdpham13 (approved):

PR #5329, opened 2023-09-14
Use constants for library group task name

rlazo:
Addressess pending comments from PR #5297

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
@@ -106,7 +106,7 @@ abstract class PublishingPlugin : Plugin<Project> {
       registerSemverCheckForReleaseTask(project, releasingProjects)
       registerPublishAllToBuildDir(project, allFirebaseLibraries)
       registerPostReleasePlugin(releasingProjects)
-      registerLibraryGroupsTaks(project, libraryGroups)
+      registerLibraryGroupsTask(project, libraryGroups)
 
       val buildMavenZip =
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -155,11 +156,11 @@ abstract class PublishingPlugin : Plugin<Project> {
     }
   }
 
-  protected fun registerLibraryGroupsTaks(
+  private fun registerLibraryGroupsTaks(
```
daymxn, line 159:
```suggestion
  private fun registerLibraryGroupsTask(
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -61,6 +61,7 @@ import org.gradle.kotlin.dsl.register
  * - [PUBLISH_RELEASING_LIBS_TO_LOCAL_TASK][registerPublishReleasingLibrariesToMavenLocalTask]
  * - [SEMVER_CHECK_TASK][registerSemverCheckForReleaseTask]
  * - [PUBLISH_ALL_TO_BUILD_TASK][registerPublishAllToBuildDir]
+ * - [LIBRARY_GROUPS_TASK][registerLibraryGroupsTaks]
```
daymxn, line 64:
```suggestion
 * - [LIBRARY_GROUPS_TASK][registerLibraryGroupsTask]
```

daymxn (commented):
Would also need to update it at the call site, but I can't comment on that in this PR since it wasn't changed.

daymxn (approved):

PR #5331, opened 2023-09-14
Add deep transitive projects to release

davidmotson:


buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -220,18 +220,40 @@ abstract class PublishingPlugin : Plugin<Project> {
...
+
     return ReleaseMetadata(librariesToRelease, releaseName)
   }
 
+  /**
+   * Finds all transitive libraries that need to be released so that the input list of project names
+   * can be released. This includes all project level dependencies, and anything in the same library
+   * group.
+   */
+  private fun getDeepTransitiveReleases(
```
daymxn, line 234: Could separate the loop into a `tailrec` function, and if there was a `LibraryGroup` class or `Object` (singleton) then this would be much cleaner as well:

```kotlin
private tailrec fun getDeepTransitiveReleases(
  librariesToRelease: List<FirebaseLibraryExtension>,
): List<FirebaseLibraryExtension> {
  val groupsToRelease = librariesToRelease.flatMap { it.projectLevelDeps }.map { it.libraryGroupName }
  val projects = LibraryGroups.projectsForGroups(groupsToRelease)

  return if(librariesToRelease == projects) librariesToRelease else getDeepTransitiveReleases(projects)
}
```
 
The alternative would be doing something like mapping library groups to adjacent library groups, eg;

```kotlin
LibraryGroup {
  fun getSiblingGroups(group: String): List<String>
}
// or...
class LibraryGroup {
  val siblings: List<LibraryGroup>
}
```

And then taking advantage of that by doing something like:
```kotlin
private tailrec fun getDeepTransitiveReleases(
  librariesToRelease: List<FirebaseLibraryExtension>,
): List<FirebaseLibraryExtension> {
  val groupsToRelease = librariesToRelease.flatMap { it.projectLevelDeps }.map { it.libraryGroup.siblings }
  
  return LibraryGroups.projectsForGroups(groupsToRelease)
}
```

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
```diff
@@ -218,18 +218,34 @@ abstract class PublishingPlugin : Plugin<Project> {
...
+
+    return ReleaseMetadata(librariesToRelease, releaseName)
+  }
+
+  /**
+   * Finds all transitive libraries that need to be released so that the input list of project names
+   * can be released. This includes all project level dependencies, and anything in the same library
+   * group.
+   */
+  private tailrec fun getDeepTransitiveReleases(
```
rlazo, line 234: For this function to be tailrecursive, wouldn't it need to have as its last argument `inputGroups`? Not sure since kotlin's refdocs have a single arg function as example.

daymxn: wdym `inputGroups`?

davidmotson:
The test that's edited captures the case this is meant to fix, if a library is pulled in without its project dependencies, those get added to the release anyway. 

I could add a test case with LGs and project deps if you think that'd be valuable

daymxn:
> The test that's edited captures the case this is meant to fix, if a library is pulled in without its project dependencies, those get added to the release anyway.
> 
> I could add a test case with LGs and project deps if you think that'd be valuable

I think it would be actually. We could always use more tests for our publishing pipeline.

davidmotson:
Done!

buildSrc/src/main/java/com/google/firebase/gradle/plugins/PublishingPlugin.kt
rlazo: I meant `inputProjects` :-D 

Talked about it with @davidmotson offline, LGTM.

rlazo (approved):

PR #5350, opened 2023-09-21
Remove `SettingsProvider` abstraction.

jrothfeder:
We don't benefit from this abstraction and as Local and Remote settings diverge it will only serve to get in our way.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettingsProvider.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettingsProvider.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2023 Google LLC
+ *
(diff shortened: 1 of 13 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/settings/RemoteSettingsProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.sessions.settings
+
+import kotlin.time.Duration
+
+internal interface RemoteSettingsProvider {
```
samedson, line 21: Do we have plans to make these diverge? It looks like Remote and Local Settings are still the same interface

PR #5351, opened 2023-09-21
Functions migration to Kotlin

emilypgoogle:
done with automated tooling, to investigate the viability of quick migrations pre API changes.


firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.kt
```diff
@@ -0,0 +1,383 @@
...
+  // Emulator settings
+  private var emulatorSettings: EmulatedServiceSettings? = null
+
+  init {
+    client = OkHttpClient()
+    serializer = Serializer()
+    this.contextProvider = Preconditions.checkNotNull(contextProvider)
+    this.projectId = Preconditions.checkNotNull(projectId)
+    val isRegion: Boolean
+    isRegion = try {
```
daymxn, line 87: Did the converter do this?

emilypgoogle: Yep

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Options for configuring the callable function.
+ *
+ *
+ * These properties are immutable once a callable function reference is instantiated.
+ */
+class HttpsCallableOptions private constructor(
+        /**
+         * Returns the setting indicating if limited-use App Check tokens are enforced for this function.
+         */
+        // If true, request a limited-use token from AppCheck.
```
emilypgoogle, line 26: Comments got doubled up here with the method and field comments, does this break javadoc?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
```diff
@@ -0,0 +1,49 @@
...
+        val limitedUseAppCheckTokens: Boolean) {
+
+  /** Builder class for [com.google.firebase.functions.HttpsCallableOptions]  */
+  class Builder {
+    /** Returns the setting indicating if limited-use App Check tokens are enforced.  */
```
emilypgoogle, line 31: This builder getter got moved to the field description. Desired?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
```diff
@@ -11,29 +11,20 @@
...
-
+/** The result of calling a HttpsCallableReference function.  */
+class HttpsCallableResult internal constructor( // The actual result data, as generic types decoded from JSON.
+        private val data: Any?) {
   /**
    * Returns the data that was returned from the Callable HTTPS trigger.
    *
-   * <p>The data is in the form of native Java objects. For example, if your trigger returned an
+   *
+   * The data is in the form of native Java objects. For example, if your trigger returned an
```
emilypgoogle, line 23: <p> was removed, is this fine?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
```diff
@@ -11,29 +11,20 @@
...
-public class HttpsCallableResult {
-  // The actual result data, as generic types decoded from JSON.
-  private final Object data;
-
-  HttpsCallableResult(Object data) {
-    this.data = data;
-  }
-
+/** The result of calling a HttpsCallableReference function.  */
+class HttpsCallableResult internal constructor( // The actual result data, as generic types decoded from JSON.
```
emilypgoogle, line 17: Is this where field comments should be going?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
```diff
@@ -11,29 +11,20 @@
...
-   * <p>The data is in the form of native Java objects. For example, if your trigger returned an
+   *
+   * The data is in the form of native Java objects. For example, if your trigger returned an
    * array, this object would be a List<Object>. If your trigger returned a JavaScript object with
-   * keys and values, this object would be a Map<String, Object>.
-   */
-  @Nullable
-  public Object getData() {
-    return data;
+   * keys and values, this object would be a Map<String></String>, Object>.
```
emilypgoogle, line 25: This parser assumed `Map<String, Object>` was HTML tags and messed with it.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallOptions.kt
```diff
@@ -45,30 +38,32 @@ class HttpsCallOptions {
...
-  OkHttpClient apply(OkHttpClient client) {
-    return client
-        .newBuilder()
-        .callTimeout(timeout, timeoutUnits)
-        .readTimeout(timeout, timeoutUnits)
-        .build();
+  companion object {
+    // The default timeout to use for all calls.
+    private const val DEFAULT_TIMEOUT: Long = 70
+    private val DEFAULT_TIMEOUT_UNITS = TimeUnit.SECONDS
```
emilypgoogle, line 67: Why did `DEFAULT_TIMEOUT_UNITS` not get converted to a const when `DEFAULT_TIMEOUT` was?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
```diff
@@ -0,0 +1,159 @@
...
+   * Changes the timeout for calls from this instance of Functions. The default is 60 seconds.
+   *
+   * @param timeout The length of the timeout, in the given units.
+   * @param units The units for the specified timeout.
+   */
+  fun setTimeout(timeout: Long, units: TimeUnit) {
+    options.setTimeout(timeout, units)
+  }
+
+  val timeout: Long
```
emilypgoogle, line 140: This created a field just for a getter. Is this desireable?

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.kt
```diff
@@ -0,0 +1,383 @@
...
+     * @param regionOrCustomDomain The region or custom domain for the HTTPS trigger, such as `"us-central1"` or `"https://mydomain.com"`.
+     */
+    @JvmStatic
+    fun getInstance(regionOrCustomDomain: String): FirebaseFunctions {
+      return getInstance(FirebaseApp.getInstance(), regionOrCustomDomain)
+    }
+
+    val instance: FirebaseFunctions
+      /** Creates a Cloud Functions client with the default app.  */
+      get() = getInstance(FirebaseApp.getInstance(), "us-central1")
```
emilypgoogle, line 381: The converter removed a getter and also created a field just for a getter here.

emilypgoogle (commented):
Left some comments on notable issues

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
```diff
@@ -0,0 +1,159 @@
...
+  // The name of the HTTPS endpoint this reference refers to.
+  // Is null if url is set.
+  private val name: String?
+
+  // The url of the HTTPS endpoint this reference refers to.
+  // Is null if name is set.
+  private val url: URL?
+
+  // Options for how to do the HTTPS call.
+  private val options: HttpsCallOptions
```
daymxn, line 34: this should actually be `package-private`, did you manually put the `private` here? Doesn't really matter in this case, but interesting thing to look out for.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
```diff
@@ -0,0 +1,49 @@
...
+ *
+ *
+ * These properties are immutable once a callable function reference is instantiated.
+ */
+class HttpsCallableOptions private constructor(
+        /**
+         * Returns the setting indicating if limited-use App Check tokens are enforced for this function.
+         */
+        // If true, request a limited-use token from AppCheck.
+        val limitedUseAppCheckTokens: Boolean) {
```
daymxn, line 27: would also need a `@JvmField` annotation, with a getter provided, to avoid a breaking change.

daymxn (commented):
Some key things I was seeing to look out for:

- `package-private` members
- public properties (that had public getters or setters) need `@JvmField` to prevent a breaking change

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
daymxn: Nah, that's fine

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
daymxn: Yeah, that's fine. Will still be interacted with the same in Kotlin, and have the same byte-code for JVM.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
daymxn: yeah, in order to retain the same implicit calling in Kotlin- it needs to be a variable.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
daymxn: Yeah that's fine, but it's worth keeping an eye out for- if it removes anything more important.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
daymxn: it really do be like that

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
daymxn: no, this is a *bad* 👎  ktfmt moment

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallOptions.kt
daymxn: Because Kotlin only allows primitives and strings to be `const`

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.kt
daymxn: yeah- seems like getters/setters are gonna be a moderate hurdle

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
emilypgoogle: This is going to cause us problems when we want to keep exposing values as `public static final` in Java, is there a workaround?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
emilypgoogle: I didn't touch this privacy, the converter must've opted to turn a package variable to a private one

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
```diff
@@ -0,0 +1,49 @@
...
+
+  /** Builder class for [com.google.firebase.functions.HttpsCallableOptions]  */
+  class Builder {
+    /** Returns the setting indicating if limited-use App Check tokens are enforced.  */
+    var limitedUseAppCheckTokens = false
+      private set
```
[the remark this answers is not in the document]
thatfiredev: I'm probably dreaming too high here, but what if we dropped the Builder completely and just had a simple `HttpsCallableOptions` class that could be passed to `getHttpsCallable()`? It would be a breaking change, but would simplify Kotlin usage, and I believe this change will come out in a release with other breaking changes anyway.

Something like:

```kotlin
class HttpsCallableOptions @JvmOverloads public constructor(
  @JvmField
  val limitedUseAppCheckTokens = false
) {

  init {
    // if the original `build()` method was doing any kind of validation,
    // we could've done it here. But in this particular case, it is not.
  }

}
```

Kotlin would call it like:

```
val options = HttpsCallableOptions(limitedUseAppCheckTokens = true)
```

Java:
```
HttpsCallableOptions options = new HttpsCallableOptions(true);
```

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
daymxn: I'm not sure I understand your question? Could you give me an example of where this would cause an issue?

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
```diff
@@ -11,29 +11,20 @@
...
+  </Object> */
+  fun getData(): Any? {
+    return data
   }
```
thatfiredev, line 29: What if we had a `component1()` here to [destructure](https://kotlinlang.org/docs/destructuring-declarations.html) `HttpsCallableResult`?

That way developers would call:
```kotlin
val (data) = task.result
```

instead of:
```kotlin
val data = task.result?.data
```

---

```suggestion
  }
  
  operator fun component1() = data
```

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * @return A Task that will be completed when the HTTPS request has completed.
+   */
+  fun call(): Task<HttpsCallableResult?> {
+    return if (name != null) {
+      functionsClient.call(name, null, options)
+    } else {
+      functionsClient.call(url!!, null, options)
+    }
+  }
```
thatfiredev, line 128: We could probably have a single function with default values here annotated with `@JvmOverloads` to let the compiler generate the other function with no args.

```
  @JvmOverloads
  fun call(data: Any? = null): Task<HttpsCallableResult?> {
    return if (name != null) {
      functionsClient.call(name, data, options)
    } else {
      functionsClient.call(url!!, data, options)
    }
  }
```

Also, do we want to have a `suspend` function version of this?

firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.kt
```diff
@@ -0,0 +1,383 @@
...
+  fun getHttpsCallable(
+          name: String, options: HttpsCallableOptions): HttpsCallableReference {
+    return HttpsCallableReference(this, name, HttpsCallOptions(options))
+  }
+
+  /** Returns a reference to the callable HTTPS trigger with the provided URL and call options.  */
+  fun getHttpsCallableFromUrl(
+          url: URL, options: HttpsCallableOptions): HttpsCallableReference {
+    return HttpsCallableReference(this, url, HttpsCallOptions(options))
+  }
```
thatfiredev, line 123: We could reduce this down to 2 functions by using default arguments and `@JvmOverloads`:

```kotlin
  @JvmOverloads
  fun getHttpsCallable(name: String, options: HttpsCallableOptions? = null): HttpsCallableReference {
    return HttpsCallableReference(this, name, HttpsCallOptions(options))
  }

  @JvmOverloads
  fun getHttpsCallableFromUrl(
          url: URL, options: HttpsCallableOptions? = null): HttpsCallableReference {
    return HttpsCallableReference(this, url, HttpsCallOptions(options))
  }
```

thatfiredev (commented):
I added some of my thoughts/suggestions/wishes for the API surface.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableOptions.kt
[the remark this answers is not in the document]
daymxn: > and I believe this change will come out in a release with other breaking changes anyway.

The intention is to avoid this, actually. We wanna [quickly] convert SDKs to Kotlin without breaking anyone, then go back and redesign the APIs as desired. It may make less sense for something as small as functions- but with the larger SDKs it'll make it significantly easier to make progressive changes once everything's in Kotlin. Although, with the new breaking change policy- we may want to revisit that idea.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
daymxn: `@JvmOverloads` would be perfect here- but it would also change the docs, since (as it currently stands) they have different docs.

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableResult.kt
daymxn: As far as changing the API- this will probably end up getting removed entirely and replaced by the method taking a generic instead. Alternatively, this could be converted into a data class instead- to automatically get `component1()`:

```kotlin
data class HttpsCallableResult internal constructor(val data: Any?)
```

Or, even better (assuming you wanted to keep the class):
```kotlin
data class HttpsCallableResult<T> internal constructor(val data: T?)
```

thatfiredev: `data` classes are discouraged in our [best practices](go/fb-kotlin-best-practices#api-surface) 🥲  _(externally visible [link](https://jakewharton.com/public-api-challenges-in-kotlin/) for non-googlers)_

But I do like the idea of having a class that takes a generic :)

daymxn (changes requested):
The test failures are valid (esp the unit tests for functions). I'll see if I can't personally take a look later

emilypgoogle:
> The test failures are valid (esp the unit tests for functions). I'll see if I can't personally take a look later

Took a look and I can't repro the unit/integ test failures locally, wondering what's causing the failure in CI

firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallableReference.kt
emilypgoogle: This became visible for testing

daymxn (approved):
So long as the smoke test is a flake, LGTM

PR #5352, opened 2023-09-21
Testing composite index queries against production 

milaGGL:
port from: https://github.com/firebase/firebase-js-sdk/pull/7632


firebase-firestore/src/androidTest/java/com/google/firebase/firestore/CompositeIndexQueryTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+import com.google.firebase.firestore.testutil.CompositeIndexTestHelper;
+import com.google.firebase.firestore.testutil.IntegrationTestUtil;
+import java.util.Map;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(AndroidJUnit4.class)
```
milaGGL, line 29: TODO(mila):add the class comment on.

ehsannas (commented):
LGTM. Please add the few additional functions (updateDoc, getDoc, and getDocs) similar to the Web.

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/CompositeIndexTestHelper.java
```diff
@@ -0,0 +1,178 @@
...
+  public Task<Void> deleteDoc(@NonNull DocumentReference document) {
+    return document.delete();
+  }
+
+  // Retrieves a single document from Firestore with test-specific fields removed.
+  @NonNull
+  public Map<String, Object> getDoc(@NonNull DocumentReference document) {
+    DocumentSnapshot docSnapshot = waitFor(document.get());
+    return removeTestSpecificFieldsFromDoc(docSnapshot.getData());
+  }
```
ehsannas, line 166: I was hoping we could return a DocumentSnapshot here, but that's not easily possible.

We could try doing it using the `MutableDocument.newFoundDocument()`, but then we'd have to use `UserDataReader` and convert the `Map<String, Object>` to `Map<String, Value>` and we'd also have to set `SnapshotVersion.NONE`. All of this to construct a synthetic document snapshot. It's not worth it I think, given that the caller can compare to the expected document's data.

milaGGL: Yes, I have wrestled with the types as well, data in a snapshot cannot be directly modified like Web does, neither we can create a new DocumentSnapshot based on the info on hand. Currently, getDoc and getDocs in most of the testings always followed by "...ToValues" or "...ToIds". So focusing on the value itself is good enough for composite index as well.

ehsannas: SG

ehsannas (approved):

rlazo (approved):

PR #5360, opened 2023-09-26
Extract SessionMaintainer interface to allow for different types of S…

jrothfeder:
…essionMaintainers to run

depending on charcteristics of the process that owns them.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -32,7 +32,8 @@ import kotlinx.coroutines.CoroutineDispatcher
  * The [FirebaseSessions] API provides methods to register a [SessionSubscriber] and delegates to a
  * [SessionMaintainer].
  */
-class FirebaseSessions(
+class FirebaseSessions
+internal constructor(
   firebaseApp: FirebaseApp,
   firebaseInstallations: FirebaseInstallationsApi,
   backgroundDispatcher: CoroutineDispatcher,
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -16,7 +16,6 @@
 
 package com.google.firebase.sessions
 
-import android.app.Application
 import android.util.Log
 import com.google.android.datatransport.TransportFactory
 import com.google.firebase.FirebaseApp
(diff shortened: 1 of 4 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -26,146 +25,52 @@ import com.google.firebase.ktx.Firebase
...
 
-    val sessionInitiateListener =
-      object : SessionInitiateListener {
-        // Avoid making a public function in FirebaseSessions for onInitiateSession.
-        override suspend fun onInitiateSession(sessionDetails: SessionDetails) {
-          initiateSessionStart(sessionDetails)
-        }
-      }
+  private val sessionMaintainer: SessionMaintainer
+  private val tag = "FirebaseSessions"
```
mrober, line 44: Why change the const TAG below and add this? Is this left over from splitting?

jrothfeder: because it requires a companion object which this class has, however, others do not. I figured since every class is going to need one of these that we'd define it in a uniform way that doesn't require adding companion objects. Thoughts?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/leader/SessionMaintainerLeader.kt
```diff
@@ -0,0 +1,170 @@
...
+        appContext.unregisterActivityLifecycleCallbacks(sessionInitiator.activityLifecycleCallbacks)
+      }
+    } else {
+      Log.e(
+        TAG,
+        "Failed to register lifecycle callbacks, unexpected context ${appContext.javaClass}."
+      )
+    }
+  }
+  override fun register(subscriber: SessionSubscriber) {
```
mrober, line 102: Will register need to be different for leaders and followers? I thought this would stay in the FirebaseSessions class.

mrober (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: Ok that's fine. Can you still make them `const` though?

jrothfeder: Strangely no 
image

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/leader/SessionMaintainerLeader.kt
jrothfeder: Yes. Followers will pass something different. The will pass different things.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/leader/SessionMaintainerLeader.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    if (subscribers.values.none { it.isDataCollectionEnabled }) {
+      Log.d(TAG, "Data Collection is disabled for all subscribers. Skipping this Session Event")
+      return
+    }
+
+    Log.d(TAG, "Data Collection is enabled for at least one Subscriber")
+
+    // This will cause remote settings to be fetched if the cache is expired.
+    sessionSettings.updateSettings()
```
visumickey, line 138: I assume this is asynchronous. So anything below will execute based on the previously cached value of settings fetch?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -26,146 +25,53 @@ import com.google.firebase.ktx.Firebase
...
 
-    val sessionInitiator =
-      SessionInitiator(
-        timeProvider,
+  init {
+    // TODO(rothbutter): create a different maintainer based on the characteristics of this process
```
visumickey, line 48: Curiousity question: How is the current maintainer leader different from the TODO maintainer and what characteristics does this depend on?

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionMaintainer.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Interface for the class that maintains all aspects of the App Quality Session for this process
+ * that includes data synchronizing across processes, sending events to our backend, and
+ * broadcasting AQS related events to listeners
+ */
+interface SessionMaintainer {
+
+  /** Register a listener for updates to the session being maintained by this class */
+  fun register(subscriber: SessionSubscriber)
```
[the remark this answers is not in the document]
jrothfeder: no

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
jrothfeder: I assure you that your curiosity will be satisfied in the next CL :)

Suffice to say that it will be significantly different from the leader.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/leader/SessionMaintainerLeader.kt
jrothfeder: No.

visumickey: In that case, is the updateSettings a blocking call? Was curious on the scenario of when will a settings change be reflecting on the client from the moment of change on the backend.

mrober: It's a suspend function and it'll pause until the settings are fetched. Then after this line, settings have been updated.

PR #5368, opened 2023-09-27
Add `SessionMaintainerFollower` and create in `FirebaseSessions` when…

jrothfeder:
… we're a follower process.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetails.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetails.kt
@@ -30,6 +30,12 @@ internal interface ProcessDetails {
 
   /** Whether the current process is running in the foreground or not. */
   val isForegroundProcess: Boolean
+
+  /** Name of this process */
+  val processName: String?
+
+  /** Name of the default process */
+  val defaultProcessName: String
 }
 
 /** Android implementation of [ProcessDetails]. */
@@ -40,23 +46,23 @@ internal class AndroidProcessDetails(context: Context) : ProcessDetails {
    * This is the app's package name unless the app overrides the android:process attribute in the
    * application block of its Android manifest file.
    */
-  private val defaultProcessName: String = context.applicationInfo.processName
+  override val defaultProcessName: String = context.applicationInfo.processName
 
-  /** The name of the current process, or null if it couldn't be found. */
-  private val currentProcessName: String? =
+  /** The name of this process, or null if it couldn't be found. */
+  override val processName: String? =
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
       Application.getProcessName()
     } else {
       findProcessName(context, Process.myPid())
     }
 
-  override val isDefaultProcess: Boolean = currentProcessName == defaultProcessName
+  override val isDefaultProcess: Boolean = processName == defaultProcessName
 
   override val isForegroundProcess: Boolean
     get() {
       val runningAppProcessInfo = RunningAppProcessInfo()
       ActivityManager.getMyMemoryState(runningAppProcessInfo)
-      return runningAppProcessInfo.importance == RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+      return runningAppProcessInfo.importance <= RunningAppProcessInfo.IMPORTANCE_FOREGROUND_SERVICE
     }
 
   /** Finds the process name for the given pid, or returns null if not found. */
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionMaintainerFollower.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionMaintainerFollower.kt
@@ -1,0 +1,50 @@
+package com.google.firebase.sessions.follower
+
+import android.util.Log
+import com.google.firebase.sessions.SessionMaintainer
+import com.google.firebase.sessions.api.FirebaseSessionsDependencies
+import com.google.firebase.sessions.api.SessionSubscriber
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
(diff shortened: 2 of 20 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetails.kt
```diff
@@ -40,23 +46,23 @@ internal class AndroidProcessDetails(context: Context) : ProcessDetails {
...
 
-  override val isDefaultProcess: Boolean = currentProcessName == defaultProcessName
+  override val isDefaultProcess: Boolean = processName == defaultProcessName
 
   override val isForegroundProcess: Boolean
     get() {
       val runningAppProcessInfo = RunningAppProcessInfo()
       ActivityManager.getMyMemoryState(runningAppProcessInfo)
-      return runningAppProcessInfo.importance == RunningAppProcessInfo.IMPORTANCE_FOREGROUND
+      return runningAppProcessInfo.importance <= RunningAppProcessInfo.IMPORTANCE_FOREGROUND_SERVICE
```
mrober, line 65: I think this should go back to `runningAppProcessInfo.importance == RunningAppProcessInfo.IMPORTANCE_FOREGROUND` because `IMPORTANCE_FOREGROUND_SERVICE` is only available after a specific api level, and it'll make our logic complicated to also count services as leaders.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionsDataRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+    val SESSION_ID = stringPreferencesKey("session_id")
+  }
+
+  val firebaseSessionDataFlow: Flow<FirebaseSessionsData> =
+    context.dataStore.data
+      .catch { exception ->
+        Log.e(tag, "Error reading stored session data.", exception)
+        emit(emptyPreferences())
+      }
+      .map { preferences -> mapSessionsData(preferences) }
```
mrober, line 47: Can this be something like `.map(::mapSessionsData)` ?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/FakeProcessDetails.kt
```diff
@@ -22,4 +22,6 @@ import com.google.firebase.sessions.ProcessDetails
 class FakeProcessDetails(
   override val isDefaultProcess: Boolean = true,
```
mrober, line 23: If we have the process name and default process name in the interface, do we need this isDefaultProcess anymore?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetails.kt
jrothfeder: the one integration test fails if we don't consider foreground services to be foreground processes https://github.com/firebase/firebase-android-sdk/blob/master/firebase-sessions/test-app/src/androidTest/kotlin/com/google/firebase/testing/sessions/FirebaseSessionsTest.kt

That suggests to me that we may need it. Thoughts?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionsDataRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+import androidx.datastore.preferences.core.stringPreferencesKey
+import androidx.datastore.preferences.preferencesDataStore
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.map
+
+data class FirebaseSessionsData(val sessionId: String?)
+
+/** Persists session data that needs to be synchronized across processes */
+class SessionsDataRepository(private val context: Context) {
```
mrober, line 34: These classes should all be `internal` visibility.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionsDataRepository.kt
jrothfeder: that didn't work

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionMaintainerFollower.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.google.firebase.sessions.api.SessionSubscriber
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+class SessionMaintainerFollower(private val sessionsDataRepository: SessionsDataRepository) :
+  SessionMaintainer {
+  val tag = "SessionMaintainerFollow"
+
+  override fun register(subscriber: SessionSubscriber) {
```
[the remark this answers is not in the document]
mrober: Is a no-op the right thing to do for followers? Should they still notify the subscribers on register? 

Perf relies on the AQS session id being set before the call to `firebaseSessions.register` returns. Crashlytics is ok without an AQS session id.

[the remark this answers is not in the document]
jrothfeder: Okay. We can force a lookup.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/testing/FakeProcessDetails.kt
jrothfeder: I'll remove.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/follower/SessionsDataRepository.kt
jrothfeder: 👍

PR #5378, opened 2023-10-02
Add FirebaseSessionsProvider and use it to detect cold starts

mrober:
Add FirebaseSessionsProvider and use it to detect cold starts. This works because the content provider gets created exactly once per app instance, regardless of multiples processes. Tested locally using test apps.


firebase-sessions/src/main/AndroidManifest.xml
```diff
@@ -18,7 +18,14 @@
   <!--<uses-sdk android:minSdkVersion="16"/>-->
 
   <application>
-    <service android:name="com.google.firebase.components.ComponentDiscoveryService"
+    <provider
+      android:name="com.google.firebase.sessions.provider.FirebaseSessionsProvider"
+      android:authorities="${applicationId}.firebasesessionsprovider"
+      android:directBootAware="true"
+      android:exported="false"
+      android:initOrder="100" />
```
visumickey, line 26: I really like this idea. But is there a way we can rely on the [FirebaseInitProvider](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-common/src/main/java/com/google/firebase/provider/FirebaseInitProvider.java) and use that to detect cold star? Are we limited by lack of access to that provider?

visumickey, line 26: I assume this is late to initialize. Hopefully that does not create a concern for us? Later we define this, the sessionID generation might get delayed and the underlying SDKs (Fireperf) might be in a wait state to get this sessionID for too long?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/provider/FirebaseSessionsProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.content.ContentValues
+import android.database.Cursor
+import android.net.Uri
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Provider to detect cold starts based on the fact the [ContentProvider] will be created exactly
+ * once per app instance, regardless of multiple processes.
+ */
+internal class FirebaseSessionsProvider : ContentProvider() {
```
visumickey, line 29: It is an internal class, any tests to make it easier to validate?

jrothfeder (approved):

firebase-sessions/src/main/AndroidManifest.xml
mrober: One of the common ways to setup firebase in a multi-process app is to disable that provider. I don't think we should rely on it. So we can have our own provider, and we're planning on moving some more logic into it.

mrober: This will still happen before application oncreate.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/provider/FirebaseSessionsProvider.kt
mrober: Added tests

PR #5390, opened 2023-10-10
Update common version

VinayGuthal:


firebase-common/gradle.properties
```diff
@@ -1,3 +1,3 @@
-version=20.4.1
-latestReleasedVersion=20.4.0
+version=20.4.2
+latestReleasedVersion=20.4.1
```
daymxn, line 2: is this expected?

VinayGuthal: yes

VinayGuthal: there is already an admrt push to release firebase-common at 20.4.1 .. This is essentially updating that

daymxn (approved):

PR #5391, opened 2023-10-10
Remove deprecated ktx packages from sessions

mrober:


jrothfeder (approved):

firebase-sessions/firebase-sessions.gradle.kts
```diff
@@ -48,8 +48,7 @@ android {
 }
 
 dependencies {
-  api("com.google.firebase:firebase-common:20.4.0")
```
VinayGuthal, line 51: We would need dependency on both firebase-common and firebase-common-ktx. This will prevent issues wrt collision when used with the old version of firebase-common-ktx

PR #5395, opened 2023-10-10
Fix client-side throttling for Play Integrity flows

rosalyntan:
Changes:
- Successful `GeneratePlayIntegrityChallenge` call should not reset `RetryManager` backoff
- Cache in-flight App Check token fetching tasks
- Update `403` errors to throttle with exponential backoff rather than a flat one-day period (to match [iOS](https://github.com/firebase/firebase-ios-sdk/blob/f3a14c249e001b3c894bb943ecc3973f50e60a49/FirebaseAppCheck/Sources/Core/Backoff/FIRAppCheckBackoffWrapper.m#L264-L268))

### DIFF
--- a/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
+++ b/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
@@ -21,7 +21,6 @@ import android.content.pm.PackageManager.NameNotFoundException;
 import android.util.Log;
 import androidx.annotation.IntDef;
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import com.google.android.gms.common.util.AndroidUtilsLight;
 import com.google.android.gms.common.util.Hex;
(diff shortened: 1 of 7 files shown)


appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
```diff
@@ -138,11 +139,13 @@ public String generatePlayIntegrityChallenge(
     }
     URL url =
         new URL(String.format(PLAY_INTEGRITY_CHALLENGE_URL_TEMPLATE, projectId, appId, apiKey));
-    return makeNetworkRequest(url, requestBytes, retryManager);
+    // We don't forward the RetryManager, as we don't want a successful
```
malcolmdeck, line 142: This is also going to stop invoking `retryManager.updateBackoffOnFailure()`, so now this generatePlayIntegrityChallenge call can spam our backend. Is that an acceptible tradeoff, or do we need to still pass the retryManager but pass in a flag instead? (which I know is more annoying, interface-wise).

malcolmdeck (changes requested):

appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/NetworkClient.java
rosalyntan: Good call, done.

malcolmdeck (approved):

PR #5397, opened 2023-10-11
Introduces a SessionDataService which is bound to by clients in each …

bryanatkinson:
Introduces a new SessionDataService that is bound to by each process and used to deliver session-ids.

### DIFF
--- a/firebase-sessions/src/main/AndroidManifest.xml
+++ b/firebase-sessions/src/main/AndroidManifest.xml
@@ -19,7 +19,7 @@
 
   <application>
     <service
-      android:name=".SessionDataService"
+      android:name="com.google.firebase.sessions.SessionDataService"
       android:enabled="true"
       android:exported="false" />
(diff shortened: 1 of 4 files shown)


firebase-sessions/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,10 @@
   <!--<uses-sdk android:minSdkVersion="16"/>-->
 
   <application>
+    <service
+      android:name=".SessionDataService"
```
mrober, line 22: Will this `.` carry on to the app when the manifests get merged? Will this have a different name for every app depending on it's app id? Or will it always be `...firebase-sessions.SessionDataService` ?

bryanatkinson: I still see it show up as `com.google.firebase.sessions.SessionDataService` in stack traces when running in an app, but I'll make this fully qualified like the other one.

mrober (approved):
LGTM knowing this is still for debugging

PR #5399, opened 2023-10-11
New datastore for holding session id and timestamp

jrothfeder:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDatastore.kt
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private fun mapSessionsData(preferences: Preferences): FirebaseSessionsData =
+    FirebaseSessionsData(
+      preferences[FirebaseSessionDataKeys.SESSION_ID],
+      preferences[FirebaseSessionDataKeys.TIMESTAMP_MS]
+    )
+}
+
+const val SESSION_CONFIGS_NAME = "firebase_session_settings"
```
mrober: I don't think we can put a `val` directly in the file like this. It will cause Kotlin to generate a class like `FirebaseSessionsDataKt` so Java can access it, and I never found a way to make that class not public. So this will cause a public api change. Instead this can go in a private companion object.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDatastore.kt
```diff
@@ -0,0 +1,78 @@
...
+import androidx.datastore.preferences.core.longPreferencesKey
+import androidx.datastore.preferences.core.stringPreferencesKey
+import androidx.datastore.preferences.preferencesDataStore
+import java.io.IOException
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.map
+
+/** Datastore for sessions information */
+data class FirebaseSessionsData(val sessionId: String?, val timestampMs: Long?)
```
mrober: Can you make these internal?

mrober: Instead of ms can we use us? The time provider provides time in microseconds.

jrothfeder: I believe you mean "µs"

mrober (approved):

PR #5400, opened 2023-10-11
Adds simple logic to only broadcast sessions for cold starts and for …

bryanatkinson:
…when the app has been backgrounded for too long. Also ignores messages that came in out of order.

Still need to hook this into the persistence layer and FirebaseSessions for actually generating sessions.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataService.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataService.kt
@@ -37,10 +37,11 @@ import java.util.UUID
 
 /** Service for providing access to session data */
 internal class SessionDataService : Service() {
-  /** Target we publish for clients to send messages to IncomingHandler. */
-  private lateinit var messenger: Messenger
 
-  val boundClients = mutableListOf<Messenger>()
+  private val boundClients = mutableListOf<Messenger>()
+  private var curSessionId: String? = null
+  private var isAppInForeground: Boolean = false
+  private var lastMsgTimeMs: Long = 0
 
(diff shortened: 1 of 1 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataService.kt
```diff
@@ -37,65 +37,81 @@ import java.util.UUID
...
-    }
+  fun handleForegrounding(msg: Message) {
+    Log.i(TAG, "SERVICE: Activity foregrounding at ${msg.getWhen()}")
 
-    fun handleBackgrounding() {
-      Log.i(TAG, "SERVICE: Activity backgrounding")
+    if (curSessionId == null) {
+      Log.i(TAG, "SERVICE: Cold start detected.")
+      newSession()
+    } else if (!isAppInForeground && msg.getWhen() - lastMsgTimeMs > MAX_BACKGROUND_MS) {
```
jrothfeder, line 72: Why are we keeping track of the foreground state of the app? Is it to prevent multiple sessions from being sent? How about update the timestamp on a foreground instead?

jrothfeder (changes requested):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataService.kt
bryanatkinson: I think we need to know that the app is already in foreground for cases like split screens or multiple windows on tablets. So if an app is already foregrounded and we get another foreground event from a different activity, we don't want to trigger a new session if it has been 30m since the last foreground/background came in.

PR #5403, opened 2023-10-12
Splits the SessionDataService into a service and client

bryanatkinson:
Created a SessionLifecycleService class which contains all the service-side code, and a SessionLifecycleClient object which handles the connection to the server and the sending and receiving of events.

Also added documentation to the code.


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * The maximum number of messages that we should queue up for delivery to the
+   * [SessionLifecycleService] in the event that we have lost the connection.
+   */
+  const val MAX_QUEUED_MESSAGES = 20
+
+  private var service: Messenger? = null
+  private var serviceBound: Boolean = false
+  private val queuedMessages = LinkedList<Message>()
```
jrothfeder, line 36: does this need to be thread-safe?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -0,0 +1,154 @@
...
+  /** Sends a message to the [SessionLifecycleService] with the given event code. */
+  private fun sendLifecycleEvent(messageCode: Int) {
+    sendLifecycleEvent(Message.obtain(null, messageCode, 0, 0))
+  }
+
+  /** Sends the given [Message] to the [SessionLifecycleService]. */
+  private fun sendLifecycleEvent(msg: Message) {
+    if (service != null) {
+      try {
+        service?.send(msg)
```
jrothfeder, line 133: If this succeeds, then why not drain the queue here as well?

jrothfeder (changes requested):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
bryanatkinson: Updated to use LinkedBlockingDeque

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  /**
+   * Should be called when any activity in this application process goes to the foreground. This
+   * will relay the event to the [SessionLifecycleService] where it can make the determination of
+   * whether or not this foregrounding event should result in a new session being generated.
+   */
+  fun foregrounded(activity: Activity): Unit {
+    sendLifecycleEvent(SessionLifecycleService.FOREGROUNDED)
```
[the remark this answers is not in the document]
bryanatkinson: Updated to use LinkedBlockingDeque

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
bryanatkinson: Updated to always try to drain the latest messages on send.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /**
+   * Sends a message to the [SessionLifecycleService] with the given event code. This will
+   * potentially also send any messages that have been queued up but not successfully delivered to
+   * thes service since the previous send.
+   */
+  private fun sendLifecycleEvent(messageCode: Int) {
+    val allMessages = ArrayList<Message>()
+    queuedMessages.drainTo(allMessages)
```
jrothfeder, line 144: Maybe this line and the line above could be extracted to a method `drainQueuedToNewList()`? You do it a couple of times.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+
+  /**
+   * Queues the given [Message] up for delivery to the [SessionLifecycleService] once the connection
+   * is established.
+   */
+  private fun queueMessage(msg: Message) {
+    Log.i(TAG, "Queueing message ${msg.what}")
+    while (queuedMessages.size >= MAX_QUEUED_MESSAGES) {
```
jrothfeder, line 185: This isn't really thread-safe (two threads could be doing this at the same time with unpredictable results). Though it's probably okay, we can also probably live without it.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -0,0 +1,181 @@
...
+      queuedMessages.removeFirst()
+    }
+    // Note: it is possible this still fails to insert if another thread inserts in the meantime.
+    // However this indicates that many lifecycle events are ocurring and so a missed message is
+    // very unlikely to result in a missed session.
+    queuedMessages.offer(msg)
```
jrothfeder, line 191: maybe log the result (true if success, false if full)?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Service for monitoring application lifecycle events and determining when/if a new session should
+ * be generated. When this happens, the service will broadcast the updated session id to all
+ * connected clients.
+ */
+internal class SessionLifecycleService : Service() {
+
+  private val boundClients = mutableListOf<Messenger>()
+  private var curSessionId: String? = null
+  private var isAppInForeground: Boolean = false
```
jrothfeder, line 42: Didn't we agree that this could be dropped?

jrothfeder (changes requested):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
bryanatkinson: Removed.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
bryanatkinson: Done

jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -0,0 +1,176 @@
...
+  /** Sends the current session id to the client connected through the given [Messenger]. */
+  private fun sendSessionToClient(client: Messenger) {
+    try {
+      client.send(
+        Message.obtain(null, SESSION_UPDATED, 0, 0).also {
+          it.setData(Bundle().also { it.putString(SESSION_UPDATE_EXTRA, curSessionId) })
+        }
+      )
+    } catch (e: Exception) {
+      if (e is DeadObjectException) {
```
mrober, line 128: Why not catch `DeadObjectException` and then catch `Exception` ?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
bryanatkinson: Done

mrober (approved):

PR #5407, opened 2023-10-12
Rename SessionCoordinator to SessionFirelogPublisher and add `getInstance` so it can be accessed from a bound service.

jrothfeder:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionFirelogPublisher.kt
```diff
@@ -35,21 +35,24 @@ internal class SessionCoordinator(
...
       eventGDTLogger.log(sessionEvent)
 
-      Log.i(TAG, "Successfully logged Session Start event: ${sessionEvent.sessionData.sessionId}")
+      Log.i(tag, "Successfully logged Session Start event: ${sessionEvent.sessionData.sessionId}")
     } catch (ex: RuntimeException) {
-      Log.e(TAG, "Error logging Session Start event to DataTransport: ", ex)
+      Log.e(tag, "Error logging Session Start event to DataTransport: ", ex)
     }
   }
+  val tag = "SessionFirelogPublisher"
```
bryanatkinson, line 51: Just out of curiosity, why have this here instead of as a const in the companion object?

bryanatkinson (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionFirelogPublisher.kt
jrothfeder: yeah, I don't know. Felt like a const in a companion object was heavy, but I suppose I can stick it there.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionFirelogPublisher.kt
```diff
@@ -49,7 +49,11 @@ internal class SessionCoordinator(
     }
   }
 
-  companion object {
-    private const val TAG = "SessionCoordinator"
+  internal companion object {
+    const val TAG = "SessionFirelogPublisher"
+
```
mrober, line 54: Can you add an instance val like this to make it easier to call?

```@JvmStatic
val instance: SessionFirelogPublisher
  get() = getInstance(Firebase.app)

mrober (approved):

PR #5410, opened 2023-10-12
Specify notification's dismiss intent target via action instead of co…

gsakakihara:
…mponent name.

* Changed createMessagingPendingIntent() to create an Intent targeting the BroadcastReceiver via action and package instead of specifying the ComponentName directly. This will make it easier to change FirebaseInstanceIdReceiver's class name in the future.

### DIFF
--- a/firebase-messaging/CHANGELOG.md
+++ b/firebase-messaging/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Unreleased
+* [changed] Specified notification's dismiss intent target via action instead of
+  component name.
+
 * [changed] Added Kotlin extensions (KTX) APIs from `com.google.firebase:firebase-messaging-ktx`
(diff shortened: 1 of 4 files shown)


xiaohegong (approved):

willxu-google:
Whats the implication of this change? ANRs with FirebaseInstanceIdReceiver broadcast exceeding 10 seconds will start reporting the app package name in the ANR instead?
Any other significant changes?

willxu-google (approved):

gsakakihara:
> Whats the implication of this change? ANRs with FirebaseInstanceIdReceiver broadcast exceeding 10 seconds will start reporting the app package name in the ANR instead? Any other significant changes?

No, this doesn't affect ANRs, it only changes how the intent is targeted so that it matches via the action instead of a hardcoded class name.

PR #5413, opened 2023-10-12
Fix a bunch of warnings on this class so that bryan can continue to have vimean superpowers.

jrothfeder:


bryanatkinson (approved):
Haha, thanks :)

mrober (approved):

PR #5416, opened 2023-10-13
Hooks up the SessionGenerator to the SessionLifecycleService

bryanatkinson:
Using the SessionGenerator to generate new sessions and create SessionDetails instead of directly creating UUID session ids in the service.

Also updated the boundClients queue to be a LinkedBlockingQueue to avoid ConcurrentModificationExceptions if a client binds while a message is being handled by the server.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
@@ -28,7 +28,7 @@ import android.os.Looper
 import android.os.Message
 import android.os.Messenger
 import android.util.Log
-import java.util.UUID
+import java.util.concurrent.LinkedBlockingQueue
 
 /**
  * Service for monitoring application lifecycle events and determining when/if a new session should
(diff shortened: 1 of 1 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -89,12 +107,14 @@ internal class SessionLifecycleService : Service() {
   private fun handleForegrounding(msg: Message) {
     Log.i(TAG, "Activity foregrounding at ${msg.getWhen()}")
 
-    if (curSessionId == null) {
+    if (!hasForegrounded) {
       Log.i(TAG, "Cold start detected.")
-      newSession()
+      hasForegrounded = true
+      broadcastSession()
```
jrothfeder, line 113: Don't we need to generate a new session here too?

jrothfeder (changes requested):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
bryanatkinson: The sessionGenerator creates an initial session right away, so on cold start we need to use that one instead of generating a new one: https://github.com/firebase/firebase-android-sdk/blob/42ec418ad57a40230636481937fca541b0c6c8ad/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionGenerator.kt#L42

jrothfeder (approved):

PR #5420, opened 2023-10-13
Wire datastore in to FirebaseLifecycleService

jrothfeder:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -28,14 +28,21 @@ import android.os.Looper
...
+import kotlinx.coroutines.launch
 
 /**
  * Service for monitoring application lifecycle events and determining when/if a new session should
  * be generated. When this happens, the service will broadcast the updated session id to all
  * connected clients.
  */
-internal class SessionLifecycleService : Service() {
+internal class SessionLifecycleService(
+  private var datastore: SessionDatastore = SessionDatastore(Firebase.app.applicationContext))
```
bryanatkinson: Is Firebase.app guaranteed to exist when this is constructed?

bryanatkinson (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
jrothfeder: I think so. Isn't it created first?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -106,15 +125,27 @@ internal class SessionLifecycleService : Service() {
...
       broadcastSession()
+      updateSessionStorage(SessionGenerator.instance.currentSession.sessionId, msg.getWhen())
+    }
+  }
+
+  private fun updateSessionStorage(timeMs: Long) {
+    updateSessionStorage(null, timeMs)
+  }
+
+  private fun updateSessionStorage(sessionId: String?, timeMs: Long) {
```
mrober, line 145: You could say `sessionId: String = null` here instead of the manual overloading with and without sessionId.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -133,14 +164,19 @@ internal class SessionLifecycleService : Service() {
...
 
-  private fun sendSessionToClient(client: Messenger) {
+  private fun maybeSendSessionToClient(client: Messenger) {
     if (hasForegrounded) {
       sendSessionToClient(client, SessionGenerator.instance.currentSession.sessionId)
+    } else {
+      // Send the value from the datastore before the first foregrounding it exists
+      val sessionData = currentSessionFromDatastore.get()
+      if (sessionData != null) {
+        sessionData.sessionId?.let { sendSessionToClient(client, it) }
```
mrober, line 177: `sessionData?.sessionId?.let` will remove the if null check.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -59,6 +59,7 @@ internal object SessionLifecycleClient {
    * The callback class that will be used to receive updated session events from the
    * [SessionLifecycleService].
    */
+  // TODO(rothbutter) should we use the main looper or is there one available in this SDK?
```
mrober, line 62: This is fine for now. I'll make the firebase common ones easy to access.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -83,6 +101,7 @@ internal class SessionLifecycleService : Service() {
         }
       }
       lastMsgTimeMs = msg.getWhen()
+      updateSessionStorage(lastMsgTimeMs) // TODO(rothbutter) don't think we need this, eh?
```
mrober, line 104: Not sure. If we don't, that will be much better. I am a bit worried about using a timestamp like this, because clocks can change with daylight savings and leap seconds and whatever. In the `WallClock` object there is `elapsedRealtime` which is a monotonic increasing clock so won't  go backwards no matter what happens on the calendar.  Otherwise an hour forward might make everything a new session.

mrober (approved):

PR #5421, opened 2023-10-13
Add composite index testing guidance to ReadMe

milaGGL:
Add guidance to ReadMe File
Add message for terraform workflow to ignore "index already exists" error message.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -121,6 +121,7 @@ jobs:
         with:
           credentials_json: ${{ secrets.GCP_SERVICE_ACCOUNT }}
       - uses: google-github-actions/setup-gcloud@v0
+      
       # create composite indexes with Terraform
       - name: Setup Terraform 
         if: contains(matrix.module, ':firebase-firestore')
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -135,7 +135,7 @@ jobs:
         if: github.event_name == 'pull_request' && contains(matrix.module, ':firebase-firestore')
         run: |
           cd firebase-firestore
-          terraform apply -var-file=../google-services.json -auto-approve
+          terraform apply -var-file=../google-services.json -auto-approve &> /dev/null
```
ehsannas, line 138: This effectively silences the output of the command. Is it intentional?

milaGGL: Yes, the terraform apply kept repeating the "index already created" error message, and what we can do to eliminate the error message is very limited. Silencing it for not to avoid alerting everyone who saw the message.

ehsannas: but what if `terraform apply` fails for _another_ legitimate reason? then we won't be able to tell the failure cause because it will be missing from the logs.

milaGGL: Terraform is used for index creation right at the moment. If there are changes like project/permission change that eventually leads to the failure in index creation, this will be reflected on the test failure (index missing). Then we need to trace back where index were created and remove the `&> /dev/null`.

ehsannas (approved):

PR #5422, opened 2023-10-13
Only support the default Firebase app

mrober:
This will make it safe to call `Firebase.app` any time from anywhere in this SDK.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
@@ -32,7 +32,7 @@ internal constructor(
   private val firebaseApp: FirebaseApp,
   internal val backgroundDispatcher: CoroutineDispatcher,
   private val sessionFirelogPublisher: SessionFirelogPublisher,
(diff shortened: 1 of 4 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -32,7 +32,7 @@ internal constructor(
   private val firebaseApp: FirebaseApp,
   internal val backgroundDispatcher: CoroutineDispatcher,
   private val sessionFirelogPublisher: SessionFirelogPublisher,
-  private val sessionGenerator: SessionGenerator,
+  sessionGenerator: SessionGenerator,
```
bryanatkinson, line 35: Why is the session generator needed here?

bryanatkinson (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
mrober: We pass it to the SessionInitiator in the init block

PR #5427, opened 2023-10-16
Hooks up the SessionLifecycleService to the FirelogPublisher, and has…

bryanatkinson:
… the FirelogPublisher handle the full SessionEvent creation since that's only relevant to Firelog.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
@@ -29,6 +29,7 @@ import android.os.Messenger
 import android.util.Log
 import com.google.firebase.Firebase
 import com.google.firebase.app
+import com.google.firebase.sessions.settings.SessionsSettings
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.atomic.AtomicReference
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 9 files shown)


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -167,6 +169,14 @@ internal class SessionLifecycleService : Service() {
       intent.getParcelableExtra<Messenger>(CLIENT_CALLBACK_MESSENGER)
     }
 
+  /**
+   * Determines if the foregrounding that occurred at the given time should trigger a new session
+   * because the app has been idle for too long.
+   */
+  private fun isSessionRestart(foregroundTimeMs: Long) =
```
mrober, line 176: The old SessionInitiator used `Duration`s not `Long`s for this check. The `WallClock` has a better clock for measuring durations that won't be affected by funny things like the user changing the time or anything with daylight savings etc. See my comment https://github.com/firebase/firebase-android-sdk/pull/5420#discussion_r1360766953

bryanatkinson: This is getting the time out of the messages themselves where we don't actually know which clock was used to generate them other than that the framework did. Switching to use our own clock would mean that we'd be keeping track of timestamps based on when the service processes each message as opposed to when the event occurred. It's true that manipulating the system clock might cause a session to be dropped or created when it shouldn't have, but IMO that's a much less important or likely scenario than delays in message delivery.

mrober (approved):

PR #5430, opened 2023-10-17
Enable setting OkHttpClient for FirebaseFunctions

### LINKED ISSUE #5404: Functions: Allow setting OkHttpClient instance
## What feature would you like to see?

I would like to be able to set or change the `OkHttpClient` instance of `FirebaseFunctions`.
Currently this is not possible because the private field is initialized in the [constructor](https://github.com/firebase/firebase-android-sdk/blob/a135c514ee4d6a48cfc6680d7a35e97e6a666736/firebase-functions/src/main/java/com/google/firebase/functions/FirebaseFunctions.java#L100).

This would enable using a shared and preconfigured `OkHttpClient` instance which could lead to improved I/O performance.

## How would you use it?

As of now I only want to configure a custom `User-Agent` string so that we can distinguish between different clients that execute callable functions in the backend.

svenjacobs:
This PR fixes #5404. I deliberately did not add the `OkHttpClient` instance to the `getInstance` factory methods because this would mean four additional variants of `getInstance`. I believe setting the HTTP client instance is a rare use case, so this is solved through an additional setter.

### DIFF
--- a/firebase-functions/src/androidTest/java/com/google/firebase/functions/FirebaseFunctionsTest.kt
+++ b/firebase-functions/src/androidTest/java/com/google/firebase/functions/FirebaseFunctionsTest.kt
@@ -19,6 +19,7 @@ import com.google.firebase.FirebaseApp
 import com.google.firebase.FirebaseOptions
 import com.google.firebase.functions.FirebaseFunctions.Companion.getInstance
 import junit.framework.TestCase.assertEquals
+import okhttp3.OkHttpClient
 import org.junit.Assert
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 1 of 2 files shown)


svenjacobs:
The PR has been rebased onto the Kotlin migration of `main`.

svenjacobs:
@thatfiredev Some explanation:

I created the `setOkHttpClient` and `getOkHttpClient` functions in Kotlin although I didn't have to. I could have just opened the variable `client`. However from the naming it might not be obvious what `client` is. There are three possibilities now.

1. Leave the setter and getter functions as is
2. Remove the functions and make `client` public accessible
  ```kotlin
  public var client: OkHttpClient
  ```

  Additionally we could add more descriptive names via annotations here. But these would only be visible from Java.

  ```kotlin
  public var client: OkHttpClient = OkHttpClient()
    @JvmName("setOkHttpClient")
    set
    @JvmName("getOkHttpClient")
    get
  ```
3. Remove the functions, rename `client` to `okHttpClient` and make it public accessible. We would have a descriptive name for both Kotlin and Java.

What do you think?

PR #5431, opened 2023-10-17
Updates the service to run on a looper on it's own thread and not use…

bryanatkinson:
… the MainLooper which is the application main thread and so could interfere with things there. Also fixes issue with missing import.


mrober (approved):

jrothfeder (approved):
image

PR #5437, opened 2023-10-17
Wait for a valid SessionID before logging the app start event.

visumickey:


firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -374,15 +386,27 @@ public synchronized void onActivityResumed(Activity activity) {
...
   }
 
+  private void dispatchAppStartTrace() {
+    // Log the app start trace in a non-main thread.
+    executorService.execute(this::logAppStartTrace);
+  }
+
+  @Override
+  public void updateSession(PerfSession session) {
+    if (session.sessionId().isEmpty()) {
```
raymondlam, line 402: When is this true that updateSession is called but the id is null:

Do we need to unregister ourshelves from session updates?

raymondlam (changes requested):
Could you add a test for when the session id is not set?

firebase-perf/src/main/java/com/google/firebase/perf/metrics/AppStartTrace.java
```diff
@@ -365,7 +370,14 @@ public synchronized void onActivityResumed(Activity activity) {
...
     onResumeTime = clock.getTime();
-    this.startSession = SessionManager.getInstance().perfSession();
+    PerfSession perfSession = SessionManager.getInstance().perfSession();
+    if (perfSession.sessionId().isEmpty()) {
+      SessionManager.getInstance().registerForSessionUpdates(sessionAwareObject);
+    } else {
+      this.startSession = SessionManager.getInstance().perfSession();
+      dispatchAppStartTrace();
+    }
+
```
bryanatkinson, line 380: Should this call `updateSession(perfSession)` instead?

bryanatkinson (approved):

PR #5447, opened 2023-10-19
Looper on a different thread for the SessionLifecycleClient's callbacks

jrothfeder:


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleClient.kt
```diff
@@ -54,13 +55,14 @@ internal object SessionLifecycleClient {
   private var serviceBound: Boolean = false
   private val queuedMessages = LinkedBlockingDeque<Message>(MAX_QUEUED_MESSAGES)
   private var curSessionId: String = ""
+  private var handlerThread: HandlerThread = HandlerThread("FirebaseSessionsClient_HandlerThread")
```
bryanatkinson, line 58: you'll need to call handlerThread.start() before creating the ClientUpdateHandler. In the service we also call stop()  in onDestroy() and so I wonder if we should stop this thread somewhere too.

jrothfeder: yeah not clear where we'd do that...

mrober (approved):

PR #5450, opened 2023-10-19
Fix changelogs after M138 mergeback.

rosalyntan:


rosalyntan:
@davidmotson @VinayGuthal should I also update the `latestReleasedVersion` in gradle.properties?

davidmotson (approved):

PR #5451, opened 2023-10-19
Add sampling, data collection, and disabled checks to newSession

mrober:


jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -110,17 +111,18 @@ internal class SessionLifecycleService() : Service() {
...
-      Log.d(TAG, "Activity foregrounding at ${msg.getWhen()}")
-      if (!hasForegrounded) {
-        Log.d(TAG, "Cold start detected.")
-        hasForegrounded = true
-        newSession()
-      } else if (isSessionRestart(msg.getWhen())) {
-        Log.d(TAG, "Session too long in background. Creating new session.")
-        newSession()
+    private fun handleForegrounding(msg: Message) =
+      CoroutineScope(Dispatchers.instance.backgroundDispatcher).launch {
```
bryanatkinson, line 115: Let's not do this on a different thread. The service guarantees thread safety by performing all the logic in the looper thread, so by moving this to the background we'd have to worry about concurrent modifications of member variables by different threads.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -142,7 +144,37 @@ internal class SessionLifecycleService() : Service() {
...
+          TAG,
+          "Sessions SDK did not have any dependent SDKs register as dependencies. Events will not be sent."
+        )
+        return
+      }
+
+      if (subscribers.values.none { it.isDataCollectionEnabled }) {
+        Log.d(TAG, "Data Collection is disabled for all subscribers. Skipping this Session Event")
+        return
+      }
```
bryanatkinson, line 161: Why do anything with subscribers from the service? It's the client's responsibility to interact with the subscribers on their containing processes.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -142,7 +144,37 @@ internal class SessionLifecycleService() : Service() {
...
+
+      Log.d(TAG, "Data Collection is enabled for at least one Subscriber")
+
+      // This will cause remote settings to be fetched if the cache is expired.
+      SessionsSettings.instance.updateSettings()
```
bryanatkinson, line 166: I think we should move this logic to the FirelogUploader and just prevent the upload there.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
```diff
@@ -196,6 +228,16 @@ internal class SessionLifecycleService() : Service() {
         sessionId.let { datastore.updateSessionId(it) }
       }
     }
+
+    /** Calculate whether we should sample events using [SessionsSettings] data. */
+    private fun shouldCollectEvents(): Boolean {
+      // Sampling rate of 1 means the SDK will send every event.
+      return randomValueForSampling <= SessionsSettings.instance.samplingRate
```
bryanatkinson, line 235: I don't think there's any need for this sampling lever since we have the enablement flag. What does it mean to sample a session? Does everything still get associated with the previous session? What does it mean for metrics? I believe we just added this to settings to give us the ability to shut things down remotely if it was too much traffic, so IMO we should either remove it altogether or move this sampling logic to the firelog uploader to just prevent upload.

bryanatkinson (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionFirelogPublisher.kt
```diff
@@ -71,8 +78,51 @@ internal class SessionFirelogPublisher(
...
+  ): Boolean {
+    if (subscribers.isEmpty()) {
+      Log.d(
+        SessionLifecycleService.TAG,
+        "Sessions SDK did not have any dependent SDKs register as dependencies. Events will not be sent."
+      )
+      return false
+    }
+
+    if (subscribers.values.none { it.isDataCollectionEnabled }) {
```
jrothfeder, line 93: why would AQS data collection be disabled on a subscriber?

mrober: We need at least one product sdk to have data collection enabled or we don't have consent to upload sessions

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
mrober: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
mrober: Done

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
mrober: Discussed in chat

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionLifecycleService.kt
mrober: Done

jrothfeder (approved):
Looks okay to me though I think we should rework this a bit. In particular, I don't think we should be querying fields on the subscriber.

PR #5452, opened 2023-10-19
Remove passthrough `register` method from `FirebaseSessions` since `F…

jrothfeder:
…irebaseSessionsDependencies` can be accessed directly.

### DIFF
--- a/firebase-crashlytics/firebase-crashlytics.gradle
+++ b/firebase-crashlytics/firebase-crashlytics.gradle
@@ -59,7 +59,6 @@ thirdPartyLicenses {
 }
 
 dependencies {
-    implementation libs.firebase.sessions
     androidTestImplementation "androidx.test.ext:junit:$androidxTestJUnitVersion"
     androidTestImplementation "androidx.test:core:$androidxTestCoreVersion"
     androidTestImplementation "com.google.truth:truth:$googleTruthVersion"
--- a/firebase-crashlytics/firebase-crashlytics.gradle
+++ b/firebase-crashlytics/firebase-crashlytics.gradle
@@ -59,6 +59,7 @@ thirdPartyLicenses {
 }
 
 dependencies {
+    implementation libs.firebase.sessions
     androidTestImplementation "androidx.test.ext:junit:$androidxTestJUnitVersion"
     androidTestImplementation "androidx.test:core:$androidxTestCoreVersion"
     androidTestImplementation "com.google.truth:truth:$googleTruthVersion"
--- a/firebase-sessions/api.txt
+++ b/firebase-sessions/api.txt
@@ -3,16 +3,50 @@ package com.google.firebase.sessions {
 
   public final class FirebaseSessions {
     method @NonNull public static com.google.firebase.sessions.FirebaseSessions getInstance();
-    method @NonNull public static com.google.firebase.sessions.FirebaseSessions getInstance(@NonNull com.google.firebase.FirebaseApp app);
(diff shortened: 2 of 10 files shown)


firebase-crashlytics/firebase-crashlytics.gradle
```diff
@@ -59,6 +59,7 @@ thirdPartyLicenses {
 }
 
 dependencies {
+    implementation libs.firebase.sessions
```
mrober, line 62: I don't think we should add sessions like this since we depend on it via `implementation(project(':firebase-sessions'))` so we can make changes to sessions and crashlytics today easily.

firebase-sessions/api.txt
```diff
@@ -3,16 +3,50 @@ package com.google.firebase.sessions {
 
   public final class FirebaseSessions {
     method @NonNull public static com.google.firebase.sessions.FirebaseSessions getInstance();
-    method @NonNull public static com.google.firebase.sessions.FirebaseSessions getInstance(@NonNull com.google.firebase.FirebaseApp app);
-    method public void register(@NonNull com.google.firebase.sessions.api.SessionSubscriber subscriber);
+    method @Deprecated @NonNull public static com.google.firebase.sessions.FirebaseSessions getInstance(@NonNull com.google.firebase.FirebaseApp app);
```
mrober, line 6: Do you think we can just remove these instance and getInstance methods? Now there is no useful public api in this class anymore. Maybe even make FirebaseSessions internal. And only the stuff in the api package be public?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/FirebaseSessions.kt
```diff
@@ -63,22 +62,6 @@ internal constructor(
...
-    FirebaseSessionsDependencies.register(subscriber)
-
-    Log.d(
-      TAG,
-      "Registering Sessions SDK subscriber with name: ${subscriber.sessionSubscriberName}, " +
-        "data collection enabled: ${subscriber.isDataCollectionEnabled}"
-    )
-  }
-
   /** Calculate whether we should sample events using [sessionSettings] data. */
```
mrober, line 65: This comment is left over from an earlier change. Can you delete it since you're already working in this file?

gradle/libs.versions.toml
```diff
@@ -59,6 +59,7 @@ kotest-runner = { module = "io.kotest:kotest-runner-junit4-jvm", version.ref = "
 kotest-assertions = { module = "io.kotest:kotest-assertions-core-jvm", version.ref = "kotest" }
 kotest-property = { module = "io.kotest:kotest-property-jvm", version.ref = "kotest" }
 quickcheck = { module = "net.java:quickcheck", version.ref = "quickcheck" }
+firebase-sessions = { group = "com.google.firebase", name = "firebase-sessions", version = "1.0.2" }
```
mrober, line 62: Here too

mrober (approved):

firebase-crashlytics/firebase-crashlytics.gradle
jrothfeder: Yeah, android studio did this for some reason and it works without it. I'm switching to vim.

firebase-sessions/api.txt
jrothfeder: I like it!

bryanatkinson (approved):

PR #5454, opened 2023-10-19
Remove AQS based session ID dependency from Fireperf.

visumickey:


raymondlam (approved):
Are we just removing the AQS session id mapping for Fireperf events but still sending AQS session events and forwarding performance's collection status?

visumickey:
> Are we just removing the AQS session id mapping for Fireperf events but still sending AQS session events and forwarding performance's collection status?

We are trying to make Fireperf not be impacted by AQS changes. Basically, for the multi process fix - we are taking an approach where Fireperf does not get impacted and then bring this change back in Fireperf. So, Fireperf will no longer get the sessionID from AQS until the multi process fix has landed.

firebase-perf/src/main/java/com/google/firebase/perf/FirebasePerfRegistrar.java
```diff
@@ -31,8 +31,6 @@
 import com.google.firebase.platforminfo.LibraryVersionComponent;
 import com.google.firebase.remoteconfig.RemoteConfigComponent;
 import com.google.firebase.sessions.FirebaseSessions;
```
mrober, line 33: Should we remove all references to `FirebaseSessions`?

visumickey: This is all cleaned up now after taking in the latest change of sessions-nine branch.

firebase-perf/CHANGELOG.md
mrober, line 1: I think this file merged slightly wrong? You're change entry should go under unreleased and 20.5.0 is already out and the Kotlin block got removed.

mrober (approved):

firebase-perf/CHANGELOG.md
visumickey: Just fixed the CHANGELOG to reflect the changes. Is the sessions-nine branch updated? I branched out of that branch.

PR #5462, opened 2023-10-20
Update latestReleasedVersion on version bump

daymxn:
Per [b/306709509](https://b.corp.google.com/issues/306709509),

This fixes an oversight in which we were not updating the `latestReleasedVersion` during our version bumps. 

Additionally, this fixes the following:

- [b/306713770](https://b.corp.google.com/issues/306713770) -> Update the `latestReleasedVersion` for libraries where it's invalid
- [b/300682493](https://b.corp.google.com/issues/300682493) -> Add condition to `moveUnreleasedChanges` task for changelog file existence

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/VersionBumpTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/VersionBumpTask.kt
@@ -24,21 +24,28 @@ import org.gradle.api.tasks.TaskAction
 import org.gradle.kotlin.dsl.provideDelegate
 
 /**
- * Bumps the `version` property of the specified [versionFile].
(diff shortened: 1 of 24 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/VersionBumpTask.kt
```diff
@@ -49,19 +56,25 @@ abstract class VersionBumpTask : DefaultTask() {
 
   @TaskAction
   fun build() {
+    project.version
```
rlazo, line 59: is this line doing something?

daymxn: Nope! Good catch, removed.

rlazo (approved):

PR #5465, opened 2023-10-20
Adds unit test for SessionLifecycleService using the fake registrar i…

bryanatkinson:
…mplementations

### DIFF
--- a/firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionLifecycleServiceTest.kt
+++ b/firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionLifecycleServiceTest.kt
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionLifecycleServiceTest.kt
```diff
@@ -0,0 +1,253 @@
...
+    service = createService()
+  }
+
+  @After
+  fun cleanUp() {
+    FirebaseApp.clearInstancesForTest()
+  }
+
+  @Test
+  @Throws(TimeoutException::class)
```
mrober, line 91: Do we need all these throws annotations?

mrober (approved):

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionLifecycleServiceTest.kt
bryanatkinson: Removed

PR #5469, opened 2023-10-23
Add common Firebase logging interface

mrober:
Add a common Firebase logging interface to make logging more consistent and make testing easier. With this we could make tests that assert specific things were logged, for example, Sessions could assert an event was logged.

### DIFF
--- a/firebase-common/src/main/java/com/google/firebase/logger/Logger.kt
+++ b/firebase-common/src/main/java/com/google/firebase/logger/Logger.kt
@@ -25,7 +25,7 @@ import java.util.concurrent.ConcurrentHashMap
  *
  * @hide
  */
-sealed class Logger private constructor(internal val tag: String) {
+sealed class Logger private constructor(val tag: String) {
   var enabled: Boolean = true
   var minLevel: Level = Level.INFO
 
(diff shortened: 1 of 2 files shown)


davidmotson (approved):

firebase-common/src/main/java/com/google/firebase/logger/Logger.kt
```diff
@@ -0,0 +1,195 @@
...
+import android.util.Log
+import androidx.annotation.VisibleForTesting
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Common logger interface that handles Android logcat logging for Firebase SDKs.
+ *
+ * @hide
+ */
+sealed class Logger
```
rlazo, line 28: While this is not part of the dev-visible public API, it's part of our inter-SDK API, and using `sealed` classes can be problematic in APIs

https://g3doc.corp.google.com/firebase/acore/g3doc/docs/kotlin_best_practices.md?cl=head#api-surface

http://goto.google.com/androidx-api-guidelines#exhaustive-when

Do we envision this class gaining more subclasses anytime soon?

mrober: I was thinking there would be logic using a `when` clause for a `FakeLogger` vs `AndroidLogger` but I ended up never needing that. So I'll remove the `sealed` since it has no purpose right now and could be problematic in the future.

PR #5470, opened 2023-10-23
Move sum&avg composite index tests to compositeIndexQueryTest.java

milaGGL:
- [x] update terraform to accept collection group index and array membership query
- [x] add new composite indexes required for testing
- [x] move sum&avg composite index tests to compositeIndexQueryTest.java(some of the field names are changed to be the same with web SDK, so that indexes in terraform are consistent)


MarkDuckworth (commented):
Same questions as web and iOS about batch writes and re-using existing field names. Otherwise, things look good.

MarkDuckworth (approved):

PR #5477, opened 2023-10-26
Make Sessions throw meaningful message when used with incompatible Perf

mrober:


firebase-sessions/src/test/kotlin/com/google/firebase/sessions/SessionLifecycleClientTest.kt
```diff
@@ -234,31 +233,31 @@ internal class SessionLifecycleClientTest {
...
+      assertThat(mattSaysHiSubscriber.sessionChangedEvents).containsExactly(SessionDetails("123"))
     }
 
   @Test
   fun handleSessionUpdate_sendsToAllSubscribersAsLongAsOneIsEnabled() =
     runTest(UnconfinedTestDispatcher()) {
       val client = SessionLifecycleClient(backgroundDispatcher() + coroutineContext)
       val crashlyticsSubscriber = addSubscriber(true, SessionSubscriber.Name.CRASHLYTICS)
-      val perfSubscriber = addSubscriber(false, SessionSubscriber.Name.PERFORMANCE)
+      val mattSaysHiSubscriber = addSubscriber(false, SessionSubscriber.Name.MATT_SAYS_HI)
```
jrothfeder, line 252: Hi Matt.

jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
```diff
@@ -39,6 +39,19 @@ object FirebaseSessionsDependencies {
...
    */
   fun addDependency(subscriberName: SessionSubscriber.Name) {
+    if (subscriberName == SessionSubscriber.Name.PERFORMANCE) {
+      throw IllegalArgumentException(
+        """
+          Incompatible versions of Firebase Perf and Firebase Sessions.
+          A safe combination would be:
+            firebase-sessions:1.1.0
+            firebase-crashlytics:18.5.0
+            firebase-perf:20.5.0
```
visumickey, line 49: Do you want the message to reflect that a particular version of a library and above or just leave it at the specified version. Leaving it at a specified version is going to need to update this comment in the near future.

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/api/FirebaseSessionsDependencies.kt
mrober: I put this for now because I don't know what will happen in future versions. I expect we will make it compatible again, so maybe only this version will have this message. If you have a better idea for wording that reflects that we can work on the message.

PR #5484, opened 2023-10-27
Remove our own ImmutableList and use similar built-in classes.

jrothfeder:
We don't wish to include guava, however, this approach of implementing our own version of ImmutableList is going to surprise most developers who expect ImmutableList to be something else.


firebase-crashlytics/src/androidTest/java/com/google/firebase/crashlytics/internal/persistence/CrashlyticsReportPersistenceTest.java
```diff
@@ -186,11 +187,14 @@ public void testLoadFinalizedReports_reportThenMultipleEvents_returnsReportWithM
         reportPersistence.loadFinalizedReports();
     assertEquals(1, finalizedReports.size());
     final CrashlyticsReport finalizedReport = finalizedReports.get(0).getReport();
+    ArrayList<Event> events = new ArrayList<>();
```
mrober, line 190: Can you say like new ArrayList<>({ testEvent, testEvent2 });

mrober (approved):

bryanatkinson (approved):

PR #5485, opened 2023-10-27
Add support for productId in transport

rlazo:
Optionally allow logging the productId together with the event in transport.

For googlers b/303136538


transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/ExternalPRequestContext.java
```diff
@@ -0,0 +1,25 @@
+package com.google.android.datatransport.cct.internal;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class ExternalPRequestContext {
```
emilypgoogle, line 8: `ExternalPRequestContext` is an odd name, why abbreviate just the P?

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/ExternalPrivacyContext.java
```diff
@@ -0,0 +1,25 @@
+package com.google.android.datatransport.cct.internal;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class ExternalPrivacyContext {
+  @Nullable
+  public abstract ExternalPRequestContext getPrequest();
```
emilypgoogle, line 10: This and the setter should probably both be `getPRequest` if we're sticking with that name.

emilypgoogle (approved):
Just a naming crit on PRequest, otherwise looks good

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/ExternalPRequestContext.java
rlazo: Discussed offline

transport/transport-backend-cct/src/main/java/com/google/android/datatransport/cct/internal/ExternalPrivacyContext.java
rlazo: Discussed offline

PR #5489, opened 2023-10-30
Adds additional integration tests for firebase-sessions

bryanatkinson:


jrothfeder (approved):

firebase-sessions/test-app/src/main/kotlin/com/google/firebase/testing/sessions/BaseActivity.kt
```diff
@@ -24,7 +24,8 @@ import android.util.Log
 import androidx.appcompat.app.AppCompatActivity
 import com.google.firebase.FirebaseApp
 
-/**  */
+/**
```
mrober, line 27: I think it's better just to delete these empty kotlindocs

mrober (approved):

PR #5494, opened 2023-10-31
Make data store pref configs multi-process safe

mrober:
Adds the process name to the config name. This is so two processes don't try to access the same config on the same file at the same time, possibly corrupting data. Data store preferences 1.1.0 will support multi-process. So when that becomes stable, we can upgrade to it and remove this util. Then all processes will share the same preferences. This will mostly only affect settings since it will be fetched on each process, while the sessions store will only happen on the service. But I made that safe too.


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataStoreConfigs.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.google.firebase.sessions.settings.SessionsSettings
+
+/**
+ * Util object for handling DataStore<Preferences> configs in multi-process apps safely.
+ *
+ * This can be removed when datastore-preferences:1.1.0 becomes stable.
+ */
+object SessionDataStoreConfigs {
+  /** Sanitized process name to use in config names. */
+  private val PROCESS_NAME = getProcessName()?.replace("\\W".toRegex(), "_") ?: ""
```
visumickey, line 32: We need some good testing behind this.

visumickey (commented):
Why would we have a preferences store once per process? With a service deciding on the sessionID generation and maintenance - do we have a need for every process owning their own references?

mrober:
> Why would we have a preferences store once per process? With a service deciding on the sessionID generation and maintenance - do we have a need for every process owning their own references?

Every process at least needs settings, to know to initialize or not. The sessions data store should only be accessed from the service but I made it safe too just to be consistent and in case stuff gets moved around, it'll still be safe.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataStoreConfigs.kt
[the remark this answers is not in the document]
mrober: Settings will fetch and cache for every process because it's needed before binding the service to decide of Sessions is enabled. This is this way to be process safe until data store preferences 1.1.0 becomes stable, then we can remove all this.

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataStoreConfigs.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Util object for handling DataStore<Preferences> configs in multi-process apps safely.
+ *
+ * This can be removed when datastore-preferences:1.1.0 becomes stable.
+ */
+object SessionDataStoreConfigs {
+  /** Sanitized process name to use in config filenames. */
+  private val PROCESS_NAME =
+    Base64.encodeToString(
+      ProcessDetailsProvider.getProcessName().encodeToByteArray(),
+      Base64.NO_WRAP or Base64.URL_SAFE, // URL safe is also filename safe.
```
bryanatkinson, line 32: Isn't a '/' url safe, but not filename safe? I don't think you can put that in a process name, but I wonder if we should wrap this reading/writing with a dynamic filename in a try catch block to make sure it doesn't cause the app to crash if it's poorly formed, or too long.

bryanatkinson (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionDataStoreConfigs.kt
mrober: I checked the doc on `URL_SAFE` it specifically says "URL and filename safe ... where - and _ are used in place of + and /" so it is safe.

PR #5497, opened 2023-10-31
Adds instrumentation tests to trigger crashes, non fatals and ANRs on…

bryanatkinson:
… multiple app processes.


firebase-sessions/test-app/src/androidTest/kotlin/com/google/firebase/testing/sessions/FirebaseSessionsIntegrationTest.kt
```diff
@@ -0,0 +1,98 @@
+package com.google.firebase.testing.sessions
```
mrober, line 1: Copyright

mrober (approved):

visumickey (approved):
These changes look good to me. Is there a way we can trigger background launch of the app and later foreground to see if we are using the older sessionID until launch and generate a new sessionID after launch?

jrothfeder (approved):

PR #5501, opened 2023-11-01
Add process details to AQS event.

jrothfeder:
also see https://github.com/firebase/firebase-android-sdk/pull/5445


firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetailsProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+    val pid = Process.myPid()
+    return getAppProcessDetails(context).find { it.pid == pid }
+      ?: buildProcessDetails(getProcessName(), pid)
+  }
+
+  /** Builds a ProcessDetails object. */
+  fun buildProcessDetails(
+    processName: String,
+    pid: Int = 0,
+    importance: Int = 0,
```
visumickey, line 69: NIT: Can we use ENUMS here instead of integer values? (if one such ENUMs exist).

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetailsProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+    isDefaultProcess: Boolean = false
+  ) = ProcessDetails(processName, pid, importance, isDefaultProcess)
+
+  /** Gets the app's current process name. If the API is not available, returns an empty string. */
+  private fun getProcessName(): String =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+      Process.myProcessName()
+    } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+      Application.getProcessName() ?: ""
```
visumickey, line 78: This code could throw warnings if some APIs are not supposed to be used (eg: deprecated) on a particular version. If  that being the case, recommend encapsulating the code to ignore compile time warnings.

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/ProcessDetailsProviderTest.kt
```diff
@@ -0,0 +1,59 @@
...
+  @Test
+  fun getCurrentProcessDetails() {
+    val processDetails =
+      ProcessDetailsProvider.getCurrentProcessDetails(firebaseApp.applicationContext)
+    assertThat(processDetails)
+      .isEqualTo(ProcessDetails("com.google.firebase.sessions.test", 0, 100, false))
+  }
+
+  @Test
+  fun getAppProcessDetails() {
```
visumickey, line 54: Should there be scenarios where there are multiple processes for testing? Not sure if it is going to add any value with the current code, but though the code would be future proof for changes anyone wants to bring to the ProcessDetailsProvider class.

visumickey (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetailsProvider.kt
jrothfeder: Not sure what you mean. Can you give an example?

firebase-sessions/src/test/kotlin/com/google/firebase/sessions/ProcessDetailsProviderTest.kt
jrothfeder: I don't believe in future proofing :)

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetailsProvider.kt
jrothfeder: encapsulate? In what way?

visumickey: Just an example on how this was used in sessions: https://github.com/firebase/firebase-android-sdk/blob/master/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/SessionEvents.kt#L66

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetailsProvider.kt
visumickey: Instead of using defaults values as `0`, I was thinking if we can have the default values as `FOREGROUND_IMPORTANCE`.

mrober (approved):
LGTM but should this merge into sessions-nine not master?

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/ProcessDetailsProvider.kt
jrothfeder: it's not under our control. https://developer.android.com/reference/android/app/ActivityManager.RunningAppProcessInfo

PR #5506, opened 2023-11-02
Fix docs unnecessarily undergo limbo resolution while resuming query bug

milaGGL:
Ported from : https://github.com/firebase/firebase-js-sdk/pull/7740


dconeybe (changes requested):
Oh, please add a change log entry.

dconeybe (approved):

PR #5508, opened 2023-11-02
Adds support for skipping tests that actually crash or ANR the test a…

bryanatkinson:
…pp during integration testing. These will still run with the scheduled 4h job.

### DIFF
--- a/firebase-sessions/test-app/test-app.gradle.kts
+++ b/firebase-sessions/test-app/test-app.gradle.kts
@@ -51,7 +51,7 @@ android {
       buildConfigField(
         "boolean",
         "SHOULD_CRASH_APP",
-        if (project.hasProperty("useReleasedVersions")) "true" else "false"
+        project.hasProperty("useReleasedVersions").toString()
       )
     }
     debug {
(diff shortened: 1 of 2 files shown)


firebase-sessions/test-app/test-app.gradle.kts
```diff
@@ -45,6 +45,24 @@ android {
   }
   kotlinOptions { jvmTarget = "1.8" }
   buildFeatures { viewBinding = true }
+  buildTypes {
+    release {
```
mrober, line 49: This can do in the defaultConfig instead of in both debug and release.

firebase-sessions/test-app/test-app.gradle.kts
```diff
@@ -45,6 +45,24 @@ android {
...
+      // We only want to actually crash the app for the scheduled runs, not the integration tests
+      buildConfigField(
+        "boolean",
+        "SHOULD_CRASH_APP",
+        if (project.hasProperty("useReleasedVersions")) "true" else "false"
```
mrober, line 54: This should already be a non-null boolean. Can you call toString or something?

mrober (approved):

firebase-sessions/test-app/test-app.gradle.kts
bryanatkinson: Done

PR #5600, opened 2023-12-14
Firestore: add TODO comments about removing special 'missing index' test logic for multidb

dconeybe:
Add a comment to Firestore's integration tests about removing special logic when testing for the "missing index" error message. Formerly, the "missing index" error message was different for the default database and a non-default database; however, their logic has been unified, and will be rolled out to production soon. Once rolled out, these tests should be updated to remove the special logic for non-default dbs.

Also, remove the "else" branch that verified a different error message when a non-default db was used because when that change rolls out that "else" block will start to fail, seemingly without reason. This also brings the test into parity with the js sdk.

A similar change was made to the js sdk in https://github.com/firebase/firebase-js-sdk/pull/7874.

Googlers see b/316359394 for details.


milaGGL (approved):
LGTM

Issue #5584, opened 2023-12-01
Initializing the SDK programmatically by using FirebaseOptions cause crash when call URLConnection.getInputStream()


yin-haoran:
Calling `URLConnection.getInputStream()` before `FirebaseApp.initializeApp()` will lead crash. Maybe due to the `Bytecode instrumented functions` `com.google.firebase.perf.network.FirebasePerfUrlConnection#openStream(java.net.URL)`.

stack info as follow:
```
java.lang.ExceptionInInitializerError
	at com.google.firebase.perf.config.RemoteConfigManager.getInstance(RemoteConfigManager.java:119)
	at com.google.firebase.perf.config.ConfigResolver.<init>(ConfigResolver.java:78)
	at com.google.firebase.perf.config.ConfigResolver.getInstance(ConfigResolver.java:86)
	at com.google.firebase.perf.application.AppStateMonitor.<init>(AppStateMonitor.java:98)
	at com.google.firebase.perf.application.AppStateMonitor.getInstance(AppStateMonitor.java:87)
	at com.google.firebase.perf.metrics.NetworkRequestMetricBuilder.<init>(NetworkRequestMetricBuilder.java:92)
	at com.google.firebase.perf.metrics.NetworkRequestMetricBuilder.builder(NetworkRequestMetricBuilder.java:84)
	at com.google.firebase.perf.network.FirebasePerfUrlConnection.openStream(FirebasePerfUrlConnection.java:63)
	at com.google.firebase.perf.network.FirebasePerfUrlConnection.openStream(FirebasePerfUrlConnection.java:47)
	at com.tony.firebasetest.MainActivity.onCreate(MainActivity.kt:38)
	at android.app.Activity.performCreate(Activity.java:8000)
	at android.app.Activity.performCreate(Activity.java:7984)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1310)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3409)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3600)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
```

I provide a [example](https://github.com/firebase/firebase-android-sdk/files/13522241/firebaseTest.zip) of this issue.

yin-haoran:
Hi @argzdev, could you see this issue please.

argzdev:
[...]

PR #5616, opened 2023-12-27
openStream should not be intercepted if Firebase Perf is not initialized

argzdev:
Proposed fix for #5584:

### Issue
Crash happens if a URL connection is opened e.g. `url.openStream()` when `TransportManager` has not been initialized.

### Investigation
1. `openStream` in `FirebasePerfUrlConnection` intercepts all urls that opens connections. 
2. `openStream` requires `TransportManager` to be initialized
3. `TransportManager` is only initialized after `FirebasePerformance` has been initialized

### Solution
If `TransportManager` is not initialized, we should simply return the URL openStream and not intercept the connection.


visumickey (approved):
This change looks good to me.

But this issue needs a more broader look to see if we can capture these network requests as well (may be initialize Transport Manager early). But that is a bigger refactor. So, this change looks good for now.

argzdev:
PR seems to be failing due to:
- App Distribution release notes needed (not sure why this is being triggered)
- `firebase-sessions:test-app:copyRootGoogleServices` failing to copy the file which causes the task for `:firebase-sessions:test-app:processDebugGoogleServices` return a `Malformed root json` error.

PR #5623, opened 2024-01-03
Add diffing for ABT Experiments in real-time RC

qdpham13:
Adds diffing logic between experiments fetched via real-time and experiments active on the device. The config keys that have been affected by the differences will be added to ConfigUpdate. Testing below:

Created experiments with https://docs.google.com/document/d/1MXxUg1JSBbh3m_Qy4tGPrGs3-Nh8FAwKmg5ZWOKQ3J8/edit

Add experiments to configs using Management API, example: https://rpc.corp.google.com/rpc/builder?rpcId=15352346049097272373

2.1) Add/Remove Experiment to Config Key.
2.2) Replace experiment in Config Key
2.3) Create new param with experiment attached

See changes in SDK

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
@@ -14,6 +14,9 @@
 
 package com.google.firebase.remoteconfig.internal;
 
+import static com.google.firebase.remoteconfig.RemoteConfigConstants.ExperimentDescriptionFieldKey.AFFECTED_PARAMETER_KEYS;
+import static com.google.firebase.remoteconfig.RemoteConfigConstants.ExperimentDescriptionFieldKey.EXPERIMENT_ID;
+
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -37,6 +40,7 @@ public class ConfigContainer {
   static final String PERSONALIZATION_METADATA_KEY = "personalization_metadata_key";
   static final String TEMPLATE_VERSION_NUMBER_KEY = "template_version_number_key";
   static final String ROLLOUT_METADATA_KEY = "rollout_metadata_key";
+  static final String ROLLOUT_ID_PREFIX = "rollout";
   public static final String ROLLOUT_METADATA_AFFECTED_KEYS = "affectedParameterKeys";
   public static final String ROLLOUT_METADATA_ID = "rolloutId";
   public static final String ROLLOUT_METADATA_VARIANT_ID = "variantId";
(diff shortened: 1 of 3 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
```diff
@@ -189,6 +193,38 @@ public boolean equals(Object o) {
...
+        continue;
+      }
+
+      // Since a config key can only have one experiment associated with it, map the key to the
+      // experiment.
+      JSONArray affectedKeys = experiment.getJSONArray(AFFECTED_PARAMETER_KEYS);
+      for (int j = 0; j < affectedKeys.length(); j++) {
+        String key = affectedKeys.getString(j);
+        JSONObject experimentsCopy = new JSONObject(experiment.toString());
+        // Removing `affectedParameterKeys` because its values never come in the same order which
```
danasilver, line 218: thinking: and because we're diffing the metadata w.r.t. the key(s) itself so we don't need to diff the key

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
```diff
@@ -189,6 +193,38 @@ public boolean equals(Object o) {
     return containerJson.toString().equals(that.toString());
   }
 
+  /** Creates a map where the key is the config key and the value if the experiment description. */
```
danasilver, line 196:
```suggestion
  /** Creates a map where the key is the config key and the value is the experiment description. */
```

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
```diff
@@ -189,6 +193,38 @@ public boolean equals(Object o) {
...
+  private Map<String, JSONObject> createExperimentsMap(JSONArray allExperiments)
```
danasilver, line 197: I believe this could be `static`? Though alternatively it could be nice to make it have a similar signature to `createRolloutParameterKeyMap` work as an instance method so we call `this.createExperimentsMap()` and `other.createExperimentsMap()`

firebase-config/src/test/java/com/google/firebase/remoteconfig/internal/ConfigContainerTest.java
```diff
@@ -138,23 +139,99 @@ public void getChangedParams_sameP13nMetadata_returnsEmptySet() throws Exception
   }
 
   @Test
```
danasilver, line 141: Nice tests for this

danasilver (commented):
Will want to add a changelog entry too, but can happen in a followup PR

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
qdpham13: updated to just use instance vars

firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigContainer.java
qdpham13: yeah that makes sense, removed this

PR #5688, opened 2024-02-01
Disconnect from RC real-time server when app is backgrounded.

danasilver:
Store the Realtime client's `HttpUrlConnection` reference on the class instance so it can used by another client instance method (`closeRealtimeHttpStream`) to disconnect when the app is backgrounded.

The rest of the changes are follow on work to handle network errors thrown when `HttpUrlConnection#disconnect` is called. Since HttpUrlConnection enforces a [strict lifecycle](https://android.googlesource.com/platform/external/okhttp/+/602d5e4/okhttp/src/main/java/com/squareup/okhttp/internal/http/HttpConnection.java#43) and this code (1) interacts with the connection in several methods/classes and (2) holds open the connection for a long time, it's likely randomly disconnecting when backgrounded will occur at an inopportune lifecycle state.

This change inserts a few conditionals to ensure we only call once-only network methods once, and error handling to catch expected errors when we disconnect intentionally.

Generally, the code could be refactored to consolidate network handling into one class, which would simplify the state that needs to be passed around and places errors must be handled.

### DIFF
--- a/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
+++ b/firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
@@ -498,7 +498,7 @@ public class ConfigRealtimeHttpClient {
             this.scheduledExecutorService,
             (completedHttpUrlConnectionTask) -> {
               Integer responseCode = null;
-              this.httpURLConnection = null;
+              httpURLConnection = null;
 
               try {
                 // If HTTP connection task failed throw exception to move to the catch block.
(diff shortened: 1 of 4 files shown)


firebase-config/src/main/java/com/google/firebase/remoteconfig/internal/ConfigRealtimeHttpClient.java
```diff
@@ -293,6 +293,15 @@ private synchronized boolean canMakeHttpStreamConnection() {
         && !isInBackground;
   }
 
+  private synchronized boolean checkAndSetHttpConnectionFlagIfNotRunning() {
```
danasilver, line 296: @qdpham13 could you add a comment with why this is combined?

qdpham13 (approved):
LGTM

danasilver:
Will add a changelog entry in a follow up PR

PR #5699, opened 2024-02-08
fix: Make ObjectValue thread safe.

tom-andersen:


firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
```diff
@@ -206,18 +237,20 @@ private void setOverlay(FieldPath path, @Nullable Value value) {
...
-  private @Nullable MapValue applyOverlay(
-      FieldPath currentPath, Map<String, Object> currentOverlays) {
+  private static @Nullable MapValue applyOverlay(
+      FieldPath currentPath, Map<String, Object> currentOverlays, Value currentValue) {
+    if (currentOverlays.isEmpty()) {
+      return null;
+    }
     boolean modified = false;
-
-    @Nullable Value existingValue = extractNestedValue(partialValue, currentPath);
```
tom-andersen, line 215: Adding currentValue as parameter, such that this method can be static. This helps reason about side effects, and also eliminates redundant work of extracting existing value from root.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
```diff
@@ -153,16 +170,18 @@ public void delete(FieldPath path) {
...
 
   public void setAll(Map<FieldPath, Value> data) {
-    for (Map.Entry<FieldPath, Value> entry : data.entrySet()) {
-      FieldPath path = entry.getKey();
-      if (entry.getValue() == null) {
-        delete(path);
-      } else {
-        set(path, entry.getValue());
+    synchronized (overlayMap) {
+      hardAssert(
```
MarkDuckworth, line 180: Was it intentional to include the `hardAssert` within the synchronized block?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
```diff
@@ -103,32 +109,41 @@ private FieldMask extractFieldMask(MapValue value) {
...
           return null;
         }
       }
-      return value.getMapValue().getFieldsOrDefault(fieldPath.getLastSegment(), null);
+      return extractSegmentValue(value, fieldPath.getLastSegment());
     }
   }
 
+  private static Value extractSegmentValue(Value value, String segment) {
+    return Values.isMapValue(value) ? value.getMapValue().getFieldsOrDefault(segment, null) : null;
```
MarkDuckworth, line 127: Was it intentional to invert the logic for extracting a value from a MapValue? Previously the code ran `Values.isMapValue(value)` on the `value` contained in the input map and returned `null` if it was not a map. Now that check is not performed on the extracted value. Seems like this could change what this method returns?

firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
```diff
@@ -206,18 +237,20 @@ private void setOverlay(FieldPath path, @Nullable Value value) {
...
    *     overlayMap}.
+   * @param currentValue The existing value at the current nesting level.
    * @return The merged data at `currentPath` or null if no modifications were applied.
    */
-  private @Nullable MapValue applyOverlay(
-      FieldPath currentPath, Map<String, Object> currentOverlays) {
+  private static @Nullable MapValue applyOverlay(
+      FieldPath currentPath, Map<String, Object> currentOverlays, Value currentValue) {
+    if (currentOverlays.isEmpty()) {
+      return null;
```
MarkDuckworth, line 246: Should this return `currentValue`?

firebase-firestore/src/androidTest/java/com/google/firebase/firestore/ServerTimestampTest.java
MarkDuckworth, line 1: These tests look like they may be related to other work

MarkDuckworth (commented):
@tom-andersen, I have a few questions. But overall the direction looks good to me

firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
tom-andersen: Good point. I will move it out!

firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
tom-andersen: No. The method description states:

```
@return The merged data at `currentPath` or null if no modifications were applied.
```

This basically just short circuiting logic before more heavy lifting of creating builder objects is done.

firebase-firestore/src/main/java/com/google/firebase/firestore/model/ObjectValue.java
tom-andersen: Semantically, this should be equivalent. The helper function is used in other contexts where value might be `null`, so I had to add check. Since this is being done in helper, it seemed like overkill to also do it in the loop. Technically the loop will work fine even without early return, but the short circuiting behavior has nice performance characteristics.

It isn't the easiest code to read, so any recommendations are welcome.

MarkDuckworth: I think the new code already is more readable. So thanks for the improvement, and clarifying comment. I don't have any recommendations for improvement, but consider a unit test if you have any concerns about equivalency.

MarkDuckworth (approved):

PR #5710, opened 2024-02-14
Get parameter value for rollout without side effects.

danasilver:
The `getString(String parameterKey)` method calls listeners as a side-effect, which isn't the behavior we want when `activate()` is called and Rollouts are published.

This duplicates the `getString` method without the side-effects (as well as a helper method to access the cache). Alternatively we could duplicate this method _within_ `ConfigGetParameterHandler`, but it'd have to be `public` which implies other callers should use it (they shouldn't - there's no other use case besides Rollouts).

Google internal: b/324275156

### DIFF
--- a/firebase-config/CHANGELOG.md
+++ b/firebase-config/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-* [fixed] Fixed an issue that could cause Remote Config personalizations to be logged early in
+* [fixed] Fixed an issue that could cause [remote_config] personalizations to be logged early in
   specific cases.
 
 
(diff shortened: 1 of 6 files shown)


danasilver:
Failed smoke tests seem unrelated. They're from Firestore with the error: `This project's quota for this operation has been exceeded. [ Exceeded quota for verifying passwords. ]` (https://firebase.corp.google.com/project/fireescape-c4819/testlab/histories/bh.150fab898fb191d1/matrices/7923701961509746284/executions/bs.832a0e2259928519/test-cases)

karenyz (approved):
good catch!! 🙏

PR #5714, opened 2024-02-14
Existing logic seems to work fine. Why is this change required?

tom-andersen:


firebase-firestore/src/main/java/com/google/firebase/firestore/util/CustomClassMapper.java
```diff
@@ -1059,7 +1059,10 @@ private static boolean shouldIncludeSetter(Method method) {
       if (method.getParameterTypes().length != 1) {
         return false;
       }
-
+      // Excluded methods
```
tom-andersen, line 1062: This was removed in other PR. So I added it back.

PR #5715, opened 2024-02-14
Implement release notes toggle

daymxn:
Per [b/324410387](https://b.corp.google.com/issues/324410387),

This implements a new property on `FirebaseLibrary` for toggling the generation of release notes. Previously, we were using the presence of Javadocs to determine this- as libraries without Javadocs don't publish release notes... at least that's what we thought. We have since realized that we have multiple libraries that publish release notes, but do *not* publish Javadocs. To accommodate these libraries, this new property can be toggled to disable release note generation.

This PR also adds appdistro to its librarygroup, as it seems that is the intended behavior.

NO_RELEASE_CHANGE


firebase-appdistribution/firebase-appdistribution.gradle
```diff
@@ -18,7 +18,9 @@ plugins {
 }
 
 firebaseLibrary {
+    libraryGroup "appdistribution"
```
daymxn, line 21: cc: @davidmotson 

Was there a reason appdistro _wasn't_ in its library group, or was it just an oversight?

rlazo: most likely an oversight

firebase-appdistribution/firebase-appdistribution.gradle
```diff
@@ -18,7 +18,9 @@ plugins {
...
     publishJavadoc = false
+    publishReleaseNotes = true
```
rlazo, line 23: we don't add this to other `true` sdks, remove?

firebase-components/firebase-dynamic-module-support/firebase-dynamic-module-support.gradle.kts
```diff
@@ -22,6 +22,7 @@ firebaseLibrary {
   testLab.enabled = false
   publishSources = true
   publishJavadoc = false
+  publishReleaseNotes = false
```
rlazo, line 25: https://firebase.google.com/support/release-notes/android?_gl=1*ab5rhc*_up*MQ..*_ga*NzA5NzI4OS4xNzA3OTQyNzMw*_ga_CW55HF8NVT*MTcwNzk0MjczMC4xLjAuMTcwNzk0MjczMC4wLjAuMA..#dynamic-feature-modules-support_v16-0-0-beta03

rlazo (approved):

rlazo (commented):
didn't mean to approve just yet

rlazo (changes requested):
see comment about dynamic module support

firebase-appdistribution/firebase-appdistribution.gradle
daymxn: Good catch, fixed.

firebase-components/firebase-dynamic-module-support/firebase-dynamic-module-support.gradle.kts
daymxn: Were these being made manually? Why is there nothing in the changelog file?

rlazo: Yes, because:


Last Updated Date | 9/1/2022 ([maven](https://maven.google.com/web/index.html?q=dynamic#com.google.firebase:firebase-dynamic-module-support:16.0.0-beta03))


So it's being a while since we made a release of it

rlazo (approved):

PR #5716, opened 2024-02-15
Add auth token to sessions

mrober:
This adds the auth token to sessions, which forced the client to validate the fid, as well as allow us to validate it on the backend.

I will make a similar change in Crashlytics in the next PR.

### DIFF
--- a/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/InstallationId.kt
+++ b/firebase-sessions/src/main/kotlin/com/google/firebase/sessions/InstallationId.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


jrothfeder (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/InstallationId.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    suspend fun create(firebaseInstallations: FirebaseInstallationsApi) =
+      try {
+        // Fetch the auth token first, so the fid will be validated.
+        val authToken = firebaseInstallations.getToken(false).await().token
+        val fid = firebaseInstallations.id.await()
+        InstallationId(fid, authToken)
+      } catch (ex: Exception) {
+        Log.w(TAG, "Error getting Firebase installation id or authentication token.", ex)
+        // If there are any failures, the fid is not validated. So return empty values for both.
```
themiswang, line 36: On iOS if token is not valid we still pass fid, token will be empty.

themiswang (approved):

firebase-sessions/src/main/kotlin/com/google/firebase/sessions/InstallationId.kt
mrober: Done

themiswang (approved):

PR #5737, opened 2024-02-23
Remove non link brackets from release notes

daymxn:
Per [b/326629027](https://b.corp.google.com/issues/326629027),

This adds a transformation in our release notes to separate brackets from their content if they're not wrapping a link. This is to avoid propagating the content to the release notes, as it shouldn't be there. In the future, we may want to add some warning for instances found (as they may have been intended to be links).

### DIFF
--- a/buildSrc/src/main/java/com/google/firebase/gradle/plugins/MakeReleaseNotesTask.kt
+++ b/buildSrc/src/main/java/com/google/firebase/gradle/plugins/MakeReleaseNotesTask.kt
@@ -162,10 +162,12 @@ abstract class MakeReleaseNotesTask : DefaultTask() {
     if (message.isBlank()) throw RuntimeException("A changelog entry message can not be blank.")
 
     val fixedMessage =
-      LINK_REGEX.replace(message) {
-        val id = it.firstCapturedValue
-        "GitHub [#$id](//github.com/firebase/firebase-android-sdk/issues/$id){: .external}"
-      }
+      message
+        .replace(LINK_REGEX) {
+          val id = it.firstCapturedValue
+          "GitHub [#$id](//github.com/firebase/firebase-android-sdk/issues/$id){: .external}"
+        }
+        .replace(NON_LINK_BRACKETS_REGEX) { it.firstCapturedValue }
 
     return "* {{${type.name.toLowerCase()}}} $fixedMessage"
   }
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/google/firebase/gradle/plugins/MakeReleaseNotesTask.kt
```diff
@@ -213,5 +215,29 @@ abstract class MakeReleaseNotesTask : DefaultTask() {
...
+     * Will match the following:
+     * ```kotlin
+     * [
+     *   "[release_config]"
+     * ]
+     * ```
+     *
+     * @see [Change.toReleaseNote]
+     */
+    private val NON_LINK_BRACKETS_REGEX = Regex("\\[(.+?)](?!\\()", RegexOption.MULTILINE)
```
emilypgoogle, line 241: This regex will match `[foo](bar.baz) lorem ipsum [qud]` and could be adjusted to not match `]` characters outside of the terminating one.

emilypgoogle (changes requested):

rlazo:
In (most)? cases, what we want is the reference to be transformed from [[app_check]] to {{app_check}}, see https://github.com/firebase/firebase-android-sdk/blob/master/appcheck/firebase-appcheck-debug/CHANGELOG.md and cl/605317607 rather than removed

rlazo:
@rachelsaunders does tooling break or complains if an unknown value is present inside the double brackets? e.g. {{notaname}}, does it complain/break or silently drops it?

rachelsaunders:
> @rachelsaunders does tooling break or complains if an unknown value is present inside the double brackets? e.g. {{notaname}}, does it complain/break or silently drops it?


If the contents within the double curly braces is not a valid string, our internal staging/build tooling will error (and maybe even break the build).  It won't just silently drop it.  So, if the string happens to be wrong here in the changelog and thus the conversion into the internal systems, we'll be able to easily catch it downstream _before_ it goes to prod.

rlazo:
> > @rachelsaunders does tooling break or complains if an unknown value is present inside the double brackets? e.g. {{notaname}}, does it complain/break or silently drops it?
> 
> If the contents within the double curly braces is not a valid string, our internal staging/build tooling will error (and maybe even break the build). It won't just silently drop it. So, if the string happens to be wrong here in the changelog and thus the conversion into the internal systems, we'll be able to easily catch it downstream _before_ it goes to prod.

That's what I was hoping for, thanks!

